PR #1012, opened 2020-04-28
Refactor Custom Secrets and AzureSQLUser

### LINKED ISSUE #887: Task: refactor sql user secret formatting
Any secret formatting that exists for Keyvault should also exist for kube. The Operators should not know the type of secret store being used.

The secret formatting is also a bit cumbersome and messy for the Ensure method. It should be pushed into its own function and called from Ensure.

The flatten option should be removed from the Keyvault client and the sql user Ensure method should instead create a list of secrets that get sent to the secret client. Feel free to add methods to the secret client for creating multiple secrets or just call the existing methods in a loop from the operator.

**Done criteria**
- no conditional functionality based on type of secret store in sql user operator
- function or functions for creating the formatted secrets
- no magical implicit functionality in Upsert...do that elsewhere

dkisselev:
Secrets Pkg:
* Return error when trying to use flatten on kube client (+test)

AzureSqlUser:
* Refactor and move custom secret formatting and reconciliation into its own function (out of Ensure)
* Clean up and standardize NamespacedName generation
* Clean up overall azuresqluser variable naming for code clarity

The flatten option has been kept in the implementation because we need to have a way to ask the secrets package to create individual secrets rather than the default behavior of working with a map/json array. The kube client now returns an error when asked to flatten because that doesn't make sense in-context (Kube only works with map objects for secrets).

Closes #887

gif

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/pkg/resourcemanager/azuresql/azuresqluser/azuresqluser.go
+++ b/pkg/resourcemanager/azuresql/azuresqluser/azuresqluser.go
@@ -130,8 +130,8 @@ func (s *AzureSqlUserManager) CreateUser(ctx context.Context, secret map[string]
 	return newUser, nil
 }
 
-// UpdateUser - Updates user password
-func (s *AzureSqlUserManager) UpdateUser(ctx context.Context, secret map[string][]byte, db *sql.DB) error {
+// RotateUserPassword - Generate and save a new password for the user
+func (s *AzureSqlUserManager) RotateUserPassword(ctx context.Context, secret map[string][]byte, db *sql.DB) error {
 	user := string(secret[SecretUsernameKey])
 	newPassword := helpers.NewPassword()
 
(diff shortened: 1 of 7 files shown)


pkg/resourcemanager/azuresql/azuresqluser/azuresqluser.go
```diff
@@ -130,8 +130,8 @@ func (s *AzureSqlUserManager) CreateUser(ctx context.Context, secret map[string]
 	return newUser, nil
 }
 
-// UpdateUser - Updates user password
-func (s *AzureSqlUserManager) UpdateUser(ctx context.Context, secret map[string][]byte, db *sql.DB) error {
+// RotateUserPassword - Generate and save a new password for the user
+func (s *AzureSqlUserManager) RotateUserPassword(ctx context.Context, secret map[string][]byte, db *sql.DB) error {
```
WilliamMortlMicrosoft, line 134: This function should be defined in the manager as well. I know that UpdateUser previously wasn't, but can you add that chance?

pkg/resourcemanager/azuresql/azuresqluser/azuresqluser_reconcile.go
```diff
@@ -149,134 +144,34 @@ func (s *AzureSqlUserManager) Ensure(ctx context.Context, obj runtime.Object, op
...
-			secrets.Flatten(true),
-		)
-		if err != nil {
-			return false, err
-		}
+	// Reconcile any requested custom secrets
+	err = s.ReconcileCustomSecrets(ctx, instance, sqlUserSecretClient, userSecret)
+	if err != nil {
+		instance.Status.Message = "failed to update custom secrets, err: " + err.Error()
+		return false, err
```
WilliamMortlMicrosoft, line 151: delete this

pkg/secrets/kube/client.go
```diff
@@ -31,6 +31,10 @@ func (k *KubeSecretClient) Create(ctx context.Context, key types.NamespacedName,
 		opt(options)
 	}
 
+	if options.Flatten {
+		return fmt.Errorf("kube secret client does not support flattened secrets")
+	}
```
WilliamMortlMicrosoft, line 36: delete

pkg/secrets/kube/client_test.go
```diff
@@ -50,6 +51,11 @@ var _ = Describe("Kube Secrets Client", func() {
 				Expect(err).To(BeNil())
 			})
 
+			Context("creating secret with secret and flatten enabled", func() {
+				err = client.Create(ctx, key, data, secrets.Flatten(true))
+				Expect(err).ToNot(BeNil())
+			})
+
```
WilliamMortlMicrosoft, line 58: delete

PR #1054, opened 2020-05-14
Add VM Extension Operator

huangpf:
This PR adds a VM extension operator for one of the crucial Compute use cases.

^-^

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -47,7 +47,7 @@ RUN apt-get update \
         golang.org/x/lint/golint \
         github.com/alecthomas/gometalinter \
         honnef.co/go/tools/... \
-        github.com/golangci/golangci-lint/cmd/golangci-lint \
+        #github.com/golangci/golangci-lint/cmd/golangci-lint \
         github.com/mgechev/revive \
         github.com/derekparker/delve/cmd/dlv 2>&1 \
     #
@@ -93,7 +93,7 @@ COPY ./Makefile ./
 RUN make install-kind
 RUN make install-kubebuilder
 RUN make install-kustomize
-RUN make install-test-dependency
+# RUN make install-test-dependency
 
 # Set the default shell to bash instead of sh
 ENV SHELL /bin/bash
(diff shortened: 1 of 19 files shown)


.devcontainer/Dockerfile
```diff
@@ -47,7 +47,7 @@ RUN apt-get update \
         golang.org/x/lint/golint \
         github.com/alecthomas/gometalinter \
         honnef.co/go/tools/... \
-        github.com/golangci/golangci-lint/cmd/golangci-lint \
+        #github.com/golangci/golangci-lint/cmd/golangci-lint \
```
frodopwns, line 50: what was the issue here? doesn't work anymore?

.devcontainer/Dockerfile
```diff
@@ -93,7 +93,7 @@ COPY ./Makefile ./
 RUN make install-kind
 RUN make install-kubebuilder
 RUN make install-kustomize
-RUN make install-test-dependency
+# RUN make install-test-dependency
```
frodopwns, line 96: is this change widely applicable or was it a local workaround?

config/default/manager_image_patch.yaml
```diff
@@ -8,7 +8,7 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
-      - image: controller:latest
+      - image: IMAGE_URL
```
frodopwns, line 11: will need to undo this change

.devcontainer/Dockerfile
huangpf: I think Janani/Justin confirmed it this target we removed/renamed. Maybe I should just remove it.

.devcontainer/Dockerfile
huangpf: I think it's a dynamic package getting downloaded into the container, and it started to error out recently. So, I had to disable it to make the container build process work. Justin sent out this full fix PR: https://github.com/Azure/azure-service-operator/pull/1059

.devcontainer/Dockerfile
huangpf: I will try to change it to ```install-test-dependencies``` as Janani indicated.

huangpf:
Closing this one, and moving it to a newly created one: https://github.com/Azure/azure-service-operator/pull/1087

PR #1097, opened 2020-05-21
 Support Operator to Add/delete User for PostgreSQL database

### LINKED ISSUE #1056: As an operator, I can use the service operator to manage a user within the PostgreSQL database
**This user story covers the following**:
Be able to create/update/delete a user within the PostgreSQL database. 

**Acceptance Criteria**
[Update this once the owner starts looking at the user story]

- [ ] The service operator can create a user within a MySQL database with role name, username and password.
- [ ] The service operator can update the user's password
- [ ] The service operator can delete a user within the MySQL database.

**Tasks**
Reference this issue when you create tasks for this user story

buhongw7583c:
Closes #1056 

Support the function to add users with specific roles to Postgresql database. 
--Be able to add a user with a predefined roles and delete the user. 
-- Able to delete a user under the postgresql database. 

The Azure Database for PostgreSQL server is created with the 3 default roles defined. 

--azure_pg_admin
--azure_superuser
--your server admin user

The server admin user account can be used to create additional users and grant those users into the azure_pg_admin role.
Also, the server admin account can be used to create less privileged users and roles that have access to individual databases and schemas but this function is not included in this PR.  

To verify the PR:
--Use the config/samples/azure_v1alpha1_postgresql_everything.yaml to config pgresql server, database and firewallrule. Ensure at least your local client has the access to azure. 
-- Apply the sample file config/samples/azure_v1alpha1_postgresqluser.yaml to add a user with a predefined role. 
Use any tool to connect the postgresqlServer (e.g. pgadmin). The default server admin password could be derived from kubectl get secret <secret name> -o yaml. By default the secret name is the pgresqlserver name. 
image

-- Delete the newly added user by `kubect delete postgresqluser <metatdata name> `. Note the username is the metadata name applied when adding the pgresql user but not the actual pgresql user name. 
From the pgadmin tool, the user has been deleted. 

image

gif

**If applicable**:
- [ ] this PR contains tests


melonrush13:
Hi Hong! I think we are missing a few files for adding a new operator. Kubebuilder generates these files for us so we don't have to manually create them! 

[Here](https://github.com/Azure/azure-service-operator/blob/master/docs/howto/newoperatorguide.md) is our documentation for creating a new operator with kubebuilder.

Here are the files we need to declare the PostgreSQLUser in from KubeBuilder:
- PROJECT
- Config/crd/kustomization.yaml
- Config/crd/patches/webhook_in_postgresqluser
- Config/rbac/postgresqluser_editor_role.yaml
- Conifg/rbac/postgresqluser_viewer_role.yaml

In step 12 of the documentation I mentioned above, we update the helm chart to include the new operator (PostgreSqlUser). We also need these two files as well
- charts/azure-service-operator-0.1.0.tgz
- charts/index.yaml

melonrush13:
We also need to add the following inside the controller files:
- new file for postgresql user test (i.e. controller/postgresqluser_controller_test.go)
- declaration of postgresqluser in controllers/suite_test.go (declare like you declared in main.go)

config/samples/azure_v1alpha1_postgresqluser.yaml
```diff
@@ -0,0 +1,33 @@
...
+  dbName: postgresqldatabase-sample
+  resourceGroup: resourcegroup-azure-operators
+  # The Azure Database for PostgreSQL server is created with the 3 default roles defined.
+  # azure_pg_admin
+  # azure_superuser
+  # your server admin user
+  roles:
+    - "azure_pg_admin"
+   # Specify a specific username for the user
+  username: hong
```
melonrush13, line 16: I think we should comment out the rest of the optional fields (username/adminsecret/adminsecretkeyvault)

Also can you change username to be a generic value instead of Hong? :)

melonrush13: This is the standard we follow with the azuresqluser, too! ((Here:))][https://github.com/Azure/azure-service-operator/blob/master/config/samples/azure_v1alpha1_azuresqluser.yaml]

main.go
```diff
@@ -536,6 +541,22 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err = (&controllers.PostgreSQLUserReconciler{
+		Reconciler: &controllers.AsyncReconciler{
+			Client:      mgr.GetClient(),
+			AzureClient: psqlusermanager,
+			Telemetry: telemetry.InitializeTelemetryDefault(
+				"PSQLUser",
+				ctrl.Log.WithName("controllers").WithName("PSQLUser"),
```
melonrush13, line 550: Should we name this `PostgreSQLUser` to stick with naming conventions? Since that is the name inside of our controller?

main.go
```diff
@@ -536,6 +541,22 @@ func main() {
...
+			),
+			Recorder: mgr.GetEventRecorderFor("PSQLUser-controller"),
```
melonrush13, line 552: We should name this `PostgreSQLUser` to stick with naming conventions, since that is the name inside of our controller!

Similar to how the rest of the PostgreSQL resources are declared

controllers/postgresqluser_controller.go
```diff
@@ -0,0 +1,44 @@
...
+
+// PSQLUserFinalizerName is the name of the finalizer
+const PSQLUserFinalizerName = "psqluser.finalizers.azure.com"
+
+// PostgreSQLUserReconciler reconciles a PSQLUser object
+type PostgreSQLUserReconciler struct {
+	Reconciler *AsyncReconciler
+}
+
+// +kubebuilder:rbac:groups=azure.microsoft.com,resources=PostgreSQLUsers,verbs=get;list;watch;create;update;patch;delete
```
melonrush13, line 32: I think this should be all lowercase. Not sure if it will affect anything if it is not, but when kube builder generates this for us, it is lowercase

buhongw7583c:
> Hi Hong! I think we are missing a few files for adding a new operator. Kubebuilder generates these files for us so we don't have to manually create them!
> 
> [Here](https://github.com/Azure/azure-service-operator/blob/master/docs/howto/newoperatorguide.md) is our documentation for creating a new operator with kubebuilder.
> 
> Here are the files we need to declare the PostgreSQLUser in from KubeBuilder:
> 
> * PROJECT
> * Config/crd/kustomization.yaml
> * Config/crd/patches/webhook_in_postgresqluser
> * Config/rbac/postgresqluser_editor_role.yaml
> * Conifg/rbac/postgresqluser_viewer_role.yaml
> 
> In step 12 of the documentation I mentioned above, we update the helm chart to include the new operator (PostgreSqlUser). We also need these two files as well
> 
> * charts/azure-service-operator-0.1.0.tgz
> * charts/index.yaml

@melonrush13 Mel, thanks so much for the great comments! I do missed several important files and the document mentioned is definitely a great help! 

All the comments and suggestions have been addressed and just one left: 
for the postgresqluser_editor_role.yaml and postgresqluser_viewer_role.yaml suggested to add under rbac folder, i checked the folder and not all the controllers have corresponding **_editor_role.yaml or **_viewer_role.yaml. Are they generated by commands or manually added? How to decide which verbs to support?  Very much appreciate your careful checking on my PR and bring so many valuable comments!

buhongw7583c:
> Hi Hong!
> 
> I tested the Happy Path and it worked! Successfully created a user on top of the database, as well as deleting successfully! I tested this out by commenting out our optional fields of adminsecret, adminsecretkeyvault and username.
> 
> I am requesting changes so we can address the missing files from KubeBuilder, add tests in the controller, and address the following edge case I ran into.
> 
> If we create a PostgreSqlUser with a improperly connected IP address, we run into the following zapperr in the reconciler. It continues to loop with the following message.
> 
> ```
> "}, "reason": "FailedReconcile", "message": "Failed to reconcile resource: 1 error occurred:\n\t* pq: no pg_hba.conf entry for host \"73.181.19.189\", user \"yc5jvKen\", database \"postgresqldatabase-sample-m\", SSL on\n\n"}
> 2020-05-27T16:38:44.958-0600	ERROR	controller-runtime.controller	Reconciler error	{"controller": "postgresqluser", "request": "default/psqluser-sample-meltwo", "error": "1 error occurred:\n\t* pq: no pg_hba.conf entry for host \"73.181.19.189\", user \"yc5jvKen\", database \"postgresqldatabase-sample-m\", SSL on\n\n"}
> github.com/go-logr/zapr.(*zapLogger).Error
> ```
> 
> I tested this out in AzureSqlUser to see if we would have the same behavior, and it does not have this zaperr in the reconcile. Here are the two status messages for comparison:
> 
> PostgreSqlUser:
> `psqluser-sample-meltwo pq: no pg_hba.conf entry for host "73.181.19.189", user "yc5jvKen", database "postgresqldatabase-sample-m", SSL on`
> 
> AzureSqlUser
> `sqluser-sample Login error: mssql: Cannot open server 'sqlservermeltest' requested by the login. Client with IP address '73.181.19.189' is not allowed to access the server. To enable access, use the Windows Azure Management Portal or run sp_set_firewall_rule on the master database to create a firewall rule for this IP address or address range. It may take up to five minutes for this change to take effect.`
> 
> We should try to handle this err in the reconciler somehow
> 
> Let me know if this makes sense or not! Thanks!

@melonrush13 Mel, sure it makes sense! I did encounter the similar issue during my test but somehow it slipped from my mind and forgot to add handlings. Now the error could be caught up and wont produce the zapperr. But the error message returned from the sdk was quite confusing and indeed it indicate that your client IP address is not permitted to access, therefore I added more information in the returned message.

melonrush13:
> > Hi Hong! I think we are missing a few files for adding a new operator. Kubebuilder generates these files for us so we don't have to manually create them!
> > [Here](https://github.com/Azure/azure-service-operator/blob/master/docs/howto/newoperatorguide.md) is our documentation for creating a new operator with kubebuilder.
> > Here are the files we need to declare the PostgreSQLUser in from KubeBuilder:
> > 
> > * PROJECT
> > * Config/crd/kustomization.yaml
> > * Config/crd/patches/webhook_in_postgresqluser
> > * Config/rbac/postgresqluser_editor_role.yaml
> > * Conifg/rbac/postgresqluser_viewer_role.yaml
> > 
> > In step 12 of the documentation I mentioned above, we update the helm chart to include the new operator (PostgreSqlUser). We also need these two files as well
> > 
> > * charts/azure-service-operator-0.1.0.tgz
> > * charts/index.yaml
> 
> @melonrush13 Mel, thanks so much for the great comments! I do missed several important files and the document mentioned is definitely a great help!
> 
> All the comments and suggestions have been addressed and just one left:
> for the postgresqluser_editor_role.yaml and postgresqluser_viewer_role.yaml suggested to add under rbac folder, i checked the folder and not all the controllers have corresponding **_editor_role.yaml or **_viewer_role.yaml. Are they generated by commands or manually added? How to decide which verbs to support? Very much appreciate your careful checking on my PR and bring so many valuable comments!

Hi Hong! So these two rbac files are generated by KubeBuilder when you do the 
`kubebuilder create api --group azure --version v1alpha1 --kind <AzureNewType>` command. 

Also, we began discussing this on Teams as Janani also noticed not every controller has these rbac files like you mentioned. We think this might be a difference in kubebuilder versions. Currently if you create a new operator with the latest version of kubebuilder, it will create these rbac files. 

When you created the operator with Kubebuilder did it generate these files for you?

docs/howto/newoperatorguide.md
```diff
@@ -312,7 +312,7 @@ go build -o bin/manager main.go
     make install
     ```
 
-    Run controller tests using `make test-existing-controllers` and deploy using `make deploy`
+    Run controller tests using `make test-integration-controllers` and deploy using `make deploy`
```
melonrush13, line 315: Thanks for fixing this!!

config/samples/azure_v1alpha1_postgresqluser.yaml
melonrush13: @buhongw7583c Can you also comment out the other two optional fields? username and adminSecret? Thanks!

Also thanks for changing the name of postgresqluser :) Good fix! :) 👍

melonrush13:
Yay! Thanks for catching that err Hong for the improper connected IP. The zaperr was gone and handled gracefully. :)

buhongw7583c:
> > > Hi Hong! I think we are missing a few files for adding a new operator. Kubebuilder generates these files for us so we don't have to manually create them!
> > > [Here](https://github.com/Azure/azure-service-operator/blob/master/docs/howto/newoperatorguide.md) is our documentation for creating a new operator with kubebuilder.
> > > Here are the files we need to declare the PostgreSQLUser in from KubeBuilder:
> > > 
> > > * PROJECT
> > > * Config/crd/kustomization.yaml
> > > * Config/crd/patches/webhook_in_postgresqluser
> > > * Config/rbac/postgresqluser_editor_role.yaml
> > > * Conifg/rbac/postgresqluser_viewer_role.yaml
> > > 
> > > In step 12 of the documentation I mentioned above, we update the helm chart to include the new operator (PostgreSqlUser). We also need these two files as well
> > > 
> > > * charts/azure-service-operator-0.1.0.tgz
> > > * charts/index.yaml
> > 
> > 
> > @melonrush13 Mel, thanks so much for the great comments! I do missed several important files and the document mentioned is definitely a great help!
> > All the comments and suggestions have been addressed and just one left:
> > for the postgresqluser_editor_role.yaml and postgresqluser_viewer_role.yaml suggested to add under rbac folder, i checked the folder and not all the controllers have corresponding **_editor_role.yaml or **_viewer_role.yaml. Are they generated by commands or manually added? How to decide which verbs to support? Very much appreciate your careful checking on my PR and bring so many valuable comments!
> 
> Hi Hong! So these two rbac files are generated by KubeBuilder when you do the
> `kubebuilder create api --group azure --version v1alpha1 --kind <AzureNewType>` command.
> 
> Also, we began discussing this on Teams as Janani also noticed not every controller has these rbac files like you mentioned. We think this might be a difference in kubebuilder versions. Currently if you create a new operator with the latest version of kubebuilder, it will create these rbac files.
> 
> When you created the operator with Kubebuilder did it generate these files for you?

Hi, Mel, yes, these files are generated by kubebuilder! I was wrong that not using kubebuilder to generate the API related files but manually add file before so missed quite a few of them... :( 
Now they are all set!

buhongw7583c:
> Hi Hong!
> 
> Thanks so much for adding these files and addressing the changes! I no longer saw Zaperr's when I create a PostgreSqlUser with an invalid PostgreSqlFirewallrule. The err message you added was great!
> 
> Although, now I am unable to delete this improperly made postgresqluser. Fortunately, we do not have a Zaperr, but it is never deleted. We should address this before merging.
> 
> This is something we are also experiencing in AzureSqlUser. I made an issue for it, too. The AzureSqlUser gives us zaperr's and never deletes. Unlike yours, which just never deletes.
> 
> I left a couple other comments, too. :) Please let me know if you have any other questions, or have issues adding the RBAC files!
> 
> Thanks!!

@

buhongw7583c:
> Hi Hong!
> 
> Thanks so much for adding these files and addressing the changes! I no longer saw Zaperr's when I create a PostgreSqlUser with an invalid PostgreSqlFirewallrule. The err message you added was great!
> 
> Although, now I am unable to delete this improperly made postgresqluser. Fortunately, we do not have a Zaperr, but it is never deleted. We should address this before merging.
> 
> This is something we are also experiencing in AzureSqlUser. I made an issue for it, too. The AzureSqlUser gives us zaperr's and never deletes. Unlike yours, which just never deletes.
> 
> I left a couple other comments, too. :) Please let me know if you have any other questions, or have issues adding the RBAC files!
> 
> Thanks!!
@melonrush13 Mel, that's a very good finding! I can also reproduce this issue as well! 

The reason is: 
Because the client IP is not allowed to access to the server, so adding or deleting a postgresqluser both failed with the error message " pq: no pg_hba.conf entry for host ..." because both functions will try to connect to DB first.  Then the reconcile (adding and deleting) will re queue the request  until the IP address issue solved. 
In this case,  if we stop the re queue and just delete the instance from service operator side, it is fine. 

While at the same time, I thought of another rare scenario:
What if a postgresql user has been added successfully first (user in postgresql db) but due to some reason (should be very rare) the client IP has no access to the server suddenly and the delete operation is triggered but will fail. If we stop the reconcile, the postgresqluser instance can be deleted from service operator but indeed the user information is still in the db. 

In that case, I would expect the end user to solve the IP issue first while keep the reconcile, then the controller will handle it after all when IP address issue solves.  
What do you think?

melonrush13:
> > Hi Hong!
> > Thanks so much for adding these files and addressing the changes! I no longer saw Zaperr's when I create a PostgreSqlUser with an invalid PostgreSqlFirewallrule. The err message you added was great!
> > Although, now I am unable to delete this improperly made postgresqluser. Fortunately, we do not have a Zaperr, but it is never deleted. We should address this before merging.
> > This is something we are also experiencing in AzureSqlUser. I made an issue for it, too. The AzureSqlUser gives us zaperr's and never deletes. Unlike yours, which just never deletes.
> > I left a couple other comments, too. :) Please let me know if you have any other questions, or have issues adding the RBAC files!
> > Thanks!!
> > @melonrush13 Mel, that's a very good finding! I can also reproduce this issue as well!
> 
> The reason is:
> Because the client IP is not allowed to access to the server, so adding or deleting a postgresqluser both failed with the error message " pq: no pg_hba.conf entry for host ..." because both functions will try to connect to DB first. Then the reconcile (adding and deleting) will re queue the request until the IP address issue solved.
> In this case, if we stop the re queue and just delete the instance from service operator side, it is fine.
> 
> While at the same time, I thought of another rare scenario:
> What if a postgresql user has been added successfully first (user in postgresql db) but due to some reason (should be very rare) the client IP has no access to the server suddenly and the delete operation is triggered but will fail. If we stop the reconcile, the postgresqluser instance can be deleted from service operator but indeed the user information is still in the db.
> 
> In that case, I would expect the end user to solve the IP issue first while keep the reconcile, then the controller will handle it after all when IP address issue solves.
> What do you think?

Great point, Hong! I haven't thought of that edge case before. I think for now it will be okay to not handle that case since it is more obscure. Down the road this may be a good thing to handle!

For this PR, for the case where the IP is never able to successfully connect, we should be able to delete the postgresqluser successfully. Currently I'm unable to delete the sqluser for this case.

[Here](https://github.com/Azure/azure-service-operator/issues/1112) is the ticket I made for this issue on AzureSqlUser! We should handle it the same way here. 

Does that work for you? If you want to discuss more we can hop on a call 4:30 mountain time on my Monday to discuss?

PROJECT
```diff
@@ -54,6 +54,10 @@ resources:
   version: v1alpha1
   kind: PostgreSQLFirewallRule
 - group: azure
+  kind: PostgreSQLUser
+  version: v1alpha1
+- group: azure
+  kind: APIMgmtAPI
```
melonrush13, line 60: quick fix - remove this second declaration of `kind:  APIMgmtAPI`

frodopwns, line 60:
```suggestion
```

melonrush13:
Also one other thing, @buhongw7583c , we should update the docs [here](https://github.com/Azure/azure-service-operator/blob/master/docs/services/postgresql/postgresql.md) to account for this new operator. Feel free to address this in another PR if you like (same comment for MySql, too!)

config/crd/patches/cainjection_in_postgresqlusers.yaml
```diff
@@ -0,0 +1,8 @@
+# The following patch adds a directive for certmanager to inject CA into the CRD
+# CRD conversion requires k8s 1.13 or later.
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    certmanager.k8s.io/inject-ca-from: $(CERTIFICATE_NAMESPACE)/$(CERTIFICATE_NAME)
```
frodopwns, line 7: the cert manager annotation here is outdated...check AzureSQLServer for the updated one

controllers/postgresqluser_controller.go
```diff
@@ -0,0 +1,44 @@
...
+const PDriverName = "postgres"
+
+// PSecretUsernameKey is the username key in secret
+const PSecretUsernameKey = "username"
+
+// PSecretPasswordKey is the password key in secret
+const PSecretPasswordKey = "password"
+
+// PSQLUserFinalizerName is the name of the finalizer
+const PSQLUserFinalizerName = "postgresqluser.finalizers.azure.com"
```
frodopwns, line 25: Are these even used anywhere?

pkg/resourcemanager/psql/psqluser/psqluser.go
```diff
@@ -0,0 +1,285 @@
...
+	psqldbdnssuffix := "database.azure.com"
+	if config.Environment().Name != "AzurePublicCloud" {
+		psqldbdnssuffix = config.Environment().SQLDatabaseDNSSuffix
+	}
+	//the host or fullserveraddress should be:
+	//for public cloud <server>.postgres.database.azure.com
+	//for China cloud <server>.postgres.database.chinacloudapi.cn
+	//for German cloud <server>.postgres.database.cloudapi.de
+	//for US government <server>.postgres.database.usgovcloudapi.net
+	fullServerAddress := fmt.Sprintf("%s.postgres."+psqldbdnssuffix, server)
```
frodopwns, line 79: Is this something we should be setting in the psql secret? I believe we have this for Azure SQL.

pkg/resourcemanager/psql/psqluser/psqluser.go
```diff
@@ -0,0 +1,285 @@
...
+
+	connString := fmt.Sprintf("host=%s user=%s password=%s port=%d dbname=%s sslmode=require connect_timeout=30", fullServerAddress, user, password, port, database)
```
frodopwns, line 81: we should consider putting a conn string in the secret too

pkg/resourcemanager/psql/psqluser/psqluser.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+
+	return db, err
+}
+
+// GrantUserRoles grants roles to a user for a given database
+func (s *PostgreSqlUserManager) GrantUserRoles(ctx context.Context, user string, roles []string, db *sql.DB) error {
+	var errorStrings []string
+	for _, role := range roles {
+		tsql := fmt.Sprintf("GRANT %s TO %q", role, user)
```
frodopwns, line 100: You should use the interpolation built into the db client to dynamically craft sql queries. This is the only way to avoid sql injection. Some places in Azure SQL Server it is impossible so we have a workaround for one or two queries. In this case I believe the normal path should work fine.

e.g. https://github.com/Azure/azure-service-operator/blob/master/pkg/resourcemanager/azuresql/azuresqluser/azuresqluser.go#L154

pkg/resourcemanager/psql/psqluser/psqluser.go
frodopwns: This is already in the secret I guess

```
  "fullyQualifiedServerName": "your_server.postgres.database.azure.com",
```

pkg/resourcemanager/psql/psqluser/psqluser.go
frodopwns: Looks like some PSQL commands like CREATE USER and possible GRANT ROLE can't use the normal string interpolation to avoid sql injection. That means you will want to use the same filtering used in the CREATE USER command on the variables here.

pkg/resourcemanager/psql/psqluser/psqluser.go
```diff
@@ -0,0 +1,285 @@
...
+	_, err := db.ExecContext(ctx, tsql)
+
+	return err
+}
+
+// UserExists checks if db contains user
+func (s *PostgreSqlUserManager) UserExists(ctx context.Context, db *sql.DB, username string) (bool, error) {
+
+	tsql := fmt.Sprintf("SELECT * FROM pg_user WHERE usename = '%s'", username)
+	res, err := db.ExecContext(ctx, tsql)
```
frodopwns, line 163: this one at least I think we can prevent sql injection the preferred way

```
	res, err := db.ExecContext(ctx, "SELECT * FROM pg_user WHERE usename = $1", username)
```

pkg/resourcemanager/psql/psqluser/psqluser.go
```diff
@@ -0,0 +1,285 @@
...
+		return false, err
+	}
+	rows, err := res.RowsAffected()
+	return rows > 0, err
+
+}
+
+// DropUser drops a user from db
+func (s *PostgreSqlUserManager) DropUser(ctx context.Context, db *sql.DB, user string) error {
+	tsql := fmt.Sprintf("DROP USER IF EXISTS %q", user)
```
frodopwns, line 174: you will want to do the filtering on the user variable here too

pkg/resourcemanager/psql/psqluser/psqluser.go
```diff
@@ -0,0 +1,285 @@
...
+		"/*",
+	}
+
+	for _, s := range badChars {
+		if idx := strings.Index(stack, s); idx > -1 {
+			return fmt.Errorf("potentially dangerous character seqience found: '%s' at pos: %d", s, idx)
+		}
+	}
+	return nil
+}
```
frodopwns, line 285: this function already exists so it should be moved to a shared location like in the helpers package

pkg/resourcemanager/psql/psqluser/psqluser.go
```diff
@@ -0,0 +1,285 @@
...
+			dbUserCustomNamespace = instance.Spec.KeyVaultSecretPrefix
+		} else {
+			dbUserCustomNamespace = "psqluser-" + instance.Spec.Server + "-" + instance.Spec.DbName
+		}
+
+		namespacedName = types.NamespacedName{Namespace: dbUserCustomNamespace, Name: instance.Name}
+	} else {
+		namespacedName = types.NamespacedName{Name: instance.Name, Namespace: instance.Namespace}
+	}
+
```
frodopwns, line 266: This isn't going to work. The operator should have no concept of what type of secret store is being used. This is a problem in AzureSQL User too and that is why there is a PR out to refactor it. 

I think we might be able to remove the prefix idea from this operator.

pkg/resourcemanager/psql/psqluser/psqluser_reconcile.go
```diff
@@ -0,0 +1,446 @@
...
+
+	// get admin creds for server
+	adminSecret, err := adminSecretClient.Get(ctx, key)
+	if err != nil {
+		instance.Status.Provisioning = false
+		instance.Status.Message = fmt.Sprintf("admin secret : %s, not found in %s", key.String(), reflect.TypeOf(adminSecretClient).Elem().Name())
+		return false, nil
+	}
+
+	adminUser := string(adminSecret[PSecretUsernameKey]) + "@" + string(instance.Spec.Server)
```
frodopwns, line 75: the fully qualified username is in the secret already

pkg/resourcemanager/psql/psqluser/psqluser_reconcile.go
```diff
@@ -0,0 +1,446 @@
...
+					formattedSecrets["password"] = DBSecret["password"]
+				}
+			} else {
+				err = sqlUserSecretClient.Delete(
+					ctx,
+					types.NamespacedName{Namespace: key.Namespace, Name: instance.Name + "-" + formatName},
+				)
+			}
+		}
+
```
frodopwns, line 245: I'd like to drop all this secret formatting stuff. There is a PR to refactor it in AzureSQL and I don't want to introduce this feature until its fixed and/or required by a user.

pkg/resourcemanager/psql/psqluser/psqluser_reconcile.go
```diff
@@ -0,0 +1,446 @@
...
+		// if the database is busy, requeue
+		errorString := err.Error()
+		if strings.Contains(errorString, "Please retry the connection later") {
+			return false, nil
+		}
+
+		// if this is an unmarshall error - igmore and continue, otherwise report error and requeue
+		if !strings.Contains(errorString, "cannot unmarshal array into Go struct field serviceError2.details") {
+			return false, err
+		}
```
frodopwns, line 102: Just an fyi...the non string parsing way to detect errors like this:

```
 if _, ok := err.(*json.UnmarshalTypeError); ok {
       // ...this is an unmarshal error
 }
```

pkg/resourcemanager/psql/psqluser/psqluser_reconcile.go
```diff
@@ -0,0 +1,446 @@
...
+				Name:      instance.Spec.DbName,
+			},
+			Target: &v1alpha1.PostgreSQLDatabase{},
+		},
+		{
+			Key: types.NamespacedName{
+				Namespace: instance.Namespace,
+				Name:      instance.Spec.Server,
+			},
+			Target: &v1alpha1.PostgreSQLServer{},
```
frodopwns, line 419: need to update this to `v1alpha2`

frodopwns: otherwise you can't deploy a psql user without the conversion webhooks

pkg/resourcemanager/psql/psqluser/psqluser.go
buhongw7583c: @frodopwns  Erin, this cannot use the SQL filtering as the username needs to be double quoted in this command if there is any special char inside, like '-' etc.

pkg/resourcemanager/psql/psqluser/psqluser.go
```diff
@@ -0,0 +1,285 @@
...
+				ctx,
+				key,
+			)
+			if err != nil {
+				instance.Status.Message = "failed to delete secret, err: " + err.Error()
+				return false, err
+			}
+		}
+	}
+
```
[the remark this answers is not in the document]
buhongw7583c: I only left this comment to confirm with you since the code change is big.  if we remove the custom secret formating, so we would  not allow user to use keyVaultSecretPrefix but we still allow customer to use keyvault to store the secret, right?

pkg/resourcemanager/psql/psqluser/psqluser.go
buhongw7583c: same question: we would remove both the keyVaultToStoreSecrets and keyVaultSecretPrefix?

PR #1100, opened 2020-05-26
Final set of changes for new pipeline

jananivMS:
Updates to pipeline to make it robust for being the primary pipeline.

- Adds logic to pick the free AKS cluster from the test pool and use that for deploy testing
- Fix issues with ACRs in the pipeline

gif

**If applicable**:
- [ ] this PR contains tests


azure-pipelines.yml
```diff
@@ -137,28 +136,51 @@ steps:
   - task: Docker@2
     displayName: Login to temporary pipeline ACR
     inputs:
-      command: login
-      containerRegistry: $(PIPELINE_CONTAINER_REGISTRY)
+      containerRegistry: 'ASO Pipeline ACR'
```
jpflueger, line 139: Just out of curiosity, it seems like there are a few different variables for this. There is still a "containerRegistry" variable set above. One task uses a variable called ["CONTAINER_REGISTRY_NAME"](https://github.com/Azure/azure-service-operator/blob/2dc02f8996c2ab3184bcb0097f0de1fb35ab3e57/azure-pipelines.yml#L291). Another task uses a variable called ["PIPELINE_ CONTAINER_REGISTRY_NAME"](https://github.com/Azure/azure-service-operator/blob/2dc02f8996c2ab3184bcb0097f0de1fb35ab3e57/azure-pipelines.yml#L230). Not saying that any of this is wrong, just that it's confusing which container registry is used and when. Do you know where the other variables come from and if they are they same as what you're setting here?

azure-pipelines.yml
```diff
@@ -137,28 +136,51 @@ steps:
...
-        $(MAJOR_VERSION).$(MINOR_VERSION).$(PATCH_VERSION)
+      containerRegistry: 'ASO Pipeline ACR'
+      repository: '$(IMAGE_NAME)'
+      command: 'buildAndPush'
+      Dockerfile: '$(System.DefaultWorkingDirectory)/Dockerfile'
+      buildContext: '$(System.DefaultWorkingDirectory)'
+      tags: '$(MAJOR_VERSION).$(MINOR_VERSION).$(PATCH_VERSION)'
+
+#  - task: Kubernetes@1
+#    displayName: Deploy to AKS - Fetch kubeconfig
```
jpflueger, line 153: Do we need this task anymore? Can it be deleted or is it temporarily disabled? Might be helpful to comment why if it's only temporary.

azure-pipelines.yml
```diff
@@ -137,28 +136,51 @@ steps:
...
+      azureSubscription: 'ASO Subscription'
+      scriptType: 'bash'
+      scriptLocation: 'inlineScript'
+      inlineScript: |
+        az resource list --tag freeforpipeline=true --query '[].{Name:name}' -o table > clusters.txt
+        clustername=$(tail -n 1 clusters.txt)
+        if [ "$clustername" == "" ]
+        then
+        echo AKS clusters unavailable
+        exit 1
```
jpflueger, line 173: You could shorten & indent this like
```
if [ "$clustername" == "" ]; then
  echo AKS clusters unavailable
  exit 1
fi
```

azure-pipelines.yml
```diff
@@ -137,28 +136,51 @@ steps:
...
+        az resource list --tag freeforpipeline=true --query '[].{Name:name}' -o table > clusters.txt
+        clustername=$(tail -n 1 clusters.txt)
+        if [ "$clustername" == "" ]
+        then
+        echo AKS clusters unavailable
+        exit 1  
+        fi
+        echo $clustername
+        echo "##vso[task.setvariable variable=chosenclustername]$clustername"
+        echo 'az aks get-credentials -g $(AKS_CLUSTER_RG) -n $clustername'
```
jpflueger, line 177: Where does this variable (AKS_CLUSTER_RG) come from? I think it's recommend to pass through variables using the "env" property so it is a little more explicit that it is coming from the pipeline. Can also help with debugging when you turn on "System Debugging" and re-run the pipeline.

azure-pipelines.yml
```diff
@@ -137,28 +136,51 @@ steps:
...
+        az aks get-credentials -g $(AKS_CLUSTER_RG) -n $clustername
+        # Set tags to not available for the selected cluster so it doesn't get used
+        az resource tag --tags 'freeforpipeline=false' -g $(AKS_CLUSTER_RG) -n $clustername --resource-type Microsoft.ContainerService/managedClusters
+      workingDirectory: '$(System.DefaultWorkingDirectory)'
```
jpflueger, line 181: You don't really need to set this because the working directory is set to this for each task by default.

azure-pipelines.yml
```diff
@@ -227,23 +249,38 @@ steps:
...
   ## TODO: Add a Health check task to ensure operator deployed to AKS works
+  
+  - task: AzureCLI@2
+    inputs:
+      azureSubscription: 'ASO Subscription'
+      scriptType: 'bash'
+      scriptLocation: 'inlineScript'
+      inlineScript: |
+        echo "Chosen AKS Cluster name"
+        echo $(chosenclustername)
```
jpflueger, line 260: Similar to a previous comment but you can use an env block to show that you're passing in a variable from the previous task. That way if the output name changes in the previous task I think it will break this task a little more explicitly.

azure-pipelines.yml
jananivMS: I think some cleanup is due for sure. There are two container registries - one is a temporary pipeline registry that we push the image to and use for testing on the cluster, and only if the deploy test succeeds we push the image to a release ACR, which is the other one. Happy to look at consolidating some of these

jananivMS: Addressed this and also added the variables needed to be configured at the begining of the pipeline for clarity

azure-pipelines.yml
jananivMS: fixed

azure-pipelines.yml
jananivMS: It does come from the pipeline variables. I documented these at the start of the pipeline

azure-pipelines.yml
jananivMS: Ok, I left it as is to make sure there isnt any new issues

azure-pipelines.yml
jananivMS: I experimented with that a bit but it looks like it works only for jobs.

PR #1102, opened 2020-05-26
Bug: AppInsights doesn't support storing secret in user defined keyvault

### LINKED ISSUE #1039: Bug: AppInsights doesn't support storing secret in user defined keyvault
AppInsights doesn't have the "keyVaultToStoreSecrets" property so it can't store the instrumentation key in a user defined KeyVault. 

**Done criteria**

- [ ] AppInsights should be able to save secrets in a user-defined keyvault

WilliamMortlMicrosoft:
closes #1039

AppInsights doesn't have the "keyVaultToStoreSecrets" property so it can't store the instrumentation key in a user defined KeyVault. 

**Done criteria**

- [ ] AppInsights should be able to save secrets in a user-defined keyvault


melonrush13:
Hi Will!

Currently testing right now, and these are the steps I went through to test the instrumentationkey.

`k create -f config/samples/azure_v1alpha1_appinsights.yaml`
Msg Status is Successfully provisioned

`k get secrets`
App insights is listed:
appinsights-sample-meltest                       Opaque                                1      9s)

`k get secrets appinsights-sample-meltest -o yaml`
Instrumentation key is listed under data

Right now I'm testing this on Master and it is already working. I try the same thing on your branch, and it performs the same way.

This is expected behavior, correct? Am I testing properly?

WilliamMortlMicrosoft:
> Hi Will!
> 
> Currently testing right now, and these are the steps I went through to test the instrumentationkey.
> 
> `k create -f config/samples/azure_v1alpha1_appinsights.yaml`
> Msg Status is Successfully provisioned
> 
> `k get secrets`
> App insights is listed:
> appinsights-sample-meltest Opaque 1 9s)
> 
> `k get secrets appinsights-sample-meltest -o yaml`
> Instrumentation key is listed under data
> 
> Right now I'm testing this on Master and it is already working. I try the same thing on your branch, and it performs the same way.
> 
> This is expected behavior, correct? Am I testing properly?

Did you see one named something like: appinsights-{resourceGroupName}-{appInsightsName}

melonrush13:
> > Hi Will!
> > Currently testing right now, and these are the steps I went through to test the instrumentationkey.
> > `k create -f config/samples/azure_v1alpha1_appinsights.yaml`
> > Msg Status is Successfully provisioned
> > `k get secrets`
> > App insights is listed:
> > appinsights-sample-meltest Opaque 1 9s)
> > `k get secrets appinsights-sample-meltest -o yaml`
> > Instrumentation key is listed under data
> > Right now I'm testing this on Master and it is already working. I try the same thing on your branch, and it performs the same way.
> > This is expected behavior, correct? Am I testing properly?
> 
> Did you see one named something like: appinsights-{resourceGroupName}-{appInsightsName}

Yes I did Will! When I make an appinsights on your branch and do a `k get secrets` I get two instances. 
One as:
`appinsights-{resourceGroupName}-{appInsightsName}`
and one as:
`{appInsightsName}
`
On master I only get
` {appInsightsName}
`

WilliamMortlMicrosoft:
> So the reason we are seeing this twice is because we are already upserting the secret already in the Ensure function [here](https://github.com/Azure/azure-service-operator/blob/80c75a62e59e928c281c8f0214364750bfc761fa/pkg/resourcemanager/appinsights/appinsights.go#L197.)
> 
> I think we can use the existing code, and then we should account for the case where the user wants to create their own separate KeyVault (which is why I think this bug was created)
> 
> We can specify this in the yaml as
> 
> ```
>  # Use the field below to optionally specify a different keyvault 
>  # to store the secrets in
>  #keyVaultToStoreSecrets: asoSecretKeyVault
> ```
> 
> And then account for this case in the Ensure function and we should be good (and delete, of course too.)

I dont think we should use the existing code... its really in the wrong location in the ensure loop, I deleted the old code out. Great catch on the delete - Added :-)

pkg/resourcemanager/appinsights/appinsights.go
```diff
@@ -153,21 +200,6 @@ func (m *Manager) Ensure(ctx context.Context, obj runtime.Object, opts ...resour
...
-		key,
-		map[string][]byte{"instrumentationKey": []byte(instKey)},
-		secrets.WithOwner(instance),
-		secrets.WithScheme(m.Scheme),
-	)
-	if err != nil {
-		instance.Status.Message = "failed to update secret, err: " + err.Error()
-		return false, err
-	}
-
```
WilliamMortlMicrosoft, line 170: removed! great catch @melonrush13  (I feel stupid for missing it)

config/samples/azure_v1alpha1_appinsights.yaml
```diff
@@ -7,3 +7,7 @@ spec:
   location: westus
   resourceGroup: resourcegroup-azure-operators
   applicationType: other # Possible values include 'web' or 'other'
+
+  # Use the field below to optionally specify a different keyvault 
+  # to store the secrets in
+  #keyVaultToStoreSecrets: asoSecretKeyVault
```
WilliamMortlMicrosoft, line 13: I had previously added the type, but not this to the sample YAML - another great catch!

pkg/resourcemanager/appinsights/appinsights.go
```diff
@@ -201,15 +233,23 @@ func (m *Manager) Delete(ctx context.Context, obj runtime.Object, opts ...resour
 		if helpers.ContainsString(catch, azerr.Type) {
 			return true, nil
 		} else if helpers.ContainsString(gone, azerr.Type) {
-			return false, nil
+			err = m.DeleteSecret(ctx,
+				i.Spec.ResourceGroup,
+				i.Name,
+				i)
+			return false, err
```
WilliamMortlMicrosoft, line 240: secrets are now being deleted :-)

melonrush13:
Will - I was able to successfully create appinsights secrets and delete them normally, as well as add a keyvault successfully and see the instrumentation key on the portal. :) Thanks for the fix!

One last question for you:
If I create an appinsights with a specified keyvault (which is already created and successfully connected) and then I delete that keyvault (and it successfully deletes) If i go to delete that appinsights now, it should successfully be able to delete to, yes?

Currently I am unable to delete that appinsights, and it is continuing to try, but never successful (been waiting about 15 min for delete)
```
2020-06-01T12:56:28.531-0600	INFO	controllers.AppInsights	ensuring vault	{"Info Type": "status", "Component": "AppInsights", "Instance": "default/appinsights-sample-keyvault"}
2020-06-01T12:56:28.534-0600	INFO	controllers.AppInsights	awaiting vault verification	{"Info Type": "requeuing", "Component": "AppInsights", "Instance": "default/appinsights-sample-keyvault"}
```

melonrush13:
> > So the reason we are seeing this twice is because we are already upserting the secret already in the Ensure function [here](https://github.com/Azure/azure-service-operator/blob/80c75a62e59e928c281c8f0214364750bfc761fa/pkg/resourcemanager/appinsights/appinsights.go#L197.)
> > I think we can use the existing code, and then we should account for the case where the user wants to create their own separate KeyVault (which is why I think this bug was created)
> > We can specify this in the yaml as
> > ```
> >  # Use the field below to optionally specify a different keyvault 
> >  # to store the secrets in
> >  #keyVaultToStoreSecrets: asoSecretKeyVault
> > ```
> > 
> > 
> > And then account for this case in the Ensure function and we should be good (and delete, of course too.)
> 
> I dont think we should use the existing code... its really in the wrong location in the ensure loop, I deleted the old code out. Great catch on the delete - Added :-)

Yes agree! I

WilliamMortlMicrosoft:
> Will - I was able to successfully create appinsights secrets and delete them normally, as well as add a keyvault successfully and see the instrumentation key on the portal. :) Thanks for the fix!
> 
> One last question for you:
> If I create an appinsights with a specified keyvault (which is already created and successfully connected) and then I delete that keyvault (and it successfully deletes) If i go to delete that appinsights now, it should successfully be able to delete to, yes?
> 
> Currently I am unable to delete that appinsights, and it is continuing to try, but never successful (been waiting about 15 min for delete)
> 
> ```
> 2020-06-01T12:56:28.531-0600	INFO	controllers.AppInsights	ensuring vault	{"Info Type": "status", "Component": "AppInsights", "Instance": "default/appinsights-sample-keyvault"}
> 2020-06-01T12:56:28.534-0600	INFO	controllers.AppInsights	awaiting vault verification	{"Info Type": "requeuing", "Component": "AppInsights", "Instance": "default/appinsights-sample-keyvault"}
> ```

Ok, so the answer is that YES that SHOULD WORK - but looking at the code, the delete of the secret would fail and I'd return false, err... do to an err being returned it would requeue forever - im pushing a small change that no longer returns errors from that.

To be honest, the code was working (sort of) as designed - as the secret was impossible to remove, but this way itll work correctly the first time.

pkg/resourcemanager/appinsights/appinsights.go
```diff
@@ -209,6 +263,10 @@ func (m *Manager) Delete(ctx context.Context, obj runtime.Object, opts ...resour
 
 	if err == nil {
 		if response.Status != "InProgress" {
+			m.DeleteSecret(ctx,
+				i.Spec.ResourceGroup,
+				i.Name,
+				i)
 			return false, nil
```
WilliamMortlMicrosoft, line 270: @melonrush13 we now always return nil for the error instead of the err if delete secret fails... this allows for the secret to already be gone

PR #1104, opened 2020-05-26
Rediscache roll actions

### LINKED ISSUE #1034: Task: Re-roll keys for RedisCache
Determine if the RedisCache API will allow the connection keys to be reset.

If so, add a RedisCacheAction operator similar to the AzureSQLAction operator to allow this.

**Done criteria**
- Users of ASO can reset RedisCache Auth keys

jpflueger:
Closes #1034 

Adds RedisCacheAction resource to trigger access key regeneration.

1. Create RedisCache resource
2. Wait until provisioned and note access keys
3. Create RedisCacheAction resource using the name from step 1
4. After it has provisioned, check the secret to verify that the access keys have changed
5. Repeat and set keyvaultToStoreSecrets

gif

**If applicable**:
- [ ] this PR contains tests


pkg/resourcemanager/rediscaches/rediscacheaction_reconcile.go
```diff
@@ -0,0 +1,114 @@
...
+		if err = m.RegeneratePrimaryAccessKey(ctx, instance.Spec.ResourceGroup, instance.Spec.CacheName); err != nil {
+			return false, err
+		}
+	}
+
+	if rollAllKeys || instance.Spec.ActionName == v1alpha1.RedisCacheActionNameRollSecondaryKey {
+		if err = m.RegenerateSecondaryAccessKey(ctx, instance.Spec.ResourceGroup, instance.Spec.CacheName); err != nil {
+			return false, err
+		}
+	}
```
frodopwns, line 45: NIT Alert: This might be one of those rare occasions where a switch is actually cleaner...let's see...

```
switch instance.Spec.ActionName {
case v1alpha1.RedisCacheActionNameRollAllKeys, v1alpha1.RedisCacheActionNameRollPrimaryKey:
              if err = m.RegeneratePrimaryAccessKey(ctx, instance.Spec.ResourceGroup, instance.Spec.CacheName); err != nil {
			return false, err
		}
case v1alpha1.RedisCacheActionNameRollAllKeys, v1alpha1.RedisCacheActionNameRollSecondaryKey:
               if err = m.RegenerateSecondaryAccessKey(ctx, instance.Spec.ResourceGroup, instance.Spec.CacheName); err != nil {
			return false, err
		}
}
```

pkg/resourcemanager/rediscaches/rediscacheaction_reconcile.go
```diff
@@ -0,0 +1,114 @@
...
+			},
+			Target: &v1alpha1.ResourceGroup{},
+		},
+		{
+			Key: types.NamespacedName{
+				Namespace: instance.Namespace,
+				Name:      instance.Spec.CacheName,
+			},
+			Target: &v1alpha1.RedisCache{},
+		},
```
frodopwns, line 95: I think your order here is in reverse. The Async Controller loops through the parents until it finds one that it can reference as owner. In this case it would reference the RG before trying the RedisCache itself.

pkg/resourcemanager/rediscaches/rediscacheaction_reconcile.go
```diff
@@ -0,0 +1,114 @@
...
+			Namespace: instance.Namespace,
+		},
+		Spec: v1alpha1.RedisCacheSpec{
+			SecretName:        instance.Spec.SecretName,
+			ResourceGroupName: instance.Spec.ResourceGroup,
+		},
+	}
+	if err = m.ListKeysAndCreateSecrets(ctx, cacheInstance); err != nil {
+		instance.Status.Provisioned = false
+		instance.Status.FailedProvisioning = true
```
frodopwns, line 60: Status.Message = err.Error()

pkg/resourcemanager/rediscaches/rediscacheaction_reconcile.go
```diff
@@ -0,0 +1,114 @@
...
+		return false, err
+	}
+
+	// successful return
+	instance.Status.Provisioned = true
+	instance.Status.FailedProvisioning = false
```
frodopwns, line 66: set a success message here

pkg/resourcemanager/rediscaches/rediscacheaction_reconcile.go
jpflueger: Wouldn't that only regenerate the primary key when you try to "rollallkeys" because the first case succeeds and doesn't hit the second case?

pkg/resourcemanager/rediscaches/rediscacheaction_reconcile.go
jpflueger: I'm still getting the RedisCache as the owner when I deploy locally. Would it be better to only declare the rediscache as a parent since it doesn't really have a resource group?

pkg/resourcemanager/rediscaches/rediscacheaction_reconcile.go
frodopwns: ah, could be, probably easy enough to test

pkg/resourcemanager/rediscaches/rediscacheaction_reconcile.go
frodopwns: Having all the possible ancestors allows for the possibility that the users might choose to create some of the dependencies out of band. I don't know why they might do this but they might create the RG, then create the redis some other way for somme reason, then maybe they still want to try to run RedisActions. In this case the Action would be owned by the RG and would be cleaned up if the RG was deleted from kube.

jpflueger: Fixed.

PR #1105, opened 2020-05-26
add support to separate out CRDs in crd folder for helm chart

### LINKED ISSUE #1005: Task: Add support to Helm chart to be able to deploy multiple instances of the operator to different namespaces in the cluster
We would need the ability to install multiple versions of operators in different namespaces in the same cluster (for the pipeline atleast). We should add the ability to specify to install the operator when some of the prereqs exist

**Done criteria**

- [ ] Helm chart can be used to install operator instances in multiple namespaces in the same cluster

cnadolny:
- Separates out the CRDs into a separate folder as per Helm3 recommendation
- Adds instructions for installing operator to a specified namespace (webhook service for CRDs to be installed in a specified namespace)

How to test:
- Install CRDS in the CRD folder (default is azureoperator-system namespace). Create first instance of ASO with Helm, default values are fine. Make sure ASO is installed and webhooks work well.
- Now install CRDs in a different namespace. Install ASO using Helm in that namespace. Make sure ASO is installed and webhooks work well.

gif

**If applicable**:
- [ ] this PR contains tests


jananivMS:
@cnadolny  - Does the documentation here need to change  for how to update the helm charts and update them in the repo when a new CRD is added? Step 12 here - https://github.com/Azure/azure-service-operator/blob/master/docs/howto/newoperatorguide.md

PR #1122, opened 2020-06-01
repackage helm chart

frodopwns:
Helm chart in master is not installing CRDs properly. This PR repackages the chart.


charts/index.yaml
```diff
@@ -3,19 +3,19 @@ entries:
...
-    created: "2020-05-20T10:20:46.240488-06:00"
+    created: "2020-06-01T12:41:48.174129-06:00"
     dependencies:
     - condition: azureUseMI
       name: aad-pod-identity
       repository: https://raw.githubusercontent.com/Azure/aad-pod-identity/master/charts
       version: 1.5.5
     description: Deploy components and dependencies of azure-service-operator
-    digest: f381ed2be22005190cffc85883f42a2f13512d7b789ba8a03d8ee4cf06235241
+    digest: ec95181a3e59179bfdb3bd92989613b1469104a39e6054acea90382795b1fe2d
```
WilliamMortlMicrosoft, line 13: did changing the digest number fix things?

charts/index.yaml
```diff
@@ -3,19 +3,19 @@ entries:
   azure-service-operator:
   - apiVersion: v2
     appVersion: 0.1.0
-    created: "2020-05-20T10:20:46.240488-06:00"
+    created: "2020-06-01T12:41:48.174129-06:00"
     dependencies:
     - condition: azureUseMI
       name: aad-pod-identity
       repository: https://raw.githubusercontent.com/Azure/aad-pod-identity/master/charts
```
WilliamMortlMicrosoft, line 10: https://raw.githubusercontent.com/Azure/aad-pod-identity/master/charts doesnt load for me

charts/index.yaml
frodopwns: nope, looks like master got contaminated with templates from the helm chart PR that is currently being reviewed...added a condition around the CRD templates that prevented them from installing

charts/index.yaml
frodopwns: hmm wonder if it moved...

@cnadolny did the aad pod chart move? seems to be in the `aad-pod-identity` directory under charts

cnadolny: This should be okay, as running the command below works:
helm repo add aad-pod-identity https://raw.githubusercontent.com/Azure/aad-pod-identity/master/charts

frodopwns: Thanks! Looks good. I was able to add the repo as well using that url.

PR #1131, opened 2020-06-03
Fix RBAC issues with postgresqluser CRD 

jananivMS:
It looks like the Postgresqluser PR added a kubebuilder directive for shortname that defined a different name for the CRD. This is causing issues when making helm chart manifests and deploying to AKS cluster. 

Removing this directive now as we are looking at shortnames in a different PR.

After this, I still have RBAC issues with this CRD. So had to enable RBAC through the kubebuilder directive in main.go. This aberration seems mostly because of not using Kubebuilder to generate the CRD


buhongw7583c:
@jananivMS  Janani, I can now successfully deploy the CRD on AKS and verify the creation of postgresqlserver/database/firewallrule/user. But seems the PR can only be reviewed with at least one of the code owners' approval.

PR #1144, opened 2020-06-07
allow ec keys to be created via the keyvaultkey operator

frodopwns:
KeyVault Key types EC and ECHSM require a Curve field. This PR adds that field and updates the sample config.


config/samples/azure_v1alpha1_keyvaultkey.yaml
```diff
@@ -6,7 +6,12 @@ spec:
...
   location: westus
   keyVault: keyvaultsample123
-  keySize: 4096
+  # type can be RSA, RSAHSM, EC, ECHSM
   type: RSA
+  # if RSA
+  keySize: 4096
+  # If EC
+  # type: EC
+  # Curve: P256 #Possible values include: 'P256', 'P384', 'P521', 'P256K'
```
frodopwns, line 15: should be lower case "curve"

pkg/resourcemanager/keyvaults/keyops.go
```diff
@@ -81,6 +81,24 @@ func (k *KeyvaultKeyClient) Ensure(ctx context.Context, obj runtime.Object, opts
 		KeyOps:        &instance.Spec.Operations,
 		KeyAttributes: &katts,
 	}
+
+	switch instance.Spec.Type {
+	case kvops.RSA, kvops.RSAHSM:
+		if instance.Spec.KeySize == 0 {
+			instance.Status.Message = "no keysize provided for rsa key"
+			instance.Status.FailedProvisioning = true
```
jpflueger, line 89: Do you need to also set `instance.Status.Provisioning = false` here (and below)?

PR #1157, opened 2020-06-11
New pipeline update: Remove Helm chart packaging stage and use checked in Helm chart instead

### LINKED ISSUE #1132: Task: Update pipeline to use checked in helm chart
Currently, the new pipeline, generates the helm chart in the pipeline to test and upload to ACR. We should change this to use the checked in Helm Chart as this enables us to make sure that the Helm chart in the repo is kept updated.

- [ ] Helm chart in the repo should be used to validate in the pipeline
- [ ] We should also have all the generated files checked in under the templates/generated folder for the helm chart

jananivMS:
Closes #1132 

- We were building the helm chart in the pipeline, but that results in us not being able to validate the Helm chart thats checked in and makes it impossible for us to catch the Helm chart not being updated in the repo. So the pipeline now uses the checked in charts and pushes that to MCR/release. 
- Also cleans up the CRDs and deployment in the cluster at the end of the deploy test

Latest run of the pipeline - https://dev.azure.com/azure/azure-service-operator/_build/results?buildId=10817&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9

gif

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -244,8 +244,6 @@ steps:
         echo $(chosenclustername)
         az aks get-credentials -g $(AKS_CLUSTER_RG) -n $(chosenclustername)
         kubectl delete -f ./charts/azure-service-operator/crds/
-        helm list
-        helm delete aso
         # Set tags to available for the selected cluster to put it back into the free pool
         echo "Setting tags back to free"
         az resource tag --tags 'freeforpipeline=true' -g $(AKS_CLUSTER_RG) -n $(chosenclustername) --resource-type Microsoft.ContainerService/managedClusters
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -207,20 +207,20 @@ steps:
       echo $img
       sed -i -e 's@azure-service-operator:latest@azureserviceoperator:'${img}'@' charts/azure-service-operator/values.yaml
       # create directory for generated files
-      mkdir charts/azure-service-operator/templates/generated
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -258,11 +242,16 @@ steps:
...
+        kubectl delete -f ./charts/azure-service-operator/crds/
+        # Remove Helm deployment
+        helm list -n $(OPERATOR_NAMESPACE)
+        helm delete aso -n $(OPERATOR_NAMESPACE)
         # Set tags to available for the selected cluster to put it back into the free pool
         echo "Setting tags back to free"
         az resource tag --tags 'freeforpipeline=true' -g $(AKS_CLUSTER_RG) -n $(chosenclustername) --resource-type Microsoft.ContainerService/managedClusters
       workingDirectory: '$(System.DefaultWorkingDirectory)'
-      failOnStandardError: true
+      failOnStandardError: false
```
WilliamMortlMicrosoft, line 254: Why wouldn't we want to fail on standard error? Can you add a comment to explain?

jananivMS: Since it’s a cleanup step that doesn’t impact the actual pipeline I changed that but now that I think of it I think it might be good to fail so we get feedback if things are failing. Will update that.

jananivMS: Fixed this now @WilliamMortlMicrosoft

Issue #1147, opened 2020-06-07
Task: investigate to use sql db for SQL commands in MySQLUser and PostgreSQLUser. 


buhongw7583c:
In MySQLUser and PostgreSQLUser, there are several SQL commands using SQL injections. 
e.g. in https://github.com/Azure/azure-service-operator/blob/d0294e15530da5a40bb42e82ce60a01213188551/pkg/resourcemanager/psql/psqluser/psqluser.go#L124

The reason using the SQL injection for MySQLUser and PostgreSQLUser SQL commands is the db.exec() provided by database/sql packages cannot executed and accepted by MySQL and PostgreSQL. 
It worth doing an investigation to see whether we can use database/sql db functions to execute the SQL Commands. 

Reference  document: https://www.calhoun.io/what-is-sql-injection-and-how-do-i-avoid-it-in-go/. . 

**Done criteria**
This is the "done" criteria for this task

- [ ] Investigate whether the MySQLUser's createuser, GrantUserRoles, DropUser command can use db.Exec to execute instead of SQL injection. 
- [ ] Investigate whether PostgreSQLUser's createuser, GrantUserRoles, DropUser commands can use db.Exec to execute instead of SQL injection.

davefellows:
Have you got an example of the code you've tried and the error you're hitting?

Not sure if this is relevant:
https://stackoverflow.com/questions/50206617/db-exec-args-always-results-in-an-error-for-my-placeholder

buhongw7583c:
@davefellows, Dave, for example, the mysql command for drop user is
```
 DROP USER 'username'@'servername';
```
We at first used the db.exec to run. 
```

	tsql := "DROP USER ?"
        user:= string("'username'@'servername'")
	_, err := db.ExecContext(ctx, tsql, user)
```
Then the error 
```
You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '?' at line 1
```
Yeah, worth checking your referred document.

buhongw7583c:
Add the result from a quick test: 
using the 'interpolateParams' set to true for mysql db connection and the db.exec works for create user and drop user. User could be added/deleted without error. 
@davefellows , great finding of the document! There should be similar one for postgresql and sql for go as well. 
```
       //connection to mysql server
        var connectionString = fmt.Sprintf("%s:%s@tcp(%s:3306)/%s?tls=skip-verify&interpolateParams=true", user, password, host, database)

	// Initialize connection object.
[...]

PR #1160, opened 2020-06-15
Corret the MySQL username format to <username>@%, and use SQL injection for MySQL CreateUser and DropUser

buhongw7583c:
This PR is part of  #1147, also fix the username bug found during customer test in engagement3. 

The PR has fixes: 
-- As customer requested,  MySQL Username should be created using the default format: ```<username>@%```, but not ```<username>@<servername>```.The '%' is the hostname and is different than the mysql login name that should be with the ```<username>@<servername> ``` format. 
https://dev.mysql.com/doc/refman/5.7/en/create-user.html
https://dev.mysql.com/doc/refman/5.7/en/account-names.html

If the user was created without specify the '@' part, it will be created using the default format.  
-- By specify the 'interpolateParams' parameter and MySQL CreateUser and DropUser now use the SQL injection to create SQL query command.  

Before the PR: 
Use the sample yaml file to create a MySQL user (should create a MySQL server and MySQL database first). Then from the connect tool like MySQL Workbench you will see the MySQL User name format is like below: 

image

After the change: 
The create user is using the default format <username>@%.  

image

To further verify the user has been created correctly, use the created username and password to login to MySQL (using mysql workbench as an example). 
Please note that when login, the username should be entered with the format <username>@<mysqlserver> which is required by MySQL. 

image

You can successfully login with the newly create mysql user and if you have granted privilege to the user, you can execute corresponding actions. e.g. Select operation. 

Delete the user and you will get the mysql user removed from both cli and the MySQL workbench. 

The introduce of 'interpolateParams' won't impact the MySQL create and delete strategy but will make it safe. 

gif

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/pkg/resourcemanager/mysql/mysqluser/mysqluser.go
+++ b/pkg/resourcemanager/mysql/mysqluser/mysqluser.go
@@ -89,7 +89,7 @@ func (s *MySqlUserManager) GrantUserRoles(ctx context.Context, user string, data
 		if err := helpers.FindBadChars(role); err != nil {
 			return fmt.Errorf("Problem found with role: %v", err)
 		}
-		//TODO: how to use SQL injection for grant command, like CreateUser and DropUser
+		//TODO: how to use SQL parameters for grant command, like CreateUser and DropUser
 		tsql := fmt.Sprintf("GRANT %s ON `%s`.* TO '%s'", role, database, user)
 		_, err := db.ExecContext(ctx, tsql)
 		if err != nil {
--- a/pkg/resourcemanager/mysql/mysqluser/mysqluser.go
+++ b/pkg/resourcemanager/mysql/mysqluser/mysqluser.go
@@ -18,7 +18,7 @@ import (
 	"github.com/Azure/azure-service-operator/api/v1alpha1"
 	"k8s.io/apimachinery/pkg/runtime"
 
-	_ "github.com/go-sql-driver/mysql" //sql drive link
+	_ "github.com/go-sql-driver/mysql" //mysql drive link
 	"k8s.io/apimachinery/pkg/types"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/resourcemanager/mysql/mysqluser/mysqluser.go
```diff
@@ -126,30 +125,26 @@ func (s *MySqlUserManager) CreateUser(ctx context.Context, server string, secret
...
-func (s *MySqlUserManager) DropUser(ctx context.Context, db *sql.DB, user string, server string) error {
-	tsql := fmt.Sprintf("DROP USER IF EXISTS '%s'@'%s'", user, server)
+func (s *MySqlUserManager) DropUser(ctx context.Context, db *sql.DB, user string) error {
+
 	if err := helpers.FindBadChars(user); err != nil {
 		return fmt.Errorf("Problem found with username: %v", err)
 	}
-	_, err := db.ExecContext(ctx, tsql)
+	tsql := "DROP USER IF EXISTS ?"
+	_, err := db.ExecContext(ctx, tsql, user)
```
davefellows, line 147: Minor: the tsql string seems kind of redundant now...
```suggestion
	_, err := db.ExecContext(ctx, "DROP USER IF EXISTS ?", user)
```

pkg/resourcemanager/mysql/mysqluser/mysqluser.go
```diff
@@ -78,19 +78,19 @@ func (s *MySqlUserManager) ConnectToSqlDb(ctx context.Context, drivername string
...
 	}
 
 	for _, role := range roles {
-		tsql := fmt.Sprintf("GRANT %s ON `%s`.* TO '%s'@'%s'", role, database, user, server)
 
 		if err := helpers.FindBadChars(role); err != nil {
 			return fmt.Errorf("Problem found with role: %v", err)
 		}
-
+		//TODO: how to use SQL injection for grant command, like CreateUser and DropUser
```
theunrepentantgeek, line 92: "SQL Injection" is the attack vector we want to prevent; probably best to use a different term for the technique we're adopting to avoid the vulnerability. #minornit
```suggestion
		//TODO: how to use SQL parameters for grant command, like CreateUser and DropUser
```

Issue #1089, opened 2020-05-20
Bug: VM Operator should check if the VM is in 'Running' state and return successful resource provisioned status


huangpf:
**Describe the bug**
VM Operator should check if the VM is in 'Running' state and return successful resource provisioned status.

**To Reproduce**
Simply create a VM using the operator and observe how quick it's declared successfully provisioned.

**Expected behavior**
The VM should be in 'Running' state when the status above is declared.

**Screenshots**
```
11379 ± k get AzureVirtualMachine                                                                                                                                     ✹ ✭
NAME                PROVISIONED   MESSAGE
rg-vmextestin-vm3   true          successfully provisioned

that is 4 seconds from creation...you want the operator to reconcile until "Running" then set success
```

**Additional context**
Probably also applies to VMSS operator which is pending on further check.

A potential fix prototype could look like this:
https://github.com/Azure/azure-service-operator/blob/master/pkg/resourcemanager/vm/reconcile.go#L54
```
if err == nil {
    if (someProvisioningState == "Running") {
        instance.Status.Provisioned = true
        instance.Status.Provisioning = false
        instance.Status.Message = resourcemanager.SuccessMsg
        instance.Status.ResourceId = *item.ID
        return true, nil
    }
    else {
        // otherwise
        return false, nil
    }
}
```

huangpf:
Hi Admin -- please help assign this issue to me. :-)

huangpf:
Assign it to @haagha from the CPlat SDK Core team.

haagha:
PR https://github.com/Azure/azure-service-operator/pull/1222

haagha:
Can we close this issue? PR is merged

matthchr:
Closing based on above

PR #1162, opened 2020-06-17
Setting Provisioned When VM is running state [#1089]

haagha:
Closes [#1089]

Sets VM provisioned state only when client returns its provisioning state as successful

If there are any tests needed for this?

gif

**If applicable**:
- [ ] this PR contains tests


pkg/resourcemanager/vm/reconcile.go
```diff
@@ -52,9 +52,14 @@ func (g *AzureVirtualMachineClient) Ensure(ctx context.Context, obj runtime.Obje
 	// to overcome the issue with the lack of idempotence of the Create call
 	item, err := g.GetVirtualMachine(ctx, resourceGroup, resourceName)
 	if err == nil {
-		instance.Status.Provisioned = true
-		instance.Status.Provisioning = false
-		instance.Status.Message = resourcemanager.SuccessMsg
+		if *item.ProvisioningState == "Succeeded" {
```
huangpf, line 55: Is there an enum or const string you can use for the ```"Succeeded"``` string comparison?

haagha: yes i can fix that

pkg/resourcemanager/vm/reconcile.go
```diff
@@ -52,9 +52,14 @@ func (g *AzureVirtualMachineClient) Ensure(ctx context.Context, obj runtime.Obje
...
+			instance.Status.Provisioned = true
+			instance.Status.Provisioning = false
+			instance.Status.Message = resourcemanager.SuccessMsg
+		} else {
+			instance.Status.Provisioned = false
+			instance.Status.Provisioning = true
+		}
```
huangpf, line 62: @haagha -- I think the following lines should be included in the truly provisioned case?
```
		instance.Status.ResourceId = *item.ID
		return true, nil
```

And the other case should not return?

haagha: populated this and more status fields for this

haagha:
@frodopwns The pipeline has been running for a couple for hours. Do you know what is average time it takes?

haagha:
rerun CI

PR #1165, opened 2020-06-18
Allow creation of API keys for App Insights

frodopwns:
Allows for the creation of App Insights API keys. These keys allow applications to read telemetry data or write annotations to metrics charts.

Create an App Insights, then create an App Insights API Key.
Go to the App insights in the portal and look to the left column for API Access. Should see the key there.

Also a secret should be created with the api key in it.


frodopwns:
thanks @jananivMS 

Addressed the stack trace for 404s from creating the key. Also repackaged the helm chart.

pkg/resourcemanager/appinsights/api_keys_reconcile.go
```diff
@@ -0,0 +1,199 @@
...
+		opt(options)
+	}
+
+	if options.SecretClient != nil {
+		c.SecretClient = options.SecretClient
+	}
+
+	instance.Status.Provisioning = true
+
+	// we have reconciled this previously, check if it already exists
```
theunrepentantgeek, line 40:
```suggestion
	// we may have reconciled this previously, check if it already exists
```

pkg/resourcemanager/appinsights/api_keys_reconcile.go
```diff
@@ -0,0 +1,199 @@
...
+			instance.Spec.AppInsights,
+			idParts[len(idParts)-1],
+		)
+		if err == nil {
+			return true, nil
+		}
+
+		return false, nil
+
+	}
```
theunrepentantgeek, line 56: Remove extra blank lines #trivial-nit

pkg/resourcemanager/appinsights/api_keys_reconcile.go
```diff
@@ -0,0 +1,199 @@
...
+			instance.Spec.ResourceGroup,
+			instance.Spec.AppInsights,
+			idParts[len(idParts)-1],
+		)
+		if err == nil {
+			return true, nil
+		}
+
+		return false, nil
+
```
frodopwns, line 55:
```suggestion
```

pkg/resourcemanager/appinsights/api_keys_reconcile.go
frodopwns: done...I think there may be a better go fmt lib that can automate this...

PR #1168, opened 2020-06-19
Flatten PSQL Server Secret when stored in Key Vault

frodopwns:
Adds `flattenSecrets` field to spec to allow secret values to be placed in individual secrets in key vault. This allows users who use a syntax like `secret://` in env vars as a way to translate secret references to secret values to skip json parsing. JSON in the secret is trickier than just flat values.

One change this makes that could affect users in some way is that instead of replacing the normal secret with flattened secrets it  creates both when flatten is set to true. This allows us to still get data from the combined secret without thinking about where else it would be.


frodopwns:
reproduced this test failure locally...looking into it

pkg/secrets/keyvault/client.go
```diff
@@ -254,50 +251,78 @@ func (k *KeyvaultSecretClient) Upsert(ctx context.Context, key types.NamespacedN
...
+	}
 
-		_, err = k.KeyVaultClient.SetSecret(ctx, vaultBaseURL, secretBaseName, secretParams)
+	/*if _, err := k.KeyVaultClient.GetSecret(ctx, vaultBaseURL, secretBaseName, secretVersion); err == nil {
+		// If secret exists we delete it and recreate it again
+		_, err = k.KeyVaultClient.DeleteSecret(ctx, vaultBaseURL, secretBaseName)
+		if err != nil {
+			return fmt.Errorf("Upsert failed: Trying to delete existing secret failed with %v", err)
+		}
+	}*/
```
WilliamMortlMicrosoft, line 270: we should remove this

pkg/secrets/keyvault/client.go
```diff
@@ -254,50 +251,78 @@ func (k *KeyvaultSecretClient) Upsert(ctx context.Context, key types.NamespacedN
...
+	result, err := k.KeyVaultClient.GetSecret(ctx, vaultBaseURL, secretName, secretVersion)
+	if err != nil {
+		return fmt.Errorf("secret does not exist" + err.Error())
+	}
+
+	stringSecret := *result.Value
+	jsonErr := json.Unmarshal([]byte(stringSecret), &data)
+	if jsonErr != nil {
+		// return fmt.Errorf("secret improperly formatted: %v", jsonErr)
+		// this is fine, no fields means this is a flat secret
```
WilliamMortlMicrosoft, line 304: meant to be commented out?

frodopwns: I think I had intended to remove that

pkg/secrets/keyvault/client.go
```diff
@@ -254,50 +251,67 @@ func (k *KeyvaultSecretClient) Upsert(ctx context.Context, key types.NamespacedN
...
 			return err
 		}
 	}
 
+	if options.Flatten {
+		for formatName := range data {
+			sName := secretName + "-" + formatName
+			_, err := k.KeyVaultClient.DeleteSecret(ctx, vaultBaseURL, sName)
+			if err != nil {
+				return err
```
Porges, line 309: should this instead try to delete as many secrets as it can instead of aborting on first failure?

controllers/suite_test.go
```diff
@@ -824,7 +824,7 @@ func setup() error {
 	if result.Response.StatusCode != 204 {
 		_, err = resourceGroupManager.CreateGroup(context.Background(), resourceGroupName, resourcegroupLocation)
 		if err != nil {
-			return fmt.Errorf("ResourceGroup creation failed")
+			return fmt.Errorf("ResourceGroup creation failed: %v", err)
```
theunrepentantgeek, line 827: Suggest using %w to create a nested error object (preserves `err` in case it contains any extended information)

pkg/resourcemanager/azuresql/azuresqluser/azuresqluser.go
```diff
@@ -204,15 +204,17 @@ func (s *AzureSqlUserManager) DeleteSecrets(ctx context.Context, instance *v1alp
...
-
 			err = secretClient.Delete(
 				ctx,
 				key,
 			)
 			if err != nil {
-				instance.Status.Message = "failed to delete secret, err: " + err.Error()
-				return false, err
+				if !strings.Contains(err.Error(), "does not exist") {
+					instance.Status.Message = "failed to delete secret, err: " + err.Error()
```
theunrepentantgeek, line 213: Since this is a user facing error message, perhaps
```suggestion
					instance.Status.Message = "Failed to delete secret: " + err.Error()
```

pkg/resourcemanager/azuresql/azuresqluser/azuresqluser_reconcile.go
```diff
@@ -243,16 +243,18 @@ func (s *AzureSqlUserManager) Ensure(ctx context.Context, obj runtime.Object, op
...
+			err = sqlUserSecretClient.Upsert(
+				ctx,
+				types.NamespacedName{Namespace: key.Namespace, Name: instance.Name},
+				formattedSecrets,
+				secrets.WithOwner(instance),
+				secrets.WithScheme(s.Scheme),
+				secrets.Flatten(true),
+			)
+			if err != nil {
+				return false, err
```
theunrepentantgeek, line 256: Should this wrap `err` to give some context that the error happened while updating SQL secrets?

pkg/secrets/keyvault/client.go
```diff
@@ -160,31 +160,28 @@ func (k *KeyvaultSecretClient) Create(ctx context.Context, key types.NamespacedN
...
+	// Initialize secret parameters
+	secretParams := keyvaults.SecretSetParameters{
+		Value:            &stringSecret,
+		SecretAttributes: &secretAttributes,
+		ContentType:      &contentType,
+	}
 
-	return nil
+	if _, err := k.KeyVaultClient.GetSecret(ctx, vaultBaseURL, secretBaseName, secretVersion); err == nil {
+		return fmt.Errorf("secret already exists %v", err)
```
theunrepentantgeek, line 179: Suggest using `%w` to wrap the error
```suggestion
		return fmt.Errorf("secret already exists: %w", err)
```

pkg/resourcemanager/azuresql/azuresqluser/azuresqluser.go
frodopwns: Yeah, I think the majority of errors we set in Status that don't come directly from Azure are likely not capitalized...not necessarily opposed but changing this one probably won't do much more than add to the inconsistency.

matthchr:
Closing this as it's quite old and hasn't been merged, and we would prefer to do something generic and cross-cutting with respect to secrets.

PR #1174, opened 2020-06-23
Improve error when SQL server doesn't have RG specified

### LINKED ISSUE #1106: Bug: Handle `InvalidApiVersionParameter` on Empty ResourceGroup / Gracefully handle 400 errs
We should remove 'omitempty' from the types files for Resourcegroup from AzureSqlServer types. file and also other types if the issue exists. This will ensure that we catch the error if resourcegroup is not specified which causes this  issue.

**Describe the bug**

The following error occurs for creating a V1Alpha1 azuresqlserver with an empty resource group or invalid named resourcegroup.

```
"Error": "sql.ServersClient#Get: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code=\"InvalidApiVersionParameter\" Message=\"The api-version '2015-05-01-preview' is invalid. The supported versions are '2020-01-01,2019-11-01,2019-10-01,2019-09-01,2019-08-01,2019-07-01,2019-06-01,2019-05-10,2019-05-01,2019-03-01,2018-11-01,2018-09-01,2018-08-01,2018-07-01,2018-06-01,2018-05-01,2018-02-01,2018-01-01,2017-12-01,2017-08-01,2017-06-01,2017-05-10,2017-05-01,2017-03-01,2016-09-01,2016-07-01,2016-06-01,2016-02-01,2015-11-01,2015-01-01,2014-04-01-preview,2014-04-01,2014-01-01,2013-03-01,2014-02-26,2014-04'.\"", "Instance": "default/sqlserver-sample-777"}
```

**To Reproduce**
`k create -f config/samples/azure_v1alpha1_azuresqlserver.yaml`
My Resource Group Name: `resourcegroup-azure-operators-mel`

**Expected behavior**
Gracefully handle 400 errors (similarly to how Redis Cache handles future errs)

**Screenshots**

**Additional context**

matthchr:
This closes #1106

Improves ResourceGroup empty handling for SQL Server creation

None

Let me know if there should be a test here too.

### DIFF
--- a/api/v1alpha1/eventhub_types.go
+++ b/api/v1alpha1/eventhub_types.go
@@ -41,8 +41,7 @@ type EventHubStorageAccount struct {
 	// ResourceGroup - Name of the storage account resource group
 	// +kubebuilder:validation:Pattern=^[-\w\._\(\)]+$
 	// +kubebuilder:validation:MinLength:1
-	// +kubebuilder:validation:Required
-	ResourceGroup string `json:"resourceGroup"`
+	ResourceGroup string `json:"resourceGroup,omitempty"`
 	// AccountName - Name of the storage account
 	// +kubebuilder:validation:MaxLength=24
 	// +kubebuilder:validation:MinLength=3
--- a/api/v1alpha1/eventhub_types.go
+++ b/api/v1alpha1/eventhub_types.go
@@ -18,7 +18,10 @@ type EventhubSpec struct {
 	Location          string                    `json:"location"`
 	Namespace         string                    `json:"namespace,omitempty"`
 	Properties        EventhubProperties        `json:"properties,omitempty"`
-	ResourceGroup     string                    `json:"resourceGroup,omitempty"`
+	// +kubebuilder:validation:Pattern=^[-\w\._\(\)]+$
+	// +kubebuilder:validation:MinLength:1
+	// +kubebuilder:validation:Required
+	ResourceGroup     string                    `json:"resourceGroup"`
 	AuthorizationRule EventhubAuthorizationRule `json:"authorizationRule,omitempty"`
 	// SecretName - Used to specify the name of the secret. Defaults to Event Hub name if omitted.
 	SecretName             string `json:"secretName,omitempty"`
@@ -37,7 +40,9 @@ type EventhubAuthorizationRule struct {
 type EventHubStorageAccount struct {
 	// ResourceGroup - Name of the storage account resource group
 	// +kubebuilder:validation:Pattern=^[-\w\._\(\)]+$
-	ResourceGroup string `json:"resourceGroup,omitempty"`
+	// +kubebuilder:validation:MinLength:1
+	// +kubebuilder:validation:Required
+	ResourceGroup string `json:"resourceGroup"`
 	// AccountName - Name of the storage account
 	// +kubebuilder:validation:MaxLength=24
 	// +kubebuilder:validation:MinLength=3
--- a/pkg/resourcemanager/azuresql/azuresqlserver/azuresqlserver_reconcile.go
+++ b/pkg/resourcemanager/azuresql/azuresqlserver/azuresqlserver_reconcile.go
@@ -124,6 +124,12 @@ func (s *AzureSqlServerManager) Ensure(ctx context.Context, obj runtime.Object, 
 
 		// TODO: This is a bit of a hack because really this check should be in the SDK.
 		// TODO: See: https://github.com/Azure/azure-sdk-for-go/issues/10712
+		// Note that this can really only happen due to bad state in etcd because the
+		// CRD for the resource requires this field. Bad state in etcd can be caused by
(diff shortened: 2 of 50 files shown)


pkg/resourcemanager/azuresql/azuresqlserver/azuresqlserver_reconcile.go
```diff
@@ -122,6 +122,13 @@ func (s *AzureSqlServerManager) Ensure(ctx context.Context, obj runtime.Object,
 	if instance.Status.Provisioning ||
 		(!specHashWasEmpty && instance.Status.SpecHash == hash) {
 
+		// TODO: This is a bit of a hack because really this check should be in the SDK.
+		// TODO: See: https://github.com/Azure/azure-sdk-for-go/issues/10712
+		if instance.Spec.ResourceGroup == "" {
+			instance.Status.Message = "A nonempty resource group must be specified."
```
Porges, line 128: "non-empty"?

Porges:
You could also put `// +kubebuilder:validation:MinLength:1` on this

matthchr:
I added some additional Kubebuilder validation

Porges:
This did get quite repetitive; we could define a `type ResourceGroupName` and put the annotations on it. The downside is that this would need to reside in some shared schema.

api/v1alpha1/eventhub_types.go
```diff
@@ -37,7 +40,9 @@ type EventhubAuthorizationRule struct {
 type EventHubStorageAccount struct {
 	// ResourceGroup - Name of the storage account resource group
 	// +kubebuilder:validation:Pattern=^[-\w\._\(\)]+$
-	ResourceGroup string `json:"resourceGroup,omitempty"`
+	// +kubebuilder:validation:MinLength:1
+	// +kubebuilder:validation:Required
+	ResourceGroup string `json:"resourceGroup"`
```
jananivMS, line 45: resource group should be optional here. This is becasue the CaptureDescription in the EventthubProperties is optional and since EventHubStorageAccount is a field in CaptureDescription, tthis can be ommitted. This is what is causing the eventhub test in the pipeline to fail.

matthchr: Wow thanks for looking so in depth. Will fix!

matthchr: Fixed this and the other failing tests -- I think it should all pass now.

matthchr: Should we make the StorageAccount field in EventHub Destination a pointer instead so that it can be nil?

Presumably, if the destination is specified, RG is required then right?

matthchr: I've left this alone (not required) for now...

I've filed an issue #1182 for this across the repo though, as I think there are some improvements which could be made.

pkg/resourcemanager/azuresql/azuresqlserver/azuresqlserver_reconcile.go
```diff
@@ -122,6 +122,13 @@ func (s *AzureSqlServerManager) Ensure(ctx context.Context, obj runtime.Object,
 	if instance.Status.Provisioning ||
 		(!specHashWasEmpty && instance.Status.SpecHash == hash) {
 
+		// TODO: This is a bit of a hack because really this check should be in the SDK.
+		// TODO: See: https://github.com/Azure/azure-sdk-for-go/issues/10712
+		if instance.Spec.ResourceGroup == "" {
+			instance.Status.Message = "A non-empty resource group must be specified."
```
jananivMS, line 128: Why do we need this check if we have the validation now? The validation should catch the case if the RG was missing or empty, correct?

matthchr: Yeah that's a good question.

I've added a comment to the code which explains the situation, but basically the only way that this can happen is if an entity in etcd already exists without the `ResourceGroup` field set. This can happen due to an upgrade from an older version of ASO, or if the customer has disabled our conversion webhooks and put say an `v1alpha1` version of this resource into etcd rather than a `v1beta1` which is what we are using as the storage version.

This is possible because we renamed from `resourcegroup` to `resourceGroup` in the YAML/JSON for this resource. 

This check is really probably unlikely to be hit in most cases because it's kinda hard to be in a situation that causes it (but it is possible see #1106 which was caused (I think) due to conversion webhooks being disabled and using a non-default version which caused this to be empty - basically the field was in the CRD (so validation of CRD passes), but the field is not capitalized correctly (due to difference in CRD spec between v1alpha1 and v1beta1), so when we go to actually use it here we get empty string.

matthchr: After some more testing we actually don't even need this check, because missing RG is caught even in the case of older API versions earlier in the stack. I've removed this

PR #1184, opened 2020-06-25
Allow assignment of roles to Azure resources via operator

frodopwns:
Allows a RoleDefinitionID to be assigned to a Service Principal for an Azure Resource.

Currently, if a user wants to deploy something like a Key Vault and give a service principal other than the creating one access they would have to create the Key Vault, add the role out of band (manually?), then update the Key Vault request with the access policies that service principal needs (read, delete, set, etc).

This PR, along with another PR #1158 would allow a user to request a Key Vault with all the policies required by all its users right off the bat. The operator would create the Key Vault first and then attempt to add policies. It would retry the policies which would give the Role Assign operator time to give the other SPs access to be given the Key Vault access policies.

- Create some resource in Azure like a Resource Group.
- Create a RoleAssign request giving the same principal running the operator a new role on that resource. Something like "Reader" should suffice.
- Check the resource's access control in the portal and ensure the principal now has "Reader" as a role.


frodopwns:
still needs tests

matthchr:
I believe that this is superseded by #1848

matthchr:
Closing this as support has been added to ASO v2 to do this (but in a slightly different way than was proposed here)

PR #1193, opened 2020-06-26
Don't run devops pipeline if the only changes are docs changes

### LINKED ISSUE #1080: Task: Update pipeline to not run tests for doc changes
For doc changes, we do not need to run the tests, which would help save time since there isn't a real use of running tests for doc changes

**Done criteria**

- [ ] Pipeline tests do not run for doc changes

matthchr:
Closes #1080

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -2,6 +2,10 @@ trigger:
   branches:
     include:
     - master
+  paths:
+    exclude:
+    - docs/*
+    - README.md
 
 resources:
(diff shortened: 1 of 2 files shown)


matthchr:
I'm not actually sure how to test this...

frodopwns:
Looks like an improvement. Anything we can do about the README.md in the root?

matthchr:
Ah, good point about the README I had forgotten about that. I think I can ignore that too.

PR #1204, opened 2020-06-30
Update to azure pipelines file

matthchr:
Updates ADO to not build of the change is just a docs change

I have also updated the ADO job to stop overriding the yaml (prior to now, the section of the YAML specifying when to build was ignored because it was overridden in the ADO backend).

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,4 +1,12 @@
+# By default, ADO runs jobs on pushes and on PRs. Since we're only tracking/after PRs, turn off 
+# tiggers on push.
 trigger:
+  branches:
+    exclude:
+    - '*'
+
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -1,4 +1,12 @@
+# By default, ADO runs jobs on pushes and on PRs. Since we're only tracking/after PRs, turn off 
+# tiggers on push.
 trigger:
+  branches:
+    exclude:
```
matthchr, line 5: It is not _100%_ clear to me that we actually want to disable this?

It might break something like `/azp run` I am not sure... It was enabled manually in the backend only for master, so if folks prefer I can leave that configuration here.

Issue #1188, opened 2020-06-25
Provide example gatekeeper policies


mkosieradzki:
It would be great for multi-tenant clusters to have example (and supported) gatekeeper policies that enable namespace-based ownership of the resources, e.g. you can only access resources from Resource Groups with names dervived from your namespace using some conventions.

This is an alternative approach to: #1173

lwindolf:
I also consider this quite important to use the operator in a somewhat secure manner. Even with #1173 I might want to have gatekeeper rules just to uncover configuration mistakes.

matthchr:
This is still something we're tracking to do

matthchr:
We're still interested in doing this

theunrepentantgeek:
We're still interested in this.

matthchr:
We still are interested in doing this

matthchr:
Still interested in doing this. We would also be willing to accept contributions from the community if they have sample policies that they're already using in production.

theunrepentantgeek:
No change from the above.

theunrepentantgeek:
No change from the above.

theunrepentantgeek:
Still no change from the above.

matthchr:
Same as above.

matthchr:
No change from above -- let's try to get copilot to write (+ test?) some small documentation for this though so that we can close this out

Issue #1173, opened 2020-06-22
As a cluster admin, I can configure separate set of credentials for deployment from different AKS namespaces


hlarubi:
Hi,

I'm new to ASO, so I might have missed something that's already in place. If so, would appreciate the pointer in the right direction or docs that describe the config I'm looking for.

This user story covers the following:
Allowing cluster administrator to create multiple service principals / managed identities in a administrator provided namespace to ensure user deployments from each AKS namespace have specific permissions (ex: users can only deploy Azure services to a specific resource group when creating ASO custom resources).

The use of a single cluster identity for service provisioning across namespaces doesn't cleanly support cluster multi-tenancy, as user permissions in a given namespace can be leveraged to provision services across any Azure resource group, even the user doesn't have direct permission to this resource group.

The goal is to enable permission model that allows AKS namespaces to be used by multiple teams with different resource groups and Azure RBAC.

**Acceptance Criteria**
User can manage Azure services using Azure Service Operator via namespace provided service principal / managed identity

Different namespaces can be configured with their own security identity and corresponding Azure RBAC

Administrator has a way to manage assignment of access to Azure services on a per namespace basis (ex: via a set of configmaps / secrets in a "managed" namespace)

frodopwns:
@hlarubi 

How important is the part about having the SPs exist in a single admin namespace as opposed to putting them in the namespaces where they would be used?

hlarubi:
@frodopwns ,

We want to be able to map who can provision services to specific resource groups via RBAC. Unless there is a way to use the identity of the deployer to determine Azure permissions or ASO is using a unique resource type that I can limit RBAC for in the user namespace, I need a way to configure an admin namespace.

If there are other ideas, I'm happy to disucss.

Thanks!

eadasiak:
I have a similar need for the operator to support multitenancy, but a slight variation on the use case where use of the operator is scoped to service teams that have been provisioned a namespace.  If they have an Azure subscription and the credentials to do so, they would be able to opt-in to using it by configuring their namespace without cluster operator intervention.  RBAC within the namespace would determine which group members within a team had permissions to configure the namespace for use by the operator (e.g,. the namespace-admins group) and to create Azure resources (e.g., namespace-edit group).  I guess the main difference between what @hlarubi has described is that there's an assumption that each service team has their own Azure subscription and can manage it independently.

[...]

PR #1206, opened 2020-07-01
POC: Proposed solution for flexible azure auth

frodopwns:
Allows flexible authentication options for the Resource Group and Azure SQL operators as a way for proving the concept.

This PR adds 2 environment variables:
- AZURE_OPERATOR_AUTH_SOURCE_MODE
- AZURE_OPERATOR_AUTH_SOURCE_NAMESPACE

AZURE_OPERATOR_AUTH_SOURCE_MODE defaults to "fallback".
Potential values:
- `fallback`: Operators tries to find service principal in namespace where resource request is created. If not found, the operator tries to find a credential in an admin namespace if AZURE_OPERATOR_AUTH_SOURCE_NAMESPACE is set. If those fail, the operator falls back to the current default of using a global credential.
- `namespace`: Operator tries to find a credential named `aso-auth` in the same namespace as the request. This is a strict mode so no alternatives will be used.
- `admin-namespace`: Operator tries to find a credential in the admin namespace designated by the env var mentioned above. The credential secret in that namespace would be named `aso-auth-{target_namespace}`. This allows admins to store secrets that grant permissions for deployment in other namespaces. This is also a strict mode.
- `global`: Operator uses the current default which is to use a global service principal.

The Async Controller would source the credential before calling `Ensure` or `Delete`.

relates to #1173 
relates to #1188 

proposes a solution for allowing multiple service principals and later managed identities to be used in multi-tenant clusters

Look at the implementation and submit your feedback. This is a POC.


controllers/async_controller.go
```diff
@@ -211,3 +225,85 @@ func (r *AsyncReconciler) Reconcile(req ctrl.Request, obj runtime.Object) (resul
 
 	return result, err
 }
+
+// AuthSourceMode determines the strategy used to soruce auth credentials when authenticating with Azure
```
davefellows, line 229:
```suggestion
// AuthSourceMode determines the strategy used to source auth credentials when authenticating with Azure
```

matthchr:
@babbageclunk I think as discussed we can close this PR based on the work you're doing?

Issue #1090, opened 2020-05-20
Bug: VM/VMSS operators should handle 400 errors in a more graceful way


huangpf:
**Describe the bug**
VM/VMSS operators should handle 400 errors in a more graceful way

**To Reproduce**
Try to create a VM with some wrong settings in the NIC name, and image name, etc.

**Expected behavior**
The operator should error out, instead of spinning

**Screenshots**
```
2020-05-20T02:37:59.206Z    ERROR    controller-runtime.controller    Reconciler error    {"controller": "azurevirtualmachine", "request": "default/rg-vmextestin-vm2", "error": "2 errors occurred:\n\t* compute.VirtualMachinesClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code=\"NicInUse\" Message=\"Network Interface rg-vmextestin-nic is used by existing resource /subscriptions/7060bca0-7a3c-44bd-b54c-4bb1e9facfac/resourceGroups/resourcegroup-vmextestin/providers/Microsoft.Compute/virtualMachines/rg-vmextestin-vm. In order to delete the network interface, it must be dissociated from the resource. To learn more, see aka.ms/deletenic.\" Details=[]\n\t* Operation cannot be fulfilled on azurevirtualmachines.azure.microsoft.com \"rg-vmextestin-vm2\": the object has been modified; please apply your changes to the latest version and try again\n\n"}
github.com/go-logr/zapr.(*zapLogger).Error
    /go/pkg/mod/github.com/go-logr/zapr@v0.1.0/zapr.go:128
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
    /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.5.0/pkg/internal/controller/controller.go:258
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
    /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.5.0/pkg/internal/controller/controller.go:232
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker
    /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.5.0/pkg/internal/controller/controller.go:211
k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1
    /go/pkg/mod/k8s.io/apimachinery@v0.17.2/pkg/util/wait/wait.go:152
[...]

PR #1222, opened 2020-07-09
wait for VM running state Closes [#1089]

haagha:
Closes #[1089]
Tested #[1090] and cannot reproduce after change

Only returns successfully provisioned when VM is in running state

**If applicable**:
- [ ] this PR contains tests


pkg/resourcemanager/vm/reconcile.go
```diff
@@ -91,10 +101,19 @@ func (g *AzureVirtualMachineClient) Ensure(ctx context.Context, obj runtime.Obje
 			switch azerr.Type {
 			case errhelp.AsyncOpIncompleteError:
 				instance.Status.Provisioning = true
+				instance.Status.Message = err.Error()
```
frodopwns, line 104: This appears to already be set at the top of the error block on line 86

pkg/resourcemanager/vm/reconcile.go
```diff
@@ -91,10 +101,19 @@ func (g *AzureVirtualMachineClient) Ensure(ctx context.Context, obj runtime.Obje
...
+
 			}
 			// reconciliation is not done but error is acceptable
 			return false, nil
 		}
+		//change to spec is required
+		if future.Response().StatusCode == 400 {
+			instance.Status.FailedProvisioning = true
+			instance.Status.Message = err.Error()
```
frodopwns, line 113: same here, redundant setting of the error message

pkg/resourcemanager/vm/reconcile.go
```diff
@@ -52,11 +54,19 @@ func (g *AzureVirtualMachineClient) Ensure(ctx context.Context, obj runtime.Obje
...
+
+		if *item.ProvisioningState == SucceededProvisioningState {
+
+			instance.Status.Provisioned = true
+			instance.Status.Provisioning = false
+			instance.Status.Message = resourcemanager.SuccessMsg
+			instance.Status.ResourceId = *item.ID
+			return true, nil
+		}
+
```
frodopwns, line 66:
```suggestion
instance.Status.State = *item.ProvisioningState
```

Then in the status.Message you could be more verbose. "The requested resource has been requested but is not yet ready." If you want.

pkg/resourcemanager/vm/reconcile.go
```diff
@@ -91,10 +101,19 @@ func (g *AzureVirtualMachineClient) Ensure(ctx context.Context, obj runtime.Obje
...
 			case errhelp.AsyncOpIncompleteError:
 				instance.Status.Provisioning = true
+				instance.Status.Message = err.Error()
+
 			}
 			// reconciliation is not done but error is acceptable
 			return false, nil
 		}
+		//change to spec is required
+		if future.Response().StatusCode == 400 {
```
frodopwns, line 111: Some SDK clients don't return full future objects on failures. Have you tested any failure scenarios here to ensure that Response() doesn't cause a nil pointer dereference here?

pkg/resourcemanager/vm/reconcile.go
```diff
@@ -119,23 +138,25 @@ func (g *AzureVirtualMachineClient) Ensure(ctx context.Context, obj runtime.Obje
 			switch azerr.Type {
 			case errhelp.AsyncOpIncompleteError:
 				instance.Status.Provisioning = true
+				instance.Status.Message = err.Error()
```
frodopwns, line 141: This is already being set at the top of the error block.

pkg/resourcemanager/vm/reconcile.go
```diff
@@ -119,23 +138,25 @@ func (g *AzureVirtualMachineClient) Ensure(ctx context.Context, obj runtime.Obje
...
-	}
-	return true, nil
+	// if instance.Status.Provisioning {
+	// 	instance.Status.Provisioned = true
+	// 	instance.Status.Provisioning = false
+	// 	instance.Status.Message = resourcemanager.SuccessMsg
+	// } else {
+	// 	instance.Status.Provisioned = false
+	// 	instance.Status.Provisioning = true
+	// }
```
frodopwns, line 158: Any reason to keep this?

pkg/resourcemanager/vm/reconcile.go
haagha: removed, thanks for pointing out.

pkg/resourcemanager/vm/reconcile.go
haagha: removed

pkg/resourcemanager/vm/reconcile.go
haagha: removed

pkg/resourcemanager/vm/reconcile.go
haagha: removed

pkg/resourcemanager/vm/reconcile.go
haagha: Seems like compute sdk returns with status code because i can get the 400's. This addresses [#1090]

pkg/resourcemanager/vm/reconcile.go
haagha: added, thanks for the suggestion

Issue #1123, opened 2020-06-01
Task: Modify Managed Identity script to have option to create identity, and not deploy aad-pod-identity


cnadolny:
Currently, the managed identity script will provision a managed identity, assign it permissions, and then install and create the required aad-pod-identity components into the cluster.

It would be useful to give the script functionality, perhaps with a flag, to provision the identity and permissions, and then print out the resource and client ID to the command line for the user.

The user would then be able to take the output client and resource ID and be able to use it with the Helm Chart. 

The managed identity docs also do not provide manual instructions anymore on how to create a managed identity:
https://github.com/Azure/azure-service-operator/blob/master/docs/howto/managedidentity.md

**Done criteria**
This is the "done" criteria for this task

- [ ] Script is modified, and Helm docs are updated

r-t-m:
I would like to add that user can have aad-pod-identity already deployed in the cluster and the only thing needed in that case is to specify which managed identity to use without any dependency installation. Besides that this will allow aad pod identity to be updated separately faster than ASO maintainers update it the chart.

matthchr:
I removed the existing script entirely in favor of clearly documenting how to create an identity (if needed). This allows customers more freedom to pick exactly the permissions they want to assign to the identity.

See: https://github.com/Azure/azure-service-operator/blob/master/docs/howto/managedidentity.md

PR #1236, opened 2020-07-17
Make aad-pod-identity install optional when using Managed Identity.

kyschouv:
When using a Managed Identity for azure-service-operator, the chart currently automatically installs aad-pod-identity. However, a user may already have aad-pod-identity installed on their cluster, or may wish to manage the install of aad-pod-identity separately from azure-service-operator. This change to the helm chart adds a separate `installAadPodIdentity` flag, which is used to control the installation of aad-pod-identity, and leaves the `azureUseMI` flag to enable the AzureIdentity binding.

It would be great if the `aadpodidbinding` in the deployment was configurable as well (so the user could deploy with whatever label they'd like), but the template that includes that looks to be auto-generated, and so I left it outside the scope of this change. This change will also unblock my current integration, so I didn't want to delay it.

refs #1123


charts/index.yaml
```diff
@@ -3,19 +3,19 @@ entries:
...
     home: https://github.com/Azure/azure-service-operator
     name: azure-service-operator
     sources:
     - https://github.com/Azure/azure-service-operator
     urls:
-    - azure-service-operator-0.1.0.tgz
-    version: 0.1.0
-generated: "2020-07-08T10:09:00.877634-06:00"
+    - azure-service-operator-0.1.1.tgz
+    version: 0.1.1
```
frodopwns, line 20: The CI/CD around the helm chart is a bit simple right now. It doesn't do anything with versioning. As such, the tar being used in the pipeline has the 10.1.0` version hardcoded. 

If we want to adhere to semver in the helm chart we will need to ensure that our current usage of MCR allows storage of multiple versions. Also we need to decide if it's worth the trouble. Helm won't currently allow us to update or delete CRDs so the value proposition is dubious.

kyschouv: Ah, sorry. I'm used to the stable repo's CI/CD where they lint the chart and reject it without a version bump. Want me to revert the version number and repack the chart?

kyschouv:
I'm returning to working on migrating our old OSBA installation to ASO, and this is still causing trouble. Could I get some resolution on this, or perhaps some documentation on how to install without Helm?

matthchr:
Hey @kyschouv - I'll look through this PR and talk with @frodopwns about this.

Sorry about the delay, we've gone through a bit of a maintainer shakeup and are still figuring out some new processes as new and tooling which has caused us to be slow on reviewing PRs.

matthchr:
@kyschov - I pushed a merge of master with your branch so the diff was cleaner (just FYI)

PR #1242, opened 2020-07-20
Add support for SKU Database to support updates

### LINKED ISSUE #1229: Bug: Updating SKU/Edition field of AzureSQLDatabase doesn't do anything
**Describe the bug**
When updating the SKU/Edition field of the AzureSQLDatabase CRD, the updated field isn't getting propagated to Azure but the CRD switches to Provisioned status which is misleading.

**To Reproduce**
1. Deploy AzureSQLDatabase
2. Update SKU/Edition field, observe that it doesn't get applied to the backing Azure resource

**Expected behavior**
The update is applied

matthchr:
Closes #1229

Adds support for updating SQL Databases

Also improved test resource cleanup and usage of the ARM LRO polling client a bit

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,6 +97,11 @@ test-cleanup-azure-resources:
 		az group delete --name $$rgname --no-wait --yes; \
 	done
 
+	for rgname in `az group list --query "[*].[name]" -o table | grep 'rg-prime$$' `; do \
+		echo "$$rgname will be deleted"; \
+		az group delete --name $$rgname --no-wait --yes; \
+	done
+
 # Build the docker image
 docker-build:
 	docker build . -t ${IMG} ${ARGS}
--- a/pkg/resourcemanager/azuresql/azuresqldb/azuresqldb.go
+++ b/pkg/resourcemanager/azuresql/azuresqldb/azuresqldb.go
@@ -18,6 +18,9 @@ import (
 type AzureSqlDbManager struct {
 }
 
+// Ensure we implement the interface we expect
+var _ SqlDbManager = &AzureSqlDbManager{}
+
 func NewAzureSqlDbManager() *AzureSqlDbManager {
 	return &AzureSqlDbManager{}
 }
@@ -52,30 +55,30 @@ func (_ *AzureSqlDbManager) GetDB(ctx context.Context, resourceGroupName string,
 }
 
 // DeleteDB deletes a DB
-func (sdk *AzureSqlDbManager) DeleteDB(ctx context.Context, resourceGroupName string, serverName string, databaseName string) (result *http.Response, err error) {
-	// TODO: Probably shouldn't return a response at all in the err case here (all through this function)
-	result = &http.Response{
-		StatusCode: 200,
-	}
+func (sdk *AzureSqlDbManager) DeleteDB(
+	ctx context.Context,
+	resourceGroupName string,
+	serverName string,
+	databaseName string) (future *sql.DatabasesDeleteFuture, err error) {
 
 	// check to see if the server exists, if it doesn't then short-circuit
 	server, err := sdk.GetServer(ctx, resourceGroupName, serverName)
 	if err != nil || *server.State != "Ready" {
-		return result, nil
+		return nil, nil
 	}
 
 	// check to see if the db exists, if it doesn't then short-circuit
 	_, err = sdk.GetDB(ctx, resourceGroupName, serverName, databaseName)
 	if err != nil {
-		return result, nil
+		return nil, nil
 	}
 
 	dbClient, err := azuresqlshared.GetGoDbClient()
 	if err != nil {
-		return result, err
+		return nil, err
 	}
 
-	future, err := dbClient.Delete(
+	result, err := dbClient.Delete(
 		ctx,
 		resourceGroupName,
 		serverName,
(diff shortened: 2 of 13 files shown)


Makefile
```diff
@@ -97,6 +97,11 @@ test-cleanup-azure-resources:
 		az group delete --name $$rgname --no-wait --yes; \
 	done
 
+	for rgname in `az group list --query "[*].[name]" -o table | grep 'rg-prime$$' `; do \
+    	echo "$$rgname will be deleted"; \
+    	az group delete --name $$rgname --no-wait --yes; \
+    done
```
frodopwns, line 103: need an explanation here since this change appears to be out of scope

matthchr: I can break it out into it's own PR.

Basically, running the tests can leak resources (at least locally?) because the tests are suffixed with rg-prime. Let me know if you want me to move this to its own PR.

Makefile
```diff
@@ -97,6 +97,11 @@ test-cleanup-azure-resources:
 		az group delete --name $$rgname --no-wait --yes; \
 	done
 
+	for rgname in `az group list --query "[*].[name]" -o table | grep 'rg-prime$$' `; do \
```
Porges, line 100: 💄  inconsistent indentation

pkg/resourcemanager/azuresql/azuresqldb/azuresqldb.go
```diff
@@ -18,6 +18,9 @@ import (
 type AzureSqlDbManager struct {
 }
 
+// Ensure we implement the interface we expect
+var _ SqlDbManager = &AzureSqlDbManager{}
```
frodopwns, line 22: I'm not familiar with this syntax, what is the "_" doing? I would say its generally more go-like to use the New function than to use a global constant.

pkg/resourcemanager/azuresql/azuresqldb/azuresqldb_reconcile.go
```diff
@@ -69,66 +66,107 @@ func (db *AzureSqlDbManager) Ensure(ctx context.Context, obj runtime.Object, opt
...
 		// resource request has been sent to ARM
 		if azerr.Type == errhelp.AsyncOpIncompleteError {
+			instance.Status.Message = "Resource request successfully submitted to Azure"
 			instance.Status.Provisioning = true
+			instance.Status.PollingURL = pollingUrl
 			return false, nil
 		}
 
+		if azerr.Type == errhelp.InvalidMaxSizeTierCombination {
+			instance.Status.FailedProvisioning = true
```
Porges, line 165: should we hide these two status fields from modification and instead provide methods that update them correctly?

pkg/resourcemanager/azuresql/azuresqldb/azuresqldb.go
matthchr: This is a type assertion asking the compiler to ensure that we've implemented this interface. The compiler elides it at compile time (because we're assigning to a `_` variable).

It's useful in places where you have a type that impl's an interface and want a more C# (or pick your OO language of choice) interface (i.e. the type implementing the interface knows that it is doing so).

In this case, one of my previous PRs had actually changed a method signature on the AzureSqlDbManager struct but missed changing it in the SqlDbManager interface -- this is just a trick so that the compiler catches that if we make the same mistake in the future.

Porges: it is asserting that the interface `SqlDbManager` is implemented correctly; if it isn’t then this line will fail to compile

pkg/resourcemanager/azuresql/azuresqldb/azuresqldb_reconcile.go
matthchr: We probably should, yes -- and maybe in the future consider merging some of them (i.e. `Provisioning` and `FailedProvisioning` could be like `ProvisioningStatus` as an enum instead of 2 bools that represent 3 distinct states (since 1 of the 4 is invalid).

matthchr: Probably in a different PR though>

pkg/resourcemanager/azuresql/azuresqldb/azuresqldb.go
Porges: [snap] 🐊

pkg/resourcemanager/azuresql/azuresqldb/azuresqldb_reconcile.go
Porges: sure

matthchr: I'll file an issue for it

frodopwns:
Created a Basic DB and then attempted to update it with a new MaxSize. Got this error: `Code="InvalidMaxSizeTierCombination" Message="The tier 'Basic' does not`

Bit, failedProvisioning was not set and the operator continued to reconcile continuously.

```
status:
  message: Code="InvalidMaxSizeTierCombination" Message="The tier 'Basic' does not
    support the database max size '4294967296'."
  pollingUrl: https://management.azure.com/subscriptions/7/resourceGroups/resourcegroup-skukazoo/providers/Microsoft.Sql/locations/westus/databaseAzureAsyncOperation/d0d85658-2f37-4426-8312-f2f15362f019?api-version=2017-10-01-preview
  resourceId: /subscriptions/e9facfac/resourceGroups/resourcegroup-skukazoo/providers/Microsoft.Sql/servers/sqlserver-skukazoo/databases/azuresqldatabase-skukazoo
  specHash: f48f4f25ea852ea25a850285d7317ad3cda293efb19017d430551cce552f785c
  state: Online
```

pkg/resourcemanager/azuresql/azuresqldb/azuresqldb_reconcile.go
matthchr: https://github.com/Azure/azure-service-operator/issues/1243

matthchr:
I fixed the issue @frodopwns mentioned where the state wasn't set quite right.

PR #1246, opened 2020-07-29
Add support for allocation of public ips with ip-tags

MarcPow:
Closes #1245

**What this PR does / why we need it**:
This PR allows public ip addresses to be allocated using ip-tags

**Special notes for your reviewer**:
I've chosen to add this as an option to v1alpha1, as opposed to forking a new version into v1beta, given the maturity of the operator and considering that this is purely additive.

**How does this PR make you feel**:
Happy!

**If applicable**:
- [ ] this PR contains documentation
- [x] this PR contains tests

PR #1249, opened 2020-08-05
Change the webhooks port to 9443

### LINKED ISSUE #1177: Bug: Operator fails to start (port 433) when PodSecurity or SCC is enabled on the cluster
**Describe the bug**
When running operator in environment where Pod Security Policy or SCC in OpenShift are enabled , operator pod crashes as it can't bind to privileged port
```
2020-06-22T13:23:53.834Z         INFO          controller-runtime.controller  Starting EventSource                   {"controller": "resourcegroup", "source": "kind source: /, Kind="}
2020-06-22T13:23:53.834Z         ERROR       setup         problem running manager         {"error": "listen tcp :443: bind: permission denied"}
github.com/go-logr/zapr.(*zapLogger).Error
                    /go/pkg/mod/github.com/go-logr/zapr@v0.1.0/zapr.go:128
main.main
                    /workspace/main.go:790
runtime.main
                    /usr/local/go/src/runtime/proc.go:203

```

**Expected behavior**
Expected Operator to use port > 1024 range and run in non-privilege mode

babbageclunk:
Closes #1177

It changes the webhook port on which the operator gets validation requests from 443 to 9443. This means that it doesn't need root privileges, which in turn means the pod can run in more restrictive environments like SCC on OpenShift.

Split out the nasty helm chart churn into a separate PR #1250.

Delighted!

**If applicable**:
- [ ] this PR contains tests


charts/azure-service-operator/crds/apiextensions.k8s.io_v1beta1_customresourcedefinition_azurepublicipaddresses.azure.microsoft.com.yaml
```diff
@@ -46,9 +41,11 @@ spec:
           properties:
             idleTimeoutInMinutes:
               type: integer
+            ipTags:
+              additionalProperties:
+                type: string
+              type: object
```
theunrepentantgeek, line 47: New properties on purpose?

babbageclunk: `ipTags` was added in #1246 - it's a `map[string]string` so I think this is right.

PR #1252, opened 2020-08-07
update port for webhook service to 443

### LINKED ISSUE #1251: samples are not working
**Describe the bug**
I am getting errors while trying out almost all the samples mentioned over here.
https://github.com/Azure/azure-service-operator/blob/master/config/samples

**To Reproduce**
example: 

1. deploy https://github.com/Azure/azure-service-operator/blob/master/config/samples/azure_v1alpha1_postgresqlserver.yaml

```
Error from server: error when creating "postgres.yml": conversion webhook for azure.microsoft.com/v1alpha1, Kind=PostgreSQLServer failed: Post https://azureoperator-webhook-service.azureoperator-system.svc:443/convert?timeout=30s: no service port 'ƻ' found for service "azureoperator-webhook-service"
```

2. deploy https://github.com/Azure/azure-service-operator/blob/master/config/samples/azure_v1alpha1_azuresqlserver.yaml

```
 k get azuresqldatabase.azure.microsoft.com/azuresqldatabase-sample

NAME                              PROVISIONED   MESSAGE
azuresqldatabase-sample-embrace                 sql.DatabasesClient#CreateOrUpdate: Failure sending request: StatusCode=404 -- Original Error: Code="ParentResourceNotFound" Message="Can not perform requested operation on nested resource. Parent resource 'sqlserver-sample-777' not found."
```

**Expected behavior**
it should create the resource in the azure successfully.

**Screenshots**

**Additional context**

madorn:
Update webhook service to listen on 443. This will prevent the need to update the current CRDs that specify 443 in their conversion config.

Closes #1251 

Without this change, creating Custom Resources that have CRD conversion webhook configs will fail.

:)

### DIFF
--- a/charts/azure-service-operator/templates/generated/~g_v1_service_azureoperator-webhook-service.yaml
+++ b/charts/azure-service-operator/templates/generated/~g_v1_service_azureoperator-webhook-service.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
 spec:
   ports:
-  - port: 9443
+  - port: 443
     targetPort: 9443
   selector:
     control-plane: controller-manager
--- a/config/webhook/service.yaml
+++ b/config/webhook/service.yaml
@@ -6,7 +6,7 @@ metadata:
   namespace: system
 spec:
   ports:
-    - port: 9443
+    - port: 443
       targetPort: 9443
   selector:
     control-plane: controller-manager


theunrepentantgeek:
There's a conflict here with recent PR #1249 which was all about moving away from port 443 so that the operator can be used in more restrictive environments. @babbageclunk will know more.

babbageclunk:
> There's a potential conflict here with recent PR #1249 which was all about moving away from port 443 so that the operator can be used in more restrictive environments. @babbageclunk will know more.

It turns out my change was overenthusiastic about updating ports to 9443. It should just have been to make the operator pod listen on 9443 (so it can start in environments where it's not running as root), but leave the webhook service listening on 443, because existing resources are configured with that port by default for the conversion webhook. (See `conversion.webhookClientConfig.service` in the generated CRD.)

babbageclunk:
/azp run

I'm going to land this change and then add another PR to update the charts.

PR #1260, opened 2020-08-27
[Bug fix] Azure SQL Managed Identity can get stuck in deleting when the database the user is in is deleted

### LINKED ISSUE #1254: Bug: When deleting AzureSQLDatabase and AzureSQLManagedUser at the same time, sometimes the managed user can get stuck
**Describe the bug**
When deleting AzureSQLDatabase and AzureSQLManagedUser at the same time, sometimes the managed user can get stuck.

**To Reproduce**
1. Create an Azure SQL DB and managed user, and delete both at the same time, or delete the DB first, wait for it to be in deleting state/deleted, and then delete the managed user.
2. Observe that sometimes the managed user gets stuck.

**Expected behavior**
Both are deleted

**Additional context**
Can see this as the state of the user:
```
k get azuresqlmanagedusers -n klt6-private-predev
NAME      PROVISIONED   MESSAGE
sqluser   true          PingContext failed: Login error: mssql: Login failed for user '<token-identified principal>'.
```

matthchr:
Closes #1254

**What this PR does / why we need it**: A few makefile cleanups and fixing a bug causing resource to get stuck in deleting state.

We currently cannot write an automated test for this scenario because the SqlManagedIdentity user only works in a real AKS cluster on Azure (due to how ManagedIdentity works). I tested it manually before sending the PR.

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ BUILD_ID ?= $(shell git rev-parse --short HEAD)
 
 # best to keep the prefix as short as possible to not exceed naming limits for things like keyvault (24 chars)
 TEST_RESOURCE_PREFIX ?= aso-$(BUILD_ID)
+KIND_CLUSTER_NAME = aso-kind
 
 # Go compiler builds tags: some parts of the test suite use these to selectively compile tests.
 BUILD_TAGS ?= all
@@ -212,17 +213,24 @@ install-bindata:
 generate-template:
 	go-bindata -pkg template -prefix pkg/template/assets/ -o pkg/template/templates.go pkg/template/assets/
 
-create-kindcluster:
-ifeq (,$(shell kind get clusters))
-	@echo "no kind cluster"
-else
-	@echo "kind cluster is running, deleteing the current cluster"
-	kind delete cluster
-endif
-	@echo "creating kind cluster"
-	kind create cluster
+# TODO: These kind-delete / kind-create targets were stolen from k8s-infra and
+# TODO: should be merged back together when the projects more closely align
+.PHONY: kind-delete
+kind-delete: install-kind
+	kind delete cluster --name=$(KIND_CLUSTER_NAME) || true
 
-set-kindcluster: install-kind
+.PHONY: kind-create
+kind-create: install-kind
+	kind get clusters | grep -E $(KIND_CLUSTER_NAME) > /dev/null;\
+	EXISTS=$$?;\
+	if [ $$EXISTS -eq 0 ]; then \
+		echo "$(KIND_CLUSTER_NAME) already exists"; \
+	else \
+		kind create cluster --name=$(KIND_CLUSTER_NAME); \
+	fi; \
+
+.PHONY: set-kindcluster
+set-kindcluster: install-kind kind-create
 ifeq (${shell kind get kubeconfig-path --name="kind"},${KUBECONFIG})
 	@echo "kubeconfig-path points to kind path"
 else
@@ -230,8 +238,6 @@ else
 	@echo  "\e[31mexport KUBECONFIG=$(shell kind get kubeconfig-path --name="kind")\e[0m"
 	@exit 111
 endif
-	make create-kindcluster
-
 	@echo "getting value of KUBECONFIG"
 	@echo ${KUBECONFIG}
 	@echo "getting value of kind kubeconfig-path"
@@ -263,7 +269,7 @@ endif
 install-kind:
 ifeq (,$(shell which kind))
 	@echo "installing kind"
-	GO111MODULE="on" go get sigs.k8s.io/kind@v0.4.0
+	GO111MODULE="on" go get sigs.k8s.io/kind@v0.8.1
 else
 	@echo "kind has been installed"
 endif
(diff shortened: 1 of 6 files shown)


pkg/resourcemanager/azuresql/azuresqlmanageduser/azuresqlmanageduser_reconcile.go
```diff
@@ -146,7 +146,24 @@ func (s *AzureSqlManagedUserManager) Delete(ctx context.Context, obj runtime.Obj
 		requestedUsername = instance.Name
 	}
 
-	db, err := s.ConnectToSqlDbAsCurrentUser(ctx, DriverName, instance.Spec.Server, instance.Spec.DbName)
+	// short circuit connection if database doesn't exist
+	_, err = s.GetDB(ctx, instance.Spec.ResourceGroup, instance.Spec.Server, instance.Spec.DbName)
```
babbageclunk, line 150: Was the underlying problem that the connect call would hang if the database doesn't exist? Should we also add a timeout on the connect? It seems like there could be other problems where the DB exists but connect still hangs - presumably that would block reconciliation.

babbageclunk (approved):
Looks good, but maybe a timeout would be worthwhile too.

Makefile
```diff
@@ -212,26 +212,30 @@ install-bindata:
...
+kind-create: install-kind
+	kind get clusters | grep -E $(KIND_CLUSTER_NAME) > /dev/null;\
+	EXISTS=$$?;\
+	if [ $$EXISTS -eq 0 ]; then \
+		echo "$(KIND_CLUSTER_NAME) already exists"; \
+	else \
+		kind create cluster --name=$(KIND_CLUSTER_NAME); \
+	fi; \
+
+set-kindcluster: install-kind kind-create
```
Porges, line 231: This should be phony too?

matthchr: Yes

pkg/resourcemanager/azuresql/azuresqlmanageduser/azuresqlmanageduser_reconcile.go
matthchr: Doesn't hang - errored with connection rejected.

Porges (approved):

pkg/resourcemanager/azuresql/azuresqlmanageduser/azuresqlmanageduser_reconcile.go
babbageclunk: Oh ok.

pkg/resourcemanager/azuresql/azuresqlmanageduser/azuresqlmanageduser_reconcile.go
```diff
@@ -146,7 +146,24 @@ func (s *AzureSqlManagedUserManager) Delete(ctx context.Context, obj runtime.Obj
...
+		catch := []string{
+			errhelp.ResourceNotFound,
+			errhelp.ParentNotFoundErrorCode,
+			errhelp.ResourceGroupNotFoundErrorCode,
+		}
+		azerr := errhelp.NewAzureErrorAzureError(err)
+		if helpers.ContainsString(catch, azerr.Type) {
+			return false, nil
+		}
+		return false, err
```
babbageclunk, line 163: Sorry I missed this first time around, but since this case is a delete success (because the DB's gone) I think you need to do the secret deletion for it as well.

matthchr: Woaah, I totally missed that. Good catch.

pkg/resourcemanager/azuresql/azuresqlmanageduser/azuresqlmanageduser_reconcile.go
matthchr: It looks like the AutoRest clients already have timeouts configured:
```
	transport := &http.Transport{
		Proxy:                 defaultTransport.Proxy,
		DialContext:           defaultTransport.DialContext,
		MaxIdleConns:          defaultTransport.MaxIdleConns,
		IdleConnTimeout:       defaultTransport.IdleConnTimeout,
		TLSHandshakeTimeout:   defaultTransport.TLSHandshakeTimeout,
		ExpectContinueTimeout: defaultTransport.ExpectContinueTimeout,
		TLSClientConfig: &tls.Config{
			MinVersion:    tls.VersionTLS12,
			Renegotiation: renengotiation,
		},
	}
```

Values seem to be this:
```
	Proxy: ProxyFromEnvironment,
	DialContext: (&net.Dialer{
		Timeout:   30 * time.Second,
		KeepAlive: 30 * time.Second,
		DualStack: true,
	}).DialContext,
	ForceAttemptHTTP2:     true,
	MaxIdleConns:          100,
	IdleConnTimeout:       90 * time.Second,
	TLSHandshakeTimeout:   10 * time.Second,
	ExpectContinueTimeout: 1 * time.Second,
```

So I think we're good there.

babbageclunk (approved):
Looks great!

PR #1266, opened 2020-09-08
Add standard SUPPORT.md file

theunrepentantgeek:

The Microsoft Open Source Program Office is recommending the use of a `SUPPORT.md` file to make it easy for users to see what level of support each repo provides.

I've taken their supplied template and customized it with information already present in our `README.md`.

Are there any other support resources/routes that should be included?

gif

**If applicable**:

### DIFF
--- a/SUPPORT.md
+++ b/SUPPORT.md
@@ -1,0 +1,9 @@
+# Support
+
+## How to file issues and get help
+
+This project uses GitHub Issues to track bugs and feature requests. Please [search](https://github.com/Azure/azure-service-operator/issues) the existing issues before filing new issues to avoid duplicates. For new issues, file your bug or feature request as [a new issue](https://github.com/Azure/azure-service-operator/issues/new/choose). The Azure Service Operator project maintainers will respond to the best of their abilities.
+
+## Microsoft Support Policy
+
+Support for this project/product is limited to the resources listed above.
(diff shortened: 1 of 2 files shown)


Porges (dismissed):

SUPPORT.md
```diff
@@ -0,0 +1,9 @@
+# Support
+
+## How to file issues and get help
+
+This project uses GitHub Issues to track bugs and feature requests. Please [search](https://github.com/Azure/azure-service-operator/issues) the existing issues before filing new issues to avoid duplicates. For new issues, file your bug or feature request as [a new issue](https://github.com/Azure/azure-service-operator/issues/new/choose). The Azure Service Operator project maintainers will respond to the best of their abilities.
```
matthchr, line 5: Should we remove this information from the read me (and replace with a section linking here?)

Just reducing document duplication?

Also I see the README says:
> Azure Service Operator is an open source project that is **not** covered by the [Microsoft Azure support policy](https://support.microsoft.com/en-us/help/2941892/support-for-linux-and-open-source-technology-in-azure). 

Should be include that here too?

matthchr (changes requested):
See comment about reducing document duplication

babbageclunk (dismissed):
I like Matt's suggestion about pointing the readme at this too.

SUPPORT.md
theunrepentantgeek: I've added that link to `SUPPORT.md`.

Won't remove the duplication from the `README.md` as the goal is to make support details more discoverable, not to require people to look in one specific place.

matthchr (approved):

PR #1275, opened 2020-09-21
Makefile targets should be PHONY

matthchr:
Makefile should specify PHONY for targets which don't produce a file of the given name (in this case, that's all of them).


babbageclunk (approved):

Makefile
```diff
@@ -208,7 +231,7 @@ endif
 install-bindata:
 	go get -u github.com/jteeuwen/go-bindata/...
 
-.PHONE:
```
Porges, line 211: 😁

Porges (approved):

PR #1276, opened 2020-09-21
Handle SQL DB conflicts (and other errors)

### LINKED ISSUE #1274: Bug: SQL DB reconcile can get stuck due to errors with long-running operation
**Describe the bug**
SQL DBs can get stuck due to certain errors which can happen to their LROs when provisioning.

**To Reproduce**
This is most commonly observed when there are multiple things attempting to manage a single object, as is the case in a multi-cluster setup or if there are manual actions being performed on a SQL Server at the same time the operator is attempting to reconcile a database of that server.

1. Create DB in server s1 with operator deployment 1
2. Create same DB in server s2 with operator deployment 2.

Observe that sometimes the second operator deployment gets an error related to `ConflictingDatabaseOperation`

**Expected behavior**
Resource should properly reconcile.

matthchr:
Closes #1274

SQL Databases can sometimes hit an error when attempting to modify them in quick succession. This can happen for a variety of reasons but is common in situations where there are multiple managers reconciling against the same Azure resource. Before this fix when this happened the DB would just get stuck forever.

This PR fixes the above issues and allows the operator to retry when this problem occurs.

This issue is difficult to write a test for given the current test framework. We should spend some time improving the framework to allow for this sort of test (probably after generator lands?)

### DIFF
--- a/pkg/errhelp/errors.go
+++ b/pkg/errhelp/errors.go
@@ -69,7 +69,7 @@ const (
 	OperationIdNotFound                            = "OperationIdNotFound"
 )
 
-func NewAzureError(err error) error {
+func NewAzureError(err error) *AzureError {
 	var kind, reason string
 	if err == nil {
 		return nil
@@ -146,13 +146,6 @@ func NewAzureError(err error) error {
 	ae.Type = kind
 
 	return &ae
-}
-
-func NewAzureErrorAzureError(err error) *AzureError {
-	if err == nil {
-		return nil
-	}
-	return NewAzureError(err).(*AzureError)
 }
 
 type AzureError struct {
--- a/pkg/errhelp/errors.go
+++ b/pkg/errhelp/errors.go
@@ -66,6 +66,7 @@ const (
 	FeatureNotSupportedForEdition                  = "FeatureNotSupportedForEdition"
 	VirtualNetworkRuleBadRequest                   = "VirtualNetworkRuleBadRequest"
 	LongTermRetentionPolicyInvalid                 = "LongTermRetentionPolicyInvalid"
+	OperationIdNotFound                            = "OperationIdNotFound"
 )
 
 func NewAzureError(err error) error {
@@ -148,6 +149,9 @@ func NewAzureError(err error) error {
 }
 
 func NewAzureErrorAzureError(err error) *AzureError {
+	if err == nil {
+		return nil
+	}
 	return NewAzureError(err).(*AzureError)
 }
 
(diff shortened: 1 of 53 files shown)


pkg/errhelp/errors.go
```diff
@@ -148,6 +149,9 @@ func NewAzureError(err error) error {
 }
 
 func NewAzureErrorAzureError(err error) *AzureError {
+	if err == nil {
```
babbageclunk, line 152: Nice. The name of this totally threw me but I can see why it's like that.

pkg/resourcemanager/azuresql/azuresqldb/azuresqldb_reconcile.go
```diff
@@ -68,9 +68,15 @@ func (db *AzureSqlDbManager) Ensure(ctx context.Context, obj runtime.Object, opt
 
 	// Before we attempt to issue a new update, check if there is a previously ongoing update
 	if instance.Status.PollingURL != "" {
+		// TODO: There are other places which use PollClient which may or may not need this treatment as well...
 		pClient := pollclient.NewPollClient()
 		res, err := pClient.Get(ctx, instance.Status.PollingURL)
-		if err != nil {
+		pollErr := errhelp.NewAzureErrorAzureError(err)
```
babbageclunk, line 74: Can Get ever return a non-AzureError error error?

babbageclunk (approved):

pkg/resourcemanager/azuresql/azuresqldb/azuresqldb_reconcile.go
matthchr: The way `NewAzureErrorAzureError` is written, even if the error doesn't match one of the predefined types it does its best to match what it can, so it won't blow up.

I don't think that get will return anything other than `autorest.DetailedError` (which is what `NewAzureErrorAzureError` parses), and there are other places in the code where I see `NewAzureErrorAzureError` has been used on the result of a call to the polling client... so I think we're ok here.

babbageclunk: That makes sense - thanks for the extra detail!

pkg/errhelp/errors.go
theunrepentantgeek: The name _completely_ mystifies me - and I've spent 15m looking at this and the use of it to try and get my head around it.

At minimum, it needs a comment to explain what it does; better if it's given a name that actually expresses what it does.

theunrepentantgeek: Given other Go code that accepts `error` and returns something else, maybe the name needs to start with `Wrap`?

theunrepentantgeek (changes requested):

pkg/errhelp/errors.go
matthchr: I've renamed this to `NewAzureError` and combined this and the existing `NewAzureError` since there's no reason to have the method that returns error (since `*AzureError` impls it anyway).

theunrepentantgeek (approved):
Thanks for renaming that method, makes more sense now. Approved!

PR #1278, opened 2020-09-24
Remove dummy caBundle value in webhook config

### LINKED ISSUE #1261: Bug: Reapplying CRDs with conversion webhooks + certmanager causes re-reconciliation
**Describe the bug**
Users have reported seeing >500 reconciliations of the AzureSqlDb object over 48 hours, even though the resource was in steady state. It's unclear if this issue is limited to AzureSqlDb or if it can happen to any ASO managed entity.

**To Reproduce**
Exact steps to reproduce this issue are unknown. First pass should be to allocate a SQL Server and SQL DB and wait to see if we can get them to reconcile frequently.

**Expected behavior**
Resources are reconciled only when they are changed or infrequently to monitor backend state.

matthchr:
- This prevents resourceVersion change when CRD is reapplied.
  - See https://github.com/kubernetes-sigs/kubebuilder/issues/1684 for more details.
  - Fixes #1261.


babbageclunk (approved):
Nice! I love when a fix is essentially "stop doing this thing that isn't needed anymore and everything will just work".

davefellows (approved):

theunrepentantgeek (approved):

PR #1283, opened 2020-10-01
Add helpers to help avoid conflicting states

matthchr:
Initial framework to avoid nonsensical Provisioning/Provisioned/FailedProvisioning states, at least until we make it an enum.

We ought to consider using these helpers in other resources too. I can file an item for it once this PR is accepted as well.

**If applicable**:
- [x] this PR contains tests


api/v1beta1/aso_types.go
```diff
@@ -25,6 +25,27 @@ type ASOStatus struct {
 	Output             string       `json:"output,omitempty"`
 }
 
+func (s *ASOStatus) SetProvisioned(msg string) {
```
davefellows, line 28: Does this need to be duplicated i.e. is there not somewhere "helper" functions that will (likely) never change between versions can go?

matthchr: What type would that helper function operator on? `v1alpha1.Status` and `v1beta1.Status` are two different types so they need two functions even if the contents of those functions is the same.

I could've avoided adding it to the `v1alpha1` type if we didn't convert back to that type in async controller, but unfortunately we do -- and having the helpers there will be useful in the future if we want to make use of them on the `v1alpha1` types as well.

davefellows: Yup, fair point.

davefellows (approved):

davefellows (approved):

Porges (approved):

theunrepentantgeek (approved):
Really looking forward to introducing that enum. #approved

Issue #1257, opened 2020-08-21
Feature: support building an OpenShift operator bundle 


babbageclunk:
At the moment, the process for getting the operator running under OpenShift with CRD conversion is complicated and requires a lot of manual edits to files (see below). We should update samples and add extra scripts/Makefile targets to simplify building an operator bundle that can be submitted to the Operator Hub.

(from @madorn)

In project root:
1. `make manifests` - generates the raw crd manifest files
2. `operator-sdk generate kustomize manifests` - creates the required kustomize directory for OLM (config/manifests). You should remove - `../scorecard` from `config/manifests/kustomization.yaml`
3.  The samples directory needs a `kustomization.yaml` to reference the CRDs: (`cd config/samples && ls | sed 's/^/- /g' | (echo "resources:" && cat) > kustomization.yaml`)
4.  `kustomize` expects unique namespace/name pairs for each sample in the directory, so some of the objects in samples should look like this patch: https://github.com/operator-framework/operator-sdk/issues/3276#issuecomment-648218797 (we could update the actual upstream samples file to facilitate this….it doesn’t seem like a big deal but not sure if you all currently have any dependencies that would break as a result…
5. Generate the bundle: `kustomize build config/manifests | operator-sdk generate bundle --overwrite --version 0.37.0`
At this point, operator-sdk has now generated a `bundle.Dockerfile`, `bundle/metadata` and `/bundle/manifests`.
Metadata contains an `annotations.yaml` and `manifests` contains the `clusterServiceVersion.yaml` and all CRD’s.
The next step would typically be to run `docker build -f bundle.Dockerfile -t $BUNDLE_IMG` .
To facilitate the lack of CRD conversion support in OLM on OpenShift 4.5, we could do the following before building the bundle:
drop in this service.yaml into the bundle/manifests directory: azureoperator-webhook-service.yaml
```yaml
kind: Service
metadata:
  name: azureoperator-webhook-service
  annotations:
    service.alpha.openshift.io/serving-cert-secret-name: webhook-server-cert
spec:
  ports:
    - port: 443
      protocol: TCP
      targetPort: 9443
  selector:
    control-plane: controller-manager
  sessionAffinity: None
  type: ClusterIP
```
And then add the following annotation to the seven CRDs that have conversion logic:
```yaml
service.beta.openshift.io/inject-cabundle: "true"
```

Once CRD conversion support is released in OpenShift 4.7, we can eliminate the azureoperator-webhook-service.yaml and annotation steps.

PR #1285, opened 2020-10-02
Add ability to build openshift bundle for azure-service-operator

madorn:
Closes #[issue number]
https://github.com/Azure/azure-service-operator/issues/1257

Allows users to easily build their own customized [Operator Lifecycle Manager](https://github.com/operator-framework/operator-lifecycle-manager) Operator bundle for Azure Service Operator.

This is only required if one wants to build their own customized bundle.

Users can currently install Azure Service Operator without building their own custom bundle via [operatorhub.io](https://operatorhub.io/operator/azure-service-operator) or the [embedded OperatorHub](https://try.openshift.com).


config/samples/kustomization.yaml
```diff
@@ -0,0 +1,59 @@
+resources:
```
babbageclunk, line 1: Oh, I misunderstood your instructions about this file - I think it makes more sense to have it committed rather than generated each time.

config/manifests/bases/azure-service-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,394 @@
+apiVersion: operators.coreos.com/v1alpha1
```
babbageclunk, line 1: This a cunning way to have a default bundle with the questions answered and instructions/icons prepopulated. I'm not very experienced with kustomize so it didn't occur to me to use it to do this, thanks!

babbageclunk (approved):
This looks great, thanks @madorn! It's probably worth adding `bundle/` and `bundle.Dockerfile` to `.gitignore` as well.

babbageclunk:
Actually, I'll add the .gitignore lines in a follow-up PR.

PR #1286, opened 2020-10-02
adding config for stale bot

sakthi-vetrivel:
Creates a stale bot to mark issues as stale about 30 days and closes stale issues after 15 days.


.github/stale.yml
```diff
@@ -0,0 +1,59 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 30
```
theunrepentantgeek, line 4:
```suggestion
daysUntilStale: 60
```

.github/stale.yml
```diff
@@ -0,0 +1,59 @@
...
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 15
```
theunrepentantgeek, line 8: My experience (as a user, on other projects) is that it feels aggressive to have a bot closing issues; it sends me the strong message that any issues I'm experiencing do not matter.

```suggestion
daysUntilClose: false
```

theunrepentantgeek (changes requested):

.github/stale.yml
davefellows: I was thinking the same thing that we don't want to close issues automatically. I imagine there are going to be a few that will be longer term things that we'll want to keep on the list...

matthchr: +1 to the above. Bugs can take longer than 60 days to fix, as can new feature requests. It feels like a bot is the reaction to unmanageable load, which I don't think we currently have.

theunrepentantgeek (approved):
Thanks for making those changes. 😀

PR #1288, opened 2020-10-06
Improve managed user state management

matthchr:
A customer reported an issue where we failed to create a managed user but the error was very sparse. This should improve the error messages and also fix a few issues where we don't properly mark the resource as `FailedProvisoning`.

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/pkg/resourcemanager/azuresql/azuresqlmanageduser/azuresqlmanageduser_reconcile.go
+++ b/pkg/resourcemanager/azuresql/azuresqlmanageduser/azuresqlmanageduser_reconcile.go
@@ -43,15 +43,14 @@ func (s *AzureSqlManagedUserManager) Ensure(ctx context.Context, obj runtime.Obj
 	_, err = s.GetDB(ctx, instance.Spec.ResourceGroup, instance.Spec.Server, instance.Spec.DbName)
 	if err != nil {
 		instance.Status.Message = errhelp.StripErrorIDs(err)
-		instance.Status.Provisioning = false
 
-		requeuErrors := []string{
+		requeueErrors := []string{
 			errhelp.ResourceNotFound,
 			errhelp.ParentNotFoundErrorCode,
 			errhelp.ResourceGroupNotFoundErrorCode,
 		}
 		azerr := errhelp.NewAzureError(err)
-		if helpers.ContainsString(requeuErrors, azerr.Type) {
+		if helpers.ContainsString(requeueErrors, azerr.Type) {
 			return false, nil
 		}
 
@@ -61,7 +60,7 @@ func (s *AzureSqlManagedUserManager) Ensure(ctx context.Context, obj runtime.Obj
 			return false, nil
 		}
 
-		// if this is an unmarshall error - igmore and continue, otherwise report error and requeue
+		// if this is an unmarshall error - ignore and continue, otherwise report error and requeue
 		if !strings.Contains(errorString, "cannot unmarshal array into Go struct field serviceError2.details") {
 			return false, err
 		}
@@ -70,7 +69,6 @@ func (s *AzureSqlManagedUserManager) Ensure(ctx context.Context, obj runtime.Obj
 	db, err := s.ConnectToSqlDbAsCurrentUser(ctx, instance.Spec.Server, instance.Spec.DbName)
 	if err != nil {
 		instance.Status.Message = errhelp.StripErrorIDs(err)
-		instance.Status.Provisioning = false
 
 		// catch firewall issue - keep cycling until it clears up
 		if strings.Contains(err.Error(), "create a firewall rule for this IP address") {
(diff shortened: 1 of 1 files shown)


pkg/resourcemanager/azuresql/azuresqlmanageduser/azuresqlmanageduser_reconcile.go
```diff
@@ -43,15 +43,14 @@ func (s *AzureSqlManagedUserManager) Ensure(ctx context.Context, obj runtime.Obj
 	_, err = s.GetDB(ctx, instance.Spec.ResourceGroup, instance.Spec.Server, instance.Spec.DbName)
 	if err != nil {
 		instance.Status.Message = errhelp.StripErrorIDs(err)
-		instance.Status.Provisioning = false
 
-		requeuErrors := []string{
+		requeueErrors := []string{
```
theunrepentantgeek, line 47: 😁

pkg/resourcemanager/azuresql/azuresqlmanageduser/azuresqlmanageduser_reconcile.go
```diff
@@ -61,7 +60,7 @@ func (s *AzureSqlManagedUserManager) Ensure(ctx context.Context, obj runtime.Obj
 			return false, nil
 		}
 
-		// if this is an unmarshall error - igmore and continue, otherwise report error and requeue
+		// if this is an unmarshall error - ignore and continue, otherwise report error and requeue
 		if !strings.Contains(errorString, "cannot unmarshal array into Go struct field serviceError2.details") {
```
theunrepentantgeek, line 64: Is the sense of this condition correct? The comment and the **if** seem to be inverted wrt each other.

theunrepentantgeek (commented):
One comment that might reflect another bug, will approve once we've discussed that.

pkg/resourcemanager/azuresql/azuresqlmanageduser/azuresqlmanageduser_reconcile.go
babbageclunk: It would be clearer if the sense of the comment was flipped to matched the if, definitely. Or if you pulled `isUnmarshalError` into a variable it probably wouldn't need a comment.

pkg/resourcemanager/azuresql/azuresqlmanageduser/azuresqlmanageduser_reconcile.go
```diff
@@ -169,7 +170,6 @@ func (s *AzureSqlManagedUserManager) Delete(ctx context.Context, obj runtime.Obj
 	db, err := s.ConnectToSqlDbAsCurrentUser(ctx, instance.Spec.Server, instance.Spec.DbName)
 	if err != nil {
 		instance.Status.Message = errhelp.StripErrorIDs(err)
-		instance.Status.Provisioning = false
```
babbageclunk, line 172: Why does this one go away? Is it just because the instance is going away anyway? Or just because Provisioning would already be false by this point?

babbageclunk (approved):
Looks good to me, other than the confusing comment @theunrepentantgeek pointed out.

pkg/resourcemanager/azuresql/azuresqlmanageduser/azuresqlmanageduser_reconcile.go
matthchr: There are a few cases:
1. Provisioning was set to false already because the resource was successfully created already. In this case, since we're just deleting, it doesn't make sense to set provisioning back to true (and we don't).
2. Provisioning was true, which means the resource wasn't successfully created yet. In this case it might make sense to set it to false since we're about to delete it, but doing so seems somewhat pointless because we're about to delete the resource. Certainly in this case, (the error case), it doesn't make sense to set provisioning to false. It might might sense to do it as a default at the beginning of delete.

Based on my discussions with Erin, provisioning was originally intended to actually reflect _when we are doing a deployment operation to Azure_. I think longer term we should revisit this and possibly just remove the notion of informing the user of that aspect anyway. We should just tell them:
```
Reconciling
Steady
ReconcileFailed
```

I think this more clearly covers all of the cases and allows us to set `Reconciling` for cases like `Delete` where we really aren't `provisioning` but we are doing _something_.

pkg/resourcemanager/azuresql/azuresqlmanageduser/azuresqlmanageduser_reconcile.go
matthchr: This comment is actually across a number of different resources (code looks basically the same), and I think it's correct although possibly phrased weirdly?

It's saying: If it is an unmarshal error, continue with this reconcile, else return. That matches the code (which returns for all errors other than unmarshal).

pkg/resourcemanager/azuresql/azuresqlmanageduser/azuresqlmanageduser_reconcile.go
babbageclunk: Makes sense, thanks!

babbageclunk (approved):

PR #1296, opened 2020-10-19
KeyVault multiple policies ObjectID fix

Subreptivus:

The range variable is overwritten at each iteration, and &policy is the same. It's end up appending a newEntry containing the same ObjectID multiple times. Simple workaround is to use a local copy of that variable.


theunrepentantgeek (approved):
Good catch! Thanks for the PR.

babbageclunk (approved):
That annoying gotcha! Good spotting, thanks!

theunrepentantgeek:
Thanks @Subreptivus.

Issue #1291, opened 2020-10-07
Consider using a different UUID library


matthchr:
**Describe the current behavior**
We're using github.com/satori/go.uuid which hasnt been updated in a while.

**Describe the improvement**
Move to a different library. Overall usage is quite small.

matthchr:
There's an API compatible version that's been updated more recently: https://github.com/gofrs/uuid

Porges:
This is currently not possible as `azure-sdk-for-go` has [a direct dependency on the bad package](https://github.com/Azure/azure-sdk-for-go/issues/3158#issuecomment-717569512).

Porges:
Maybe I can use `go.mod` rewrites…

matthchr:
@porges did you do this?

Porges:
@matthchr waiting for https://github.com/Azure/autorest.go/pull/520 to land in autorest, then we can update our dependencies.

PR #1306, opened 2020-10-27
Remove use of satori uuid package

Porges:
For #1291. Had to fix this with a go.mod rewrite as there are dependencies like the Go Azure SDK that depend on the satori package directly.


go.mod
```diff
@@ -4,7 +4,6 @@ go 1.13
 
 require (
 	github.com/Azure/azure-sdk-for-go v44.0.0+incompatible
-	github.com/Azure/go-autorest v14.2.0+incompatible // indirect
```
Porges, line 7: these other removals happened due to `go mod tidy`

matthchr (approved):
LGTM provided tests pass

Porges:
@matthchr what's the test coverage like? are we safe?

matthchr:
@Porges the test coverage isn't as good as it could/should be - but it's what we have.

I see this package is used by us in azuresqlmanageduser and keyvault. We definitely have tests for KeyVault and from what I can see at least `TestKeyvaultControllerWithAccessPolicies` should cover a path where this package is being used.

Porges:
I might hold off on this for a bit while Azure SDK for Go updates their references. Then I shouldn’t need to do mod rewriting.

Porges:
Tracking this PR in autorest: https://github.com/Azure/autorest.go/pull/520

PR #1309, opened 2020-11-04
Don't use global credentials in low-level methods

babbageclunk:
This is a refactoring to pull the use of global credentials up to the level of the reconcile loop for resources to enable switching where we look for credentials in a future PR. This is part of implementing #1173 based on the work in #1206.

This PR doesn't change the behaviour of the operator - the controller loops will still use global credentials. Making this mechanical change first in a separate PR to simplify reviewing.

Tips for reviewers - the first three commits are the most interesting ones, the others are basically just mechanical. Although the resource groups one was slightly fiddlier than the others.

gif

### DIFF
--- a/main.go
+++ b/main.go
@@ -837,6 +837,7 @@ func main() {
 		Reconciler: &controllers.AsyncReconciler{
 			Client: mgr.GetClient(),
 			AzureClient: loadbalancer.NewAzureLoadBalancerClient(
+				config.GlobalCredentials(),
 				secretClient,
 				mgr.GetScheme(),
 			),
--- a/main.go
+++ b/main.go
@@ -158,6 +158,7 @@ func main() {
 
 	redisCacheFirewallRuleManager := rcfwr.NewAzureRedisCacheFirewallRuleManager()
 	appInsightsManager := resourcemanagerappinsights.NewManager(
+		config.GlobalCredentials(),
 		secretClient,
 		scheme,
 	)
@@ -870,7 +871,7 @@ func main() {
 	if err = (&controllers.AppInsightsApiKeyReconciler{
 		Reconciler: &controllers.AsyncReconciler{
 			Client:      mgr.GetClient(),
-			AzureClient: resourcemanagerappinsights.NewAPIKeyClient(secretClient, scheme),
+			AzureClient: resourcemanagerappinsights.NewAPIKeyClient(config.GlobalCredentials(), secretClient, scheme),
 			Telemetry: telemetry.InitializeTelemetryDefault(
 				"AppInsightsApiKey",
 				ctrl.Log.WithName("controllers").WithName("AppInsightsApiKey"),
--- a/main.go
+++ b/main.go
@@ -173,13 +173,15 @@ func main() {
 	}
 	eventhubClient := resourcemanagereventhub.NewEventhubClient(secretClient, scheme)
 	sqlServerManager := resourcemanagersqlserver.NewAzureSqlServerManager(
+		config.GlobalCredentials(),
 		secretClient,
 		scheme,
 	)
-	sqlDBManager := resourcemanagersqldb.NewAzureSqlDbManager()
-	sqlFirewallRuleManager := resourcemanagersqlfirewallrule.NewAzureSqlFirewallRuleManager()
-	sqlVNetRuleManager := resourcemanagersqlvnetrule.NewAzureSqlVNetRuleManager()
+	sqlDBManager := resourcemanagersqldb.NewAzureSqlDbManager(config.GlobalCredentials())
+	sqlFirewallRuleManager := resourcemanagersqlfirewallrule.NewAzureSqlFirewallRuleManager(config.GlobalCredentials())
+	sqlVNetRuleManager := resourcemanagersqlvnetrule.NewAzureSqlVNetRuleManager(config.GlobalCredentials())
 	sqlFailoverGroupManager := resourcemanagersqlfailovergroup.NewAzureSqlFailoverGroupManager(
+		config.GlobalCredentials(),
 		secretClient,
 		scheme,
 	)
@@ -191,14 +193,16 @@ func main() {
 		scheme,
 	)
 	sqlUserManager := resourcemanagersqluser.NewAzureSqlUserManager(
+		config.GlobalCredentials(),
 		secretClient,
 		scheme,
 	)
 	sqlManagedUserManager := resourcemanagersqlmanageduser.NewAzureSqlManagedUserManager(
+		config.GlobalCredentials(),
 		secretClient,
 		scheme,
 	)
-	sqlActionManager := resourcemanagersqlaction.NewAzureSqlActionManager(secretClient, scheme)
+	sqlActionManager := resourcemanagersqlaction.NewAzureSqlActionManager(config.GlobalCredentials(), secretClient, scheme)
 
 	err = (&controllers.StorageAccountReconciler{
 		Reconciler: &controllers.AsyncReconciler{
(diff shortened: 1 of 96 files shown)


pkg/resourcemanager/config/credentials.go
```diff
@@ -0,0 +1,61 @@
...
+	OperatorKeyvault() string
+}
+
+type credentials struct {
+	clientID       string
+	clientSecret   string
+	tenantID       string
+	subscriptionID string
+
+	// TODO: not sure whether these are part of the credentials or
```
matthchr, line 24: IMO they both are.

pkg/resourcemanager/config/credentials.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// SubscriptionID is a target subscription for Azure resources.
+func (c credentials) SubscriptionID() string {
+	return c.subscriptionID
+}
+
+// UseMI() specifies if managed service identity auth should be used. Used for
+// aad-pod-identity
+func (c credentials) UseMI() bool {
```
matthchr, line 54: Minor: Worth expanding to `UseManagedIdentity`? Even though I know what we're talking about here I find `MI` a somewhat obtuse abbreviation.

pkg/resourcemanager/iam/authorizers.go
```diff
@@ -230,7 +230,7 @@ func getAuthorizerForResource(resource string) (autorest.Authorizer, error) {
 		a = autorest.NewBearerAuthorizer(token)
 
 	case OAuthGrantTypeDeviceFlow:
```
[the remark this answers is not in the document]
matthchr: Nothing you need to do about that in this PR though

pkg/resourcemanager/iam/authorizers.go
```diff
@@ -260,13 +260,14 @@ func GetMSITokenForResource(resource string) (*adal.ServicePrincipalToken, error
 }
 
 // GetResourceManagementTokenHybrid retrieves auth token for hybrid environment
-func GetResourceManagementTokenHybrid(activeDirectoryEndpoint, tokenAudience string) (adal.OAuthTokenProvider, error) {
+// TODO(creds-refactor): this seems to be unused, remove it.
```
matthchr, line 263: Planning to do this in a follow-up PR, or was this intended to be part of this PR?

pkg/resourcemanager/keyvaults/keyvault.go
```diff
@@ -49,9 +51,12 @@ func getVaultsClient() (keyvault.VaultsClient, error) {
 	return vaultsClient, nil
 }
 
-func getObjectID(ctx context.Context, tenantID string, clientID string) (*string, error) {
+// TODO(creds-refactor): can we just use the tenant and client id from
+// the creds here? Depends whether the client code uses them to
+// override the creds.
+func getObjectID(ctx context.Context, creds config.Credentials, tenantID string, clientID string) (*string, error) {
```
matthchr, line 57: This function is used in 2 places. One of those places is `CreateVaultWithAccessPolicies` which as far as I can tell is only used in tests... may be worth moving to tests directly if that's really a test helper method not used at all in production code.

The other location is `ParseAccessPolicy` which seems to use a clientId/tenantId other than the one in the credentials (supplied by the user?)

pkg/resourcemanager/keyvaults/keyvault.go
```diff
@@ -114,7 +119,7 @@ func ParseNetworkPolicy(ruleSet *v1alpha1.NetworkRuleSet) keyvault.NetworkRuleSe
 }
 
 // ParseAccessPolicy - helper function to parse access policies from Kubernetes spec
-func ParseAccessPolicy(policy *v1alpha1.AccessPolicyEntry, ctx context.Context) (keyvault.AccessPolicyEntry, error) {
+func ParseAccessPolicy(policy *v1alpha1.AccessPolicyEntry, ctx context.Context, creds config.Credentials) (keyvault.AccessPolicyEntry, error) {
```
matthchr, line 122: minor: consider making `ctx` the first param?

pkg/resourcemanager/keyvaults/keyvault_manager.go
```diff
@@ -13,8 +13,6 @@ import (
 	"github.com/Azure/azure-service-operator/pkg/resourcemanager"
 )
 
-var AzureKeyVaultManager KeyVaultManager = &azureKeyVaultManager{}
```
matthchr, line 16: Why remove this?

matthchr: Instead just do `var _ KeyVaultManager = &azureKeyVaultManager{}` seems reasonable?

pkg/resourcemanager/storages/storageaccount/storageaccount.go
```diff
@@ -80,9 +81,9 @@ func ParseNetworkPolicy(ruleSet *v1alpha1.StorageNetworkRuleSet) storage.Network
 	}
 }
 
-func getStoragesClient() (storage.AccountsClient, error) {
-	storagesClient := storage.NewAccountsClientWithBaseURI(config.BaseURI(), config.SubscriptionID())
-	a, err := iam.GetResourceManagementAuthorizer()
+func getStoragesClient(creds config.Credentials) (storage.AccountsClient, error) {
```
matthchr, line 84: minor: `getStorageClient`?

matthchr (dismissed):
That was quite a PR to review!

pkg/resourcemanager/iam/authorizers.go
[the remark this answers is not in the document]
babbageclunk: Yeah, it does seem like it could only possibly be useful when debugging, and even then it's dubious. I'll add a comment that this should be removed.

main.go
```diff
@@ -128,77 +129,87 @@ func main() {
...
 		setupLog.Info("Instantiating secrets client for keyvault " + keyvaultName)
-		secretClient = keyvaultSecrets.New(keyvaultName)
+		secretClient = keyvaultSecrets.New(keyvaultName, config.GlobalCredentials())
 	}
 
-	apimManager := resourceapimanagement.NewManager()
-	apimServiceManager := apimservice.NewAzureAPIMgmtServiceManager()
-	vnetManager := vnet.NewAzureVNetManager()
-	resourceGroupManager := resourcemanagerresourcegroup.NewAzureResourceGroupManager()
+	// TODO(creds-refactor): construction of these managers will need
```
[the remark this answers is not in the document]
babbageclunk: Yeah, I think this is a good change we should make - the small type-specific `*Reconciler` types should go away. This comment wasn't very clear because when I added it I didn't fully know where the managers were going to move to, but actually the construction will happen inside the `AsyncReconciler` rather than in the type-specific ones (that's the follow-up change I'm working on at the moment). I'll update it in the meantime.

pkg/resourcemanager/config/credentials.go
babbageclunk: Thanks, I've removed the comment.

pkg/resourcemanager/iam/authorizers.go
babbageclunk: Gah, I meant to do a pass across the PR to finalise these TODOs, sorry! May as well remove it here.

pkg/resourcemanager/keyvaults/keyvault.go
babbageclunk: I've removed the comment - I've left the method for now even though it's used from tests. If it was only used in the tests for keyvault I'd move it, but it's kind of tricky to share testing code between two packages.

pkg/resourcemanager/keyvaults/keyvault_manager.go
babbageclunk: This was being used by code in other packages - I needed to remove the global variable because it needs to have creds when it's constructed. I'll reinstate the interface assertion part though!

matthchr (approved):

PR #1312, opened 2020-11-16
[WIP] Construct clients/managers inside AsyncReconciler.Reconcile

babbageclunk:
This defers constructing the ARM client instances until we get a request, so that in future the reconciler can get credentials from a location specific to the resource being reconciled.

Add resourcemanager.ClientFactory type, and change AsyncReconciler to accept a ClientFactory so that it can construct the client on each request.

Add a NewARMClient function conforming to ClientFactory every resource package, and pass those in to the reconcilers.

This is part of implementing #1173 based on the work in #1206.

gif

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/controllers/async_controller.go
+++ b/controllers/async_controller.go
@@ -60,10 +60,10 @@ func (r *AsyncReconciler) Reconcile(req ctrl.Request, obj runtime.Object) (resul
 	var secretClient secrets.SecretClient
 	if keyvaultName == "" {
 		r.Telemetry.LogInfoByInstance("status", "keyvault name is empty", req.String())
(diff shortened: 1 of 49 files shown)


matthchr:
@babbageclunk what's the status of this PR?

babbageclunk:
I've been keeping it up to date but it's on-hold while we work out what our multitenancy approach will be - the operator-per-namespace approach doesn't really require this change from my perspective. Could close it for now and resurrect if we decide to go down this route?

babbageclunk:
Closing this for now - we're working on operator-per-namespace first, then will come back to this after that.

PR #1314, opened 2020-11-18
Add support for MySQL Administrators

matthchr:
**What this PR does / why we need it**: Adds support for creating MySQL admins

gif

**If applicable**:
- [x] this PR contains tests


controllers/mysql_combined_test.go
```diff
@@ -53,6 +53,27 @@ func TestMySQLHappyPath(t *testing.T) {
...
+	mySQLAdmin := &azurev1alpha1.MySQLServerAdministrator{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "admin",
+			Namespace: "default",
+		},
+		Spec: azurev1alpha1.MySQLServerAdministratorSpec{
+			ResourceGroup:     rgName,
+			Server:            mySQLServerName,
+			AdministratorType: azurev1alpha1.MySQLServerAdministratorTypeActiveDirectory,
+			// The below fields are for a user managed identity which exists in the ASO-CI subscription
```
matthchr, line 65: Possibly we need to support creating managed identities via the operator too in which case we could do an e2e test that did that first and then used it here

matthchr: I've asked Akin about this

pkg/resourcemanager/mysql/aadadmin/reconcile.go
```diff
@@ -0,0 +1,218 @@
...
+		instance.Status.SetProvisioned(resourcemanager.SuccessMsg)
+		instance.Status.State = admin.Status
+		return true, nil
+	}
+
+	tenantUUID, err := uuid.FromString(instance.Spec.TenantId)
+	if err != nil {
+		instance.Status.SetFailedProvisioning(fmt.Sprintf("failed to parse tenant UUID %s", instance.Spec.TenantId))
+		// TODO: Do we have to manage state here?
+		// instance.Status.State = admin.Status
```
babbageclunk, line 65: It seems like we'd always want the status to be the most up-to-date view we have, right? It's not bad the way it is, but if there were more early returns and you wanted to avoid the repetition you could use a defer.

babbageclunk (dismissed):
Looks great!

controllers/mysqlserveradministrator_controller.go
```diff
@@ -0,0 +1,28 @@
...
+// MySQLServerAdministratorReconciler reconciles a MySQLServerAdministrator object
+type MySQLServerAdministratorReconciler struct {
+	Reconciler *AsyncReconciler
+}
+
+// +kubebuilder:rbac:groups=azure.microsoft.com,resources=mysqlserveradministrators,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=azure.microsoft.com,resources={mysqlserveradministrators/status,mysqlserveradministrators/finalizers},verbs=get;update;patch
+
+func (r *MySQLServerAdministratorReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	return r.Reconciler.Reconcile(req, &azurev1alpha1.MySQLServerAdministrator{})
```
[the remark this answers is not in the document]
matthchr: I wonder if this file is generated by kubebuilder. It might be which might be why we're following the pattern? Unsure.

pkg/resourcemanager/mysql/aadadmin/reconcile.go
```diff
@@ -0,0 +1,218 @@
...
+
+func NewMySQLServerAdministratorManager(creds config.Credentials) *MySQLServerAdministratorManager {
+	return &MySQLServerAdministratorManager{
+		Creds: creds,
+	}
+}
+
+func newMySQLServerAdministratorsClient(creds config.Credentials) mysql.ServerAdministratorsClient {
+	adminClient := mysql.NewServerAdministratorsClientWithBaseURI(config.BaseURI(), creds.SubscriptionID())
+	a, _ := iam.GetResourceManagementAuthorizer(creds)
```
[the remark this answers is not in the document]
matthchr: I'll fix it here just so we're not propagating a bad pattern.

We should probably do a sweep to fix all the others too.

pkg/resourcemanager/mysql/aadadmin/reconcile.go
matthchr: The problem I'm having is it's actually unclear to me what the `State` field even is... and I see most resources don't consistently use it.

controllers/mysql_combined_test.go
matthchr: Sounds like we don't need to support managed identity creating at least for now, so this is probably the best path forward at least temporarily.

pkg/resourcemanager/mysql/aadadmin/reconcile.go
matthchr: I've fixed this to manually setting Succeeded/Failed which is what a bunch of resources seem to do...

controllers/mysqlserveradministrator_controller.go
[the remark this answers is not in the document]
babbageclunk: Hmm yeah - not sure where we would put the rbac annotations. Maybe we'd just have them in yaml instead? How do we do that for the code-gen resources?

babbageclunk (approved):

controllers/mysqlserveradministrator_controller.go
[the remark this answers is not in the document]
matthchr: I think the actual annotations can be anywhere - they aren't even affiliated with a type (although it's convenient to put them right above one)

PR #1317, opened 2020-11-24
Update MySQL user roles when users are modified

### LINKED ISSUE #1311: Bug: SQL Users whose spec is updated to change assigned roles do not update roles in the backing DB
**Describe the bug**
SQL Users whose spec is updated to change assigned roles do not update roles in the backing DB. I believe that this applies to all SQL users we currently support, but need to reproduce to be sure.

**To Reproduce**
Create user, specify some roles, remove 1 and update user, check DB to see if that role was removed for that user.

**Expected behavior**
The user is no longer a member of that role (as if it has been removed via `EXEC sp_droprolemember 'Role', 'userName';`)

matthchr:
This is related to but does not resolve #1311

We need to decide if we want to support this functionality at all. It's "easiest" for MySQL and harder for Azure SQL (I haven't even looked into PostgreSQL).

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/controllers/mysql_combined_test.go
+++ b/controllers/mysql_combined_test.go
@@ -9,10 +9,14 @@ import (
 	"context"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 
 	azurev1alpha1 "github.com/Azure/azure-service-operator/api/v1alpha1"
 	"github.com/Azure/azure-service-operator/api/v1alpha2"
+	"github.com/Azure/azure-service-operator/pkg/resourcemanager/mysql"
+	"github.com/Azure/azure-service-operator/pkg/resourcemanager/mysql/mysqluser"
 )
 
 func TestMySQLHappyPath(t *testing.T) {
@@ -32,6 +36,7 @@ func TestMySQLHappyPath(t *testing.T) {
 	RequireInstance(ctx, t, tc, mySQLServerInstance)
 
 	// Create a mySQL replica
+
 	mySQLReplicaInstance := v1alpha2.NewReplicaMySQLServer(mySQLReplicaName, rgName, rgLocation, mySQLServerInstance.Status.ResourceId)
 	mySQLReplicaInstance.Spec.StorageProfile = nil
 
@@ -76,6 +81,8 @@ func TestMySQLHappyPath(t *testing.T) {
 
 	ruleName := GenerateTestResourceNameWithRandom("mysql-fw", 10)
 
+	// This rule opens access to the public internet, but in this case
+	// there's literally no data in the database
 	ruleInstance := &azurev1alpha1.MySQLFirewallRule{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      ruleName,
@@ -85,11 +92,14 @@ func TestMySQLHappyPath(t *testing.T) {
 			Server:         mySQLServerName,
 			ResourceGroup:  rgName,
 			StartIPAddress: "0.0.0.0",
-			EndIPAddress:   "0.0.0.0",
+			EndIPAddress:   "255.255.255.255",
 		},
 	}
 
 	EnsureInstance(ctx, t, tc, ruleInstance)
+
+	// Create user and ensure it can be updated
+	RunMySQLUserHappyPath(ctx, t, mySQLServerName,mySQLDBName, rgName)
 
 	// Create VNet and VNetRules -----
 	RunMySqlVNetRuleHappyPath(t, mySQLServerName, rgLocation)
@@ -98,4 +108,70 @@ func TestMySQLHappyPath(t *testing.T) {
 	EnsureDelete(ctx, t, tc, mySQLDBInstance)
 	EnsureDelete(ctx, t, tc, mySQLServerInstance)
 	EnsureDelete(ctx, t, tc, mySQLReplicaInstance)
+}
+
+func RunMySQLUserHappyPath(ctx context.Context, t *testing.T, mySQLServerName string, mySQLDBName string, rgName string) {
+	assert := assert.New(t)
+
+	// Create a user in the DB
+	username := GenerateTestResourceNameWithRandom("user", 10)
+	user := &azurev1alpha1.MySQLUser{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      username,
+			Namespace: "default",
+		},
+		Spec: azurev1alpha1.MySQLUserSpec{
+			ResourceGroup: rgName,
+			Server:        mySQLServerName,
+			DbName:        mySQLDBName,
+			Username:      username,
+			Roles:         []string{"SELECT"},
+		},
+	}
+	EnsureInstance(ctx, t, tc, user)
+
+	// Update user
+	namespacedName := types.NamespacedName{Name: username, Namespace: "default"}
+	err := tc.k8sClient.Get(ctx, namespacedName, user)
+	assert.NoError(err)
+
+	updatedRoles := []string{"UPDATE", "DELETE", "CREATE", "DROP"}
+	user.Spec.Roles = updatedRoles
+	err = tc.k8sClient.Update(ctx, user)
+	assert.NoError(err)
+
+	// TODO: Ugh this is fragile, the path to the secret should probably be set on the status?
+	// See issue here: https://github.com/Azure/azure-service-operator/issues/1318
+	secretNamespacedName := types.NamespacedName{Name: mySQLServerName, Namespace: "default"}
+	adminSecret, err := tc.secretClient.Get(ctx, secretNamespacedName)
+	adminUser := string(adminSecret["fullyQualifiedUsername"])
+	adminPassword := string(adminSecret[mysqluser.MSecretPasswordKey])
+	fullServerName := string(adminSecret["fullyQualifiedServerName"])
+
+	db, err := mysql.ConnectToSqlDB(
+		ctx,
+		mysql.MySQLDriverName,
+		fullServerName,
+		mySQLDBName,
+		mysql.MySQLServerPort,
+		adminUser,
+		adminPassword)
+	assert.NoError(err)
+
+	assert.Eventually(func() bool {
+		roles, err := mysql.ExtractUserRoles(ctx, db, username, mySQLDBName)
+		assert.NoError(err)
+
+		if len(roles) != len(updatedRoles) {
+			return false
+		}
+
+		for _, role := range updatedRoles {
+			if _, ok := roles[role]; !ok {
+				return false
+			}
+		}
+
+		return true
+	}, tc.timeout, tc.retry, "waiting for DB user to be updated")
 }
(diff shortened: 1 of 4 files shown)


controllers/mysql_combined_test.go
```diff
@@ -99,3 +109,70 @@ func TestMySQLHappyPath(t *testing.T) {
...
+	namespacedName := types.NamespacedName{Name: username, Namespace: "default"}
+	err := tc.k8sClient.Get(ctx, namespacedName, user)
+	assert.NoError(err)
+
+	updatedRoles := []string{"UPDATE", "DELETE", "CREATE", "DROP"}
+	user.Spec.Roles = updatedRoles
+	err = tc.k8sClient.Update(ctx, user)
+	assert.NoError(err)
+
+	// TODO: Ugh this is fragile, the path to the secret should probably be set on the status?
```
babbageclunk, line 143: Yeah, really good call.

controllers/mysql_combined_test.go
```diff
@@ -99,3 +109,70 @@ func TestMySQLHappyPath(t *testing.T) {
...
+		}
+
+		for _, role := range updatedRoles {
+			if _, ok := rolesForThisDB[role]; !ok {
+				return false
+			}
+		}
+
+		return true
+	}, tc.timeout, tc.retry, "waiting for DB user to be updated")
```
babbageclunk, line 177: Test for removing a user's role as well?

pkg/helpers/sqlrole.go
```diff
@@ -0,0 +1,33 @@
+package helpers
+
+type SQLRoleAction string
+
+const (
+	RoleActionDelete = "delete"
+	RoleActionCreate = "create"
+)
+
+type SQLRoleAndAction struct {
```
babbageclunk, line 10:
```suggestion
type SQLRoleAction struct {
```

pkg/helpers/sqlrole.go
```diff
@@ -0,0 +1,33 @@
...
+	Role   string
+	Action SQLRoleAction
+}
+
+func DiffCurrentAndExpectedSQLRoles(currentRoles map[string]struct{}, expectedRoles map[string]struct{}) []SQLRoleAndAction {
```
babbageclunk, line 15: I think this would be cleaner returning adds and deletes as two slices - all the callers need to split them out them anyway. If you're worried about getting them muddled up you could make them a struct with Added and Deleted fields.

pkg/helpers/sqlrole_test.go
```diff
@@ -0,0 +1,84 @@
...
+				switch roleAndAction.Action {
+				case RoleActionCreate:
+					adds[roleAndAction.Role] = struct{}{}
+				case RoleActionDelete:
+					removes[roleAndAction.Role] = struct{}{}
+				}
+
+			}
+
+			assert.Equal(c.expectedRoleCreates, adds)
```
babbageclunk, line 79: If you made `DiffCurrentAndExpectedSQLRoles` return two slices you could use assert.ElementsMatch to compare them independent of order here.

pkg/resourcemanager/mysql/mysqlhelper.go
```diff
@@ -88,30 +93,112 @@ func ConnectToSQLDBAsCurrentUser(
...
+// ExtractUserRoles extracts the roles the user has. The result is a map of database to roles
+func ExtractUserRoles(ctx context.Context, user string, db *sql.DB) (map[string]map[string]struct{}, error) {
+	if err := helpers.FindBadChars(user); err != nil {
+		return nil, errors.Wrapf(err, "problem found with username")
+	}
+
+	// TODO: Possibly need to fully qualify this with database name?
+	rows, err := db.QueryContext(ctx, "SHOW GRANTS FOR ?", user)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to list grants for user %s", user)
```
babbageclunk, line 127: Failed is redundant here - just `listing grants for user blah` would be better.

pkg/resourcemanager/mysql/mysqlhelper.go
```diff
@@ -88,30 +93,112 @@ func ConnectToSQLDBAsCurrentUser(
...
+	}
+	defer rows.Close()
+
+	result := make(map[string]map[string]struct{})
+	for rows.Next() {
+		var row string
+		err := rows.Scan(&row)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to iterate returned rows")
```
babbageclunk, line 136: Same here.

pkg/resourcemanager/mysql/mysqlhelper.go
```diff
@@ -88,30 +93,112 @@ func ConnectToSQLDBAsCurrentUser(
...
 		// an administrator user and then doing whatever they want without SQL injection.
 		// See https://github.com/go-sql-driver/mysql/blob/3b935426341bc5d229eafd936e4f4240da027ccd/connection.go#L198
 		// for specifics of what go-mysql-driver supports.
-		tsql := fmt.Sprintf("GRANT %s ON `%s`.* TO ?", role, database)
-		_, err := db.ExecContext(ctx, tsql, user)
-		if err != nil {
-			errorStrings = append(errorStrings, err.Error())
+		switch roleAndAction.Action {
+		case helpers.RoleActionCreate:
+			tsql := fmt.Sprintf("GRANT %s ON `%s`.* TO ?", roleAndAction.Role, database)
```
babbageclunk, line 189: The GRANT and REVOKE commands can accept multiple roles, so you could just issue two statements (or one if one of the slices was empty) rather than one per role changed.

pkg/resourcemanager/mysql/mysqlhelper.go
```diff
@@ -88,30 +93,112 @@ func ConnectToSQLDBAsCurrentUser(
...
+}
+
+// ExtractUserRoles extracts the roles the user has. The result is a map of database to roles
+func ExtractUserRoles(ctx context.Context, user string, db *sql.DB) (map[string]map[string]struct{}, error) {
+	if err := helpers.FindBadChars(user); err != nil {
+		return nil, errors.Wrapf(err, "problem found with username")
+	}
+
+	// TODO: Possibly need to fully qualify this with database name?
+	rows, err := db.QueryContext(ctx, "SHOW GRANTS FOR ?", user)
```
babbageclunk, line 125: The output of SHOW GRANTS is really intended for humans, can we get the information we need by querying the information schema? That would hopefully give us individual rows for each role so you wouldn't need the nasty parsing. (Trying to find a clear reference for doing this - I'm more familiar with the postgres ones for this kind of info.)

babbageclunk (dismissed):
This looks good, but possibly there's a way to get the role information in a more machine-readable format.

pkg/resourcemanager/mysql/mysqlhelper.go
babbageclunk: I just tried it out, I think selecting from `information_schema.user_privileges` would get what you want, since we're granting on everything?

controllers/mysql_combined_test.go
```diff
@@ -76,6 +81,8 @@ func TestMySQLHappyPath(t *testing.T) {
 
 	ruleName := GenerateTestResourceNameWithRandom("mysql-fw", 10)
 
+	// This rule opens access to the public internet, but in this case
+	// there's literally no data in the database
```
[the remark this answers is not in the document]
matthchr: Yeah, I originally tried to get the IP and then realized it wasn't going to work because we dont know what routing looks like outside of the docker container/VM.

controllers/mysql_combined_test.go
matthchr: Filed an issue for this here: https://github.com/Azure/azure-service-operator/issues/1318

controllers/mysql_combined_test.go
matthchr: This test actually tests both at once. I set the desired roles from `SELECT` to `UPDATE, DELETE, CREATE, DROP`

matthchr: So for the test to pass `SELECT` had to have been removed.

pkg/helpers/sqlrole.go
matthchr: The enum was already called that

pkg/helpers/sqlrole.go
matthchr: Will do the struct instead.

pkg/helpers/sqlrole.go
matthchr: This goes away with the change below though, so resolving this.

pkg/helpers/sqlrole_test.go
matthchr: Had them return two maps (since order doesn't matter and duplicates don't make sense), so kept using `Equal`.

pkg/resourcemanager/mysql/mysqlhelper.go
matthchr: Nice

pkg/resourcemanager/mysql/mysqlhelper.go
matthchr: Trying this out

matthchr: This works. I've changed to this which is quite a lot cleaner and makes me feel significantly better about the chances this is going to work in practice with real users and permissions. I left a note stating that this works only because we're limited in what permissions we manage (only per-DB, rather than per-table or per-column).

controllers/mysql_combined_test.go
babbageclunk: Oh sorry, missed that!

pkg/resourcemanager/mysql/mysqlhelper.go
```diff
@@ -88,31 +90,72 @@ func ConnectToSQLDBAsCurrentUser(
...
+func ExtractUserRoles(ctx context.Context, db *sql.DB, user string, database string) (map[string]struct{}, error) {
+	if err := helpers.FindBadChars(user); err != nil {
+		return nil, errors.Wrapf(err, "problem found with username")
+	}
+
+	// Note: This works because we only assign permissions at the DB level, not at the table, column, etc levels -- if we assigned
+	// permissions at more levels we would need to do something else here such as join multiple tables or
+	// parse SHOW GRANTS with a regex.
+	tsql := fmt.Sprintf("SELECT PRIVILEGE_TYPE from information_schema.SCHEMA_PRIVILEGES WHERE GRANTEE = '''%s''@''%%''' and TABLE_SCHEMA = ?", user)
+	rows, err := db.QueryContext(ctx, tsql, database)
```
babbageclunk, line 103: Grantee can be a parameter here, can't it? You'd still have to build the weird formatting with the quotes and @ - that's a bit easier if you don't need to sql-quote it.
```suggestion
	tsql := "select privilege_type from information_schema.schema_privileges where grantee = ? and table_schema= ?"
	formattedUser := fmt.Sprintf("'%s'@'%'", user)
	rows, err := db.QueryContext(ctx, tsql, formattedUser, database)
```

I tend to make queries all lowercase, not sure what the standard is - a lot of people prefer SQL keywords in uppercase.

matthchr: Yeah all of the existing queries we have are all uppercase for the keywords at least. I'll fix this

babbageclunk (approved):

PR #1332, opened 2020-12-10
Update Helm chart to latest released operator version

babbageclunk:
- Use more best practices for Helm charts as defined here: https://helm.sh/docs/chart_best_practices/conventions/
  - Update chart to contain recently added resources.
  - Update chart version to 1.0.0, so that we can begin following semver more closely.

This is based on #1316 but rolls back the controller-tools dependency to 0.2.5.

As noted in a [comment on the previous PR](https://github.com/Azure/azure-service-operator/pull/1316#issuecomment-736139050), we haven't updated the aad-pod-identity dependency yet because they have a list parameter that doesn't merge nicely with user-provided config. When the fix for [this bug](https://github.com/Azure/aad-pod-identity/issues/885) is released we'll update this dependency.


theunrepentantgeek (approved):
:shipit:

PR #1336, opened 2020-12-20
Continue reconciling Azure SQL managed users for unknown errors

babbageclunk:
These could be setup issues like the AAD admin not being set on the server - giving up on the first try means the only way to get the user created successfully is to edit the resource to remove the finalizer, delete it, then re-add. The Ensure for non AAD users doesn't use this pattern of returning true, nil when it can't reach the server.

**If applicable**:
- [ ] this PR contains tests


pkg/resourcemanager/azuresql/azuresqlmanageduser/azuresqlmanageduser_reconcile.go
```diff
@@ -81,9 +81,8 @@ func (s *AzureSqlManagedUserManager) Ensure(ctx context.Context, obj runtime.Obj
 			return false, nil
 		}
 
-		// Other failures are terminal
 		instance.Status.SetFailedProvisioning(instance.Status.Message)
-		return true, nil
+		return false, nil
```
theunrepentantgeek, line 85: Ergh. I dislike returning boolean values unless they have really simple meanings - better to define an enumeration and be declarative. Something for later, perhaps.

theunrepentantgeek (approved):

pkg/resourcemanager/azuresql/azuresqlmanageduser/azuresqlmanageduser_reconcile.go
babbageclunk: Agreed! Especially here for Ensure and for Delete - the meaning of these booleans are pretty unclear, you basically need to read the AsyncReconciler code to see how they're used. I'm going to make an issue to clean them up - but obviously it's spread across all of the manager types unfortunately.

matthchr (approved):

PR #1339, opened 2020-12-22
Allow Mysql AAD User DB Name to be empty.

cdmurph32:
Leaving the DBName field empty will cause the roles to be granted on the server level, rather than just for a database.

This allows user privileges to be granted at the server level for all databases. This also allows for the following privileges to be used: `RELOAD, PROCESS, SHOW DATABASES, REPLICATION SLAVE, REPLICATION CLIENT, CREATE USER`

PostgreSQL probably needs similar work. MySQLUser could easily take advantage since it calls the same helper functions. That said, this is all that is required for the Adobe work item.

gif

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/api/v1alpha1/mysqlaaduser_types.go
+++ b/api/v1alpha1/mysqlaaduser_types.go
@@ -15,9 +15,8 @@ type MySQLAADUserSpec struct {
 	// +kubebuilder:validation:Required
 	Server string `json:"server"`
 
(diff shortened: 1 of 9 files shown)


matthchr:
@babbageclunk I think your #1357 supersedes this, can you confirm and if so we should probably close this?

babbageclunk:
We ended up taking a different approach to solve this problem, as seen in #1357.

PR #1342, opened 2020-12-27
feat: added in azure container registry resource

chrismellard:
New Azure Container Registry operator.

First contribution here and had some dramas getting the Makefile to function well on a Mac. These were

- `sed -i'' -e` causes a backup with an extension of `-e` for me. To resolve for me on a Mac required a space like `sed -i '' -e`
- Redundant `sed` command in `make update` which calls `make docker-image` that already uses `sed` to replace the kustomize patch image version of the controller
- `kind>=0.7` necessitated `--loglevel=trace` => `-v=4` or pick another verbosity level beyond 4!
- `kind>=0.7` necessitated specifying `--name <kind-cluster-name<>` on any `kind` commands
- `kind>=0.7` has deprecated `kind get kubeconfig-path` so replaced this with `kubectl config current-context` to check we're pointing to kind cluster

Other than that the rest of the changes are all ACR operator specific.


matthchr:
Closing this in favor of the ASO v2 PR that was merged here: #2074

Issue #1327, opened 2020-12-08
Bug: Helm install missing POD_NAMESPACE on deployment, CRDs don't work


kyschouv:
**Describe the bug**
We're installing using the Helm chart, and there are a couple of issues. First of all, pods for the `azureoperator-controller-manager` deployments log an error from the `manager` container complaining that the `POD_NAMESPACE` environment variable isn't set. If the deployment is manually patched, then the pod eventually outputs this error:

```
ERROR   setup   problem running manager {"error": "no matches for kind \"MySQLAADUser\" in version \"azure.microsoft.com/v1alpha1\""}

github.com/go-logr/zapr.(*zapLogger).Error
        /go/pkg/mod/github.com/go-logr/zapr@v0.1.0/zapr.go:128
main.main
        /workspace/main.go:980
runtime.main
        /usr/local/go/src/runtime/proc.go:203
```

In both cases, ASO is unusable.

**To Reproduce**

Install chart version 0.1.0 using the following values file (we're using Kubernetes version 1.18.10):

Values file:
```
azureSubscriptionID: {subscription}
azureTenantID: {tenant}
azureUseMI: true
```

**Additional context**
We're using Terraform to manage this deployment, but that shouldn't impact anything, as it's still just using Helm under the hood.

Is the Helm chart still the recommended install method? Should we be using Operator Lifecycle Manager or something to manage the install? Will that work with a managed identity? (we're trying to use managed identities wherever possible).

babbageclunk:
Thanks @kyschouv - I should have released an update for the helm chart when I did the last release, it looks like that will be the cause of both the POD_NAMESPACE and MySQLAADUser errors. Sorry about that - I'll update it today!

Deploying with Helm is still the recommended install method, although there are some issues with CRD updates for Helm that are still open - discussion here: https://github.com/helm/helm/issues/5871
We're not sure how best to handle that problem.

Installing with OLM would also work, but the latest version isn't on operatorhub yet either - I'll start that process too. (It would preventthe issue of the chart and image being out of sync.) OLM can support managed identity.

kyschouv:
Do you have an indication of whether Helm or OLM will be a more stable method of installation for the time being (i.e. next 3-6 months)? It's feeling like OLM is probably the way to go, especially as CRDs are likely to be added and potentially changed, and that Helm issue has been open for nearly a year and a half now without resolution. However, if it's not being updated as frequently, that might point at Helm being the best choice? And just dealing with CRD challenges?

[...]

PR #1345, opened 2021-01-05
Refer to a specific ASO image tag in the helm chart

babbageclunk:
Extend the helm-chart-manifests target to get the latest release from the github  API and include it in the chart values.yaml so that deploying the chart will get the matching image from the registry.

This will prevent errors caused by the helm chart referring to :latest but not including CRDs for the types included in the released operator image (if there is a delay between image release and the chart being updated).

Fixes the underlying problem in #1327 - now the helm chart won't be broken in that situation because it will explicitly still refer to the previously-released image.

gif


charts/azure-service-operator/templates/generated/apps_v1_deployment_azureoperator-controller-manager.yaml
```diff
@@ -18,16 +18,6 @@ spec:
...
-        - --secure-listen-address=0.0.0.0:8443
-        - --upstream=http://127.0.0.1:8080/
-        - --logtostderr=true
-        - --v=10
-        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.5.0
-        name: kube-rbac-proxy
-        ports:
-        - containerPort: 8443
-          name: https
       - args:
```
matthchr, line 21: Unsure if this change is expected, but it looks harmless so signing off.

matthchr (approved):

charts/azure-service-operator/templates/generated/apps_v1_deployment_azureoperator-controller-manager.yaml
babbageclunk: Yeah, I looked at that - as far as I can tell it's just moved the section from one place to another, presumably some non-determinism in the generator?

Issue #1351, opened 2021-01-11
Remove ClientID option on KeyVault AccessPolicy


matthchr:
**Describe the current behavior**
We have `ClientID` today and we try to translate that into an `ObjectID`.

**Describe the improvement**
It would be great if this worked universally but unfortunately depending on what permissions the ASO service principal or managed identity has on the subscription it may or may not have permission to call the API required to look up the `ObjectID` from the `ClientID`. The KeyVault API requires an `ObjectID` so rather than trying to be smart we should just make the customer supply the same ID that KeyVault expects.

matthchr:
Leaving this open as an FYI, but not intending to fix this at this time as V1 is in critical fixes only mode.

matthchr:
Closing this given ASOv1 is in maintenance mode and this isn't a critical issue

PR #1352, opened 2021-01-11
Provision AccessPolicy and KeyVault at same time

### LINKED ISSUE #1298: Bug: AzureKeyVault accesspolicy issues
**Describe the bug**
1. If we update the access policy by adding or removing a new entity, the access policy won't get updated in azure but the status of the CRD is successfully provisioned.
2. If we add/update an invalid identity to access policy (like a client id that doesn't exist at all), it doesn't show any error message in the crd status and the status of CRD gets successfully provisioned.

If you attempt to do the above with `networkPolicies` everything works as expected.

matthchr:
The operator was originally reconciling AccessPolicy's after the rest of the KV
had been created (see: #1158). This isn't actually required because even doing
this there are tons of reasons that this can fail. I've filed #1351 to track
removing ClientID from the CRD in a future API version as there are a ton of
obscure ways that we can fail to translate that ID into an ObjectID.

Closes #1298

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/pkg/resourcemanager/keyvaults/keyvault.go
+++ b/pkg/resourcemanager/keyvaults/keyvault.go
@@ -13,6 +13,14 @@ import (
 	auth "github.com/Azure/azure-sdk-for-go/services/graphrbac/1.6/graphrbac"
 	"github.com/Azure/azure-sdk-for-go/services/keyvault/mgmt/2018-02-14/keyvault"
 	kvops "github.com/Azure/azure-sdk-for-go/services/keyvault/v7.0/keyvault"
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/to"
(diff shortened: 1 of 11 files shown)


pkg/resourcemanager/keyvaults/keyvault.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/Azure/azure-service-operator/pkg/resourcemanager/iam"
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/to"
+	"github.com/pkg/errors"
```
babbageclunk, line 25: Reorder these imports too?

babbageclunk (dismissed):
Looks great!

babbageclunk (approved):

Issue #1346, opened 2021-01-05
Improve how database users are modeled for MySQL and PostgreSQL


matthchr:
**Describe the current behavior**
Our current Kubernetes object graph models database users at a per-database level across all supported database types. This has the advantage of allowing us to configure ownership so that if a user is created for a particular database (and both the user and database are being managed by ASO) then deleting the database will also delete the users in Kubernetes.

The downside of this though is that MySQL and PostgreSQL allow more flexible user management, where a single user may have permissions to many databases. While it's technically possible to have users with permissions to multiple databases in our current model it requires creating many ASO entities all pointing to the same database user. This is not recommended as the reconciliation for each entity will be in competition. Ownership is difficult as well in this model, as deleting one database should delete the user in Kubernetes assigned to that database, but _not actually delete the user in the database_ since there are other Kubernetes entities also referring to the same database user.

Even worse than the above: there are permissions which are _server_ level permissions and cannot be granted at the database level. We do not have a good way to model this currently.

**Describe the improvement**
Database users could be modelled more closely to how the various database products actually manage them, where really the user is at the server level and then is granted permissions to 1 or more databases. A CRD might look something like this:

```yaml
apiVersion: azure.microsoft.com/v1alpha2
kind: MySQLUser
metadata:
  name: mysqluser-sample
spec:
  server: mysqlserver-sample-777
  resourceGroup: resourcegroup-azure-operators
  roles:
    - "REPLICATION CLIENT"
    - "PROCESS"
  databaseRoles:
    mysqldatabase-sample:
      - "SELECT"
      - "INSERT"
      - "UPDATE"
      - "DELETE"
```

This has the following advantages:
1. Easily models what we currently allow.
2. Allows server level permissions.
3. Allows a single user (managed by a single entity in Kubernetes) to have permissions to multiple databases.

The main disadvantage of this approach is that if you want to delete the user entity in Kubernetes when the database is deleted you can no longer do that, as it would now be owned by the database server.

**Additional context**
This would be a **breaking change** and would (eventually) apply to PostgreSQL and MySQL database users, both AAD and normal "flavors" of user.

babbageclunk:
We've had some feedback pointing out that that the user model for Azure SQL matches well with the resources we have now, so we'll be leaving AzureSQLUser and AzureSQLManagedUser as they are. These changes will only be made to MySQLUser, MySQLAADUser and PostgreSQLUser types.

matthchr:
[...]

PR #1357, opened 2021-01-13
Add MySQLUser and MySQLAADUser v1alpha2

babbageclunk:
The new versions of MySQLUser and MySQLAADUser have roles per-database and server-level roles rather than a user being specific to one database. This adds conversions from v1alpha1 -> v1alpha2 (these are always possible), and supports conversion back to v1alpha1 as long as the user doesn't have any server-level roles and has roles in exactly one database. It also includes the yaml situps to connect the conversion webhooks.

This implements #1346 for MySQL. The initial plan had been to split this change into two PRs with the reconciliation changes in a follow-up, but because envtest doesn't support conversion webhooks the CI tests would have prevented this, so reconciliation and controller tests are included as the last two commits.

gif

**If applicable**:
- [x] this PR contains documentation (in the form of samples)
- [x] this PR contains tests


api/v1alpha1/mysqlaaduser_conversion.go
```diff
@@ -0,0 +1,69 @@
...
+	// ObjectMeta
+	dst.ObjectMeta = src.ObjectMeta
+
+	// Spec
+	dst.Spec.ResourceGroup = src.Spec.ResourceGroup
+	dst.Spec.Server = src.Spec.Server
+	dst.Spec.AADID = src.Spec.AADID
+	dst.Spec.Username = src.Spec.Username
+
+	// All roles should be interpreted as database roles in v1alpha2
```
matthchr, line 25: minor: This says what you're doing, but so does the code. I think it makes more sense to expand on the why, for example: "because that was the only kind of role supported by v1alpha1"

api/v1alpha1/mysqluser_conversion.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// Spec
+	dst.Spec.ResourceGroup = src.Spec.ResourceGroup
+	dst.Spec.Server = src.Spec.Server
+	dst.Spec.AdminSecret = src.Spec.AdminSecret
+	dst.Spec.AdminSecretKeyVault = src.Spec.AdminSecretKeyVault
+	dst.Spec.Username = src.Spec.Username
+	dst.Spec.KeyVaultToStoreSecrets = src.Spec.KeyVaultToStoreSecrets
+
+	// All roles should be interpreted as database roles in v1alpha2
```
matthchr, line 27: Same minor comment about sharing the why here

api/v1alpha1/mysqlaaduser_conversion.go
```diff
@@ -0,0 +1,69 @@
...
+	return nil
+}
+
+func (dst *MySQLAADUser) ConvertFrom(srcRaw conversion.Hub) error {
+	src := srcRaw.(*v1alpha2.MySQLAADUser)
+
+	// Converting a v1alpha2 user into a v1alpha1 one is only allowed
+	// if it has exactly one database...
+	if len(src.Spec.DatabaseRoles) != 1 {
+		return errors.Errorf("can't convert user %q to v1alpha1 MySQLAADUser because it has privileges in %d databases", src.ObjectMeta.Name, len(src.Spec.DatabaseRoles))
```
matthchr, line 42: super minor: Would it be worth doing `%T` and passing `dst` so we get the full type name (and also so that refactors catch it).

This is like... super mega minor though.

matthchr: Same thoughts apply to the standard MySQLUser logs too but not going to bother repeating them there.

api/v1alpha2/mysqlaaduser_conversion.go
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package v1alpha2
+
+func (*MySQLAADUser) Hub() {}
```
matthchr, line 6: Is this required for impling an interface? If so should we include an interface assertion in here too?

This file looks real weird to me but it also is just doing what the other conversion files are doing I see...

api/v1alpha2/mysqlaaduser_types.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// MySQLAADUserSpec defines the desired state of MySQLAADUser
+type MySQLAADUserSpec struct {
+	// +kubebuilder:validation:MinLength:1
```
matthchr, line 14: Use `=` not `:` for the `MinLength` stuff

api/v1alpha2/mysqlaaduser_types.go
```diff
@@ -0,0 +1,76 @@
...
+	ResourceGroup string `json:"resourceGroup"`
+
+	// The server-level roles assigned to the user.
+	Roles []string `json:"roles"`
+
+	// The database-level roles assigned to the user (keyed by
+	// database name).
+	DatabaseRoles map[string][]string `json:"databaseRoles,omitempty"`
+
+	// AAD ID is the ID of the user in Azure Active Directory.
```
matthchr, line 30: minor: I think often we have the required stuff ordered to the top of the file.

api/v1alpha2/mysqluser_conversion.go
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package v1alpha2
+
+func (*MySQLUser) Hub() {}
```
matthchr, line 6: Same question here

api/v1alpha2/mysqluser_types.go
```diff
@@ -0,0 +1,71 @@
...
+	// +kubebuilder:validation:Required
+	Server string `json:"server"`
+
+	// +kubebuilder:validation:Pattern=^[-\w\._\(\)]+$
+	// +kubebuilder:validation:MinLength:1
+	// +kubebuilder:validation:Required
+	ResourceGroup string `json:"resourceGroup"`
+
+	// The server-level roles assigned ot the user.
+	Roles []string `json:"roles,omitempty"`
```
matthchr, line 27: I think roles/database roles are optional, move them down below the `// optional` comment?

api/v1alpha2/mysqluser_types.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// MySQLUserSpec defines the desired state of MySqlUser
+type MySQLUserSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+	// +kubebuilder:validation:MinLength:1
```
matthchr, line 17: `=` not `:`

config/crd/patches/webhook_in_mysqlusers.yaml
```diff
@@ -5,6 +5,7 @@ kind: CustomResourceDefinition
 metadata:
   name: mysqlusers.azure.microsoft.com
 spec:
+  preserveUnknownFields: false
```
matthchr, line 8: Is there a story here? Wondering why we need to be explicit here but I don't see the same explicitness in other webhook defs (for example azuresqlusers)

config/samples/azure_v1alpha2_mysqlaaduser.yaml
```diff
@@ -0,0 +1,29 @@
...
+  # When creating a user for a "normal" (non-managed identity) user or
+  # group, this is the OID of the user or group.
+  aadId: 00000000-0000-0000-0000-000000000000
+  # Specify a list of server-level privileges with roles. Privileges
+  # could be one or more of the following: RELOAD, PROCESS, SHOW
+  # DATABASES, REPLICATION SLAVE, REPLICATION CLIENT, CREATE USER
+  # roles:
+  #  - PROCESS
+
+  # Soecify the database-level privileges this user should have.
```
matthchr, line 21:
```suggestion
  # Specify the database-level privileges this user should have.
```

config/samples/azure_v1alpha2_mysqlaaduser.yaml
```diff
@@ -0,0 +1,29 @@
...
+  databaseRoles:
+    mysqldatabase-sample:
+      # The privileges could be one or more of the following: SELECT,
```
matthchr, line 24:
```suggestion
      # The privileges can be one or more of the following: SELECT,
```

I think?

config/samples/azure_v1alpha2_mysqluser.yaml
```diff
@@ -0,0 +1,33 @@
...
+  server: mysqlserver-sample
+  dbName: mysqldatabase-sample
+  resourceGroup: resourcegroup-azure-operators
+  # Specify a list of server-level privileges with roles. Privileges
+  # could be one or more of the following: RELOAD, PROCESS, SHOW
+  # DATABASES, REPLICATION SLAVE, REPLICATION CLIENT, CREATE USER
+  # roles:
+  #  - PROCESS
+
+  # Soecify the database-level privileges this user should have.
```
matthchr, line 15:
```suggestion
  # Specify the database-level privileges this user should have.
```

config/samples/azure_v1alpha2_mysqluser.yaml
```diff
@@ -0,0 +1,33 @@
...
+  databaseRoles:
+    mysqldatabase-sample:
+      # The privileges could be one or more of the following: SELECT,
```
matthchr, line 18:
```suggestion
      # The privileges can be one or more of the following: SELECT,
```
I think?

matthchr (approved):

api/v1alpha2/mysqlaaduser_conversion.go
babbageclunk: Yeah, this is a marker method required for `v1alpha2.MySQLAADUser` to be the hub type for conversions. Really it's so they can specify the interfaces for conversions. Definitely feels pretty odd. I'll add assertions in both the v1 and v2 files.

api/v1alpha2/mysqlaaduser_types.go
babbageclunk: I've done this change in the v1alpha1 file too - I don't think there are any backwards compatibility concerns with that, right?

babbageclunk: Oh actually in that case Roles was required so I haven't changed it.

config/crd/patches/webhook_in_mysqlusers.yaml
babbageclunk: Hmm - I copied it from the other webhook patches (eg MySQLServer) after getting an error on deploy. The only ones that have it are the enabled ones (AzureSQLUser only has one version) - I'm going to add it to all of them to save the next person (or future me) the same confusion.

babbageclunk: Ha, although I can understand your confusion, since AzureSQLUser *does* have webhooks, just not conversions.

matthchr (approved):

matthchr:
Looks like CI failed - could be related to your changes as the tests that failed were some of the negative test cases in MySQLUser?

babbageclunk:
It turns out that envtest [doesn't yet support conversion webhooks](https://github.com/kubernetes-sigs/controller-runtime/issues/998). That means the tests for this PR wouldn't pass without the reconciliation and tests being updated to use the v1alpha2 versions of MySQLUser and MySQLAADUser, so I've included those in this PR (the last two commits).

pkg/resourcemanager/mysql/mysqlaaduser/reconcile.go
```diff
@@ -9,23 +9,24 @@ import (
...
-
-	_ "github.com/go-sql-driver/mysql" //sql drive link
-	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/types"
+	"github.com/Azure/azure-service-operator/pkg/resourcemanager/config"
+	"github.com/Azure/azure-service-operator/pkg/resourcemanager/mysql"
+	mysqlserver "github.com/Azure/azure-service-operator/pkg/resourcemanager/mysql/server"
 )
 
+const mysqlDatabase = "mysql"
```
matthchr, line 28: Is it worth a comment here explaining why this magic database name makes sense? (I assume that this is the default DB name at the server level that lets you manage users?)

pkg/resourcemanager/mysql/mysqlaaduser/reconcile.go
```diff
@@ -150,8 +153,7 @@ func (m *MySQLAADUserManager) Delete(ctx context.Context, obj runtime.Object, op
 	}
 
 	// short circuit connection if database doesn't exist
```
matthchr, line 155:
```suggestion
	// short circuit connection if server doesn't exist
```

pkg/resourcemanager/mysql/mysqlhelper.go
```diff
@@ -90,74 +90,119 @@ func ConnectToSQLDBAsCurrentUser(
...
-// ExtractUserRoles extracts the roles the user has. The result is the set of roles the user has for the requested database
-func ExtractUserRoles(ctx context.Context, db *sql.DB, user string, database string) (map[string]struct{}, error) {
-	if err := helpers.FindBadChars(user); err != nil {
-		return nil, errors.Wrapf(err, "problem found with username")
+func formatUser(user string) string {
+	// Wrap the user name in the weird formatting MySQL uses.
+	return fmt.Sprintf("'%s'@'%%'", user)
+}
+
+// extractUserRoles extracts the roles the user has. The result is the set of roles the user has for the requested database
```
matthchr, line 98:
```suggestion
// ExtractUserDatabaseRoles extracts the per-database roles that the user has. The user can have different permissions to each database. The details of access are returned in the map, keyed by database name.
```

pkg/resourcemanager/mysql/mysqlhelper.go
```diff
@@ -168,7 +213,53 @@ func GrantUserRoles(ctx context.Context, user string, database string, roles []s
 	return nil
 }
 
-func addRoles(ctx context.Context, db *sql.DB, database string, user string, roles map[string]struct{}) error {
+// EnsureUserDatabaseRoles revokes and grants database roles as needed
+// so they match the ones passed in. It returns warning messages if
```
matthchr, line 217: I'm not totally clear how the comment: 
> It returns warning messages if some databases don't exist

is correct?

1. Is `ExtractUserDatabaseRoles` succeeding (but with 0 roles) if the DB doesn't exist? It's not clear to me if that's an error case or an "empty return" case.
2. Other errors when adding/removing roles will also just get reported as warnings right? As a matter of fact based on my read of this function and its caller it seems like even if we cannot assign some roles to some DBs we will still say `Provisioned = true` and `State = "Succeeded"` - that seems probably not correct?

I can sorta see the reasoning around bypassing DB not exists (as maybe next reconcile loop it will), but wonder if we need to be more explicit about checking for errors that aren't that one. Even for that error it might make sense to leave this resource in a "provisioned=false" state as it's not done with all that was asked?

config/samples/azure_v1alpha2_mysqlaaduser.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: azure.microsoft.com/v1alpha2
+kind: MySQLAADUser
+metadata:
+  name: mysqlaaduser-sample
+spec:
+  server: mysqlserver-sample
+  dbName: mysqldatabase-sample
```
matthchr, line 7: There is no `dbname` field at this location in the new CRD?

config/samples/azure_v1alpha2_mysqluser.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: azure.microsoft.com/v1alpha2
+kind: MySQLUser
+metadata:
+  name: mysqluser-sample
+spec:
+  server: mysqlserver-sample
+  dbName: mysqldatabase-sample
```
matthchr, line 7: There is no `dbname` field at this location in the new CRD?

config/samples/azure_v1alpha2_mysqluser.yaml
```diff
@@ -0,0 +1,33 @@
...
+  resourceGroup: resourcegroup-azure-operators
```
matthchr, line 8: super duper mega minor: IMO put `resourceGroup` above `server` as it's the "parent" in the ownership hierarchy and it feels more natural to read that top to bottom (parent to child to childs-child and so on)

matthchr: Same applies to above sample too

controllers/mysql_combined_test.go
```diff
@@ -152,24 +157,31 @@ func RunMySQLUserHappyPath(ctx context.Context, t *testing.T, mySQLServerName st
...
 		mysql.MySQLDriverName,
 		fullServerName,
-		mySQLDBName,
+		"mysql",
 		mysql.MySQLServerPort,
 		adminUser,
 		adminPassword)
 	assert.NoError(err)
 
+	expectedRoles := mysql.SliceToSet(updatedRoles)
```
matthchr, line 166: May be lacking some tests around the the "DBName doesn't exist" case - it was explicitly mentioned in one of the comments but not sure that this test covers it?

pkg/resourcemanager/mysql/mysqluser/mysqluser_reconcile.go
```diff
@@ -23,6 +23,8 @@ import (
 	keyvaultSecrets "github.com/Azure/azure-service-operator/pkg/secrets/keyvault"
 )
 
+const mysqlDatabase = "mysql"
```
matthchr, line 26: Same comment about possibly worth leaving a "what the heck is this" comment? I think this is a magical DB name that is the default DB but not 100% sure

pkg/resourcemanager/mysql/mysqluser/mysqluser_reconcile.go
```diff
@@ -147,22 +149,27 @@ func (s *MySqlUserManager) Ensure(ctx context.Context, obj runtime.Object, opts
...
-		return false, fmt.Errorf("no roles specified for database user")
+	err = mysql.EnsureUserServerRoles(ctx, db, user, instance.Spec.Roles)
+	if err != nil {
+		err = errors.Wrap(err, "ensuring server roles")
+		instance.Status.Message = err.Error()
+		return false, err
 	}
 
-	err = mysql.GrantUserRoles(ctx, user, instance.Spec.DbName, instance.Spec.Roles, db)
+	warnings, err := mysql.EnsureUserDatabaseRoles(ctx, db, user, instance.Spec.DatabaseRoles)
```
matthchr, line 159: Same comment here about if we really know that what is ending up in warnings here is something we're ok classifying as a warning.

matthchr (commented):
Some minor comments asking for improved comments in a few places, and then one sorta big comment about the behavior when there are failures reconciling DB permissions.

pkg/resourcemanager/mysql/mysqlhelper.go
babbageclunk: Whoops, too hasty editing there, thanks!

pkg/resourcemanager/mysql/mysqlaaduser/reconcile.go
babbageclunk: Nice call - I've moved that constant into `pkg/resourcemanager/mysql ` and renamed the server port and driver name constants there so that they don't stutter.

config/samples/azure_v1alpha2_mysqlaaduser.yaml
babbageclunk: D'oh!

config/samples/azure_v1alpha2_mysqluser.yaml
babbageclunk: Yup yup, definitely agree about the hierarchy here.

pkg/resourcemanager/mysql/mysqlhelper.go
babbageclunk: This makes a lot of sense - after our discussion I'll accumulate the errors for different databases into one error from the Ensure function (even though we still apply the roles for the later databases on an error) and have the user still be provisioned=false in the reconciler.

matthchr (approved):

PR #1358, opened 2021-01-13
Fix bugs related to secret management and improve documentation

### LINKED ISSUE #1356: Bug: Creating Azure SQL user for SQL Server that wasn't created by operator fails
**Describe the bug**
Get an error like so:
```
NAME       PROVISIONED   MESSAGE
mattuser                 admin secret : default/matthchr-sql-test-serv-secret, not found in KeyvaultSecretClient: secret does not existkeyvault.BaseClient#GetSecret: Failure responding to request: StatusCode=404 -- Original Error: autorest/azure: Service returned an error. Status=404 Code="SecretNotFound" Message="A secret with (name/id) default-matthchr-sql-test-serv-secret was not found in this key vault. If you recently deleted this secret you may be able to recover it using the correct recovery command. For help resolving this issue, please see https://go.microsoft.com/fwlink/?linkid=2125182"
```

**To Reproduce**
1. Provision an Azure SQL server and database out of band
2. Attempt to create a user in that database using ASO.
3. Observe that creation fails because of ASOs inability to find the secret it's looking for

**Expected behavior**
The secret is found and the user is provisioned correctly.

### LINKED ISSUE #1360: Bug: Generated Kubernetes secret names don't match documentation
**Describe the bug**
The [documentation](https://github.com/Azure/azure-service-operator/blob/master/docs/howto/secrets.md) describes Kubernetes secret names as being the same as the resource's metadata.name. However, it appears that the actual name ends up being `{type}.{group}.{resource-name}`.

matthchr:
Closes #1356
Closes #1360

Add new AZURE_SECRET_NAMING_VERSION mode

The new mode allows us to fix inconsistencies in how secrets
were named without making a breaking change.
    
 - AppInsights created secrets in the same namespace as the resource but with name: "appinsights-<resourceGroup>-<resourceName>"
 - Storage created secrets in the same namespace as the resource but with name: "storage-<resourceGroup>-<resourceName>"
 - AzureSQL resources created resources with a different naming scheme as well.
 - Other resources created a secret in the same namespace with the secret name being the resource name.
    
The new V2 mode ensures that all resources create secrets in KeyVault and/or Kubernetes with a consistent naming pattern.

Return proper error if we cannot deserialize secret.

Improved secrets documentation

**If applicable**:
- [x] this PR contains tests


controllers/helpers.go
```diff
@@ -444,3 +442,61 @@ func GenerateRandomSshPublicKeyString() string {
 	sshPublicKeyData := string(ssh.MarshalAuthorizedKey(publicRsaKey))
 	return sshPublicKeyData
 }
+
+//CreateVaultWithAccessPolicies creates a new key vault and provides access policies to the specified user - used in test
+func CreateVaultWithAccessPolicies(ctx context.Context, creds config.Credentials, groupName string, vaultName string, location string, clientID string) error {
```
matthchr, line 447: This was copied from resourcemanager/keyvaults/keyvault.go as it's only used in test and it kept confusing me by being in the reconcile file.

controllers/suite_test.go
```diff
@@ -91,13 +90,18 @@ func setup() error {
 	// the purpose of these tests
 	envy.Set("POD_NAMESPACE", "azureoperator-system")
 
+	// Force the secret naming version to 2 for these tests
+	// Note: This can be removed to run the tests in the old v1 naming scheme (they
+	// should still pass)
+	envy.Set("AZURE_SECRET_NAMING_VERSION", "2")
```
matthchr, line 96: Wondering if I should make the default v2 and then force people to set it back to v1. That's what I did in the Helm chart but I didn't do it in code... should I do it in both?

docs/howto/secrets.md
```diff
@@ -1,28 +1,76 @@
...
+2. You can use a Key Vault with "Soft delete" enabled. However, you cannot use a Key Vault with "Purge Protection" enabled, as this prevents the
+   secrets from being deleted and causes issues if a resource with the same name is re-deployed.
 
-```
-export AZURE_OPERATOR_KEYVAULT=OperatorSecretKeyVault
-```
+## Secret naming
+
+There are two versions of secret naming used by the Azure Service Operator. The secret naming version is controlled by the `AZURE_SECRET_NAMING_VERSION` field of the `azureoperatorsettings` secret.
+Valid values are `1` and `2`. Version `2` is the default when installing the operator via Helm. If no value is specified for `AZURE_SECRET_NAMING_VERSION` the default is version `1` for backwards compatibility.
```
matthchr, line 38: As mentioned above, wondering if I should change this so that the default is 2 in code as well.

go.mod
```diff
@@ -13,38 +14,51 @@ require (
...
 	gopkg.in/ini.v1 v1.55.0 // indirect
 	k8s.io/api v0.18.6
-	k8s.io/apimachinery v0.18.6
+	k8s.io/apimachinery v0.18.8
+	k8s.io/apiserver v0.18.6 // indirect
 	k8s.io/client-go v0.18.6
-	sigs.k8s.io/controller-runtime v0.6.2
-	sigs.k8s.io/controller-tools v0.2.5 // indirect
+	k8s.io/utils v0.0.0-20200603063816-c1c6865ac451 // indirect
+	sigs.k8s.io/controller-runtime v0.6.0
```
matthchr, line 59: Why downgrade controller-runtime?

pkg/resourcemanager/azuresql/azuresqluser/azuresqluser.go
```diff
@@ -203,17 +197,14 @@ func (m *AzureSqlUserManager) DeleteSecrets(ctx context.Context, instance *v1alp
...
-
-			err = secretClient.Delete(
-				ctx,
-				key,
-			)
-			if err != nil {
-				instance.Status.Message = "failed to delete secret, err: " + err.Error()
-				return false, err
-			}
+		err = secretClient.Delete(
```
matthchr, line 200: This might be a problem because of the permissions that RBC has set up?

Might need to move this to best effort for now?

pkg/resourcemanager/azuresql/azuresqluser/azuresqluser.go
```diff
@@ -240,24 +231,23 @@ func (m *AzureSqlUserManager) GetOrPrepareSecret(ctx context.Context, instance *
 	return secret
 }
 
-// GetNamespacedName gets the namespaced-name
-func GetNamespacedName(instance *v1alpha1.AzureSQLUser, secretClient secrets.SecretClient) types.NamespacedName {
-	var namespacedName types.NamespacedName
-	keyVaultEnabled := reflect.TypeOf(secretClient).Elem().Name() == "KeyvaultSecretClient"
+func MakeSecretKey(secretClient secrets.SecretClient, instance *v1alpha1.AzureSQLUser) secrets.SecretKey {
+	if secretClient.GetSecretNamingVersion() == secrets.SecretNamingV1 {
+		// This check is really bad but we're keeping it for legacy reasons...
```
matthchr, line 236: Why is it bad? (Explain)

pkg/resourcemanager/azuresql/azuresqluser/azuresqluser_reconcile.go
```diff
@@ -234,19 +240,24 @@ func (s *AzureSqlUserManager) Ensure(ctx context.Context, obj runtime.Object, op
 					formattedSecrets["username"] = []byte(user)
 
 				case "password":
-					formattedSecrets["password"] = DBSecret["password"]
+					formattedSecrets["password"] = userSecret["password"]
 				}
 			} else {
-				err = sqlUserSecretClient.Delete(
```
matthchr, line 240: Same comment here -- it looks as if this error was previously just ignored, which may be important if customers have blocked deletes?

pkg/resourcemanager/azuresql/azuresqluser/azuresqluser.go
matthchr: Nope it's ok we were doing this before too.

pkg/resourcemanager/config/env.go
```diff
@@ -61,6 +62,17 @@ func ParseEnvironment() error {
 	creds.operatorKeyvault = envy.Get("AZURE_OPERATOR_KEYVAULT", "")         // operatorKeyvault()
 	testResourcePrefix = envy.Get("TEST_RESOURCE_PREFIX", "t-"+helpers.RandomString(6))
 	podNamespace, err = envy.MustGet("POD_NAMESPACE")
+	secretNamingVersionInt := ParseIntFromEnvironment("AZURE_SECRET_NAMING_VERSION")
+
+	// If this isn't set, default to the oldest support version
+	if secretNamingVersionInt == 0 || secretNamingVersionInt == 1 {
+		secretNamingVersion = secrets.SecretNamingV1
+	} else if secretNamingVersionInt == 2 {
```
matthchr, line 70: Are we ok with these code level defaults or should we be defaulting to 2?

pkg/secrets/keyvault/client.go
```diff
@@ -78,33 +92,32 @@ func IsKeyVaultAccessible(kvsecretclient secrets.SecretClient) bool {
...
-	if strings.Contains(err.Error(), errhelp.Forbidden) {
+	err = client.Delete(ctx, key)
+	if err != nil && strings.Contains(err.Error(), errhelp.Forbidden) {
 		return false
 	}
 
 	return true
 }
 
+// TODO: This isn't even used except in test, maybe just delete it?
```
matthchr, line 108: What are thoughts on me removing this in favor of just a single method `Upsert`?

controllers/suite_test.go
theunrepentantgeek: Yes, for consistency.

theunrepentantgeek (commented):
Partial review, I got about half way through.

Does this actually need the versioning config for the secrets config?

Could we instead default to the new one with fallback to the old style if it's missing? Back compat for existing users, new features for new users, or amy I missing something important?

pkg/resourcemanager/config/env.go
matthchr: Changed to default to 2 here as well.

api/v1alpha1/storageaccount_types.go
```diff
@@ -39,7 +39,7 @@ type StorageAccountSpec struct {
 // StorageAccountSku the SKU of the storage account.
 type StorageAccountSku struct {
 	// Name - The SKU name. Required for account creation; optional for update.
-	// Possible values include: 'StandardLRS', 'StandardGRS', 'StandardRAGRS', 'StandardZRS', 'PremiumLRS', 'PremiumZRS', 'StandardGZRS', 'StandardRAGZRS'
+	// Possible values include: 'Standard_LRS', 'Standard_GRS', 'Standard_RAGRS', 'Standard_ZRS', 'Premium_LRS', 'Premium_ZRS', 'Standard_GZRS', 'Standard_RAGZRS'
```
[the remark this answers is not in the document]
matthchr: Added

charts/azure-service-operator/templates/secret.yaml
```diff
@@ -20,4 +20,7 @@ data:
   {{- end }}
   {{- if .Values.azureOperatorKeyvault }}
   AZURE_OPERATOR_KEYVAULT: {{ .Values.azureOperatorKeyvault | b64enc | quote }}
+  {{- end }}
+  {{- if .Values.azureSecretNamingVersion }}
+  AZURE_SECRET_NAMING_VERSION: {{ .Values.azureSecretNamingVersion | b64enc | quote }}
   {{- end }}
```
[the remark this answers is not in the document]
matthchr: Added

config/default/manager_image_patch.yaml
```diff
@@ -51,6 +51,12 @@ spec:
                 key: AZURE_CLOUD_ENV
                 name: azureoperatorsettings
                 optional: true
+          - name: AZURE_SECRET_NAMING_VERSION
+            valueFrom:
+              secretKeyRef:
+                name: azureoperatorsettings
+                key: AZURE_SECRET_NAMING_VERSION
+                optional: true
```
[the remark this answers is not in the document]
matthchr: I think there's risk of migration trouble either way. If the customer was using the old style (all of our existing customers), they need to manually move their secrets to the new naming scheme (the secret _content_ is the same), which means there's some risk if they forget to force it back to v1 when upgrading.

controllers/appinsights_controller_test.go
```diff
@@ -7,23 +7,29 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
-	azurev1alpha1 "github.com/Azure/azure-service-operator/api/v1alpha1"
+	"github.com/stretchr/testify/assert"
```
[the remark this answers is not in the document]
matthchr: There's no good reason. In ASO the `controllers` tests use this and the `pkg/resourcemanager` and `api` tests use `gomega`. The `resourcemanager` tests also aren't run on CI (but the `api` tests are).

Which is all a long way of saying, I'm following the existing "pattern" here but the pattern is bad and we should have a concerted effort to fix the test + CI situation... but not in this PR =)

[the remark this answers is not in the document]
matthchr: Also some tests are written in this style and others are written in the BDD style with ginkgo. (I think it's the tests that use gomega that all also use BDD + ginkgo).

My personal preference would be to hold what we have in k8sinfra as the north start and move to gomega + non-BDD, but probably a discussion for another time.

controllers/azuresqlaction_controller_test.go
```diff
@@ -29,7 +30,14 @@ func RunSQLActionHappy(t *testing.T, server string) {
 	//Get SQL credentials to compare after rollover
 	secret := &v1.Secret{}
 	assert.Eventually(func() bool {
-		err := tc.k8sClient.Get(ctx, types.NamespacedName{Name: server, Namespace: "default"}, secret)
+		var secretName string
+		if tc.secretClient.GetSecretNamingVersion() == secrets.SecretNamingV1 {
+			secretName = server
+		} else {
+			secretName = "azuresqlserver-" + server
+		}
```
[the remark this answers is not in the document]
matthchr: Done

pkg/secrets/keyvault/client.go
```diff
@@ -52,23 +60,29 @@ func getVaultsURL(ctx context.Context, vaultName string) string {
...
+func IsKeyVaultAccessible(client secrets.SecretClient) bool {
 	ctx := context.Background()
-	key := types.NamespacedName{Name: "", Namespace: ""}
+	key := secrets.SecretKey{Name: "test", Namespace: "default", Kind: "test"}
+
+	data, err := client.Get(ctx, key)
+	if err == nil {
+		// TODO: It's not expected we ever take this path -- should this be an error?
+		return false
+	}
```
theunrepentantgeek, line 84: Comparing `err == nil` is jarring. Is that correct?

pkg/secrets/keyvault/client.go
```diff
@@ -307,55 +307,112 @@ func (k *KeyvaultSecretClient) Delete(ctx context.Context, key types.NamespacedN
...
+			if err != nil {
+				return err
+			}
+		}
+		return nil
 	} else {
-		secretName = key.Name
+		secretName, err := k.makeSecretName(key)
+		if err != nil {
+			return errors.Wrapf(err, "couldn't make secret name for %+v", key)
```
theunrepentantgeek, line 332: Don't we run a PII disclosure risk using `%+v` here with a secret?

pkg/secrets/keyvault/client.go
```diff
@@ -307,55 +307,112 @@ func (k *KeyvaultSecretClient) Delete(ctx context.Context, key types.NamespacedN
...
+	options := &secrets.Options{}
+	for _, opt := range opts {
+		opt(options)
+	}
+
+	if options.Flatten {
+		for _, suffix := range options.FlattenSuffixes {
+			secretName, err := k.makeSecretName(secrets.SecretKey{Name: key.Name + "-" + suffix, Namespace: key.Namespace, Kind: key.Kind})
+			if err != nil {
+				return errors.Wrapf(err, "couldn't make secret name for %+v", key)
```
theunrepentantgeek, line 321: Don't we run a PII disclosure risk using `%+v` here with a secret?

theunrepentantgeek (commented):
Overall looks good - within the limits of my understanding of the code anyway. 
A handful of comments.

Something feels repetitive - as though there's a missing abstraction, but I can't put my finger on it. I'll have another review with fresh eyes tomorrow.

pkg/resourcemanager/azuresql/azuresqluser/azuresqluser_reconcile.go
```diff
@@ -9,22 +9,49 @@ import (
...
 
+func GetAdminSecretKey(adminSecretName string, namespace string) secrets.SecretKey {
+	// TODO: Is there a better way to get TypeMeta here?
+	return secrets.SecretKey{Name: adminSecretName, Namespace: namespace, Kind: reflect.TypeOf(v1beta1.AzureSqlServer{}).Name()}
+}
+
+func (s *AzureSqlUserManager) getAdminSecret(ctx context.Context, instance *v1alpha1.AzureSQLUser) (map[string][]byte, error) {
+	adminSecretClient := s.SecretClient
+	adminSecretName := instance.Spec.AdminSecret
+	if len(instance.Spec.AdminSecret) == 0 {
```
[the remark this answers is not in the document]
matthchr: Fixed

pkg/resourcemanager/config/env.go
```diff
@@ -65,6 +66,20 @@ func ParseEnvironment() error {
...
+		return errors.Wrapf(err, "couldn't get AZURE_SECRET_NAMING_VERSION env variable")
+	}
+
+	// If this isn't set, default to the oldest support version
+	if secretNamingVersionInt == 1 {
+		secretNamingVersion = secrets.SecretNamingV1
+	} else if secretNamingVersionInt == 0 || secretNamingVersionInt == 2 {
+		secretNamingVersion = secrets.SecretNamingV2
+	} else {
+		return errors.Errorf("secret naming version must be between 0 and 2 but was %d", secretNamingVersionInt)
```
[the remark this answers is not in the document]
matthchr: Yeah good point, fixed.

pkg/resourcemanager/config/env.go
```diff
@@ -117,3 +132,11 @@ func ParseBoolFromEnvironment(variable string) bool {
 	}
 	return value
 }
+
+func ParseIntFromEnvironment(variable string) (int, error) {
+	value, err := strconv.Atoi(envy.Get(variable, "0"))
+	if err != nil {
+		return 0, errors.Wrapf(err, "invalid input value specified for int %v: \"%v\"", variable, value)
```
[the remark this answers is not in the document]
matthchr: Good catch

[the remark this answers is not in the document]
matthchr: This whole error message was poorly written as well, have improved

pkg/resourcemanager/storages/storageaccount/storageaccount.go
```diff
@@ -219,14 +219,18 @@ func (m *azureStorageManager) StoreSecrets(ctx context.Context, resourceGroupNam
...
+		secretKey,
 		data,
 		secrets.WithOwner(instance),
 		secrets.WithScheme(m.Scheme),
 	)
 }
+
+func (m *azureStorageManager) makeSecretKey(instance *v1alpha1.StorageAccount) secrets.SecretKey {
+	if m.SecretClient.GetSecretNamingVersion() == secrets.SecretNamingV1 {
+		return secrets.SecretKey{Name: fmt.Sprintf("storageaccount-%s-%s", instance.Spec.ResourceGroup, instance.Name), Namespace: instance.Namespace, Kind: instance.TypeMeta.Kind}
```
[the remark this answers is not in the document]
matthchr: Agreed - fixed.

pkg/secrets/keyvault/client.go
matthchr: It _is_ correct but there's a clearer way to write this method that also has the advantage of working if we previously had leaked the dummy key we're using to test.

Really this method needs to go away entirely (what it's doing is really not good), but I've kept it here for now because I didn't want to remove functionality that I don't fully understand why we have.

pkg/secrets/keyvault/client.go
matthchr: `key` is just the name of the secret, so `+v` is safe.

With that said maybe I should just override `Stringer` and use that for the key.... that probably is safer.

matthchr: Fixed this and also improved error message of `makeSecretName` so wrapping was not needed here (the error from `makeSecretName` now includes the requisite details via the `Stringer` interface).

pkg/secrets/keyvault/client.go
matthchr: As above, fixed this

config/default/manager_image_patch.yaml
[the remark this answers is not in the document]
theunrepentantgeek: But by leaving the default at v1, we're always going to be in this exact situation - with existing customers using v1. It isn't going to be easier to change the default to v2 when we have *more* users.

controllers/appinsights_controller_test.go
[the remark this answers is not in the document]
theunrepentantgeek: Fair enough.

pkg/resourcemanager/config/env.go
```diff
@@ -110,10 +125,20 @@ func GetExpectedConfigurationVariables() []ConfigRequirementType {
 }
 
 func ParseBoolFromEnvironment(variable string) bool {
-	value, err := strconv.ParseBool(envy.Get(variable, "0"))
+	env := envy.Get(variable, "0")
+	value, err := strconv.ParseBool(env)
 	if err != nil {
-		log.Printf("WARNING: invalid input value specified for bool %v: \"%v\", disabling\n", variable, value)
+		log.Printf("WARNING: invalid input value specified for %q, expected bool, actual: \"%v\". Disabling\n", variable, env)
```
theunrepentantgeek, line 131: Can you use %q to provide the quotes, given that `env` is a string?

pkg/secrets/keyvault/client.go
theunrepentantgeek: The reworked method looks clearer. 👍

theunrepentantgeek (dismissed):

pkg/resourcemanager/config/env.go
matthchr: Yeah good suggestion - done.

pkg/secrets/keyvault/client.go
babbageclunk: I like that.

pkg/secrets/keyvault/client.go
```diff
@@ -307,55 +309,112 @@ func (k *KeyvaultSecretClient) Delete(ctx context.Context, key types.NamespacedN
...
-	params := keyvault.KeyCreateParameters{
-		Kty:           keyvault.RSA,
-		KeySize:       &ksize,
-		KeyOps:        &kops,
-		KeyAttributes: &katts,
-	}
-	k.KeyVaultClient.CreateKey(ctx, vaultBaseURL, name, params)
-	return nil
+func (k *SecretClient) makeLegacySecretName(key secrets.SecretKey) (string, error) {
+	// The legacy behavior had differences between different kinds of resources - we need to preserve those differences
```
babbageclunk, line 380: Vestigial from when you were going to do all of the name generation here?

babbageclunk (approved):
This looks good, my only concern was how we ensure that the V1 secret naming path is tested. Thanks for discussion about that!

pkg/resourcemanager/azuresql/azuresqlmanageduser/azuresqlmanageduser.go
```diff
@@ -243,3 +212,15 @@ func findBadChars(stack string) error {
...
+
+func makeSecretKey(secretClient secrets.SecretClient, instance *v1alpha1.AzureSQLManagedUser) secrets.SecretKey {
+	if secretClient.GetSecretNamingVersion() == secrets.SecretNamingV1 {
+		if len(instance.Spec.KeyVaultSecretPrefix) != 0 { // If KeyVaultSecretPrefix is specified, use that for secrets
+			return secrets.SecretKey{Name: instance.Spec.KeyVaultSecretPrefix, Namespace: instance.Namespace, Kind: instance.TypeMeta.Kind}
+		}
+		return secrets.SecretKey{Name: instance.Name, Namespace: instance.Namespace, Kind: instance.TypeMeta.Kind}
+	}
+
+	// TODO: Ignoring prefix here... we ok with that?
```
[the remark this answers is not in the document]
matthchr: Yes. It's also taking an opportunity to remove some of the "extra" secrets configuration that only AzureSQLUser/AzureSQLManagedUser have that don't match the other types and doesn't really seem to add a lot of value to what customers can do (since it's not needed to avoid conflicts anymore it seems just extra complexity on our side to support it without much in the way of customer value add)

controllers/azuresql_combined_test.go
```diff
@@ -58,14 +58,17 @@ func TestAzureSqlServerCombinedHappyPath(t *testing.T) {
...
 	secret := &v1.Secret{}
 	assert.Eventually(func() bool {
-		err = tc.k8sClient.Get(ctx, types.NamespacedName{Name: sqlServerName, Namespace: sqlServerInstance.Namespace}, secret)
-
-		if err == nil {
-			if (secret.ObjectMeta.Name == sqlServerName) && (secret.ObjectMeta.Namespace == sqlServerInstance.Namespace) {
-				return true
-			}
+		var expectedServerSecretName string
+		if tc.secretClient.GetSecretNamingVersion() == secrets.SecretNamingV1 {
```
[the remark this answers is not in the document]
matthchr: As discussed, I will make some changes to the pipeline to do this

pkg/resourcemanager/config/env.go
```diff
@@ -65,6 +66,20 @@ func ParseEnvironment() error {
 		return errors.Wrapf(err, "couldn't get POD_NAMESPACE env variable")
 	}
 
+	secretNamingVersionInt, err := ParseIntFromEnvironment("AZURE_SECRET_NAMING_VERSION")
+	if err != nil {
+		return errors.Wrapf(err, "couldn't get AZURE_SECRET_NAMING_VERSION env variable")
+	}
+
+	// If this isn't set, default to the oldest support version
```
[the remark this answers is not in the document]
matthchr: That's because it is! Fixed.

pkg/secrets/keyvault/client.go
matthchr: Oh, yep. Removed.

pkg/secrets/keyvault/client.go
matthchr: Done

config/default/manager_image_patch.yaml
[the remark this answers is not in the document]
matthchr: Yeah agreed - I was thinking that this would ease migration for some existing customers but better to default the same in Helm + the code for sure.

controllers/async_controller.go
```diff
@@ -72,20 +72,21 @@ func (r *AsyncReconciler) Reconcile(req ctrl.Request, obj runtime.Object) (resul
...
 
-	if len(KeyVaultName) != 0 {
+	if len(keyVaultName) != 0 {
 		// Instantiate the KeyVault Secret Client
-		keyvaultSecretClient = keyvaultsecretlib.New(KeyVaultName, config.GlobalCredentials())
+		keyvaultSecretClient = keyvaultsecretlib.New(keyVaultName, config.GlobalCredentials(), config.SecretNamingVersion())
 
 		r.Telemetry.LogInfoByInstance("status", "ensuring vault", req.String())
 
+		// TODO: It's really awkward that we do this so often?
```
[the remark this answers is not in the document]
matthchr: Yeah I _think_ this is more about error code handling and setting the right message on the resource if we can't access the KeyVault, although doing a pre-check so that resources don't need to might be part of it as well.

Obviously as you point out, no amount of looking can guarantee your leap succeeds so pre-checking feels sorta pointless, and an alternative where we have an error handling library that can help us determine error causes presumably would mean that rather than doing this check once we can just farm it out to the error checker in all the places we interact with KV at which point there's not much value in doing this it seems to me as well? Especially since we need error handling in those other places already so it's not like we're forcing some there when previously we had none...

I've filed #1369 to track this as I don't think we want to bloat this PR further by me doing it here.

controllers/azuresql_combined_test.go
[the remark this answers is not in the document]
matthchr: Made these changes

controllers/secret_naming_version_test.go
```diff
@@ -0,0 +1,427 @@
...
+	kvsecrets "github.com/Azure/azure-service-operator/pkg/secrets/keyvault"
+)
+
+// NOTE: Tests in this file are intended to be run twice, once in the v1 secret naming mode
+// and once in the v2 secret naming mode. In some instances they parallel tests
+// which don't assert on secret name or shape.
+// The naming of the tests is important! It's how the makefile knows to run just these tests
+// in the old V1 mode.
+// Not every resource is included here, only resources which for legacy reasons in the old v1
+// secret naming mode had secret naming schemes which were different than the standard.
```
babbageclunk, line 34: Nice.

babbageclunk (approved):

PR #1371, opened 2021-02-05
Prevent deadlocks in reconcilation loop

matthchr:
Adds a context with a 5 minute timeout for all reconcile loop passes to ensure that if for whatever reason something gets stuck, it doesn't block all reconciliation forever.

Also ensure that we specify a timeout when connecting to Azure SQL to avoid deadlocks in that codepath (see https://github.com/denisenkom/go-mssqldb/issues/609)

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/controllers/async_controller.go
+++ b/controllers/async_controller.go
@@ -26,9 +26,10 @@ import (
 )
 
 const (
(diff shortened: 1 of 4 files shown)


babbageclunk (approved):
Good to have this as a backstop!

PR #1373, opened 2021-02-09
Fix release scripts

matthchr:
**If applicable**:
- [ ] this PR contains tests


Porges:
Is the fix just to remove Helm?

Porges (approved):

matthchr:
No it's also to switch away from `$()` syntax because that's specific to DevOps and use env variables instead ($... or ${})

PR #1376, opened 2021-02-11
Bug fixes for release 1.0.19504

matthchr:
- Fix issue where we were checking if a KeyVault was accessible very frequently. Instead just rely on error handling in the resource itself if creating a secret fails. This also gives a better error message.
- Fix bug in templates preventing webhook from getting correct certificate injected by `cert-manager`.


babbageclunk (approved):
Looks good - maybe expand the description?

matthchr:
Description expanded

PR #1395, opened 2021-03-18
RedisCache - ensure FailedProvisioning=false when Provisioned=true

### LINKED ISSUE #1394: Bug: RedisCache can be Provisioned and FailedProvisioning
**Describe the bug**
We've had reports of RedisCache being in the state Provisioned=true, FailedProvisioning=true.
From reading the code this can happen because it initially failed (possibly a temporary error that we aren't checking for), and then on a subsequent periodic reconciliation the ARM query reported that it was deployed successfully. The code that updates the state to Provisioned=true doesn't unset FailedProvisioning.

**To Reproduce**
* Deploy a RedisCache and hit a transient (but unknown to us) error
* The cache will be marked as failed
* Wait for the error to go away and the cache to be reconciled
* The cache will be marked as Provisioned=true, FailedProvisioning=true

**Expected behavior**
Ideally the transient error should be recognised so that the cache stays in provisioning until it passes, but at the least when it goes to provisioned the failed provisioning flag should be unset.

babbageclunk:
A user reported a case where a RedisCache was provisioned and working but was marked as FailedProvisioning=true. This seems to have happened because it initially failed (possibly a temporary error that we aren't checking for), and then on a subsequent periodic reconciliation the ARM query reported that it was deployed successfully. Use the SetProvisioned helper method (and friends) to prevent this state.

Fixes #1394 

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/pkg/resourcemanager/rediscaches/redis/rediscache_reconcile.go
+++ b/pkg/resourcemanager/rediscaches/redis/rediscache_reconcile.go
@@ -50,13 +50,13 @@ func (rc *AzureRedisCacheManager) Ensure(ctx context.Context, obj runtime.Object
 				instance.Status.Message = err.Error()
 				return false, err
 			}
-			instance.Status.State = string(newRc.ProvisioningState)
 
 			if newRc.StaticIP != nil {
 				instance.Status.Output = *newRc.StaticIP
 			}
 
 			instance.Status.ResourceId = *newRc.ID
+			instance.Status.State = string(newRc.ProvisioningState)
 			instance.Status.SetProvisioned(resourcemanager.SuccessMsg)
 			return true, nil
 		}
@@ -90,8 +90,6 @@ func (rc *AzureRedisCacheManager) Ensure(ctx context.Context, obj runtime.Object
 			instance.Status.SetProvisioning("RedisCache exists but may not be ready")
 			return false, nil
 		} else if helpers.ContainsString(catchKnownError, azerr.Type) {
-			// TODO: Not sure why we want this to be provisioning=false?
-			instance.Status.Provisioning = false
 			return false, nil
 		} else {
 			// serious error occured, end reconcilliation and mark it as failed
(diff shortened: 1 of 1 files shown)


pkg/resourcemanager/rediscaches/redis/rediscache_reconcile.go
```diff
@@ -50,16 +50,14 @@ func (rc *AzureRedisCacheManager) Ensure(ctx context.Context, obj runtime.Object
 				instance.Status.Message = err.Error()
 				return false, err
 			}
-			instance.Status.Message = resourcemanager.SuccessMsg
 			instance.Status.State = string(newRc.ProvisioningState)
```
matthchr, line 53: minor: I would put the state setting next to the message/provisioned/provisioning setting

pkg/resourcemanager/rediscaches/redis/rediscache_reconcile.go
```diff
@@ -90,18 +87,15 @@ func (rc *AzureRedisCacheManager) Ensure(ctx context.Context, obj runtime.Object
 
 		// handle the error
 		if helpers.ContainsString(catchInProgress, azerr.Type) {
-			instance.Status.Message = "RedisCache exists but may not be ready"
+			instance.Status.SetProvisioning("RedisCache exists but may not be ready")
 			return false, nil
 		} else if helpers.ContainsString(catchKnownError, azerr.Type) {
+			// TODO: Not sure why we want this to be provisioning=false?
```
matthchr, line 93: It comes down to a question of what we want `provisioning` to mean... I think what the original ASO authors were trying to convey with it was: "I am currently waiting for Azure to actually make the thing", and so they didn't count "waiting in Kubernetes for some reason" as `provisioning = true`.

I don't think that's really a good interpretation though because it's not stable: We set `provisioning = false` here but also return `false` which means "try again" - causing the reconcile loop to come back through and it the `SetProvisioning` on line 71... so the resource is going to blip between provisioning/not until whatever `knownError` is resolved. I think it's clearer to just say that it's provisioning the whole time and also include the message if needed explaining what is going on.

Since we already set the message to include the error details on line 74 I think you can just delete this line.

matthchr (dismissed):
LGTM - left a comment on your TODO though suggesting a minor change

matthchr (approved):

PR #1407, opened 2021-04-12
mysqlserver: Keep track of the polling URL when creating

babbageclunk:

Previously this was being dropped on the floor unless the create call returned an AsyncOpIncompleteError. This works fine as long as the server eventually is available. One situation where it fails is creating a replica at the same time as the leader server (or very quickly after). In that case the create server call returns no error, but querying the poll URL shows that the creation failed because the SourceServerID didn't match anything. Without capturing the poll URL on the initial creation the reconciliation gets stuck because it thinks that the 404 it gets when checking for the server indicates that it's just not ready yet - provisioning never goes back to false and it doesn't retry the creation. This could happen in other eventual error cases as well.

Re-enabled the replica creation in the MySQL test - we shouldn't be hitting the timeout errors we were seeing with it any longer.

gif

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/pkg/resourcemanager/mysql/server/reconcile.go
+++ b/pkg/resourcemanager/mysql/server/reconcile.go
@@ -190,9 +190,9 @@ func (m *MySQLServerClient) Ensure(ctx context.Context, obj runtime.Object, opts
 			instance.Status.Provisioned = false
 			instance.Status.FailedProvisioning = true
 			return true, nil
-
 		}
 
+		instance.Status.PollingURL = pollURL
 		instance.Status.Message = "request submitted to Azure"
 	}
 	return false, nil
--- a/pkg/resourcemanager/mysql/server/reconcile.go
+++ b/pkg/resourcemanager/mysql/server/reconcile.go
@@ -123,6 +123,7 @@ func (m *MySQLServerClient) Ensure(ctx context.Context, obj runtime.Object, opts
 				instance.Status.ResourceId = *server.ID
 				instance.Status.State = string(server.UserVisibleState)
 				instance.Status.SpecHash = hash
+				instance.Status.PollingURL = ""
 				return true, nil
 			}
 		}
@@ -176,7 +177,7 @@ func (m *MySQLServerClient) Ensure(ctx context.Context, obj runtime.Object, opts
 				if azerr.Type == errhelp.AsyncOpIncompleteError {
 					instance.Status.PollingURL = pollURL
 				}
-				instance.Status.Message = "Postgres server exists but may not be ready"
+				instance.Status.Message = "MySQL server exists but may not be ready"
 				instance.Status.Provisioning = true
 				return false, nil
 			}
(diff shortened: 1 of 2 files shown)


theunrepentantgeek (approved):

pkg/resourcemanager/mysql/server/reconcile.go
```diff
@@ -190,9 +190,9 @@ func (m *MySQLServerClient) Ensure(ctx context.Context, obj runtime.Object, opts
 			instance.Status.Provisioned = false
 			instance.Status.FailedProvisioning = true
 			return true, nil
-
 		}
 
+		instance.Status.PollingURL = pollURL
```
matthchr, line 195: Should we clear this when the operation is done?

pkg/resourcemanager/mysql/server/reconcile.go
```diff
@@ -190,9 +190,9 @@ func (m *MySQLServerClient) Ensure(ctx context.Context, obj runtime.Object, opts
 			instance.Status.Provisioned = false
 			instance.Status.FailedProvisioning = true
 			return true, nil
-
```
matthchr, line 193: Above on line 179 it says: ""Postgres server exists but may not be ready" - wrong DB type, this is mySQL.

Maybe fix that here too?

pkg/resourcemanager/mysql/server/reconcile.go
```diff
@@ -190,9 +190,9 @@ func (m *MySQLServerClient) Ensure(ctx context.Context, obj runtime.Object, opts
 			instance.Status.Provisioned = false
 			instance.Status.FailedProvisioning = true
 			return true, nil
```
matthchr, line 192: if this path was taken, wouldn't it mean that we're "done" and no future reconciliations will happen until the spec is changed again?

Just trying to get it straight in my head what exactly was happening here that caused things to get stuck and the only way I see is by taking this path (since all others return false which means keep trying?)

matthchr: but if that's the case, your fix won't fix it because this path would still get taken and then we wouldn't come around again?

Do we need to determine the specific error that's being hit and put it into the known errors collection so that we go around again?

pkg/resourcemanager/mysql/server/reconcile.go
babbageclunk: Oh, I had fixed that but accidentally removed it when I ripped out my logging - will redo.

pkg/resourcemanager/mysql/server/reconcile.go
babbageclunk: No, this error handling wasn't the path that caused the problem. The `CreateServerIfValid` call was returning no error, skipping the `if err != nil` and setting `Provisioning=true, Message="request submitted to Azure"`, returning `false, nil`. Then the next time through it would go down the `if instance.Status.Provisioning` branch, the GetServer on line 84 would return a 404 error but because there's no polling URL nothing would happen - no state change, and as long as the server never appeared the reconciliation would just keep happening forever.

matthchr: Doh, now I get it!

pkg/resourcemanager/mysql/server/reconcile.go
babbageclunk: Yeah, good idea - I'll clear it when we set Provisioned to true. It's probably still useful until then.

matthchr (approved):

theunrepentantgeek (approved):

babbageclunk:
/azp run

(looks like an intermittent failure in an Azure SQL secret naming test)

babbageclunk:
Looks like an intermittent failure in an Azure SQL legacy secret naming test.

/azp run

PR #1408, opened 2021-04-13
Draft of multitenancy document

babbageclunk:

This lays out our plan for supporting multitenancy in ASO to allow different groups to manage Azure resources using their own credentials within one k8s cluster.

**If applicable**:
- [ ] this PR contains tests


docs/multitenancy/proposal.md
```diff
@@ -0,0 +1,71 @@
...
+
+### CRDs and webhooks
+
+`CustomResourceDefinitions`, `ValidatingWebhookConfigurations` and `MutatingWebhookConfigurations` are cluster-wide so they can't be managed on a per-operator basis.
+In a multitenant configuration installing and managing the CRDs and webhooks will be a separate step from deploying operators (and will require cluster admin privileges).
+
+The conversion webhooks in the CRDs and the validating and mutating webhook configurations that are defined for different ASO resources will all point to the `azureoperator-webhook-service` service deployed in the `azureoperator-system` namespace (by default).
+The service will be backed by a deployment of the normal ASO operator running in a restricted webhook-only mode - it won't be watching any namespaces for Azure resources and won't have access to any Azure credentials.
+This mode will be triggered by the configuration setting `AZURE_WEBHOOK_ONLY=1`.
+With this set no other configuration values will be required, and no resource watches will be started - the operator will just respond to webhook requests from the API server.
```
davefellows, line 29:
```suggestion
With this setting no other configuration values will be required, and no resource watches will be started - the operator will just respond to webhook requests from the API server.
```

docs/multitenancy/proposal.md
```diff
@@ -0,0 +1,71 @@
...
+If the operator is configured to use managed identity (`AZURE_USE_MI=1`), then the `AzureIdentity` and `AzureIdentityBinding` will be found in that namespace as well.
+Each instance of the operator will use the service account defined in its namespace.
+(This is all the same as the current behaviour of the operator.)
+
+A new configuration attribute `AZURE_TARGET_NAMESPACES` will control which namespaces that operator will monitor.
+This should be a comma-separated list of names, which could include the namespace containing the operator deployment (but doesn't have to).
+Having no namespaces in the list or omitting the setting from the configuration will result in the operator watching all namespaces for ASO resources as it does now.
+
+// TODO: How to ensure that each namespace is managed by at most one operator? Having different operators (with potentially different credentials) fighting over a single resource would be messy. Similarly having both a global ASO and per-namespace instances would cause problems.
```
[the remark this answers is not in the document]
matthchr: Maybe we can look to other operators to see if they have any protections against this sort of thing.

docs/multitenancy/proposal.md
```diff
@@ -0,0 +1,71 @@
...
+The service will be backed by a deployment of the normal ASO operator running in a restricted webhook-only mode - it won't be watching any namespaces for Azure resources and won't have access to any Azure credentials.
+This mode will be triggered by the configuration setting `AZURE_WEBHOOK_ONLY=1`.
+With this set no other configuration values will be required, and no resource watches will be started - the operator will just respond to webhook requests from the API server.
+
+### RBAC
+
+At the moment installing a global instance of ASO creates a `ClusterRole` defining all of the permissions the operator requires on the different Azure resources (`azureoperator-manager-role`), and then a `ClusterRoleBinding` granting those permissions for resources in all namespaces (`azureoperator-manager-rolebinding`).
+A multitenant deployment won't have the global `ClusterRoleBinding`; instead each managed namespace N will need to contain a `RoleBinding` connecting the global `ClusterRole` with the service account of the operator that will manage N's resources.
```
matthchr, line 34: We'll have to talk about how to make management of this easy... or maybe just "easier" since it's inherently somewhat complicated

docs/multitenancy/proposal.md
```diff
@@ -0,0 +1,71 @@
...
+
+The biggest downside (other than complexity) is that the different operator installations will be separately managed and we can't assume that they will all be upgraded at the same time.
+The CRDs and webhook service deployment will need to be upgraded by the cluster administrator before any of the tenant operators can be upgraded.
+Conversely if there's a release of the operator that removes an old version of a resource definition, the cluster administrator will need to ensure that all of the tenant operators and resources have been upgraded beyond that version before the core CRDs and webhook service are upgraded.
+The [storage-version-migrator operator](https://github.com/kubernetes-sigs/kube-storage-version-migrator) might be useful for this.
+
+### Alternatives
+
+An earlier plan for implementing multitenancy in ASO was to keep the operator deployment as it is now (one per cluster) but add flexibility in how the credentials would be looked up.
```
matthchr, line 48: I think the eventual plan is still to do this, right?

As discussed for other customers such as CAPZ the N copies of operator pattern doesn't work.

docs/multitenancy/proposal.md
```diff
@@ -0,0 +1,71 @@
...
+The operator needs to support creating Azure resources using multiple credentials, managed by different groups, within one cluster.
+This will be enabled by changing the operator configuration to allow specifying a set of the cluster's namespaces to watch for resources to manage, and then running a deployment of the ASO operator per tenant.
+
+Each deployment of the operator will run in a separate namespace (rather than the current default `azureoperator-system`) and read its configuration from a secret named `azureoperatorsettings` in that namespace.
+If the operator is configured to use managed identity (`AZURE_USE_MI=1`), then the `AzureIdentity` and `AzureIdentityBinding` will be found in that namespace as well.
+Each instance of the operator will use the service account defined in its namespace.
+(This is all the same as the current behaviour of the operator.)
+
+A new configuration attribute `AZURE_TARGET_NAMESPACES` will control which namespaces that operator will monitor.
+This should be a comma-separated list of names, which could include the namespace containing the operator deployment (but doesn't have to).
```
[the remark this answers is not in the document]
davefellows: I suggest we stick with a simple list until we get an ask for anything else.

docs/multitenancy/proposal.md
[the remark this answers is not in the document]
davefellows: I'd say this is fine to punt for v1. Once we get to the point of building better deployment tooling to help ensure upgrades are done in the right way, maybe we can add this validation to that as well?
Should we add a section to this doc for logging/observability so at least we can ensure it's dead obvious in the logs if something like this accidentally happened.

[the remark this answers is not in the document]
babbageclunk: Yeah, that's a good suggestion about logging, I'll add a section to capture it.

docs/multitenancy/proposal.md
babbageclunk: Definitely - I'm vaguely picturing a tool to deploy it that would take the list of namespaces as input somehow and create the rolebindings at the same time as creating the deployment.

docs/multitenancy/proposal.md
babbageclunk: Oh good point. I don't think the credentials-per-namespace model of #1206 worked well for them either though - their model was to have multiple clusters in one namespace that could each be tagged with different credentials. I'm not sure how we'd map that into ASO: 
* By referring to credentials directly on ASO resources? 
* Some sub-namespace grouping of resources? 
* Resource group? That might be too restrictive.

Possibly it would make more sense to support their model directly without the rest of the credential fallback in #1206? (Although that would still require the code changes in ASO to make credential lookup per-request rather than when the reconciler is created.)

docs/multitenancy/proposal.md
```diff
@@ -0,0 +1,71 @@
+# Multitenancy approach
+
+This document presents our proposed approach for implementing multitenancy support in Azure Service Operator.
+
+## Overview
+
+The operator needs to support creating Azure resources using multiple credentials, managed by different groups, within one cluster.
```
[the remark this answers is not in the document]
babbageclunk: Yeah, good call - I'll add that here.

docs/multitenancy/proposal.md
```diff
@@ -0,0 +1,79 @@
...
+
+The biggest downside (other than complexity) is that the different operator installations will be separately managed and we can't assume that they will all be upgraded at the same time.
+The CRDs and webhook service deployment will need to be upgraded by the cluster administrator before any of the tenant operators can be upgraded.
+Conversely if there's a release of the operator that removes an old version of a resource definition, the cluster administrator will need to ensure that all of the tenant operators and resources have been upgraded beyond that version before the core CRDs and webhook service are upgraded.
+The [storage-version-migrator operator](https://github.com/kubernetes-sigs/kube-storage-version-migrator) might be useful for this.
+
+### Alternatives
+
+An earlier plan for implementing multitenancy in ASO was to keep the operator deployment as it is now (one per cluster) but add flexibility in how the credentials would be looked up.
```
matthchr, line 53: minor: tie this back in with "option 2" above?

docs/multitenancy/proposal.md
```diff
@@ -0,0 +1,79 @@
...
+A proof of concept to find credentials based on a resource's namespace can be seen in [PR #1206](https://github.com/Azure/azure-service-operator/pull/1206).
+This would be simpler in terms of deployments, but a global operator with access to different tenants' credentials carries the risk of applying the wrong credentials when reconciling an Azure resource.
+The operator-per-tenant architecture has been chosen to reduce that risk, since any given operator will only have access to the credentials it's configured with.
```
matthchr, line 56: Maybe rephrase this a bit as: "This would be simpler in terms of deployments, but a global operator with access to different tenants' credentials carries the risk of applying the wrong credentials when reconciling an Azure resource, a risk that may be unacceptable for customers who are very security conscious. We have chosen to implement the operator-per-tenant architecture first to enable the most security conscious customers. Along the way we will hopefully be able to lay the foundation for other multitenancy configurations as well."

Or something like that?

My main thinking here is just that this sounds like "that other way has failings that are insurmountable and so we're not going to do it" when in reality it's more like: "Both ways have failings but this way caters to the most security conscious customers, and we can always add the intermediate multi-tenancy (option 2) later"

matthchr (approved):

docs/multitenancy/proposal.md
```diff
@@ -0,0 +1,79 @@
+# Multitenancy approach
+
+This document presents our proposed approach for implementing multitenancy support in Azure Service Operator.
+
+## Overview
+
+The operator needs to support creating Azure resources using multiple credentials, managed by different groups, within one cluster. There are (at least) two ways we could handle multiple credentials in ASO:
+1. Allow running multiple operators in one cluster each with its own credentials and set of namespaces to watch for Azure resources.
+1. Keep the operator cluster-global but add a credential lookup when reconciling so that the identity used to communicate with Azure can be picked on a per-namespace (or per-resource) basis. A proof of concept of this approach can be seen in [PR #1206](https://github.com/Azure/azure-service-operator/pull/1206).
```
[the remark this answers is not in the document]
babbageclunk: That was deliberate - markdown will renumber them - but probably silly, no reason to make it confusing to someone reading the text file. Updated!

docs/multitenancy/proposal.md
babbageclunk: I like this wording, thanks!

docs/multitenancy/proposal.md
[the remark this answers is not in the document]
theunrepentantgeek: Isn't there supposed to be a blank line before the first item of a list in markdown?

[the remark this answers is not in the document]
babbageclunk: It renders ok with View File...
I'll add a blank to be on the safe side.

docs/multitenancy/proposal.md
```diff
@@ -0,0 +1,81 @@
...
+
+Initially we'll be working on option 1, but in the long term we expect to support both options for different use cases.
+
+Option 1 will be enabled by changing the operator configuration to allow specifying a set of the cluster's namespaces to watch for resources to manage, and then running a deployment of the ASO operator per tenant.
+Each deployment of the operator will run in a separate namespace (rather than the current default `azureoperator-system`) and read its configuration from a secret named `azureoperatorsettings` in that namespace.
+If the operator is configured to use managed identity (`AZURE_USE_MI=1`), then the `AzureIdentity` and `AzureIdentityBinding` will be found in that namespace as well.
+Each instance of the operator will use the service account defined in its namespace.
+(This is all the same as the current behaviour of the operator.)
+
+A new configuration attribute `AZURE_TARGET_NAMESPACES` will control which namespaces that operator will monitor.
```
eadasiak, line 20: Would the configuration attribute be dynamically read so that the list can be modified without interrupting the operator (say, from a configMap)?  In our environments, namespaces can be requested (and removed) by tenants on an ad-hoc basis, so we wouldn't necessarily want to have to redeploy the ASO just to manage the list of namespaces monitored, and the operator should handle the removal of a monitored namespace gracefully.

babbageclunk: Hmm, good point. You can avoid having to redeploy by using `kubectl rollout restart deployment azureoperator-controller-manager` but ideally it would happen automatically on changing the config. There's discussion of enabling that declaratively in kubernetes/kubernetes#29761 and kubernetes/kubernetes#24957, but I haven't got to the end of the second one yet - since it's from 2016 and still open, I'm not sure we want to wait for that.

I'd definitely want to do it by bouncing the operator process to ensure that everything has a consistent view. One option would be to mount the secret into a volume, watch that file with inotify and restart when it changes. The downside of that would be that all operator pods would restart at pretty much the same time. It's less of a problem for something like an operator where it just means that reconciles are delayed, compared with something that's serving network requests. 

Do we think that would be an issue? Will people be using multiple replicas of the operator in one deployment at all?

matthchr: > Will people be using multiple replicas of the operator in one deployment at all?

Do we even support this today? I don't think we do.

babbageclunk: I just tried it out. It works, although the different pods fight over reconciling the resource - it's noisy in the logs but ok since they're using the same config. I'm not sure whether anyone would want it though - maybe throughput if you're really creating a lot of resources?

eadasiak: Thanks @matthchr and @babbageclunk.  I think your comments both address my concern and also indicates that I got my scoping confused when I was considering the usage of the config attribute.  Since tenants would be running their own deployment of the operator, it would be entirely within their purview to manage the list of namespaces and handle the restarts accordingly.  

When I wrote this, I had option 2 on the brain and was imagining a centrally-managed list that would have to go through a longer change request and release cycle that can last weeks on our side.  Apologies for that!

docs/multitenancy/proposal.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Advantages and constraints
+
+The primary benefit of this architecture is that it enables different tenants to manage their operators independently along with their credentials and resources.
+Additionally in the case of operators using managed identities it removes the need to replicate `AzureIdentity` and `AzureIdentityBinding` resources between namespaces to ensure that the pod identity machinery configures the node identities correctly for the operator.
+
+The biggest downside (other than complexity) is that the different operator installations will be separately managed and we can't assume that they will all be upgraded at the same time.
+The CRDs and webhook service deployment will need to be upgraded by the cluster administrator before any of the tenant operators can be upgraded.
+Conversely if there's a release of the operator that removes an old version of a resource definition, the cluster administrator will need to ensure that all of the tenant operators and resources have been upgraded beyond that version before the core CRDs and webhook service are upgraded.
```
dharmab, line 49: A version compatibility matrix between the cluster admin components (CRD and webhook) and tenant components (namespaced operators) would be useful for validating this.

docs/multitenancy/proposal.md
```diff
@@ -0,0 +1,81 @@
...
+We have chosen to implement the operator-per-tenant architecture first to enable the most security conscious customers.
+Along the way we will hopefully be able to lay the foundation for other multitenancy configurations as well.
+
+Once that's ready we'll come back to implementing more flexible credential matching.
+There are also potential ASO users who would benefit from being able to assign credentials directly to resources, which can be implemented as an extension of the per-namespace credential handling.
+
+## Upgrading
+
+As discussed above, the CRDs and webhook service will need to be upgraded to a given version before any of the per-tenant operators can be upgraded.
+If a tenant operator is upgraded beyond the global version, the new operator pods will crash complaining about an unknown custom resource version until the deployment image version is rolled back.
```
dharmab, line 66: Is there a possibility of an inverse problem where a CRD or webhook update causes a previously functional tenant operator to no longer function?

If so, is a rollback of the CRD and webhook possible without disrupting any other tenant operators? Or would the outdated operator be forced to update?

docs/multitenancy/proposal.md
ryanzhang-oss: > I just tried it out. It works, although the different pods fight over reconciling the resource

Don't the controllers request a leader lock before starting to reconcile?

babbageclunk: > Don't the controllers request a leader lock before starting to reconcile?

Not by default at the moment - there's an `--enable-leader-election` option, but the deployment that gets created doesn't set it. (Nothing stopping a user from setting it in the deployment though.)

docs/multitenancy/proposal.md
babbageclunk: It's definitely possible for that to happen, especially in the case of dropping a version out of the CRD, but also if we incorrectly make a change to an old version. 

The difficulty with rolling back to an earlier version of the CRD and webhook processor is that there might already be resources using the new version (especially if the kube-storage-version-migrator is in use). Just replacing the CRD+webhooks with the previous operator version would cause errors because those resources would be a version that no longer exists in the CRD or conversion code. I think the right process would be to:

- Change the storage version selected on the CRD to the latest one supported by the previous version of the operator.
- Wait for the `StorageVersionMigration` to be completed so that all of the stored resources are downgraded to the previous version (which will still be in the CRD and recognised in the conversion code).
- Then roll back the CRDs and conversion webhook operator deployment.

matthchr (approved):

PR #1409, opened 2021-04-19
Switch to use the distroless nonroot base image and user

### LINKED ISSUE #1406: use nonroot user for manager container
The root user in the container is used

**Describe the improvement**
Instead use nonroot user

**Additional context**
As far as I am aware there is no need to have root permissions to run the operator.

babbageclunk:
We've already changed the webhook listener to start on 9443, so there's no need for the manager to run as root - this reduces the attack surface if the pod is compromised somehow.

Closes #1406 

gif


matthchr (approved):
Looks great! Love the GIF.

Issue #1392, opened 2021-03-17
Request: Openshift Operator Hub still using an old version of operator


lgmorand:
**Describe the bug**
Operator Hub offers version 0.37 of ASO. Can someone update it to a more recent version of the operator like a 1.x ?

**To Reproduce**
OpenShift > open Operator Hub

![operator-hub](https://user-images.githubusercontent.com/6757079/111494031-69578e80-873e-11eb-9590-2f4bbf95b5e5.png)

babbageclunk:
Yes - I'm working on this right now.

babbageclunk:
PRs for the updated bundle in the OpenShift and operatorhub stores.

https://github.com/operator-framework/community-operators/pull/4174
https://github.com/operator-framework/community-operators/pull/4175

matthchr:
Closing this as the above PRs have been merged

PR #1410, opened 2021-04-20
Update operatorhub bundle generation

babbageclunk:
Required for #1392 (although the bundle will still need to be updated afterwards).

The current operatorhub bundle needs updating, because it's a long way behind the current release. There were some changes made to the version of the bundle in the community-operators repository that were not propagated back here. This PR captures those changes, and extends the generation to emit a bundle that can be submitted, so we can keep it up to date for users of Openshift and OLM.

gif


Makefile
```diff
@@ -1,3 +1,6 @@
+PUBLIC_REPO=mcr.microsoft.com/k8s/azureserviceoperator
+PLACEHOLDER_IMAGE=controller:latest
+
 # Image URL to use all building/pushing image targets
 IMG ?= controller:latest
```
matthchr, line 5: Should this be using `PLACEHOLDER_IMAGE` now?

Makefile
```diff
@@ -342,9 +345,13 @@ endif
 # Current operator version
 VERSION ?= 0.37.0
 
+.PHONY: generate-operator-bundle
+generate-operator-bundle: LATEST_VERSION := $(shell curl -sL https://api.github.com/repos/Azure/azure-service-operator/releases/latest  | jq '.tag_name' --raw-output )
```
matthchr, line 349: Is this the latest version or the latest tag? Above in `helm-chart-manifests` this same value is called `LATEST_TAG`, but here it's `LATEST_VERSION` and `LATEST_TAG` includes the `$(PUBLIC_REPO)` prefix. We should be consistent in what we're calling it (since as far as I can tell it's the same concept in both places?)

Makefile
```diff
@@ -342,9 +345,13 @@ endif
...
+generate-operator-bundle: LATEST_TAG := $(PUBLIC_REPO):$(LATEST_VERSION)
+generate-operator-bundle: LATEST_SHA := $(shell docker pull $(LATEST_TAG) > /dev/null && docker image inspect $(LATEST_TAG) | jq --raw-output '.[0].Id' )
 generate-operator-bundle: manifests
 	kustomize build config/manifests | operator-sdk generate bundle --version $(VERSION) --channels stable --default-channel stable --overwrite
 	# This is only needed until CRD conversion support is released in OpenShift 4.6.x/Operator Lifecycle Manager 0.16.x
 	scripts/add-openshift-cert-handling.sh
-	# Rather than modify config/rbac manifests, replace CSV's default serviceAccount with azure-service-operator
-	sed -i 's/serviceAccountName: default/serviceAccountName: azure-service-operator/g' bundle/manifests/azure-service-operator.clusterserviceversion.yaml
+	# Inject the SHA-based container reference into the bundle.
+	scripts/inject-container-reference.sh "$(PUBLIC_REPO)@$(LATEST_SHA)"
```
matthchr, line 357: What's the argument for SHA here vs TAG? Is it some OLM/OperatorHub requirement? If so maybe note that in the comment?

config/manifests/bases/azure-service-operator.clusterserviceversion.yaml
```diff
@@ -165,6 +165,14 @@ spec:
       kind: KeyVault
       name: keyvaults.azure.microsoft.com
       version: v1alpha1
+    - description: Add an AAD user to an existing MySQL database.
```
matthchr, line 168: Just confirming -- this file is generated somehow is that right?

config/manifests/kustomization.yaml
```diff
@@ -1,3 +1,7 @@
 resources:
 - ../default
 - ../samples
+
+patches:
+  - patches/serviceaccountname_in_deployment.yaml
```
matthchr, line 6: Just confirming that this makes sense to do in the case of Helm deployments as well? Any backcompat concerns?

scripts/inject-container-reference.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+sha_reference="$1"
+now="$(date --rfc-3339=seconds)"
+csvfile="bundle/manifests/azure-service-operator.clusterserviceversion.yaml"
```
matthchr, line 6: This isn't a csvfile?

Makefile
babbageclunk: I've updated this.

Makefile
babbageclunk: I'm not sure whether it's required - I was maintaining it because that's what our existing one uses. 
Looking at bundles in the community-operators repo most of them don't use that, so I'll simplify it to match what we do for helm.

Makefile
babbageclunk: I'll make them consistent.

config/manifests/bases/azure-service-operator.clusterserviceversion.yaml
babbageclunk: No, unfortunately it isn't. If the CRDs aren't added here they'll still show up in the output cluster service version, but without any description. I don't think there's a downside to that other that in the Openshift UI for creating resources there won't be any explanatory text, but I figured I'd add it here.

There are doc comments on the Go types for the resources, but controller-gen puts that into the schema, and the `operator-sdk bundle` command doesn't use them to populate this field. (The existing ones on our resources aren't great, but we could update them if they were used.) As far as I can tell there isn't a description field on the [CRD k8s type](https://kubernetes.io/docs/reference/kubernetes-api/extend-resources/custom-resource-definition-v1/) for describing the resource as a whole.

A possible option would be to add another build step before this to extract the schema descriptions from the files in  `config/crd/bases`, and compile them into a patch that got stitched into the CSV file that then gets fed into `operator-sdk bundle`? (Plus updating the existing doc comments.)

config/manifests/kustomization.yaml
babbageclunk: No, this isn't used in the helms chart - it's only used to produce the yaml that's fed into the operator bundle. Although it's pretty confusing that this directory is called config/manifests - it would be clearer to call it `config/operator-hub`. I'll rename it.

scripts/inject-container-reference.sh
babbageclunk: It's a Cluster Service Version file. :grin: I blame the Openshift people for picking an overloaded acronym! But I'll rename the variable to something less confusing.

config/manifests/kustomization.yaml
babbageclunk: Oh weird - for some reason it also needs to be passed the `--kustomize-dir` if it's not `config/manifests`, even though the content of that is what's sent on stdin.

scripts/inject-container-reference.sh
matthchr: haha I totally missed that!

Makefile
```diff
@@ -339,12 +342,16 @@ else
...
 generate-operator-bundle: manifests
-	kustomize build config/manifests | operator-sdk generate bundle --version $(VERSION) --channels stable --default-channel stable --overwrite
+	rm -r bundle
+	@echo "Latest released tag is $(LATEST_TAG)"
+	kustomize build config/operator-bundle | operator-sdk generate bundle --version $(LATEST_TAG) --channels stable --default-channel stable --overwrite --kustomize-dir config/operator-bundle
 	# This is only needed until CRD conversion support is released in OpenShift 4.6.x/Operator Lifecycle Manager 0.16.x
 	scripts/add-openshift-cert-handling.sh
-	# Rather than modify config/rbac manifests, replace CSV's default serviceAccount with azure-service-operator
-	sed -i 's/serviceAccountName: default/serviceAccountName: azure-service-operator/g' bundle/manifests/azure-service-operator.clusterserviceversion.yaml
+	# Inject the SHA-based container reference into the bundle.
```
matthchr, line 353: minor: It's not SHA based anymore?

config/manifests/bases/azure-service-operator.clusterserviceversion.yaml
matthchr: Yeah, might be worth filing an issue for this. What I'm seeing that is awkward here is:
1. If I add a new resource, how do I know I need to modify this file? Is there any automation that will detect a mismatch between supported resources and resources listed in the ClusterServiceVersion file?
2. Having to duplicate descriptions of resources as you pointed out is awkward.
3. Having our README duplicated in here is also awkward (because presumably we'll update it and it'll get out of date).

matthchr (dismissed):

config/manifests/bases/azure-service-operator.clusterserviceversion.yaml
babbageclunk: I'll file an issue for it. The good part is that a new resource will still get included in the bundle, just with a blank description in the creation UI - not ideal but not a deal-breaker either. The descriptions aren't actually duplicates... but only because the current doc comments on the types are pretty generic at the moment. Moving them over should be part of that work too.

The README isn't quite a duplicate of our normal one because the install instructions are specific to OLM - that's also not ideal. I'll make an issue for that too.

matthchr (approved):
Looks great!

PR #1418, opened 2021-04-22
Update helm chart for 1.0.22275 operator release

babbageclunk:
Updates the helm chart to refer to the new operator image. I incremented the chart version to 1.1.22275 since it was already past the operator version, but at least this way we can keep updating it when new versions are released. Also updated the appVersion to match the image.


babbageclunk:
I'm not sure why all of the line-wrapping changed in the CRDs. Could be helm version? Mine is 3.5.4 which looks like the latest version, about a week old.

matthchr:
My preference for chart version is to keep it at 3 digits: `1.2.0` for this release. Of course that won't match the container image version -- maybe we can update the container image to start with `1.2.xxx` as well?

Wondering if we need a document that outlines what exactly to do for the release process, since there are a few steps and a few things that have to get changed...

matthchr:
I think the line-wrapping has to do with the version of Kustomize you're running? See https://github.com/kubernetes-sigs/kustomize/issues/947 (which I think we've referenced before).

We probably need to set up a dockerfile and run the generation and helm stuff in a container so that we guarantee everybody is using the same versions of the tools...

babbageclunk:
> My preference for chart version is to keep it at 3 digits: 1.2.0 for this release. Of course that won't match the container image version -- maybe we can update the container image to start with 1.2.xxx as well?
Ok - so we'd just increment it with each new release - this is 1.2.0, the next is 1.3.0...
Keeping the patch number for bugfixes for the chart, I guess?

> Wondering if we need a document that outlines what exactly to do for the release process, since there are a few steps and a few things that have to get changed...
Yeah, I was planning on adding sections for this and the operator bundle to the existing document once they were done (and I was clear about what actually needs to be done 😄).

babbageclunk:
> See above comment about `1.1.xxxx` vs `1.2.0`.

Done. (Weird that I can't reply to this comment?)

> I think the line-wrapping has to do with the version of Kustomize you're running? See kubernetes-sigs/kustomize#947 (which I think we've referenced before).

Chasing down the rabbit hole to go-yaml/yaml#455 and go-yaml/yaml#670 it seems like this isn't going to be fixed any time soon. It doesn't seem to cause any problem in the CRDs (they're in the API server without the linebreaks if you dump them as JSON so I think they're stripped out), and it's from the current version of kustomize so I think we should leave them in.

matthchr (approved):

PR #1421, opened 2021-04-27
Make tenant and subscription IDs optional in the config

babbageclunk:
Having the tenant and subscription required causes issues in the community-operators CI for the operator bundle, because it tries to deploy the operator without any configuration. There is precedent here: client ID is already marked as optional.  Additionally tenant and subscription would need to become optional as part of implementing the operator-per-namespace multitenancy model. 

If the fields are left out the error is correctly reported on resources when they are reconciled.

### DIFF
--- a/config/default/manager_image_patch.yaml
+++ b/config/default/manager_image_patch.yaml
@@ -28,11 +28,13 @@ spec:
               secretKeyRef:
                 name: azureoperatorsettings
                 key: AZURE_TENANT_ID
+                optional: true
           - name: AZURE_SUBSCRIPTION_ID
             valueFrom:
               secretKeyRef:
                 name: azureoperatorsettings
                 key: AZURE_SUBSCRIPTION_ID
+                optional: true
           - name: AZURE_USE_MI
             valueFrom:
               secretKeyRef:
--- a/go.sum
+++ b/go.sum
@@ -631,6 +631,7 @@ k8s.io/utils v0.0.0-20200603063816-c1c6865ac451/go.mod h1:jPW/WVKK9YHAvNhRxK0md/
 sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.7/go.mod h1:PHgbrJT7lCHcxMU+mDHEm+nx46H4zuuHZkDP6icnhu0=
 sigs.k8s.io/controller-runtime v0.6.5 h1:DSRu6E4FBeVwd/p8niskCVWnX5TSC6ZT9L/OIWOBK7s=
 sigs.k8s.io/controller-runtime v0.6.5/go.mod h1:WlZNXcM0++oyaQt4B7C2lEE5JYRs8vJUzRP4N4JpdAY=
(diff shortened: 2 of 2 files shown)


matthchr (commented):
> Having them being required causes issues in the community-operators CI for the operator bundle, because it tries to deploy the operator without any configuration.

Might be worth following up/mentioning that this is a reasonable desire? The ability to decouple: "Deploy this" and "make CRDs"?

> They're not more required than client ID which is already marked as optional

ClientID is optional if you're using Managed Identity, so this isn't quite true I don't think.

> Tenant and subscription would need to become optional as part of implementing the operator-per-namespace multitenancy model anyway

Agreed!


One thought though... if the operator isn't crashing when it's initially deployed, are we "safe" in the deployment order that we have now in Helm and elsewhere? Wondering if this might cause issues because the operator will allow itself to start with these fields empty, and since we don't refresh them when the secret is updated (even if that's just 2s later?) they won't take effect?

I hadn't thought of that when I said we should just do this but wondering if that's a reason we shouldn't?

babbageclunk:
> Might be worth following up/mentioning that this is a reasonable desire? The ability to decouple: "Deploy this" and "make CRDs"?

Good call, I'll add that.

> > They're not more required than client ID which is already marked as optional
> 
> ClientID is optional if you're using Managed Identity, so this isn't quite true I don't think.

Is that right? In the [install docs](https://github.com/Azure/azure-service-operator/blob/master/docs/howto/deploy.md#managed-identity-authentication) we still set client ID for managed identity, it's just client secret that isn't set. Although maybe it's not needed and we could be getting it from the `AzureIdentity`?

> One thought though... if the operator isn't crashing when it's initially deployed, are we "safe" in the deployment order that we have now in Helm and elsewhere? Wondering if this might cause issues because the operator will allow itself to start with these fields empty, and since we don't refresh them when the secret is updated (even if that's just 2s later?) they won't take effect?
> 
> I hadn't thought of that when I said we should just do this but wondering if that's a reason we shouldn't?

Hmm, so if we create the deployment first and then the secret, there's a chance that the operator comes up with empty config and then never sees the secret once it's created. While if the pod has at least one non-optional env variable, it will crash until the secret is created. Is that right? I'll have a look at the helm chart to see if that could be a problem.

babbageclunk:
Running the helm chart in --dry-run shows that the secret is created first, and when deploying the operator bundle the instructions are to create the secret first, so I think we're fine on that front.

matthchr (approved):

babbageclunk:
Seems like the tests are failing to get any available clusters?

babbageclunk:
Hmm, looking in the portal, all of the clusters in the CI subscription are tagged with `freeforpipeline=false`. :(

matthchr:
I think theres a bug where they can get orphaned with that... not exactly sure how but I had to once before clear the flag manually.

babbageclunk:
Ok, I've reset them all to true.

babbageclunk:
This time it failed because there was a timeout creating a resource group? :(

babbageclunk:
Woot!

PR #1424, opened 2021-05-07
Update CRDs to v1 apiversion

### LINKED ISSUE #1270: CRDs should be updated to v1 from v1beta1
Currently our CRDs are `apiextensions.k8s.io/v1beta1`

**Describe the improvement**
We should move to `apiextensions.k8s.io/v1`

matthchr:
Closes #1270.

  - Update CRDs to use v1 of the CRD API.
  - Forces preserveUnknownFields to false.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,8 +11,7 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
-# Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
-CRD_OPTIONS ?= "crd:crdVersions=v1beta1"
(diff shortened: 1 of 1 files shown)


matthchr:
I tested upgrade on a simple cluster that already had the old CRDs deployed into it. New CRD application went fine. I don't _think_ we need to worry about having existing resources doing things in the cluster (mine didn't have much).

babbageclunk (approved):
Looks good! It sounds like there are still a number of problems with switching the webhook configurations to v1 as well. :frowning_face:

PR #1428, opened 2021-05-12
Update patch targets to CRD v1

babbageclunk:
These were missed when the CRD generation was updated to produce v1 (#1424). Weirdly there's nothing in the CI pipeline that uses the patched CRDs - I'll add an issue for that.

gif


matthchr (approved):
Thanks for catching and fixing this so quickly @babbageclunk

babbageclunk:
Yay, thanks for fixing this @Porges!

PR #1475, opened 2021-05-14
Add /hack/ directory to IGNORE_FILTERS

Porges:
This adds the `hack` directory (where `k8s-infra`-sourced code lives) to the `IGNORE_FILTERS` set in `check-changes.sh`, which means that the ASO pipeline will not need to be re-run when only `k8s-infra` code changes.

### DIFF
--- a/scripts/check-changes.sh
+++ b/scripts/check-changes.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e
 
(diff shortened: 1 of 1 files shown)


theunrepentantgeek (approved):
:shipit:

matthchr (approved):

PR #1482, opened 2021-05-17
Create PropertyBag data structure and test.

### LINKED ISSUE #1479: Feature: Property Bag data type
Design and implement the property bag data structure, ready for injection into all Storage types, both resources and object types. 

Essentially the bag is a `map[string]string` but with some helper functions to make the generated code easier to use.

theunrepentantgeek:

Provides an implementation of a PropertyBag that we can use within our storage types to cache additional information not directly supported by the declared properties.

gif

**If applicable**:
- [x] this PR contains tests

Closes #1479


hack/generated/pkg/genruntime/property_bag.go
```diff
@@ -0,0 +1,89 @@
...
+	case float64:
+		bag[property] = strconv.FormatFloat(v, 'e', -1, 64)
+	case bool:
+		if v {
+			bag[property] = "true"
+		} else {
+			bag[property] = "false"
+		}
+	default:
+		// Default to treating as a JSON blob
```
Porges, line 51: The default case here subsumes all the above cases as they will all serialize as specified, except for `string`. Worth just having that?

Porges (commented):
Just one thing to think about.

hack/generated/pkg/genruntime/property_bag.go
theunrepentantgeek: Unexpected - but all the tests pass with the simpler implementation, so I'm happy.

Porges (approved):

PR #1485, opened 2021-05-18
Fix docs for StorageAutogrow

theunrepentantgeek:
**What this PR does / why we need it**:

Updates the documentation for `StorageAutogrow` to correctly show the available values as `Enabled` and `Disabled`.

Thanks @maggiexian for reporting the issue.

**How does this PR make you feel**:
[#puppies](https://media.giphy.com/media/MlIz6P8tUKAIeo1jgN/giphy.gif)

**If applicable**:
- [x] this PR contains documentation
- [ ] this PR contains tests

Closes #1483

Issue #1483, opened 2021-05-17
Bug: Description of property 'storageAutogrow' is wrong


maggiexian:
**Describe the bug**
The description of property 'storageAutogrow' is wrong. 
https://github.com/Azure/azure-service-operator/blob/master/charts/azure-service-operator/crds/apiextensions.k8s.io_v1beta1_customresourcedefinition_mysqlservers.azure.microsoft.com.yaml#L216 
When I set the value to ''StorageAutogrowEnabled'' as the description suggested, the deploy failed with "Original Error: Code=\"InvalidStorageAutoGrow\" Message=\"Invalid Storage AutoGrow Value\" "

**Expected behavior**
It turns out the allowed values are "Enabled" and "Disabled". Please fix the property definition.

theunrepentantgeek:
Thanks for the bug report @maggiexian.

PR #1490, opened 2021-05-18
Improve validation for some fields with enumerations

theunrepentantgeek:

Adds validation for a few properties that have a fixed set of values, ensuring errors are reported by kubectl instead of waiting for an ARM failure.

Followup to #1483 and #1485

[#spiffy](https://media.giphy.com/media/gg2sGd9VH2QQg2Lq4D/giphy.gif)


api/v1alpha2/sharedazuredbs_types.go
```diff
@@ -50,6 +51,7 @@ const (
 	PSQLMemoryOptimized SkuTier = "MemoryOptimized"
 )
 
+// +kubebuilder:validation:Enum=Enabled;Disabled
```
matthchr, line 54: minor: You have two formats here, the shorter/simpler one and the longer/more verbose one. Feels like we should have a single format we prefer?

In most of the existing ASO code it seems to be the `Enabled;Disabled` format, but the generator is doing `{"Enabled","Disabled"}`. Unsure which we should pick but feels like 1 style per file is better?

matthchr (approved):

babbageclunk (approved):

api/v1alpha2/sharedazuredbs_types.go
theunrepentantgeek: This syntax works, and I didn't get the other syntax to work in my testing - so I'd prefer to leave this as is.

theunrepentantgeek: (Which is not to say the more verbose syntax is wrong, maybe I had a local issue.)

matthchr: What about the more verbose syntax you use above?
// +kubebuilder:validation:Enum={"Basic","GeneralPurpose","MemoryOptimized"}

PR #1492, opened 2021-05-18
Improve record/replay tests

### LINKED ISSUE #1432: Need better error if recording tests body mismatches
It's currently very difficult to debug a recording test body mismatch. You see a bunch of error messages like this:
> [controller:test-integration-envtest-cover] E0317 02:54:57.574042   20848 generic_controller.go:211] NamespaceController "msg"="Error during reconcile" "error"="getting deployment \"/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/k8sinfratest-rg-kjqcwl/providers/Microsoft.Resources/deployments/k8s_5b3a7c53-665f-57f5-b02a-b01a67869999\" from ARM: Get \"https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/k8sinfratest-rg-kjqcwl/providers/Microsoft.Resources/deployments/k8s_5b3a7c53-665f-57f5-b02a-b01a67869999?api-version=2019-10-01\": Requested interaction not found" "name"="k8sinfratest-sbnamespace-kbvici" "namespace"="k8s-infra-test-ns" "action"="MonitorDeployment"

BUT:

1. It doesn't mention that the reason for the "Requested interaction not found" is from VCR at all (you need to infer that).
2. It doesn't say what the body was (so you can't tell _why_ it mismatched).
3. The controller doesn't understand that this should be a fatal error so it doesn't actually fail the test until 10m later when it hits a timeout.

It's possible that an easy out here is to panic with a lot more details about what is going on. That would give us a full stacktrace... although I think we need to be careful because panicing in a go test if not trapped at the top level ends up making the tests kindof look like they passed I think?

Porges:
A few changes here:

### Check bodies in requests

This makes sure the request/response pair is for the correct request (by default, `go-vcr` only checks the HTTP method/URL). To achieve this we blank out any dates/timestamps that are found in bodies, since they cause a mismatch when the test is rerun.

### Fix "count" header addition

We had code to track the request count at a per-URL level so that we handle multiple requests to the same URL properly in recordings (so that a series of polling GETs can get back different responses, for example), but this was not being added properly. Now it is.

### Better messages when go-vcr interaction cannot be found

At the moment when running record/replay tests and an interaction (request+response) cannot be located, it is hard to figure out what went wrong. This PR adds a wrapper around the `go-vcr` code to add more context about what went wrong.

An example message in the output looks like:

```
*** Cannot find go-vcr recording for request (no responses recorded for this method/URL): DELETE https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/k8sinfratest-rg-kjqcwl?api-version=2020-06-01
```

Ideally we would `panic` when this happens, but we can't at the moment (the code explains why).

Closes #1432.

VCR error

### DIFF
--- a/hack/generated/pkg/testcommon/error_translating_roundtripper.go
+++ b/hack/generated/pkg/testcommon/error_translating_roundtripper.go
@@ -71,24 +71,16 @@ func (w errorTranslation) RoundTrip(req *http.Request) (*http.Response, error) {
 	}
 
 	// find all request bodies for the specified method/URL combination
-	urlString := req.URL.String()
-	var bodiesForMethodAndURL []string
-	for _, interaction := range w.ensureCassette().Interactions {
-		if urlString == interaction.URL && req.Method == interaction.Request.Method &&
-			req.Header.Get(COUNT_HEADER) == interaction.Request.Headers.Get(COUNT_HEADER) {
-			bodiesForMethodAndURL = append(bodiesForMethodAndURL, interaction.Request.Body)
-			break
-		}
-	}
+	matchingBodies := w.findMatchingBodies(req)
 
-	if len(bodiesForMethodAndURL) == 0 {
+	if len(matchingBodies) == 0 {
 		fmt.Printf("\n*** Cannot find go-vcr recording for request (no responses recorded for this method/URL): %s %s (attempt: %s)\n\n", req.Method, req.URL.String(), req.Header.Get(COUNT_HEADER))
 		return nil, originalErr
 	}
 
 	// locate the request body with the shortest diff from the sent body
 	shortestDiff := ""
-	for i, bodyString := range bodiesForMethodAndURL {
+	for i, bodyString := range matchingBodies {
 		diff := cmp.Diff(sentBodyString, bodyString)
 		if i == 0 || len(diff) < len(shortestDiff) {
 			shortestDiff = diff
@@ -97,4 +89,18 @@ func (w errorTranslation) RoundTrip(req *http.Request) (*http.Response, error) {
 
 	fmt.Printf("\n*** Cannot find go-vcr recording for request (body mismatch): %s %s\nShortest body diff: %s\n\n", req.Method, req.URL.String(), shortestDiff)
 	return nil, originalErr
+}
+
+// finds bodies for interactions where request method, URL, and COUNT_HEADER match
+func (w errorTranslation) findMatchingBodies(r *http.Request) []string {
+	urlString := r.URL.String()
+	var result []string
+	for _, interaction := range w.ensureCassette().Interactions {
+		if urlString == interaction.URL && r.Method == interaction.Request.Method &&
+			r.Header.Get(COUNT_HEADER) == interaction.Request.Headers.Get(COUNT_HEADER) {
+			result = append(result, interaction.Request.Body)
+		}
+	}
+
+	return result
 }
--- a/hack/generated/pkg/testcommon/error_translating_roundtripper.go
+++ b/hack/generated/pkg/testcommon/error_translating_roundtripper.go
@@ -39,7 +39,7 @@ type errorTranslation struct {
 	cassette *cassette.Cassette
 }
 
-func (w errorTranslation) ensure_cassette() *cassette.Cassette {
+func (w errorTranslation) ensureCassette() *cassette.Cassette {
 	if w.cassette == nil {
 		cassette, err := cassette.Load(w.cassetteName)
 		if err != nil {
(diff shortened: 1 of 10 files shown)


hack/generated/pkg/testcommon/error_translating_roundtripper.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/dnaeon/go-vcr/cassette"
+	"github.com/dnaeon/go-vcr/recorder"
+	"github.com/google/go-cmp/cmp"
+)
+
+// translateErrors wraps the given Recorder to handle any "Requested interaction not found"
+// and log better information about what the expected request was.
+//
+// By default the error will be returned to the controller which might ignore/retry it
+// and not log any useful information. So instead here we find  the recorded request with
```
matthchr, line 23:
```suggestion
// and not log any useful information. So instead here we find the recorded request with
```

hack/generated/pkg/testcommon/error_translating_roundtripper.go
```diff
@@ -0,0 +1,100 @@
...
+// the body that most closely matches what was sent and report the "expected" body.
+//
+// Ideally we would panic on this error but we don't have a good way to deal with the following
+// problem at the moment:
+// - during record the controller does GET (404), PUT, … GET (OK)
+// - during playback the controller does GET (which now returns OK), DELETE, PUT, …
```
matthchr, line 29: I thought that the ordering/count header existed specifically to make the first GET here 404?

hack/generated/pkg/testcommon/error_translating_roundtripper.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+type errorTranslation struct {
+	recorder     *recorder.Recorder
+	cassetteName string
+
+	cassette *cassette.Cassette
+}
+
+func (w errorTranslation) ensure_cassette() *cassette.Cassette {
```
matthchr, line 42: Is this implementing an interface or something?

If not, why not `ensureCassette`?

hack/generated/pkg/testcommon/test_context.go
```diff
@@ -134,7 +133,7 @@ func createRecorder(testName string, recordReplay bool) (autorest.Authorizer, st
 			return false
 		}
 
-		r.Body = ioutil.NopCloser(&b)
+		r.Body = io.NopCloser(&b)
```
Porges, line 136: 🔫

hack/generated/pkg/testcommon/error_translating_roundtripper.go
```diff
@@ -0,0 +1,100 @@
...
+		if bodyErr != nil {
+			// see invocation of SetMatcher in the createRecorder, which does this
+			panic("io.ReadAll(req.Body) failed, this should always succeed because req.Body has been replaced by a buffer")
+		}
+
+		sentBodyString = string(bodyBytes)
+	}
+
+	// find all request bodies for the specified method/URL combination
+	urlString := req.URL.String()
```
matthchr, line 74: minor: lines 74 to 87 kinda look like they could be factored out into a nicely named method for a bit more clarity?

matthchr (approved):
LGTM - left a few comments/questions though.

hack/generated/pkg/testcommon/error_translating_roundtripper.go
Porges: Right, let me get that restored and see if I can make this all work properly...

matthchr: Since you fixed the counting header, is this still true or does this comment need updating?

hack/generated/pkg/testcommon/test_context.go
```diff
@@ -172,18 +181,23 @@ func createRecorder(testName string, recordReplay bool) (autorest.Authorizer, st
...
-			request.Header.Get(COUNT_HEADER) == i.Headers.Get(COUNT_HEADER)
+		return nil
 	})
 
 	return authorizer, subscriptionID, r, nil
 }
 
+var dateMatcher *regexp.Regexp = regexp.MustCompile(`\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(.\d+)?Z`)
+
+// hideDates replaces all ISO8601 datetimes with a fixed value
```
matthchr, line 196: minor: Include the reason why we want this here in the comment?

hack/generated/pkg/testcommon/error_translating_roundtripper.go
Porges: There’s still one rogue DELETE happening and I'm not sure why. Will stick an issue to investigate later...

Porges: https://github.com/Azure/azure-service-operator/issues/1497

PR #1495, opened 2021-05-19
Create storage conversion helper functions

theunrepentantgeek:

Create storage variants of all resources and complex object types, injecting helper conversion functions for conversion back and forth. These helper functions will be used to build the actual conversions in a later PR.

The crux of operation is the StorageTypeFactory (see `storage_type_factory.go`) which uses some distinct utility classes to make the required changes.

Also includes:

* Some improvements to debugger output for various types
* Use of new abstractions PropertyContainer and FunctionContainer
* Improvements to field & parameter names, and other code gardening

[#headingintherightdirection](https://media.giphy.com/media/cAfaWIcWr7qus/giphy.gif)

**If applicable**:
- [x] this PR contains tests


hack/generator/pkg/codegen/storage/storage_type_factory.go
```diff
@@ -0,0 +1,208 @@
...
+	err := f.process()
+	if err != nil {
+		return nil, err
+	}
+
+	return f.types, nil
+}
+
+func (f *StorageTypeFactory) process() error {
+	err := f.pendingStorageConversion.Process(f.createStorageVariant)
```
babbageclunk, line 67: I'm not sure the queues are buying you much here - each one is only populated from one place, and they're processed one after the other, so they could have just been slices. That said, they don't really hurt either. I was just expecting the population logic to be more complicated.

babbageclunk (approved):
This looks great - I really like the split-out type- and property converter.

hack/generator/pkg/astmodel/storage_conversion_function.go
```diff
@@ -314,7 +331,14 @@ func (fn *StorageConversionFunction) createConversions(receiver TypeDefinition)
...
 			if err != nil {
 				// An error was returned; this can happen even if a conversion was created as well.
-				errs = append(errs, err)
+				if first {
+					klog.V(2).Infof("Generating conversion function %s() for %s", fn.name, receiver.name)
+					first = false
+				}
+
+				// err will include the details of which property, don't need to log here
+				klog.V(2).Infof("Error: %s", err)
```
matthchr, line 340: This code threw me a bit:

1. The variable `first`... is tracking the first conversion error so that we can log a special log for the first error? It's not clear to me why this is useful though?
2. Why log the `err` here but then throw it away and return a totally new error later rather than do something with `kerrors.NewAggregate` like you had before?
3. The `properties` variable could maybe be named `erroredProperties` or `failedProperties` or something, given it's actually just tracking failed properties?

matthchr: It feels like if the `err` context is going to be useful to users, then we should include it in the returned error (so that context is maintained for why the failure occurred and bubbles up to the point of failure)

hack/generator/pkg/astmodel/storage_conversion_function.go
```diff
@@ -325,7 +349,37 @@ func (fn *StorageConversionFunction) createConversions(receiver TypeDefinition)
...
+		if len(properties) == 1 {
+			label = "property"
+		}
+		return errors.Errorf("Errors with %d %s: %s", len(properties), label, strings.Join(properties, "; "))
+	}
+
+	return nil
+}
+
+// AsPropertyContainer converts a type into a property container
```
matthchr, line 363:
```suggestion
// asPropertyContainer converts a type into a property container
```

hack/generator/pkg/codegen/pipeline_create_storage_types.go
```diff
@@ -7,247 +7,67 @@ package codegen
...
+			// Create a factory for each group (aka service) and divvy up the types
+			factories := make(map[string]*storage.StorageTypeFactory)
+			for name, def := range types {
 
-				if astmodel.ARMFlag.IsOn(d.Type()) {
-					// Skip ARM definitions, we don't need to create storage variants of those
+				ref, ok := name.PackageReference.AsLocalPackage()
+				if !ok {
+					// Skip definitions from non-local packages
+					// (should never happen)
```
matthchr, line 32: Consider logging a warning in this case at the least (assuming it really should never happen?)

hack/generator/pkg/codegen/pipeline_create_storage_types.go
```diff
@@ -7,247 +7,67 @@ package codegen
...
 					errs = append(errs, err)
 					continue
 				}
 
-				finalDef := def.WithDescription(descriptionForStorageVariant(d))
-				storageTypes[finalDef.Name()] = finalDef
+				result.AddTypes(t)
 			}
 
 			if len(errs) > 0 {
```
matthchr, line 64: exceedingly minor: you can write this as:
```
err := kerrors.NewAggregate(errs)
if err != nil {
    return nil, err
}
```

It's exactly the same, _but_ ignores `nil` errors in the collection if you somehow had one and reads like 5% nicer?

Feel free to ignore this

hack/generator/pkg/codegen/pipeline_mark_storage_version.go
```diff
@@ -72,6 +72,15 @@ func groupResourcesByVersion(types astmodel.Types) (map[unversionedName][]astmod
 	result := make(map[unversionedName][]astmodel.TypeDefinition)
 
 	for _, def := range types {
+
+		// TODO: Remove this
```
matthchr, line 76: If you have a bug tracking it, reference the bug here?

If not, don't worry about it.

hack/generator/pkg/codegen/storage/storage_type_factory.go
```diff
@@ -0,0 +1,208 @@
...
+
+import (
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/astmodel"
+	"github.com/pkg/errors"
+	"k8s.io/klog/v2"
+)
+
+// StorageTypeFactory is used to create storage types for a specific service
+type StorageTypeFactory struct {
+	service                    string                                                  // Name of the group we're handling (used mostly for logging)
```
matthchr, line 16: copying my comment from previous review over: why not call this `group`, since that's what it's called elsewhere in the code? I feel like using `service` terminology here is adding additional terminology overhead. People already need to understand what a `group` and `version` are (to understand what storage conversions even do in the context of Kuberentes)

hack/generator/pkg/codegen/storage/hub_version_marker.go
```diff
@@ -0,0 +1,35 @@
...
+	result := &HubVersionMarker{}
+
+	result.visitor = astmodel.TypeVisitorBuilder{
+		VisitResourceType: result.markResourceAsStorageVersion,
+	}.Build()
+
+	return result
+}
+
+// MarkAsStorageVersion mark the supplied type definition as the storage version
```
matthchr, line 26:
```suggestion
// MarkAsStorageVersion marks the supplied type definition as the storage version
```

hack/generator/pkg/codegen/storage/property_converter.go
```diff
@@ -0,0 +1,146 @@
...
+package storage
+
+import (
+	"fmt"
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/astmodel"
+	"github.com/pkg/errors"
+)
+
+// PropertyConverter is used to convert the properties of object types as required for storage variants
+type PropertyConverter struct {
```
matthchr, line 15: minor, naming: Is this a `PropertyConverter` or a `PropertyConversionBuilder` or `Applier` or something?

It's not actually _doing_ conversions? It's just putting conversions onto types?

hack/generator/pkg/codegen/storage/property_converter.go
```diff
@@ -0,0 +1,146 @@
...
+// useBaseTypeForEnumerations replaces an enumeration with its underlying base type
+func (p *PropertyConverter) useBaseTypeForEnumerations(
+	tv *astmodel.TypeVisitor, et *astmodel.EnumType, ctx interface{}) (astmodel.Type, error) {
+	return tv.Visit(et.BaseType(), ctx)
+}
+
+// shortCircuitNamesOfSimpleTypes redirects TypeNames that reference resources or objects into our
+// storage namespace, and replaces TypeNames that point to simple types (enumerations or
+// primitives) with their underlying type.
+func (p *PropertyConverter) shortCircuitNamesOfSimpleTypes(
```
matthchr, line 76: Moving comment from last review over here:
Is the main reason that you need to do this for things like validated types where they had an alias before but since you're dropped validations there's no need for an alias now?

matthchr: expanding: It might be helpful to explain (in this methods comment maybe?) the purpose of this function in addition to what it does.

hack/generator/pkg/codegen/storage/property_converter.go
```diff
@@ -0,0 +1,146 @@
...
+			// Something went wrong, return the error
+			return nil, err
+		}
+		if prop != nil {
+			// We have the conversion we need, return it promptly
+			return prop, nil
+		}
+	}
+
+	return nil, fmt.Errorf("failed to find a conversion for property %v", property.PropertyName())
```
matthchr, line 56: minor: Useful to also include the property type in the error?

hack/generator/pkg/codegen/storage/storage_type_factory.go
```diff
@@ -0,0 +1,208 @@
...
+	types                      astmodel.Types                                          // All the types for this group
+	pendingStorageConversion   astmodel.TypeNameQueue                                  // Queue of types that need storage variants created for them
+	pendingConversionInjection astmodel.TypeNameQueue                                  // Queue of types that need conversion functions injected
+	pendingMarkAsHubVersion    astmodel.TypeNameQueue                                  // Queue of types that need to be flagged as the hub storage version
+	idFactory                  astmodel.IdentifierFactory                              // Factory for creating identifiers
```
matthchr, line 21: minor, style: Is the description in a tabbed-over string a standard golang style? (@babbageclunk knows maybe?)

It reads pretty cleanly here, but also isn't a "golang-style" comment as it's not starting with the name of the variable it refers to?

Mostly just wondering for my own knowledge.

hack/generator/pkg/codegen/storage/storage_type_factory.go
matthchr: Agree: it wasn't clear to me if the queue type was actually needed here or not (I think not?)

matthchr: Would the whole relationship between `types` and `pendingStorageConversion` be simpler if instead you just did away with `pendingStorageConversion` entirely, had `Add` add to `f.Types`, and then had a `result astmodel.Types` which was produced by `process`?

So:
```
    inputTypes astmodel.Types // input types to be processed by process() - this is filled out by Add
```

The set of types that are `pendingStorageConversion` then is just all `inputTypes`? So you loop over that and call `createStorageVariant` for each one, except now that you're not doing it using the queue-processing function you have a bit more flexibility in return type, so you could just have `createStorageVariant` return the new `TypeDefinition` it created, which you could store in a local `result astmodel.Types` variable inside `process()`?  This code ends up looking a lot more like a mini-pipeline stage (in that it's dealing with sets of types?)

I feel like the separation of the set of input types and the set of output types helps reason about things like `createStorageVariant` more easily, as that operates on `f.types`, but if `f.types` is simultanously being modified to include the result of added storage variants there's more confusion in "what exactly is in `f.types`" than there has to be?

hack/generator/pkg/codegen/storage/type_converter.go
```diff
@@ -0,0 +1,147 @@
...
+import (
+	"fmt"
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/astmodel"
+	"github.com/pkg/errors"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+)
+
+// TypeConverter is used to create a storage variant of an API type
+type TypeConverter struct {
+	visitor           astmodel.TypeVisitor // Visitor used to achieve the required modification
```
matthchr, line 17: minor: Same question about comments to the side rather than above (and prefixed with var name)

hack/generator/pkg/codegen/storage/storage_type_factory.go
```diff
@@ -0,0 +1,208 @@
...
+		pendingStorageConversion:   astmodel.MakeTypeNameQueue(),
+		pendingConversionInjection: astmodel.MakeTypeNameQueue(),
+		pendingMarkAsHubVersion:    astmodel.MakeTypeNameQueue(),
+		idFactory:                  idFactory,
+		conversionMap:              make(map[astmodel.PackageReference]astmodel.PackageReference),
+		functionInjector:           NewFunctionInjector(),
+		resourceHubMarker:          NewHubVersionMarker(),
+	}
+
+	result.typeConverter = NewTypeConverter(result.types)
```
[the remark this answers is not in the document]
theunrepentantgeek: Agreed. I've oscilated a bit on the style for initializing interrelated members, I think this works better.

hack/generator/pkg/codegen/storage/function_injector.go
```diff
@@ -0,0 +1,46 @@
...
+	return ot.WithFunction(fn), nil
+}
+
+// injectFunctionIntoResource takes the function provided as a context and includes it on the
+// provided resource type
+func (_ *FunctionInjector) injectFunctionIntoResource(
+	_ *astmodel.TypeVisitor, rt *astmodel.ResourceType, ctx interface{}) (astmodel.Type, error) {
+	fn := ctx.(astmodel.Function)
+	return rt.WithFunction(fn), nil
+}
```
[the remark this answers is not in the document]
theunrepentantgeek: They are simple functions, but giving them names makes the intentions super clear when the `TypeVisitorBuilder` is used.

hack/generator/pkg/codegen/storage/hub_version_marker.go
```diff
@@ -0,0 +1,35 @@
...
+func (m *HubVersionMarker) MarkAsStorageVersion(def astmodel.TypeDefinition) (astmodel.TypeDefinition, error) {
+	return m.visitor.VisitDefinition(def, nil)
+}
+
+// markResourceAsStorageVersion marks the supplied resource as the canonical hub (storage) version
+func (m *HubVersionMarker) markResourceAsStorageVersion(
+	_ *astmodel.TypeVisitor, rt *astmodel.ResourceType, _ interface{}) (astmodel.Type, error) {
+	return rt.MarkAsStorageVersion(), nil
+}
```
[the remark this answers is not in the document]
theunrepentantgeek: I think there's benefit in giving intention revealing names. #wontfix

hack/generator/pkg/astbuilder/builder.go
```diff
@@ -291,9 +291,9 @@ func ReturnNoError() dst.Stmt {
 //
 // errors.Wrap(err, <message>)
```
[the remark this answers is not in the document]
theunrepentantgeek: Added a comment.

hack/generator/pkg/astmodel/std_references.go
```diff
@@ -7,31 +7,31 @@ package astmodel
...
-	JsonReference    PackageReference = MakeExternalPackageReference("encoding/json")
-	ReflectReference PackageReference = MakeExternalPackageReference("reflect")
-	TestingReference PackageReference = MakeExternalPackageReference("testing")
+	FmtReference     = MakeExternalPackageReference("fmt")
+	JsonReference    = MakeExternalPackageReference("encoding/json")
+	ReflectReference = MakeExternalPackageReference("reflect")
+	TestingReference = MakeExternalPackageReference("testing")
 
 	// References to our Libraries
 	GenRuntimeReference PackageReference = MakeExternalPackageReference(genRuntimePathPrefix)
```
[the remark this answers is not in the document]
theunrepentantgeek: Done.

hack/generator/pkg/astmodel/storage_conversion_factories.go
```diff
@@ -123,10 +124,16 @@ func createTypeConversion(
 		}
 	}
 
+	var debugDescriptionOfDestination strings.Builder
```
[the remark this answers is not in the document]
theunrepentantgeek: Added a comment.

hack/generator/pkg/astmodel/storage_conversion_function.go
theunrepentantgeek: Changed.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
theunrepentantgeek: Added warning

hack/generator/pkg/codegen/pipeline_mark_storage_version.go
theunrepentantgeek: Works for me.

hack/generator/pkg/codegen/pipeline_create_storage_types.go
theunrepentantgeek: Works for me.

hack/generator/pkg/codegen/pipeline_mark_storage_version.go
theunrepentantgeek: Don't have one.

hack/generator/pkg/codegen/storage/storage_type_factory.go
theunrepentantgeek: Thanks, I thought I'd fixed all of those.

hack/generator/pkg/codegen/storage/hub_version_marker.go
theunrepentantgeek: Fixed.

hack/generator/pkg/codegen/storage/property_converter.go
theunrepentantgeek: It changes the types of properties passed to `ConvertProperty()` - it does the conversion.

hack/generator/pkg/codegen/storage/property_converter.go
theunrepentantgeek: Improve the comments

hack/generator/pkg/codegen/storage/property_converter.go
theunrepentantgeek: Included

hack/generator/pkg/codegen/storage/storage_type_factory.go
theunrepentantgeek: I'm sure I've seen this - can change it if not idiomatic

hack/generator/pkg/codegen/storage/type_converter.go
theunrepentantgeek: Moved.

hack/generator/pkg/astmodel/storage_conversion_function.go
theunrepentantgeek: Done.

matthchr (commented):
A few more minor suggestions - last main big topic is the queues + processing which I think @babbageclunk and I had comments on. (Not necessarily agitating for a specific change there but you hadn't responded to those comments yet so unsure if you're planning changes there or not).

hack/generator/pkg/astmodel/types.go
```diff
@@ -204,3 +203,22 @@ func (types Types) ResolveResourceStatusDefinition(
 	// preserve outer spec name
 	return resourceStatusDef.WithName(statusName), nil
 }
+
+// Process applies a func to transform all members of this set of type definitions, returning a new set of type
+// definitions containing the results of the transfomration, or possibly an error
```
matthchr, line 208:
```suggestion
// definitions containing the results of the transformation, or possibly an error
```

matthchr (approved):
LGTM!

theunrepentantgeek:
MC> A few more minor suggestions - last main big topic is the queues + processing which I think 
MC> @babbageclunk and I had comments on. 

Sorry, I missed responding to those comments - I've removed the use of queues completely.

hack/generator/pkg/astmodel/types.go
theunrepentantgeek: Missed this fix - included in my current branch instead.

PR #1499, opened 2021-05-20
Use correct DNS suffix for MySQLAADUser

### LINKED ISSUE #1496: Bug: MySQLAADUser doesn't use correct DB URL in national clouds
**Describe the bug**
It always assumes that the DB is in public Azure. See the comment [here](https://github.com/Azure/azure-service-operator/blob/master/pkg/resourcemanager/mysql/mysqlhelper.go#L37).

Ideally we wouldn't need to use the DB secret to get this value, we could get it from the Go SDK similarly to how we do for Azure SQL (see [this bug in the Azure Go SDK](https://github.com/Azure/azure-sdk-for-go/issues/13749)), but in the shorter term we can just rely on the DB secret.

matthchr:
Fix bug where the public DNS suffix was always used even in nonpublic environments. 

Closes #1496


matthchr:
I tested this manually and confirmed when the `AZURE_CLOUD_ENV` is set to `AzureUSGovernmentCloud` we correctly return the `mysql.database.usgovcloudapi.net` value now.

As with other AAD things this is a bit tricky to test in an automated fashion with our current test infrastructure.

Porges (approved):

babbageclunk (approved):

PR #1503, opened 2021-05-21
Update webhook patch files to produce valid v1 CRDs

babbageclunk:
I really should have tried deploying the resulting CRDs after the changes in #1428, rather than just seeing that the patches applied cleanly after the update. It turns out the `webhookClientConfig` field has been moved down into a parent `webhook` key, which also has a required list of acceptable conversion review versions.

See https://kubernetes.io/docs/reference/kubernetes-api/extend-resources/custom-resource-definition-v1/#CustomResourceDefinitionSpec

**If applicable**:
- [ ] this PR contains tests


matthchr (approved):
Great catch!

PR #1504, opened 2021-05-21
One pager design on improving object structure

matthchr:


theunrepentantgeek (commented):
I like this - and we certainly need a solution to the problem described. An approach like this is kinda self documenting for consumers (which is good), but it also further distances our objects from the ARM originals (which we thought was desirable early on). So I'm a bit conflicted.

matthchr:
@porges mentioned his preference for solving this problem is to lean the other way: Make the properties specific to the operator "special" and leave the Azure properties at the top:

**Spec**:
```yaml
apiVersion: microsoft.storage.infra.azure.com/v1alpha1api20190401
kind: StorageAccount
metadata:
  name: samplekubestorage
  namespace: default
spec:
  azureName: mykubestorage
  location: westcentralus
  kind: BlobStorage
  sku:
    name: Standard_LRS
  accessTier: Hot
  tags:
    tag1: tag1
    tag2: tag2
  forOperator:
    secretConfiguration: 
       ...
    owner:
      name: k8sinfra-sample-rg  
```

**Status**:
```yaml
status:
  id: /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Storage/storageAccounts/mykubestorage
  location: westcentralus
  kind: BlobStorage
  sku:
    name: Standard_LRS
  accessTier: Hot
  tags:
    tag1: tag1
    tag2: tag2
  operatorStatus:
    deploymentId: /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/Microsoft.Deployments/deployments/1234
    goalSeekingState: GoalMet
```

This has the advantage of making the properties the customer cares most about less nested in the case of `spec`, although I'm not sure that the fields the customer cares the most about are less nested in the `status` case, as they likely care about the state of the resource in the operator more than the specific fields which are set in Azure (since presumably they're going to match the `spec` pretty much)

babbageclunk:
> This has the advantage of making the properties the customer cares most about less nested in the case of `spec`, although I'm not sure that the fields the customer cares the most about are less nested in the `status` case, as they likely care about the state of the resource in the operator more than the specific fields which are set in Azure (since presumably they're going to match the `spec` pretty much)

Do we have control of how the fields are ordered out of kubectl? If the `operatorStatus` fields were first in the status output that would give them enough prominence, wouldn't it?

matthchr:
> Do we have control of how the fields are ordered out of kubectl? If the operatorStatus fields were first in the status output that would give them enough prominence, wouldn't it?

In the `-o yaml` ordering I don't think we do, but in the default display I think we do via the `additionalPrinterColumns` stuff

docs/design/clarifying-object-strucutre.md
```diff
@@ -0,0 +1,111 @@
...
+result of a GET on the resource) and properties that we are presenting. For example, we may want to have a status field
+for `deploymentId` documenting the deployment ID used to create the resource. If we do that there's no easy way for the
+customer to understand that field is provided by ASO and not by Storage.
+
+## Proposal
+We introduce an additional level of hierarchy specifically to clarify what is coming or going directly from or to Azure.
+This is similar to what Crossplane does (see their 
+[MySQLServer](https://github.com/crossplane/provider-azure/blob/master/examples/database/mysqlserver.yaml)).
+
+On the `spec` we could call this something like `forAzure` or `azureSpec`, and on the `status` `fromAzure` or `azureStatus`.
```
theunrepentantgeek, line 63: I think this has come up in discussion elsewhere, but worth capturing here - this pushes all the stuff that customers are interested in _down_ a level. Maybe we should instead push ASO stuff down into a container?

matthchr: Yes, I will update the spec to reflect this

docs/design/clarifying-object-structure.md
```diff
@@ -0,0 +1,106 @@
...
+A: No, because `state` and `deploymentId` are currently resource annotations, not fields in the `status`, and on the `spec` the only
+   fields which might belong in `operatorSpec` are `owner` and `azureName` which as per the above FAQ are staying at the top level.
+
+# Open questions
+
+1. What should we call the property which we push things down "into"? Here are some options:
+  * `operator` (for both)
+  * `operatorData` (for both)
+  * `forOperator` / `fromOperator`
+  * `operatorSpec` / `operatorStatus`
```
babbageclunk, line 105: I prefer this option from the ones listed.

babbageclunk (approved):

docs/design/clarifying-object-structure.md
matthchr: I think I also prefer this

PR #1505, opened 2021-05-24
Fix operator bundle generation issues found testing upgrades

babbageclunk:
Testing upgrading an operator bundle from 0.37.0 to 1.0.22275 with a custom index and `CatalogSource` revealed a few problems in the bundle generation.

* The image reference was `mcr.microsoft.com/k8s/azureserviceoperator@1.0.22275` when it should have been `mcr.microsoft.com/k8s/azureserviceoperator:1.0.22275`
* The `spec.webhookdefinitions` items in the [`ClusterServiceVersion`](https://docs.openshift.com/container-platform/4.5/rest_api/operatorhub_apis/clusterserviceversion-operators-coreos-com-v1alpha1.html#specification) need to have `admissionReviewVersions` specified. This is a bit weird since the webhook configurations we generate at the moment are v1beta1 (because of controller-runtime limitations), but patching version fields in anyway produces the desired output.
* The CRD webhook config needed to be updated to match v1 (done in #1503).

With these fixed the bundle upgrades successfully, phew!

I'll follow this up with a document on how to test bundle upgrading (with a view to turning that into an automated test) and create a community-operators PR for the bundle.


Makefile
```diff
@@ -356,9 +356,8 @@ generate-operator-bundle: manifests
 	# This is only needed until CRD conversion support is released in OpenShift 4.6.x/Operator Lifecycle Manager 0.16.x
 	scripts/add-openshift-cert-handling.sh
 	# Inject the container reference into the bundle.
-	scripts/inject-container-reference.sh "$(PUBLIC_REPO)@$(LATEST_TAG)"
+	scripts/inject-container-reference.sh "$(PUBLIC_REPO):$(LATEST_TAG)"
```
matthchr, line 359: Is there a markdown document describing the usage of this build target (i.e. what to do once you've run it?) If so might be worth referencing that markdown via a comment in the makefile.

config/operator-bundle/patches/admissionreviewversions_in_webhook.yaml
```diff
@@ -0,0 +1,17 @@
+# This adds an admissionReviewVersions field to the validation webhook
+# items.  It's not exactly valid, since it's a v1beta1
```
matthchr, line 2: Should we file a bug for the operator bundle folks?

matthchr (approved):

Makefile
babbageclunk: Not yet - that's the doc I mention in the description.

config/operator-bundle/patches/admissionreviewversions_in_webhook.yaml
babbageclunk: Good call - I'm doing that now.

babbageclunk: Actually, while I was writing up a bug for them I realised there was nothing stopping us from switching to v1 ValidationWebhookConfiguration, which would make this less weird. So I did that instead, although it required patching the file to remove v1 from admissionReviewVersions (since those are not supported by controller-runtime).

babbageclunk: I realised the patches aren't applied when running the tests, so any test relying on a webhook failed. Unfortunately there's no option in the kubebuilder webhook directive to manage the versions directly. 
Actually I could just do it directly in the manifests make target like the `preserveUnknownFields` fix... I'll land this one as it is and do that subsequently.

PR #1507, opened 2021-05-24
Make validator and defaulter more extensible

### LINKED ISSUE #1430: Defaulting and Validating handlers should be extensible
Each resource can only implement `admissions.Defaulter` and `admissions.Validator` once. As long as all of the validation/defaulting we want to do is understood at code-generation time that's fine, but once it's not the implementations of these interfaces generated by the code generator will block any handcrafted extensions.

We probably want to support some sort of `BiggerValidator` and `BiggerDefaulter` interfaces:
```
type ValidationFunc func() error

type BiggerValidator interface {
    validateCreates() []ValidationFunc
    validateUpdates(old runtime.Object)  []ValidationFunc
    validateDeletes()  []ValidationFunc
}
```

Then we could change the implementation of the code generated for `Validate` and `Default` to check if the type implemets the `BiggerValidator`/`BiggerDefaulter` types and call those validations/defaults as well?

matthchr:
As the code generator gets fully fleshed out, we're running into cases where we need to perform special validation or defaulting logic on all resources. Additionally, the ability to manually add to validations/defaults for certain resources will be needed.

See https://github.com/Azure/azure-service-operator/blob/master/docs/design/defaulter-validator.md for the design of this feature.

Closes #1430


matthchr:
I know this looks like a pretty large PR, but ~2300 lines are golden files.

hack/generator/pkg/astmodel/function.go
```diff
@@ -27,3 +27,85 @@ type Function interface {
...
+	requiredPackages *PackageReferenceSet
+}
+
+// Name returns the unique name of this function
+// (You can't have two functions with the same name on the same object or resource)
+func (k *objectFunction) Name() string {
+	return k.name
+}
+
+func (k *objectFunction) RequiredPackageReferences() *PackageReferenceSet {
```
matthchr, line 51: Missing comments

Porges (approved):

Porges:
Closes #1430?

matthchr:
So it does! Added to the description

PR #1511, opened 2021-05-25
Modify Swagger pipeline stage to allow for more augmentations

Porges:
This allows us to write multiple augmenters for information from Swagger and they will all be applied to any corresponding Resource types.

----

It isn’t included in this PR but here is an example of what an `augmenter` can look like, using `TypeMerger`. The pattern of matching up `TypeName`s structurally is probably going to be common with pulling information from Swagger:

```golang
// flattenAugmenter copies across the "flatten" property from Swagger
func flattenAugmenter(allTypes astmodel.Types) augmenter {
	return func(main astmodel.Type, swagger astmodel.Type) (astmodel.Type, error) {
		var merger = astmodel.NewTypeMerger(func(ctx interface{}, left, right astmodel.Type) (astmodel.Type, error) {
			// return left (= main) if the two sides are not ObjectTypes
			return left, nil
		})

		merger.Add(func(main, swagger astmodel.TypeName) (astmodel.Type, error) {
			// when we merge two typenames we always return the main type name
			// however, *as a side effect*, we augment this named type with the information
			// from the corresponding swagger type.

			// this allows us to handle cases where names differ greatly from JSON schema to Swagger,
			// we rely on the structure of the types to tell us which types are the same despite having different names

			// TODO: do we need to resolve these until we hit a non-TypeName?
			newType, err := merger.Merge(allTypes[main].Type(), allTypes[swagger].Type())
			if err != nil {
				return nil, err
			}

			allTypes[main] = allTypes[main].WithType(newType)

			return main, nil
		})

		merger.Add(func(main, swagger *astmodel.ObjectType) (astmodel.Type, error) {
			props := main.Properties()
			for ix, mainProp := range props {
				if swaggerProp, ok := swagger.Property(mainProp.PropertyName()); ok {
					// first copy over flatten property
					mainProp = mainProp.SetFlatten(swaggerProp.Flatten())

					// now recursively merge property types
					newType, err := merger.Merge(mainProp.PropertyType(), swaggerProp.PropertyType())
					if err != nil {
						return nil, err
					}

					props[ix] = mainProp.WithType(newType)
				}
			}

			return main.WithProperties(props...), nil
		})

		return merger.Merge(main, swagger)
	}
}
```


hack/generator/pkg/codegen/pipeline_swagger_augment.go
```diff
@@ -0,0 +1,117 @@
...
+			}
+
+			// put all types into a new set
+			newTypes := make(astmodel.Types)
+			for _, t := range types {
+				newTypes.Add(t)
+			}
+
+			// build the augmenter we will use
+			augmenter := fuseAugmenters( /* put your augmenters here */ )
```
Porges, line 62: At the moment this is a noöp due to having no augmenters, but I hope you can get the idea 🙂

babbageclunk (approved):
Looks good!

hack/generator/pkg/codegen/pipeline_swagger_augment.go
```diff
@@ -0,0 +1,117 @@
...
+			klog.V(1).Infof("Loaded Swagger data (%v resources, %v other types)", len(swaggerTypes.resources), len(swaggerTypes.otherTypes))
+
+			statusTypes, err := generateStatusTypes(swaggerTypes)
+			if err != nil {
+				return nil, err
+			}
+
+			// put all types into a new set
+			newTypes := make(astmodel.Types)
+			for _, t := range types {
```
matthchr, line 57: minor: Would it be clearer to instead below loop over `types` and just have there be an `else` clause that adds the type to the set?

So something like:

```
for typeName, typeDef := range newTypes {
    if resource, ok := typeDef.Type().(*astmodel.ResourceType); ok {
        ...
        // Don't need to do newTypes[typeName] = ... overwrite here anymore, instead can just add it to newTypes which I think plays nicer with types.Add assertions and more closely follows pattern of other pipeline stages).
    } else {
        newTypes.Add(typeDef)
    }
}
```

hack/generator/pkg/codegen/pipeline_swagger_augment_status.go
```diff
@@ -17,96 +17,35 @@ import (
 	"sync"
 
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/klog/v2"
```
matthchr, line 20: This file is technically not a pipeline stage anymore -- wondering if it should be renamed and/or moved someplace else? (A subdirectory of codegen, or maybe into astmodel?)

matthchr (approved):

hack/generator/pkg/codegen/pipeline_swagger_augment.go
Porges: it would but it's done like this so that the augmenters can update other types aside from the one that is passed in, see, e.g. `flattenAugmenter` in the PR description. 

When that augmenter hits an object type on each side of spec/status, it will find any properties with the same name and augment them as well, so that e.g. if one side has `subThing SpecItem` and the other `subThing StatusItem` then `SpecItem` will get augmented with the `StatusItem` despite them having entirely different names in the JSON/Swagger.

matthchr: Ah I missed that nuance in the sample code you had posted. I'll go ahead and resolve this.

Porges: @matthchr yeah, it isn’t obvious because there’s no distinguishing between immutable and mutable uses of `Types`...

Porges: @matthchr actually I have changed this. I decided "load status info from Swagger" and "augment Spec with Status info" should be two separate stages. Have changed back to the "clearer" way.

Porges:
Subsumed by #1548

PR #1524, opened 2021-05-28
Fix small ARM conversion bug

matthchr:
If there are no properties to actually perform conversion on, just return nil.


matthchr:
Not sure if linting on the golden files would've caught this, but another situation where the golden files actually had a compiler error in it and I just didn't notice it initially.

Porges (approved):

PR #1526, opened 2021-05-28
User-Agent version

### LINKED ISSUE #1525: UserAgent of v2 controller should report version
A description of how things are today.

At the moment we have:

https://github.com/Azure/azure-service-operator/blob/662f514774ce413e45308dec7b25400a3b4fe0e0/hack/generated/pkg/armclient/raw_client.go#L26

This should also report the version so we can track bugs and usage based on requests.

**Describe the improvement**

It should look like: `aso-controller/x.y.z`.

Porges:
Closes #1525

Adds version to the User-Agent string so it will look like `aso-controller/v2.0.0-alpha`.


Taskfile.yml
```diff
@@ -5,11 +5,16 @@ output: prefixed
...
   GENERATOR_APP: aso-gen
   GENERATOR_ROOT: ./hack/generator/
+  GENERATOR_PACKAGE: github.com/Azure/azure-service-operator/hack/generator
 
   CONTROLLER_APP: aso-controller
   CONTROLLER_ROOT: ./hack/generated/
+  CONTROLLER_PACKAGE: github.com/Azure/azure-service-operator/hack/generated
 
-  CONTROLLER_DOCKER_IMAGE: azure-service-operator:v2.0.0-alpha
+  # Once the v2 controller is the main controller we can drive this from git tags
```
matthchr, line 14: minor: Is this a TODO?

Taskfile.yml
```diff
@@ -5,11 +5,16 @@ output: prefixed
...
+  # via e.g. `git describe --tags --dirty` which would give better info for dev builds
+  VERSION: v2.0.0-alpha
```
matthchr, line 16: `VERSION` and `SHORT_VERSION` are also set in the `generator:build` step already... is that a problem?

matthchr (approved):

Taskfile.yml
Porges: Marked as such

Taskfile.yml
Porges: Removed `VERSION` which `generator:build` didn’t use, and added a note to this comment to update `generator` once this is the primary version for the code. At the moment we really just want it in the controller’s `User-Agent` for metrics purposes.

babbageclunk (approved):

PR #1530, opened 2021-05-28
Add v2 PublicIPAddress resource and CRUD test

matthchr:
Adds PublicIPAddress, and fixes existing ResourceGroup test to use Patch rather than Update.

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/hack/generated/controllers/crd_networking_publicip_test.go
+++ b/hack/generated/controllers/crd_networking_publicip_test.go
@@ -1,0 +1,86 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+package controllers_test
+
+import (
+	"context"
+	"testing"
+
+	. "github.com/onsi/gomega"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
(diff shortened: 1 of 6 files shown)


hack/generated/controllers/crd_networking_publicip_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+	objectKey, err := client.ObjectKeyFromObject(publicIPAddress)
+	g.Expect(err).ToNot(HaveOccurred())
+
+	// ensure state got updated in Azure
+	g.Eventually(func() *int {
+		updatedIP := &network.PublicIPAddresses{}
+		g.Expect(testContext.KubeClient.Get(ctx, objectKey, updatedIP)).To(Succeed())
+		return updatedIP.Status.Properties.IdleTimeoutInMinutes
+	}, remainingTime(t)).Should(Equal(&idleTimeoutInMinutes))
```
Porges, line 74: this works despite being a pointer comparison‽

Porges: so it ends up using `reflect.DeepEqual` which will chase pointers…

Porges (approved):

PR #1532, opened 2021-05-30
Trial test helpers

Porges:

Makes tests less boilerplatey.

WIP, please provide feedback on how good/bad this is.

gif

**If applicable**:
- [ ] this PR contains tests


hack/generated/controllers/crd_cosmosdb_test.go
```diff
@@ -18,13 +17,9 @@ import (
 func Test_CosmosDB_CRUD(t *testing.T) {
 	t.Parallel()
 
-	g := NewGomegaWithT(t)
-	ctx := context.Background()
-	testContext, err := testContext.ForTest(t)
-	g.Expect(err).ToNot(HaveOccurred())
+	testContext := testContext.ForTest(t)
```
babbageclunk, line 20: I think you could go the whole hog and call the local `tc`. (But now I've got the Top Cat theme tune stuck in my head.)

babbageclunk (approved):
The updated tests look heaps nicer!

hack/generated/controllers/crd_cosmosdb_test.go
Porges: that’s a good suggestion because the `Expect` calls are now very clunky

hack/generated/pkg/testcommon/kube_test_context.go
```diff
@@ -115,6 +125,9 @@ type KubePerTestContext struct {
 	KubeClient client.Client
 	Ensure     *Ensure
 	Match      *KubeMatcher
+
```
matthchr, line 128: minor: remove this newline maybe?

hack/generated/pkg/testcommon/kube_test_context_helpers.go
```diff
@@ -0,0 +1,102 @@
...
+	return ktc.G.Eventually(actual, ktc.RemainingTime())
+}
+
+func (ktc *KubePerTestContext) CreateAndWait(obj runtime.Object) {
+	ktc.G.Expect(ktc.KubeClient.Create(ktc.Ctx, obj)).To(gomega.Succeed())
+	ktc.G.Eventually(obj, ktc.RemainingTime()).Should(ktc.Match.BeProvisioned())
+}
+
+func (ktc *KubePerTestContext) CreateNewTestResourceGroupAndWait() *v1alpha1api20200601.ResourceGroup {
+	rg, err := ktc.CreateNewTestResourceGroup(WaitForCreation)
```
matthchr, line 63: Might be worth commenting here that the reason this exists separate from `CreateAndWait` is that the `RG` one also registers a cleanup function to ensure RG deletion even if the test fails or times out

matthchr: I was originally wondering why bother with this since `ResourceGroup` is a `runtime.Object` and could use the above helper too

hack/generated/pkg/testcommon/kube_test_context_helpers.go
```diff
@@ -0,0 +1,102 @@
...
+	ktc.G.Expect(ktc.KubeClient.Delete(ktc.Ctx, obj)).To(gomega.Succeed())
+	ktc.G.Eventually(obj, ktc.RemainingTime()).Should(ktc.Match.BeDeleted())
+}
+
+type Subtest struct {
+	Name string
+	Test func(testContext KubePerTestContext)
+}
+
+func (ktc *KubePerTestContext) RunParallelSubtests(tests ...Subtest) {
```
matthchr, line 86: Comment?

hack/generated/pkg/testcommon/kube_test_context_helpers.go
matthchr, line 63: Minor: I noticed in the body of this method that we do `log.Printf`, when I think it should actually be `tc.T.Logf()` - if you agree maybe worth fixing in this PR?

hack/generated/controllers/crd_cosmosdb_test.go
```diff
@@ -6,7 +6,6 @@ Licensed under the MIT license.
 package controllers_test
 
 import (
-	"context"
 	"testing"
 
 	. "github.com/onsi/gomega"
```
matthchr, line 11: Is the expectation that these tests don't need to use any of the built-in gomega assertions because we've got wrappers on `testContext`? And if they _do_ need an assertion then it would be via `testContext.G`?

If so, maybe remove this import from all the tests? Hopefully the fact this isn't imported at all will help guide people to using the right assertions

hack/generated/pkg/testcommon/kube_test_context_helpers.go
```diff
@@ -0,0 +1,102 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/Azure/azure-service-operator/hack/generated/_apis/microsoft.resources/v1alpha1api20200601"
+	"github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+)
```
matthchr, line 16: Rather than having an "extensions" file here, would it make more sense to just split the `KubePerTestContext` definition out into its own file and then move these things there with it?

It feels sorta awkward splitting these methods off from the methods defined in the file with `KubePerTestContext`?

hack/generated/pkg/testcommon/kube_test_context_helpers.go
```diff
@@ -0,0 +1,102 @@
...
+	ktc.G.Eventually(obj, ktc.RemainingTime()).Should(ktc.Match.BeProvisioned())
+}
+
+func (ktc *KubePerTestContext) CreateNewTestResourceGroupAndWait() *v1alpha1api20200601.ResourceGroup {
+	rg, err := ktc.CreateNewTestResourceGroup(WaitForCreation)
+	ktc.Expect(err).ToNot(gomega.HaveOccurred())
+	return rg
+}
+
+func (ktc *KubePerTestContext) Get(key types.NamespacedName, obj runtime.Object) {
```
matthchr, line 68: I'm a bit worried that this method (as well as `Update`, etc) are super generically named on a catch-all object. I think for now there's not too much issue because the total method count is still pretty low but I could pretty easily see some confusion arising in test authors... "Get what"?

Not sure we need to do anything about it now though.

matthchr (approved):
Looks good to me -- left a few organizational comments

hack/generated/controllers/crd_cosmosdb_test.go
Porges: It’s only `Expect`/`Eventually` that come off `testContext`, you still need `gomega` imported for all the actual assertions.

hack/generated/pkg/testcommon/kube_test_context_helpers.go
Porges: Updated them to have a `Resource` suffix, which helps a little 🙂

hack/generated/pkg/testcommon/kube_test_context_helpers.go
Porges: Done.

hack/generated/pkg/testcommon/kube_test_context_helpers.go
Porges: Done.

Porges: really only one test (the resource group test) will use CreateAndWait for an RG. Perhaps this could have a better name…

hack/generated/controllers/crd_cosmosdb_test.go
matthchr: Ahh, I see.

Issue #1531, opened 2021-05-29
Bug: conversion webhook service not found


ramjigit:
**Describe the bug**
Error from server: error when creating "azuredb.yaml": conversion webhook for azure.microsoft.com/v1alpha1, Kind=AzureSqlDatabase failed: Post "https://webhook-service.system.svc:443/convert?timeout=30s": service "webhook-service" not found

**To Reproduce**
<Fill in the steps>

**Expected behavior**

**Screenshots**

**Additional context**

ramjigit:
Not able to create Azuresqlserver or DB ,getting webhook issue

babbageclunk:
Hi @ramjigit, digging into it it looks to be a problem with our recent release - the webhook service and namespace should have been rewritten by `kustomize` to have the prefix `azureoperator-`. We're working on a fix at the moment.

cpflaume:
I have a similar issue: When I run 'make manifests' or 'make install', the path for the webhook is not rewritten properly:

What is looks like
```yaml
spec:
  conversion:
    strategy: Webhook
    webhook:
      clientConfig:
        caBundle: xxx
        service:
          name: webhook-service
          namespace: system
```

What it did look like in the past or should look like

```yaml
spec:
  conversion:
    strategy: Webhook
    webhook:
      clientConfig:
        caBundle: xxx
        service:
          name: azureoperator-webhook-service
          namespace: azureoperator-system
          path: /convert
          port: 443
```
This currently blocks me to contribute to this project...

cpflaume:
I think this happend with the apiextensions version change vom v1beta1 to v1.

Possible fix: 
https://github.com/Azure/azure-service-operator/blob/034cf475628ecc0990a733c5fb3a8f2ea65a5a8f/config/crd/kustomizeconfig.yaml#L8

```yaml
 path: spec/conversion/webhook/clientConfig/service/name
```

Same would apply to the namespace.

babbageclunk:
Yes, this is absolutely right - thanks for the fix!

ramjigit:
I tested it now ,I still face the same issue

babbageclunk:
Hmm - can you please run:
```sh
make manifests
kustomize build config/default | grep "conversion:" -B 8 -A 6
```

And show the output of the last command? Also the output of:
```sh
kubectl get crd azuresqlservers.azure.microsoft.com -o jsonpath="{.spec.conversion.webhook.clientConfig.service}"
```

And:
```sh
[...]

PR #1535, opened 2021-05-31
fix path so kustomize adds prefixes to the webhook service

cpflaume:
The path seems to have changed with the upgrade of the CRD to version v1.

refs #1531

Not sure if it fixes all related issues, but worked when generating the CRD yamls locally.


babbageclunk (approved):
I tested this and it fixes the problem. Thanks!

ramjigit:
Please merge this with master

PR #1538, opened 2021-06-01
Specification for Preservation of ARM API Versioning

### LINKED ISSUE #1468: Handle apiVersion property specially
The desired state is as follows (I think):

1. `APIVersion` exists only on the storage type. We need it here because we need to remember what version the customer created the CRD with so that we can use that version to submit it to ARM.
2. `APIVersion` doesn't exist on customer visible types: It's already set in `metav1.TypeInfo` on each resource, so we don't need it again inside each resource, it's duplicate information appearing at what looks like a weird level in the CRD.
3. `APIVersion` _is_ included in the JSON payload to ARM, because it's required to inform ARM about the specific APIVersion requested to create the resource.

theunrepentantgeek:

Initial specification for how we will preserve the correct API version and ensure we use it when interacting with ARM.

Resolves #1468

Direct link for viewing with embedded diagrams:
https://github.com/Azure/azure-service-operator/blob/spec/api-versioning/docs/design/api-versions.md

gif

**If applicable**:


docs/design/api-versions.md
```diff
@@ -0,0 +1,32 @@
...
+
+If we have a valid **v1** `Person`, trying to submit that through the **v2** ARM API will fail because it's missing a postal address. 
+
+## Proposed Solution
+
+We need to preserve the original API Version of each resource, and use that to create an appropriate resource for ARM.
+
+### API Preservation
+
+When generating the storage variants of each resource, we'll inject a new `ApiVersion` property of type **string**.
```
babbageclunk, line 21: Does this need updating? In the whiteboard discussion we had you were talking about a GVK property instead?

theunrepentantgeek: Yes, I've now updated the spec.

babbageclunk (approved):

babbageclunk (approved):
Nice - like the diagrams.

PR #1541, opened 2021-06-01
Add postrequisite support to PipelineStage

theunrepentantgeek:

We already have prerequisite support where a pipeline stage can declare that another stage must run before it does. This adds the mirror capabiltiy for a pipeline stage to declare that another stage must run after this one.

This is useful for pipeline stages that might be dropped out in different editions of the pipeline, as it allows that stage to declare all of the relationships needed.

gif

**If applicable**:
- [x] this PR contains tests


Porges (approved):

hack/generator/pkg/codegen/pipeline_stage.go
```diff
@@ -59,6 +62,24 @@ func (stage PipelineStage) RequiresPrerequisiteStages(prerequisites ...string) P
 	return stage
 }
 
+// RequiresPostrequisiteStages declares which stages must be executed after this one has completed
+// This is not completely isomorphic with RequiresPrerequisiteStages as are supporting stages that are sometimes omitted
```
matthchr, line 66: Language here: `as are supporting stages`?

As there are?

matthchr (approved):

hack/generator/pkg/codegen/pipeline_stage.go
theunrepentantgeek: Revised the comment.

PR #1543, opened 2021-06-01
Add v2 LoadBalancer and test

matthchr:
Adds support for networking.LoadBalancer

Includes a helper method for generating references between resources.

**If applicable**:
- [x] this PR contains tests


hack/generated/controllers/crd_networking_loadbalancer_test.go
```diff
@@ -0,0 +1,124 @@
...
+				},
+				// TODO: The below stuff isn't really necessary for LB CRUD but is required for VMSS...
+				InboundNatPools: []network.InboundNatPool{
+					{
+						Name: "MyFancyNatPool",
+						Properties: &network.InboundNatPoolPropertiesFormat{
+							FrontendIPConfiguration: network.SubResource{
+								Reference: genruntime.ResourceReference{
+									// TODO: This is still really awkward
+									ARMID: testContext.MakeARMId(rg.Name, "Microsoft.Network", "loadBalancers", lbName, "frontendIPConfigurations", lbFrontendName),
```
matthchr, line 87: This is _really_ awkward. We may want to find a way to work around this with an admissions webhook or custom hook in this controller...

hack/generated/pkg/testcommon/test_context.go
```diff
@@ -210,3 +213,13 @@ func (tc PerTestContext) NewTestResourceGroup() *resources.ResourceGroup {
...
 }
+
+func (tc PerTestContext) MakeARMId(resourceGroup string, provider string, params ...string) string {
+	if len(params)%2 != 0 {
+		panic("ARM Id params must come in resourceKind/name pairs")
+	}
+
+	suffix := strings.Join(params, "/")
+
+	return fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/%s/%s", tc.AzureSubscription, resourceGroup, provider, suffix)
```
theunrepentantgeek, line 224: This will have a trailing `/` if `params` is empty - is that correct?

theunrepentantgeek (commented):
One quick question.

hack/generated/pkg/testcommon/test_context.go
matthchr: No, but really `params` min length should be 2. I'll add a check for that.

hack/generated/controllers/crd_networking_loadbalancer_test.go
Porges: @matthchr we should be able to build this automatically from the Swagger specs

matthchr: @porges can you expand? This includes a reference to a different resource that:

1. Isn't _really_ a resource (it's a resource created by a property above).
2. Doesn't exist yet and whose name is variable (based on the spec of this resource)

Porges: @matthchr, not that it will be automatic, but e.g. the Swagger spec has:

```
/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/loadBalancers/{loadBalancerName}/frontendIPConfigurations/{frontendIPConfigurationName}
```

However, since as you say it’s not _really_ a resource we won’t have anything to attach it to…

Porges (approved):

hack/generated/controllers/crd_networking_loadbalancer_test.go
```diff
@@ -0,0 +1,113 @@
...
+				PublicIPAllocationMethod: network.PublicIPAddressPropertiesFormatPublicIPAllocationMethodStatic,
+			},
+		},
+	}
+
+	tc.CreateResourceAndWait(publicIPAddress)
+	defer tc.DeleteResourceAndWait(publicIPAddress)
+
+	// It should be created in Kubernetes
+	g.Expect(publicIPAddress.Status.Id).ToNot(BeNil())
```
Porges, line 53: we should probably remove this check from this test

matthchr: Fixed

PR #1545, opened 2021-06-02
Allow property assignments to pull values from functions

theunrepentantgeek:

Allows generated conversions to read a value from a suitably named source function for writing into a field on the destination type. Having this will allow resolution of #1468 by having a function on an API variant (e.g. `v1alpha1api20210601.Person`) that is used to populate a field on the associated storage variant (e.g. `v1alpha1api20210601storage.Person`).

Implementing a generic solution was cleaner than trying to do lots of semi-nasty special casing. Introducing `ValueFunction` avoided some wide spread changes to existing implementations of `Function`.

There's also a little tidy up around the golden tests as they were previously using private (lowercase) field names when our generated types actually use public (CapitalCase) naming.

gif

**If applicable**:
- [x] this PR contains tests


hack/generator/pkg/astmodel/one_of_json_marshal_function.go
```diff
@@ -122,3 +122,8 @@ func (f *OneOfJSONMarshalFunction) AsFunc(
 func (f *OneOfJSONMarshalFunction) RequiredPackageReferences() *PackageReferenceSet {
 	return NewPackageReferenceSet(MakeExternalPackageReference("encoding/json"))
 }
+
+// ReturnTypes returns nil as we can't represent byte[] in our type system
+func (f *OneOfJSONMarshalFunction) ReturnTypes() []Type {
+	return nil
```
babbageclunk, line 128: Presumably this isn't a problem because ReturnTypes is just used for References?

hack/generator/pkg/conversions/property_assignment_function.go
```diff
@@ -295,33 +324,60 @@ func (fn *PropertyAssignmentFunction) createConversions(receiver astmodel.TypeDe
...
+			// Generate a conversion to assign the property from the function
+			conv, err := fn.createFunctionConversion(sourceFunction, destinationProperty)
 			if err != nil {
 				// An error was returned, we abort creating conversions for this object
-				return errors.Wrapf(err, "creating conversion for property %q of %q", receiverProperty.PropertyName(), receiver.Name())
+				return errors.Wrapf(
+					err,
+					"creating conversion setting property %q from function %s()",
+					destinationName,
+					sourceFunction.Name())
```
babbageclunk, line 377: These two chunks (and the `createPropertyConversion` and `createFunctionConversion` methods below) are very similar. Would it be feasible to make wrappers for sourceProperty and sourceFunction to make them expose the same interface (where one reads from a selector and the other reads from a callexpr)  and then use that in a single `createPropertyConversion` method?

babbageclunk (approved):
Looks good! Maybe the possibility of reducing some code duplication in the handling of functions vs properties? But not a deal breaker.

hack/generator/pkg/astmodel/one_of_json_marshal_function.go
theunrepentantgeek: I've removed this method - have added a new interface `ValueFunction` which is implemented only where needed, eliminating the need to add `ReturnTypes()` to a bunch of existing functions.

hack/generator/pkg/conversions/property_assignment_function.go
theunrepentantgeek: I've created two wrapper end points - `ReadableConversionEndpoint` and `WritableConversionEndpoint` that achieve a nice reduction in duplication.

hack/generator/pkg/conversions/property_assignment_function.go
```diff
@@ -271,58 +269,45 @@ func (fn *PropertyAssignmentFunction) generateAssignments(
...
+
+		// Generate a conversion from one endpoint to another
+		conv, err := fn.createConversion(sourceEndpoint, destinationEndpoint)
+		if err != nil {
+			// An error was returned, we abort creating conversions for this object
+			return errors.Wrapf(
+				err,
+				"creating conversion to %s by %s",
+				sourceEndpoint,
+				destinationEndpoint)
```
babbageclunk, line 307: Are the arguments here the right way around?

babbageclunk (approved):
Nice - I like the readable/writable endpoint split!

hack/generator/pkg/conversions/property_assignment_function.go
theunrepentantgeek: They need to be swapped; will do that in my next PR.

PR #1548, opened 2021-06-02
Flattening

### LINKED ISSUE #1447: Flatten "properties" to avoid unnecessary nesting
For example in the generated `Redis` types, there is a field `Properties` which probably ought to have its fields promoted up a level.

This is a pattern that is quite common (for example Storage account also has a `Properties`)

Porges:
Closes #1447

Applies the `x-ms-client-flatten` attribute when it is present in the Swagger.

Highly recommend reviewing [ignoring whitespace](https://github.com/Azure/azure-service-operator/pull/1548/files?diff=split&w=1).

gif

**If applicable**:
- [x] this PR contains tests


Taskfile.yml
```diff
@@ -43,7 +43,7 @@ tasks:
     - task: generator:ci
     - task: controller:generate-crds # required here to prevent it running twice in controller:ci, remove once go-task dedupes dependencies
     - task: controller:ci
-    - task: crossplane:ci
+    #- task: crossplane:ci
```
Porges, line 46: Need to fix SQL not being able to be flattened.

matthchr: Before you merge this PR or after?

Porges: It feels bad to comment out things to pass CI ;)

hack/generator/pkg/codegen/testdata/Flattening/FlattenedPropertiesAreFlattened.golden
```diff
@@ -0,0 +1,16 @@
+// Code generated by k8s-infra. DO NOT EDIT.
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+package v1alpha1api20200101
+
+//Generated from: https://test.test/schemas/2020-01-01/test.json
+type Test struct {
+	InnerFlat    *Inner `json:"inner_flat,omitempty"`
+	InnerNotFlat *Inner `json:"inner_not_flat,omitempty"`
```
Porges, line 9: Note that this doesn’t do the desired thing yet, due to `gojsonschema` not exposing custom attributes.

hack/generated/controllers/crd_cosmosdb_test.go
```diff
@@ -29,15 +29,13 @@ func Test_CosmosDB_CRUD(t *testing.T) {
...
-			Properties: documentdb.DatabaseAccountCreateUpdateProperties{
-				DatabaseAccountOfferType: documentdb.DatabaseAccountCreateUpdatePropertiesDatabaseAccountOfferTypeStandard,
-				Locations: []documentdb.Location{
-					{
-						LocationName: &tc.AzureRegion,
-					},
+			Location:                 &tc.AzureRegion,
+			Owner:                    testcommon.AsOwner(rg.ObjectMeta),
+			Kind:                     &kind,
+			DatabaseAccountOfferType: documentdb.DatabaseAccountsSpecPropertiesDatabaseAccountOfferTypeStandard,
```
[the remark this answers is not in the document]
theunrepentantgeek: I had the thought of removing duplicate words (including plural/single matches) from enumeration names.

For example, 
* `DatabaseAccountsSpecPropertiesDatabaseAccountOfferTypeStandard` would become
* `Database` `Accounts` `Spec` `Properties` `Database` `Account` `Offer` `Type` Standard` would become
* `Database` `Accounts` `Spec` `Properties` `Offer` `Type` `Standard` would become
* `DatabaseAccountsSpecPropertiesOfferTypeStandard` 

which is better. 

IIRC there were some odd edge cases though, so I dropped it. Worth revisiting.

hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
```diff
@@ -184,6 +185,134 @@ func (builder *convertToARMBuilder) referencePropertyHandler(
...
+
+	// build (x || y || …)
+	cond := conds[0]
+	for _, c := range conds[1:] {
+		cond = &dst.BinaryExpr{
+			X:  cond,
+			Op: token.LOR,
+			Y:  c,
+		}
+	}
```
theunrepentantgeek, line 295: I think this should be a helper in `astbuilder` - something like

``` go
func Or(expr ...dst.Expr) dst.Expr
```

Might want to write `And()` while you're there.

hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
```diff
@@ -184,6 +185,134 @@ func (builder *convertToARMBuilder) referencePropertyHandler(
...
+			List: []dst.Stmt{
+				astbuilder.SimpleAssignment(
+					astbuilder.Selector(dst.NewIdent(builder.resultIdent), string(toPropName)),
+					token.ASSIGN,
+					&dst.UnaryExpr{
+						Op: token.AND,
+						X: &dst.CompositeLit{
+							Type: toPropTypeName.AsType(builder.codeGenerationContext),
+						},
+					}),
```
theunrepentantgeek, line 310: I've got an `astbuilder` helper for `CompositeLit` in one of my PRs, you may want to steal it.

hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
```diff
@@ -184,6 +185,134 @@ func (builder *convertToARMBuilder) referencePropertyHandler(
...
+		}
+	}
+
+	// build if (conds…) { target.prop = &TargetType{} }
+	return &dst.IfStmt{
+		Cond: cond,
+		Body: &dst.BlockStmt{
+			List: []dst.Stmt{
+				astbuilder.SimpleAssignment(
+					astbuilder.Selector(dst.NewIdent(builder.resultIdent), string(toPropName)),
```
theunrepentantgeek, line 303: I think we're using this pattern in a few places - maybe we need `astbuilder.QualifiedAssignment(expr, string, ...)`

hack/generator/pkg/astmodel/conversion_function_builder.go
```diff
@@ -168,8 +170,8 @@ func (builder *ConversionFunctionBuilder) BuildConversion(params ConversionParam
 	var sourceSB strings.Builder
 	params.SourceType.WriteDebugDescription(&sourceSB, types)
 	var destinationSB strings.Builder
-	params.SourceType.WriteDebugDescription(&destinationSB, types)
-	panic(fmt.Sprintf("don't know how to perform conversion for %T -> %T", sourceSB.String(), destinationSB.String()))
+	params.DestinationType.WriteDebugDescription(&destinationSB, types)
```
theunrepentantgeek, line 173: Whoops! Thanks. 😊

hack/generator/pkg/astmodel/property_definition.go
```diff
@@ -24,9 +24,29 @@ type PropertyDefinition struct {
 	propertyType                     Type
 	description                      string
 	hasKubebuilderRequiredValidation bool
+	flatten                          bool           // maps to x-ms-client-flatten: should the propertyType be flattened into the parent
+	flattenedFrom                    []PropertyName // a list of property names from whence this property was flattened
```
theunrepentantgeek, line 28: Elsewhere, you have in `buildFlattenedAssignment` a check like this:

``` go
if len(toProp.FlattenedFrom()) > 1 {
	// this doesn't appear to happen anywhere in the JSON schemas currently
	panic("need to implement multiple levels of flattening!")
}
```

Doesn't that mean that `flattenedFrom` will only ever have one member?

hack/generator/pkg/codegen/pipeline_flatten_properties.go
```diff
@@ -0,0 +1,146 @@
...
+
+			// safety check:
+			if err := checkForDuplicateNames(newProps); err != nil {
+				return nil, err
+			}
+
+			result := it.WithoutProperties().WithProperties(newProps...)
+
+			return result, nil
+		},
```
theunrepentantgeek, line 60: I really dislike putting methods inline like this - the code becomes impenetrable. Recommend pulling out the anonymous functions as methods, giving them an intention revealing name.

hack/generator/pkg/codegen/pipeline_flatten_properties.go
```diff
@@ -0,0 +1,146 @@
...
+			flattenedProps = append(flattenedProps, innerProps...)
+		} else {
+			flattenedProps = append(flattenedProps, prop)
+		}
+	}
+
+	return flattenedProps, nil
+}
+
+func flattenPropType(propType astmodel.Type, defs astmodel.Types) ([]*astmodel.PropertyDefinition, error) {
```
theunrepentantgeek, line 110: I think what you want here is a nested `TypeVisitor` as otherwise you're going to have problems. 
If you alternatively really want to just smoosh everything down, handle `MetaType` by using `Unwrap()`.

hack/generator/pkg/codegen/pipeline_flatten_properties_test.go
```diff
@@ -0,0 +1,75 @@
...
+		g.Expect(fs).To(HaveLen(1))
+
+		var f astmodel.Function
+		for _, single := range fs {
+			f = single
+			break
+		}
+
+		g.Expect(f.Name()).To(Equal("Flatten"))
+	*/
```
theunrepentantgeek, line 74: Commented code?

hack/generator/pkg/codegen/pipeline_status_from_swagger.go
```diff
@@ -60,40 +59,35 @@ func augmentResourcesWithStatus(idFactory astmodel.IdentifierFactory, config *co
...
 					newTypes.Add(typeDef)
 				}
 			}
 
-			// all non-resources are added regardless of whether they are used
-			// if they are not used they will be pruned off by a later pipeline stage
-			newTypes.AddAll(statusTypes.otherTypes)
-
-			klog.V(1).Infof("Found status information for %v resources", found)
+			klog.V(1).Infof("Found status information for %v resources", matchedResources)
```
theunrepentantgeek, line 90: Is level 1 correct here?

hack/generator/pkg/codegen/pipeline_status_from_swagger.go
```diff
@@ -106,7 +100,19 @@ type statusTypes struct {
...
-	otherTypes []astmodel.TypeDefinition
+	otherTypes astmodel.Types
+}
+
+func (st statusTypes) findResourceType(typeName astmodel.TypeName) (astmodel.Type, bool) {
+	if statusDef, ok := st.resourceTypes.tryFind(typeName); ok {
+		klog.V(4).Infof("Swagger information found for %v", typeName)
+		return statusDef, true
+	} else {
+		klog.V(4).Infof("Swagger information missing for %v", typeName)
```
theunrepentantgeek, line 111: Suggest that messages about missing information should be more important than when it's found - maybe make this `V(3)`?

hack/generator/pkg/codegen/testdata/Flattening/FlattenedPropertiesAreFlattened.golden
theunrepentantgeek: What needs to happen? Does this block merging?

matthchr:
This will probably have conflicts with #1556 - we should coordinate.

hack/generator/pkg/astmodel/property_definition.go
Porges: At the moment. In the future some team might go crazy and squash multiple levels… it’s easier to have this do the right thing and then fail later in the pipeline.

hack/generator/pkg/codegen/pipeline_flatten_properties_test.go
Porges: oops. this was regarding an alternate implementation idea that I abandoned

hack/generator/pkg/codegen/pipeline_status_from_swagger.go
Porges: We insert an error type (below) so that if the type is used you will get a warning logged. These are diagnostic if you want to see what _would_ be missing for pruned types, so they aren't relevant to normal usage.

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
```diff
@@ -224,6 +225,99 @@ func (builder *convertFromARMBuilder) ownerPropertyHandler(
...
+	fromPropType, err := allTypes.FullyResolve(fromProp.PropertyType())
+	if err != nil {
+		panic(err)
+	}
+
+	var fromPropObjType *astmodel.ObjectType
+	var objOk bool
+	// (2.) resolve any optional type
+	if fromPropOptType, ok := fromPropType.(*astmodel.OptionalType); ok {
+		// TODO: need to generate a nil check in the generated code?
```
matthchr, line 269: Does this TODO need to be done? The fact that this works suggests otherwise?

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
```diff
@@ -224,6 +225,99 @@ func (builder *convertFromARMBuilder) ownerPropertyHandler(
...
+func (builder *convertFromARMBuilder) buildFlattenedAssignment(toProp *astmodel.PropertyDefinition, fromProp *astmodel.PropertyDefinition) []dst.Stmt {
+	if len(toProp.FlattenedFrom()) > 1 {
+		// this doesn't appear to happen anywhere in the JSON schemas currently
+		panic("need to implement multiple levels of flattening!")
+	}
+
+	allTypes := builder.codeGenerationContext.GetAllReachableTypes()
+
+	// the from shape here must be:
+	// 1. maybe a typename, pointing to…
```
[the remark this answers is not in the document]
matthchr: oooh I see now. It's cascading downwards?

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
```diff
@@ -224,6 +225,99 @@ func (builder *convertFromARMBuilder) ownerPropertyHandler(
...
+
+		// (4.) resolve the inner object type
+		fromPropObjType, objOk = elementType.(*astmodel.ObjectType)
+	} else {
+		// (4.) resolve the inner object type
+		fromPropObjType, objOk = fromPropType.(*astmodel.ObjectType)
+	}
+
+	if !objOk {
+		panic("property marked as flattened from non-object type, which doesn’t make sense")
```
matthchr, line 284: Log some prop names and whatnot

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
```diff
@@ -224,6 +225,99 @@ func (builder *convertFromARMBuilder) ownerPropertyHandler(
...
+	allTypes := builder.codeGenerationContext.GetAllReachableTypes()
+
+	// the from shape here must be:
+	// 1. maybe a typename, pointing to…
+	// 2. maybe optional, wrapping …
+	// 3. maybe a typename, pointing to…
+	// 4. an object type
+
+	// (1.) resolve any outer typename
+	fromPropType, err := allTypes.FullyResolve(fromProp.PropertyType())
```
matthchr, line 260: would extracting the discovery of `fromPropObjType` to a function make things any clearer? Or just clutter the type with an extra function?

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
```diff
@@ -224,6 +225,99 @@ func (builder *convertFromARMBuilder) ownerPropertyHandler(
...
+
+	// *** Now generate the code! ***
+	toPropName := toProp.PropertyName()
+	nestedProp, ok := fromPropObjType.Property(toPropName)
+	if !ok {
+		panic("couldn't find source of flattened property")
+	}
+
+	stmts := builder.typeConversionBuilder.BuildConversion(astmodel.ConversionParameters{
+		Source:            astbuilder.Selector(astbuilder.Selector(dst.NewIdent(builder.typedInputIdent), string(fromProp.PropertyName())), string(toPropName)),
```
matthchr, line 295: This line is pretty long - break maybe?

hack/generator/pkg/codegen/pipeline_flatten_properties.go
Porges: Here I only want to check the top-level type, and there are only a few types that can be flattened. `TypeVisitor` will walk through the whole tree.

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
```diff
@@ -224,6 +225,99 @@ func (builder *convertFromARMBuilder) ownerPropertyHandler(
 	return []dst.Stmt{result}
 }
 
+func (builder *convertFromARMBuilder) flattenedPropertyHandler(
```
matthchr, line 228: Some of the other methods don't do this (shame on me), but it might be nice to have a skeleton of what code is actually generated from this in the comment of the method.

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
```diff
@@ -224,6 +225,99 @@ func (builder *convertFromARMBuilder) ownerPropertyHandler(
...
+	}
+
+	// *** Now generate the code! ***
+	toPropName := toProp.PropertyName()
+	nestedProp, ok := fromPropObjType.Property(toPropName)
+	if !ok {
+		panic("couldn't find source of flattened property")
+	}
+
+	stmts := builder.typeConversionBuilder.BuildConversion(astmodel.ConversionParameters{
```
matthchr, line 294: minor: maybe put the `astmodel.ConversionParameters` onto a new line? It gets a bit lost at the end of this one IMO

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
```diff
@@ -224,6 +225,99 @@ func (builder *convertFromARMBuilder) ownerPropertyHandler(
...
+		SourceType:        nestedProp.PropertyType(),
+		Destination:       astbuilder.Selector(dst.NewIdent(builder.receiverIdent), string(toPropName)),
+		DestinationType:   toProp.PropertyType(),
+		NameHint:          string(toProp.PropertyName()),
+		ConversionContext: nil,
+		AssignmentHandler: nil,
+		Locals:            builder.locals,
+	})
+
+	// failed?
```
matthchr, line 305: Is this a failure (panic?) or an expected case where there just wasn't anything to do? The comment doesn't say enough so IMO should either be removed or expanded

hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
```diff
@@ -66,6 +66,7 @@ func newConvertToARMFunctionBuilder(
 		result.fixedValuePropertyHandler("Type"),
 		result.fixedValuePropertyHandler("ApiVersion"),
 		result.propertiesWithSameNameHandler,
+		result.flattenedPropertyHandler,
```
matthchr, line 69: minor: order matters here I think, and you want to be above `propertiesWithSameNameHandler` I'm pretty sure

matthchr: Like you were in the `from_arm` direction

hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
```diff
@@ -184,6 +185,134 @@ func (builder *convertToARMBuilder) referencePropertyHandler(
...
+		// (3.) resolve any inner typename
+		toPropTypeName = optType.Element().(astmodel.TypeName)
+		toPropType, err = allTypes.FullyResolve(optType.Element())
+		if err != nil {
+			panic(err)
+		}
+	}
+
+	// (4.) we have found the underlying object type
+	toPropObjType := toPropType.(*astmodel.ObjectType)
```
matthchr, line 235: Same question, if this part would be clearer as a method that was just called here?

Wondering further: Is this and the code in the other direction actually the same (modulo setting the `needToInitializeToProp`)? Maybe you could refactor to a shared function: `func lookupFlattenedType(t astmodel.Type) (*astmodel.ObjectType, wasOptional bool, error)`?

hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
```diff
@@ -184,6 +185,134 @@ func (builder *convertToARMBuilder) referencePropertyHandler(
...
+	if needToInitializeToProp {
+		result = []dst.Stmt{builder.buildToPropInitializer(fromProps, toPropTypeName, toPropName)}
+	}
+
+	// Copy each from-prop into the to-prop
+	for _, fromProp := range fromProps {
+		// find the corresponding inner property on the to-prop type
+		toSubProp, ok := toPropObjType.Property(fromProp.PropertyName())
+		if !ok {
+			panic("unable to find source of flattened property")
```
matthchr, line 250: Log specifics

hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
```diff
@@ -184,6 +185,134 @@ func (builder *convertToARMBuilder) referencePropertyHandler(
...
+		}
+
+		// generate conversion
+		stmts := builder.typeConversionBuilder.BuildConversion(astmodel.ConversionParameters{
+			Source:            astbuilder.Selector(dst.NewIdent(builder.receiverIdent), string(fromProp.PropertyName())),
+			SourceType:        fromProp.PropertyType(),
+			Destination:       astbuilder.Selector(astbuilder.Selector(dst.NewIdent(builder.resultIdent), string(toPropName)), string(toSubProp.PropertyName())),
```
matthchr, line 257: long line - linebreak?

hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
```diff
@@ -184,6 +185,134 @@ func (builder *convertToARMBuilder) referencePropertyHandler(
...
+	// Copy each from-prop into the to-prop
+	for _, fromProp := range fromProps {
+		// find the corresponding inner property on the to-prop type
+		toSubProp, ok := toPropObjType.Property(fromProp.PropertyName())
+		if !ok {
+			panic("unable to find source of flattened property")
+		}
+
+		// generate conversion
+		stmts := builder.typeConversionBuilder.BuildConversion(astmodel.ConversionParameters{
```
matthchr, line 254: minor, style: Put the `astmodel.ConversionParameters` on the next line maybe?

hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
```diff
@@ -184,6 +185,134 @@ func (builder *convertToARMBuilder) referencePropertyHandler(
...
+			SourceType:        fromProp.PropertyType(),
+			Destination:       astbuilder.Selector(astbuilder.Selector(dst.NewIdent(builder.resultIdent), string(toPropName)), string(toSubProp.PropertyName())),
+			DestinationType:   toSubProp.PropertyType(),
+			NameHint:          string(toSubProp.PropertyName()),
+			ConversionContext: nil,
+			AssignmentHandler: nil,
+			Locals:            builder.locals,
+		})
+
+		// failed?
```
matthchr, line 265: Same q as other file - this comment makes it not clear if this is expected or an error case. Seems expected given you're not panicing? Might be worth noting when this can happen.

hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
```diff
@@ -184,6 +185,134 @@ func (builder *convertToARMBuilder) referencePropertyHandler(
...
+			return nil
+		}
+
+		result = append(result, stmts...)
+	}
+
+	return result
+}
+
+func (builder *convertToARMBuilder) buildToPropInitializer(fromProps []*astmodel.PropertyDefinition, toPropTypeName astmodel.TypeName, toPropName astmodel.PropertyName) dst.Stmt {
```
matthchr, line 276: minor: linebreaks for long function signature?

hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
```diff
@@ -184,6 +185,134 @@ func (builder *convertToARMBuilder) referencePropertyHandler(
 	)
 }
 
+func (builder *convertToARMBuilder) flattenedPropertyHandler(
```
matthchr, line 188: Same comment as above file about being nice to have rough code shape in comment

hack/generator/pkg/astmodel/code_generation_context.go
```diff
@@ -142,7 +142,9 @@ func (codeGenContext *CodeGenerationContext) GetAllReachableTypes() Types {
 	for _, pkgImport := range codeGenContext.packageImports.AsSlice() {
 		types, found := codeGenContext.GetTypesInPackage(pkgImport.packageReference)
 		if found {
-			result.AddTypes(types)
+			for k, v := range types {
```
matthchr, line 145: Why this change?

hack/generator/pkg/codegen/pipeline_status_from_swagger.go
```diff
@@ -40,9 +39,9 @@ added to the Status field of the Resource type, after we have renamed all the st
 avoid any conflicts with existing Spec types that have already been defined.
 
 */
-func augmentResourcesWithStatus(idFactory astmodel.IdentifierFactory, config *config.Configuration) PipelineStage {
+func addStatusFromSwagger(idFactory astmodel.IdentifierFactory, config *config.Configuration) PipelineStage {
 	return MakePipelineStage(
-		"augmentStatus",
+		"statusFromSwagger",
```
matthchr, line 44: minor:
```suggestion
		"addStatusFromSwagger",
```

hack/generator/pkg/codegen/pipeline_status_from_swagger.go
matthchr: @theunrepentantgeek are you suggesting it should be lower or higher? I'd say given this log isn't very verbose that it's a 0 or 1.

hack/generator/pkg/codegen/pipeline_flatten_properties.go
```diff
@@ -0,0 +1,146 @@
...
+
+import (
+	"context"
+
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/astmodel"
+	"github.com/pkg/errors"
+)
+
+func flattenProperties() PipelineStage {
+	return MakePipelineStage("flatten-properties", "Apply flattening to properties marked for flattening", applyPropertyFlattening)
```
matthchr, line 16: minor - change this to `flattenProperties` to match other casing of stages?

hack/generator/pkg/codegen/testdata/Flattening/FlattenedPropertiesAreFlattened.golden
matthchr: I _think_ @Porges is referring to our inability to write golden file tests with Swagger input, which means we can't actually _set_ the flattening flag.

hack/generator/pkg/codegen/testdata/Flattening/FlattenedPropertiesAreFlattened.json
```diff
@@ -0,0 +1,29 @@
+{
+    "$comment": "Check that flattened definitions are inlined into parent type",
```
matthchr, line 2: If you're planning on merging this, probably worth a big TODO in the comment saying that it doesn't actually generate correct code because of the `gojsonschema` point you mentioned above.

hack/generator/pkg/codegen/pipeline_status_from_swagger.go
```diff
@@ -16,18 +16,17 @@ import (
 	"strings"
 	"sync"
 
-	kerrors "k8s.io/apimachinery/pkg/util/errors"
-
 	"github.com/Azure/azure-service-operator/hack/generator/pkg/astmodel"
 	"github.com/Azure/azure-service-operator/hack/generator/pkg/config"
```
matthchr, line 20: minor: I think this needs ordering fixups

hack/generator/pkg/codegen/pipeline_flatten_properties.go
```diff
@@ -0,0 +1,146 @@
...
+			return errors.Errorf("Flattening caused duplicate property name %q", p.PropertyName())
+		}
+
+		names[p.PropertyName()] = struct{}{}
+	}
+
+	return nil
+}
+
+// containerName is the type that will hold the flattened properties, i.e. act as a container for them
```
matthchr, line 85: Comment doesn't seem to match function name (or parameters?)

hack/generator/pkg/codegen/pipeline_flatten_properties.go
```diff
@@ -0,0 +1,146 @@
...
+			return astmodel.IdentityVisitOfFlaggedType(this, it, ctx)
+		},
+	}.Build()
+}
+
+func checkForDuplicateNames(props []*astmodel.PropertyDefinition) error {
+	names := make(map[astmodel.PropertyName]struct{})
+	for _, p := range props {
+		if _, ok := names[p.PropertyName()]; ok {
+			return errors.Errorf("Flattening caused duplicate property name %q", p.PropertyName())
```
matthchr, line 76: really minor, style: `Flattening` -> `flattening` because I'm pretty sure for some reason Go style is to not uppercase first word in error.

hack/generator/pkg/codegen/pipeline_flatten_properties.go
```diff
@@ -0,0 +1,146 @@
...
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package codegen
+
+import (
+	"context"
+
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/astmodel"
```
matthchr, line 11: minor: ordering

hack/generator/pkg/codegen/pipeline_flatten_properties_test.go
```diff
@@ -0,0 +1,75 @@
...
+ * Licensed under the MIT license.
+ */
+
+package codegen
+
+import (
+	"context"
+	"testing"
+
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/astmodel"
```
matthchr, line 12: minor: ordering

hack/generator/pkg/codegen/pipeline_status_augment.go
```diff
@@ -0,0 +1,137 @@
...
+				return nil, err
+			}
+		}
+
+		return main, nil
+	}
+}
+
+// flattenAugmenter copies across the "flatten" property from Swagger
+func flattenAugmenter(allTypes astmodel.Types) augmenter {
```
matthchr, line 68: Wondering if there are other things in the future we want to merge in from the Swagger types, will we need to duplicate most of this main type vs Swagger type (main prop vs swagger prop) logic?

Maybe when we get a second one of these we should refactor this to have a pluggable last bit for the last method (line 114?) Or maybe you pull the first few anonymous functions out into named functions so that if we do end up with other impls they look like:

```
merger.Add(defaultType)
merger.Add(defaultTypeName)
merger.Add(lookupSwaggerType)
merger.Add(... // Your custom handler here)
```

matthchr (approved):
I think the core of this came out pretty clean actually, nice!

hack/generator/pkg/codegen/pipeline_status_from_swagger.go
theunrepentantgeek: I was more thinking that having actionable warnings at a more significant level than information means that it's easy to get just those warnings when needed. 

Having both branches at `V(4)` means that you've got no choice other than sift through all the log messages to find the ones where it's missing.

hack/generator/pkg/codegen/pipeline_status_augment.go
```diff
@@ -0,0 +1,137 @@
...
+		"augmentSpecWithStatus",
+		"Merges information from Status into Spec",
+		func(ctx context.Context, types astmodel.Types) (astmodel.Types, error) {
+
+			newTypes := make(astmodel.Types)
+			newTypes.AddTypes(types)
+
+			// build the augmenter we will use:
+			augmenter := fuseAugmenters(
+				flattenAugmenter(newTypes), // note that this can alter newTypes as a side-effect, so it must be preloaded with all the types
```
babbageclunk, line 25: This sounds iffy. I can understand it needing all of the types loaded for lookups, but why does the augmenter mutate types?

I read ahead and see the place where the update happens for typenames. Why is that needed, rather than just waiting until the loop below hits the definitions corresponding to the type names.

Ohhhhh, I think I get it - the loop below only handles resources, because those are the only things that have spec and status matched up between the json and swagger. Is that right?

hack/generator/pkg/codegen/pipeline_status_from_swagger.go
```diff
@@ -16,18 +16,17 @@ import (
...
 	"github.com/Azure/azure-service-operator/hack/generator/pkg/jsonast"
```
babbageclunk, line 21: Split these out into a separate stanza?

hack/generator/pkg/astmodel/property_definition.go
```diff
@@ -24,9 +24,29 @@ type PropertyDefinition struct {
...
+	result := *property
+	result.flattenedFrom = append([]PropertyName{name}, result.flattenedFrom...)
+	return &result
+}
+
+func (property *PropertyDefinition) WasFlattened() bool {
+	return len(property.flattenedFrom) > 0
+}
+
+func (property *PropertyDefinition) WasFlattenedFrom(name PropertyName) bool {
```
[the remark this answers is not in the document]
babbageclunk: Ah, I think I see this from the arm conversion - it's to handle the possibility of multiple levels of flattening?

babbageclunk (approved):
I'm with Matt, this came out really cleanly! Lots of questions but I think those are just getting my head around what's going on rather than problems.

hack/generator/pkg/astmodel/code_generation_context.go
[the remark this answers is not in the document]
Porges: It can find duplicates, somehow. I didn't think about it too much 🙂

hack/generator/pkg/astmodel/conversion_function_builder.go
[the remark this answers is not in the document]
Porges: Very!

hack/generator/pkg/astmodel/property_definition.go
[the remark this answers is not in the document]
Porges: @babbageclunk exactly. I'll make this comment clearer.

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
Porges: The code here doesn't know, really. It has asked the conversion builder to build a conversion for the inner types and that wasn't able to be built, so the overall conversion is not able to be built. The top-level handler will emit an error about which types cannot be converted, if no other conversion handler can do it.

I'll expand the comment here.

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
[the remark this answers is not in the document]
Porges: yes, the optional can contain a typename as well

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
Porges: It's hard to do that here because mostly it relies on the other conversion handlers to generate the inner conversion. I'll do a sketch 🙂

hack/generator/pkg/codegen/pipeline_flatten_properties_test.go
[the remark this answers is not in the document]
Porges: see other comment thread https://github.com/Azure/azure-service-operator/pull/1548/files#r648750613

hack/generator/pkg/codegen/pipeline_status_augment.go
Porges: It's the bit about identifying "the same" type via object structure. We can't do that just by name, we can only do it while we are walking/exploring the resource types. When we identify that two types are the same we update the spec side of that type in-place.

Perhaps it would be clearer to output a set of new types instead?

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
matthchr: Yeah it's tricky because the guts are actually not generated by this, but yeah just a really simple skeleton that sketches out what "flattening a property" actually entails at a high level.

hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
Porges: added

hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
Porges: I'm not sure it would shorten much here?

hack/generator/pkg/astmodel/armconversion/shared.go
```diff
@@ -100,8 +100,8 @@ func generateTypeConversionAssignments(
 	return result
 }
 
-// NewARMTransformerImpl creates a new interface with the specified ARM conversion functions
-func NewARMTransformerImpl(
+// NewARMConversionInterfaceImpl creates a new interface with the specified ARM conversion functions
```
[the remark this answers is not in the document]
Porges: done!

hack/generator/pkg/astmodel/code_generation_context.go
Porges: duplicates were being found. I didn't think much about why.

hack/generator/pkg/codegen/pipeline_flatten_properties.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// containerName is the type that will hold the flattened properties, i.e. act as a container for them
+func flattenObjectType(objectType *astmodel.ObjectType, defs astmodel.Types) ([]*astmodel.PropertyDefinition, error) {
+	var flattenedProps []*astmodel.PropertyDefinition
+
+	props := objectType.Properties()
+	for _, prop := range props {
+		if prop.Flatten() {
+			innerProps, err := flattenPropType(prop.PropertyType(), defs)
```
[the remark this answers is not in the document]
Porges: HE SAID THE NAME

hack/generator/pkg/codegen/testdata/Flattening/FlattenedPropertiesAreFlattened.json
Porges: agreed

hack/generated/controllers/crd_cosmosdb_test.go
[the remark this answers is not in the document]
Porges: Added an issue to track: https://github.com/Azure/azure-service-operator/issues/1591

hack/generator/pkg/codegen/testdata/Flattening/FlattenedPropertiesAreFlattened.golden
Porges: Added issue to track: https://github.com/Azure/azure-service-operator/pull/1548#discussion_r647856588

hack/generated/controllers/crd_storageaccount_test.go
```diff
@@ -33,13 +33,11 @@ func Test_StorageAccount_CRUD(t *testing.T) {
 			Location: tc.AzureRegion,
 			Owner:    testcommon.AsOwner(rg.ObjectMeta),
 			Kind:     storage.StorageAccountsSpecKindBlobStorage,
-			Sku: storage.Sku{
-				Name: storage.SkuNameStandardLRS,
+			Sku: storage.StorageAccounts_Spec_Sku{
```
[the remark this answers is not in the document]
Porges: Yeah, will address here: https://github.com/Azure/azure-service-operator/issues/1591

hack/generated/controllers/crd_storageaccount_test.go
```diff
@@ -33,13 +33,11 @@ func Test_StorageAccount_CRUD(t *testing.T) {
...
+				Name: storage.StorageAccountsSpecSkuNameStandardLRS,
```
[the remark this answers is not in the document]
Porges: Tracking: https://github.com/Azure/azure-service-operator/issues/1591

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
Porges: Now done.

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
Porges: I gave it a go but it was messy. There should be a better way to refactor this to make it work with arbitrary depth as well, but will leave it for now.

hack/generator/pkg/codegen/pipeline_flatten_properties.go
```diff
@@ -0,0 +1,146 @@
...
+func applyPropertyFlattening(
+	ctx context.Context,
+	defs astmodel.Types) (astmodel.Types, error) {
+	visitor := makeFlatteningVisitor(defs)
+
+	result := make(astmodel.Types)
+	for name, def := range defs {
+		newDef, err := visitor.VisitDefinition(def, name)
+		if err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
Porges: done

hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
Porges: So originally I was thinking this is a performance optimization, since the code must iterate to find all the flattened properties. However, there's a correctness issue here, when a flattened property contains a property that has the same name! So yes, this should come first. I'll also need a good test for it but will delay until golden files can support flattening properly…

Porges: Tracking: https://github.com/Azure/azure-service-operator/issues/1597

hack/generator/pkg/codegen/pipeline_flatten_properties.go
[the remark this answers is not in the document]
Porges: Actually, I'm undoing this. The visitor already reports a good error!

hack/generator/pkg/codegen/pipeline_status_augment.go
Porges: Agree that we will probably want to pull parts of this out to be shared. But I won't do it in this PR until we have a use-case where we know what we want.

hack/generator/pkg/codegen/pipeline_status_from_swagger.go
Porges: I'll bump missing to `V(3)`.

hack/generator/pkg/astmodel/armconversion/convert_to_arm_function_builder.go
Porges: I did try this at first but it didn't come out very clean, the methods are a bit lopsided. I can try again when I look to redo it a bit to support multiple levels of flattening, but I'd rather push to get the PR done now.

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
Porges: Added a very high-level description.

hack/generator/pkg/codegen/pipeline_status_augment.go
Porges: I realized a few things at once: #1 this step of doing side-effect updates on the named types is completely unnecessary, since we can merge the types and return an unnamed object which will be named later (& #2 in fact, that's why we are now getting worse names in a bunch of cases!)

I've removed the side-effects and everything still works.

hack/generator/pkg/codegen/pipeline_flatten_properties.go
Porges: In this case I don't think there are much better names to use: the functions are very specific to the flattening visitor so it would just be something like `visitObjectTypeAndFlatten` which isn't any better than the `VisitObjectType:` field inside the `flatteningVisitor` method… 🤔

Taskfile.yml
Porges: seems to work again, perhaps #1556 fixed it

hack/generator/pkg/codegen/pipeline_flatten_properties.go
theunrepentantgeek: Meh. I still think having them inline makes the method too long to keep in your head as it's read, but it's not a blocker.

PR #1549, opened 2021-06-04
Add new CosmosDBSQLDatabase CRD

matthchr:
Without this, the existing CosmosDB CRD is not very useful, as it's just an account that cannot hold data

gif

**If applicable**:
- [x] this PR contains tests


cdmurph32:
I appreciate it @matthchr!

pkg/resourcemanager/cosmosdb/sqldatabase/cosmosdbsql.go
```diff
@@ -0,0 +1,275 @@
...
+
+	// Need to poll the polling URL, so not done yet!
+	return false, nil
+}
+
+func (m *AzureCosmosDBSQLDatabaseManager) Delete(ctx context.Context, obj runtime.Object, opts ...resourcemanager.ConfigOption) (bool, error) {
+	options := &resourcemanager.Options{}
+	for _, opt := range opts {
+		opt(options)
+	}
```
babbageclunk, line 152: `options` isn't used in this method - is it just for side-effects in the ConfigOption callbacks?

pkg/resourcemanager/cosmosdb/sqldatabase/cosmosdbsql.go
```diff
@@ -0,0 +1,275 @@
...
+
+/*
+ * ARMClient implementation
+ */
+
+func (m *AzureCosmosDBSQLDatabaseManager) Ensure(ctx context.Context, obj runtime.Object, opts ...resourcemanager.ConfigOption) (bool, error) {
+	options := &resourcemanager.Options{}
+	for _, opt := range opts {
+		opt(options)
+	}
```
babbageclunk, line 44: `options` isn't used in this method - is it just for side-effects in the ConfigOption callbacks?

pkg/resourcemanager/cosmosdb/sqldatabase/cosmosdbsql.go
```diff
@@ -0,0 +1,275 @@
...
+			errhelp.AlreadyExists,
+			errhelp.FailoverGroupBusy,
+		}
+		unrecoverableErrors := []string{
+			errhelp.InvalidFailoverGroupRegion,
+		}
+		return errhelp.IsErrorFatal(err, append(allowedErrors, notFoundErrorCodes...), unrecoverableErrors)
+	}
+	instance.Status.SetProvisioning("Resource request successfully submitted to Azure")
+	instance.Status.PollingURL = future.PollingURL()
```
babbageclunk, line 142: Is there anywhere that clears the polling url once the long-running operation completes? Should there be?

pkg/resourcemanager/cosmosdb/sqldatabase/cosmosdbsql.go
```diff
@@ -0,0 +1,275 @@
...
+
+		if helpers.ContainsString(finished, azerr.Type) {
+			return false, nil
+		}
+		instance.Status.Message = fmt.Sprintf("delete failed with: %s", err.Error())
+		return false, err
+	}
+
+	// Not done yet, keep trying
+	return false, nil
```
babbageclunk, line 192: Shouldn't this be `true, nil`? I think the sense of this return is "does the object still exist?", and returning false here causes AsyncReconciler to remove the finalizer.

pkg/resourcemanager/cosmosdb/sqldatabase/cosmosdbsql.go
```diff
@@ -0,0 +1,275 @@
...
+	if !ok {
+		return nil, fmt.Errorf("failed type assertion on kind: %s", obj.GetObjectKind().GroupVersionKind().String())
+	}
+	return db, nil
+}
+
+func makeTags(tags map[string]string) map[string]*string {
+	result := make(map[string]*string)
+	for key, value := range tags {
+		result[key] = &value
```
babbageclunk, line 243: Is this going to end up with all the values being the same?
```suggestion
		localValue := value
		result[key] = &localValue
```

pkg/resourcemanager/cosmosdb/sqldatabase/cosmosdbsql.go
```diff
@@ -0,0 +1,275 @@
...
+	var createUpdateOptions *documentdb.CreateUpdateOptions
+	if instance.Spec.Throughput != nil || autoscaleSettings != nil {
+		createUpdateOptions = &documentdb.CreateUpdateOptions{
+			Throughput:        instance.Spec.Throughput,
+			AutoscaleSettings: autoscaleSettings,
+		}
+	}
+
+	return documentdb.SQLDatabaseCreateUpdateParameters{
+		//Location: &instance.Spec.Location, // TODO: Is this really optional? TODO: if this passes with this commented out, maybe just remove it?
```
babbageclunk, line 266: Presumably this can be removed. 😁

pkg/resourcemanager/cosmosdb/sqldatabase/resource_differ.go
```diff
@@ -0,0 +1,147 @@
...
+func doResourcesMatch(expected *documentdb.SQLDatabaseResource, actual *documentdb.SQLDatabaseGetPropertiesResource) bool {
+	var expectedIDCoalesced *string
+	if expected != nil {
+		expectedIDCoalesced = expected.ID
+	}
+
+	var actualIDCoalesced *string
+	if actual != nil {
+		actualIDCoalesced = actual.ID
+	}
```
babbageclunk, line 17: I'm not sure this buys you anything - why not just use `expected.ID` where you're using `*expectedIDCoalesced`? Oh crap, is it because ID is a `*string`? Gross. Ok, as you were.

pkg/resourcemanager/cosmosdb/sqldatabase/resource_differ.go
```diff
@@ -0,0 +1,147 @@
...
+
+	if (expectedIDCoalesced == nil) != (actualIDCoalesced == nil) {
+		return false
+	}
+	if expectedIDCoalesced != nil && actualIDCoalesced != nil && *expectedIDCoalesced != *actualIDCoalesced {
+		return false
+	}
+
+	return true
```
babbageclunk, line 26:
```suggestion
	if expectedIDCoalesced == nil && actualIDCoalesced == nil {
		return true
	}

	return *expectedIDCoalesced == *actualIDCoalesced
```

pkg/resourcemanager/cosmosdb/sqldatabase/resource_differ.go
```diff
@@ -0,0 +1,147 @@
...
+	if expectedThroughputCoalesced != nil && actualThroughputCoalesced != nil &&
+		*expectedThroughputCoalesced != *actualThroughputCoalesced {
+		return false
+	}
+
+	if !doAutoscaleSettingsMatch(expectedAutoscaleCoalesced, actualAutoscaleCoalesced) {
+		return false
+	}
+
+	return true
```
babbageclunk, line 60: I think you can use a similar transformation to simplify this.

pkg/resourcemanager/cosmosdb/sqldatabase/resource_differ.go
```diff
@@ -0,0 +1,147 @@
...
+		return false
+	}
+
+	if expectedMaxThroughputCoalesced != nil && actualMaxThroughputCoalesced != nil {
+		if *expectedMaxThroughputCoalesced != *actualMaxThroughputCoalesced {
+			return false
+		}
+	}
+
+	return true
```
babbageclunk, line 84: And this.

pkg/resourcemanager/cosmosdb/sqldatabase/resource_differ.go
```diff
@@ -0,0 +1,147 @@
...
+			return false
+		}
+
+		if (expectedValue == nil) != (actualValue == nil) {
+			return false
+		}
+
+		if expectedValue != nil && actualValue != nil && *expectedValue != *actualValue {
+			return false
+		}
```
babbageclunk, line 121: Same here except with a `continue` rather than a `return true`.

pkg/resourcemanager/cosmosdb/sqldatabase/resource_differ.go
```diff
@@ -0,0 +1,147 @@
...
+	var actualResourceCoalesced *documentdb.SQLDatabaseGetPropertiesResource
+	if actual != nil {
+		actualResourceCoalesced = actual.Resource
+	}
+
+	if !doResourcesMatch(expectedResourceCoalesced, actualResourceCoalesced) {
+		return false
+	}
+
+	return true
```
babbageclunk, line 102:
```suggestion
	return doResourcesMatch(expectedResourceCoalesced, actualResourceCoalesced)
```

pkg/resourcemanager/cosmosdb/sqldatabase/resource_differ.go
```diff
@@ -0,0 +1,147 @@
...
+		return false
+	}
+
+	if expected.SQLDatabaseCreateUpdateProperties != nil {
+		if !doesThroughputMatch(expected.SQLDatabaseCreateUpdateProperties.Options, actualThroughput.ThroughputSettingsGetProperties) {
+			return false
+		}
+	}
+
+	return true
```
babbageclunk, line 146:
```suggestion
	if expected.SQLDatabaseCreateUpdateProperties == nil {
		return true
	}
	return doesThroughputMatch(expected.SQLDatabaseCreateUpdateProperties.Options, actualThroughput.ThroughputSettingsGetProperties)
```

babbageclunk (approved):
Looks great - mostly minor comments and questions. 
I think the sense of one of the returns from `CosmosDBSQLDatabase.Delete` is wrong, and there might be a pointer-capture bug in the `makeTags` func.

config/samples/azure_v1alpha1_cosmosdbsqldatabase.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: azure.microsoft.com/v1alpha1
+kind: CosmosDBSQLDatabase
+metadata:
+  name: cosmosdbsql-sample-1
+spec:
+  resourceGroup: resourcegroup-azure-operators
+  cosmosDBAccount: cosmosdb-sample-1
+
+  # Manual throughput (RU/s) for the database expressed in units of 100 request
```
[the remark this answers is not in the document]
matthchr: Yes you cannot "not set" it and then set it afterwards.

controllers/cosmosdbsqldatabase_controller_test.go
```diff
@@ -0,0 +1,194 @@
...
+	cosmosDBAccountName string,
+	dbName string,
+	expectedThroughput int32) (bool, error) {
+
+	throughputSettings, err := client.GetSQLDatabaseThroughput(ctx, rgName, cosmosDBAccountName, dbName)
+	if err != nil {
+		return false, errors.Wrap(err, "err getting cosmos sql DB throughput from Azure")
+	}
+
+	if throughputSettings.ThroughputSettingsGetProperties == nil {
```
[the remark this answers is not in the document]
matthchr: The way it works in C# basically means that any nil short-circuits the rest of the checks to the right and just returns the nil. I'm not sure exactly how it'd work with `*` though. Even if it didn't, this could become:
```
actual := throughputSettings.ThroughputSettingsGetProperties?.Resource?.Throughput?

if actual != nil {
    return *actual == expected
}

return false
```

In the actual diffing code, a lot of that coalescing could be turned into:
`a?.b?.c == d?.e?.f` too

pkg/resourcemanager/cosmosdb/sqldatabase/cosmosdbsql.go
matthchr: It's done in the polling helper (which is awkward and side-effecty but is how it's set up right now)

pkg/resourcemanager/cosmosdb/sqldatabase/cosmosdbsql.go
matthchr: Ahh! I think you're right.

Isn't there a linter for this? I thought @Porges had enabled it.

api/v1alpha1/cosmosdbsqldatabase_types.go
```diff
@@ -0,0 +1,75 @@
...
+	// AutoscaleSettings contains the user specified autoscale configuration.
+	AutoscaleSettings *AutoscaleSettings `json:"autoscaleSettings,omitempty"`
+
+	// Tags are key-value pairs associated with the resource.
+	Tags map[string]string `json:"tags,omitempty"`
+}
+
+type AutoscaleSettings struct {
+	// +kubebuilder:validation:Min=0
+	// MaxThroughput is the autoscale max RU/s of the database. This must not be specified if Throughput is specified.
```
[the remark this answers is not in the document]
matthchr: technically no, you can specify an empty autoscaleSettings.

Practically speaking we ought to just document it like you say as IMO it's clearer. Will change

api/v1alpha1/cosmosdbsqldatabase_types.go
```diff
@@ -0,0 +1,75 @@
...
+	// +kubebuilder:validation:Pattern=^[-\w\._\(\)]+$
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Required
+	// ResourceGroup is the resource group the CosmosDBSQLDatabase will be created in.
+	ResourceGroup string `json:"resourceGroup"`
+
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Required
+	// CosmosDBAccount is the account that the SQL database will be created in.
+	CosmosDBAccount string `json:"cosmosDBAccount"`
```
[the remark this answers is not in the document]
matthchr: Account is probably sufficient.

pkg/resourcemanager/cosmosdb/sqldatabase/resource_differ.go
matthchr: So the reason I didn't have any `return true`'s in this code is because it's really easy to accidentally write "these two match, we're good!" and forget that you're examining an object with 4 properties and you've only checked 2 of them. So I was striving the for "find mismatches, else if none true" to avoid that as I found myself making that mistake and then catching it like 2-3 times while writing these.

If folks don't think that's valuable in these cases (where there's only a single field) then I can change to do as you say. It felt like if another field was added to the struct later and somebody came into this method though, it'd be really easy to just "add stuff to the bottom" and not realize that actually it's wrong you need to go get rid of that `return true` as well.

config/samples/azure_v1alpha1_cosmosdbsqldatabase.yaml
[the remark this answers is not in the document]
babbageclunk: Ah, got it - might be worth noting in the sample?

pkg/resourcemanager/cosmosdb/sqldatabase/cosmosdbsql.go
babbageclunk: Yeah, I think there was! Maybe it's only running on the codegen stuff?

config/samples/azure_v1alpha1_cosmosdbsqldatabase.yaml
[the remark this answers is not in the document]
matthchr: I've added documentation about this as well.

pkg/resourcemanager/cosmosdb/sqldatabase/cosmosdbsql.go
matthchr: No it was just boilerplate from other controllers. I've removed the `options` construction for now and left `opts` unused.

pkg/resourcemanager/cosmosdb/sqldatabase/cosmosdbsql.go
matthchr: Yeah it's only running on the codegen stuff right now it looks like (because it's run through the taskfile which is only invoked on the hack/* stuff).

pkg/resourcemanager/cosmosdb/sqldatabase/resource_differ.go
matthchr: Unless any objection, will leave it as is for future extensibility and hopefully reduction in bugs.

babbageclunk: Yeah, that reasoning makes sense to me.

pkg/resourcemanager/cosmosdb/sqldatabase/cosmosdbsql.go
```diff
@@ -0,0 +1,294 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	cosmosDBSQLDatabaseClient, err := cosmosdb.GetCosmosDBSQLDatabaseClient(m.Creds)
+	if err != nil {
+		return false, errors.Wrapf(err, "failed to create cosmos DB SQL database client")
+	}
+
+	// TODO: What if we have a pollURL already but it's for a create
```
babbageclunk, line 155: I think you've handled this now, right? I guess you were typing this and then thought "darn it, now I have to go fix that!".

babbageclunk (approved):
Polling on delete looks good - nice catch!

pkg/resourcemanager/cosmosdb/sqldatabase/cosmosdbsql.go
matthchr: Oh yea - will fix in a follow-on PR though just to not invalidate my CI pass

PR #1550, opened 2021-06-09
Build aso-controller image

Porges:
Build and generate a Docker image containing the v2 controller. (Just for demo purposes now, we will need a release workflow as well.)

gif


theunrepentantgeek (approved):

babbageclunk (approved):

.github/workflows/pr-validation.yml
```diff
@@ -20,8 +20,30 @@ jobs:
...
+      - name: Build docker image
+        run: |
+          container_id=${{steps.devcontainer.outputs.id}}
+          docker exec "$container_id" task controller:docker-build
+          docker cp "$container_id:/workspace/hack/generated/bin" .
+          mv bin/*.tar azure-service-operator.tar
+
+      - name: Archive Docker image
+        uses: actions/upload-artifact@v2
+        with:
```
matthchr, line 43: In the ADO job we publish to an MCR feed instead of uploading to artifacts. Do we want to do something similar here?

Porges: This is just for the PR validation run, so you can download the output from it. We’ll add another workflow for the actual release.

PR #1556, opened 2021-06-10
Generate the latest resource versions from the 2019 spec

### LINKED ISSUE #1488: Ensure that the types we're generating are actually the latest version
The ARM JSON schemas don't necessarily have the latest API version types exposed via the `2015-05-01` specification which is what we're relying on today. We should ensure that we're actually generating the latest APIVersion for each type we're interested in, which may require use updating to use a newer "rollup version" of the JSON schema.

matthchr:
- Name every resource we're generating explicitly to ensure that we don't accidentally include resources we 
   didn't expect (or write tests for).
 - Use constants rather than string literals in tests when performing `HEAD` resource.

Closes #1488


matthchr:
Build is failing because [this PR](https://github.com/Azure/azure-resource-manager-schemas/pull/1755) needs to merge, then I will update submodule to point to that commit

hack/generator/pkg/conversions/property_assignment_function.go
```diff
@@ -90,6 +92,8 @@ func NewPropertyAssignmentToFunction(
 		knownLocals:     astmodel.NewKnownLocalsSet(idFactory),
 	}
 
+	result.knownLocals.Add("destination")
```
theunrepentantgeek, line 95: Maybe add a comment that I (@unrepentantgeek) intend to do something that's less hacky here.

hack/generator/pkg/conversions/property_assignment_function.go
```diff
@@ -64,6 +64,8 @@ func NewPropertyAssignmentFromFunction(
 		knownLocals:     astmodel.NewKnownLocalsSet(idFactory),
 	}
 
+	result.knownLocals.Add("source")
```
theunrepentantgeek, line 67: Here too: maybe add a comment that I (@unrepentantgeek) intend to do something that's less hacky here.

babbageclunk (approved):
This looks good to me. I don't think the refHandler code to support including other files is too hacky - is the concern the timing of when it happens?

theunrepentantgeek (approved):

matthchr:
> This looks good to me. I don't think the refHandler code to support including other files is too hacky - is the concern the timing of when it happens?

The bit that felt kinda hack was identifying when it's a ref. Some refs we assume are to types, and some refs (well, this _one_) isn't to a type but a whole file we need to include. The bit I was wondering if there was a better way was how to differentiate between the two. What I did does work though.

PR #1558, opened 2021-06-10
Allow resources to have additional properties

theunrepentantgeek:

Currently resources only have `Spec` and `Status` properties, but we have a need (for storage versions) to have additional properties as well. This PR extends support to allow additional properties on resources, but treating `Spec` and `Status` as special cases.

First consumer will be for Storage versions where the original group-version-kind of the API type needs to be preserved.

gif

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/hack/generator/pkg/astmodel/resource_type.go
+++ b/hack/generator/pkg/astmodel/resource_type.go
@@ -288,15 +288,30 @@ func (resource *ResourceType) EmbeddedProperties() []*PropertyDefinition {
 	}
 }
 
-// WithProperty creates a new ObjectType with another property attached to it
+// WithProperty creates a new ResourceType with another property attached to it
 // Properties are unique by name, so this can be used to Add and Replace a property
 func (resource *ResourceType) WithProperty(property *PropertyDefinition) *ResourceType {
-	// Create a copy of objectType to preserve immutability
+	// Create a copy to preserve immutability
 	result := resource.copy()
 	result.properties[property.propertyName] = property
 
 	return result
 }
+
+// WithoutProperty creates a new ResourceType that's a copy without the specified property
+func (resource *ResourceType) WithoutProperty(name PropertyName) *ResourceType {
+
+	if name == "Status" || name == "Spec" {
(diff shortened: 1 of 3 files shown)


hack/generator/pkg/astmodel/resource_type.go
```diff
@@ -286,26 +288,73 @@ func (resource *ResourceType) EmbeddedProperties() []*PropertyDefinition {
...
 	}
 
 	if resource.status != nil {
-		statusProperty := NewPropertyDefinition("Status", "status", resource.status).
-			WithTag("json", "omitempty")
-		result = append(result, statusProperty)
+		result = append(result, resource.createStatusProperty())
 	}
 
 	return result
```
babbageclunk, line 330: Shouldn't `Properties()` return the non-spec/status properties too?

babbageclunk (approved):
It's nice to review a couple of PRs with so few files! Lately everything seems to have 50 changed files.

This looks good but I think there's an oversight in the `Properties()` method.

hack/generator/pkg/astmodel/resource_type.go
theunrepentantgeek: And thus we prove that I need to write the test even for code I think I can't possibly get wrong! 😊

theunrepentantgeek: Fixed.

hack/generator/pkg/astmodel/resource_type.go
```diff
@@ -24,6 +24,7 @@ type ResourceType struct {
 	status           Type
 	isStorageVersion bool
 	owner            *TypeName
+	properties       map[PropertyName]*PropertyDefinition
```
matthchr, line 27: @theunrepentantgeek - Sorry I missed looking at this PR on my Friday, but why do we need this change for the storage versions? The property doesn't have to be on the top level resource type I don't think. It really should be in spec in the storage version too, shouldn't it?

Maybe we can sync up offline or in the meeting today?

PR #1559, opened 2021-06-10
Add AZURE_TARGET_NAMESPACES to restrict the namespaces the operator watches

babbageclunk:
Add a new configuration value AZURE_TARGET_NAMESPACES which restricts the operator to only reconciling Azure resources within the listed namespaces (comma-separated).
This enables running multiple operators that each watch their own set of namespaces and work with different credentials.

Testing is a pain for settings like this - I've used the same approach as the secret-naming version testing for now, but we need to come up with a better way to do this. The next part of the multitenancy work is a setting to put the operator into webhook-only mode, so that will have the same problem. 

gif

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -100,7 +100,7 @@ steps:
         arch=$(go env GOARCH)
         go mod download
         make install-kubebuilder
-        make install-test-dependencies
+        make install-test-tools
         make generate-test-certs
       workingDirectory: '$(System.DefaultWorkingDirectory)'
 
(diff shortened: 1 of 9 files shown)


theunrepentantgeek (approved):
Looks good.

azure-pipelines.yml
```diff
@@ -156,6 +156,26 @@ steps:
       BUILD_ID: $(Build.BuildId)
     workingDirectory: '$(System.DefaultWorkingDirectory)'
 
+  # TODO: There is no way to run steps in parallel in Azure pipelines but ideally this step would run in parallel
+  # TODO: with the above testing step to reduce overall runtime
+  - script: |
```
matthchr, line 161: Maybe we should just make a separate `envtest` context, and set it up with different config options, and then run the two of those in parallel?

controllers/targetnamespace_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+	newName := func() string {
+		return "storageacct" + helpers.RandomString(6)
+	}
+
+	createNamespaces(ctx, t, "watched", "unwatched")
+
+	configuredNamespaces := envy.Get("AZURE_TARGET_NAMESPACES", "")
+
+	instance := v1alpha1.StorageAccount{
```
matthchr, line 43: rather than `instance` and `instance2`, instead use more clear names like `instanceDefaultNamespace` and `instanceWatchedNamespace` ?

pkg/resourcemanager/config/env.go
```diff
@@ -142,3 +144,16 @@ func ParseIntFromEnvironment(variable string) (int, error) {
 	}
 	return value, nil
 }
+
+func ParseStringListFromEnvironment(variable string) []string {
```
matthchr, line 148: Missing a unit test for this?

main.go
```diff
@@ -122,9 +123,25 @@ func main() {
...
+		setupLog.Error(err, "unable to parse settings required to provision resources in Azure")
+		os.Exit(1)
+	}
+
+	setupLog.V(0).Info("Configuration details", "Configuration", resourcemanagerconfig.ConfigString())
+
+	targetNamespaces := resourcemanagerconfig.TargetNamespaces()
+	var cacheFunc cache.NewCacheFunc
+	if targetNamespaces != nil {
+		setupLog.V(0).Info("Restricting operator cache", "namespaces", targetNamespaces)
```
matthchr, line 137: minor: This is technically pretty much duplicate as it gets logged by `ConfigString()` anyway?

It might be worth keeping just because it adds some nice explicitness/clarity for customers looking at logs, but I don't see any extra value we'd be gaining from it over and above the log above on line 132

matthchr (approved):

azure-pipelines.yml
babbageclunk: We talked through some options for this but won't do them in this PR.

main.go
babbageclunk: Yeah, I think that was really just for me while I was implementing it - removed!

babbageclunk:
Hmm, that doesn't seem good. Going to try one more time, otherwise I'm going to have to work something else out tomorrow.

babbageclunk:
Yay, thanks for merging @matthchr!

PR #1562, opened 2021-06-14
Simplify Property Conversion Context

theunrepentantgeek:

Tidies up use of the `PropertyConversionContext`, making it impossible to pass the wrong conversion direction into a `PropertyAssignmentFunction`. Also renames the factory method to `NewPropertyConversionContext`.

gif


babbageclunk (approved):
Nice.

PR #1566, opened 2021-06-14
Remove dangerous format specifiers

### LINKED ISSUE #1470: Review use of Format Specifiers for potential information leakage
The Go format specifier `%v` (and variations) can dump an entire struct into a log message, potentially resulting in information leakage (e.g. if the struct contains an API key or other secret).

We should identify which format specifiers might cause this kind of leakage and ensure we're only using them where it's safe and there's no alternative.

theunrepentantgeek:

Removes all uses of the format specifiers `%v` and `%+v` as those dump entire objects into the log, which very likely will include secrets (e.g. credentials etc); we don't want those exposed in the logs.

Closes #1470

gif

### DIFF
--- a/hack/generated/pkg/genruntime/errors_test.go
+++ b/hack/generated/pkg/genruntime/errors_test.go
@@ -6,6 +6,7 @@ Licensed under the MIT license.
 package genruntime
 
 import (
+	"strings"
 	"testing"
 
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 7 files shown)


hack/generated/pkg/genruntime/errors_test.go
```diff
@@ -112,7 +112,13 @@ func TestOwnerNotFound_RemembersCause(t *testing.T) {
 
 	g.Expect(errors.Cause(err)).To(Equal(cause))
 
-	fmtedErr := fmt.Sprintf("%+v", err)
+	var builder strings.Builder
+	for err != nil {
```
babbageclunk, line 116: I don't think this loop is necessary - the wrapper error should include the context of the errors it wraps. The stacktrace that's needed for the error line assertions can be retrieved using [this technique](https://pkg.go.dev/github.com/pkg/errors?utm_source=godoc#hdr-Retrieving_the_stack_trace_of_an_error_or_wrapper).

babbageclunk (approved):

hack/generated/pkg/genruntime/errors_test.go
theunrepentantgeek: The `ReferenceNotFound` error wasn't including the nested context, but now it does. I'll query the stack trace separately.

theunrepentantgeek: Done.

Porges (approved):

babbageclunk (approved):

matthchr (approved):

PR #1567, opened 2021-06-16
Create Test Package

theunrepentantgeek:

Eliminates most duplication of test support infrastructure, allowing reuse across multiple packages. 

Doesn't eliminate all copies of `FakeFunction` as we can't move the uses within `astmodel` out without causing a package reference cycle.

Will be making more reuse of methods in an upcoming PR, testing functions.

gif


hack/generator/pkg/test/fake_function.go
```diff
@@ -13,12 +13,14 @@ import (
...
 type FakeFunction struct {
-	name       string
-	Imported   *astmodel.PackageReferenceSet
-	Referenced astmodel.TypeNameSet
-	returnType astmodel.Type
-	idFactory  astmodel.IdentifierFactory
+	name         string
+	Imported     *astmodel.PackageReferenceSet
+	Referenced   astmodel.TypeNameSet
+	TypeReturned astmodel.Type
```
babbageclunk, line 22: super minor: I have a slight preference for ReturnType here. 
Oh, I see that you did this to make it consistent with Imported and References. Present tense (Imports, References) feels more natural for those too, and I quite like that they can also be read as plural nouns as well since they're sets. 🤷

babbageclunk (approved):
Looks good!

hack/generator/pkg/test/fake_function.go
theunrepentantgeek: Also, the function `ReturnType()` is on the `Function` interface, so I couldn't use `ReturnType` for the member name here.

PR #1568, opened 2021-06-16
Add Disk (from microsoft.compute/20200930) to generated resources

babbageclunk:
This enables the Disk resource in code generation and adds a test that exercises creating, updating and deleting it.

The test framework is really nice now!

gif

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/hack/generated/controllers/crd_disk_test.go
+++ b/hack/generated/controllers/crd_disk_test.go
@@ -1,0 +1,77 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+package controllers_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	compute "github.com/Azure/azure-service-operator/hack/generated/_apis/microsoft.compute/v1alpha1api20200930"
+	"github.com/Azure/azure-service-operator/hack/generated/pkg/testcommon"
+)
+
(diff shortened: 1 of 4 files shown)


hack/generated/controllers/crd_disk_test.go
```diff
@@ -0,0 +1,77 @@
...
+func Test_Disk_CRUD(t *testing.T) {
+	t.Parallel()
+
+	tc := globalTestContext.ForTest(t)
+
+	rg := tc.CreateNewTestResourceGroupAndWait()
+
+	// Create a disk.
+	standardSkuName := compute.DiskSkuNameStandardLRS
+	size := 500
```
matthchr, line 27: minor: `sizeInGb`?

matthchr (approved):
I feel like these always look so clean!

@babbageclunk How was the experience, were there parts that were confusing/hard to understand? Did you look over the generated resource and see if it looked "reasonable"?

babbageclunk:
> How was the experience, were there parts that were confusing/hard to understand? Did you look over the generated resource and see if it looked "reasonable"?

Pretty straightforward, nothing difficult (might have been lucky with the resource, no weird warts on this one). I did look through the resource, it seemed sane, although lots of pointers which make it a little fiddly to construct types sometimes. One thing I think would be nicer in the generated output would be to split the files up differently. At the moment the types are lost in the surrounding ARM and storage conversion functions - it might be better to have the types in one file, all the conversion functions in another, maybe webhooks in another (although there's not that much of them at the moment).

PR #1570, opened 2021-06-16
Fix generated comments to reflect azure-service-operator

theunrepentantgeek:

Our generated code had comments indicating the source was `k8s-infra` and `k8s-infra-gen` - we change these to read `azure-service-operator` while retaining the ability to detect and clean up generated code from older versions.

gif


hack/generator/pkg/astmodel/generation_consts.go
```diff
@@ -8,7 +8,14 @@ package astmodel
 var (
 	CodeGenerationComments []string = []string{
 		// Note that this format is actually an official specification in go:  https://github.com/golang/go/issues/13560
-		"// Code generated by k8s-infra. DO NOT EDIT.", // TODO: Update this when the generated is moved/renamed
+		"// Code generated by azure-service-operator. DO NOT EDIT.", // TODO: Update this when the generated is moved/renamed
```
babbageclunk, line 11: Maybe remove the TODO? Although I guess we still want to update it if it's moved again.

hack/generator/pkg/astmodel/generation_consts.go
```diff
@@ -8,7 +8,14 @@ package astmodel
...
+	}
+
+	// Required for backward compatibility.
+	// Presence of any one string will result in the file being treated as a generated file that we are responsible for
+	CodeGenerationCommentsDeprecated []string = []string{
```
matthchr, line 16: Is the plan to delete this later?

Feels like we should maybe just delete this now and tell the 3 or 4 of us to manually clean up the directory once when we pull? Cleaner than having legacy here before we're even released?

matthchr (approved):

babbageclunk (approved):

hack/generator/pkg/astmodel/generation_consts.go
matthchr: I agree - remove TODO. I think the TODO was in reference to what @theunrepentantgeek just did (i.e. that it shouldn't say k8sinfra)

matthchr, line 11: IMO we should differentiate between ASO and the tool that generates aso.

Maybe this is ASO-gen or something?

theunrepentantgeek: Nah, that TODO has been there for ages. I'll remove it though.

hack/generator/pkg/astmodel/generation_consts.go
theunrepentantgeek: Yeah, didn't take a lot to persuade me. :grin:

hack/generator/pkg/astmodel/generation_consts.go
theunrepentantgeek: Changed to `azure-service-operator-codegen`.

PR #1574, opened 2021-06-17
Provide support for property conversions of JSON properties

theunrepentantgeek:

We have properties of type `v1.JSON` but they were not supported by the property conversions needed for storage variants.

At the same time, we generalized two existing property converters to allow support for known types to be easily added.

gif

**If applicable**:
- [x] this PR contains tests


hack/generator/pkg/astmodel/std_references.go
```diff
@@ -30,6 +30,7 @@ var (
 	ControllerRuntimeAdmission   = MakeExternalPackageReference("sigs.k8s.io/controller-runtime/pkg/webhook/admission")
 	GitHubErrorsReference        = MakeExternalPackageReference("github.com/pkg/errors")
 
+
```
matthchr, line 33: minor: Extra newline here

matthchr (approved):

hack/generator/pkg/conversions/testdata/CopyKnownReferenceProperty.golden
```diff
@@ -0,0 +1,33 @@
...
+// Licensed under the MIT license.
+package vCurrent
+
+import (
+	"github.com/Azure/azure-service-operator/hack/generated/_apis/Verification/vNext"
+	"github.com/Azure/azure-service-operator/hack/generated/pkg/genruntime"
+)
+
+type Person struct {
+	KnownReference genruntime.KnownResourceReference `json:"known-reference"`
```
matthchr, line 12: minor: This should probably be `knownReference` in the JSON (even though it doesn't matter at all)

hack/generator/pkg/astmodel/std_references.go
theunrepentantgeek: Fixed by `go fmt`

PR #1577, opened 2021-06-18
Config transformers and filters now error if no matches

matthchr:
  - Ensure that all config transformers and filters
    raise an error by default if they don't match anything.
  - Removed config options that had no matches.


matthchr:
I got tired of fighting with it for VMSS (where I'm mucking with the config quite a bit to make sure the resource comes out looking good).

theunrepentantgeek (approved):

PR #1579, opened 2021-06-19
Add new pipeline stage that merges groups

### LINKED ISSUE #1578: StorageTypeFactory assumes that types do not reference one another across group boundaries
Today, in the `createStorageTypes` pipeline stage, a `StorageTypeFactory` is created for each group, and is populated with all of the types in that group. The `types` collection built up for each `StorageTypeFactory` contains only types from that group.

This disallows any cross-group type references, which I _think_ is reasonable, except that there are instances of cross-group references - for example `microsoft.compute` refers to `microsoft.compute.extensions` - see the [VirtualMachineScaleSetExtension](https://schema.management.azure.com/schemas/2020-12-01/Microsoft.Compute.json) type.

**Describe the improvement**
Either `StorageTypeFactory` needs to tolerate this layout, or we need to come up with a stage that automatically does a graph-walk and moves things that have cross-group dependencies into mirrored types in the same group.

matthchr:
Fixes #1578

- In the case that group A references group B, we need to pull group B's types into group A.
- Also fixes a bug in the Crossplane golden files tests which combined with this new pipeline stage could cause all types to get pruned.

### DIFF
--- a/hack/generator/pkg/codegen/pipeline_collapse_cross_group_refs.go
+++ b/hack/generator/pkg/codegen/pipeline_collapse_cross_group_refs.go
@@ -1,0 +1,57 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package codegen
+
+import (
+	"context"
+
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/astmodel"
+	"github.com/pkg/errors"
+)
+
+// collapseCrossGroupReferences finds and removes references between API groups. This isn't particularly common
+// but does occur in a few instances, for example from Microsoft.Compute -> Microsoft.Compute.Extensions.
+func collapseCrossGroupReferences() PipelineStage {
+	return MakePipelineStage(
+		"collapseCrossGroupReferences",
+		"Finds and removes cross group references",
+		func(ctx context.Context, types astmodel.Types) (astmodel.Types, error) {
+			resources := astmodel.CollectResourceDefinitions(types)
+			result := make(astmodel.Types)
+
+			for resourceName := range resources {
+				walker := newTypeWalker(types, resourceName)
+				updatedTypes, err := walker.Walk(types[resourceName])
+				if err != nil {
+					return nil, errors.Wrapf(err, "failed walking types")
+				}
+
+				for _, newDef := range updatedTypes {
+					err := result.AddAllowDuplicates(newDef)
+					if err != nil {
+						return nil, err
+					}
+				}
+			}
+
+			return result, nil
+		})
+}
+
+func newTypeWalker(types astmodel.Types, resourceName astmodel.TypeName) *astmodel.TypeWalker {
+	visitor := astmodel.TypeVisitorBuilder{}.Build()
+	walker := astmodel.NewTypeWalker(types, visitor)
+	walker.AfterVisit = func(original astmodel.TypeDefinition, updated astmodel.TypeDefinition, ctx interface{}) (astmodel.TypeDefinition, error) {
+		if !resourceName.PackageReference.Equals(updated.Name().PackageReference) {
+			// Note: If we ever find this generating colliding names, we might need to introduce a unique suffix.
+			// For now though it doesn't seem to, so preserving the shorter names as they're clearer.
+			updated = updated.WithName(astmodel.MakeTypeName(resourceName.PackageReference, updated.Name().Name()))
+		}
+		return astmodel.IdentityAfterVisit(original, updated, ctx)
+	}
+
+	return walker
+}
--- a/hack/generator/pkg/codegen/code_generator.go
+++ b/hack/generator/pkg/codegen/code_generator.go
@@ -119,6 +119,9 @@ func createAllPipelineStages(idFactory astmodel.IdentifierFactory, configuration
 		// Strip out redundant type aliases:
 		removeTypeAliases(),
 
+		// Collapse cross group references
+		collapseCrossGroupReferences(),
+
 		// De-pluralize resource types
 		// (Must come after type aliases are resolved)
 		improveResourcePluralization().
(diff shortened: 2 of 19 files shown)


hack/generator/pkg/codegen/pipeline_collapse_cross_group_refs.go
```diff
@@ -0,0 +1,57 @@
...
+				if err != nil {
+					return nil, errors.Wrapf(err, "failed walking types")
+				}
+
+				for _, newDef := range updatedTypes {
+					err := result.AddAllowDuplicates(newDef)
+					if err != nil {
+						return nil, err
+					}
+				}
```
theunrepentantgeek, line 37: I think you could replace this loop with the new `OverlayWith()` method.

hack/generator/pkg/codegen/pipeline_collapse_cross_group_refs.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+// collapseCrossGroupReferences finds and removes references between API groups. This isn't particularly common
+// but does occur in a few instances, for example from Microsoft.Compute -> Microsoft.Compute.Extensions.
+func collapseCrossGroupReferences() PipelineStage {
+	return MakePipelineStage(
+		"collapseCrossGroupReferences",
+		"Finds and removes cross group references",
+		func(ctx context.Context, types astmodel.Types) (astmodel.Types, error) {
+			resources := astmodel.CollectResourceDefinitions(types)
```
theunrepentantgeek, line 22: Is this deterministing in sort order of the results? If not, or if the return type is a map, then sometimes we might merge `Microsoft.Compute` into `Microsoft.Compute.Extensions` intead of the other way around.

theunrepentantgeek (changes requested):
One question that we need to answer before merging, otherwise good.

hack/generator/pkg/codegen/pipeline_collapse_cross_group_refs.go
matthchr: A good point - will improve. Although in the case of your example Microsoft.Compute always references Microsoft.Compute.Extensions, which doesn't itself reference Microsoft.Compute I don't believe (there's not a cycle). But there could be one so I'll ensure we handle it.

matthchr: Oh wait actually what we're doing now is ok: If there _were_ a cycle then the stuff in Microsoft.Compute which Microsoft.Compute.Extensions referenced would get copied into Microsoft.Compute.Extensions, and the stuff in Microsoft.Compute.Extensions which was referenced by Microsoft.Compute will be moved into Microsoft.Compute.

So the net is in a full cycle that both packages will contain everything in the other package - which is the desired behavior so that we're not crossing groups for references.

hack/generator/pkg/codegen/pipeline_collapse_cross_group_refs.go
matthchr: I'm not finding such a method?

hack/generator/pkg/codegen/pipeline_collapse_cross_group_refs.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func newTypeWalker(types astmodel.Types, resourceName astmodel.TypeName) *astmodel.TypeWalker {
+	visitor := astmodel.TypeVisitorBuilder{}.Build()
+	walker := astmodel.NewTypeWalker(types, visitor)
+	walker.AfterVisit = func(original astmodel.TypeDefinition, updated astmodel.TypeDefinition, ctx interface{}) (astmodel.TypeDefinition, error) {
+		if !resourceName.PackageReference.Equals(updated.Name().PackageReference) {
+			// Note: If we ever find this generating colliding names, we might need to introduce a unique suffix.
+			// For now though it doesn't seem to, so preserving the shorter names as they're clearer.
+			updated = updated.WithName(astmodel.MakeTypeName(resourceName.PackageReference, updated.Name().Name()))
```
babbageclunk, line 51: Presumably in the case of a collision we'll panic rather than just overwriting?

babbageclunk (approved):
Pretty neat use of TypeWalker!

hack/generator/pkg/codegen/pipeline_collapse_cross_group_refs.go
matthchr: A collision would be `err := result.AddAllowDuplicates(newDef)` above returning an error - meaning we had two types that were named the same but are structurally different. Right now that'd just block the whole pipeline and we'd have to come and add handling to deal with it.

hack/generator/pkg/codegen/pipeline_collapse_cross_group_refs.go
theunrepentantgeek: Makes sense.

hack/generator/pkg/codegen/pipeline_collapse_cross_group_refs.go
theunrepentantgeek: SEe PR #1598

theunrepentantgeek (approved):

hack/generator/pkg/codegen/pipeline_collapse_cross_group_refs.go
matthchr: Resolving this as there was agreement that it is correct over Teams.

hack/generator/pkg/codegen/pipeline_collapse_cross_group_refs.go
matthchr: I went to do this, but realized that `OverlayWith` doesn't have the `AllowDuplicates` behavior. I don't think we want to just pick a winner in the case of a collision (which is what `OverlayWith` does) as that is making an implicit assumption that the loser has the same structure as the winner. The loop as written ensures that in cases where there is a collision, the two types are identical and thus can be treated the same.

theunrepentantgeek: Fair enough.

PR #1595, opened 2021-06-23
Fix Storage Conversion generation bug

matthchr:
If a map type is a TypeName that is not pointing to a local package, just use the type as is rather than trying to look it up (which will    fail).


hack/generator/pkg/conversions/property_conversions.go
```diff
@@ -971,6 +971,13 @@ func copyKnownType(name astmodel.TypeName, methodName string) func(*TypedConvers
 		}
 
 		return func(reader dst.Expr, writer func(dst.Expr) []dst.Stmt, generationContext *astmodel.CodeGenerationContext) []dst.Stmt {
+			if methodName == "DeepCopy" {
```
matthchr, line 974: Open to a cleaner way to do this - a parameter to the function indicating if we should dereference or not?

hack/generator/pkg/codegen/storage/property_converter.go
```diff
@@ -90,6 +91,11 @@ func (p *PropertyConverter) useBaseTypeForEnumerations(
 func (p *PropertyConverter) shortCircuitNamesOfSimpleTypes(
 	tv *astmodel.TypeVisitor, tn astmodel.TypeName, ctx interface{}) (astmodel.Type, error) {
 
+	// for nonlocal packages, preserve the name as is
+	if _, ok := tn.PackageReference.AsLocalPackage(); !ok {
```
matthchr, line 95: We don't generate the storage types in the Golden files tests right now. Once we do, this issue would be caught there by the existing golden file test that has a variety of `v1.JSON` types.

Given that, I didn't try to add a test specifically for this. Once we enable storage type generation in golden files we should be covered.

hack/generator/pkg/conversions/property_conversions.go
theunrepentantgeek: I don't think this will generate the right code. 

For one, need to store the value in a local to dereference it, and for another if this is being assigned to an optional property, it will be wrapped in AddressOf (`&`) giving some really odd results.

hack/generator/pkg/conversions/testdata/CopyJSONProperty.golden
```diff
@@ -16,7 +16,7 @@ type Person struct {
 func (person *Person) AssignPropertiesFromPerson(source *vNext.Person) error {
 
 	// JSONBlob
-	person.JSONBlob = source.JSONBlob.DeepCopy()
+	person.JSONBlob = *source.JSONBlob.DeepCopy()
```
theunrepentantgeek, line 19: I don't think this compiles.

And, if `person.JSONBlob` was optional, we'd generate this:

``` go
person.JSONBlob = &*source.JSONBlob.DeepCopy()
```
which also won't compile.

theunrepentantgeek (changes requested):

hack/generator/pkg/conversions/property_conversions.go
matthchr: We don't need to store the value in a local to dereference it -- you just need a local for making a pointer (not for dereferencing one), see: https://play.golang.org/p/0Oc53WcpnJr

I noted below in the comment: 
> // This dereference is always safe because we ensured that both source and destination are always
> // non optional

Is this not correct? Just because it _looks required_ here doesn't necessarily mean the final destination is required?

In the case of VMSS, here's the type in question that has the `v1.JSON`:
```
//Generated from: https://schema.management.azure.com/schemas/2020-12-01/Microsoft.Compute.Extensions.json#/definitions/genericExtension
type GenericExtension struct {
	// +kubebuilder:validation:MinLength=1
	//Publisher: Microsoft.Compute/extensions - Publisher
	Publisher string `json:"publisher"`

	//Settings: Microsoft.Compute/extensions - Settings
	Settings map[string]v1.JSON `json:"settings"`

	// +kubebuilder:validation:MinLength=1
	//Type: Microsoft.Compute/extensions - Type
	Type string `json:"type"`

	// +kubebuilder:validation:MinLength=1
	//TypeHandlerVersion: Microsoft.Compute/extensions - Type handler version
	TypeHandlerVersion string `json:"typeHandlerVersion"`
}
```

and here's the generated method that this code helps make:
```
        // Settings
	settingMap := make(map[string]v1.JSON)
	for settingKey, settingValue := range genericExtension.Settings {
		// Shadow the loop variable to avoid aliasing
		settingValue := settingValue
		settingMap[settingKey] = *settingValue.DeepCopy()
	}
	destination.Settings = settingMap
```

Is the real issue here that when dealing with a `v1.JSON` it should end up as `*v1.JSON` on the storage type?

theunrepentantgeek: The conversion pipeline works by unwrapping/simplifying endpoints - so optional properties are simplified to required properties, and then rewrapped once the underlying code for the required value is generated.

The required invariant is that the `reader` and `writer` have to be generated code that matches the type specified of the matching endpoint. I _think_ you have maintained that, but will add some tests to verify.

hack/generator/pkg/codegen/storage/property_converter.go
theunrepentantgeek: I think we should be able to add a test - the test coverage of the conversions package is intended to catch issues here, rather than forcing a more involved analysis if/when it fails as a part of storage conversion.

matthchr:
Thanks @theunrepentantgeek for the push to the branch with additional tests and some extra fixes

hack/generator/pkg/codegen/storage/property_converter.go
theunrepentantgeek: Have added a test - which found some interesting edge cases we needed to handle.

hack/generator/pkg/conversions/testdata/CopyJSONProperty.golden
theunrepentantgeek: Added tests to check, and it looks fine.

hack/generator/pkg/conversions/property_conversions.go
matthchr: Resolving this as you did add some tests and fixed some corner cases that worked but were ugly.

theunrepentantgeek (approved):

PR #1596, opened 2021-06-23
Some small test improvements

matthchr:
  - Move region to WestUS2 since some resources have a default quota of
    0 in West US (VMs in particular).
  - Wait for resource group deletion at the end of each test (this is important
    for reproducibility as well as making sure we don't leak resources.
  - Re-record tests.

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/hack/generated/pkg/testcommon/error_translating_roundtripper.go
+++ b/hack/generated/pkg/testcommon/error_translating_roundtripper.go
@@ -76,20 +76,20 @@ func (w errorTranslation) RoundTrip(req *http.Request) (*http.Response, error) {
 	matchingBodies := w.findMatchingBodies(req)
 
 	if len(matchingBodies) == 0 {
-		w.t.Logf("\n*** Cannot find go-vcr recording for request (no responses recorded for this method/URL): %s %s (attempt: %s)\n\n", req.Method, req.URL.String(), req.Header.Get(COUNT_HEADER))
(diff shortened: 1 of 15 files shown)


Porges (approved):

hack/generated/pkg/testcommon/error_translating_roundtripper.go
```diff
@@ -76,20 +76,20 @@ func (w errorTranslation) RoundTrip(req *http.Request) (*http.Response, error) {
...
-		w.t.Logf("\n*** Cannot find go-vcr recording for request (no responses recorded for this method/URL): %s %s (attempt: %s)\n\n", req.Method, req.URL.String(), req.Header.Get(COUNT_HEADER))
+		w.t.Fatalf("\n*** Cannot find go-vcr recording for request (no responses recorded for this method/URL): %s %s (attempt: %s)\n\n", req.Method, req.URL.String(), req.Header.Get(COUNT_HEADER))
 		return nil, originalErr
 	}
 
 	// locate the request body with the shortest diff from the sent body
 	shortestDiff := ""
 	for i, bodyString := range matchingBodies {
-		diff := cmp.Diff(sentBodyString, bodyString)
+		diff := cmp.Diff(bodyString, sentBodyString)
```
matthchr, line 86: I think we had this backwards before: `cmp.Diff(x, y)` shows `y-x`, so changes in `y` that are not in `x` are shown with a `+`.

I think that means we want `y` to be the "new" body and `x` to be the "old" body.

Porges: Makes sense 🙂

PR #1600, opened 2021-06-24
Helm Chart: Added support for using existing secret for Azure Service Operator

ErikLundJensen:
Update of the Helm Chart.

In AKS cluster the secrets come from Azure Key Vault using for example External-Secrets. This pull-request makes it possible to use an existing secret instead of always creating a azure-service-operator secret.


matthchr (approved):
LGTM other than a minor comment

PR #1606, opened 2021-06-26
Persist original ARM API version for use by generic controller

theunrepentantgeek:

Adds `OriginalVersion()` function to API Spec types to return the actual ARM API version requested by users. 
Adds `OriginalVersion` property to Storage Spec types to store that version for later use.
Adds `OriginalGVK()` function to Storage Resource types to return the GVK of the original version.
Removes `StorageTypeFactory` in favour of separate pipeline sections

Also includes
- Unit tests for `OriginalVersion()` function generation
- Unit tests for `OriginalGVK()` function generation
- Fix bug in prerequisite checks for pipeline stages
- Fix bug in `ErroredType` where `WriteDebugDescription()` would crash if there was no wrapped type

gif

**If applicable**:
- [x] this PR contains tests


hack/generator/pkg/codegen/storage/storage_type_factory.go
```diff
@@ -32,25 +35,43 @@ func NewStorageTypeFactory(group string, idFactory astmodel.IdentifierFactory) *
...
+	for _, def := range definitions {
+		f.referenceTypes.Add(def)
+
+		if rt, ok := astmodel.AsResourceType(def.Type()); ok {
+			// We have a resource type
+			if tn, ok := astmodel.AsTypeName(rt.SpecType()); ok {
+				// Keep track of all our spec types
+				f.specTypes = f.specTypes.Add(tn)
+			}
+			if tn, ok := astmodel.AsTypeName(rt.StatusType()); ok {
```
matthchr, line 63: minor: Would a resource without a spec sneaking in here cause problems? Would it be worth failing-fast in that case (could `panic` if you don't want to pollute this method signature?).

Ignore this if there's no issue and a spec-less resource isn't going to blow up later in this stage.

hack/generator/pkg/codegen/storage/storage_type_factory.go
```diff
@@ -61,38 +82,32 @@ func (f *StorageTypeFactory) Types() (astmodel.Types, error) {
 	return f.outputTypes, nil
 }
 
+// process carries out our transformations
+// Each step reads from outputTypes and puts the results back in there
 func (f *StorageTypeFactory) process() error {
 
-	f.outputTypes = make(astmodel.Types)
+	f.outputTypes = f.referenceTypes.Copy()
```
matthchr, line 89: minor: Here with the `outputTypes` nomenclature I kinda liked the `inputTypes` which matched up?

hack/generator/pkg/codegen/storage/storage_type_factory.go
```diff
@@ -61,38 +82,32 @@ func (f *StorageTypeFactory) Types() (astmodel.Types, error) {
...
 
-	// Create storage variants and stash them in outputTypes so injectConversions can find them
-	storageVariants, err := f.inputTypes.Process(f.createStorageVariant)
+	// Inject OriginalVersion() methods into all our spec types
+	modifiedSpecTypes, err := f.outputTypes.Process(f.injectOriginalVersionMethod)
 	if err != nil {
 		return err
 	}
+	f.outputTypes = f.outputTypes.OverlayWith(modifiedSpecTypes)
```
matthchr, line 96: Same question here: Previously you had `AddTypes` which ensure that there was no collision (it panics if there was). Now you don't and so a collision is silently dropped. If there's an expectation that the `modifiedSpecTypes` doesn't trample something existing in `outputTypes` then that `panic` seems to be asserting a useful property and we might not want to lose it?

matthchr: Maybe it's moot because there's not ever anything in `outputTypes` to start with anyway? Even if that's the case I think the clarity of having the assert (and risk of failure) is better in case of future refactors that added a step before this and this made it possible for there to be something in `outputTypes` when you get here.

hack/generator/pkg/codegen/pipeline/create_storage_types.go
```diff
@@ -61,16 +61,14 @@ func CreateStorageTypes(idFactory astmodel.IdentifierFactory) Stage {
...
 
 			err := kerrors.NewAggregate(errs)
 			if err != nil {
 				return nil, err
 			}
 
-			unmodified := types.Except(result)
-			result.AddTypes(unmodified)
-			return result, nil
+			return types.OverlayWith(storageTypes), nil
```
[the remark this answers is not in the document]
matthchr: After re-reading the `process()` function I do indeed see that there is overlap and that we want the `storageTypes` to win because we've modified some existing `Spec` types to have an additional method.

Will leave the above here not because it's correct but because I was initially confused... wondering if a comment or a variable name change might help?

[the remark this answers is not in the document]
matthchr: Actually further reading the code I don't quite follow how `storageTypes` isn't just a superset of `types`? In which case there's no need for overlay you can just return `storageTypes`?

hack/generator/pkg/codegen/storage/storage_type_factory.go
```diff
@@ -161,31 +196,56 @@ func (f *StorageTypeFactory) injectConversions(definition astmodel.TypeDefinitio
...
+// downstreamDef is the definition closer to our hub type in our directed conversion graph
+func (f *StorageTypeFactory) injectConversionsBetween(
+	upstreamDef astmodel.TypeDefinition, downstreamDef astmodel.TypeDefinition) (*astmodel.TypeDefinition, error) {
+
 	// Create conversion functions
-	knownTypes := make(astmodel.Types)
-	knownTypes.AddTypes(f.inputTypes)
-	knownTypes.AddTypes(f.outputTypes)
+	knownTypes := f.referenceTypes.Copy()
+	knownTypes = knownTypes.OverlayWith(f.outputTypes)
```
matthchr, line 210: This is throwing me for a loop a bit. The call chain is:
1. `f.outputTypes.Process(f.injectConversions)` line 106
2. `f.injectConversionsBetween(definition, nextDef)` line 199

and inside `injectConversionsBetween` we then build a set of `knownTypes` (which is `f.referenceTypes + f.outputTypes`) which is then provided to `PropertyConversionContext` for the purposes of looking up types for the conversion.

I understand the objective (build a `knownTypes` collection that we use for lookup), but how it's being done is confusing me a bit:
1. We recompute this set for each invocation of `f.injectConversions`, so we're O(N^2).
2. Since this is reusing the same `f.outputTypes` in various locations and functions you have to pay close attention to temporally where this is called "what is in `outputTypes` at the time this function is invoked"? This took me a second to understand and would be cleaner IMO if the set in question had just been a parameter rather than a struct variable that's constantly being mutated.

I think r.e. (2) above that I had raised a similar comment in a previous PR and you pointed out that some of that was required because of the signature of `Process`. I'm wondering if the usage of `Process` is helping or hurting our complexity here. I can't quite imagine what the code would look like more purely-functional but I'm wondering if the few extra lines of loops would be offset by clearer method signatures and lack of `output` reuse. Bonus that it would also make it easy to just pass in the `knownTypes` from above rather than recomputing it for each element.

matthchr: Technically you can dodge the O(N^2) problem by just precomputing `knownTypes` and stashing it on `f` but I think that leads us further down the "state mutation happening at different times is critical to the code" and also leads to a smell where you have a struct variable that really exists to pass a parameter to a method (and is useless before and after that call)

matthchr (commented):
Not done yet but posting what I have

hack/generator/pkg/astmodel/object_type.go
```diff
@@ -240,7 +240,10 @@ func (objectType *ObjectType) References() TypeNameSet {
 		results.AddAll(property.PropertyType().References())
 	}
 
-	// Not collecting types from functions deliberately.
```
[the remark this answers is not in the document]
theunrepentantgeek: Nobody can remember - and there don't seem to be any untoward side effects. 

Elmo

hack/generator/pkg/astbuilder/composite_literals.go
```diff
@@ -0,0 +1,48 @@
...
+// structType is an expression to handle both structs from the current package and imported ones requiring qualification
+func NewCompositeLiteralDetails(structType dst.Expr) *CompositeLiteralDetails {
+	return &CompositeLiteralDetails{
+		structType: structType,
+	}
+}
+
+// AddField adds initialization of another field
+// Returns the receiver to allow method chaining when desired
+func (details *CompositeLiteralDetails) AddField(name string, value dst.Expr) *CompositeLiteralDetails {
```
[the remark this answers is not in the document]
theunrepentantgeek: I'm on the fence, for the reasons you stated and others. Changed to the imperative form.

hack/generator/pkg/codegen/pipeline/create_storage_types.go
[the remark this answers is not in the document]
theunrepentantgeek: I've collapsed the two sets into a single one just called `types`, and I've added more documentation to `process()` explaining what's going on.

hack/generator/pkg/codegen/storage/storage_type_factory.go
```diff
@@ -11,13 +11,16 @@ import (
...
 	"github.com/Azure/azure-service-operator/hack/generator/pkg/conversions"
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/functions"
 )
 
-// StorageTypeFactory is used to create storage inputTypes for a specific api group
+// StorageTypeFactory is used to create storage types for a specific api group
 type StorageTypeFactory struct {
 	group             string                                                  // Name of the group we're handling (used mostly for logging)
-	inputTypes        astmodel.Types                                          // All the types for this group
+	referenceTypes    astmodel.Types                                          // All the types for this group
```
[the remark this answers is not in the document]
theunrepentantgeek: Yup. Done.

hack/generator/pkg/codegen/storage/storage_type_factory.go
theunrepentantgeek: Can we have a resource without a spec? What would that even mean? There'd be nothing to submit to ARM, right?

hack/generator/pkg/codegen/storage/storage_type_factory.go
theunrepentantgeek: Simplified.

hack/generator/pkg/codegen/storage/storage_type_factory.go
theunrepentantgeek: Let me know how you think this looks with the changes to collapse `referenceTypes` and `outputTypes` together.

hack/generator/pkg/codegen/storage/storage_type_factory.go
theunrepentantgeek: This is why I collapsed `referenceType` and `outputTypes` into the same set - there's no need to recompute it.

hack/generator/pkg/codegen/storage/storage_type_factory.go
matthchr: > Can we have a resource without a spec?

Without a bug, no I don't think we can. My question here was basically, you're doing these two checks for spec and status and if somehow there was no spec and no status this loop wouldn't error it'd just skip adding anything - in other places I've seen this code there's a `!ok` case on the spec path that blows up if there's no spec. Did you want that here, or is it really not important in this context and if we wanted to assert on that we should do it elsewhere elsewhere?

I'm fine with either answer.

hack/generator/pkg/codegen/storage/storage_type_factory.go
matthchr: This bit looks good to me now

hack/generator/pkg/codegen/storage/storage_type_factory.go
```diff
@@ -154,38 +187,61 @@ func (f *StorageTypeFactory) injectConversions(definition astmodel.TypeDefinitio
...
 	if !ok {
 		// No next type so nothing to do
 		// (this is expected if the type is discontinued or we're looking at the hub type)
 		return nil, nil
 	}
 
+	return f.injectConversionsBetween(definition, nextDef)
+}
+
+// inject conversion methods between the two specified definitions
```
matthchr, line 200: minor: 
```suggestion
// injectConversionsBetween injects conversion methods between the two specified definitions
```

hack/generator/pkg/codegen/storage/testdata/Test_StorageTypeFactory_GeneratesExpectedResults-v20211231storage.golden
```diff
@@ -0,0 +1,54 @@
...
+	Items           []v20211231storage.Person `json:"items"`
+}
+
+//Storage version of v20211231.Person_Spec
+type Person_Spec struct {
+	CustomerProgram *string `json:"customerProgram,omitempty"`
+	FamilyName      *string `json:"familyName,omitempty"`
+	FullName        *string `json:"fullName,omitempty"`
+	KnownAs         *string `json:"knownAs,omitempty"`
+	OriginalVersion string  `json:"original-version"`
```
matthchr, line 44: shouldn't this be `originalVersion` in the JSON?

hack/generator/pkg/codegen/storage/type_converter.go
```diff
@@ -40,11 +46,63 @@ func NewTypeConverter(types astmodel.Types) *TypeConverter {
 	return result
 }
 
-// ConvertDefinition applies our type conversion to a specific type definition
-func (t *TypeConverter) ConvertDefinition(def astmodel.TypeDefinition) (astmodel.TypeDefinition, error) {
-	result, err := t.visitor.VisitDefinition(def, nil)
+// ConvertResourceDefinition applies our type conversion to a specific resource type definition
+func (t *TypeConverter) ConvertResourceDefinition(def astmodel.TypeDefinition) (astmodel.TypeDefinition, error) {
```
matthchr, line 50: It's not clear to me how this function and `ConvertObjectDefinition` are different? I don't see any resource specific handling here?

hack/generator/pkg/codegen/storage/type_converter.go
```diff
@@ -63,16 +121,22 @@ func (t *TypeConverter) convertResourceType(
 	resource *astmodel.ResourceType,
 	ctx interface{}) (astmodel.Type, error) {
 
+	// Storage resource needs OriginalGVP() function injected too
```
matthchr, line 124:
```suggestion
	// Storage resource needs OriginalGVK() function injected too
```

hack/generator/pkg/codegen/storage/type_converter.go
```diff
@@ -91,6 +155,13 @@ func (t *TypeConverter) convertObjectType(
 	}
 
 	objectType := astmodel.NewObjectType().WithProperties(properties...)
+
+	if converterContext.isSpec {
+		// Inject OriginalVersion property
+		originalVersion := astmodel.NewPropertyDefinition("OriginalVersion", "original-version", astmodel.StringType)
```
matthchr, line 161: Same comment as in the golden file, I think this should be `originalVersion` not `original-version`

hack/generator/pkg/functions/original_gvk_function.go
```diff
@@ -0,0 +1,89 @@
...
+
+// NewOriginalGVKFunction creates a new OriginalGVKFunction
+func NewOriginalGVKFunction(idFactory astmodel.IdentifierFactory) *OriginalGVKFunction {
+	return &OriginalGVKFunction{
+		idFactory: idFactory,
+	}
+}
+
+// Name returns the name of this function, which is always OriginalGVK()
+func (o OriginalGVKFunction) Name() string {
```
matthchr, line 42: I think since you're returning a `*OriginalGVKFunction` and asserting `var _ astmodel.Function = &OriginalGVKFunction{}` that the receiver here should also be a ptr (this applies to all the functions here)

hack/generator/pkg/functions/original_version_function.go
```diff
@@ -0,0 +1,84 @@
...
+
+// NewOriginalVersionFunction creates a new OriginalVersionFunction
+func NewOriginalVersionFunction(idFactory astmodel.IdentifierFactory) *OriginalVersionFunction {
+	return &OriginalVersionFunction{
+		idFactory: idFactory,
+	}
+}
+
+// Name returns the name of this function, which is always OriginalVersion()
+func (o OriginalVersionFunction) Name() string {
```
matthchr, line 38: I think since you're returning a `*OriginalVersionFunction` and asserting `var _ astmodel.ValueFunction = &OriginalVersionFunction {}` that the receiver here should also be a ptr (this applies to all the functions here)

hack/generator/pkg/functions/original_version_function_test.go
```diff
@@ -0,0 +1,28 @@
...
+package functions
+
+import (
+	"testing"
+
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/astmodel"
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/test"
+)
+
+func Test_OriginalVersionFunction_GeneratesExpectedCode(t *testing.T) {
```
matthchr, line 15: Nice, I like these per-function golden tests. Good idea.

hack/generator/pkg/codegen/storage/storage_type_factory.go
```diff
@@ -32,67 +35,77 @@ func NewStorageTypeFactory(group string, idFactory astmodel.IdentifierFactory) *
...
 			return nil, err
 		}
 	}
 
-	return f.outputTypes, nil
+	return f.types, nil
 }
 
+// process carries out our transformations
+// Each step reads from outputTypes and puts the results back in there
```
matthchr, line 86: minor: There is no more `outputTypes`

hack/generator/pkg/codegen/pipeline/create_storage_types.go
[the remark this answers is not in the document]
matthchr: I think I'm still missing why `storageTypes` here isn't a superset of `types`?

Don't we:
1. Add all types to a `StorageTypeFactory`.
2. Return all the types we added, plus new types generated, from the `StorageTypeFactory`?

So `storageTypes already contains all the types? No `types.OverlayWith(storageTypes)` needed?

matthchr (approved):

hack/generator/pkg/codegen/storage/storage_type_factory.go
matthchr: I think you've resolved the issue with recomputing (item 1 above).

I do still wonder if the `types.Process(...)` approach which leads itself to a mutable collection (`types`) stored on the `StorageTypeFactory` is hurting us more than it's helping though, and if a set of pure (or more-pure) functions would make the sequence of mutations clearer even if we had to do away with `types.Process(...)` and just write a loop.

hack/generator/pkg/codegen/storage/type_converter.go
babbageclunk, line 124:
```suggestion
	// Storage resource needs OriginalGVK() function injected too
```

hack/generator/pkg/codegen/storage/type_converter.go
```diff
@@ -149,3 +220,8 @@ func (_ *TypeConverter) descriptionForStorageVariant(definition astmodel.TypeDef
 
 	return result
 }
+
+type typeConverterContext struct {
+	isSpec   bool
+	isStatus bool
```
babbageclunk, line 226: `isStatus` doesn't appear to be used anywhere - maybe typeConverterContext could just be a bool or enum value?

babbageclunk (approved):
This looks great, although I think the change you discussed of breaking StorageTypeFactory into a set of pipeline stages will be nicer.

hack/generator/pkg/codegen/pipeline/create_storage_types.go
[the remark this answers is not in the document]
theunrepentantgeek: I've gotten rid of `storage_types_factory` as it was turning into a God class; instead there are a set of individual pipeline stages that cooperate to make the required changes.

hack/generator/pkg/codegen/storage/storage_type_factory.go
theunrepentantgeek: If we have a resource with no spec, I think it will blow up before we get here. I don't see the need to check.

hack/generator/pkg/codegen/storage/storage_type_factory.go
theunrepentantgeek: I've gotten rid of `storage_types_factory`, replacing it with a set of individual pipeline stages.

hack/generator/pkg/codegen/storage/storage_type_factory.go
theunrepentantgeek: I've removed `storage_type_factory` entirely.

hack/generator/pkg/codegen/storage/testdata/Test_StorageTypeFactory_GeneratesExpectedResults-v20211231storage.golden
theunrepentantgeek: Yes, but I've removed `storage_type_factory` entirely. #nofixneeded

hack/generator/pkg/codegen/storage/type_converter.go
theunrepentantgeek: They were going to become different, in another WIP branch. But now I've switched to separate stages, they are the same and I've collapsed them into a single method.

hack/generator/pkg/codegen/storage/type_converter.go
theunrepentantgeek: Removed into a separate stage.

hack/generator/pkg/functions/original_gvk_function.go
theunrepentantgeek: Fixed.

hack/generator/pkg/functions/original_version_function.go
theunrepentantgeek: Fixed.

hack/generator/pkg/codegen/storage/storage_type_factory.go
theunrepentantgeek: there is no more `storage_type_factory` too! 😁

hack/generator/pkg/codegen/storage/testdata/Test_StorageTypeFactory_GeneratesExpectedResults-v20211231storage.golden
```diff
@@ -0,0 +1,54 @@
+// Code generated by azure-service-operator-codegen. DO NOT EDIT.
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+package v20211231storage
+
+import (
```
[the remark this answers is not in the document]
theunrepentantgeek: There will be, that's in a separate WIP branch for now.

hack/generator/pkg/codegen/storage/type_converter.go
theunrepentantgeek: No longer need this struct at all.

hack/generator/pkg/codegen/storage/storage_type_factory.go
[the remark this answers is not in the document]
theunrepentantgeek: I've removed `storage_type_factory` completely.

hack/generator/pkg/codegen/storage/storage_type_factory.go
```diff
@@ -103,6 +116,7 @@ func (f *StorageTypeFactory) process() error {
 // visitor is a type visitor that will do the creation
```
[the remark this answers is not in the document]
theunrepentantgeek: Yup. gone now.

babbageclunk (approved):

hack/generator/pkg/functions/original_gvk_function.go
```diff
@@ -0,0 +1,110 @@
...
+	hasOriginalVersionProperty bool
+}
+
+// Ensure OriginalGVKFunction properly implements Function
+var _ astmodel.Function = &OriginalGVKFunction{}
+
+type OriginalVersionKind string
+
+const (
+	ReadProperty = OriginalVersionKind("property")
```
matthchr, line 39: I hate that go doesn't have first class enum support, and the solution of consts _kinda_ works but if the consts don't have the enum name in them it's sometimes hard to pick out which values belong to which type.

I know in some places we do something like `OriginalVersionKindReadProperty` (which is sorta how you'd do it in C, I think?), but we don't have that consistently and it is awkwardly long.

Does this bug other folks too? What do people prefer?

hack/generator/pkg/test/asserts.go
```diff
@@ -28,3 +29,27 @@ func AssertFileGeneratesExpectedCode(t *testing.T, fileDef *astmodel.FileDefinit
 
 	g.Assert(t, testName, buf.Bytes())
 }
+
+// AssertPackagesGenerateExpectedCode creates a golden file for each package represented in the passed set of type
+// definitions, asserting that the generated content is expected
+func AssertPackagesGenerateExpectedCode(t *testing.T, types astmodel.Types, prefix string) {
```
matthchr, line 35: It'd be nice to have the files each in a subfolder with the name of the test? Otherwise feels like we'll get just a crazy number of golden files and it'll be hard to tell which apply to what? I'm actually struggling with that in this PR even with the tests you've added.

matthchr (approved):
LGTM - Left a bunch of nits and one comment about golden test file structure.

hack/generator/pkg/codegen/pipeline/testdata/TestInjectPropertyAssignmentFunctions-v20211231storage.golden
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"github.com/Azure/azure-service-operator/testing/microsoft.person/v20211231storage"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+//Storage version of v20211231.Person
+type Person struct {
```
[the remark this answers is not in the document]
matthchr: I realized they are testing something but just with split files which I didn't realize - as mentioned much further below it'd be nice to have the related test golden files in the same subfolder?

hack/generator/pkg/astmodel/function_container.go
```diff
@@ -11,4 +11,19 @@ type FunctionContainer interface {
 	// Functions returns all the function implementations
 	// A sorted slice is returned to preserve immutability and provide determinism
 	Functions() []Function
+
+	// HasFunctionWithName determines if this resource has a function with the given name
+	HasFunctionWithName(name string) bool
+}
+
+// AsFunctionContainer converts a type into a function container
+func AsFunctionContainer(theType Type) (FunctionContainer, bool) {
```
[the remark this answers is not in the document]
theunrepentantgeek: Added comment.

hack/generator/pkg/astmodel/property_container.go
```diff
@@ -15,3 +15,15 @@ type PropertyContainer interface {
 	// Property returns the property and true if the named property is found, nil and false otherwise
 	Property(name PropertyName) (*PropertyDefinition, bool)
 }
+
+// AsPropertyContainer converts a type into a property container
+func AsPropertyContainer(theType Type) (PropertyContainer, bool) {
```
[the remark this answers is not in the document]
theunrepentantgeek: Added comment.

hack/generator/pkg/codegen/pipeline/create_storage_types.go
```diff
@@ -8,69 +8,53 @@ package pipeline
...
-					continue
-				}
-
-				factory, ok := factories[ref.Group()]
-				if !ok {
-					klog.V(3).Infof("Creating storage factory for %s", ref.Group())
-					factory = storage.NewStorageTypeFactory(ref.Group(), idFactory)
-					factories[ref.Group()] = factory
-				}
+			// Isolate both resources and complex objects
```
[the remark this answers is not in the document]
theunrepentantgeek: Performance is still pretty good, so I went with the first suggestion, using a couple of well named lambda variables to improve clarity.

hack/generator/pkg/codegen/pipeline/create_storage_types.go
```diff
@@ -8,69 +8,53 @@ package pipeline
...
+				storageDef, err := typeConverter.ConvertDefinition(def)
 				if err != nil {
-					errs = append(errs, err)
-					continue
+					return nil, errors.Wrapf(err, "creating storage variant of %q", name)
 				}
 
-				result.AddTypes(t)
+				storageTypes.Add(storageDef)
+				conversionGraph.AddLink(name.PackageReference, storageDef.Name().PackageReference)
```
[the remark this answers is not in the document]
theunrepentantgeek: I'll move creation of the storage graph entirely out into a separate stage for clarity; will do that in a separate PR though.

hack/generator/pkg/codegen/pipeline/testdata/TestInjectPropertyAssignmentFunctions-v20211231storage.golden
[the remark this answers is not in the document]
theunrepentantgeek: We should update the golden file tests to use subfolders in a later PR.
Also, once all the storage stuff has finished, we can (and should) remove the suppressions from the codegen golden tests.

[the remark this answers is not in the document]
theunrepentantgeek: I created issue #1626 to track.

hack/generator/pkg/codegen/storage/property_injector.go
```diff
@@ -0,0 +1,45 @@
...
+	result.visitor = astmodel.TypeVisitorBuilder{
+		VisitObjectType:   result.injectPropertyIntoObject,
+		VisitResourceType: result.injectPropertyIntoResource,
+	}.Build()
+
+	return result
+}
+
+// Inject modifies the passed type definition by injecting the passed property
+func (fi *PropertyInjector) Inject(def astmodel.TypeDefinition, prop *astmodel.PropertyDefinition) (astmodel.TypeDefinition, error) {
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

hack/generator/pkg/codegen/storage/types_tools.go
```diff
@@ -0,0 +1,80 @@
...
+ */
+
+package storage
+
+import (
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/astmodel"
+)
+
+// FindResourceTypes walks the provided set of TypeDefinitions and returns all the resource types
+func FindResourceTypes(types astmodel.Types) astmodel.Types {
```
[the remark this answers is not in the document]
theunrepentantgeek: Good idea.

hack/generator/pkg/functions/original_gvk_function.go
theunrepentantgeek: I used very simple names here to optimize the API when consuming; adding a unique suffix would make that clunky. 

Having first class enum support would be preferable - but I guess Go is designed to push people towards many simpler packages.

hack/generator/pkg/test/asserts.go
theunrepentantgeek: Agreed. See #1626

hack/generator/pkg/codegen/storage/type_converter.go
```diff
@@ -20,12 +22,15 @@ type TypeConverter struct {
...
+	// idFactory is a reference to our IdentifierFactory singleton
+	idFactory astmodel.IdentifierFactory
 }
 
 // NewTypeConverter creates a new instance of the utility type
-func NewTypeConverter(types astmodel.Types) *TypeConverter {
+func NewTypeConverter(types astmodel.Types, idFactory astmodel.IdentifierFactory) *TypeConverter {
 	result := &TypeConverter{
 		types:             types,
+		idFactory:         idFactory,
```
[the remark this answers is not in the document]
theunrepentantgeek: Good catch; no longer used, after the refactoring. Removed.

hack/generator/pkg/codegen/pipeline/create_storage_types_test.go
```diff
@@ -0,0 +1,50 @@
...
+	idFactory := astmodel.NewIdentifierFactory()
+	// Test Resource V1
+
+	specV1 := test.CreateSpec(pkg2020, "Person", fullNameProperty, familyNameProperty, knownAsProperty)
+	statusV1 := test.CreateStatus(pkg2020, "Person")
+	resourceV1 := test.CreateResource(pkg2020, "Person", specV1, statusV1)
+
+	// Test Resource V2
+
+	specV2 := test.CreateSpec(pkg2021, "Person",
```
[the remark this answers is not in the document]
theunrepentantgeek: It's not; I split it that way because they're all properties, but I agree it's a mess. Split as one parameter per line as per our conventions.

hack/generator/pkg/codegen/pipeline/inject_original_gvk_function.go
```diff
@@ -0,0 +1,55 @@
...
+// This function allows us to recover the original version used to create each custom resource, giving the operator the
+// information needed to interact with ARM using the correct API version.
+func InjectOriginalGVKFunction(idFactory astmodel.IdentifierFactory) Stage {
+
+	result := MakeStage(
+		injectOriginalGVKFunctionId,
+		"Inject the function OriginalGVK() into each Resource type",
+		func(ctx context.Context, types astmodel.Types) (astmodel.Types, error) {
+			injector := storage.NewFunctionInjector()
+			result := types.Copy()
```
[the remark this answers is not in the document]
theunrepentantgeek: Good idea. Done.

hack/generator/pkg/codegen/pipeline/inject_original_gvk_function.go
```diff
@@ -0,0 +1,55 @@
...
+
+			specs := storage.FindResourceTypes(types)
```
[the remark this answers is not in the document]
theunrepentantgeek: Variable has been renamed to `resources`. Good catch.

hack/generator/pkg/codegen/pipeline/inject_original_gvk_function.go
```diff
@@ -0,0 +1,55 @@
...
+				} else {
+					fn = functions.NewOriginalGVKFunction(functions.ReadFunction, idFactory)
+				}
+
+				defWithFn, err := injector.Inject(def, fn)
+				if err != nil {
+					return nil, errors.Wrapf(err, "injecting OriginalVersion() into %s", name)
+				}
+
+				result[defWithFn.Name()] = defWithFn
```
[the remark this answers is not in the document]
theunrepentantgeek: This stage adds no types, only updates existing ones.

hack/generator/pkg/codegen/pipeline/inject_original_version_function.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/astmodel"
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/codegen/storage"
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/functions"
+)
+
+// injectOriginalVersionFunctionStageId is the unique identifier for this pipeline stage
+const injectOriginalVersionFunctionStageId = "injectOriginalVersionFunction"
+
+// InjectOriginalVersionFunction injects the function OriginalVersion() into each Spec type
```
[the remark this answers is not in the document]
theunrepentantgeek: ARM types don't need this, and I think they're filtered out.

hack/generator/pkg/codegen/pipeline/inject_original_version_function.go
```diff
@@ -0,0 +1,46 @@
...
+// This function allows us to recover the original version used to create each custom resource, giving the operator the
+// information needed to interact with ARM using the correct API version.
+func InjectOriginalVersionFunction(idFactory astmodel.IdentifierFactory) Stage {
+
+	return MakeStage(
+		injectOriginalVersionFunctionStageId,
+		"Inject the function OriginalVersion() into each Spec type",
+		func(ctx context.Context, types astmodel.Types) (astmodel.Types, error) {
+			injector := storage.NewFunctionInjector()
+			result := types.Copy()
```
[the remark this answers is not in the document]
theunrepentantgeek: Again, this stage doesn't create any new types, it only modifies existing ones. The defensive assert would stop it working.

hack/generator/pkg/codegen/pipeline/inject_original_version_property.go
```diff
@@ -0,0 +1,59 @@
...
+// injectOriginalVersionPropertyId is the unique identifier for this pipeline stage
+const injectOriginalVersionPropertyId = "injectOriginalVersionProperty"
+
+// InjectOriginalVersionProperty injects the property OriginalVersion into each Storage Spec type
+// This property gets populated by reading from the OriginalVersion() function previously injected into the API Spec
+// types, allowing us to recover the original version used to create each custom resource, and giving the operator the
+// information needed to interact with ARM using the correct API version.
+func InjectOriginalVersionProperty() Stage {
+
+	result := MakeStage(
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

hack/generator/pkg/codegen/pipeline/inject_original_version_property.go
```diff
@@ -0,0 +1,59 @@
...
+		injectOriginalVersionPropertyId,
+		"Inject the property OriginalVersion into each Storage Spec type",
+		func(ctx context.Context, types astmodel.Types) (astmodel.Types, error) {
+			injector := storage.NewPropertyInjector()
+			result := types.Copy()
```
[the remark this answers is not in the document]
theunrepentantgeek: No new types - only changes to existing ones.

hack/generator/pkg/codegen/pipeline/inject_original_version_property.go
```diff
@@ -0,0 +1,59 @@
...
+
+			storageSpecs := storage.FindSpecTypes(types).Where(func(definition astmodel.TypeDefinition) bool {
+				fc, ok := astmodel.AsFunctionContainer(definition.Type())
+				if !ok {
+					// Exclude everything except resources and complex objects
```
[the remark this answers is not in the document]
theunrepentantgeek: Reworked the filter.

hack/generator/pkg/codegen/pipeline/inject_original_version_function.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/codegen/storage"
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/functions"
+)
+
+// injectOriginalVersionFunctionStageId is the unique identifier for this pipeline stage
+const injectOriginalVersionFunctionStageId = "injectOriginalVersionFunction"
+
+// InjectOriginalVersionFunction injects the function OriginalVersion() into each Spec type
+// This function allows us to recover the original version used to create each custom resource, giving the operator the
+// information needed to interact with ARM using the correct API version.
```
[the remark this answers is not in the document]
theunrepentantgeek: We inject the `OriginalVersion()` function before we create any storage types. I've documented this.

hack/generator/pkg/codegen/pipeline/inject_property_assignment_functions_test.go
```diff
@@ -0,0 +1,52 @@
...
+	idFactory := astmodel.NewIdentifierFactory()
+	// Test Resource V1
+
+	specV1 := test.CreateSpec(pkg2020, "Person", fullNameProperty, familyNameProperty, knownAsProperty)
+	statusV1 := test.CreateStatus(pkg2020, "Person")
+	resourceV1 := test.CreateResource(pkg2020, "Person", specV1, statusV1)
+
+	// Test Resource V2
+
+	specV2 := test.CreateSpec(pkg2021, "Person",
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

hack/generator/pkg/codegen/pipeline/create_storage_types.go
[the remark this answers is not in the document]
matthchr: Yeah that's even better, good idea.

PR #1584, opened 2021-06-21
Create TypeNameProcessingRegister

theunrepentantgeek:
**What this PR does / why we need it**:

I have a need in `StorageTypeFactory` to track all of the `TypeName` references encountered, checking to see if we process a `TypeDefinition` for each one. I've factored this out as a utility type and intend to use it from at least two different locations. 

We'd brainstormed the name `DynamicTypeNameCheckList` for this type, but that didn't work quite so well in use. 

I've also changed the API for `TypeNameSet` to remove the returns from mutators, and cleaned up code that was reliant upon them.

**How does this PR make you feel**:
gif

**If applicable**:
- [x] this PR contains tests

PR #1607, opened 2021-06-26
Improve the API for TypeNameSet

theunrepentantgeek:

The API design of `TypeNameSet`, requiring the result of `Add()` to be assigned back to the original variable resulted in a number of recurring bugs as it was easy to forget to do this. I'm changing the API to a more regular set style.

Extracted from PR #1584, which I might now abandon.

gif


hack/generator/pkg/codegen/embeddedresources/remover.go
```diff
@@ -273,11 +273,21 @@ func findSubResourcePropertiesTypeNames(types astmodel.Types) (map[astmodel.Type
...
+				result[owner] = astmodel.NewTypeNameSet(*specPropertiesTypeName)
+			} else {
+				result[owner].Add(*specPropertiesTypeName)
+			}
 		}
+
 		if statusPropertiesTypeName != nil {
-			result[owner] = result[owner].Add(*statusPropertiesTypeName)
+			if result[owner] == nil {
+				result[owner] = astmodel.NewTypeNameSet(*statusPropertiesTypeName)
```
Porges, line 287: to make these (and the one below) simpler we could have a version of `map[astmodel.TypeName]astmodel.TypeNameSet` that auto-vivifies

Porges (approved):

PR #1608, opened 2021-06-28
Don't fail on flattening collisions

Porges:
The new VMSS resources do this (a collision on `Type` property). Technically we can handle this case properly but it will require more work.

We can handle the VMSS case because the user can't normally overwrite the top-level `Type` property, so any `Type` property in the CRD must point to the flattened `Type`.

### DIFF
--- a/hack/generator/pkg/codegen/pipeline/flatten_properties.go
+++ b/hack/generator/pkg/codegen/pipeline/flatten_properties.go
@@ -53,7 +53,7 @@ func makeFlatteningVisitor(defs astmodel.Types) astmodel.TypeVisitor {
 
 			// safety check:
 			if err := checkForDuplicateNames(newProps); err != nil {
-				klog.V(1).Infof("Flattening caused duplicate property names, skipping flattening")
+				klog.Warningf("Flattening caused duplicate property names, skipping flattening")
 				return it, nil // nolint:nilerr
 			}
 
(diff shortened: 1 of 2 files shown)


hack/generator/pkg/codegen/pipeline/flatten_properties.go
```diff
@@ -52,7 +53,8 @@ func makeFlatteningVisitor(defs astmodel.Types) astmodel.TypeVisitor {
 
 			// safety check:
 			if err := checkForDuplicateNames(newProps); err != nil {
-				return nil, err
+				klog.V(1).Infof("Flattening caused duplicate property names, skipping flattening")
```
matthchr, line 56: Maybe this is an error or warning rather than info?

Also do we want to file a bug to do something more intelligent here in the future?

matthchr (approved):

hack/generator/pkg/codegen/pipeline/flatten_properties.go
Porges: Yes I'm going to open a task.

hack/generator/pkg/codegen/pipeline/flatten_properties.go
```diff
@@ -52,7 +53,8 @@ func makeFlatteningVisitor(defs astmodel.Types) astmodel.TypeVisitor {
 
 			// safety check:
 			if err := checkForDuplicateNames(newProps); err != nil {
-				return nil, err
+				klog.Warningf("Flattening caused duplicate property names, skipping flattening")
```
matthchr, line 56: Maybe log the error here since that includes the property name?

Porges: good point

matthchr (approved):

PR #1611, opened 2021-06-29
Update Spec for how the correct API version will be used with ARM

theunrepentantgeek:

Completes the spec descrbing how we will ensure we use the right Spec/Status formats when interacting with ARM.

The interface `GeneratedResource` is designed to plug in cleanly to the existing reconciler, let me know if you see any problems with that.

gif


docs/design/api-versions.md
```diff
@@ -16,13 +16,21 @@ If we have a valid **v1** `Person`, trying to submit that through the **v2** ARM
 
 ## Proposed Solution
 
-We need to preserve the original API Version of each resource, and use that to create an appropriate resource for ARM.
+**Preserve the API Version**: the original API version used to create the custom resource (CR) will be preserved so that we know which API version to use when interacting with ARM.
+
+**New conversion interfaces**: to avoid having to version-convert an entire resource when we are interested in only its `Spec` or `Status`, we'll define and implement new interfaces to allow direct conversion.
```
matthchr, line 21: New interfaces on the `spec` and `status`? I'm 98% sure the answer is yes but might be worth clarifying here how "new interfaces" provides the "avoiding having to version-convert an entire resource" problem.

matthchr: Oh maybe you go into this below

docs/design/api-versions.md
```diff
@@ -32,8 +40,169 @@ For each API spec, generated `AssignPropertiesTo*()` method  will read the value
...
 
-SequenceDiagram TBC.
+We don't want the overhead of doing a complete conversion of each resource if we're only interested in conversion of the associated `Spec` or `Status`. For example, when submitting a request to ARM we need only do a `Spec` conversion; similiarly, when processing a response from ARM, we need only do a `Status` conversion.
+
+We already have the required infrastructure in the form of the `AssignProperties*()` methods, making it simple to implement these new interfaces.
+
+For conversion of `Spec` types:
+
+``` go
+type ConvertibleSpec interface {
```
matthchr, line 52: Not immediately grasping why this helps us since `Spec` already implements `Convertible` that you mention below? Why can't we just use that?

docs/design/api-versions.md
```diff
@@ -32,8 +40,169 @@ For each API spec, generated `AssignPropertiesTo*()` method  will read the value
...
+    ConvertToSpec(destination ConvertibleSpec) error
+}
+```
+
+As we have done with implementations of the traditional [`Convertible`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/conversion#Convertible) interface, these will be implemented on Specs that are further from our hub, and will chain towards that hub.
+
+For conversion of `Status` types, a similar approach:
+
+``` go
+type ConvertibleStatus interface {
```
matthchr, line 66: Same question about why not just use `Convertible` which the Status types should already implement?

docs/design/api-versions.md
```diff
@@ -32,8 +40,169 @@ For each API spec, generated `AssignPropertiesTo*()` method  will read the value
...
+```
+
+## Reconciler integration
+
+When the reconciler is working with ARM, we'll use the `OriginalVersion` property on the storage spec to pivot from the current version to the original version.
+
+To keep the reconciler a simple as possible, we'll handle this with generated code hidden behind a simple interface:
+
+``` go
+type GeneratedResource interface {
```
matthchr, line 82: minor: I think a name other than `GeneratedResource` would make sense here? We already have an existing interface which is `KubernetesResource`, documented as:

```
// KubernetesResource is an Azure resource. This interface contains the common set of
// methods that apply to all ASO resources.
type KubernetesResource interface {
	// Owner returns the ResourceReference of the owner, or nil if there is no owner
	Owner() *ResourceReference

	// TODO: I think we need this?
	// KnownOwner() *KnownResourceReference

	// AzureName returns the Azure name of the resource
	AzureName() string

	// Some types, but not all, have a corresponding:
	// 	SetAzureName(name string)
	// They do not if the name must be a fixed value (like 'default').

	// TODO: GetAPIVersion here?

	// TODO: I think we need this
	// SetStatus(status interface{})
}
```

Probably makes sense to just expand that? Certainly this interface and it seem to have a lot in common, and there are some TODOs on `KubernetesResource` that look a lot like what you're proposing here.

docs/design/api-versions.md
```diff
@@ -32,8 +40,169 @@ For each API spec, generated `AssignPropertiesTo*()` method  will read the value
...
+    // GetSpec returns the spec of the resource ready for 
+    // serialization to ARM
+    GetSpec(scheme runtime.Scheme) (genruntime.ARMTransformer, error)
```
matthchr, line 85: In keeping with the above, if this were on a `KubernetesResource` interface maybe this is `GetSpecARMConverter` or something to differentiate from another possible method `GetSpec()` which just returned the spec as a `runtime.Object` or something?

matthchr: Same with the below methods, consider sprinkling some `ARM` specific stuff in the name to contrast with another possible method which creates the corresponding Kubernetes type rather than the ARM type.

docs/design/api-versions.md
matthchr: Or if we want a separate interface we could embed that interface into this interface.

docs/design/api-versions.md
```diff
@@ -32,8 +40,169 @@ For each API spec, generated `AssignPropertiesTo*()` method  will read the value
...
+    // Use gvk to create an empty resource of the expected version 
+    originalObj, err := scheme.New(gvk)
+    if err != nil {
+        return errors.Wrapf(
+            err, "unable to create requested version %s of resource", gvk)
+    }
+    resource, ok := originalObj.(genruntime.GeneratedResource)
+    if !ok {
+        return errors.Wrapf(
+            err, "unable to convert %T to genruntime.GeneratedResource", originalObj)
```
matthchr, line 150: minor: Don't want to wrap `err` here because the issue was `!ok` not `err != nil`

docs/design/api-versions.md
```diff
@@ -32,8 +40,169 @@ For each API spec, generated `AssignPropertiesTo*()` method  will read the value
...
+    }
+
+    // Extract the Spec and populate
+    resultSpec := resource.GetSpec()
```
matthchr, line 154: This `GetSpec` also takes `scheme`, but I actually wonder if what you've written is nicer and we want 2 methods on two interfaces?

```go
type KubernetesResource interface {
    GetSpec() interface{}
    // That other stuff it already has
}

type StorageResource interface {
    GetARMSpec(scheme runtime.Scheme) (genruntime.ARMTransformer, error)
}
```

This would convey that: there are two different things we're doing, one of which can fail and one of which cannot. Storage types could implement both interfaces

If you don't want to return `interface{}` on the first one and have to cast, we could also have a `GetSpecAsARMTransformer` for the first interface, although I suspect we'll end up wanting `GetSpec() interface` anyway since right now we're doing that via reflection I'm pretty sure.

Then the storage types could implement `KubernetesResource` _and_ `StorageResource` (or possibly `StorageResource` can embed `KubernetesResource` and we can then change the reconciler to reconcile on that type rather than `KubernetesResource` which is what it's doing now

docs/design/api-versions.md
```diff
@@ -32,8 +40,169 @@ For each API spec, generated `AssignPropertiesTo*()` method  will read the value
...
+        return errors.Wrapf(
+            err, "unable to convert %T to genruntime.GeneratedResource", originalObj)
+    }
+
+    // Return an empty Status from the empty resource
+    return resource.NewStatus(scheme)
+}
+```
+
+After the reconciler has populated an empty `Status` instance, `SetStatus()` is called for the update. It will be the appropriate API `Spec` as that is returned by `NewStatus()`, so all we need to do is to populate our Hub status.
```
matthchr, line 184:
```suggestion
After the reconciler has populated an empty `Status` instance, `SetStatus()` is called for the update. It will be the appropriate API `Status` as that is returned by `NewStatus()`, so all we need to do is to populate our Hub status.
```

matthchr: I think?

docs/design/api-versions.md
theunrepentantgeek: Spec's done - and can't implement the standard `Convertible` interface as those methods require conversion to/from the hub resource type. Specs aren't resources, and only resources should implement that one.

docs/design/api-versions.md
theunrepentantgeek: I'll add a brief introductory paragraph here to indicate that this section is high level.

docs/design/api-versions.md
theunrepentantgeek: And, same answer as for Specs - they don't currently implement `Convertible`, and can't.

docs/design/api-versions.md
theunrepentantgeek: Yes, we need a better name than `GeneratedResource`, but that's the best I could come up with.

Not sure it makes sense to put onto `KubernetesResource`, but _maybe_. 

I'm not familiar with the purpose of `KubernetesResource` - if it fits, then sure. But, given interfaces represent abstractions, if the new methods don't fit the abstraction, we should use a new interface.

docs/design/api-versions.md
theunrepentantgeek: These methods don't return ARM types; they return implementations that can create ARM types.

docs/design/api-versions.md
theunrepentantgeek: Fixed.

docs/design/api-versions.md
theunrepentantgeek: I don't want to have two different interfaces because that pushes responsibilities onto the reconciler to know whether it's working with an API resource or a Storage resource (because it needs to know which interface to use). The goal is to abstract away those details so that the reconciler doesn't need to worry.

docs/design/api-versions.md
theunrepentantgeek: Correct. Fixed.

docs/design/api-versions.md
theunrepentantgeek: What about `VersionAwareResource` as that's the key functionality hidden behind the curtain - it will automatically do the right thing wrt versions.

docs/design/api-versions.md
matthchr: Doh, I missed that it only applied to top level types somehow. Got it.

docs/design/api-versions.md
matthchr: I suggested `GetARMSpecConverter` not `GetARMSpec` - I agree that `GetARMSpec` would be misleading, but I also think that `GetSpec` is misleading because this isn't really returning a spec, it's returning a thing that can be used to make an ARM type from a spec, and `GetSpec` seems better reserved for a method that actually returns the spec itself (and has no required parameters, unlike the required `scheme` here)

docs/design/api-versions.md
matthchr: > But, given interfaces represent abstractions, if the new methods don't fit the abstraction, we should use a new interface.

The original intent of `KubernetesResource` was to define an interface which must be implemented for a resource to be reconciled by the generic reconciler. If it ends up being composed of other interfaces I think that's fine, but I think we should have a single top level interface that is the canonical "This is what you must implement for generic_reconciler/azure_deployment_reconciler to work. This also eases the amount of casting done in the controller because it just has to cast once (to the top level interface) which has everything it's going to need.

matthchr: Ah sorry, actually the interface I was thinking of was:
```
type MetaObject interface {
	runtime.Object
	metav1.Object
	KubernetesResource
}
```
`genruntime.MetaObject` - so yeah we should decide what makes sense in `KubernetesResource` and what makes sense in another interface, but we should put whatever that other interface is into `MetaObject` alongside `KubernetesResource`

matthchr: happy to chat about this in person if you wanna discuss it more

docs/design/api-versions.md
matthchr: Yeah, fair enough.

In retrospect I think that this suggestion was really born from the `GetSpec` name making me think that the impl on the API type was really wanting to be `GetSpec() interface{}` when that's not its intent. A name that more clearly exposes the "Get me a thing that can make me an ARM Spec" helps resolve that, and agree that the goal of a common interface for getting that kind of thing between API and storage types is worthwhile.

docs/design/api-versions.md
matthchr: On the naming: `VersionAwareResource` feels kinda weird because I'm not sure I would call the API types "version aware" - they only really know their own version, yet they would be implementing this interface. I think that interface name fits well if we were only implementing it on storage types.

Maybe?
```go
// TODO: ARMConverterSupplier is a real awkward name, could be something else.
type ARMConverterSupplier interface {
        GetARMSpecConverter(scheme runtime.Scheme) (genruntime.ARMTransformer, error)
        NewARMStatusConverter(scheme runtime.Scheme) (genruntime.FromARMConverter, error)
}

type KubernetesResource interface {
        ARMConverterSupplier

	// Owner returns the ResourceReference of the owner, or nil if there is no owner
	Owner() *ResourceReference

	// AzureName returns the Azure name of the resource
	AzureName() string

	SetStatus(status interface{}) error
}
```

I pulled `SetStatus` out into the other interface because it felt like it was really pretty generic? In my head at least `KuberentesResource` would really end up having these 4 methods:
```go
GetStatus() interface{}
SetStatus(status interface{}) error
GetSpec() interface{}
SetSpec(spec interface{}) error
```

Although I am not sure we'll need the `SetSpec` one, or the `GetStatus` one because of the directionality of the operator.

A random aside: It's awkward we use both `Transformer` and `Converter` suffixes in the `FromARMConverter`/`ARMTransformer` interfaces. We should probably change `ARMTransformer` to `ARMConverter`? Then it's built of two embedded interfaces `FromARMConverter` and `ToARMConverter`?

theunrepentantgeek: As discussed in our design review, I'll put the required members directly on `KubernetesResource`, rather than defining a new resource interface. I've updated the spec to reflect this.

docs/design/api-versions.md
theunrepentantgeek: I'm moving the conversion into the reconciler - which both removes the need for a `scheme` parameter, and avoids confusion here because `GetSpec()` just returns the spec on the resource, no conversion required.

docs/design/api-versions.md
theunrepentantgeek: I'm moving the conversion into the reconciler - which both removes the need for a `scheme` parameter, and avoids confusion here because `GetSpec()` just returns the spec on the resource, no conversion required.

matthchr (approved):

docs/hugo/content/architecture-decision-records/ADR-2021-06-API-Version-Recovery.md
```diff
@@ -0,0 +1,172 @@
+# API Version Recovery
+
+## Context
+
+In addition to structural changes, there may be behaviour changes between ARM API versions. It is therefore important that we use the correct API version - the version requested by the user - when interacting with ARM, to ensure that we get the expected behaviour.
```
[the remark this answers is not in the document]
theunrepentantgeek: I've ammended the decision section to add some of your ideas, but don't think they belong in the context section (which is supposed to frame the problem).

PR #1614, opened 2021-07-01
Design for customer facing resource states

### LINKED ISSUE #1448: Design customer visible states and their meanings
Azure ARM resources have a `provisioningState`, which we're currently exposing directly as a `resource-state` annotation on the resource. We'd eventually like to move that to a field in `Status` which we're setting (I think).

There are also "states" we'd like to put the resource in on the Kuberentes side. An example Kubernetes state is `OwnerNotFound`. This is a state that a resource would idle in while its owner doesn't exist (for example, a `StorageAccount` where the `ResourceGroup` the account is supposed to be in doesn't exist).

There is also a possibly orthogonal state: `GoalStatus` whose values `Seeking`, `Reached`, and `Blocked` (names all placeholders) respectively mean that the controller is attempting to drive to the goal state, has reached the goal state, or has encountered an error which prevents the goal state from being reached. This would be the enum that attempts to replace the 3-tuple of booleans in ASO.

The overarching question is: **How should we model these states on our resources**? Some questions:
1. Should these two state-sources (Azure vs Kubernetes) be distinct fields on `status`, or somehow merged together?
2. Is there a relationship between when an error is set (yet another field on `status` we should be thinking about) and what the state(s) say?
3. What about error(s) that are non-fatal? There are lots of temporary errors which can be retried through - do we leave the state in `Seeking` but specify an error, or do we just log (or event?) on the error and not show it on the resource itself?
4. Does it make sense to make use of [conditions](https://github.com/kubernetes/community/blob/d7e353b47e73becde78b1be451058602e585cfd1/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties)? (I don't think they fit here, given the directive: "For observations that are well known and apply to all instances, a regular field is preferred") but in some ways they fit quite nicely...

Note: `ProvisioningState` is an ARM concept and so doesn't necessarily map cleanly on to handcrafted data plane resources. I think we're going to want at least a single field which is common across all resources which customers can look at to gauge the state at a high level.

matthchr:
Closes #1448

A design for how to expose the state of a resource to customers in ASO v2


matthchr:
I filed #1615 for actually implementing this. We can decide if we want to do it in alpha-0 or some other alpha release.

docs/design/resource-states.md
```diff
@@ -0,0 +1,213 @@
...
+	ResourceStateAnnotation  = "resource-state.infra.azure.com"
+	ResourceErrorAnnotation  = "resource-error.infra.azure.com"
+```
+
+This was done mostly to avoid the additional work to code-generate the correct fields in each resources `status` and set them. Since it's possible that the whole `status` can get lost, some of the things may need to stay 
+as annotations (`deploymentId` possibly), while others are obviously status and can be derived again if lost. We took the temporary approach of putting it all in annotations while we learned more about what the best practices were.
+
+#### Problems with this apporach
+* The `ResourceStateAnnotation` is reusing the same enum to describe state as the `armclient` uses for deployments. This is problematic because that only supports `Succeeded`, `Failed`, `Deleting`, and `Accepted`.
+  `Succeeded`, `Failed`, and `Deleting` make sense in the context of the operator but `Accepted` doesn't really. There's no state for `InProgress` or `Reconciling`.
```
Porges, line 50: also Failed and Deleting don't talk about the resource they talk about the deployment; the deployment can be Deleting but the resource still exists.

matthchr: Yeah, I'll update the doc to call that out more clearly.

matthchr:
Based on comments during live review, will update to be more like CAPI's.
Also change the "Deploying" to "Reconciling"

matthchr:
Saw some formatting issues -- will fix those

devigned (approved):
I left a little grammar nit, but otherwise lgtm.

The uniformity of providing a `Ready` condition feels like a good choice. It simplifies the Azure reconciliation cycle and is easily consumed by tooling. I also like the use of reason and severity to provide guidance on progress as well as failure states (still progressing, user intervention needed, and terminal). 

I don't have a strong opinion about `kstatus`. It seems like the tooling would pick up on `Ready` as effectively as `kstatus`.

docs/design/resource-states.md
```diff
@@ -0,0 +1,221 @@
...
+	ResourceId         string              `json:"resourceId,omitempty"`
+```
+
+#### Problems with this approach
+* `Provisioning`, `Provisioned` and `FailedProvisioning` are mutually exclusive but the structure of the `status` doesn't convey that well. Additionally the fact that there are multiple
+  fields means it's easier for bugs to creep in where we set one but didn't clear another.
+* The `State` field is not widely used by all resources, and seems to conflict with `Provisioning`/`Provisioned`/`FailedProvisioning`. It's not clear when you should look at one or the other.
+
+### ASO v2 (autogenerated)
+The autogenerated resources don't current have anything in their `Status` which conveys the Status _of the operator_. For the initial alpha releases, these fields are instead set as annotations on the object:
```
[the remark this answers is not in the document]
matthchr: Fixed this and pushed an update.

docs/design/resource-states.md
```diff
@@ -0,0 +1,221 @@
+# Proposal for reporting resource Status
+
+## What Status are we talking about?
+There are two types of Status that we're interested understanding and reporting to the customer when running an operator that creates resources in Azure:
```
CecileRobertMichon, line 4:
```suggestion
There are two types of Status that we're interested in understanding and reporting to the customer when running an operator that creates resources in Azure:
```

docs/design/resource-states.md
```diff
@@ -0,0 +1,221 @@
...
+* Using annotations means that it's not clear to the customer what the possible valid states are for the fields that are enums.
+* Users expect to look in `Status` for status related fields. Having it exposed as an annotation was never intended as a long term solution, just a hack so that we could get things up and running.
+
+## Propsal
+
+### Goals
+* The status of the operator should be reported as a subsection in the `Status` of the resource itself. See [#1504](https://github.com/Azure/azure-service-operator/pull/1504).
+* It should be clear to customers where to look to determine what the current status of their resource is.
+* The status should include information in both normal and failure cases.
+* When a failure occurs, it should be clear to customers that a failure has happened and what the cause of the failure was.
```
CecileRobertMichon, line 62:
```suggestion
* When a failure occurs, it should be clear to users that a failure has happened and what the cause of the failure was.
```

CecileRobertMichon: would it make sense to also add as a goal `When a failure occurs, it should be clear to users whether that failure is transient and the operator can self-recover.` or something like that? to distinguish terminal (eg. VM size unavailable vs transient errors (eg. internal server error, will retry)

docs/design/resource-states.md
```diff
@@ -0,0 +1,221 @@
...
+* ACK reports status through a [variety of conditions](https://github.com/aws-controllers-k8s/runtime/blob/8191f606c5975d5ba0a1433fab695533980733ba/apis/core/v1alpha1/conditions.go), including `ACK.Adopted`, `ACK.resourceSynced`, `ACK.Terminal`, etc.
+* Cluster API originally used a `phase` and `failureReason`/`failureMessage` pattern, but has since [moved to use conditions](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/20200506-conditions.md) and is deprecating the old pattern.
+* Pod uses a combination of [phase](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase) and [conditions](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions), including `PodScheduled`, `ContainersReady`, `Initalized`, and `Ready`.
+
+It seems that most of these (and many core Kubernetes resources) use a `Ready` condition to specify the status.
+
+### More on conditions
+These best practices were gathered from [conditions API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties)
```
CecileRobertMichon, line 77: would include something like `As of July 2021` since these are likely to evolve

docs/design/resource-states.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Reference
+1. [Article on conditions and status reporting in Kubernetes](https://dev.to/maelvls/what-the-heck-are-kubernetes-conditions-for-4je7). Note that this article is a bit old as Cluster API uses conditions now, but it gives a great overview of the topic.
+2. [Reccomended shape of Conditions](https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1623-standardize-conditions)
+3. [Update condition guidance](https://github.com/kubernetes/community/pull/4521) - and the actual guidance itself is [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties)
+4. [More condition reccomendations](https://github.com/kubernetes-sigs/cluster-api/issues/1658#issuecomment-584794287)
+
+## Open questions
+1. There are some small differences between this proposal and what CAPI is doing
+   * CAPI says `Reason` is optional, but we're making it required.
```
CecileRobertMichon, line 191: I believe the _reason_ `Reason` (😄 ) is optional is that it should only be set when the condition is False. There is no Reason on True Conditions

CecileRobertMichon: I guess the official recommendations do say it should be required though

> reason is required and must not be empty. The actor setting the value should always describe why the condition is the way it is, even if that value is "unknown unknowns". No other actor has the information to make a better choice.

need to go back and check why that's not the case in CAPI

docs/design/resource-states.md
matthchr: yes! This definitely was a goal but wasn't explicitly called out - thanks.

docs/design/resource-states.md
matthchr: Given that there are infinitely many failure cases and really only one success case, it feels "ok" to give a (somewhat useless) `Reason` like `Success` for the one success case, as it then saves everybody from then having think about an empty reason.

With that said the guidance isn't really crystal clear as `Reason` is clearly optional [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties).

CecileRobertMichon: Got a better answer on the CAPI slack: https://kubernetes.slack.com/archives/C8TSNPY4T/p1625615995097600

In case you can't open the link:


> cecile Yesterday at 4:59 PM
Does anyone know/remember why CAPI conditions have Reason as optional? The KEP from apimachinery says it should be required:
 reason is required and must not be empty. The actor setting the value should always describe why the condition is the way it is, even if that value is “unknown unknowns”. No other actor has the information to make a better choice.

> fabrizio.pandini  6 hours ago
Umhh, this comment does not match what the API types are defining few lines above in the doc
        // The reason for the condition’s last transition in CamelCase.
	// The specific API may choose whether or not this field is considered a guaranteed API.
	// This field may not be empty.

> fabrizio.pandini  6 hours ago
I personally don’t think that always adding a reason to a condition makes sense,
thing e.g,. when a condition signals everything is ok, like e.g. Machine Ready true (adding a reason in this case will be just redundant)
Instead, I think that adding a reason in case the condition are signalling a problem makes much more sense

docs/design/resource-states.md
matthchr: Fixed

docs/design/resource-states.md
matthchr: I'm not entirely sure I follow this answer. It seems to say:
1. Actually each specific API can choose if `Reason` is required or not.
2. Fabrizio (and a few others who thumbs up-ed his answer) don't see much reason to include `Reason` when the condition signals OK.

I _agree_ with point 2... but it's a really minor quibble. I can't imagine people getting confused by a reason that was included in the success case, and always including reason has the advantage that you can make a nice simple statement about reason: "It's always set". More importantly than that though, it matches the KEP. The other places where the CAPI diverged from the KEP were explicitly [called out](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/20200506-conditions.md#kubernetes-conditions).

@porges, @theunrepentantgeek, @babbageclunk - you have thoughts on this topic?

matthchr:
Note to self to add another open question: `LastTransitionTime` in the spec is documented as: "the last time the condition transitioned from one status to another." `Status` is a particular field, which can oftentimes not change even while the underlying `Reason`, `Severity`, or `Message` does change.

Do we want to define `LastTransitionTime` as the last time _anything_ changed, or just the last time the `Status` field changed?

docs/design/resource-states.md
```diff
@@ -0,0 +1,222 @@
...
+	ResourceErrorAnnotation  = "resource-error.infra.azure.com"
+```
+
+This was done mostly to avoid the additional work to code-generate the correct fields in each resources `status` and set them. Since it's possible that the whole `status` can get lost, some of the things may need to stay 
+as annotations (`deploymentId` possibly), while others are obviously status and can be derived again if lost. We took the temporary approach of putting it all in annotations while we learned more about what the best practices were.
+
+#### Problems with this apporach
+* The `ResourceStateAnnotation` is reusing the same enum to describe state as the `armclient` uses for deployments. This is problematic because that only supports `Succeeded`, `Failed`, `Deleting`, and `Accepted`.
+  `Succeeded`, `Failed`, and `Deleting` make sense in the context of the operator but `Accepted` doesn't really. There's no state for `InProgress` or `Reconciling`.
+* There's an awkward tension between sharing the state of the ARM deployment and the state of the resource. They're often the same but sometimes not (such as in the case the deployment is being deleted but the resource is not).
```
theunrepentantgeek, line 51: They conincidentally share some of the same names for their state but I agree that they're not really the same thing. Reusing the same enum for mutliple things that conincidentally seem similar is a good way to get tangled up later.

docs/design/resource-states.md
```diff
@@ -0,0 +1,222 @@
...
+## Propsal
+
+### Goals
+* The status of the operator should be reported as a subsection in the `Status` of the resource itself. See [#1504](https://github.com/Azure/azure-service-operator/pull/1504).
+* It should be clear to users where to look to determine what the current status of their resource is.
+* The status should include information in both normal and failure cases.
+* When a failure occurs, it should be clear to the user that a failure has happened and what the cause of the failure was.
+* When a failure occurs, there should be an error or code that is programatically consumable (for automation, etc).
+* When a failure occurs, there should be an error or text that is human consumable and ideally more informative than the programatically consumable error code.
+* When a failure occurs, it should be clear to the user whether that failure is transient and the operator can self-recover.
```
theunrepentantgeek, line 65: I like the explicit list of goals. 👍

docs/design/resource-states.md
```diff
@@ -0,0 +1,222 @@
...
+
+It seems that most of these (and many core Kubernetes resources) use a `Ready` condition to specify the status.
+
+### More on conditions
+These best practices were gathered from [conditions API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) as of July 2021.
+
+> Some resources in the v1 API contain fields called phase, and associated message, reason, and other status fields. The pattern of using phase is deprecated. Newer API types should use conditions instead. Phase was essentially a state-machine enumeration field, that contradicted system-design principles and hampered evolution, since adding new enum values breaks backward compatibility. Rather than encouraging clients to infer implicit properties from phases, we prefer to explicitly expose the individual conditions that clients need to monitor. Conditions also have the benefit that it is possible to create some conditions with uniform meaning across all resource types, while still exposing others that are unique to specific resource types. See #7856 for more details and discussion.
```
theunrepentantgeek, line 80: The `#7876` is resolving as a link to an ASO issue, which doesn't exist - make it an explicit link back into the original repo.

docs/design/resource-states.md
```diff
@@ -0,0 +1,222 @@
...
+
+> In condition types, and everywhere else they appear in the API, Reason is intended to be a one-word, CamelCase representation of the category of cause of the current status, and Message is intended to be a human-readable phrase or sentence, which may contain specific details of the individual occurrence. Reason is intended to be used in concise output, such as one-line kubectl get output, and in summarizing occurrences of causes, whereas Message is intended to be presented to users in detailed status explanations, such as kubectl describe output.
```
theunrepentantgeek, line 82: Do you mean `camelCase` (leading lowercase with uppercase word separators) or `PascalCase` (leading uppercase with uppercase word separators)?

Also suggest "... Reason is intended to be a single compound-word, ... "

theunrepentantgeek: Though, if this is a quote ... keep it intact ...

docs/design/resource-states.md
```diff
@@ -0,0 +1,222 @@
...
+
+> reason is required and must not be empty. The actor setting the value should always describe why the condition is the way it is, even if that value is "unknown unknowns". No other actor has the information to make a better choice.
+
+Some other best practices on [Kubernetes design principles](https://github.com/kubernetes/community/blob/4c9ef2d/contributors/design-proposals/architecture/principles.md):
+
+> Object status must be 100% reconstructable by observation. Any history kept must be just an optimization and not required for correct operation.
+
+> Do not define comprehensive state machines for objects with behaviors associated with state transitions and/or "assumed" states that cannot be ascertained by observation.
+
+### Details
```
theunrepentantgeek, line 98: Is this section "details of our design" or still quoting from elsewhere?

docs/design/resource-states.md
```diff
@@ -0,0 +1,222 @@
...
+
+The structure of the `Condition` is based on the [reccomended shape of conditions KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1623-standardize-conditions) as well as the [Cluster API conditions propsal](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/20200506-conditions.md) and is as follows:
+
+```go
+// ConditionSeverity expresses the severity of a Condition Type failing.
+type ConditionSeverity string
+
+const (
+  // ConditionSeverityError specifies that a failure of a condition type
+  // should be viewed as an error.
```
theunrepentantgeek, line 110: Is this where we should be explicit that an error will require external intervention to address? (vs a warning that we might be able to resolve automatically.)

docs/design/resource-states.md
```diff
@@ -0,0 +1,222 @@
...
+
+  // ConditionSeverityWarning specifies that a failure of a condition type
+  // should be viewed as a warning, but that things could still work.
+  ConditionSeverityWarning ConditionSeverity = "Warning"
+
+  // ConditionSeverityInfo specifies that a failure of a condition type
+  // should be viewed as purely informational, and that things could still work.
+  ConditionSeverityInfo ConditionSeverity = "Info"
+
+  // ConditionSeverityNone will only be set if the condition is in state "True".
```
theunrepentantgeek, line 121: This comment contradicts the comment on `Severity`, below.

docs/design/resource-states.md
```diff
@@ -0,0 +1,222 @@
...
+   // +required
+   Status corev1.ConditionStatus `json:"status"`
+
+   // Severity with which to treat failures of this type of condition.
+   // For conditions which have positive polarity (Status == True is their normal/healthy state), this will be omitted when Status == True
+   // For conditions which have negative polarity (Status == False is their normal/healthy state), this will be omitted when Status == False.
+   // +optional
+   Severity ConditionSeverity `json:"severity,omitempty"`
+
+   // LastTransitionTime is the last time the condition transitioned from one status to another.
```
theunrepentantgeek, line 142:
```suggestion
   // LastTransitionTime is the last time the condition changed
```

docs/design/resource-states.md
```diff
@@ -0,0 +1,222 @@
...
+   // +required
+   LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty"`
+
+   // The reason for the condition's last transition.
+   // Reasons should be CamelCase with no spaces. A reason is always provided, this field will not be empty
```
theunrepentantgeek, line 147: Should this be `camelCase` or `PascalCase` as described earlier? Maybe qualify this with an explanation in parenthesis to be explicit.

```suggestion
   // Reasons should be camelCase (leading lowercase, with Uppercase word boundaries) with no spaces. A reason is always provided, this field will not be empty
```

docs/design/resource-states.md
```diff
@@ -0,0 +1,222 @@
...
+   Reason string `json:"reason"`
+
+   // A human readable message indicating details about the transition. This field may be empty
+   // +optional
+   Message string `json:"message,omitempty"`
+}
+```
+
+### The `Ready` condition
+The above discusses support for arbitrary conditions. In practice at least for now, ASO will only expose a single `Ready` condition.
```
theunrepentantgeek, line 158:
```suggestion
The `Condition` definition above discusses support for arbitrary conditions. In practice, at least for now, ASO will only expose a single `Ready` condition.
```

docs/design/resource-states.md
```diff
@@ -0,0 +1,222 @@
...
+
+#### The `Reason` field
+Inside of the `Ready` condition we always include a `Reason` which provides a programatically consumable reason that the resource is in the given state. 
+`Reason` is derived from multiple sources depending on where in the resource lifecycle we are. The value in `Reason` may be set by the operator itself, or be set to the result of an error or response code received from Azure. 
+
+The operator defines the following reasons:
+
+| Reason                 | Severity                        |   Meaning                                                                                                                                                                 |
+|------------------------|---------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Reconciling            | Info                            | A request has been submitted to Azure. No error has been received from Azure, but one may occur in the future.                                                            |
```
theunrepentantgeek, line 169: Maybe need to be explicit that we may be waiting to find out whether it worked or not.

docs/design/resource-states.md
```diff
@@ -0,0 +1,222 @@
...
+| WaitingForOwner        | Warning                         | The `owner` of this resource cannot be found in Kubernetes. Progress is blocked until the `owner` is created.                                                             |
+| Deleting               | Info                            | The resource is being deleted.                                                                                                                                            |
+| Succeeded              | None (""/empty)                 | The `spec` has successfully been applied. No additional changes are being attempted at this time.                                                                         |
+
+There are no failure conditions (`Severity = Error`) specified here because currently all fatal errors come directly from Azure. When an error response is received from Azure, the `Code` from Azure is set as the `Reason`, and the `Message` from Azure
+is set as the `Message`.
```
theunrepentantgeek, line 175: GitHub respects line breaks in markdown, so your paragraphs are going look a little weird when rendered. Each paragraph should be a single line of text. #nitpick

docs/design/resource-states.md
```diff
@@ -0,0 +1,222 @@
...
+
+#### `Severity` meaning in the context of the `Ready` condition
+
+* `Info` means that everything is proceeded as expected. This is the "happy path".
+* `Warning` means that something is wrong, but we haven't given up. The resource will continue to be reconciled until we can progress past the cause of the `Warning`. Users should examine the `Warning` as some may be due to the operator waiting for action from them, as in the case of `Reason=WaitingForOwner, Severity=Warning`. Others may be due to transient unavailability in Azure, as in the case of `Reason=InternalServerError, Severity=Warning`.
+* `Error` means that we were unable to reconcile the resource successfully. A `Ready` condition with `Status=False, Reason=<something>, Severity=Error` is no longer attempting to be reconciled. The user must make an update to the resource to fix one or more configuration problems.
```
theunrepentantgeek, line 181: Maybe list these in the reverse order, starting from `Error`?

docs/design/resource-states.md
theunrepentantgeek: A reason like `ReconciliationComplete` would make it clear that the reason (heh) `Ready` is now **true** is that we've done all our stuff. Removes ambiguity about how we got here.

docs/design/resource-states.md
```diff
@@ -0,0 +1,222 @@
...
+
+Possible outcomes:
+
+|      Status of deployment                                                  |               State of resource                                                                                   |
+|----------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------|
+| 4xx HTTP error code when creating deployment                               | Resource does not exist                                                                                           |
+| Deployment accepted, but fails after polling                               | Resource may or may not exist (depends on the RP)                                                                 |
+| Deployment succeeds                                                        | Resource has `Failed` `provisioningState` - The spec doesn't say this is impossible but it's probably very rare   |
+| Deployment succeeds                                                        | Resource has `Succeeded` `provisioningState`                                                                      |
```
theunrepentantgeek, line 219: I use a Visual Studio Code plugin that will reformat tables for clarity. See [Markdown All in One](https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one)

```suggestion
| Status of deployment                         | State of resource                                                                                               |
| -------------------------------------------- | --------------------------------------------------------------------------------------------------------------- |
| 4xx HTTP error code when creating deployment | Resource does not exist                                                                                         |
| Deployment accepted, but fails after polling | Resource may or may not exist (depends on the RP)                                                               |
| Deployment succeeds                          | Resource has `Failed` `provisioningState` - The spec doesn't say this is impossible but it's probably very rare |
| Deployment succeeds                          | Resource has `Succeeded` `provisioningState`                                                                    |

```

theunrepentantgeek (changes requested):
Looks really good. A bunch of comments, nothing particularly important.

docs/design/resource-states.md
```diff
@@ -0,0 +1,222 @@
...
+## Reference
+1. [Article on conditions and status reporting in Kubernetes](https://dev.to/maelvls/what-the-heck-are-kubernetes-conditions-for-4je7). Note that this article is a bit old as Cluster API uses conditions now, but it gives a great overview of the topic.
+2. [Reccomended shape of Conditions](https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1623-standardize-conditions)
+3. [Update condition guidance](https://github.com/kubernetes/community/pull/4521) - and the actual guidance itself is [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties)
+4. [More condition reccomendations](https://github.com/kubernetes-sigs/cluster-api/issues/1658#issuecomment-584794287)
+
+## Open questions
+1. There are some small differences between this proposal and what CAPI is doing
+   * CAPI says `Reason` is optional, but we're making it required.
+   * CAPI's `Type` field is an alias to `string` called `ConditionType` and ours is just `string`.
```
matthchr, line 193: I've changed my mind on this one -- will update to use a string alias and define a set of possible values. That more clearly shows that what we currently support is just `Ready` and if we add more they should be added uniformly across the operator (we don't have the same sub-project situation that CAPI does)

docs/design/resource-states.md
matthchr: Yes, exactly. The ARM deployment state also by definition mirrors the resource same _much of the time_ (but not all the time), so it's extra confusing to use as it almost seems right but actually isn't.

docs/design/resource-states.md
matthchr: Fixed

docs/design/resource-states.md
matthchr: This is a quote, and camelCase is apparently a Microsoft-ism according to [Wikipedia](https://en.wikipedia.org/wiki/Camel_case):
> Camel case is often used as a naming convention in computer programming, but is an ambiguous definition due the to the optional capitalization of the first letter. Some programming styles prefer camel case with the first letter capitalised, others not. For clarity, this article calls the two alternatives upper camel case (initial uppercase letter, also known as Pascal case) and lower camel case (initial lowercase letter, also known as dromedary case). Some people and organizations, notably Microsoft, use the term camel case only for lower camel case. Pascal case means only upper camel case. 

I'm not going to "fix" this here because it's a quote, but I've modified the comment in the `Condition` type to say: "upper CamelCase (PascalCase)" for clarity.

docs/design/resource-states.md
matthchr: It's details of our design. I've moved the `More on conditions` and `Examining other projects like ASO` to above (outside of) the `Proposal` section to make this clearer, as really neither of those sections are actually the proposal.

docs/design/resource-states.md
matthchr: Yes. I've updated the comments here quite a bit based on an actual copy of this code I have in a branch. It should be clearer now.

docs/design/resource-states.md
matthchr: Fixed

docs/design/resource-states.md
matthchr: Fixed this and clarified (it's supposed to be PascalCase which sometimes is also referred to as upper CamelCase or just CamelCase apparently in the Go community (as I've seen at least 2 instances of that)).

docs/design/resource-states.md
matthchr: See my comment here: https://github.com/Azure/azure-service-operator/pull/1614#issuecomment-875930699

I've added this as an open question and we should decide which we prefer. CAPI says it just changes when `Status` changes, but I checked their actual implementation and it really changes anytime the condition changes at all (even if it's from status False -> status False.

The KEP for conditions says:
> 	// Last time the condition transitioned from one status to another.
	// This should be when the underlying condition changed.  If that is not known, then using the time when the API field changed is acceptable

Which makes me think we should make it as you comment here...

docs/design/resource-states.md
matthchr: Fixed

docs/design/resource-states.md
matthchr: Rephrased this to say that

docs/design/resource-states.md
matthchr: Fixed

docs/design/resource-states.md
matthchr: Done

docs/design/resource-states.md
matthchr: I've updated to add a new section called Open questions (answered), where I posed and answered this question. Have updated the comment to match what you suggested.

docs/design/resource-states.md
matthchr: Fixed

docs/design/resource-states.md
theunrepentantgeek: Huh. I remember those terms back as far as my Uni days. Good enough.

theunrepentantgeek (approved):
Looks good to me. Thanks for the updates.

CecileRobertMichon:
> Do we want to define LastTransitionTime as the last time anything changed, or just the last time the Status field changed?

FWIW, this is how CAPI answers that question:

```
// NOTE: If a condition already exists, the LastTransitionTime is updated only if a change is detected
// in any of the following fields: Status, Reason, Severity and Message.
```

matthchr:
Thanks @CecileRobertMichon - I've also adopted that approach.

docs/design/resource-states.md
babbageclunk: I can see that Reason is redundant when Ready is true (it's like that Tolstoy quote about happy families), but I think the simplicity of having it always present is worth that wrinkle.

babbageclunk (approved):

PR #1616, opened 2021-07-01
Inject Hub() function on hub storage resources

theunrepentantgeek:

To support use of the `Convertible` interface, the hub storage version needs to implement a `Hub()` function. This pipeline stage modifies resources that have been flagged as "The storage version" by injecting this necessary function.

gif

**If applicable**:
- [x] this PR contains tests

Requires PR #1606 to be merged first, as this reuses test support functions from that PR, and then rebasing on `master`.

Harvested from PR #1533.


hack/generator/pkg/codegen/golden_files_test.go
```diff
@@ -153,6 +153,7 @@ func NewTestCodeGenerator(testName string, path string, t *testing.T, testConfig
 			"injectOriginalVersionFunction",
 			"injectOriginalVersionProperty",
 			"injectPropertyAssignmentFunctions",
+			pipeline.InjectHubFunctionStageId,
```
Porges, line 156: makes me thing a pipeline should just be a type `{ string name; run(Types, (idFactory?)) Types}` and then this would be `injectHubFunction.name`

Porges (approved):

hack/generator/pkg/codegen/golden_files_test.go
theunrepentantgeek: I don't see how this would work - whether the factory methods return functions or objects, we don't have a reference to the value here, so we can't dereference it to access name. #pleaseexplainmore

PR #1618, opened 2021-07-04
Convert Direction from enum into a discriminated union

theunrepentantgeek:

Code that needed to make a decision based on the direction of conversion always needed to handle "other" values with a panic; converting the type into an interface allows us to ensure that only valid implementations are provided, simplifying consumption.

This type will be used more frequently in upcoming PRs to implement type conversion.

Harvested from PR #1533

**Prerequisites**
- [x] PR #1606 to merge 
- [x] Rebasing on `master`

Let us take a moment and lament how `SelectString()` and `SelectType` would be a single method if Go had generics right now.

gif

### DIFF
--- a/hack/generator/pkg/conversions/property_assignment_function.go
+++ b/hack/generator/pkg/conversions/property_assignment_function.go
@@ -154,15 +154,10 @@ func (fn *PropertyAssignmentFunction) Equals(f astmodel.Function) bool {
 
 // AsFunc renders this function as an AST for serialization to a Go source file
 func (fn *PropertyAssignmentFunction) AsFunc(generationContext *astmodel.CodeGenerationContext, receiver astmodel.TypeName) *dst.FuncDecl {
-	var description string
-	switch fn.direction {
-	case ConvertFrom:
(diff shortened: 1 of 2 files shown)


hack/generator/pkg/conversions/property_assignment_function.go
```diff
@@ -268,19 +263,14 @@ func (fn *PropertyAssignmentFunction) generateAssignments(
 // createConversions iterates through the properties on our receiver type, matching them up with
 // our other type and generating conversions where possible
 func (fn *PropertyAssignmentFunction) createConversions(receiver astmodel.TypeDefinition) error {
-	var sourceEndpoints map[string]ReadableConversionEndpoint
-	var destinationEndpoints map[string]WritableConversionEndpoint
+	// When converting FROM, otherDefinition.Type() is our source
+	// When converting TO, receiver.Type() is our source
+	// and conversely for our destination
+	sourceType := fn.direction.SelectType(fn.otherDefinition.Type(), receiver.Type())
+	destinationType := fn.direction.SelectType(receiver.Type(), fn.otherDefinition.Type())
```
babbageclunk, line 270: This bit is fiddly but I think still better than the original.

babbageclunk (approved):

PR #1619, opened 2021-07-05
Ensure DB connections are closed when we've finished with them

### LINKED ISSUE #1576: Bug: PostgreSqlUser opens more and more connections
**Describe the bug**
User is created and permissions are granted. PGSqlConnection idles. During time more and more connections are created and all idles. It takes some time but then the database accepts no more connection because of 100 idle connections

**To Reproduce**
Create a PostgreSqlUser CR and wait for provisioned and a little bit more.

**Expected behavior**
User is created and permissions are granted. PGSqlConnection is closed.

**Screenshots**
grafik

**Additional context**
I guess the connections are from this part:

https://github.com/Azure/azure-service-operator/blob/master/pkg/resourcemanager/psql/psqluser/psqluser_reconcile.go

There are just a view users

grafik

babbageclunk:
Closes #1576 

Before this change, the different resources would leak database connections whenever they were reconciled. Add `defer db.Close()` to all of the places that open database connections so that they are closed at the end of the method.

I started off making a cache to keep the `*sql.DB`s in so that they could be reused between reconciliations, but then the difficulty was ensuring we didn't keep the DBs in the cache indefinitely. One option was to hook up the database controller so it could remove the cached entry when the database was deleted, but that's quite fiddly architecturally. I also wrote [some code](https://github.com/Azure/azure-service-operator/compare/master...babbageclunk:bug/postgres-connections) to clear DBs from the cache if they haven't been requested after a fixed amount of time, which would work but would be really difficult to test (to an extent that is probably not worth it). 

gif

### DIFF
--- a/controllers/mysql_combined_test.go
+++ b/controllers/mysql_combined_test.go
@@ -27,7 +27,7 @@ func TestMySQLHappyPath(t *testing.T) {
 	ctx := context.Background()
 
 	// Add any setup steps that needs to be executed before each test
-	rgLocation := "eastus2"
+	rgLocation := "westus2"
 	rgName := tc.resourceGroupName
 	mySQLServerName := GenerateTestResourceNameWithRandom("mysql-srv", 10)
 	mySQLReplicaName := GenerateTestResourceNameWithRandom("mysql-rep", 10)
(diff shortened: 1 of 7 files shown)


Porges (approved):

theunrepentantgeek (approved):
Looks good. :shipit:

matthchr (approved):

johgoe:
@matthchr @babbageclunk  Is there already a timeline for the next release, which includes this fix?

matthchr:
@johgoe There's no hard deadline but I would imagine sometime in the next week. Are you waiting for this change? (I assume yes given you're asking here).

johgoe:
@matthchr Thank you for you fast reply. Currently we have to restart our PostgreSQL Server too often. We are happy to hear that the release should be available sometime next week.

PR #1621, opened 2021-07-05
Fix subtest name generation

Porges:
Subtests stared with `ParallelSubtests` were sharing the same `Namer` as parent tests, which is a) wrong and b) could lead to race conditions.

Each subtest now gets its own `Namer`, based on its test name.

Regenerated recordings for affected tests. I also removed 2 more headers from recordings to reduce recording diff size.


theunrepentantgeek (approved):
Looks good to me. :shipit:

babbageclunk (approved):

PR #1622, opened 2021-07-05
Move Property Assignment Function into the functions package

theunrepentantgeek:

Moves `PropertyAssignmentFunction` into the `functions` package so that it's more discoverable and reusable. Required removing some functionality from the function so that it could stay in the `conversions` package to avoid circular references.

Reuse of this function is required by WIP to create a `ConvertResourceFunction`, necessary for implementing the `Convertible` interface.

Harvested from PR #1533

**Prerequisites**
- [x] Merge PR #1606
- [x] Merge PR #1618 
- [ ] Rebase on `master`

gif

**If applicable**:
- [x] this PR contains tests


hack/generator/pkg/conversions/property_conversions.go
```diff
@@ -145,6 +145,15 @@ func CreateTypeConversion(
 	return nil, err
 }
 
+
+func NameOfPropertyAssignmentFunction(name astmodel.TypeName, direction Direction, idFactory astmodel.IdentifierFactory) string {
+	nameOfOtherType := idFactory.CreateIdentifier(name.Name(), astmodel.Exported)
+	return direction.SelectString(
+		"AssignPropertiesFrom"+nameOfOtherType,
+		"AssignPropertiesTo"+nameOfOtherType)
```
babbageclunk, line 153: Maybe clearer as:
```suggestion
	return "AssignProperties" + direction.SelectString("From", "To") + nameOfOtherType
```

babbageclunk: Actually maybe `.SelectString("From", "To")` could be built-in? Although that's not too bad as-is.

hack/generator/pkg/conversions/readable_conversion_endpoint.go
```diff
@@ -60,3 +60,41 @@ func MakeReadableConversionEndpointForValueFunction(
 func (r ReadableConversionEndpoint) String() string {
 	return r.description
 }
+
+// Reader generates an expression to read our endpoint
```
babbageclunk, line 64:
```suggestion
// Read generates an expression to read our endpoint
```

babbageclunk (approved):

hack/generator/pkg/conversions/property_conversions.go
theunrepentantgeek: I'm doing a Code Gardening PR that cleans up a few things like this, will include there.

hack/generator/pkg/conversions/readable_conversion_endpoint.go
theunrepentantgeek: Fixed.

PR #1623, opened 2021-07-05
Pin golangci-lint version

Porges:
This caused CI failure recently due to new linters being added.


theunrepentantgeek (approved):
:shipit:

PR #1625, opened 2021-07-06
Support user specified MySQLServer secrets

### LINKED ISSUE #1590: MySQL Server username and password should be able to be specified by the user
ASO autogenerates a username and password for admin access to the database, which is stored in a secret which ASO creates.

**Describe the improvement**
The customer should be allowed to provide a reference to a Kubernetes secret in the same namespace for the password, and instead of autogenerating username and password we will use the provided one. We also need to consider rollover cases to ensure that this pattern will work for rollover.

For the username it should be optionally specified on the MySQLServer itself (probably not secret reference required as the username isn't really a secret)

Open questions:
1. Do we allow secret references across namespaces, or only within the namespace that the resource is deployed?

matthchr:
Closes #1590

 - The specified secret must be a Kubernetes secret.
 - The specified secret must contain a "username" and "password" field.
 - The specified secret must be in the same namespace as the MySQLServer.
 - If the specified secret doesn't exist, reconciliation will be blocked
   until the secret does exist. Once the secret is created, reconciliation
   will continue as normal.
 - The operator does not make the user specified secret owned by
   the MySQLServer.
 - The operator still creates a secret containing connection string details
   and username/password for the server. This secret is named as it was
   before. This means that the customer specified username and password
   are consumed to create this secret, but other resources such as MySQLUser
   still consume the generated secret file.

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/controllers/mysql_combined_test.go
+++ b/controllers/mysql_combined_test.go
@@ -10,12 +10,14 @@ import (
 	"reflect"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
 	azurev1alpha1 "github.com/Azure/azure-service-operator/api/v1alpha1"
 	"github.com/Azure/azure-service-operator/api/v1alpha2"
+	"github.com/Azure/azure-service-operator/pkg/helpers"
 	"github.com/Azure/azure-service-operator/pkg/resourcemanager/mysql"
 	"github.com/Azure/azure-service-operator/pkg/resourcemanager/mysql/mysqluser"
 	"github.com/Azure/azure-service-operator/pkg/secrets"
@@ -113,7 +115,7 @@ func TestMySQLHappyPath(t *testing.T) {
 }
 
 func RunMySQLUserHappyPath(ctx context.Context, t *testing.T, mySQLServerName string, mySQLDBName string, rgName string) {
-	assert := assert.New(t)
+	assert := require.New(t)
 
 	// Create a user in the DB
 	username := GenerateTestResourceNameWithRandom("user", 10)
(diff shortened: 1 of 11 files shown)


docs/services/mysql/mysql.md
```diff
@@ -17,15 +21,16 @@ The value for kind, `MySQLServer` is the Custom Resource Definition (CRD) name.
...
-- `fullyqualifiedservername` : Fully qualified name of the MySQL server such as mysqlserver.mysql.database.azure.com
-- `mysqlservername` : MySQL server name
-- `username` : Server admin
-- `password` : Password for the server admin
-- `fullyqualifiedusername` : Fully qualified user name that is required by some apps such as <username>@<mysqlserver>
+- `fullyQualifiedServerName`: Fully qualified name of the MySQL server. Example: `mysqlserver.mysql.database.azure.com`.
+- `mySqlServerName`: MySQL server name.
+- `username`: Server admin account name.
+- `password`: Server admin account password.
+- `fullyQualifiedUsername`: Fully qualified user name that is required by some apps. Example: `<username>@<mysqlserver>`.
```
theunrepentantgeek, line 33: Suggest turning this into a table:

```suggestion
| Field                      | Content                                                  | Example                                 |
| -------------------------- | -------------------------------------------------------- | --------------------------------------- |
| `fullyQualifiedServerName` | Fully qualified name of the MySQL server.                | `mysqlserver.mysql.database.azure.com`. |
| `mySqlServerName`          | MySQL server name.                                       |                                         |
| `username`                 | Server admin account name.                               |                                         |
| `password`                 | Server admin account password.                           |                                         |
| `fullyQualifiedUsername`   | Fully qualified user name that is required by some apps. | `<username>@<mysqlserver>`.             |
```

pkg/resourcemanager/mysql/mysqlaaduser/reconcile.go
```diff
@@ -199,7 +199,7 @@ func (m *MySQLAADUserManager) GetServer(ctx context.Context, resourceGroupName,
 	// they're not used getting the server.
 	// TODO: This feels a bit dodgy, consider taking secret client and
 	// scheme just so we can pass them in here.
-	client := mysqlserver.NewMySQLServerClient(m.Creds, nil, nil)
+	client := mysqlserver.NewMySQLServerClient(m.Creds, nil, nil, nil)
```
theunrepentantgeek, line 202: Not a fan of all the **nil** values - they're very obscure.

theunrepentantgeek (approved):
Looks good, only minor comments.

pkg/resourcemanager/mysql/server/reconcile.go
```diff
@@ -41,27 +52,42 @@ func (m *MySQLServerClient) Ensure(ctx context.Context, obj runtime.Object, opts
...
 	// If a replica is requested, ensure that source server is specified
-	if createmode == mysql.CreateModeReplica {
+	if createMode == mysql.CreateModeReplica {
 		if len(instance.Spec.ReplicaProperties.SourceServerId) == 0 {
 			instance.Status.Message = "Replica requested but source server unspecified"
 			return true, nil
 		}
 	}
 
+	adminCreds, err := m.GetUserProvidedAdminCredentials(ctx, instance)
```
babbageclunk, line 68: Should we make this resource reconcile when the secret is updated? Are people going to expect the user details to be updated in that case? (Is updating the credentials like that supported in ARM?)

pkg/resourcemanager/mysql/server/reconcile.go
```diff
@@ -325,46 +349,104 @@ func (m *MySQLServerClient) UpdateServerNameInSecret(ctx context.Context, secret
...
-	secret["mySqlServerName"] = []byte(instance.Name)
+	// Populate secret fields derived from username and password
+	secret[UsernameSecretKey] = []byte(username)
+	secret[PasswordSecretKey] = []byte(password)
+	secret[FullyQualifiedUsernameSecretKey] = []byte(makeFullyQualifiedUsername(username, instance.Name))
+	secret[MySQLServerNameSecretKey] = []byte(instance.Name)
 	return secret, nil
 }
+
+type MySQLCredentials struct {
```
babbageclunk, line 416: Doc comment.

babbageclunk (approved):
This looks great. I wonder whether there's a way to have a change to the secret trigger a reconcile?

pkg/resourcemanager/mysql/server/reconcile.go
matthchr: We probably should, yes. It's a bit of feature creep because we didn't support rolling credentials before (since the user couldn't specify them), but really we should now.

I'll look into it but may do it as a separate PR as it'll have another set of tests ensuring that the update actually works, and the actual watching of that secret may add some complexity.

matthchr: I've filed #1634 - will follow up with a separate PR tackling that.

controllers/mysql_combined_test.go
```diff
@@ -197,3 +199,81 @@ func RunMySQLUserHappyPath(ctx context.Context, t *testing.T, mySQLServerName st
...
+
+	// Create the mySQLServer object and expect the Reconcile to be created
+	mySQLServerInstance := v1alpha2.NewDefaultMySQLServer(mySQLServerName, rgName, rgLocation)
+	mySQLServerInstance.Spec.AdminSecret = adminSecretName
+	RequireInstance(ctx, t, tc, mySQLServerInstance)
+
+	// Create a mySQL replica
+	mySQLReplicaInstance := v1alpha2.NewReplicaMySQLServer(mySQLReplicaName, rgName, rgLocation, mySQLServerInstance.Status.ResourceId)
+	mySQLReplicaInstance.Spec.StorageProfile = nil
+	EnsureInstance(ctx, t, tc, mySQLReplicaInstance)
```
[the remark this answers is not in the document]
matthchr: The replica has two "modes":
1. The old mode, where it reads the username and password from the primary secret created by ASO.
2. The new mode, where if _on the replica_ the user specifies `AdminSecret`, it instead gets the secret directly from that location. It still reads the primary secret created by ASO for all of the other fields it needs (URLs mostly).

My intent here was to test mode 1, but actually I can do that faster by just asserting on the shape of the generated secret after creating the primary server. As long as the generated secret looks the same as it does when not using `AdminSecret` on the primary, the replica should work.

For mode 2, I didn't write a test for it. The additional code is pretty trivial (a single `if`) and as you say, additional creations of replicas are burdensome because of how long they take to make.

I'll change this to just assert the shape of the secret from the primary is as expected and drop the replica creation part.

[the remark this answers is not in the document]
matthchr: Actually you're right, I can just remove this entirely because the MySQLUser test already requires the username/password so this is already being verified by that.

pkg/resourcemanager/mysql/server/reconcile.go
matthchr: Fixed (and the public method below)

pkg/resourcemanager/mysql/mysqlaaduser/reconcile.go
matthchr: Fixed this.

pkg/resourcemanager/mysql/mysqluser/mysqluser.go
```diff
@@ -56,7 +56,9 @@ func (m *MySqlUserManager) GetDB(ctx context.Context, resourceGroupName string,
 
 // GetServer retrieves a server
 func (m *MySqlUserManager) GetServer(ctx context.Context, resourceGroupName, serverName string) (mysqlmgmt.Server, error) {
-	client := mysqlserver.NewMySQLServerClient(m.Creds, m.SecretClient, m.Scheme)
+	// TODO: It's only ok to pass nil for KubeReader here because we know it's not needed to perform GET server.
```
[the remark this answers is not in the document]
matthchr: Fixed this.

docs/services/mysql/mysql.md
matthchr: Done

PR #1627, opened 2021-07-07
Create conversion graph

theunrepentantgeek:
**What this PR does / why we need it**:

Adds a pipeline stage that computes the conversion graph required for our storage versioning.

**Prerequisites**
- [x] Merge PR #1618 
- [x] Merge PR #1622
- [x] Merge PR #1644
- [x] Rebase on `master`

**How does this PR make you feel**:
gif

**If applicable**:
- [x] this PR contains tests

PR #1628, opened 2021-07-07
Implement the Convertible interface on non-hub resources

theunrepentantgeek:

Implements the required `ConvertTo()` and `ConvertFrom()` methods needed by the `conversions.Convertible` interface used by the controller

Note that the stage is disabled for two reasons. First is that we don't want to activate this functionality until post `codegen-alpha-0`. Second is that we need to turn on a bunch of features together, otherwise the internal validation of the controller will reject our webhooks.

**Prerequisites**

- [x] Merge PR #1618
- [x] Merge PR #1622 
- [x] Merge PR #1627 
- [x] Merge PR #1644
- [x] Rebase on `master`

gif

**If applicable**:
- [x] this PR contains tests


hack/generator/pkg/astmodel/interface_implementation.go
```diff
@@ -57,6 +57,11 @@ func (iface *InterfaceImplementation) References() TypeNameSet {
 	return results
 }
 
+// FunctionCount returns the number of included functions
+func (iface *InterfaceImplementation) FunctionCount() int {
```
matthchr, line 61: Would it make more sense to just expose this as `Functions()` like exists on `ResourceType`? (Can we do that since this is also embedded there?)

hack/generator/pkg/codegen/pipeline/implement_convertible_interface.go
```diff
@@ -0,0 +1,81 @@
...
+
+func createConvertibleInterfaceImplementation(
+	name astmodel.TypeName,
+	resource *astmodel.ResourceType,
+	conversionGraph *storage.ConversionGraph,
+	idFactory astmodel.IdentifierFactory) *astmodel.InterfaceImplementation {
+	var conversionFunctions []astmodel.Function
+
+	for _, fn := range resource.Functions() {
+		if propertyAssignmentFn, ok := fn.(*functions.PropertyAssignmentFunction); ok {
```
matthchr, line 73: A comment explaining what the relationship between the PropertyAssignmentFunctions and and Convertible might be useful here?

matthchr, line 73: Is it expected that there are resources without a `PropertyAssignmentFunctions`, or should that be an error condition?

hack/generator/pkg/codegen/pipeline/testdata/TestInjectConvertibleInterface-v20211231storage.golden
```diff
@@ -0,0 +1,41 @@
+// Code generated by azure-service-operator-codegen. DO NOT EDIT.
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+package v20211231storage
```
matthchr, line 4: Are all 4 of these golden files generated from that 1 test? (I think yes based on the name?). I think we have an item for this already but they should be in their own subfolder.

hack/generator/pkg/functions/resource_conversion_function_test.go
```diff
@@ -0,0 +1,107 @@
...
+	. "github.com/onsi/gomega"
+
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/astmodel"
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/conversions"
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/test"
+)
+
+// Test_ResourceConversionFunction_DirectConversion_GeneratesExpectedCode tests the code when the ConvertTo() and
+// ConvertFrom() functions are directly converting to/from the Hub type, without any intermediate step.
+func Test_ResourceConversionFunction_DirectConversion_GeneratesExpectedCode(t *testing.T) {
```
matthchr, line 20: Same question if we can use one of the standard json file based golden tests once this pipeline stage is actually fully enabled.

If so leave a TODO here as well?

hack/generator/pkg/functions/resource_conversion_function.go
```diff
@@ -0,0 +1,308 @@
...
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/astbuilder"
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/astmodel"
+)
+
+// ResourceConversionFunction implements conversions to/from our hub type
+// Existing PropertyAssignment functions are used to implement stepwise conversion
+//
+// Direct conversion from the hub type:
+//
+// func (r <receiver>) ConvertFrom(hub conversion.Hub) error {
```
matthchr, line 23: minor: use the `From|To` notation you used below to clarify that this can also happen in either direction?

hack/generator/pkg/functions/resource_conversion_function.go
```diff
@@ -0,0 +1,308 @@
...
+//
+// Indirect conversion, multiple steps via an intermediate instance
+//
+// func (r <receiver>) Convert<From|To>(hub conversion.Hub) error {
+// 	var source <vNext>
+// 	err := source.ConvertFrom(hub)
+// 	if err != nil {
+// 		return errors.Wrap(err, "converting from hub to source")
+// 	}
+// 	err = person.AssignPropertiesFromPerson(&source)
```
matthchr, line 39: `From|To`

hack/generator/pkg/functions/resource_conversion_function.go
matthchr: Or maybe because you aren't using it super consistently below either (and doing so might make the examples awkward for example in the error messages, etc) instead you could just preface this with something like: "Examples show the `From` direction. The only difference between the two is that all instances of `From` are replaced with `To` in the `To` direction." or something?

hack/generator/pkg/functions/resource_conversion_function.go
```diff
@@ -0,0 +1,308 @@
...
+// Existing PropertyAssignment functions are used to implement stepwise conversion
+//
+// Direct conversion from the hub type:
+//
+// func (r <receiver>) ConvertFrom(hub conversion.Hub) error {
+// 	source, ok := hub.(*<hub.Type>)
+// 	if !ok {
+// 		return fmt.Errorf("expected <hub.Type> but received %T instead", hub)
+// 	}
+// 	return person.AssignPropertiesFromPerson(source)
```
matthchr, line 28: Is `person` here supposed to be `r` or something? It's not clear where it came from. Same in the example below.

matthchr: If you want to use a placeholder for the `receiver` type, then maybe also show that same placeholder in the assignment functions?

`r.AssignPropertiesFrom<Receiver>(source)`?

hack/generator/pkg/functions/resource_conversion_function.go
```diff
@@ -0,0 +1,308 @@
...
+}
+
+// directConversion creates a simple direct conversion between the two types
+//
+// <local>, ok := <hubAsInterface>.(<actualHubType>)
+// if !ok {
+//     return errors.Errorf("expected <actualHubType> but received %T instead", <hubAsInterface>)
+// }
+//
+// return receiver.AssignProperties(To|From)<type>(<local>)
```
matthchr, line 140: minor: `<receiver>` here?

matthchr (approved):
Some minor comments but looks good overall

hack/generator/pkg/functions/resource_conversion_function_test.go
theunrepentantgeek: No. The purpose of the test is to verify the behavour of the stage in isolation, to help troubleshoot any issues as specific to this stage.

hack/generator/pkg/codegen/pipeline/testdata/TestInjectConvertibleInterface-v20211231storage.golden
theunrepentantgeek: I've worked out how to put them into a subfolder but the change has a wide blast radius due to the change being needed in a shared method from the `test` package. I'll keep the changes out of this PR and will do them separately.

hack/generator/pkg/codegen/pipeline/implement_convertible_interface.go
```diff
@@ -0,0 +1,81 @@
...
+func ImplementConvertibleInterface(idFactory astmodel.IdentifierFactory) Stage {
+
+	stage := MakeStage(
+		ImplementConvertibleInterfaceStageId,
+		"Implement the Convertible interface on each non-hub Resource type",
+		func(ctx context.Context, state *State) (*State, error) {
+			injector := astmodel.NewInterfaceInjector()
+
+			modifiedTypes := make(astmodel.Types)
+			resources := storage.FindResourceTypes(state.Types())
```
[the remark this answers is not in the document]
matthchr: Would you consider pulling it forward? I've got at least one upcoming PR where I'd like to use the Status one.

hack/generator/pkg/codegen/pipeline/implement_convertible_interface.go
matthchr: but this function isn't ever invoked with hub resources anyway, because of the
```
				if resource.IsStorageVersion() {
					// The hub storage version doesn't implement Convertible
					continue
				}
```
in the loop yes?

That's why I was wondering if you wanted to tolerate types without those functions or if it made sense to just fail fast if you found one since that meant something suspicious was going on.

hack/generator/pkg/functions/resource_conversion_function.go
matthchr: Yeah FWIW I agree. In theory I like `<From|To>` but in practice it makes things a bit harder to read. I think my main thing here is just be consistent, if you don't use them that's fine you can just have a preface or something that explains _roughly_ how the other direction might look. The comments are meant to help us understand the code rather than be an exact, faithful replica of the template that is being applied so I think either is ok.

theunrepentantgeek: At the moment I'm leaning towards keeping them in - it's still easier to parse the comment with `<To|From>` than the actual code, and it cues readers into the complexity that's present.

Issue #1591, opened 2021-06-22
Improve generated typenames 


Porges:
**Describe the current behavior**
See: https://github.com/Azure/azure-service-operator/pull/1548#discussion_r647855845

Includes depluralization of parent type names when creating type names for subresources.

Porges:
Marking as `alpha-1` since the generated names do not affect CRDs, only the intermediate/internal go code.

theunrepentantgeek:
From #1459:

Consider performing the pluralization improvements step for spec and status of resources too.
Add a golden test for pluralization?

theunrepentantgeek:
Whether we still need to do this depends on the results of #1758.

matthchr:
Still need to do this, but not critical because we don't give code-dependency guarantees right now

PR #1631, opened 2021-07-07
Improve flattening

Porges:
1. Flattening needed to visit arrays, maps, etc to match up inner types.
2. Added a panic if we have missed flattening any properties marked `flatten: true`; improve reporting of panics during code generation process to include package/typename.
3. If types are not modified then keep using the original type. This improves the typenames a bit (#1591).
4. When matching up spec/status types to augment specs with status, ignore any optionality when matching up types. This makes things work properly with the handcrafted schemas where often the `required` attribute has been forgotten.


hack/generator/pkg/codegen/pipeline/flatten_properties.go
```diff
@@ -67,17 +67,28 @@ func makeFlatteningVisitor(defs astmodel.Types) astmodel.TypeVisitor {
 
 			return result, nil
 		},
-		VisitFlaggedType: func(this *astmodel.TypeVisitor, it *astmodel.FlaggedType, ctx interface{}) (astmodel.Type, error) {
-			// skip ARM types, do not flatten
-			if it.HasFlag(astmodel.ARMFlag) {
```
matthchr, line 72: Seems sorta weird to visit ARM types and then unflatten them later? Why bother visiting them at all?

hack/generator/pkg/codegen/pipeline/status_augment.go
```diff
@@ -63,32 +64,37 @@ func fuseAugmenters(augments ...augmenter) augmenter {
...
-			// when we merge two typenames we know that (structurally) they must
-			// be the ‘same’ type, even if they have different names
+		// need to resolve main type
+		merger.Add(func(main astmodel.TypeName, swagger astmodel.Type) (astmodel.Type, error) {
+			// Connascence alert! This is in coöperation with the code in
+			// determine_resource_ownership.go:updateChildResourceDefinitionsWithOwner
+			// which requires the typenames to identify which ChildResources match
+			// which Resources. The ChildResource types should never actually be
+			// used, so it's okay that we don't visit them here.
+			if strings.HasSuffix(main.Name(), "ChildResource") {
```
matthchr, line 83: minor: You could stick this in a const so that somebody interested could also find all usages of the cost which in theory would take them to the other location as well (although I think the comment is great)

matthchr (approved):

hack/generator/pkg/codegen/pipeline/flatten_properties.go
Porges: Instead of skipping them here I remove the `flatten: true` on any properties in an ARM type. This lets us capture if we haven't handled any `flatten: true` properties later.

hack/generator/pkg/codegen/pipeline/status_augment.go
```diff
@@ -63,32 +64,37 @@ func fuseAugmenters(augments ...augmenter) augmenter {
 
 // flattenAugmenter copies across the "flatten" property from Swagger
 func flattenAugmenter(allTypes astmodel.ReadonlyTypes) augmenter {
```
theunrepentantgeek, line 66: I've read this a dozen times and I still have no idea what this method is doing. 

Strongly suggest extracting giving all those anonymous functions some intention revealing names.

theunrepentantgeek (changes requested):

hack/generator/pkg/codegen/pipeline/flatten_properties.go
matthchr: Ahh somehow I was thinking that `removeFlatten` was actually unflattening them - I see it now. Resolving this.

hack/generator/pkg/codegen/pipeline/status_augment.go
Porges: can you take another look after the big comment expansion I’ve done? the functions are highly contextual; in a language with proper pattern-matching they would distinct branches in a function; here they could be given names like `flattenAugment_handleTypeTypeName` and `flattenAugment_handleTypeNameType` but I don’t think that’s going to help much

hack/generator/pkg/astmodel/allof_type.go
```diff
@@ -134,15 +134,15 @@ func (allOf AllOfType) RequiredPackageReferences() *PackageReferenceSet {
 // Equals returns true if the other Type is a AllOf that contains
 // the same set of types
 func (allOf *AllOfType) Equals(t Type) bool {
+	if allOf == t {
+		return true // short-circuit
+	}
+
```
theunrepentantgeek, line 140: Won't this always fail because (here) they're different types? I thought the type conversion was needed first.

hack/generator/pkg/codegen/pipeline/status_augment.go
theunrepentantgeek: That's much more decipherable now! Thanks.

theunrepentantgeek (approved):

hack/generator/pkg/astmodel/allof_type.go
Porges: It will do a comparison with what is inside the interface, so equivalent to doing a pointer comparison.

PR #1635, opened 2021-07-08
Put short-circuiting code in one place

Porges:
A little tidy-up.


babbageclunk (approved):
Looks great! This way around makes it a much nicer idiom in the codebase.

hack/generator/pkg/astmodel/validated_type.go
```diff
@@ -156,6 +156,16 @@ func (v *ValidatedType) ElementType() Type {
...
 
+func (v *ValidatedType) WithElement(t Type) Type {
+	if v.element.Equals(t) {
+		return v
+	}
+
+	result := *v
+	result.element = t
+	return &result
+}
```
theunrepentantgeek, line 167: When cloning types, hadn't we agreed on using a memberwise `copy()` func to ensure all the fields were handled correctly? ('cause we had some issues early on with "copies" that shared state.)

theunrepentantgeek (commented):
Overall, I like the approach of ensuring we share more immutable values, reducing memory churn. Do any of your new withers need to use an internal `copy()` func to follow our prior conventions?

hack/generator/pkg/astmodel/optional_type.go
```diff
@@ -19,16 +19,32 @@ type OptionalType struct {
...
-var _ MetaType = &OptionalType{}
+// cache of commonly-used values
+var cachedOptionals = map[Type]*OptionalType{
+	BoolType:   {BoolType},
+	FloatType:  {FloatType},
+	IntType:    {IntType},
+	StringType: {StringType},
+	UInt32Type: {UInt32Type},
+	UInt64Type: {UInt64Type},
+}
```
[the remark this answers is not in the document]
Porges: I just thought it might be nice to help avoid many “duplicate” instantiations, and will speed up equality checks and suchlike slightly. From my numbers it saves about five thousand allocations.

hack/generator/pkg/astmodel/validated_type.go
Porges: Here we don’t touch any of the interior-mutable types like maps or slices, so it is safe. I’d prefer not to pessimize this…

theunrepentantgeek: Fair enough.

theunrepentantgeek (approved):
Looks good to me.

Giphy

matthchr (approved):

PR #1637, opened 2021-07-09
Improve failures in recorded tests

Porges:
Change record-replay errors to not panic but instead return an error which the controller can handle to set an Error-level condition on the resource. This should cause the tests to fail in a more appropriate location and we will know where exactly in the test the problem occurred.


hack/generated/controllers/generic_controller.go
```diff
@@ -90,7 +90,7 @@ func registerWebhook(mgr ctrl.Manager, obj runtime.Object) error {
 		Complete()
 }
 
-func RegisterAll(mgr ctrl.Manager, applier armclient.Applier, objs []runtime.Object, log logr.Logger, options Options) error {
+func RegisterAll(mgr ctrl.Manager, applier armclient.Applier, objs []runtime.Object, options Options) error {
```
Porges, line 93: removed `log` parameter here as there is already one in `Options`

hack/generated/pkg/reconcilers/deployment_error_classifier.go
```diff
@@ -1,3 +1,8 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
```
Porges, line 4: how?!

theunrepentantgeek (commented):
Looks good so far.

hack/generated/pkg/reconcilers/deployment_error_classifier.go
matthchr: #1643

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
```diff
@@ -109,18 +109,34 @@ func NewAzureDeploymentReconciler(
 	}
 }
 
+func (r *AzureDeploymentReconciler) checkForFatalReconciliationError(err error) {
```
matthchr, line 112: minor: Maybe leave a comment here explaining what actually can trigger these?

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
```diff
@@ -109,18 +109,34 @@ func NewAzureDeploymentReconciler(
...
 	if err != nil {
 		r.log.Error(err, "error determining create or update action")
 		r.recorder.Event(r.obj, v1.EventTypeWarning, "DetermineCreateOrUpdateActionError", err.Error())
+
+		r.checkForFatalReconciliationError(err)
+
 		return ctrl.Result{}, err
 	}
 
+	r.log.Info("Reconciling Azure resource %s, will perform: %s", r.obj.AzureName(), action)
```
matthchr, line 131: Is this equivalent to `V(0)`?

hack/generated/pkg/testcommon/kube_test_context_envtest.go
```diff
@@ -75,15 +77,18 @@ func createEnvtestContext(perTestContext PerTestContext) (*KubeBaseTestContext,
...
 		controllers.GetKnownStorageTypes(),
-		perTestContext.logger,
 		controllers.Options{
 			CreateDeploymentName: func(obj metav1.Object) (string, error) {
 				// create deployment name based on test name and kubernetes name
 				result := uuid.NewSHA1(uuid.Nil, []byte(perTestContext.TestName+"/"+obj.GetNamespace()+"/"+obj.GetName()))
 				return fmt.Sprintf("k8s_%s", result.String()), nil
 			},
 			RequeueDelay: requeueDelay,
+			Options: controller.Options{
```
matthchr, line 87: Should we specify the max concurrent reconciles here too?

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
```diff
@@ -109,18 +109,34 @@ func NewAzureDeploymentReconciler(
 	}
 }
 
+func (r *AzureDeploymentReconciler) checkForFatalReconciliationError(err error) {
+	var fatal FatalReconciliationError
+	if errors.As(err, &fatal) {
+		r.SetResourceProvisioningState(armclient.FailedProvisioningState)
+		r.SetResourceError(fatal.Message)
```
Porges, line 116: @matthchr do we need to call `Update` after doing this?

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
Porges: I think there might not be any `V`s on this type, but will check

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
matthchr: Yes, there's not an update baked in at the top level in `generic_controller` right now, so we own calling it before we return if we want to.

matthchr: and in order for modifications to be persisted, we must call update*

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
```diff
@@ -108,20 +108,56 @@ func NewAzureDeploymentReconciler(
...
+// checkForFatalReconciliationError:
+// - at the moment this is only raised during tests if we determine that a record-replay
+//   test has gone off the rails, but it will be useful for scenarios in the real world as well
+func (r *AzureDeploymentReconciler) checkForFatalReconciliationError(ctx context.Context, err error) error {
+	var fatal FatalReconciliationError
+	if !errors.As(err, &fatal) {
+		return err
+	}
+
+	condition := r.PositiveConditions.MakeFalseCondition(
```
theunrepentantgeek, line 120: This is weird? Is transiting through `PositiveConditions` the only way to make a negative one?

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
```diff
@@ -108,20 +108,56 @@ func NewAzureDeploymentReconciler(
 	}
 }
 
+// checkForFatalReconciliationError:
+// - at the moment this is only raised during tests if we determine that a record-replay
+//   test has gone off the rails, but it will be useful for scenarios in the real world as well
+func (r *AzureDeploymentReconciler) checkForFatalReconciliationError(ctx context.Context, err error) error {
```
theunrepentantgeek, line 114: This method is named as though it's a predicate that returns a bool, but does something quite other. 

I'm having trouble coming up with a different name, and I believe that's because the method is doing two different things - checking to see whether there's a fatal error, and storing that error on a condition.

Suggest replacing this method with two. 

Firstly, a conversion method:
``` go
func AsFataReconciliationError(err error) (FatalReconciliationError, bool)
```

Secondly, a method to store the error as a condition:
``` go
func (r *AzureDeploymentReconciler) SetErrorCondition(err error)
```

You can then replace all calls to `checkForFatalReconciliationError` with:
``` go
if fatal, ok := AsFatalReconciliationError(err); ok {
    r.SetErrorCondition(fatal)
}
```

theunrepentantgeek (approved):
Looks good. A couple of minor comments, including one on method naming/design.

gif

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
Porges: quote from `positive_condition_builder`:

> PositiveConditionBuilder makes positive polarity conditions. A positive polarity condition is a condition where Status==True is the "happy path" state, and Status==False is a failure state.

Our condition "Ready" is positive; this sets Ready to False.

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
Porges: right but then I’m going to have to repeat that everywhere… I’ll still split it up into smaller pieces but I want to retain the check & apply method

Porges: how about `Turn(Appropriate)ErrorsIntoConditions` 😛

ideally we'd have a top-level place where we could do a single check for this and then write it inline like you had, but it's not easy when we are working inside the controller-runtime framework

Porges: …which I have done!

theunrepentantgeek (approved):

PR #1638, opened 2021-07-09
Add tparallel linter and fix

Porges:
If subtests are going to run in parallel the parent test needs to be marked as well; this linter catches it.


theunrepentantgeek (approved):
Good catch. #shipit

Rocket Launch

PR #1639, opened 2021-07-09
Enforce alternatives to %v and %+v

### LINKED ISSUE #1585: Add linter to catch use of %v and %+v
**Describe the improvement**

Those format specifiers can cause leakage of PII in logs, which is a Bad Thing(tm).

We've eliminated all the uses of `%v` and `%+v` but should now turn on the linter to make sure we stay clean.

theunrepentantgeek:

The `%v` and `%+v` format specifiers blindly dump all of the contents of any struct into the string - which can result in disclosure of secrets and other PII into log files where it is uncontrolled; this is a potential security and GDPR issue we need to prevent.

This PR adds a check to enforce that the format specifiers `%v` and `%+v` are not used. 

Closes #1585 

gif


api/v1alpha2/mysqlserver_types.go
```diff
@@ -33,7 +33,7 @@ type MySQLServerSpec struct {
 	// If AdminSecret is specified but a secret with the given name is not found in the same namespace
 	// as the MySQLServer, then reconciliation will block until the secret is created.
 	// If this is not specified, a username and password will be automatically generated.
-	AdminSecret            string               `json:"adminSecret,omitempty"`
+	AdminSecret string `json:"adminSecret,omitempty"`
```
matthchr, line 36: oops

matthchr: The fact that the old ASO build lets me merge this angers me... maybe will make a fix to that.

hack/generator/pkg/astmodel/kubebuilder_validations.go
```diff
@@ -37,20 +37,34 @@ func GenerateKubebuilderComment(validation KubeBuilderValidation) string {
...
+	switch v := value.Interface().(type) {
+	case int, int16, int32, int64:
+		return fmt.Sprintf("%d", v)
+	case bool:
+		return fmt.Sprintf("%t", v)
+	case string:
+		return v
+	default:
+		klog.Error(fmt.Sprintf("unexpected value for kubebuilder comment - %s", value.Kind()))
+		return "%%UNKNOWN%%"
```
matthchr, line 64: I think it's appropriate to panic here rather than emit incorrect code

pkg/helpers/retry_test.go
```diff
@@ -16,7 +16,7 @@ func TestRetryTimeout(t *testing.T) {
 	})
 	stop := time.Now().Sub(start)
 	if stop < 5*time.Second {
-		t.Errorf("retry ended too soon: %v", stop)
+		t.Errorf("retry ended too soon: %s", stop)
```
matthchr, line 19: Not that there's any issue with this change, but wondering if we want to just exclude `_test` files from the check?

pkg/resourcemanager/azuresql/azuresqluser/azuresqluser.go
```diff
@@ -11,6 +11,7 @@ import (
 
 	azuresql "github.com/Azure/azure-sdk-for-go/services/preview/sql/mgmt/v3.0/sql"
 	_ "github.com/denisenkom/go-mssqldb"
+	"github.com/pkg/errors"
```
matthchr, line 14: without an amendment to the makefile (that builds old ASO), these can regress. I don't see that in this PR.

matthchr: The Taskfile targets don't run for these packages because they're not in `hack`

matthchr (approved):

pkg/helpers/retry_test.go
theunrepentantgeek: If we have a `%v` in a test that dumps, say, the API key that we use for testing, wouldn't that be an issue?  The whole PR is about being a bit paranoid about disclosures, so I'm happier checking the `_test` files too.

hack/generator/pkg/astmodel/kubebuilder_validations.go
theunrepentantgeek: Changed.

pkg/resourcemanager/azuresql/azuresqluser/azuresqluser.go
theunrepentantgeek: I'll make a TODO for following up with changes to the Makefile to match.

Or do we want to migrate ASO to the same Task approach?

PR #1640, opened 2021-07-12
Eliminate most use of hard coded strings for pipeline stage names

theunrepentantgeek:

Largely eliminates the use of magic strings to identify pipeline stages, which will help with future stability. 

For a couple of stages where the prerequisites are intrinsic to correct behaviour, I moved the declaration of the prerequisite into the stage itself where it is less likely to be inadvertently removed.

Also includes a handful of other code gardening changes.

**Prerequisites**

- [x] Merge #1627 
- [x] Merge #1628 
- [x] Merge #1644
- [x] Rebase on `master`

gif


hack/generator/pkg/codegen/golden_files_test.go
```diff
@@ -146,41 +146,42 @@ func NewTestCodeGenerator(testName string, path string, t *testing.T, testConfig
...
-			//pipeline.InjectHubFunctionStageId,
-			"reportTypesAndVersions")
+			pipeline.DeleteGeneratedCodeStageID,
+			pipeline.CheckForAnyTypeStageID,
+			pipeline.CreateStorageTypesStageID,
+			pipeline.InjectOriginalGVKFunctionStageID,
+			pipeline.InjectOriginalVersionFunctionStageID,
+			pipeline.InjectOriginalVersionPropertyStageID,
+			pipeline.InjectPropertyAssignmentFunctionsStageID,
+			//pipeline.InjectHubFunctionStageID,
```
matthchr, line 156: minor: Is this commented out for a reason? I think yes and it will be enabled later? Might be worth a TODO here

matthchr (approved):
LGTM!

hack/generator/pkg/codegen/golden_files_test.go
theunrepentantgeek: I don't think we'll want to include this stage in tests, but we can't remove it until it's formally included in the pipeline. Comment added.

matthchr (approved):

PR #1642, opened 2021-07-12
Update generated servicebus queue/topic to 2021 API

### LINKED ISSUE #1555: Use the 2021-01-01-preview version of ServiceBus API for topics/queue's
Currently using the 2018-01-01-preview because there is not support in the ServiceBus RP for 2021-01-01-preview. They are releasing it as we speak though so we should be able to start using it for topics and queue's as well.

matthchr:
Closes #1555


theunrepentantgeek (approved):
Assuming the recorded tests are correct, looks good to me.

PR #1644, opened 2021-07-13
Introduce Pipeline State

theunrepentantgeek:

Modify our pipeline to have a more state that's passed from stage to stage, giving us more flexibility in what each stage can do - now they can do more than just return a modified set of `TypeDefinitions`.

gif


hack/generator/pkg/codegen/pipeline/state.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/astmodel"
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/codegen/storage"
+)
+
+// State is an immutable instance that captures the information being passed along the pipeline
+type State struct {
+	types           astmodel.Types           // set of types generated so far
+	conversionGraph *storage.ConversionGraph // graph of transitions between packages in our conversion graph
```
matthchr, line 16: This feels sorta awkward because the notion of a pipeline is agnostic of what steps it contains, but this `State` assumes certain steps, which makes the pipeline less generic than it would otherwise be. This state is passed through all the pipeline stages but it's hard to tell when authoring a stage which of the state variables will be set and which will not - you need to know where in the pipeline your stage will go and then look around into the definitions of other stages to see who sets what (I'm assuming a world where there is more state than there is now).

matthchr: I wonder if we'd rather have a less typesafe (but more generic) `stageData map[string]interface{}` which contained data from a given stage (indexed by the stage name). Then stages could declare they have pre-requisites and then look up the data from their prerequisite stage via its name (which they will already have because they defined it as a pre-requisite as well?). This feels like it would play nicer with the existing graph-ordering features we have (pre/post-requisite), while keeping the pipeline agnostic of what specific stages it is comprised of.

The other possibility rather than a state object that contains information about the type graph (which is my interpretation of what `conversionGraph` is), is to encode the information into `types` and avoid `state` altogether. @porges and I have run into situations where we wanted to include metadata on a specific type so that future stages would know what to do for cases like flattening or references. I believe we both ended up making use of property annotations (in our cases the data was per-property) for that case. The downside of property annotations is that they're rendered in the generated code, which we don't want to require in all cases. Would some sort of generic "annotations/metadata on a type" solve your problem here? That would allow you to reconstruct the graph in a future stage rather than passing it forward.

theunrepentantgeek: The purist in me considered some kind of generic solution, but the pragmatist in me won the argument (by a narrow margin) that it was something we didn't need _yet_. 

I like the idea of a later stage being able to request _"give me the information produced by **this** prior stage"_, it would be a good way to solve the problem when we need a more generic solution.

matthchr (approved):
I agree with the desire to have some sort of higher level state transference between pipeline stages, but do think it should be made more generic so as to not encode specifics about the pipeline into the generic pipeline API.

With that said, I am fine with this for now as an intermediate step.

PR #1646, opened 2021-07-13
Update ASO v1 dependencies

matthchr:
We're specifically interested in updating controller-runtime, as that gives us access to some new filtered watches which we may need.


controllers/async_controller.go
```diff
@@ -9,12 +9,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/Azure/azure-service-operator/pkg/resourcemanager"
-	"github.com/Azure/azure-service-operator/pkg/resourcemanager/config"
-	"github.com/Azure/azure-service-operator/pkg/secrets"
-	keyvaultsecretlib "github.com/Azure/azure-service-operator/pkg/secrets/keyvault"
-	telemetry "github.com/Azure/azure-service-operator/pkg/telemetry"
-	multierror "github.com/hashicorp/go-multierror"
+	"github.com/hashicorp/go-multierror"
```
theunrepentantgeek, line 12: Why use this vs `"k8s.io/apimachinery/pkg/util/errors"` which we've used elsewhere for error aggregation?

controllers/azurevirtualmachine_controller.go
```diff
@@ -17,7 +19,7 @@ type AzureVirtualMachineReconciler struct {
 // +kubebuilder:rbac:groups=azure.microsoft.com,resources=azurevirtualmachines,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=azure.microsoft.com,resources={azurevirtualmachines/status,azurevirtualmachines/finalizers},verbs=get;update;patch
 
-func (r *AzureVirtualMachineReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+func (r *AzureVirtualMachineReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
theunrepentantgeek, line 22: Do you not get lint errors about `ctx` being unused on all of these?

theunrepentantgeek (approved):
A couple of questions, nothing major.

gif

controllers/async_controller.go
matthchr: This is V1 ASO, it used this package and not the other one we use for V2. Didn't make sense to change it since I'm just updating dependencies right now not trying to change functionality.

controllers/azurevirtualmachine_controller.go
matthchr: I don't think we have such a linter for ASO v1. Although I am pretty sure we don't have one for ASO v2 either? You have a pretty good point that I should pass it down though...

matthchr: I confirmed that we do not have a linter that catches this currently enabled for the generator or generated (V2 ASO).
I have also passed this ctx to the function called here, which means it's now plumbed all the way through (previously the function called here manufactured a context from `context.Background()`)

PR #1648, opened 2021-07-14
Update ASO V2 dependencies

matthchr:


hack/generated/controllers/generic_controller.go
```diff
@@ -194,7 +193,9 @@ func (gr *GenericReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	// https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/controllers.md, which says:
 	// Never mutate original objects! Caches are shared across controllers, this means that if you mutate your "copy"
 	// (actually a reference or shallow copy) of an object, you'll mess up other controllers (not just your own).
-	obj = obj.DeepCopyObject()
+	obj = obj.DeepCopyObject().(client.Object)
+
+	gr.Log.V(0).Info("Reconcile invoked", "kind", fmt.Sprintf("%T", obj))
```
matthchr, line 198: minor: Is there a cleaner way to log the type being reconciled in the `logr.Logger` interface (since it doesn't seem to support format specifiers?)

hack/generated/go.sum
```diff
@@ -6,18 +6,39 @@ cloud.google.com/go v0.44.2/go.mod h1:60680Gw3Yr4ikxnPRS/oxxkBccT6SA1yMk63TGekxK
 cloud.google.com/go v0.45.1/go.mod h1:RpBamKRgapWJb87xiFSdk4g1CME7QZg3uwTez+TSTjc=
 cloud.google.com/go v0.46.3 h1:AVXDdKsrtX33oR9fbCMu/+c1o8Ofjq6Ku/MInaLVg5Y=
 cloud.google.com/go v0.46.3/go.mod h1:a6bKKbmY7er1mI7TEI4lsAkts/mkhTSZK8w33B4RAg0=
+cloud.google.com/go v0.50.0/go.mod h1:r9sluTvynVuxRIOHXQEHMFffphuXHOMZMycpNR5e6To=
```
matthchr, line 9: Sure whatever you say golang

hack/generated/controllers/generic_controller.go
theunrepentantgeek: The best suggestion I have is to break out an intention revealing local:

```suggestion
    objKind := fmt.Sprintf("%T", obj)
	gr.Log.V(0).Info("Reconcile invoked", "kind", objKind)
```

Another way would be to introduce a function on each of our generated types that generates a description to use here, but that's a lot of overhead.

hack/generated/pkg/testcommon/test_context.go
```diff
@@ -75,9 +75,19 @@ func (tc TestContext) ForTest(t *testing.T) (PerTestContext, error) {
 		return PerTestContext{}, errors.Wrapf(err, "creating ARM client")
 	}
 
+	// As of https://github.com/Azure/go-autorest/releases/tag/autorest%2Fv0.11.9, the autorest
+	// client reuses HTTP clients among instances. We add handlers to the HTTP client based on
+	// the specific test in question, which means that clients cannot be reused.
+	// We explicitly create a new http.Client so that one tests recordings don't
```
theunrepentantgeek, line 81: #nitpick #innerpedant
```suggestion
	// We explicitly create a new http.Client so that the recording from one test doesn't
```

theunrepentantgeek (approved):
A couple of questions, but nothing of import.

hack/generated/controllers/controller_resources.go
```diff
@@ -7,19 +7,20 @@ package controllers
...
+func GetKnownStorageTypes() []client.Object {
 	knownTypes := getKnownStorageTypes()
 
 	knownTypes = append(knownTypes, new(resources.ResourceGroup))
 
 	return knownTypes
 }
 
-func GetKnownTypes() []runtime.Object {
+func GetKnownTypes() []client.Object {
```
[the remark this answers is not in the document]
matthchr: `client.Object` is actually a combo of `runtime.Object` and `metav1.Object`. You can see them explain it [here](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.9.2/pkg/client/object.go#L45).

This is actually a nice win because we actually want to make sure all of our CRD types are `client.Object` (have both `runtime,Object` and `metav1.Object`), so using this everywhere helps ensure that.

hack/generated/controllers/crd_disk_test.go
```diff
@@ -53,8 +53,7 @@ func Test_Disk_CRUD(t *testing.T) {
 	disk.Spec.NetworkAccessPolicy = &networkAccessPolicy
 	patcher.Patch(disk)
 
-	objectKey, err := client.ObjectKeyFromObject(disk)
-	tc.Expect(err).ToNot(HaveOccurred())
```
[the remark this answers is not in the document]
Porges: it’s better when there’s no result and then you can `tc.Expect(x.SomeCall()).To(Succeed())`

Porges (approved):

hack/generated/pkg/testcommon/test_context.go
matthchr: Fixed

PR #1649, opened 2021-07-14
Implement conversions for Specs and Statuses

theunrepentantgeek:

Defines the new interfacea `CovertibleSpec` and `ConvertibleStatus` that will be consumed by the generic controller, and implements them on our generated types.

Prerequisites
- [x] PR #1627 
- [x] PR #1628
- [x] PR #1640 
- [x] Rebase on `master`

gif

**If applicable**:
- [x] this PR contains tests


hack/generated/pkg/genruntime/convertible_spec.go
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package genruntime
+
+type ConvertibleSpec interface {
```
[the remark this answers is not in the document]
babbageclunk: From talking with you, I understand that the From and To directions are needed because without it there'd need to be circular references between different versions. Can you add a comment to that effect? :grin:

hack/generator/pkg/codegen/pipeline/implement_convertible_status_interface.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/astmodel"
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/conversions"
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/functions"
+)
+
+// ImplementConvertibleStatusInterfaceStageId is the unique identifier for this pipeline stage
+const ImplementConvertibleStatusInterfaceStageId = "implementConvertibleStatusInterface"
+
+func ImplementConvertibleStatusInterface(idFactory astmodel.IdentifierFactory) Stage {
```
babbageclunk, line 21: The code of this and the Spec stage are basically the same, with just the word "spec" replaced with "status". Can this just be one stage factory that gets passed in some kind of strategy that encapsulates which of spec or status should be used?

hack/generator/pkg/functions/pivot_conversion_function.go
```diff
@@ -0,0 +1,173 @@
...
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/conversions"
+)
+
+// PivotConversionFunction implements a pivot that's used when a conversion reaches a hub type but hasn't yet found the
+// innermost required conversion needed.
+//
+// We use this for Spec types, implementing genruntime.ConvertibleSpec, and for Status types, implementing
+// genruntime.ConvertibleStatus.
+//
+// If the two instances involved in conversion are not in the same "spoke" leading to this "hub" version, we'll pivot to
```
babbageclunk, line 24: I got confused and asked you about this and you explained it very nicely with a diagram, then you said you should add some ascii art to this.

hack/generator/pkg/functions/pivot_conversion_function.go
```diff
@@ -0,0 +1,173 @@
...
+	// idFactory is a reference to an identifier factory used for creating Go identifiers
+	idFactory astmodel.IdentifierFactory
+}
+
+// Ensure we properly implement the function interface
+var _ astmodel.Function = &PivotConversionFunction{}
+
+// NewSpecPivotConversionFunction creates a pivot conversion function that works to convert between two Spec types
+// implementing the interface genruntime.ConvertibleSpec.
+// hubType is the TypeName of our hub type
```
babbageclunk, line 51: `hubType` isn't mentioned in this function - is this from a previous version?

hack/generator/pkg/functions/pivot_conversion_function.go
```diff
@@ -0,0 +1,173 @@
...
+		direction:     direction,
+		idFactory:     idFactory,
+	}
+
+	return result
+}
+
+// NewStatusPivotConversionFunction creates a pivot conversion function that works to convert between two Status types
+// implementing the interface genruntime.ConvertibleStatus.
+// hubType is the TypeName of our hub type
```
babbageclunk, line 69: Ditto.

hack/generator/pkg/functions/pivot_conversion_function.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+func (fn *PivotConversionFunction) declarationDocComment(receiver astmodel.TypeName, parameter string) string {
+	return fn.direction.SelectString(
+		fmt.Sprintf("populates our %s from the provided %s", receiver.Name(), parameter),
+		fmt.Sprintf("populates the provided %s from our %s", parameter, receiver.Name()))
+}
+
+func (fn *PivotConversionFunction) Equals(otherFn astmodel.Function) bool {
+	rcf, ok := otherFn.(*ChainedConversionFunction)
```
babbageclunk, line 166:
```suggestion
	rcf, ok := otherFn.(*PivotConversionFunction)
```

babbageclunk (approved):
This looks great - I think there's maybe a little bit of code duplication between spec and status stages that could be removed like you've done in the conversion functions.

hack/generator/pkg/functions/pivot_conversion_function.go
theunrepentantgeek: Added comment and diagram.

hack/generator/pkg/functions/pivot_conversion_function.go
theunrepentantgeek: Yup. Fixed up the doccomment.

hack/generator/pkg/functions/pivot_conversion_function.go
theunrepentantgeek: Fixed.

hack/generator/pkg/functions/pivot_conversion_function.go
theunrepentantgeek: Fixed.

hack/generator/pkg/functions/pivot_conversion_function.go
matthchr: A link to the design doc on this section with a non-ascii diagram would also work

hack/generator/pkg/functions/pivot_conversion_function.go
```diff
@@ -0,0 +1,204 @@
...
+//   --> v2storage.Person.ConvertTo(v2.Person)
+//     --> v3storage.Person.ConvertTo(v2.Person)              // Pivot!
+//       --> v2.Person.ConvertFrom(v3storage.Person)          // Change of direction
+//         --> v2storage.Person.ConvertFrom(v3storage.Person)
+//
+// Note that conversions like this always pivot through the hub version.
+//
+type PivotConversionFunction struct {
+	// nameFrom is the name for this function when converting FROM a provided instance
+	nameFrom string
```
matthchr, line 64: I wrote this up before you refactored but then I think it was lost when this file got renamed, but you could instead of having a `nameFrom`/`nameTo`, just assign a single `name` field in the `PivotConversionFunction` constructors.

Something like:
`name: direction.SelectString("ConvertSpecFrom", "ConvertSpecTo")`

It feels a bit weird to have the name dynamically selected at generation time when it's actually known significantly earlier (at constructor called time).

Minor though.

hack/generator/pkg/functions/pivot_conversion_function.go
matthchr: Although I see you've got the ascii diagram already so maybe nevermind =)

hack/generator/pkg/codegen/pipeline/implement_chained_conversion_interface.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/astmodel"
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/conversions"
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/functions"
+)
+
+// ImplementConvertibleSpecInterfaceStageId is the unique identifier for this pipeline stage
+const ImplementConvertibleSpecInterfaceStageId = "implementConvertibleSpecInterface"
+
+func ImplementConvertibleSpecInterface(idFactory astmodel.IdentifierFactory) Stage {
```
matthchr, line 21: This file is `implement_chained_conversion_interface` but the one with status is `implement_convertible_status_interface` - feels like there should be a pattern that isn't being followed by one of them?

matthchr (approved):

hack/generator/pkg/astbuilder/conditions.go
```diff
@@ -45,3 +58,32 @@ func IfNotNil(toCheck dst.Expr, statements ...dst.Stmt) *dst.IfStmt {
...
 	}
 }
+
+// IfOk checks a boolean ok variable and if it is ok runs the given statements
+//
+//	if ok {
+//		<statements>
+//	}
+//
+func IfOk(statements ...dst.Stmt) *dst.IfStmt {
```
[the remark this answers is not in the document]
matthchr: or at least for the first case...

hack/generator/pkg/functions/pivot_conversion_function.go
matthchr: Oh wait apparently that comment was saved - I see it above. It applies here too though I guess.
Github weird after refreshing in the middle of putting comments I guess.

hack/generator/pkg/codegen/pipeline/implement_convertible_status_interface.go
theunrepentantgeek: I tried this refactoring, but the code for the stage factory was pretty hard to read; I think it's better to leave them as separate (but similar) stages.

hack/generator/pkg/functions/chained_conversion_function.go
```diff
@@ -0,0 +1,291 @@
...
+//     }
+//
+//     return r.AssignPropertiesFrom(source)
+// }
+//
+type ChainedConversionFunction struct {
+	// nameFrom is the name for this function when converting FROM a provided instance
+	nameFrom string
+	// nameTo is the name for this function when converting TO a provided instance
+	nameTo string
```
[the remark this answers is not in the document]
theunrepentantgeek: Used to need both names - when the pivot was also implemented by this function, but no longer needed now that's been factored out into a separate type. Will simplify as suggested.

hack/generator/pkg/functions/pivot_conversion_function.go
theunrepentantgeek: No longer need both names, so will simplify as suggested.

hack/generator/pkg/astbuilder/builder.go
```diff
@@ -28,13 +28,17 @@ func CheckErrorAndReturn(otherReturns ...dst.Expr) dst.Stmt {
 // CheckErrorAndWrap checks if the err is non-nil, and if it is returns it, wrapped with additional information
 //
 // if err != nil {
-//      return errors.Wrapf(err, <message>, <args>)
+//      return errors.Wrap(f)(err, <message>, <args>)
```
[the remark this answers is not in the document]
theunrepentantgeek: I've made the comment more explicit by showing both cases.

hack/generator/pkg/astbuilder/builder.go
```diff
@@ -363,6 +367,18 @@ func Selector(expr dst.Expr, names ...string) *dst.SelectorExpr {
 		exprs...).(*dst.SelectorExpr)
 }
 
+// CompareEqual generates a == comparison between the two expressions
+//
+// <lhs> == <rhs>
+//
+func CompareEqual(lhs dst.Expr, rhs dst.Expr) *dst.BinaryExpr {
```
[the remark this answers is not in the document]
theunrepentantgeek: Using `Equal()` collided with the Gomega matcher of the same name during testing, so I went with a different name. I'll rename them to `AreEqual()` and `AreNotEqual()` to improve discoverability.

hack/generator/pkg/astbuilder/conditions.go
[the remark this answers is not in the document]
theunrepentantgeek: Good idea - but with potentially a wider blast radius. I'll take note to do this in a separate code gardening followup PR.

hack/generator/pkg/codegen/golden_files_test.go
```diff
@@ -155,6 +155,9 @@ func NewTestCodeGenerator(testName string, path string, t *testing.T, testConfig
 			pipeline.InjectPropertyAssignmentFunctionsStageID,
 			// TODO: Once the stage is enabled in the pipeline, we may need to remove it here for testing
 			//pipeline.InjectHubFunctionStageID,
+			//pipeline.ImplementConvertibleInterfaceStageId,
```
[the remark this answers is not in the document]
theunrepentantgeek: Yes; we don't want to turn on these two stages until post `codegen-alpha-0`.

hack/generator/pkg/codegen/pipeline/implement_convertible_spec_interface.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/functions"
+)
+
+// ImplementConvertibleSpecInterfaceStageId is the unique identifier for this pipeline stage
+const ImplementConvertibleSpecInterfaceStageId = "implementConvertibleSpecInterface"
+
+func ImplementConvertibleSpecInterface(idFactory astmodel.IdentifierFactory) Stage {
+	stage := MakeStage(
+		ImplementConvertibleSpecInterfaceStageId,
+		"Inject ConvertSpecTo() and ConvertSpecFrom() to implement genruntime.ConvertibleSpec",
```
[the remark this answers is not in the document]
theunrepentantgeek: Good idea. Done.

hack/generator/pkg/codegen/pipeline/implement_convertible_spec_interface.go
```diff
@@ -0,0 +1,85 @@
...
+
+			newTypes := state.Types().OverlayWith(modifiedTypes)
+			return state.WithTypes(newTypes), nil
+		})
+
+	stage.RequiresPrerequisiteStages(InjectPropertyAssignmentFunctionsStageID)
+	return stage
+}
+
+func createConvertibleSpecInterfaceImplementation(
```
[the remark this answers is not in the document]
theunrepentantgeek: It's already documented elsewhere; will add a comment pointing to that location so it's easier to find.

hack/generator/pkg/codegen/pipeline/testdata/TestInjectConvertibleStatusInterface/microsoft.person-v20200101.golden
```diff
@@ -0,0 +1,186 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+type Person struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              Person_Spec   `json:"spec,omitempty"`
+	Status            Person_Status `json:"status,omitempty"`
```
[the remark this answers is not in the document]
theunrepentantgeek: I tried combining the two stages together, but there are enough differences that the resulting stage factory was hard to read; I've therefore left it the way it was.

The two interfaces need to be separate because we want to catch any case where the wrong interface is passed around as early as possible. This is less important right now, but as I write the code to integrate into the generic controller, having the compiler catch any issues will be most useful.

I understand (and agree with) the sentiment that the golden files can be hard to read - but I'm not sure combining them actually makes things easier to understand. I've tried to somewhat mitigate this by adding comments detailing what needs to be checked in each golden file.

hack/generator/pkg/codegen/golden_files_test.go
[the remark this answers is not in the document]
matthchr: I figured but was just doublechecking

hack/generated/pkg/genruntime/convertible_status.go
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package genruntime
+
+type ConvertibleStatus interface {
```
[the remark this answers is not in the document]
matthchr: I was just suggesting that since the size of possible Specs is large and the size of possible "correct types" to pass to one of these functions is small/one, that we're always going to run a large risk of runtime error regardless and so we could optimize for slightly less code if we wanted to.

hack/generator/pkg/codegen/pipeline/testdata/TestInjectConvertibleStatusInterface/microsoft.person-v20200101.golden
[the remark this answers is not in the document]
matthchr: Does it make sense to generate types with just spec or just status having the interface impled, but not the other one? Or is the expectation that these two stages are always used together?

hack/generator/pkg/codegen/testdata/ARMCodeGeneratorPipeline.golden
```diff
@@ -0,0 +1,46 @@
+Expected Pipeline Stages for ARM Code Generation
+------------------------------------------------
+loadSchema                                     Load and walk schema
```
matthchr, line 3: Hmm, this file being added seems wrong? I think it exists already but maybe with a different name?

theunrepentantgeek: Odd, it's definitely surplus. Will remove.

hack/generator/pkg/codegen/pipeline/implement_convertible_status_interface.go
babbageclunk: Fair enough!

PR #1650, opened 2021-07-14
Prune unused types after flattening

matthchr:
* Flattening can cause types which were previously being used to no longer be used. 
   Run a prune after flattening to ensure that we remove these unneeded types.
* Fix bug in Function References() where many functions were incorrectly referring to 
   their receiverType, resulting in types remaining included when they should have been
   removed.
* Fix small bug in embedded type handling exposed by this change so that 
   the golden tests pass.

### DIFF
--- a/hack/generator/pkg/astmodel/one_of_json_marshal_function.go
+++ b/hack/generator/pkg/astmodel/one_of_json_marshal_function.go
@@ -43,10 +43,11 @@ func (f *OneOfJSONMarshalFunction) Equals(other Function) bool {
 	return false
 }
 
-// References returns the set of references for the underlying object.
+// References returns the set of types to which this function refers.
+// SHOULD include any types which this function references but its receiver doesn't.
+// SHOULD NOT include the receiver of this function.
 func (f *OneOfJSONMarshalFunction) References() TypeNameSet {
-	// Defer this check to the owning object as we only refer to its properties and it
-	return f.oneOfObject.References()
+	return nil
 }
 
 // AsFunc returns the function as a go dst
--- a/hack/generator/pkg/codegen/code_generator.go
+++ b/hack/generator/pkg/codegen/code_generator.go
@@ -145,6 +145,9 @@ func createAllPipelineStages(idFactory astmodel.IdentifierFactory, configuration
 		// Effects the "flatten" property of Properties:
 		pipeline.FlattenProperties(),
 
+		// Remove types which may not be needed after flattening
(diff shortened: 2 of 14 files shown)


hack/generator/pkg/codegen/code_generator.go
```diff
@@ -145,6 +145,9 @@ func createAllPipelineStages(idFactory astmodel.IdentifierFactory, configuration
 		// Effects the "flatten" property of Properties:
 		pipeline.FlattenProperties(),
 
+		// Remove types which may not be needed after flattening
+		pipeline.StripUnreferencedTypeDefinitions(),
```
theunrepentantgeek, line 149: Might be an idea to add a postrequisite to the flattening stage to capture the requirement for post-processing.

theunrepentantgeek (approved):
Good catch. A couple of minor comments.

gif

hack/generator/pkg/astmodel/one_of_json_marshal_function.go
```diff
@@ -43,10 +43,11 @@ func (f *OneOfJSONMarshalFunction) Equals(other Function) bool {
...
 }
 
-// References returns the set of references for the underlying object.
+// References returns the set of types to which this function refers.
+// SHOULD include any types which this function references but its receiver doesn't.
+// SHOULD NOT include the receiver of this function.
 func (f *OneOfJSONMarshalFunction) References() TypeNameSet {
-	// Defer this check to the owning object as we only refer to its properties and it
-	return f.oneOfObject.References()
+	return nil
```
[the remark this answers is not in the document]
matthchr: Yeah in the context of this type, the field `oneOfObject` is the receiver.

hack/generator/pkg/codegen/code_generator.go
matthchr: Doing this resulted in:
```
E0715 10:26:59.857238    9739 gen_types.go:33] Error creating code generator: postrequisite "stripUnreferenced" of stage "flattenProperties" satisfied too early
```

I think because we have multiple copies of `stripUnreferenced` and the pre/postrequisite stuff doesn't understand that?

matthchr: Will leave this as is for now, we can investigate improving the pre/post stuff to maybe support this in the future (although there are times where you only want a single copy of a stage... starting to get complicated to express in pre/post?)

PR #1653, opened 2021-07-16
Clean up autogenerated properties

matthchr:
* Improve errors from property type transformers further.
* Add missed reference properties.
* Exclude some readonly properties

### DIFF
--- a/hack/generator/azure-arm.yaml
+++ b/hack/generator/azure-arm.yaml
@@ -15,12 +15,6 @@ typeFilters:
     group: microsoft.servicebus
     version: v*api20210101preview
     because: We want to export this particular preview version
-  # TODO: This is temporarily included because the SB RP doesn't support queue/topic
-  # TODO: in 20210101preview right now.
-  - action: include
-    group: microsoft.servicebus
-    version: v*api20180101preview
-    because: We want to export this particular preview version
   - action: prune
     version: '*preview'
     because: preview SDK versions are excluded by default
@@ -210,7 +204,6 @@ typeTransformers:
     target:
       name: string
     because: Modeling this as an enum doesn't work well in the context of CRDs because new regions are regularly added
-
   - name: "*"
     property: Condition
     ifType:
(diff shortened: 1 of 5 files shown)


hack/generator/azure-arm.yaml
```diff
@@ -15,12 +15,6 @@ typeFilters:
     group: microsoft.servicebus
     version: v*api20210101preview
     because: We want to export this particular preview version
-  # TODO: This is temporarily included because the SB RP doesn't support queue/topic
-  # TODO: in 20210101preview right now.
-  - action: include
-    group: microsoft.servicebus
-    version: v*api20180101preview
-    because: We want to export this particular preview version
```
theunrepentantgeek, line 23: Should this change be in this PR?

hack/generator/pkg/codegen/pipeline/add_cross_resource_references.go
```diff
@@ -18,7 +18,7 @@ import (
 	"github.com/Azure/azure-service-operator/hack/generator/pkg/config"
 )
 
-var armIDDescriptionRegex = regexp.MustCompile("(?i).*/subscriptions/.*?/resourceGroups/.*")
+var armIDDescriptionRegex = regexp.MustCompile("(?i)(.*/subscriptions/.*?/resourceGroups/.*|ARM ID|Resource ID)")
```
theunrepentantgeek, line 21: This regex looks like it is matching a URL - why does it have spaces in it?

hack/generator/pkg/codegen/pipeline/add_cross_resource_references.go
```diff
@@ -196,16 +200,32 @@ func newKnownReferencesMap(configuration *config.Configuration) map[referencePai
 			typeName: astmodel.MakeTypeName(configuration.MakeLocalPackageReference("microsoft.storage", "v1alpha1api20210401"), "VirtualNetworkRule"),
 			propName: "Id",
 		}: true,
+		{
+			typeName: astmodel.MakeTypeName(configuration.MakeLocalPackageReference("microsoft.storage", "v1alpha1api20210401"), "EncryptionIdentity"),
+			propName: "UserAssignedIdentity",
+		}: true,
+		{
+			typeName: astmodel.MakeTypeName(configuration.MakeLocalPackageReference("microsoft.storage", "v1alpha1api20210401"), "ResourceAccessRule"),
```
theunrepentantgeek, line 208: Suggest factoring out repeated package references into locals. #minor

theunrepentantgeek (commented):
Surprised to see no changes to golden files - what effect do these changes have on the generated properties?

hack/generator/azure-arm.yaml
matthchr: Yeah this is the old version that isn't actually needed anymore (we don't export any types from it)

hack/generator/pkg/codegen/pipeline/add_cross_resource_references.go
matthchr: This regex is basically a heurisitc looking through the property comment to guess if it is an ARM ID. It has a URL in it not because the comment must contain a URL, but because many comments for ARM IDs say: "This is of the form: /subscriptions/<>/resourceGroups/<>/microsoft.storage/..."

The other two fields I added are additional detection that are common - the comments often say "ARM ID" or "Resource ID"

This is really a subset of the heuristics I came up with for the Swagger repo linter (which hasn't been merged yet).

matthchr:
@theunrepentantgeek asked:
> Surprised to see no changes to golden files - what effect do these changes have on the generated properties?

Unfortunately, we don't have any golden files for flattening right now, since they're sourced from the Swagger specification which we don't have a way to write golden files test for. The only change in this PR that would impact golden files is that flattening + "reference" fix, which we *should* have a test for but because of the above limitation we don't.

Everything else in this PR is about controlling specific types generated from the specs we're including, or improving error messages, neither of which show in the golden files.

theunrepentantgeek (approved):
Thanks for answering my questions, happy for you to merge.

PR #1654, opened 2021-07-19
Support MySQLServer secret rollover

matthchr:
Allows updates to the user supplied credentials for MySQLServer to change the credentials on the server.

I reworked MySQLServer reconcile a little bit to use some of the newer functionality we have... it should be a little cleaner (and a little more correct) than it was before.

**If applicable**:
- [x] this PR contains tests


controllers/mysqlserver_controller.go
```diff
@@ -24,7 +33,66 @@ func (r *MySQLServerReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 func (r *MySQLServerReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// In order to watch a secret associated with this server:
+	// * We cannot use "For" - this automatically reconciles the type in question, and since the type
+	//   is a secret we don't want that.
+	// * We cannot use "Owns" - this reconciles the resources owner.
+	// * We must use Watches
+
+	// TODO:
+	// In order to actually make a change to the MySQLServer password, you must issue an update to the server which
+	// includes "administratorLogin": "<user>", "administratorLoginPassword": "<password>",
```
babbageclunk, line 44: Is this TODO still to be done?

controllers/mysqlserver_controller.go
```diff
@@ -24,7 +33,66 @@ func (r *MySQLServerReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+
+	// Add a field indexer which we will use to find servers
```
babbageclunk, line 46:
```suggestion
	// Add a field indexer which we will use to find servers by AdminSecret name.
```

controllers/mysqlserver_controller.go
```diff
@@ -24,7 +33,66 @@ func (r *MySQLServerReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+// likely scoped by a label selector? That requires additional work as we would need to manage that label
+// in the MySQLServer reconcile loop probably.
+func makeMapFunc(mgr ctrl.Manager) func(o client.Object) []reconcile.Request {
+	return func(o client.Object) []reconcile.Request {
+		// Safety check that we're looking at a secret, if not nothing to do
+		if _, ok := o.(*corev1.Secret); !ok {
+			return nil
+		}
+
+		// This should be fast since the list if items it's going through should always be cached
```
babbageclunk, line 72:
```suggestion
		// This should be fast since the list of items it's going through should always be cached
```

controllers/mysqlserver_controller.go
```diff
@@ -24,7 +33,66 @@ func (r *MySQLServerReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+		obj := rawObj.(*v1alpha2.MySQLServer)
+		return []string{obj.Spec.AdminSecret}
+	})
+	if err != nil {
+		return err
+	}
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&v1alpha2.MySQLServer{}).
+		Watches(&source.Kind{Type: &corev1.Secret{}}, handler.EnqueueRequestsFromMapFunc(makeMapFunc(mgr))).
```
babbageclunk, line 57: This is neat!

pkg/resourcemanager/mysql/server/reconcile.go
```diff
@@ -52,6 +53,23 @@ func (m *MySQLServerClient) Ensure(ctx context.Context, obj runtime.Object, opts
 		return true, err
 	}
 
+	pClient := pollclient.NewPollClient(m.Creds)
```
babbageclunk, line 56: It's much nicer having this logic pulled out to the top of the method!

pkg/resourcemanager/mysql/server/reconcile.go
```diff
@@ -87,142 +105,107 @@ func (m *MySQLServerClient) Ensure(ctx context.Context, obj runtime.Object, opts
...
 	}
+
+	instance.Status.SetPollingURL(pollURL, api.PollingURLKindCreateOrUpdate)
+	instance.Status.SetProvisioning("request submitted to Azure")
+
 	return false, nil
 }
 
+func (m *MySQLServerClient) calculateHash(spec azurev1alpha2.MySQLServerSpec, secret map[string][]byte) string {
+	return helpers.Hash256(struct {
```
babbageclunk, line 200: Cunning!

babbageclunk (approved):
Looks nice! The reconcile changes make it a bit simpler to follow as well.

controllers/mysqlserver_controller.go
matthchr: Oh, nope. Good eye. This was a note to myself that you must have both. There's a bug (I emailed them) if you omit the username it just hangs forever.

pkg/resourcemanager/mysql/server/reconcile.go
matthchr: I thought it was a bit hacky but didn't have a better way, and it does do the job.

controllers/mysqlserver_controller.go
matthchr: Yeah, it's a bit awkward right now because we end up watching all secrets. I'm not 100% sure we can avoid that, we might be able to with https://github.com/kubernetes-sigs/controller-runtime/blob/master/designs/use-selectors-at-cache.md but I couldn't work out how to do it and it's an optimization not required for functionality so I've left a note but we should be ok for now I think.

PR #1657, opened 2021-07-22
Fail pipeline if any type contains ErroredType with errors

matthchr:
* Adds a new stage requiring that all generated types do not contain an `ErroredType` with errors (warnings are ok). 
* Missing `Status` types now produce an `ErroredType` with errors (before it was warnings). This is required for future work where resource Status's will begin to have more information in addition to just the response from Azure, such as `Conditions` and possibly other things going forward.

### DIFF
--- a/hack/generator/pkg/codegen/pipeline/apply_export_filters.go
+++ b/hack/generator/pkg/codegen/pipeline/apply_export_filters.go
@@ -15,26 +15,28 @@ import (
 	"github.com/Azure/azure-service-operator/hack/generator/pkg/config"
 )
 
+const ApplyExportFiltersStageID = "filterTypes"
+
 // ApplyExportFilters creates a Stage to reduce our set of types for export
 func ApplyExportFilters(configuration *config.Configuration) Stage {
-	return MakeLegacyStage(
-		"filterTypes",
+	return MakeStage(
+		ApplyExportFiltersStageID,
 		"Apply export filters to reduce the number of generated types",
-		func(ctx context.Context, types astmodel.Types) (astmodel.Types, error) {
-			return filterTypes(configuration, types)
+		func(ctx context.Context, state *State) (*State, error) {
+			return filterTypes(configuration, state)
 		})
 }
 
 // filterTypes applies the configuration include/exclude filters to the generated definitions
 func filterTypes(
 	configuration *config.Configuration,
-	definitions astmodel.Types) (astmodel.Types, error) {
+	state *State) (*State, error) {
 
 	newDefinitions := make(astmodel.Types)
 
-	filterer := configuration.BuildExportFilterer(definitions)
+	filterer := configuration.BuildExportFilterer(state.types)
 
-	for _, def := range definitions {
+	for _, def := range state.types {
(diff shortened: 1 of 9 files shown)


Porges (approved):

matthchr:
Going to change this to an "error check phase" rather than something specific to status types. So the stage will just look through all of the types for any `ErroredType` with `len(errors) != 0` catalog them all and fail. This covers status but also other things in the future (in theory).

Porges (approved):

hack/generator/pkg/codegen/pipeline/apply_export_filters.go
```diff
@@ -15,10 +15,12 @@ import (
 	"github.com/Azure/azure-service-operator/hack/generator/pkg/config"
 )
 
+const ApplyExportFiltersStageID = "filterTypes"
+
 // ApplyExportFilters creates a Stage to reduce our set of types for export
 func ApplyExportFilters(configuration *config.Configuration) Stage {
 	return MakeLegacyStage(
```
theunrepentantgeek, line 22: Any chance you've got time to change this to use `MakeStage()` instead? #minor

hack/generator/pkg/codegen/pipeline/verify_no_errored_types.go
```diff
@@ -0,0 +1,102 @@
...
+				_, err := visitor.visitor.Visit(def.Type(), erroredTypeVisitorContext{name: def.Name()})
+				if err != nil {
+					return nil, errors.Wrapf(err, "failed while visiting %q", def.Name())
+				}
+			}
+
+			err := kerrors.NewAggregate(visitor.errs)
+			if err != nil {
+				return nil, err
+			}
```
theunrepentantgeek, line 38: This seems overly complex - why not use an approach similar to the existing `AsObjectType()` and `AsResourceType()` methods? Introduce `AsErroredType()`, use it on `def.Type()` and then access the errors you want right there.

theunrepentantgeek (changes requested):
I'm going to hit _Request Changes_ because I think there's a simpler way to achieve the same result.

hack/generator/pkg/codegen/pipeline/verify_no_errored_types.go
matthchr: Because errors can also be on a `property` in an `ObjectType` or on `spec` or `status` in a `ResourceType`, so you can't just look one level deep to find them. For example today our `Status` error is on `resource.Status`. Unless you're suggesting that we implement a new method on `Type` that's like `Errors()` or something, and then each type can look in its specific viable locations for errors.

theunrepentantgeek: Ahh, that makes sense. Objections withdrawn.

theunrepentantgeek (approved):

hack/generator/pkg/codegen/pipeline/apply_export_filters.go
matthchr: Changed

PR #1658, opened 2021-07-22
Add operator mode configuration

babbageclunk:
This allows running the operator in either webhooks-only or watchers-only mode, as well as the default of running both. There's a new configuration setting `AZURE_OPERATOR_MODE` which defaults to `both` (the current behaviour) but can also be set to either `watchers` or `webhooks`.

This (combined with the [target namespaces configuration](https://github.com/Azure/azure-service-operator/pull/1559)) enables multiple instances of the operator to be deployed in one k8s cluster in a multi-tenant configuration:
* One global instance deployed in webhooks-only mode with `ValidatingWebhookConfiguration` and `MutatingWebhookConfiguration` resources that refer to it.
* One or more instances in watchers-only mode with separate Azure credentials and different target namespaces being monitored.

See the [multitenancy proposal](https://github.com/Azure/azure-service-operator/blob/master/docs/multitenancy/proposal.md) for more detail.

gif

**If applicable**:
- [x] this PR contains tests


controllers/suite_test.go
```diff
@@ -176,752 +139,19 @@ func setup() error {
...
-				ctrl.Log.WithName("controllers").WithName("PostgreSQLFirewallRule"),
-			),
-			Recorder: k8sManager.GetEventRecorderFor("PostgreSQLFirewallRule-controller"),
-			Scheme:   k8sManager.GetScheme(),
-		},
-	}).SetupWithManager(k8sManager)
-	if err != nil {
-		return err
-	}
-
```
matthchr, line 885: I like all this red

pkg/resourcemanager/config/config.go
```diff
@@ -33,10 +33,28 @@ var (
...
+// OperatorMode determines whether we'll run watchers and/or webhooks.
+type OperatorMode string
+
+const (
+	OperatorModeBoth     = OperatorMode("both")
+	OperatorModeWebhooks = OperatorMode("webhooks")
+	OperatorModeWatchers = OperatorMode("watchers")
+)
+
+func (m OperatorMode) IncludesWebhooks() bool {
```
matthchr, line 50: Would this be more cleanly done with an enum that was an `int` done as a bitmask where `Includes(mode OperatorMode)` bool?

pkg/resourcemanager/config/env.go
```diff
@@ -68,6 +68,17 @@ func ParseEnvironment() error {
 	}
 	targetNamespaces = ParseStringListFromEnvironment("AZURE_TARGET_NAMESPACES")
 
+	operatorMode = OperatorMode(envy.Get("AZURE_OPERATOR_MODE", "both"))
```
matthchr, line 71: minor: use the `operatorMode` const value here rather than hardcoding `both`

pkg/resourcemanager/config/env.go
```diff
@@ -68,6 +68,17 @@ func ParseEnvironment() error {
...
+	switch operatorMode {
+	case OperatorModeBoth, OperatorModeWebhooks, OperatorModeWatchers:
+	default:
+		return errors.Errorf(`operator mode must be one of "both", "webhooks" or "watchers" but was %q`, operatorMode)
+	}
+
+	if operatorMode == OperatorModeWebhooks && len(targetNamespaces) > 0 {
```
matthchr, line 78: Doesn't this presuppose that there are only the two modes?

Would this be better written as
```suggestion
	if !operatorMode.IncludesWatchers() && len(targetNamespaces) > 0 {
```

and then change the error message accordingly to say: "mode must contain watchers to specify targetNamespaces"?

matthchr (commented):
Do we need some tests that ensure that this mode works somehow?

pkg/resourcemanager/config/config.go
babbageclunk: My initial reaction is that I don't think so, but I'm just going to try it that way to see.

babbageclunk: It came out pretty nice!

controllers/suite_test.go
```diff
@@ -91,13 +55,21 @@ func setup() error {
...
 	// envy.Set("AZURE_SECRET_NAMING_VERSION", "1")
 
-	err := resourcemanagerconfig.ParseEnvironment()
+	err := config.ParseEnvironment()
 	if err != nil {
 		return err
 	}
 
+	if len(envy.Get("TEST_EMIT_ASO_LOGS", "")) > 0 {
+		ctrl.SetLogger(zap.New(func(o *zap.Options) {
```
matthchr, line 64: Any reason not to just do this all the time?

pkg/resourcemanager/config/env.go
```diff
@@ -157,3 +166,13 @@ func ParseStringListFromEnvironment(variable string) []string {
 	}
 	return items
 }
+
+func ParseOperatorMode(value string) (OperatorMode, error) {
```
matthchr, line 170: minor: Maybe throw a comment on this?

matthchr (commented):
Looks good to me, holding approval just over the enum question but once that's answered can just approve.

pkg/resourcemanager/config/config.go
babbageclunk: D'oh, I rebased this PR onto `upstream/master` from home yesterday without pulling from `origin/operator-mode` first (which would have brought in the enum change). All fixed now.

controllers/suite_test.go
babbageclunk: It's really noisy - it would make the logs huge in the full test run, especially with them all running in parallel.

pkg/resourcemanager/config/env.go
babbageclunk: Oh, the updated version does have a comment! :)

azure-pipelines.yml
```diff
@@ -180,6 +180,46 @@ jobs:
...
+    - script: |
+        set -e
+        export PATH=$PATH:$(go env GOPATH)/bin:$(go env GOPATH)/kubebuilder/bin
+        export KUBEBUILDER_ASSETS=$(go env GOPATH)/kubebuilder/bin
+        make test-webhooks-only-mode
+      displayName: Run webhooks-only mode tests
+      condition: or(eq(variables['check_changes.SOURCE_CODE_CHANGED'], 'true'), eq(variables['Build.SourceBranch'], 'refs/heads/master'))
+      continueOnError: 'false'
+      env:
+        GO111MODULE: on
```
matthchr, line 194: I feel great sadness for the way this job is now. (That's not your fault nor did I expect you to fix it here, it just means I wish it were better!).

pkg/resourcemanager/config/operatormode.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// String converts the mode into a readable value.
+func (m OperatorMode) String() string {
+	switch m {
+	case OperatorModeWatchers:
+		return "watchers"
+	case OperatorModeWebhooks:
+		return "webhooks"
+	case OperatorModeBoth:
```
matthchr, line 41: Read in context, `operatorMode=both` is maybe a bit weird in that it begs the question "both what"?

This also assumes there shall be no further modes, as if there are `both` becomes confusing. What does it mean when there are 3 modes?

Wondering if we want `both` to instead be like `webhooksAndWatchers` or `webhooks|watchers` or something?

matthchr (approved):

pkg/resourcemanager/config/operatormode.go
babbageclunk: Yeah, fair point - how about `webhooks-and-watchers`? I'd rather it was all lowercase, and using bitwise-or makes it read kind of the opposite from the actual meaning.

babbageclunk: And it's ok if it's verbose since it's the default.

PR #1661, opened 2021-07-23
Replace boolean isSpecType with enum

matthchr:
Just some small code cleanup

### DIFF
--- a/hack/generator/pkg/astmodel/armconversion/shared.go
+++ b/hack/generator/pkg/astmodel/armconversion/shared.go
@@ -21,7 +21,7 @@ type conversionBuilder struct {
 	armTypeIdent               string
 	codeGenerationContext      *astmodel.CodeGenerationContext
 	idFactory                  astmodel.IdentifierFactory
-	isSpecType                 bool
+	typeKind                   TypeKind
 	methodName                 string
 	kubeType                   *astmodel.ObjectType
 	armType                    *astmodel.ObjectType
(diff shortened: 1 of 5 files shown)


hack/generator/pkg/astmodel/armconversion/shared.go
```diff
@@ -31,9 +31,9 @@ type conversionBuilder struct {
 type TypeKind int
 
 const (
-	OrdinaryType TypeKind = iota
-	SpecType
-	StatusType
+	TypeKindOrdinary TypeKind = iota
+	TypeKindSpec
+	TypeKindStatus
```
theunrepentantgeek, line 36: Suggest using the type name as a suffix - `OrdinaryTypeKind`, `SpecTypeKind`, and `StatusTypeKind` - as that's easier to read. 

If you limited the scope of the enum by making it internal, you could even get away with `ordinary`, `spec` and `status` as the values.

theunrepentantgeek: Oh, I see elsewhere that they're used outside the package - but my first suggestion is still useful.

theunrepentantgeek (approved):
Looks good - one minor comment about naming.

gif

hack/generator/pkg/astmodel/armconversion/shared.go
matthchr: We don't always put the enum type in the enum value name, but when we do it's as a prefix as far as I can tell. See `ValidationKind`, `GenerationPipeline`, `ExportFilter`, `TypeFilter`, etc.

There also seems to be at least some of this standard in the community as well, seeing as the [proposal for Go2](https://github.com/golang/go/issues/19814) also uses prefixes (in its example).

I wasn't able to find anywhere we used a suffix (there are places where we don't prefix at all). For now I think I'm going to leave this as is. We can maybe have an overall style discussion on what we want to do here (I'm happy to change this I just want it to match what we do elsewhere too).

PR #1663, opened 2021-07-26
Move kind targets to Taskfile

matthchr:
- Also fix localhost docker registry

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -11,11 +11,15 @@ vars:
   CONTROLLER_ROOT: ./hack/generated/
   CONTROLLER_PACKAGE: github.com/Azure/azure-service-operator/hack/generated
 
+  SCRIPTS_ROOT: ./scripts
+
   # TODO: Once the v2 controller is the main controller we can drive this from git tags
   # via e.g. `git describe --tags --dirty` which would give better info for dev builds.
   # When doing this, also update the generator (currently using SHORT_VERSION below).
   VERSION: v2.0.0-alpha
   CONTROLLER_DOCKER_IMAGE: azure-service-operator:{{.VERSION}}
+
+  LOCAL_REGISTRY_CONTROLLER_DOCKER_IMAGE: localhost:5000/azure-service-operator:latest
 
   CROSSPLANE_APP: crossplane-gen
   CROSSPLANE_ROOT: ./hack/crossplane/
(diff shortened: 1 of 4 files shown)


Taskfile.yml
```diff
@@ -177,14 +181,15 @@ tasks:
     - "controllers/**/*.go"
     - "pkg/**/*.go"
     cmds:
-    - go build -ldflags "-X {{.CONTROLLER_PACKAGE}}/pkg/version.BuildVersion={{.VERSION}}" -o ./bin/{{.CONTROLLER_APP}}
+    - CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags "-X {{.CONTROLLER_PACKAGE}}/pkg/version.BuildVersion={{.VERSION}}" -o ./bin/{{.CONTROLLER_APP}}
```
theunrepentantgeek, line 184: Do we absolutely need to nail down the `OS` and `ARCH` like this?

scripts/deploy_testing_secret.sh
```diff
@@ -9,10 +9,7 @@ set -o pipefail
 
 REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 
-bins="${REPO_ROOT}/hack/tools/bin"
-k="$bins/kubectl"
-
-cat <<EOF | $k apply -f -
+cat <<EOF | kubectl apply -f -
```
theunrepentantgeek, line 12: Why do we need to pipe an empty stream into `kubectl`?

babbageclunk: The syntax is ~~a bit~~ very weird but this is saying "cat everything up to the EOF marker (on line 23) to `kubectl apply -f -`".
https://stackoverflow.com/questions/2500436/how-does-cat-eof-work-in-bash

babbageclunk (approved):
These make iterating on the controller (like adding debugging) really nice!

scripts/deploy_testing_secret.sh
babbageclunk: I think people sometimes prefer this to the alternative `kubectl apply -f - <<EOF` because it lets you have the input of the pipeline on the left (although you still have the heredoc below it). It makes a bit more sense when there are more commands in the pipeline.

matthchr: Yeah my intent here as well was not to rewrite this script, just fix it to work in a taskfile rather than a Makefile.

Taskfile.yml
matthchr: I think we want to set them. According to [this](https://www.digitalocean.com/community/tutorials/building-go-applications-for-different-operating-systems-and-architectures), `GOOS` and `GOARCH` are used to set the _target_ OS and Arch. Since the intent of this binary is that it be used and run inside of the container built in the `controller:docker-build` step, which is a Linux container, I think it makes sense to force Linux here so that the result of this step will always work with the `controller:docker-build` step.

This command should work just fine on Windows - it will just produce a binary targeting Linux (which is what we want AFAIK).

theunrepentantgeek: Given that the binary is going to be used inside the container, this makes sense. :+1:

scripts/deploy_testing_secret.sh
theunrepentantgeek: Aside from cursing the arcane syntax choices of bash, all good here.

PR #1666, opened 2021-07-27
Fix webhooks in ASOv2

matthchr:
Fixes webhook configuration in ASOv2.

When we add conversion webhooks we will need to revisit this as they aren't currently hooked up.

### DIFF
--- a/hack/generated/config/certmanager/certificate.yaml
+++ b/hack/generated/config/certmanager/certificate.yaml
@@ -1,0 +1,26 @@
+# The following manifests contain a self-signed issuer CR and a certificate CR.
+# More document can be found at https://docs.cert-manager.io
+# WARNING: Targets CertManager 1.1 check https://docs.cert-manager.io/en/latest/tasks/upgrading/index.html for
+# breaking changes
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: selfsigned-issuer
+  namespace: system
+spec:
+  selfSigned: {}
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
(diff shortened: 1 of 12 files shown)


hack/generated/config/certmanager/certificate.yaml
```diff
@@ -0,0 +1,28 @@
+# The following manifests contain a self-signed issuer CR and a certificate CR.
+# More document can be found at https://docs.cert-manager.io
+# WARNING: Targets CertManager 0.11 check https://docs.cert-manager.io/en/latest/tasks/upgrading/index.html for
+# breaking changes
+
+# TODO: The above comments are a lie
```
theunrepentantgeek, line 6: gif

matthchr: Hah I should fix this before merge. I think we use certmanager like 1.4 now so just need to fix that bit.

hack/generated/main.go
```diff
@@ -40,15 +39,12 @@ func main() {
...
 
-	syncPeriod := 30 * time.Minute
-
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:             scheme,
 		MetricsBindAddress: metricsAddr,
 		LeaderElection:     enableLeaderElection,
 		LeaderElectionID:   "controllers-leader-election-azinfra-generated",
 		Port:               9443,
-		SyncPeriod:         &syncPeriod, // TODO: Don't leave this set, just playing with it
```
Porges, line 51: :D

Porges (approved):

PR #1669, opened 2021-07-28
Improve the external API of the PropertyBag 

theunrepentantgeek:

The original API of the property bag was going to be troublesome to generate code to consume, partly because of the way it combined checking for existence with deserialization. I've split those functions out into separate methods, which will allow code generation to avoid name collisions by using a nested scope:

``` go
if propertyBag.Contains("FullName") {
    var fullName string
    err := propertyBag.Pull("FullName", &fullName)
    if err != nil {
        return errors.NewError("failed to read property FullName from propertyBag")
    }

    destination.FullName = fullName
}
```

Split out as a separate PR to help keep the main PropertyBag usage PR smaller.

gif

**If applicable**:
- [x] this PR contains tests


hack/generated/pkg/genruntime/property_bag.go
```diff
@@ -7,25 +7,39 @@ package genruntime
...
+	return found
+}
+
 // Add is used to add a value into the bag; exact formatting depends on the type.
 // Any existing value will be overwritten.
+// property is the name of the item to put into the bag
+// value is the instance to be stashed away for later
 func (bag PropertyBag) Add(property string, value interface{}) error {
 	switch v := value.(type) {
 	case string:
```
matthchr, line 45: minor: This isn't in the code you actually changed, but I'm curious why `string` is called out here but not other primitive types?

Feels like you could probably get away without any special casing and just always use `json.Marshal`? I am assuming that `json.Marshal(str)` just returns `str` but I don't know for sure.

hack/generated/pkg/genruntime/property_bag_test.go
```diff
@@ -6,23 +6,113 @@
...
+
+func TestPropertyBag_NewPropertyBag_WithMultipleItemBag_ReturnsBagWithExpectedContent(t *testing.T) {
+	g := NewWithT(t)
+	original := NewPropertyBag()
+	g.Expect(original.Add("Answer", 42)).To(Succeed())
+	g.Expect(original.Add("Halloween", "31OCT")).To(Succeed())
+	g.Expect(original.Add("Christmas", "25DEC")).To(Succeed())
+
+	bag := NewPropertyBag(original)
+	g.Expect(bag).To(HaveKey("Answer"))
```
matthchr, line 55: You aren't asserting that the actual content is correct here, just that it has the key?
Shouldn't you also check to make sure that the key maps to the correct value?

Same comment applies below in a couple other tests, but won't repeat.

matthchr: Hmm I see in the other tests (that you already had) that you may have those cases covered already.

Maybe this just becomes minor: rather than `ReturnsBagWithExpectedContent` use `ReturnsBagWithExpectedKeys` to make it clear you only care about asserting on keys here?

hack/generated/pkg/genruntime/property_bag.go
theunrepentantgeek: I don't recall - might have been to avoid the overhead of calling `json.Marshal` when not required, or it may have been to handle round trip formatting issues. I do remember it was deliberate though.

matthchr: Might be worth leaving a comment there as to the reason? Especially if there was some problem if it wasn't handled specially?

matthchr (approved):
Approved as the active comments I have are pretty minor, and/or are on sections of the code that actually weren't changed as part of the PR.

Feel free to merge once you've taken a look at them.

hack/generated/pkg/genruntime/property_bag.go
theunrepentantgeek: I have a laundry list of minor code-gardening things to do post `codegen-alpha-0`, I'll add this to that.

hack/generated/pkg/genruntime/property_bag_test.go
theunrepentantgeek: Other tests check that the values stored can be recalled cleanly, so I'm only worried in this test whether the expected keys are put into the bag. I'll rename the tests as you suggest.

PR #1672, opened 2021-07-29
Code gardening for code generation

theunrepentantgeek:

Our `SimpleAssignment()` method has always been brittle because it's so easy to misuse by passing the wrong token value. This also makes it more complex to consume - I've wanted to simplify things for a while.

* `SimpleAssignment()` has been modified so it always does assignment 

* New method `SimpleDeclaration()` uses `:=` to always declare a variable. The first parameter is a string, simplifying all the calling sites.

* New method `SetVariable()` handles the very few remaining cases where the choice between declaration and assignment was dynamic.

The helper `EnsureStatementBlock()` has been subsumed into `StatementBlock()` to simplify usage.

Extends usage of helper functions to simplify code.

Moves some local variable name reservations later, until we know for sure the name is needed. This avoids a handful of cases where we'd reserve a local variable name and not use it, forcing other local variables to use clumsier names in the final output.

I wanted some of this for a WIP branch so I extracted it out into a separate PR.

gif

### DIFF
--- a/hack/generator/pkg/astbuilder/assignments.go
+++ b/hack/generator/pkg/astbuilder/assignments.go
@@ -6,20 +6,22 @@
 package astbuilder
 
 import (
+	"fmt"
 	"go/token"
 
 	"github.com/dave/dst"
 )
 
 // SimpleAssignment performs a simple assignment like:
-//     <lhs> := <rhs>       // tok = token.DEFINE
-// or  <lhs> = <rhs>        // tok = token.ASSIGN
-func SimpleAssignment(lhs dst.Expr, tok token.Token, rhs dst.Expr) *dst.AssignStmt {
+//
+//     <lhs> = <rhs>
+//
+func SimpleAssignment(lhs dst.Expr, rhs dst.Expr) *dst.AssignStmt {
 	return &dst.AssignStmt{
 		Lhs: []dst.Expr{
 			dst.Clone(lhs).(dst.Expr),
 		},
-		Tok: tok,
+		Tok: token.ASSIGN,
 		Rhs: []dst.Expr{
 			dst.Clone(rhs).(dst.Expr),
 		},
--- a/hack/generator/pkg/astbuilder/assignments.go
+++ b/hack/generator/pkg/astbuilder/assignments.go
@@ -44,6 +44,21 @@ func SimpleDeclaration(lhs dst.Expr, rhs dst.Expr) *dst.AssignStmt {
 	}
 }
 
+// SetVariable allows for either variable declaration or assignment by passing the required token
+// Only token.DEFINE and token.ASSIGN are supported, other values will panic.
+// Use SimpleAssignment or SimpleDeclaration by preference
+func SetVariable(lhs dst.Expr, tok token.Token, rhs dst.Expr) *dst.AssignStmt {
+	if tok == token.ASSIGN {
+		return SimpleAssignment(lhs, rhs)
+	}
+
+	if tok == token.DEFINE {
+		return SimpleDeclaration(lhs, rhs)
+	}
+
+	panic(fmt.Sprintf("token %q not supported in VariableAssignment", tok))
+}
+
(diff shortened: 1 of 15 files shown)


hack/generator/pkg/astbuilder/assignments.go
```diff
@@ -6,26 +6,59 @@
...
 			dst.Clone(rhs).(dst.Expr),
 		},
 	}
 }
 
+// SimpleDeclaration performs a simple assignment like:
+//
+//     <id> := <rhs>
+//
+func SimpleDeclaration(id string, rhs dst.Expr) *dst.AssignStmt {
```
matthchr, line 35: We might need to be careful about nomenclature here. [This](https://blog.golang.org/declaration-syntax) and [this](https://gobyexample.com/variables) both use declaration to refer to _just_ the `var x` bit.

In other places in the code we for example have `LocalVariableDeclaration` (which makes a `var <ident> <type>`) and `VariableDeclaration` which makes a `var <ident> <type>` as well (but with a different return type).

It seems like _also_ using `Declaration` to refer to `a := b` might cause confusion?

I think we use the `Definition` word less (and the token is `DEFINE`, not `DECLARE`) - does `SimpleDefinition` do better? Or maybe we need to make it `SimpleDeclareAndAssign` or something?

hack/generator/pkg/astbuilder/assignments.go
```diff
@@ -6,26 +6,59 @@
...
+		Tok: token.DEFINE,
+		Rhs: []dst.Expr{
+			dst.Clone(rhs).(dst.Expr),
+		},
+	}
+}
+
+// SetVariable allows for either variable declaration or assignment by passing the required token
+// Only token.DEFINE and token.ASSIGN are supported, other values will panic.
+// Use SimpleAssignment or SimpleDeclaration by preference
```
matthchr, line 49:
```suggestion
// Use SimpleAssignment or SimpleDeclaration if possible
```
or something?

For whatever reason the "by preference" ending here makes it sound like: "You can use whatever you prefer" when we're really saying "you _should_ prefer something other than this if possible"

hack/generator/pkg/astbuilder/assignments.go
```diff
@@ -6,26 +6,59 @@
...
+func SetVariable(lhs dst.Expr, tok token.Token, rhs dst.Expr) *dst.AssignStmt {
```
matthchr, line 50: Minor: Rather than `SetVariable` this could possibly just be `Assign` (or `Assignment`) in contrast to `SimpleAssignment`/`SimpleDeclaration`?

That keeps with the nomenclature of the `dst` and so hopefully makes this easier to find and more closely matches the naming of the above two methods (which I think fits given that this is "like those but with more flexibility and less safety")?

matthchr: When I hear variable I tend to think of local (or global) variables and not expressions like `a.b.c.d` too, whereas this supports both

matthchr (approved):
Looks good. Have some comments related to the naming of things though.

hack/generator/pkg/astbuilder/assignments.go
theunrepentantgeek: I'm not entirely wedded to the naming ... but ... I used the term declaration because the Go Language feature for `:=` is literally called _Short Variable Declarations_.

Though, I agree that having the token as `DEFINE` is inconsistent.

I'll rename the method to `ShortDeclaration()` to better align with the GoLang terminology.

hack/generator/pkg/astbuilder/assignments.go
theunrepentantgeek: Good idea. I've reworded the comment to be much stronger.

hack/generator/pkg/astbuilder/assignments.go
theunrepentantgeek: Renamed to `AssignmentStatement()`.

PR #1673, opened 2021-07-30
Increase CI timeout

matthchr:


theunrepentantgeek (approved):
Wow, for a file with only three minor changes, the GitHub diff sure does get confuzzled! 

Looks good. I'll merge this as requested via chat.

gif

PR #1674, opened 2021-07-30
Use Update (PUT) not Patch when modifying spec and status

matthchr:
Patch has a number of issues:
1. You aren't guaranteed that you're operating on the latest version of the resource, so you may be patching in a way that is actually nonsensical.
2. It's difficult to remove things

Rather than fight with Patch, moved to use Update (full replace) everywhere. As part of this work also refactored a little to make the controller more deterministic.

gif

**If applicable**:
- [x] this PR contains tests


matthchr:
Note: There are lots more refactors that we need to do to `azure_deployment_reconciler.go` but I'm putting them off given the work I have in that space with conditions and also the reality that the crufty-ness of the controller isn't going to be impacting customers whereas some of the resource shape stuff we haven't finished yet does.

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
```diff
@@ -254,12 +258,28 @@ func (r *AzureDeploymentReconciler) SpecSignature() (string, error) {
 	return hex.EncodeToString(hash[:]), nil
 }
 
+func (r *AzureDeploymentReconciler) UpdateBeforeCratingDeployment(
```
matthchr, line 261: Creating... not Crating

hack/generated/controllers/recordings/Test_CosmosDB_CRUD.yaml
```diff
@@ -172,14 +172,14 @@ interactions:
       - application/json
       Test-Request-Attempt:
       - "0"
-    url: https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/asotest-rg-dnhfns/providers/Microsoft.Resources/deployments/k8s_f9f7e648-5320-5c32-8775-febefed1fa3a?api-version=2019-10-01
+    url: https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/asotest-rg-dnhfns/providers/Microsoft.Resources/deployments/k8s_f9f7e648-5320-5c32-8775-febefed1fa3a?api-version=2019-10-01
```
theunrepentantgeek, line 175: What's the canonical form here? Was `resourcegroups` wrong or is `resourceGroups` now wrong?

theunrepentantgeek: To clarify - I think we should be using the canonical forms of the ARM URIs even if either case is supported.

hack/generated/pkg/armclient/template_client.go
```diff
@@ -311,14 +312,28 @@ func (atc *AzureTemplateClient) HeadResource(ctx context.Context, id string, api
...
+	if len(params)%2 != 0 {
+		panic("ARM Id params must come in resourceKind/name pairs")
 	}
 
-	start := "/subscriptions/" + subscriptionId
-	remaining := strings.Join(segments, "/")
+	suffix := strings.Join(params, "/")
 
-	return start + "/" + remaining, nil
+	return fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/%s/%s", subscription, resourceGroup, provider, suffix)
```
theunrepentantgeek, line 338: Can/should `MakeResourceGroupScopeARMId()` delegate to `MakeSubscriptionScopeARMId()`?

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
```diff
@@ -110,6 +110,8 @@ func NewAzureDeploymentReconciler(
 }
 
 func (r *AzureDeploymentReconciler) CreateOrUpdate(ctx context.Context) (ctrl.Result, error) {
+	r.logObj("reconciling resource")
```
theunrepentantgeek, line 113: Do we have the information here to log _which_ resource we're reconciling?

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
```diff
@@ -128,6 +130,8 @@ func (r *AzureDeploymentReconciler) CreateOrUpdate(ctx context.Context) (ctrl.Re
 }
 
 func (r *AzureDeploymentReconciler) Delete(ctx context.Context) (ctrl.Result, error) {
+	r.logObj("deleting resource")
```
theunrepentantgeek, line 133: Ditto here - useful to log _which_ resource is being deleted.

hack/generated/controllers/crd_disk_test.go
```diff
@@ -48,10 +48,10 @@ func Test_Disk_CRUD(t *testing.T) {
 	armId := *disk.Status.Id
 
 	// Perform a simple patch.
-	patcher := tc.NewResourcePatcher(disk)
+	old := disk.DeepCopy()
 	networkAccessPolicy := compute.DiskPropertiesNetworkAccessPolicyDenyAll
 	disk.Spec.NetworkAccessPolicy = &networkAccessPolicy
-	patcher.Patch(disk)
+	tc.Patch(old, disk)
```
theunrepentantgeek, line 54: Should this comment/code still be using the term _Patch_?

theunrepentantgeek (commented):
A couple of questions, but mostly looks good.

gif

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
matthchr: It's automatically included as a variable pair on `r.log` already (`namespace` and `name`, as well as the specific controller type that is doing the logging). So everything that `r.log` logs has it automatically - which is nice because it keeps the code where we log short as it's just the "extra stuff" that has to be passed there.

Agree it does look a bit weird at first glance though.

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
matthchr: Same response as above -- already included

hack/generated/controllers/crd_disk_test.go
matthchr: This really is a Patch, so yes I think so.

If we wanted to use update here we'd need to also add an additional GET up above (so that we were operating on the latest version of the object). I figured keeping the Patch helper for tests was fine as it's a bit more succinct.

hack/generated/controllers/recordings/Test_CosmosDB_CRUD.yaml
matthchr: I'm not actually 100% sure there is an _official_ cannonicalized form which is used consistently across Azure.

With that said, the [ARM docs](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-resource#resourceid) use `resourceGroups`, and other providers such as `Microsoft.Storage` seem to use `storageAccounts` (not `storageaccounts`), so I think it makes sense to use that form as it's:
1. More readable.
2. The one used at least in the official docs I've seen.

hack/generated/pkg/armclient/template_client.go
matthchr: It can't without changing the way they're both writter, as `MakeSubscriptionScopeARMId` explicitly forbids tacking things after `/subscriptions/<>/` but before the `providers` section. If I changed `MakeSubscriptionScopeARMId` to allow that it wouldn't be as safe because it'd allow other (incorrect) formats too presumably.

We could probably refactor an internal method that both delegated to, but doing so would probably end up basically just being a slightly less generic `strings.Join`, so it doesn't seem worth it?

hack/generated/controllers/recordings/Test_CosmosDB_CRUD.yaml
theunrepentantgeek: Works for me. 👍

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
theunrepentantgeek: Ahh, good technique. Nice.

hack/generated/pkg/armclient/template_client.go
theunrepentantgeek: Fair enough.

hack/generated/controllers/crd_disk_test.go
theunrepentantgeek: Ok.

theunrepentantgeek (approved):
Looks good.

gif

PR #1675, opened 2021-07-30
Add new conditions type

matthchr:
Adds the `Conditions` types discussed in #1614

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/hack/generated/pkg/genruntime/conditions/conditions.go
+++ b/hack/generated/pkg/genruntime/conditions/conditions.go
@@ -1,0 +1,175 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package conditions
+
+import (
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type Conditions []Condition
+
+// FindIndexByType returns the index of the condition with the given ConditionType if it exists.
+// If the Condition with the specified ConditionType is doesn't exist, the second boolean parameter is false.
+func (c Conditions) FindIndexByType(conditionType ConditionType) (int, bool) {
+	for i := range c {
+		condition := c[i]
+		if condition.Type == conditionType {
+			return i, true
+		}
+	}
+
+	return -1, false
+}
+
+// TODO: Hah, name...
+type Conditioner interface {
+	GetConditions() Conditions
+	SetConditions(conditions Conditions)
+}
+
+// ConditionSeverity expresses the severity of a Condition.
+type ConditionSeverity string
+
+const (
+	// ConditionSeverityError specifies that a failure of a condition type
+	// should be viewed as an error. Errors are fatal to reconciliation and
+	// mean that the user must take some action to resolve
+	// the problem before reconciliation will be attempted again.
+	ConditionSeverityError ConditionSeverity = "Error"
+
+	// ConditionSeverityWarning specifies that a failure of a condition type
+	// should be viewed as a warning. Warnings are informational. The operator
+	// may be able to retry and resolve the warning without any action from the user, but
+	// in some cases user action to resolve the warning will be required.
+	ConditionSeverityWarning ConditionSeverity = "Warning"
+
+	// ConditionSeverityInfo specifies that a failure of a condition type
+	// should be viewed as purely informational. Things are working.
+	// This is the happy path.
+	ConditionSeverityInfo ConditionSeverity = "Info"
+
+	// ConditionSeverityNone specifies that there is no condition severity.
+	// For conditions which have positive polarity (Status == True is their normal/healthy state), this will set when Status == True
+	// For conditions which have negative polarity (Status == False is their normal/healthy state), this will be set when Status == False.
+	// Conditions in Status == Unknown always have a severity of None as well.
+	// This is the default state for conditions.
+	ConditionSeverityNone ConditionSeverity = ""
+)
+
+type ConditionType string
+
+const (
+	// ConditionTypeReady is a condition indicating if the resource is ready or not.
+	// A ready resource is one that has been successfully provisioned to Azure according to the
+	// resource spec. It has reached the goal state. This usually means that the resource is ready
+	// to use, but the exact meaning of Ready may vary slightly from resource to resource. Resources with
+	// caveats to Ready's meaning will call that out in the resource specific documentation.
+	ConditionTypeReady = "Ready"
+)
+
+var _ fmt.Stringer = Condition{}
+
+// Condition defines an extension to status (an observation) of a resource
+type Condition struct {
+	// Type of condition.
+	// +kubebuilder:validation:Required
+	Type ConditionType `json:"type"`
+
+	// Status of the condition, one of True, False, or Unknown.
+	// +kubebuilder:validation:Required
+	Status metav1.ConditionStatus `json:"status"`
+
+	// Severity with which to treat failures of this type of condition.
+	// For conditions which have positive polarity (Status == True is their normal/healthy state), this will be omitted when Status == True
+	// For conditions which have negative polarity (Status == False is their normal/healthy state), this will be omitted when Status == False.
+	// This is omitted in all cases when Status == Unknown
+	// +kubebuilder:validation:Optional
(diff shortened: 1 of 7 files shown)


hack/generated/pkg/genruntime/conditions/conditions.go
```diff
@@ -0,0 +1,166 @@
...
+package conditions
+
+import (
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// TODO: Hah, name...
+type Conditioner interface {
```
theunrepentantgeek, line 15: Yeah, not a fan of this name either. 

But the best I've got to offer is `ConditionProvider` and `ConditionCointainer`, neither of which is better.

hack/generated/pkg/genruntime/conditions/conditions.go
```diff
@@ -0,0 +1,166 @@
...
+
+// Copy returns an independent copy of the Condition
+func (c Condition) Copy() Condition {
+	return c
+}
+
+// String returns a string representation of this condition
+func (c Condition) String() string {
+	return fmt.Sprintf(
+		"Condition [%s], Status = %q, Severity = %q, Reason = %q, Message = %q, LastTransitionTime = %q",
```
theunrepentantgeek, line 109: Is `%q` correct for a `metav1.Time` type?

hack/generated/pkg/genruntime/conditions/conditions.go
```diff
@@ -0,0 +1,166 @@
...
+
+// SetCondition sets the provided Condition on the Conditioner. The condition is only
+// set if the new condition is in a different state than the existing condition of
+// the same type.
+func SetCondition(o Conditioner, new Condition) {
+	if o == nil {
+		return
+	}
+
+	conditions := o.GetConditions()
```
theunrepentantgeek, line 126: Since it's our interface, why not have `GetConditions()` return a `map[ConditionType]Condition` that allows for easy updates/checks?

hack/generated/pkg/genruntime/conditions/conditions.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	// TODO: do we sort conditions here? CAPI does.
+
+	o.SetConditions(conditions)
+}
+
+// GetCondition gets the Condition with the specified type from the provided Conditioner.
+// If no Condition matches that type, the boolean second parameter is false. If a condition
+// was found, the boolean is true.
```
theunrepentantgeek, line 152: I find these easier to read with the positive case coming first.

```suggestion
// Returns the condition and true if a condition with the specified type is found, or 
// an empty condition and false if not.
```

theunrepentantgeek (commented):
A few questions to consider; I'll probably approve once those are answered.

gif

hack/generated/pkg/genruntime/conditions/conditions.go
matthchr: Cluster API does:
```
type Setter interface {
	Getter
	SetConditions(clusterv1.Conditions)
}
```

So it's used as `conditions.Setter`, which is maybe not bad? I didn't split the interface into `Get` and `Set`, but I could.

hack/generated/pkg/genruntime/conditions/conditions.go
matthchr: If we return a `map` is that we can't fuss with ordering if we want - which I think we might end up wanting/needing to do in order to make updates work well (CAPI does it). Obviously we could do ordering inside of the interface implementation but that means we end up duplicating the ordering code in a lot of implementations.

Given that (for better or worse) the actual property on status is a slice, I think it makes sense for the interface to expose that lowest level primitive. If we want to write a wrapper that uses that interface to get the slice and then convert it to a map, work with that and then convert it back, we can already do that above the level of the interface (which saves us having to write code generation to do the to/from map logic in each impl).

One more (quite minor) consideration is that I was modeling this after the CAPI implementation (which we also referenced for the `Condition` shape) and their interface looks (mostly) like this

hack/generated/pkg/genruntime/conditions/conditions.go
```diff
@@ -0,0 +1,166 @@
...
+const (
+	// ConditionSeverityError specifies that a failure of a condition type
+	// should be viewed as an error. Errors are fatal to reconciliation and
+	// mean that the user must take some action to resolve
+	// the problem before reconciliation will be attempted again.
+	ConditionSeverityError ConditionSeverity = "Error"
+
+	// ConditionSeverityWarning specifies that a failure of a condition type
+	// should be viewed as a warning. Warnings are informational. The operator
+	// may be able to retry through the warning without any action from the user, but
```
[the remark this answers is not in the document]
matthchr: Fixed

hack/generated/pkg/genruntime/conditions/conditions.go
matthchr: Yes, `metav1.Time` embeds `time.Time`, and `time.Time` has a [String()](https://pkg.go.dev/time#Time.String) function which renders text that looks like: `2021-08-02 10:18:10.230766841 -0700 PDT`. So `%q` is just wrapping that in quotes.

hack/generated/pkg/genruntime/conditions/conditions.go
matthchr: Fixed

hack/generated/pkg/genruntime/conditions/conditions.go
```diff
@@ -0,0 +1,166 @@
...
+	// This is the default state for conditions.
+	ConditionSeverityNone ConditionSeverity = ""
+)
+
+type ConditionType string
+
+const (
+	// ConditionTypeReady is a condition indicating if the resource is ready or not.
+	// A ready resource is one that has been successfully provisioned to Azure according to the
+	// resource Spec.
```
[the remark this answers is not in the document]
matthchr: I think our users would appreciate that, but I'm also a bit hesitant assuming that just because a resource is successfully provisioned in Azure that it is "ready to use".

What this means (at a technical level, for ARM resources) is that the resource we have sent to Azure was successfully provisioned. The long running provisioning operation completed and the resource is in the `Provisioned` state. I am not sure that we can authoritatively say more than that without having a deep understanding of the specific resource in question?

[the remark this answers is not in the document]
matthchr: I realized that this status is actually broader than ARM, because it'll be on all resources regardless of if they are ARM or not. I tried to craft something generic that was helpful while also leaving a bit of room for uncertainty... I came up with this:
> 	// ConditionTypeReady is a condition indicating if the resource is ready or not.
	// A ready resource is one that has been successfully provisioned to Azure according to the
	// resource spec. It has reached the goal state. This usually means that the resource is ready
	// to use, but the exact meaning of Ready may vary slightly from resource to resource. Resources with
	// caveats to Ready's meaning will call that out in the resource specific documentation.

Let me know if that sounds good or if you have suggestions to change it.

hack/generated/pkg/genruntime/conditions/conditions.go
matthchr: maybe let's sync on this in person?

I'm not understanding how what you're suggesting solves for example the ordering problem I mentioned, nor why we necessarily need it in this interface rather than in a wrapper if we want. The actual data is a slice and has to be a slice (because that's what the Kubernetes conditions spec requires) so any time we act on a map we must convert it back to a slice at the end.

hack/generated/pkg/genruntime/conditions/conditions.go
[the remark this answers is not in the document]
theunrepentantgeek: That works for me.

hack/generated/pkg/genruntime/conditions/conditions.go
matthchr: Synced up in person and I'll go ahead and make the change as discussed

theunrepentantgeek: Looks good.

theunrepentantgeek (approved):
Looks good. 

gif

PR #1684, opened 2021-08-03
Don't CommitUpdate twice in a single reconcile loop

matthchr:
Multiple updates in a single loop means that there are multiple changes with new resourceVersions floating around. This then leads to races where you might end up reconciling on the resource from after update1 when you really want the change from after update2.

Since the two-updates was just an optimization, removed it.

### DIFF
--- a/hack/generated/controllers/generic_controller.go
+++ b/hack/generated/controllers/generic_controller.go
@@ -39,7 +39,7 @@ type GenericReconciler struct {
 	Name                 string
 	GVK                  schema.GroupVersionKind
 	Controller           controller.Controller
-	RequeueDelay         time.Duration
+	RequeueDelayOverride time.Duration
 	CreateDeploymentName func(obj metav1.Object) (string, error)
 }
 
@@ -54,11 +54,6 @@ type Options struct {
 }
 
 func (options *Options) setDefaults() {
-	// default requeue delay to 5 seconds
-	if options.RequeueDelay == 0 {
-		options.RequeueDelay = 5 * time.Second
-	}
-
 	// override deployment name generator, if provided
 	if options.CreateDeploymentName == nil {
 		options.CreateDeploymentName = createDeploymentName
@@ -135,7 +130,7 @@ func register(mgr ctrl.Manager, reconciledResourceLookup map[schema.GroupKind]sc
 		Log:                  log.WithName(controllerName),
 		Recorder:             mgr.GetEventRecorderFor(controllerName),
 		GVK:                  gvk,
-		RequeueDelay:         options.RequeueDelay,
+		RequeueDelayOverride: options.RequeueDelay,
 		CreateDeploymentName: options.CreateDeploymentName,
 	}
 
@@ -225,10 +220,13 @@ func (gr *GenericReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		return ctrl.Result{}, err
 	}
 
-	// TODO: This actually means we don't get exponential backoff which Kubernetes does by default,
-	// TODO: but we need this for test today to get fast reconciles.
-	if result.Requeue && result.RequeueAfter == time.Duration(0) {
-		result.RequeueAfter = gr.RequeueDelay
+	// If we have a requeue delay override, set it for all situations where
+	// we are requeueing.
+	hasRequeueDelayOverride := gr.RequeueDelayOverride != time.Duration(0)
+	isRequeueing := result.Requeue || result.RequeueAfter > time.Duration(0)
+	if hasRequeueDelayOverride && isRequeueing {
+		result.RequeueAfter = gr.RequeueDelayOverride
+		result.Requeue = true
 	}
 
 	return result, nil
--- a/hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
+++ b/hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
@@ -12,7 +12,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
-	"strconv"
 	"strings"
 	"time"
 
@@ -47,8 +46,6 @@ const (
 	ResourceStateAnnotation  = "resource-state.infra.azure.com"
 	ResourceErrorAnnotation  = "resource-error.infra.azure.com"
 	ResourceSigAnnotationKey = "resource-sig.infra.azure.com"
-	// PreserveDeploymentAnnotation is the key which tells the applier to keep or delete the deployment
-	PreserveDeploymentAnnotation = "x-preserve-deployment"
 )
 
 // TODO: Do we actually want this at the controller level or this level?
@@ -184,23 +181,6 @@ func (r *AzureDeploymentReconciler) GetDeploymentNameOrDefault() string {
 
 func (r *AzureDeploymentReconciler) SetDeploymentName(name string) {
 	genruntime.AddAnnotation(r.obj, DeploymentNameAnnotation, name)
-}
-
-func (r *AzureDeploymentReconciler) GetShouldPreserveDeployment() bool {
-	preserveDeploymentString, ok := r.obj.GetAnnotations()[PreserveDeploymentAnnotation]
-	if !ok {
-		return false
-	}
-
-	preserveDeployment, err := strconv.ParseBool(preserveDeploymentString)
-	// Anything other than an error is assumed to be false...
-	// TODO: Would we rather have any usage of this key imply true (regardless of value?)
-	if err != nil {
-		// TODO: Log here
-		return false
-	}
-
-	return preserveDeployment
 }
 
 func (r *AzureDeploymentReconciler) SetResourceProvisioningState(state armclient.ProvisioningState) {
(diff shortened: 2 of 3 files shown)


hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
```diff
@@ -593,90 +606,39 @@ func (r *AzureDeploymentReconciler) MonitorDeployment(ctx context.Context) (ctrl
...
+	r.log.Info(
+		"Deployment in terminal state",
+		"DeploymentID", deployment.ID,
+		"State", deployment.ProvisioningStateOrUnknown(),
+		"Error", deployment.ErrorOrEmpty())
 
+	// It is possible that we delete the deployment here and then are unable to persist the details of the created
+	// resource to etcd below. If this happens, a subsequent reconciliation will attempt to GET the deployment which will
+	// fail. That will trigger us to throw the deployment ID away and create a new one (which will end up being a no-op
+	// because the Azure resource already exists). Since it's expected that this scenario of events is rare, we don't
```
matthchr, line 628:
```suggestion
	// because the Azure resource already exists). Since it's expected that this sequence of events is rare, we don't
```

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
```diff
@@ -593,90 +606,39 @@ func (r *AzureDeploymentReconciler) MonitorDeployment(ctx context.Context) (ctrl
...
-	if deployment.IsTerminalProvisioningState() {
-		// we are done
-		return ctrl.Result{}, nil
-	}
+	deployment.ID = ""
+	deployment.Name = ""
 
-	r.log.V(3).Info("Deployment still running")
-	return ctrl.Result{Requeue: true, RequeueAfter: retryAfter}, err
+	// TODO: Do people prefer this or just putting this method body inline
```
matthchr, line 640: Thoughts on this?

hack/generated/controllers/generic_controller.go
```diff
@@ -225,10 +220,10 @@ func (gr *GenericReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		return ctrl.Result{}, err
 	}
 
-	// TODO: This actually means we don't get exponential backoff which Kubernetes does by default,
-	// TODO: but we need this for test today to get fast reconciles.
-	if result.Requeue && result.RequeueAfter == time.Duration(0) {
+	// If we have a hardcoded requeue delay, force that
+	if gr.RequeueDelay != time.Duration(0) && (result.Requeue || result.RequeueAfter > time.Duration(0)) {
 		result.RequeueAfter = gr.RequeueDelay
+		result.Requeue = true
```
Porges, line 226: `Requeue` is ignored if `RequeueAfter` is set… we should probably rethink this a bit, since we probably want the `controller-runtime` backoff…

matthchr: I think that this is keeping the `controller-runtime` backoff?

If the configured `gr.RequeueDelay` is `0` (the default in everything but the recording tests) then we don't do anything here and just return `result` as-is - which is what we want to do because that's what the underlying controller set.
If the configured `RequeueDelay` is nonzero (the tests) we just force the delay to be the specified delay, which yes deprives us of the backoff but that's what we want in the tests.

I think that's _better_ than what we had before because before the actual live behavior was that we were always setting `result.RequeueAfter = gr.RequeueAfter` whenever `result.RequeueAfter == 0` which was all the time because we don't set it in the controller?

matthchr: Possibly I need to extract the if parameters out into clearly named variables to make this more obvious though?

Porges: I was misreading it a bit. But yes, renaming to something like `RequeueDelayOverride` would help!

Porges (approved):

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
```diff
@@ -560,7 +540,40 @@ func (r *AzureDeploymentReconciler) CreateDeployment(ctx context.Context) (ctrl.
...
+
+	err := r.Update(deployment, status)
+	if err != nil {
+		return ctrl.Result{}, errors.Wrap(err, "updating obj")
+	}
+
+	err = r.CommitUpdate(ctx)
+	if err != nil {
+		// This is a superfluous error as per https://github.com/kubernetes-sigs/controller-runtime/issues/377
+		// The correct handling is just to ignore it and we will get an event shortly with the updated version to patch
```
theunrepentantgeek, line 570: Should we be treating _all_ errors as superfluous, or should we be checking in case an unexpected kind of error comes back?

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
```diff
@@ -593,90 +606,39 @@ func (r *AzureDeploymentReconciler) MonitorDeployment(ctx context.Context) (ctrl
...
-		if idErr != nil {
-			return ctrl.Result{}, errors.Wrap(idErr, "getting resource ID from resource")
-		}
-
-		s, _, statusErr := r.getStatus(ctx, resourceID)
-		if statusErr != nil {
-			return ctrl.Result{}, errors.Wrap(statusErr, "getting status from ARM")
-		}
+	r.log.V(4).Info("Monitoring deployment", "action", string(CreateOrUpdateActionMonitorDeployment), "id", deploymentID, "state", deployment.ProvisioningStateOrUnknown())
+	r.recorder.Event(r.obj, v1.EventTypeNormal, string(CreateOrUpdateActionMonitorDeployment), fmt.Sprintf("Monitoring Azure deployment ID=%q, state=%q", deploymentID, deployment.ProvisioningStateOrUnknown()))
```
theunrepentantgeek, line 610: Suggest formatting the parameters across separate lines
```suggestion
	r.recorder.Event(
		r.obj, 
		v1.EventTypeNormal, 
		string(CreateOrUpdateActionMonitorDeployment), 
		fmt.Sprintf("Monitoring Azure deployment ID=%q, state=%q", deploymentID, deployment.ProvisioningStateOrUnknown()))
```

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
matthchr: This comment applies specifically to `NotFound` which is dealt with via `client.IgnoreNotFound(err)` in the return below. The comment is a line above that because it's long,

I could change "this" to "NotFound" maybe?

hack/generated/controllers/generic_controller.go
matthchr: Changed this to `gr.RequeueDelayOverride` and also refactored the if statement to make it clearer.

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
matthchr: Changed it to:
> NotFound is a superfluous error as per https://github.com/kubernetes-sigs/controller-runtime/issues/377

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
matthchr: Did this - also with the above line which was basically just as long

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
matthchr: There were some other occurrences of this comment and I changed them too

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
matthchr: I created a method that is only used in a single place and was second guessing whether it was just cleaner to do it inline in this method. That makes this method longer though...

I've just removed the TODO and kept the `handleDeploymentFinished` for now. We can always change it later if we don't like it.

theunrepentantgeek: A single use method is often still a good idea because it gives that piece of functionality a name, and allows consideration in isolation.

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
theunrepentantgeek: Err, ahh, my comment was more that currently _every_ error returned from `r.CommitUpdate()` is treated as superfluous regardless of the details, and I was asking if that was correct.

So I was asking about the couple of lines ABOVE the comment, not the lines below it.

theunrepentantgeek: Having pulled the code and looked at in detail, it does the right thing - but it's an odd idiom and easy to misread.

theunrepentantgeek (approved):
Concerns resolved.

gif

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
matthchr: Yeah that was my thought when I wrote it, but sometimes you can also overuse it and end up in really deep method chains. For some reason I was hesitant about it when I wrote it but in reviewing the PR it seems just net-good.

PR #1687, opened 2021-08-04
Use Github Action to add needs-triage label

Porges:
Bot doesn't seem to be working so let's do this instead. Seems simpler overall as well!

### DIFF
--- a/.github/workflows/add-needs-triage-label.yml
+++ b/.github/workflows/add-needs-triage-label.yml
@@ -13,6 +13,6 @@ jobs:
       issues: write
     steps:
       - name: Label issues
-        uses: andymckay/labeler@1825c391d68ade591efa053af2472e5b0e2e2d9e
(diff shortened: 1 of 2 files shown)


.github/workflows/add-needs-triage-label.yml
```diff
@@ -0,0 +1,18 @@
...
+      - opened
+      
+jobs:
+  label_issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - name: Label issues
+        uses: andymckay/labeler@1825c391d68ade591efa053af2472e5b0e2e2d9e
```
matthchr, line 16: I think `@1.0.3` would also work and would be clearer?

matthchr (approved):

.github/workflows/add-needs-triage-label.yml
Porges: Added comment

PR #1688, opened 2021-08-04
Generate DeepCopy() methods for genruntime types

matthchr:
Some genruntime types are used by the generated code, while others are used only by the handcrafted controller. Types used
by the generated code need DeepCopy() functionality.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -195,7 +195,7 @@ tasks:
   controller:test-integration-envtest:
     desc: Run integration tests with envtest using record/replay.
     dir: "{{.CONTROLLER_ROOT}}"
-    deps: [controller:generate-crds]
+    deps: [controller:generate-kustomize]
     cmds:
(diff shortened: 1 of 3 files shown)


Taskfile.yml
```diff
@@ -45,15 +45,15 @@ tasks:
     desc: Run all CI checks.
     cmds: 
     - task: generator:ci
-    - task: controller:generate-crds # required here to prevent it running twice in controller:ci, remove once go-task dedupes dependencies
+    - task: controller:generate-kustomize # required here to prevent it running twice in controller:ci, remove once go-task dedupes dependencies
```
theunrepentantgeek, line 48: Doesn't **task** dedup now? See PR #1690

theunrepentantgeek (approved):

Taskfile.yml
matthchr: Yeah, had conflicts with that. Now fixed.

PR #1689, opened 2021-08-04
Add new conditions pipeline stages

matthchr:
Add pipeline stage to put conditions on all status types.

**If applicable**:
- [x] this PR contains tests


matthchr:
Ah, looks like this has to be rebase onto #1688

I will leave this as is for now and rebase it once 1688 makes it to master. Comments still welcome on this even though CI is failing (it should pass after rebase)

hack/generator/pkg/astmodel/conditioner_interface.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/dave/dst"
+
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/astbuilder"
+)
+
+const (
+	ConditionsProperty = "Conditions"
+)
+
+func NewConditionerInterfaceImpl(
```
matthchr, line 20: Methods missing comments in here...

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
```diff
@@ -218,6 +219,23 @@ func (builder *convertFromARMBuilder) ownerPropertyHandler(
 	return []dst.Stmt{result}
 }
 
+// conditionsPropertyHandler generates conversions the "Conditions" status property. This property is set by the controller
+// after each reconcile and so currently is not preserved.
+func (builder *convertFromARMBuilder) conditionsPropertyHandler(
+	toProp *astmodel.PropertyDefinition,
+	_ *astmodel.ObjectType) []dst.Stmt {
+
+	if toProp.PropertyName() != builder.idFactory.CreatePropertyName(astmodel.ConditionsProperty, astmodel.Exported) || builder.typeKind != TypeKindStatus {
```
matthchr, line 228: maybe clarify this if statement a bit

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
```diff
@@ -218,6 +219,23 @@ func (builder *convertFromARMBuilder) ownerPropertyHandler(
 	return []dst.Stmt{result}
 }
 
+// conditionsPropertyHandler generates conversions the "Conditions" status property. This property is set by the controller
```
theunrepentantgeek, line 222:
```suggestion
// conditionsPropertyHandler generates conversions for the "Conditions" status property. This property is set by the controller
```

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
```diff
@@ -218,6 +219,23 @@ func (builder *convertFromARMBuilder) ownerPropertyHandler(
...
+// after each reconcile and so currently is not preserved.
```
theunrepentantgeek, line 223: "currently" implies that maybe it should in future? Is this right?
```suggestion
// after each reconcile and does not need to be preserved.
```

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
```diff
@@ -218,6 +219,23 @@ func (builder *convertFromARMBuilder) ownerPropertyHandler(
...
+func (builder *convertFromARMBuilder) conditionsPropertyHandler(
+	toProp *astmodel.PropertyDefinition,
+	_ *astmodel.ObjectType) []dst.Stmt {
+
+	if toProp.PropertyName() != builder.idFactory.CreatePropertyName(astmodel.ConditionsProperty, astmodel.Exported) || builder.typeKind != TypeKindStatus {
+		return nil
+	}
+
+	// Returning an empty statement allows us to "consume" this match and not proceed to the next handler.
```
theunrepentantgeek, line 232: Returning an empty statement feels like a bit of a hack, esp given it's not idiomatic Go to return sentinel values. Consider a later refactoring where each handler returns `([]dst.Stmt, bool)`?

hack/generator/pkg/codegen/pipeline/add_status_conditions.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/astmodel"
+)
+
+const AddStatusConditionsStageID = "addStatusConditions"
+
+func AddStatusConditions(idFactory astmodel.IdentifierFactory) Stage {
+	return MakeLegacyStage(
+		AddStatusConditionsStageID,
+		"Adds conditions to all status types and implements genruntime.Conditioner on all resources",
```
theunrepentantgeek, line 21:
```suggestion
		"Adds the property 'Conditions' to all status types and implements genruntime.Conditioner on all resources",
```

hack/generator/pkg/codegen/pipeline/add_status_conditions.go
```diff
@@ -0,0 +1,82 @@
...
+				}
+				resourceType = resourceType.WithInterface(conditionerImpl)
+
+				result.Add(def.WithType(resourceType))
+			}
+			remaining := defs.Except(resourceDefs).Except(statusDefs)
+			// Add everything that's left
+			for _, def := range remaining {
+				result.Add(def)
+			}
```
theunrepentantgeek, line 51: I think you can do this more simply using the `Overlay()` method on `Types`.

hack/generator/pkg/codegen/pipeline/add_status_conditions.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// addConditions adds a "Conditions" property to the status type
+func addConditions(statusType astmodel.Type) (astmodel.Type, error) {
+	visitor := astmodel.TypeVisitorBuilder{
+		VisitObjectType: addConditionsToObject,
+	}.Build()
+
+	return visitor.Visit(statusType, nil)
+}
```
theunrepentantgeek, line 64: Use `PropertyInjector` to do the heavy lifting

hack/generator/pkg/codegen/pipeline/add_status_conditions_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+	initialState := NewState().WithTypes(types)
+
+	finalState, err := RunTestPipeline(
+		initialState,
+		AddStatusConditions(idFactory))
+	g.Expect(err).To(Succeed())
+
+	// When verifying the golden file, check to ensure that the Conditions property on the Status type looks correct,
+	// and that the conditions.Conditioner interface is properly implemented.
```
theunrepentantgeek, line 38:
```suggestion
	// and that the conditions.Conditioner interface is properly implemented on the resource.
```

theunrepentantgeek (commented):
You've said it needs work (including rebase) so I won't approve, but looks good.

gif

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
matthchr: Good eye

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
matthchr: There's a theoretical future where we have support for more than just the `Ready` condition, and other conditions aren't set in each reconcile loop pass possibly?

It seems an unlikely future though, especially because if we don't rederive the conditions each pass there are cases (such as when `Status` gets lost) where we might be making assumptions about what we have and not updating them - have made the change you suggested as it more clearly explains the _now_, and we can leave the future for the future.

hack/generator/pkg/astmodel/armconversion/convert_from_arm_function_builder.go
matthchr: I've made a note to do this but will do it in a later PR. Given that the infrastructure that uses this function signature is shared across to/from ARM (and there are a number of methods in each) it'll be a pretty big refactor.

Agree it'd make it a bit clearer though.

hack/generator/pkg/codegen/pipeline/add_status_conditions.go
matthchr: Fixed

hack/generator/pkg/codegen/pipeline/add_status_conditions.go
matthchr: Yep, you're right. Moved to use that.

hack/generator/pkg/codegen/pipeline/add_status_conditions_test.go
matthchr: Fixed

hack/generator/pkg/codegen/pipeline/add_status_conditions.go
matthchr: That is nice!

The only problem is that I want to make sure to error if we're trying to add a property that already exists. It seems like maybe `PropertyInjector` should have that behavior by default too? Otherwise we'll be silently overwriting properties.

I'll delete my code here and update `PropertyInjector` to do that.

matthchr:
I've rebased this now so it should be good to merge once approved

hack/generator/pkg/astmodel/property_injector.go
```diff
@@ -25,19 +27,34 @@ func NewPropertyInjector() *PropertyInjector {
 
 // Inject modifies the passed type definition by injecting the passed property
 func (pi *PropertyInjector) Inject(def TypeDefinition, prop *PropertyDefinition) (TypeDefinition, error) {
-	return pi.visitor.VisitDefinition(def, prop)
+	result, err := pi.visitor.VisitDefinition(def, prop)
+	if err != nil {
+		return TypeDefinition{}, errors.Wrapf(err, "failed to inject property into %q", def.Name())
```
theunrepentantgeek, line 32: Maybe include the name of the property too (in case the error has another cause)?
```suggestion
		return TypeDefinition{}, errors.Wrapf(err, "failed to inject property %q into %q", prop.Name(), def.Name())
```

hack/generator/pkg/codegen/pipeline/add_status_conditions.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/pkg/errors"
+
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/astmodel"
+)
+
+const AddStatusConditionsStageID = "addStatusConditions"
+
+func AddStatusConditions(idFactory astmodel.IdentifierFactory) Stage {
+	return MakeLegacyStage(
```
theunrepentantgeek, line 19: Ideally new stages wouldn't use `MakeLegacyStage()`, they'd use `MakeStage()` instead.

hack/generator/pkg/codegen/pipeline/add_status_conditions.go
```diff
@@ -0,0 +1,56 @@
...
+				result.Add(updatedDef)
+			}
+
+			resourceDefs := astmodel.FindResourceTypes(defs)
+			for _, def := range resourceDefs {
+				resourceType := def.Type().(*astmodel.ResourceType)
+
+				conditionerImpl, err := astmodel.NewConditionerInterfaceImpl(idFactory, resourceType)
+				if err != nil {
+					return nil, errors.Wrapf(err, "couldn't create conditioner impl for %q", def.Name())
```
theunrepentantgeek, line 46:
```suggestion
					return nil, errors.Wrapf(err, "couldn't create genruntime.Conditioner implementation for %q", def.Name())
```

theunrepentantgeek (approved):
Looks good.

gif

hack/generator/pkg/astmodel/property_injector.go
matthchr: Fixed

hack/generator/pkg/codegen/pipeline/add_status_conditions.go
matthchr: Fixed

hack/generator/pkg/codegen/pipeline/add_status_conditions.go
matthchr: Fixed

PR #1692, opened 2021-08-05
Release pipeline for v2 builds

### LINKED ISSUE #1632: Create release pipeline for ASO v2
We have a release pipeline for ASO v1, done via Azure DevOps

**Describe the improvement**
We also need a release pipeline for ASO V2. @porges has done some work on this but I don't think it's done yet.

Porges:
Closes #1632 

gif


.github/workflows/create-release.yml
```diff
@@ -0,0 +1,40 @@
...
+  build-and-push:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # required to access tags
+          submodules: 'true'
+
+      - name: Build & run devcontainer image
```
matthchr, line 20: Above with the "build and push Devcontainer image" task in `build-devcontainer-image.yml` we use the `docker/build-push-action@v2`. It seems that has a `push` option (which could be set to false?)

On the other hand I see in `pr-validation` we do basically this what you're doing here.

A couple questions about this:
1. Should we be using that `build-and-push` task more universally?
2. If the answer to the above is no, should we maybe have a comment here and in `pr-validation` that basically says "these two things should be kept in sync"?

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,40 @@
...
+        # this always builds a new image from scratch rather than from the build-devcontainer-image workflow output
+        # so that we pick up the latest versions of everything
+        id: devcontainer
+        run: |
+          docker build --tag devcontainer:latest .devcontainer
+          container_id=$(docker create -w /workspace -v $GITHUB_WORKSPACE:/workspace -v /var/run/docker.sock:/var/run/docker.sock devcontainer:latest)
+          docker start "$container_id"
+          echo "::set-output name=id::$container_id"
+        
+      - name: Login to registry
```
matthchr, line 30: Does the `docker/login-action` not work for us here for some reason?

matthchr: I ask just because that's what we use in the other files

.github/workflows/master-validation.yml
```diff
@@ -17,13 +17,21 @@ jobs:
...
         with:
+          fetch-depth: 0 # required to access tags
           submodules: 'true'
 
-      - name: Build devcontainer image
-        run: docker build --tag devcontainer:1.0 .devcontainer
+      - name: Build & run devcontainer image
+        # this always builds a new image from scratch rather than from the build-devcontainer-image workflow output
+        # so that we pick up the latest versions of everything
+        id: devcontainer
```
matthchr, line 26: Same question about leaving a comment here noting that if you change this you should also probably change the other usages of this pattern in this folder?

Taskfile.yml
```diff
@@ -182,16 +175,30 @@ tasks:
...
     dir: "{{.CONTROLLER_ROOT}}"
     deps: [controller:build]
     sources:
     - Dockerfile
     - ./bin/{{.CONTROLLER_APP}}
     cmds:
     - docker build . -t {{.CONTROLLER_DOCKER_IMAGE}}
+
+  controller:docker-build-and-save:
+    desc: Generates the {{.CONTROLLER_APP}} Docker file.
```
matthchr, line 188: minor: This is sorta confusingly phrased? A dockerfile is the input to create a docker image, yes? maybe:
```suggestion
    desc: Builds the {{.CONTROLLER_APP}} Docker image and saves it using docker save.
```

Taskfile.yml
```diff
@@ -182,16 +175,30 @@ tasks:
...
+    dir: "{{.CONTROLLER_ROOT}}"
+    deps: [controller:docker-build]
+    cmds:
     - docker save {{.CONTROLLER_DOCKER_IMAGE}} > bin/{{.CONTROLLER_DOCKER_IMAGE}}.tar
 
+  controller:docker-push-remote:
```
matthchr, line 194: minor: Maybe fix my awkwardly named `controller:docker-push-local-reg` to `controller:docker-push-local` to match this?

hack/generated/config/manager/manager_image_patch.yaml
```diff
@@ -8,7 +8,7 @@ spec:
     spec:
       containers:
         # Change the value of image field below to your controller image URL
-        - image: localhost:5000/azure-service-operator:latest
+        - image: localhost:5000/azureserviceoperator:latest
```
matthchr, line 11: Do we need to somehow disambiguate this from the docker image published for v1?

matthchr: Or will `v1` just always be `azureserviceoperator:1.x` and `v2` is always `azureserviceoperator:2.x`? If so that make sense.

scripts/build-version.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python3
+
+import re
+import subprocess
+import sys
+
+def to_semver(gitVersion):
```
matthchr, line 7: minor: `git_version` in Python is more idiomatic.

You can also throw on a `: string` if you want. It doesn't do anything but helps aid in clarity.

scripts/build-version.py
```diff
@@ -0,0 +1,66 @@
...
+
+    if len(sys.argv) != 2:
+        me = sys.argv[0]
+        sys.stderr.write(f"usage: {me} <tag-prefix>\n")
+        sys.stderr.write(f"example: {me} v2\n")
+        import doctest
+        print(f"Running tests: {doctest.testmod()}")
+        sys.exit(1)
+
+    versionPrefix = sys.argv[1]
```
matthchr, line 57: minor: `version_prefix`. Same with other variables below.

Taskfile.yml
```diff
@@ -182,16 +175,30 @@ tasks:
...
+    desc: Pushes the {{.CONTROLLER_APP}} Docker image.
+    dir: "{{.CONTROLLER_ROOT}}"
+    deps: [controller:docker-build]
+    cmds:
+    - docker tag {{.CONTROLLER_DOCKER_IMAGE}} "{{.DOCKER_PUSH_TARGET}}/{{.CONTROLLER_DOCKER_IMAGE}}"
```
matthchr, line 199: Where does `DOCKER_PUSH_TARGET` come from? I don't see it set anywhere?

matthchr: Ah, it's an environment variable that must be set by the caller. Just wondering, if it's not set is the error obvious?

matthchr (commented):
Looks good to me - some comments and questions.

It would be good to also have a document on what exactly the release process is. We have one for ASOv1 [here](https://github.com/Azure/azure-service-operator/blob/master/docs/howto/release.md). Maybe you could update that to make a section for ASOv1 and another section for ASOv2 and write down the steps?

matthchr:
One other thought: Do we already have the infrastructure set up to run `task controller:generate-kustomzie` and include that in our artifacts or something? We'll need more than just the actual docker image for a release (even if that release doesn't include a Helm chart yet because we haven't set one up)

.github/workflows/create-release.yml
Porges: The login has to happen inside the (dev)container, not on the host.

hack/generated/config/manager/manager_image_patch.yaml
Porges: @matthchr, yes exactly. And we won’t push v2 to `:latest` until v2 is complete.

Taskfile.yml
Porges: I did have a check here but dropped it during merge. Will add it back in.

Porges:
> One other thought: Do we already have the infrastructure set up to run `task controller:generate-kustomzie` and include that in our artifacts or something? We'll need more than just the actual docker image for a release (even if that release doesn't include a Helm chart yet because we haven't set one up)

This will be in a separate PR for bundling the artifacts we need.

Porges:
> Looks good to me - some comments and questions.
> 
> It would be good to also have a document on what exactly the release process is. We have one for ASOv1 [here](https://github.com/Azure/azure-service-operator/blob/master/docs/howto/release.md). Maybe you could update that to make a section for ASOv1 and another section for ASOv2 and write down the steps?
> 
> Could do this in a separate PR though, doesn't necessarily have to be in this one

Yes, definitely 😁

.github/workflows/create-release.yml
Porges: The problem is that we need things to happen inside the devcontainer. Now, Github has [support for running jobs in a container](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer) but we need the container to really be “cached container image + potential updates to the Dockerfile” so that you can make changes in a PR, and I can’t see a way to achieve that.

Porges: Maybe if I have two jobs, one that builds an image and the other one that uses it…

matthchr: Don't go out of your way to do a big change for this. I was just curious about the two approaches is all. This is fine and in some ways "better" because it's not opaque and we can see what is happening.

hack/generated/config/manager/manager_image_patch.yaml
matthchr: `:latest` is basically "useless" in our use case anyway I think, since pin releases to a particular version even for ASOv1 so that people don't accidentally get updates when we push new container images.

With that said, what you're saying makes sense. Looks like a good plan to me!

matthchr (approved):

.github/workflows/create-release.yml
Porges: …but that’s going to be tricky, would have to push each image to the registry.

Porges: Going to leave this for now :)

theunrepentantgeek (approved):
Looks good.

gif

PR #1694, opened 2021-08-05
Force top level status properties to be optional

matthchr:
There are times when `Status` doesn't exist, such as when an object is first created. There's a bit of an unwritten rule that:
1. `Status` is not a pointer on the resource.
2. All top level status properties have `omitempty` set.

This way when there isn't a status, there's not a bunch of extra `""`'s returned.

This is especially important for some multi-write cases between spec and status, as `""`'s returned can overwrite in-memory data before it has a chance to be sent via a `client.Status().Update()` call

gif

### DIFF
--- a/hack/generator/pkg/codegen/pipeline/make_status_properties_optional.go
+++ b/hack/generator/pkg/codegen/pipeline/make_status_properties_optional.go
@@ -1,0 +1,76 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package pipeline
+
+import (
+	"context"
+
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/klog/v2"
+
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/astmodel"
+)
+
+const MakeStatusPropertiesOptionalStageID = "makeStatusPropertiesOptional"
+
+// MakeStatusPropertiesOptional makes all top level Status properties optional. This is required because Status itself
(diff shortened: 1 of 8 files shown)


Porges (approved):

hack/generator/pkg/codegen/pipeline/make_status_properties_optional.go
```diff
@@ -0,0 +1,76 @@
...
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/klog/v2"
+
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/astmodel"
+)
+
+const MakeStatusPropertiesOptionalStageID = "makeStatusPropertiesOptional"
+
+// MakeStatusPropertiesOptional makes all top level Status properties optional. This is required because Status itself
+// is actually optional (it's not set initially) and if any top level properties are not optional the end up
```
matthchr, line 20:
```suggestion
// is actually optional (it's not set initially) and if any top level properties are not optional they end up
```

hack/generator/pkg/codegen/pipeline/make_status_properties_optional.go
```diff
@@ -0,0 +1,76 @@
...
+			}
+
+			remaining := state.Types().Except(result)
+			result.AddTypes(remaining)
+
+			return state.WithTypes(result), nil
+		})
+}
+
+// addConditions adds a "Conditions" property to the status type
```
matthchr, line 56: This doc is wrong

PR #1695, opened 2021-08-05
Update Azure SDK version and update Helm chart parameters

### LINKED ISSUE #1379: Add variable for image gcr.io/kubebuilder/kube-rbac-proxy:v0.5.0
My k8s cluster only has access to my private image-registry and not to gcr.io.
Please extend the helm-chart and add a variable for the image of kube-rbac-proxy so I can replace it with the private-registry path.

**Describe the improvement**
Support private-registry only.

matthchr:
Closes #1379

* Use updated Azure Go SDK Environment for MySQL URLs
* Add Helm parameter for kube-rbac-proxy


controllers/mysqlserver_controller.go
```diff
@@ -68,7 +68,7 @@ func makeMapFunc(mgr ctrl.Manager) func(o client.Object) []reconcile.Request {
 		// This should be fast since the list of items it's going through should always be cached
 		// locally with the shared informer.
 		// Unfortunately we don't have a ctx we can use here... maybe need to file a bug?
```
theunrepentantgeek, line 70: Do file an issue and give the # here.

pkg/resourcemanager/mysql/mysqlhelper.go
```diff
@@ -31,43 +31,8 @@ const DriverName = "mysql"
...
-	} else if envName == "AzureGermanCloud" {
-		return "mysql.database.cloudapi.de"
-	}
-
-	return "mysql.database.azure.com"
-}
-
 func GetFullSQLServerName(serverName string) string {
-	return serverName + "." + GetMySQLDatabaseDNSSuffix()
+	return serverName + "." + config.Environment().MySQLDatabaseDNSSuffix
```
theunrepentantgeek, line 35: Good idea to move this to configuration.

theunrepentantgeek (approved):
Looks good.

gif

pkg/resourcemanager/mysql/mysqlhelper.go
matthchr: Making sure I understand: Are you saying that the change made here is good, _or_ that I should move this whole `GetFullSQLServerName` method to `config`?

I _think_ you're saying the first thing but just wanted to confirm.

controllers/mysqlserver_controller.go
matthchr: Filed - although when I had talked with the controller-runtime folks before they weren't super optimistic about being able to do anything about this...

pkg/resourcemanager/mysql/mysqlhelper.go
theunrepentantgeek: I was confirming that _the change you'd already done_ was a good one, moving knowledge of the DNS suffix for MySQL Databases out of our code to where it can be customized without a recompilation.

matthchr: Thanks for the confirmation

controllers/mysqlserver_controller.go
```diff
@@ -67,14 +67,14 @@ func makeMapFunc(mgr ctrl.Manager) func(o client.Object) []reconcile.Request {
 
 		// This should be fast since the list of items it's going through should always be cached
 		// locally with the shared informer.
-		// Unfortunately we don't have a ctx we can use here... maybe need to file a bug?
-		ctx, cancel := context.WithTimeout(context.Background(), 10 * time.Second)
+		// Unfortunately we don't have a ctx we can use here, see https://github.com/kubernetes-sigs/controller-runtime/issues/1628
```
babbageclunk, line 70: Nice

babbageclunk (commented):
LGTM!

babbageclunk (approved):

PR #1696, opened 2021-08-06
Use Conditions rather than annotations to determine reconcile state

### LINKED ISSUE #1615: Implement the customer visible state model defined in #1614
This is a tracking item for implementing what is decided on by the referenced design in #1614

matthchr:
Closes #1615 

gif

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/hack/generated/pkg/genruntime/helpers.go
+++ b/hack/generated/pkg/genruntime/helpers.go
@@ -11,7 +11,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 )
 
-func NewObjectFromObject(obj client.Object, scheme *runtime.Scheme) (client.Object, error) {
+// NewObjectFromExemplar creates a new client.Object with the same GVK as the provided client.Object.
+// The supplied client.Object is not changed and the returned client.Object is empty.
+func NewObjectFromExemplar(obj client.Object, scheme *runtime.Scheme) (client.Object, error) {
 	gvk, err := apiutil.GVKForObject(obj, scheme)
 	if err != nil {
 		return nil, err
--- a/hack/generated/pkg/genruntime/helpers.go
+++ b/hack/generated/pkg/genruntime/helpers.go
@@ -1,0 +1,26 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+package genruntime
+
+import (
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+)
+
+func NewObjectFromObject(obj client.Object, scheme *runtime.Scheme) (client.Object, error) {
+	gvk, err := apiutil.GVKForObject(obj, scheme)
+	if err != nil {
+		return nil, err
+	}
+
+	// Create a fresh destination to deserialize to
+	newObj, err := scheme.New(gvk)
+	if err != nil {
+		return nil, err
+	}
+	return newObj.(client.Object), nil
+}
--- a/hack/generated/pkg/reconcilers/deployment_error_classifier_test.go
+++ b/hack/generated/pkg/reconcilers/deployment_error_classifier_test.go
@@ -8,14 +8,20 @@ package reconcilers_test
 import (
 	"testing"
 
+	. "github.com/onsi/gomega"
+
 	"github.com/Azure/azure-service-operator/hack/generated/pkg/armclient"
 	"github.com/Azure/azure-service-operator/hack/generated/pkg/reconcilers"
-	. "github.com/onsi/gomega"
 )
 
 var badRequestError = armclient.DeploymentError{
 	Code:    "BadRequest",
 	Message: "That was not a good request",
+}
+
+var badRequestError2 = armclient.DeploymentError{
+	Code:    "BadRequest",
+	Message: "There was something wrong with that request",
 }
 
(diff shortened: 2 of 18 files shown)


hack/generated/pkg/genruntime/helpers.go
```diff
@@ -0,0 +1,26 @@
...
+
+package genruntime
+
+import (
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+)
+
+func NewObjectFromObject(obj client.Object, scheme *runtime.Scheme) (client.Object, error) {
```
matthchr, line 14: Need comment

theunrepentantgeek: Is there a better name for this method? Maybe `NewInstanceFromExemplar()`?

hack/generated/pkg/reconcilers/deployment_error_classifier_test.go
```diff
@@ -43,47 +49,97 @@ func newError(inner ...armclient.DeploymentError) *armclient.DeploymentError {
...
 func Test_NilError_IsRetryable(t *testing.T) {
 	g := NewGomegaWithT(t)
-	g.Expect(reconcilers.ClassifyDeploymentError(nil)).To(Equal(reconcilers.DeploymentErrorRetryable))
+	expected := reconcilers.DeploymentErrorDetails{
+		Classification: reconcilers.DeploymentErrorRetryable,
+		Code:           reconcilers.UnknownErrorCode,
+		Message:        reconcilers.UnknownErrorMessage,
+	}
+	g.Expect(reconcilers.ClassifyDeploymentError(nil)).To(Equal(expected))
 }
```
theunrepentantgeek, line 58: I think you could combine all these tests into a single test with subtests, using the `cases := struct{...}{...}` style we've used elsewhere. 

I can't decide whether you should, though.

hack/generated/pkg/reconcilers/deployment_error_translator.go
```diff
@@ -0,0 +1,67 @@
...
+
+package reconcilers
+
+import (
+	"github.com/Azure/go-autorest/autorest/azure"
+
+	"github.com/Azure/azure-service-operator/hack/generated/pkg/armclient"
+)
+
+func stringOrEmptyString(value interface{}) string {
```
theunrepentantgeek, line 14: Does this need to support `fmt.Stringer` too?

theunrepentantgeek (approved):
Looks good.

gif

hack/generated/pkg/reconcilers/deployment_error_classifier_test.go
matthchr: I thought about it when I initially wrote these. It ended up not seeming worth it though because the bulk of the lines of code in each test is defining the input error and the expected `DeploymentErrorDetails`, which is code that isn't duplicate. Combining into a single test with a loop to execute the list of inputs then would only really be saving us the duplicate assertion lines (and `NewGomegaWithT(t)`), which there is only one of in each test.

Given that the reduction in duplicate code isn't all that much, and the `struct {...}{...}` style tests have other disadvantages (harder to run just a single test, harder to break into them with a debugger) it felt better to go with this style. Especially because some of the cases do subtly differ from one another in a way that would increase the inner-loops complexity slightly in the `struct {...}{...}` testcase style (supporting multiple error inputs, supporting nil error input).

The `struct {...}{...}` feels more suited to tests where you have a moderate number of assertions which are being repeated across various styles of input. Bonus if you can craft that input itself in a loop somehow.

Happy to change this though if people disagree.

hack/generated/pkg/reconcilers/deployment_error_translator.go
matthchr: If this were going to be an exported method or used in other contexts then yea, supporting `fmt.Stringer` seems like it'd be a good idea.

In this context my intent was really as a helper method for parsing the horrible `[]map[string]interface{}` error from the Azure SDK and so it has a pretty narrow expected usage context. If it was going to be used more broadly I think we'd also want it to not silently allow non-string (and non `fmt.Stringer`) and just turn those into `""`. That case feels more like an `err` possibly? Adding an `err` though muddies the actual usage as then I'd have to discard it (see where it's called down below for what I mean).

matthchr: I've added a comment to this method explaining its behavior (which I'm hoping will help steer anybody else in this package away from it unless they want this specific behavior)

hack/generated/pkg/genruntime/helpers.go
matthchr: I think `NewObject` is important because the thing we're making is a `client.Object`. The second half though I don't have an issue changing: `NewObjectFromExemplar`?

I've made that change and also added a comment.

theunrepentantgeek: Looks good.

hack/generated/pkg/reconcilers/deployment_error_classifier_test.go
theunrepentantgeek: Fair enough.

hack/generated/pkg/reconcilers/deployment_error_translator.go
theunrepentantgeek: Looks good.

babbageclunk (commented):
This looks great to me - I was a bit confused about the positive condition builder at first but I'm alright with that now. A few super-minor suggestions.

PR #1717, opened 2021-08-12
Add new options to support KeyVault soft delete

### LINKED ISSUE #1686: Bug: Deleting and recreating resource while KeyVault soft delete is enabled causes an error
**Describe the bug**
If ASO is configured to store secrets in KeyVault, and that KeyVault has SoftDelete enabled, then when an ASO resource is deleted and recreated with the same name an error will occur because the newly created resource is unable to create a secret with the same name as a currently soft-deleted secret.

**To Reproduce**
1. Use KeyVault for secret storage.
2. Enable soft delete in the KeyVault.
3. Create any resource with a secret (AzureSQLServer for example).
4. Delete that resource
5. Create the same resource again
6. Observe the error

**Expected behavior**
The recreation of a previously deleted resource works.

matthchr:
Closes #1686

Adds two new boolean flags `PURGE_DELETED_KEYVAULT_SECRETS` and `RECOVER_SOFT_DELETED_KEYVAULT_SECRETS`.

If `PURGE_DELETED_KEYVAULT_SECRETS` is set to `True`, ASO will issue a Purge request after every KeyVault `Delete` request.
If `RECOVER_SOFT_DELETED_KEYVAULT_SECRETS` is set to `True` and ASO hits the error `ObjectIsDeletedButRecoverable` when creating a secret, a `Recover` request will be issued to KeyVault and then the operator will try to create the key again (until it can succeed) 

Testing this is awkward. The more "modes" we have the more awkward testing becomes. We have a solution for this in the V2 operator because we run multiple instances of envtest (which has multiple instances of the controller which could have different configurations).

**If applicable**:
- [x] this PR contains tests


pkg/secrets/keyvault/client.go
```diff
@@ -183,7 +194,57 @@ func (k *SecretClient) Upsert(ctx context.Context, key secrets.SecretKey, data m
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func isErrSecretWasSoftDeleted(err error) bool {
+	var azureErr = &azure.RequestError{}
```
babbageclunk, line 207:
```suggestion
	var azureErr *azure.RequestError
```

pkg/secrets/keyvault/client.go
```diff
@@ -183,7 +194,57 @@ func (k *SecretClient) Upsert(ctx context.Context, key secrets.SecretKey, data m
...
+	if errors.As(err, &azureErr) {
+		if azureErr == nil && azureErr.ServiceError == nil {
```
babbageclunk, line 209: I don't think `azureErr` can be nil at this point. Which is lucky because `azureErr.ServiceError` would panic, right? Should this have been `||`?

babbageclunk (approved):
This looks great - a couple of questions in the keyvault secret client changes.

pkg/secrets/keyvault/client.go
matthchr: lol yes this should have been `||`

pkg/secrets/keyvault/client.go
matthchr: Good catch

pkg/secrets/keyvault/client.go
matthchr: but I'll go ahead and just drop this initial check as you're right, I don't think it can be `nil` here either.

matthchr: (kept the `azureErr.ServiceErrpr == nil` bit, just dropped the `azureErr == nil` bit)

controllers/suite_test.go
```diff
@@ -55,6 +55,11 @@ func setup() error {
 	// Uncomment the below to run the tests in the old v1 naming mode
 	// envy.Set("AZURE_SECRET_NAMING_VERSION", "1")
 
+	// TODO: Should this just default to true?
```
matthchr, line 58: Need to decide on this before merging the PR

matthchr: @babbageclunk do you have thoughts here?

babbageclunk: Without this on, everyone who has soft-delete on in their KV is potentially going to hit the same problem, right? (And it's not going to affect KVs without soft-delete.) There's versioning so that it's not lossy in the way that a purge would be. How fine-grained are the permissions? Is permission to recover the same as deleting? 
I think defaulting it to true would make sense.

matthchr: Yeah it's not lossy, and KeyVault is actually requiring softdelete to be enabled starting sometime end of this year I think so it will become a common problem.

matthchr: There is a different permission (recover) that is required to perform a recovery. But if we don't have such a permission and we're in this case we have the choice of:

Failing because we can't create the key
Failing because we can't recover the key

So either way... it's going to fail in this situation.

babbageclunk: Yeah, not much we can do in that case, and the recovery behaviour is much better than purging on delete. I say do it!

theunrepentantgeek (approved):
Looks good. A couple of minor comments.

gif

controllers/suite_test.go
theunrepentantgeek: I don't see a downside - worst case is that we fail in a scenario where we would have failed _anyway_.

The upside is that we cleanly handle a situation without failing, improving resilience.

charts/azure-service-operator/values.yaml
```diff
@@ -25,6 +25,18 @@ azureUseMI: False
 # azureSecretNamingVersion allows choosing the algorithm used to derive secret names. Version 2 is recommended.
 azureSecretNamingVersion: "2"
 
+# purgeDeletedKeyVaultSecrets determines if the operator should issue a secret Purge request in addition
+# to Delete when deleting secrets in Azure Key Vault. This only applies to secrets that are stored in Azure Key Vault.
+# It does nothing if the secret is stored in Kubernetes.
+purgeDeletedKeyVaultSecrets: False
```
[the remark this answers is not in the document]
matthchr: They're not mutually exclusive, you could set both to true if you say, wanted to purge on delete but recover on failure in case somehow something wasn't purged.

Do you think they _should_ be mutually exclusive? They could be combined into a `softDeleteHandling` or something, that was either `Purge|Recover|None`

controllers/azuresql_combined_test.go
```diff
@@ -338,3 +339,47 @@ func TestAzureSqlServerCombinedHappyPath(t *testing.T) {
...
+
+func TestAzureSqlServer_KeyVaultSoftDelete_CreateDeleteCreateAgain(t *testing.T) {
+	t.Parallel()
+	defer PanicRecover(t)
+	ctx := context.Background()
+	require := require.New(t)
+
+	rgLocation := "westus2"
+
+	// Create a KeyVault with soft delete enabled that we can use to perform our tests
```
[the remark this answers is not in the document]
matthchr: That's a _very_ good question. I have a helper method that does it for the client tests but I'm not using it here. It might not be getting set at all.

Very good eye.

controllers/suite_test.go
matthchr: Done, the default is now true

controllers/azuresql_combined_test.go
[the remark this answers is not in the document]
matthchr: Fixed this

PR #1719, opened 2021-08-16
Fix EOLN settings so devcontainer works on Windows

theunrepentantgeek:

Trying to use the devcontainer on Windows was failing because the EOL characters in certain files were incorrect.

It turns out the option in `.gitattributes` is case sensitive ... because _reasons_ ... so they were having no effect. 

Also changed a couple other needed files to use Linux LF endings.

gif


Porges (approved):

.gitattributes
```diff
@@ -1,4 +1,7 @@
-*.sh     text eol=LF
-Makefile text eol=LF
+*.sh     text eol=lf
+*.py     text eol=lf
+*.golden eol=lf
+Makefile text eol=lf
+Taskfile text eol=lf
```
Porges, line 5:
```suggestion
Taskfile.{yml,yaml} text eol=lf
```

Porges (approved):

PR #1723, opened 2021-08-16
Remove envtest job timeout for now

babbageclunk:
It seems like there's a problem in the job that's causing it to be killed by the timeout, but the way the job is killed prevents us from seeing the output which would let us fix the underlying problem in whichever test is failing.

[Example of the job timeout](https://dev.azure.com/azure/azure-service-operator/_build/results?buildId=27948&view=logs&j=ca395085-040a-526b-2ce8-bdc85f692774&t=cb33ee24-f492-54d2-727a-ae0a7bb84b2c)

gif


matthchr (approved):

theunrepentantgeek (approved):

babbageclunk:
> /azp run

Wasn't it already running? Had it failed?

theunrepentantgeek (approved):
Still looks good.

gif

Issue #1722, opened 2021-08-16
Bug: Status types don't understand Swagger discriminator and so are missing fields


matthchr:
**Describe the bug**
When the Swagger specification has a type which is modelling inheritance, such as [Cosmos DB BackupPolicy](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cosmos-db/resource-manager/Microsoft.DocumentDB/stable/2021-05-15/cosmos-db.json#L8392), the Status types don't realize this and end up generating a type that looks like:
```go
//Generated from:
type BackupPolicy_Status struct {
	// +kubebuilder:validation:Required
	Type BackupPolicyType_Status `json:"type"`
}
```

**Expected behavior**
These types should probably be modelled in k8s as a [Kubernetes union type](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/1027-api-unions/README.md), which is basically a type with a discriminator and N mutually exclusive properties.

matthchr:
This is the "Status side" of #1721

matthchr:
@porges did your PR #1912 solve this?

matthchr:
@porges Can you confirm if this is actually still an issue?

theunrepentantgeek:
Resolving as fixed. In the current Swagger WIP branch, `BackupPolicy` is correctly turned into a union type as we want.

In `api\documentdb\v1beta20210515\database_account_types_gen.go` we have:

``` go
type BackupPolicy struct {
	// Continuous: Mutually exclusive with all other properties
	Continuous *ContinuousModeBackupPolicy `json:"continuous,omitempty"`

	// Periodic: Mutually exclusive with all other properties
	Periodic *PeriodicModeBackupPolicy `json:"periodic,omitempty"`
}
```

Matching this, in `api\documentdb\v1beta20210515\database_account_spec_arm_types_gen.go` we have:

``` go
type BackupPolicy_ARM struct {
	// Continuous: Mutually exclusive with all other properties
	Continuous *ContinuousModeBackupPolicy_ARM `json:"continuous,omitempty"`

	// Periodic: Mutually exclusive with all other properties
	Periodic *PeriodicModeBackupPolicy_ARM `json:"periodic,omitempty"`
}
```

which in turn has the desired `MarshalJSON` and `UnmarshalJSON` methods:

``` go
// MarshalJSON defers JSON marshaling to the first non-nil property, because BackupPolicy_ARM represents a discriminated union (JSON OneOf)
[...]

Issue #1721, opened 2021-08-16
Spec types with OneOf must roundtrip with discriminator


matthchr:
**Describe the current behavior**
ASOv2 code-generated ARM types that contain a property whose type is a `oneOf` serialize correct in one direction (outbound to ARM), but don't serialize correctly inbound (from ARM).

**Describe the improvement**
Serialization should be supported in both directions. We will need this for Spec types in order to do diffing.

The Cosmos DB (DocumentDB) `BackupPolicy` which we already generate shows exactly how things are shaped currently. It has a  `MarshalJSON` defined but no `UnmarshalJSON`. You can see the Swagger associated with this type at [Cosmos DB BackupPolicy](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cosmos-db/resource-manager/Microsoft.DocumentDB/stable/2021-05-15/cosmos-db.json#L8392).

**Additional context**
I _think_ it is the case that all `oneOf` types in the ARM schema are actually attempting to model the discriminator pattern linked above. Without a discriminator it's not possible (or at least very hard) to do serialization from ARM into a Spec type (how do you know which field it matches?). With a discriminator though it's much easier.

Provided all usages of `oneOf` are actually with a discriminator, we should be able to figure that out and make use of it to implement `UnmarshalJSON` on Spec types.

matthchr:
This is the "spec side" of #1722

Issue #1366, opened 2021-01-29
Resources should diff with Azure instead of using SpecHash


matthchr:
**Describe the current behavior**
When a reconciliation is performed on a resource its spec is serialized to JSON and hashed, then diffed against a stored hash to determine if anything has changed.

**Describe the improvement**
We should be diffing against the state of the resource in Azure, not the state of the resource in Kuberentes. Without diffing against Azure we cannot notice when something has changed behind the back of the operator and correct it. Additionally, diffing against Azure means that the 10h "slow" reconciliation events have a chance to correct resources which got into a bad state on the Kubernetes side (possibly due to a bug in a previous version of the operator for instance).

matthchr:
Closing this in favor of #1491

PR #1724, opened 2021-08-17
Suppress classes causing test failures

theunrepentantgeek:

A very small number of ARM State classes representing OneOf conditions can't currently round trip as they override marshaling _**to**_ JSON without overriding marshaling _**from**_ JSON (See #1721 #1722)

We want to revisit these and fix things up as a part of implementing diff'ing (#1366), so I took a very simple approach to suppressing the test failures.

Also included, some improvements to the way the output of property tests is formatted (the default uses a 75 column wordwrap which is a bit narrow for our use).

gif


Porges (approved):

matthchr (approved):

hack/generator/pkg/codegen/pipeline/json_serialization_test_cases.go
```diff
@@ -48,22 +48,47 @@ func InjectJsonSerializationTests(idFactory astmodel.IdentifierFactory) Stage {
...
+// NeedsTest returns true if we should generate a testcase for the specified definition
 func (s *objectSerializationTestCaseFactory) NeedsTest(def astmodel.TypeDefinition) bool {
 	_, ok := astmodel.AsPropertyContainer(def.Type())
-	return ok
+	if !ok {
+		// Can only generate tests for property containers
+		return false
+	}
+
+	// Check for types that we need to suppress - these are ARM types that don't currently round trip because they're
```
matthchr, line 78: Might make sense to put this comment by the specific classes suppressions, and not here in the generic suppression handling?

matthchr: In case there are other reasons for suppression that aren't due to OneOf

PR #1728, opened 2021-08-18
Folder reorganization

Porges:
Don’t read me yet.


matthchr:
**Some thoughts** (ordered roughly in order of severity):

Do we really want the generator and controller to be the same go module? I sorta liked the fact that they were two separate modules with two distinct sets of dependencies - it lets us keep the generator dependencies down while the controller dependencies will be larger (because it's gonna need all the k8s stuff).

controller_resources_gen.go, controller_resources.go, and generic_controller.go shouldn't be in `test/controller`, Those should go next to the controller code in `internal` probably? It's registering the api types with the generic controller, and the generic controller itself (not the entrypoint the the main lib the entrypoint uses).

Part of me likes having the controller and the generator next to one another, but I wonder if it'd be better if the generator wasn't in `v2`, as it's not really v2 of the generator? I guess so long as the generator and the generated code are in the same repo we're constantly presented with situations that push us towards treating/versioning them similarly (such as with go folder structure) even though really they're distinct. 

I think it's kinda awkward having `crossplane` inside `v2` as well as then we're not having parity with folder structure. I would expect: `/crossplane/api` which compares with `/v2/api`, not `/v2/crossplane/api` when there is also `/v2/api`. With that said I think the crossplane folder here is entirely temporary and we could just avoid committing it for now if we wanted. Alternatively we can put it in v2 and just move it later, so I'm not saying this must be moved _today_, just that it's not where I think it should be if that part of things were going public soon.

Other possible changes:
- `specs` has such overloaded meaning. We could just call it `submodules`? I don't hate `specs` though if we decide to keep it that.
- I'm not sold on `tests`. The reference says "Additional external test apps and test data. Feel free to structure the /test directory anyway you want." Is anything in there really "external"? It doesn't feel like it to me. I guess the win is that the integration-y tests and test-common stuff is separated though, which I do like... [CAPZ](https://github.com/kubernetes-sigs/cluster-api-provider-azure/tree/main) does have e2e tests in `test/e2e` but also has `internal/test`. [Cluster API AWS](https://github.com/kubernetes-sigs/cluster-api-provider-aws/tree/main/controllers) seems to put at least some envtest tests right next to the controllers inside `controllers` rather than in `test/e2e` although they have that too. After typing a block of text on this, I dislike it less than I started with, especially if the generator/controller single module is changed.

Porges:
Github needs per-paragraph reaction emoji.

matthchr:
FWIW I played around with it some more and the tests thing I'm less sure about now. I think I like having the tests themselves near the code in question, but then "test library stuff" does feel like it should be in `test/common` or `internal/testcommon`.

matthchr:
@porges Are you abandoning this in favor of doing it in smaller chunks?

Porges:
> @Porges Are you abandoning this in favor of doing it in smaller chunks?

Yes; leaving it here for reference at the moment.

PR #1731, opened 2021-08-20
Improve recording default timeouts and controller backoff timing

matthchr:
Improve recording test default timeouts

  * In the abscence of a timeout specified on the cmdline, use recording mode to determine what the timeout should be. When recording, timeout is longer (15m), and when replaying the timeout is shorter (2m).

Improve controller exponential backoff and timing

* Reduced the maximum wait time (it was >15m). Polling speed is now variable based on test mode. In replay mode, polling is fast. In recording mode, polling is slower. 

gif

### DIFF
--- a/hack/generated/controllers/generic_controller.go
+++ b/hack/generated/controllers/generic_controller.go
@@ -13,11 +13,13 @@ import (
 	"github.com/benbjohnson/clock"
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
+	"golang.org/x/time/rate"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/conversion"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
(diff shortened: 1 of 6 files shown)


babbageclunk:
Great gif but I have to use the inspector to delete it otherwise I can't read anything on the page.

hack/generated/controllers/generic_controller.go
```diff
@@ -239,3 +241,15 @@ func (gr *GenericReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 
 	return result, nil
 }
+
+// NewRateLimiter creates a new workqueue.Ratelimiter for use controlling the speed of reconciliation.
+// In throttles individual requests exponentially and also controls for multiple requests.
```
theunrepentantgeek, line 246:
```suggestion
// It throttles individual requests exponentially and also controls for multiple requests.
```

theunrepentantgeek (approved):
Looks good.

gif

hack/generated/controllers/generic_controller.go
matthchr: Fixed

PR #1732, opened 2021-08-20
Remove all status validations

matthchr:
Removes validation from status types.

This has the unfortunate impact that we won't be able to share types between spec and status... which is pretty unfortunate. See the long explanation in remove_status_property_validations.go for the reason behind this change.

Note: Testing this is difficult because of our testing gap on status types, but it _is_ covered by the upcoming AKS E2E test (not in PR yet) as without this that test fails.


Porges (approved):

theunrepentantgeek (approved):
Looks good.

gif

PR #1738, opened 2021-08-23
Update kind

matthchr:
The one we have is a bit old and I _think_ has a bug where sometimes nodes won't start? The new one at least works for me consistently.

Also deleted the makefiles in hack because we don't use them for anything anymore.


theunrepentantgeek (approved):
Works for me.

gif

PR #1741, opened 2021-08-24
Rename _apis directory to apis and commit it

Porges:
This is the first step of reorganizing the code for v2.


theunrepentantgeek (approved):
Woot!

Hulk Lift!
gif

hack/generated/apis/microsoft.documentdb/versions_matrix.md
```diff
@@ -0,0 +1,56 @@
+|                                                               | v1alpha1api20210515 |
+|---------------------------------------------------------------|---------------------|
```
matthchr, line 2: Do we actually want the version matrix files committed? I sorta thought they were just for our consumption? Maybe we should gitignore `md` files in `apis`?

Taskfile.yml
```diff
@@ -246,14 +246,13 @@ tasks:
...
     sources:
-    - "_apis/**/*_gen.go" # depends on all generated types
+    - "apis/**/*_gen.go" # depends on all generated types
     cmds:
-    - mkdir -p ./_apis
-    - find ./_apis -type f -name "zz_generated.*" -delete
+    - mkdir -p ./apis
+    - find ./apis -type f -name "zz_generated.*" -delete
     - if [ -d "./config/crd/bases" ]; then find "./config/crd/bases" -type f -delete; fi
-    # must CD into _apis as it is ignored by default (this is on purpose to not kill gopls!)
```
matthchr, line 254: I think we no longer need to `cd apis` when running controller-gen since it isn't prefixed with `_` and so isn't ignored? So `./...` will match it?

matthchr (approved):
Approved but had a few minor comments

hack/generated/apis/microsoft.documentdb/versions_matrix.md
theunrepentantgeek: No, I always intended that these would be committed as a reference for which types have been generated in each specific version. I've got a ToDo to work out why the headers are now being omitted from the tables.

PR #1745, opened 2021-08-24
Reduce test polling speed some to reduce flakiness

matthchr:
Obivously this isn't the ideal way to do things, but I can't think of another "easy" fix. The real solution is to stop modifying status and spec at the same time, but I believe in order to do that we need to move some fields out of Spec and into Status, which will take some effort. This should unblock us for now.

### DIFF
--- a/hack/generated/pkg/testcommon/kube_per_test_context.go
+++ b/hack/generated/pkg/testcommon/kube_per_test_context.go
@@ -247,9 +247,11 @@ func (tc *KubePerTestContext) DefaultTimeout() time.Duration {
 
 // PollingIntervalReplaying is the polling interval to use when replaying.
 // TODO: Setting this really low sometimes seems to cause
-// TODO: updating resource: Operation cannot be fulfilled: the object has been modified; please apply your changes to the latest version and try again
-// TODO: I don't know why -- possibly we're starving APIServer of cycles because of fast polling loops which prevents it from processing cache updates or something?
-var PollingIntervalReplaying = 200 * time.Millisecond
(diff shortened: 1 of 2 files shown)


hack/generated/pkg/testcommon/kube_per_test_context.go
```diff
@@ -247,9 +247,11 @@ func (tc *KubePerTestContext) DefaultTimeout() time.Duration {
 
 // PollingIntervalReplaying is the polling interval to use when replaying.
 // TODO: Setting this really low sometimes seems to cause
-// TODO: updating resource: Operation cannot be fulfilled: the object has been modified; please apply your changes to the latest version and try again
-// TODO: I don't know why -- possibly we're starving APIServer of cycles because of fast polling loops which prevents it from processing cache updates or something?
-var PollingIntervalReplaying = 200 * time.Millisecond
+// TODO: updating resource: Operation cannot be fulfilled: the object has been modified; please apply your changes to the latest version and try again.
+// TODO: This happens when the test sees a Status update and makes an update to the resource while racing with the Spec update
+// TODO: in azure_deployment_reconciler CommitUpdate. If we fix https://github.com/Azure/azure-service-operator/issues/1744 we can reduce this
+// TODO: further.
```
theunrepentantgeek, line 253: Minor ambiguity - does "reduce this further" mean "reduce the time between polling" or "reduce how often we poll"?

theunrepentantgeek (approved):

hack/generated/pkg/testcommon/kube_per_test_context.go
matthchr: Fixed (I think).

PR #1748, opened 2021-08-25
Unroll self-referential types

matthchr:
This "unrolls" types that have a property that refers to themselves. This _doesn't_ remove all possible cycles, just a particular type of cycle which is common and that we have a "solution" for.

**If applicable**:
- [x] this PR contains tests


hack/generator/pkg/codegen/pipeline/unroll_recursive_types.go
```diff
@@ -0,0 +1,188 @@
...
+// }
+// This usually happens for Error types (CloudError and the like). The recursive references must be removed because
+// controller-tools doesn't support generating "references" (JSON $ref) so it can't support recursive types today.
+// See https://github.com/kubernetes-sigs/controller-tools/issues/489 for more information.
+// Unrolling these types, while required for controller-tools to function, means
+// THAT THE TYPES WE GENERATE DON'T EXACTLY CONFORM TO THE PUBLISHED API!
+// In practice, we think this is ok because the types that reference themselves only return a depth of 1.
+// If we were to unroll all loops (rather than just types that directly reference themselves like we're doing here) that
+// "in practice" observation may no longer hold, so we avoid doing it here (it's also more complicated to do that).
+// TODO: We could require you to name the type in config or code in order to get it unrolled? There are currently 43 types being unrolled by this
```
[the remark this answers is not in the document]
matthchr: I'm definitely not opposed to changing the config we have. In some ways it feels like there are parts that would be better split apart? The Swagger repo has a config per-service type (so one for `microsoft.network`, one for `microsoft.compute`, etc). That keeps the configs somewhat small even though there are tons and tons of services.

In any case, a discussion for someplace other than here =)

Did you have any thoughts on the question I asked above?

hack/generator/pkg/codegen/testdata/RecursiveTypes/Unrolls_recursive_type_multiple_loops.golden
```diff
@@ -0,0 +1,20 @@
...
+type Test struct {
+	Error *Error `json:"error,omitempty"`
+}
+
+//Generated from: https://test.test/schemas/2020-01-01/test.json#/definitions/Error
+type Error struct {
+	Error1 *Error_Unrolled `json:"error1,omitempty"`
+	Error2 *Error_Unrolled `json:"error2,omitempty"`
+	Name   *string         `json:"name,omitempty"`
+}
```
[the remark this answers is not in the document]
matthchr: I'm not sure I follow - they're already JSON annotated and should already deserialize from what ARM is returning to us. All I did was limit the recursive depth.

So the following will deserialize fine:
```json
{
    "name": "Outer error",
    "error1": {
        "name": "Inner error 1"
    },
    "error2": {
        "name": "Inner error 2"
    }
}
```

Whereas this (which is allowed as per the Swagger) will not anymore:
```json
{
    "name": "Outer error",
    "error1": {
        "name": "Inner error 1",
        "error1": {
            "name": "inner inner error 1"
        }
    },
    "error2": {
        "name": "Inner error 2",
        "error2": {
            "name": "inner inner error 2"
        }
    }
}
```

You can see where this comes up in the [ARM common API definitions](https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/common-api-details.md#error-response-content). The actual ARM error types usually look like [this](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/common-types/resource-management/v2/types.json#L285) - note that a lot of specs (Batch, etc) don't use the same type but instead redefine a type with a similar structure (which is why we can't just name a single type to unroll and be done here).

So the answer to "will it deserialize" is yes. We represent a subset of what's actually allowed by the API so it will always deserialize _something_. The harder question is, will that something be *complete*, and that boils down to if services actually in practice nest details inside their error details (so 3+ levels instead of 2). My observation is that they don't and that aligns with ARMs example, but obviously the type they've defined is broader than that so there's no guarantee.

hack/generator/pkg/astmodel/type_visitor.go
```diff
@@ -128,23 +128,29 @@ func IdentityVisitOfPrimitiveType(_ *TypeVisitor, it *PrimitiveType, _ interface
 	return it, nil
 }
 
-func identityVisitObjectTypePerPropertyContext(_ *PropertyDefinition, ctx interface{}) interface{} {
-	return ctx
+func identityVisitObjectTypePerPropertyContext(_ *ObjectType, _ *PropertyDefinition, ctx interface{}) (interface{}, error) {
+	return ctx, nil
```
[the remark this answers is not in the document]
matthchr: I did. The `UnrollObjectTypeProperty` function adheres to this signature, and has a possible error return if/when `AddAllowDuplicates` returns an `err`.

The whole reasoning for this is so that we can have a "per-property handler" for the visitor rather than a per-object type handler. Before this there were a good number of places where we were duplicating 95% of the default object visitor code because there wasn't a per-property hook.

hack/generator/pkg/codegen/pipeline/unroll_recursive_types.go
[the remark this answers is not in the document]
theunrepentantgeek: I'm happy with the approach you've taken here.

hack/generator/pkg/codegen/testdata/RecursiveTypes/Unrolls_recursive_type_multiple_loops.golden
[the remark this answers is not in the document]
theunrepentantgeek: Thanks for clarifying.

theunrepentantgeek (approved):

hack/generator/pkg/codegen/testdata/RecursiveTypes/Unrolls_recursive_type_multiple_loops.golden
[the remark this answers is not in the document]
matthchr: Discussed this and decided to only do this for things with `Error` in the name, and also log an error or warning if we run into one that doesn't match that description.

hack/generator/pkg/codegen/pipeline/unroll_recursive_types.go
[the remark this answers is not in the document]
matthchr: Discussed this and decided to only do this for things with `Error` in the name, and also log an error or warning if we run into one that doesn't match that description.

hack/generator/pkg/astmodel/type_visitor.go
```diff
@@ -128,23 +128,29 @@ func IdentityVisitOfPrimitiveType(_ *TypeVisitor, it *PrimitiveType, _ interface
...
 }
 
 var IdentityVisitOfObjectType = MakeIdentityVisitOfObjectTypeWithPerPropertyContext(identityVisitObjectTypePerPropertyContext)
 
-// MakeIdentityVisitOfObjectTypeWithPerPropertyContext creates an visitor function which creates a per-property context before visiting each
+// MakeIdentityVisitOfObjectTypeWithPerPropertyContext creates a visitor function which creates a per-property context before visiting each
 // property of the ObjectType
 func MakeIdentityVisitOfObjectTypeWithPerPropertyContext(
```
babbageclunk, line 139: This name is too wordy - it would be better to make this just `MakeIdentityVisitOfObjectType`. Since there's no way to call it without passing a `makeCtx` func, and there's no other simpler `MakeIdentityVisitOfObjectType` in the namespace that it needs to be distinguished from, it would be as clear without the extra words in the name. 

Basically the name should say what the function does (makes an identity object type visitor), not how it does it (by making a new context for every property).

hack/generator/pkg/codegen/pipeline/unroll_recursive_types.go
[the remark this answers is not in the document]
babbageclunk: From discussion in chat it sounds like the plan is to do it automatically for any directly self-referential  type with "Error" in the name, and then raise an error for any other directly self-referential type? Or actually, if they don't get fixed automatically will we get a crash later on? I guess it would still be useful to fail fast at this point rather than getting a much more obscure error later.

That sounds good to me.

hack/generator/pkg/codegen/pipeline/unroll_recursive_types.go
```diff
@@ -0,0 +1,188 @@
...
+func UnrollRecursiveTypes() Stage {
+	return MakeStage(
+		UnrollRecursiveTypesStageID,
+		"Unrolls directly recursive types, as they are not supported by controller-gen",
```
babbageclunk, line 48: super minor nit: To make the tense consistent with other pipeline stages.
```suggestion
		"Unroll directly recursive types since they are not supported by controller-gen",
```

hack/generator/pkg/codegen/pipeline/unroll_recursive_types.go
```diff
@@ -0,0 +1,188 @@
...
+		func(ctx context.Context, state *State) (*State, error) {
+
+			result := make(astmodel.Types)
+
+			// Find object types that reference themselves
+
+			fixer := newSimpleRecursiveTypeFixer()
+
+			for _, def := range state.Types() {
+				updatedDef, err := fixer.visitor.VisitDefinition(def, simpleRecursiveTypeFixerContext{name: def.Name()})
```
babbageclunk, line 58: Maybe make this a method on the fixer? Feels like it's reaching into fixer's guts a bit.

hack/generator/pkg/codegen/pipeline/unroll_recursive_types.go
```diff
@@ -0,0 +1,188 @@
...
+	visitor := astmodel.TypeVisitorBuilder{
+		VisitObjectType: astmodel.MakeIdentityVisitOfObjectTypeWithPerPropertyContext(result.UnrollObjectTypeProperty),
+		VisitTypeName:   result.UnrollRecursiveReference,
+	}.Build()
+	result.visitor = &visitor
+
+	return result
+}
+
+func (s *simpleRecursiveTypeFixer) UnrollObjectTypeProperty(ot *astmodel.ObjectType, prop *astmodel.PropertyDefinition, ctx interface{}) (interface{}, error) {
```
babbageclunk, line 107: Shouldn't this be private? It feels like part of the internal implementation. (Maybe you made it public so it could be called from `astmodel.MakeIdentityVisit...`? I'm _nearly_ sure you don't need to.)

hack/generator/pkg/codegen/pipeline/unroll_recursive_types.go
```diff
@@ -0,0 +1,188 @@
...
+	typedCtx := ctx.(simpleRecursiveTypeFixerContext)
+	name := typedCtx.name
+
+	if !isPropertyMatchingTypeName(name, prop) {
+		return ctx, nil
+	}
+
+	// We know we have at least one property we need to remove. There might be more than 1 direct reference, do a
+	// quick scan for others...
+	toRemove := map[astmodel.PropertyName]struct{}{
```
babbageclunk, line 117: Why do this - wouldn't each self-referential property remove itself? Ahhh - it's because then each property would generate a different unrolled type with itself removed but the other self-references still there, is that right? Could you expand this comment to that effect?

Feels a little bad that this process is repeated for each property - I guess there's no sharing the context between the different properties... 

_sneaky idea that's definitely bad_
You could mutate the original ctx with the self-ref property names and then return a derived context?

hack/generator/pkg/codegen/pipeline/unroll_recursive_types.go
```diff
@@ -0,0 +1,188 @@
...
+
+	err := s.newTypes.AddAllowDuplicates(unrolledDef)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error adding unrolled type %q", name)
+	}
+
+	return typedCtx.WithUnrolledName(unrolledName), nil
+}
+
+func (s *simpleRecursiveTypeFixer) UnrollRecursiveReference(this *astmodel.TypeVisitor, it astmodel.TypeName, ctx interface{}) (astmodel.Type, error) {
```
babbageclunk, line 145: This should be private too.

hack/generator/pkg/codegen/pipeline/unroll_recursive_types.go
```diff
@@ -0,0 +1,188 @@
...
+	if !typedCtx.mustUnroll {
+		return astmodel.IdentityVisitOfTypeName(this, it, ctx)
+	}
+
+	klog.Warningf("%q references itself, removing the self-reference by unrolling to %q", typedCtx.name, typedCtx.unrolledName)
+	return astmodel.IdentityVisitOfTypeName(this, typedCtx.unrolledName, typedCtx)
+}
+
+// asTypeName checks if the type is an astmodel.TypeName or can be unwrapped into an astmodel.TypeName.
+// This differs from astmodel.AsTypeName() in that it unwraps Arrays and Maps too. Use with caution.
```
babbageclunk, line 157: Oh cunning - so this handles `SomeError` with `[]SomeError` or `map[string]SomeError` fields as well? That's spicy!

babbageclunk: Although this is one of those cases where I think it should be in its own package so another pipeline stage doesn't accidentally use it.

hack/generator/pkg/codegen/testdata/RecursiveTypes/Unrolls_recursive_type_list_property.golden
```diff
@@ -0,0 +1,19 @@
...
+package v1alpha1api20200101
+
+//Generated from: https://test.test/schemas/2020-01-01/test.json
+type Test struct {
+	Error *Error `json:"error,omitempty"`
+}
+
+//Generated from: https://test.test/schemas/2020-01-01/test.json#/definitions/Error
+type Error struct {
+	Errors []Error_Unrolled `json:"errors,omitempty"`
```
babbageclunk, line 13: So cool

babbageclunk (approved):
This looks great! Just minor comments.

hack/generator/pkg/astmodel/type_visitor.go
```diff
@@ -128,23 +128,29 @@ func IdentityVisitOfPrimitiveType(_ *TypeVisitor, it *PrimitiveType, _ interface
...
-	makeCtx func(prop *PropertyDefinition, ctx interface{}) interface{}) func(this *TypeVisitor, it *ObjectType, ctx interface{}) (Type, error) {
+	makeCtx func(ot *ObjectType, prop *PropertyDefinition, ctx interface{}) (interface{}, error)) func(this *TypeVisitor, it *ObjectType, ctx interface{}) (Type, error) {
```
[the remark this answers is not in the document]
matthchr: done

hack/generator/pkg/astmodel/type_visitor.go
matthchr: > Basically the name should say what the function does (makes an identity object type visit func), not how it does it (by making a new context for every property).

I sometimes struggle with this =)

Originally I was thinking that I really wanted to highlight the difference between `MakeIdentityVisitOfObjectTypeWithPerPropertyContext` and `IdentityVisitOfObjectType`, but you're right that in practice I'm not sure that distinction helps all that much.

I've done this.

hack/generator/pkg/codegen/pipeline/unroll_recursive_types.go
matthchr: Fixed

hack/generator/pkg/codegen/pipeline/unroll_recursive_types.go
matthchr: No you're right, I was being careless here. Good catch

hack/generator/pkg/codegen/pipeline/unroll_recursive_types.go
matthchr: Fixed

hack/generator/pkg/codegen/pipeline/unroll_recursive_types.go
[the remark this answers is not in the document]
matthchr: This worked out quite well. If I run into a type without "error" in its name, I specifically raise an error. Turns out there are none currently (all 43 types that do this have Error in their name), which is exactly what we were going for.

hack/generator/pkg/codegen/pipeline/unroll_recursive_types.go
matthchr: Surprising how much cleaner that makes the call. Great suggestion.

hack/generator/pkg/codegen/pipeline/unroll_recursive_types.go
matthchr: I've moved the whole `SimpleRecursiveTypeFixer` to its own package (a subpackage of pipeline)

hack/generator/pkg/codegen/pipeline/unroll_recursive_types.go
matthchr: I've updated the comment to clarify here.

The mutating ctx feels worse than O(N^2) on properties - it's not ideal but any fix that I can come up with for it feels even worse?

PR #1754, opened 2021-08-26
Partial update of Golden tests

theunrepentantgeek:

Building on the changes in PR #1737, updating some of our pipeline golden tests to use the new format.

Creating a partial PR to keep the size down since I uncovered some reentrancy bugs in a couple of pipeline stages. #yaytests

Includes renaming of the updated tests to `TestGolden` in alignment with our desired naming convention (so that we can create a Task to update them all as a batch.)

Also: 

* Updates a couple of packages
* Increases timeout for lint to run in CI

**Prerequisites** 

* [x] Merge PR #1737 
* [x] Rebase on `master`

The change introduces diff notation into the PR, which is also used by GitHub to highlight changes, so care is needed when reading the updates to Golden files.

gif

**If applicable**:
- [x] this PR contains tests


hack/generator/pkg/functions/property_assignment_function.go
```diff
@@ -76,17 +74,17 @@ func NewPropertyAssignmentFunction(
...
 		conversions:        make(map[string]StoragePropertyConversion),
-		knownLocals:        astmodel.NewKnownLocalsSet(idFactory),
 		receiverName:       idFactory.CreateIdentifier(receiver.Name().Name(), astmodel.NotExported),
 		parameterName:      direction.SelectString("source", "destination"),
 	}
 
 	// Flag receiver and parameter names as used
-	result.knownLocals.Add(result.receiverName)
-	result.knownLocals.Add(result.parameterName)
+	knownLocals := astmodel.NewKnownLocalsSet(idFactory)
```
Porges, line 82: perhaps as convenience `NewKnownLocalsSet` should also take `...string`s

Porges (approved):

hack/generator/pkg/codegen/pipeline/create_storage_types_test.go
```diff
@@ -14,7 +14,7 @@ import (
 	"github.com/Azure/azure-service-operator/hack/generator/pkg/test"
 )
 
-func TestCreateStorageTypes(t *testing.T) {
+func TestGolden_CreateStorageTypes(t *testing.T) {
```
Porges, line 17: `Taskfile.yml` needs to be updated for this to work. Do it in this PR?

hack/generator/pkg/functions/property_assignment_function.go
theunrepentantgeek: Would only have an effect in one place - but still a good idea. :+1:

PR #1759, opened 2021-08-27
Controller logging improvements

matthchr:
We weren't consistent with what we were logging at what level. Some log level 0 messages were getting emitted every time through the reconcile loop, others were much less verbose. When writing a log statement, it was hard to determine where in the 0 to 10 spectrum to put your log without crawling all of the other log statements around you to rebuild your mental model of what levels we were using.

There is some relatively standard guidance [from Kubernetes](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md), but the problem is remembering it, especially given that they have 6 levels discussed there many of which are just described as "More details" which of course begs the question: "When do I choose "details" and when do I choose "more details""

The solution I came up with was to:
- Name some log levels and use the names. _Hopefully_ the names are sorta descriptive.
- Provide documentation/examples that are specific to our use case (authoring a controller) on each named level for the kinds of logs that should go into that level.

I'm open to changing the log level names or even doing something else entirely, as long as it solves the fundamental problem of  helping us keep our logs consistent.

I also set the log level in the container to 1 (I believe the default is 0) so that if we have issues escalated to us we have a bit more.

### DIFF
--- a/hack/generated/config/manager/manager.yaml
+++ b/hack/generated/config/manager/manager.yaml
@@ -25,6 +25,7 @@ spec:
       containers:
       - args:
         - --enable-leader-election
+        - --v=2
         image: controller:latest
         name: manager
         resources:
--- a/hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
+++ b/hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
@@ -28,11 +28,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
-	"github.com/Azure/azure-service-operator/hack/generated/pkg/genruntime/conditions"
-	"github.com/Azure/azure-service-operator/hack/generated/pkg/ownerutil"
-
 	"github.com/Azure/azure-service-operator/hack/generated/pkg/armclient"
 	"github.com/Azure/azure-service-operator/hack/generated/pkg/genruntime"
+	"github.com/Azure/azure-service-operator/hack/generated/pkg/genruntime/conditions"
+	. "github.com/Azure/azure-service-operator/hack/generated/pkg/logging"
+	"github.com/Azure/azure-service-operator/hack/generated/pkg/ownerutil"
 	"github.com/Azure/azure-service-operator/hack/generated/pkg/reflecthelpers"
 	"github.com/Azure/azure-service-operator/hack/generated/pkg/util/kubeclient"
 )
@@ -129,7 +129,7 @@ func (r *AzureDeploymentReconciler) CreateOrUpdate(ctx context.Context) (ctrl.Re
 }
 
 func (r *AzureDeploymentReconciler) Delete(ctx context.Context) (ctrl.Result, error) {
-	r.logObj("deleting resource")
+	r.logObj("reconciling resource")
 
 	action, actionFunc, err := r.DetermineDeleteAction()
 	if err != nil {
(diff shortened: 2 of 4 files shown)


hack/generated/config/manager/manager.yaml
```diff
@@ -25,6 +25,7 @@ spec:
       containers:
       - args:
         - --enable-leader-election
+        - --v=1
```
matthchr, line 28: Should we set this higher?

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
```diff
@@ -369,8 +369,6 @@ func (r *AzureDeploymentReconciler) DetermineCreateOrUpdateAction() (CreateOrUpd
 	}
 
 	if hasOngoingDeployment {
-		r.log.V(3).Info("Have existing deployment ID, will monitor it", "deploymentID", ongoingDeploymentID)
```
theunrepentantgeek, line 372: If something stalls and never makes progress, this log line might be the key to spotting the situation. 
Suggest restoring this, but at Verbose level.

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
```diff
@@ -883,7 +882,7 @@ func (r *AzureDeploymentReconciler) isOwnerReady(ctx context.Context) (bool, err
 	if err != nil {
 		var typedErr *genruntime.ReferenceNotFound
 		if errors.As(err, &typedErr) {
-			r.log.V(4).Info("Owner does not yet exist", "NamespacedName", typedErr.NamespacedName)
+			r.log.V(Verbose).Info("Owner does not yet exist", "NamespacedName", typedErr.NamespacedName)
```
theunrepentantgeek, line 885: Might this be more important? An owner that never gets created is a real possibility and a blocker we'd want to be able to troubleshoot. Unless the information is shown elsewhere?
```suggestion
			r.log.V(Info).Info("Owner does not yet exist", "NamespacedName", typedErr.NamespacedName)
```

hack/generated/config/manager/manager.yaml
theunrepentantgeek: While we're in Alpha, yes - we kind of expect issues. Change it back to 1 when we hit Beta, maybe?

theunrepentantgeek (approved):
This is a really good idea.

I have a couple nitpicks on levels for some messages, but otherwise looks great.

gif

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
matthchr: I removed this because we actually already logged this above on line 355 pretty much, so this line was duplicate. Acutally the above line is better because it also includes the current `Ready` condition and `hasChanged`.

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
matthchr: It'll be shown on the actual resource state too, but you're right I can promote this to info.

I even said for info:
> an Info log probably shouldn't be emitted every time through the reconcile loop, at least in the happy path.

and this would log every loop _but_ it's not the happy path so at least as per that comment it's ok. Will fix.

hack/generated/config/manager/manager.yaml
matthchr: I've put this up to 2 for now. That means we're not logging full API requests/responses (which we get at 3) but we're logging everything else.

Avoiding full requests/responses (except in our test logs for now) seems prudent given they could theoretically contain secrets at the moment since we don't do any pruning AND we don't put secrets into their own Kubernetes secrets object. We'll need to do something about that - I've got #1762 filed tracking that.

hack/generated/pkg/reconcilers/azure_deployment_reconciler.go
matthchr: Fixed this

PR #1761, opened 2021-08-28
Add visualization of the repository

theunrepentantgeek:

A [recent GitHub blog post](https://next.github.com/projects/repo-visualization) shared an experimental way to visualize the content of a repository, as a way for contributors to see what the repo contains. I thought it would be worth trying out.

Look at the rendered version of README.md to see the visualization. There may be a better place to put this.

gif

**If applicable**:


Porges:
[link to image](https://raw.githubusercontent.com/Azure/azure-service-operator/0b1089b9fb3c6cf01f6daf1f9490552f9be51726/diagram.svg)

Unfortunately the generated SVG is not hyperlinked. It looks like eventually the `apis` dir will swamp everything else!

theunrepentantgeek:
It would be interesting if the tool was extended to allow specifying subdirectories, and maybe exclusions too ... would be neat to visualise just the generator, or just the controller sans `apis`.

README.md
```diff
@@ -53,6 +53,21 @@ Please search open issues [here](https://github.com/Azure/azure-service-operator
 
 For more information, see [SUPPORT.md](SUPPORT.md).
 
+## Repository Visualizations
```
matthchr, line 56: Wondering if it makes more sense for these links to be in the individual project README's rather than the root one? My thinking with this root README was just as a landing page that answered the "What is the ASO project, what are the versions, which should I choose" and everything else was in the v1 or v2 README (so that people who want v1 didn't have to see v2's stuff and vice versa).

.github/workflows/visualize-repo.yml
```diff
@@ -0,0 +1,33 @@
+name: Render Visualization of Azure Service Operator v2
```
matthchr, line 1: This looks like it updates the visualization for v1 and v2? Maybe the title needs to be fixed?

README.md
theunrepentantgeek: I've moved the diagrams somewhat, though not necessarily into README files - as those are often about using the operator, not modifying it. The new locations are better than here, but still not great.

We need to add specific documentation for people who want to do development vs people wanting to run the operator.

.github/workflows/visualize-repo.yml
theunrepentantgeek: Fixed.

matthchr (approved):

README.md
matthchr: Yeah good point, if the intended audience is developers (and not users) then CONTRIBUTING or the development.md for v1 is I think a great place to put them. Approved!

PR #1765, opened 2021-08-30
Fix conversion webhooks in operator bundle

### LINKED ISSUE #1718: Bug: Unable to deploy resources on new (1.27207) ASO deployment with error: context deadline exceeded.
**Describe the bug**
I've installed Azure Service Operator (1.27207) using Operator Lifecycle Manager. It's installed on an AKS cluster. This cluster has the same configuration as I've used successfully with previous versions of ASO, except that it's using a newer version of Kubernetes than previously (1.21.2) and it's using Azure networking (due to the requirement from aad pod identity). I'm trying to deploy a MySQLServer resource (and others related to it), but I'm getting `context deadline exceeded`. This error is persisting through manager controller restarts. I don't think it's related to aad-pod-identity, as I'm successfully using that with Azure KeyVault CSI elsewhere in the cluster.

```
2021-08-13T18:30:24.316Z        INFO    controllers.MySQLServer exiting reconciliation  {"Info Type": "status", "Component": "MySQLServer", "Instance": "test/my-test-server"}
2021-08-13T18:30:24.316Z        ERROR   controller-runtime.manager.controller.mysqlserver       Reconciler error        {"reconciler group": "azure.microsoft.com", "reconciler kind": "MySQLServer", "name": "my-test-server", "namespace": "test", "error": "1 error occurred:\n\t* context deadline exceeded\n\n"}
```

**To Reproduce**
1. Deploy ASO to a cluster through OLM using Managed Identity.
2. Try to deploy a MySQLServer resource.

**Expected behavior**
The resource is deployed successfully, and/or status is updated.

babbageclunk:
The previously released operator bundle wasn't setting up conversion webhooks properly, so any custom resource with multiple versions would encounter problems reconciling. This updates the bundle generation to use the latest release of operator-sdk, and corrects the webhook config on the multi-version CRDs so that they refer to the webhook service OLM creates, since that's the one that it sets up certificates for.

The azureoperator-webhook-service has been removed. 

Since OLM manages the certificate setup for the webhooks, I've also removed the cert-manager annotations from the CRDs, and removed the cert-manager setup from the installation instructions.

Closes #1718 

gif

**If applicable**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -399,6 +399,9 @@ generate-operator-bundle: manifests
 	kustomize build config/operator-bundle | operator-sdk generate bundle --version $(LATEST_TAG) --channels stable --default-channel stable --overwrite --kustomize-dir config/operator-bundle
 	# Building the docker bundle requires a tests/scorecard directory.
 	mkdir -p bundle/tests/scorecard
+	# Remove the webhook service - OLM will create one when installing
+	# the bundle.
+	rm bundle/manifests/azureoperator-webhook-service_v1_service.yaml
 	# Inject the container reference into the bundle.
 	scripts/inject-container-reference.sh "$(PUBLIC_REPO):$(LATEST_TAG)"
 	# Update webhooks to use operator namespace and remove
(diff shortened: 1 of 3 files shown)


matthchr (approved):
LGTM - the bit on `yq`'ing the files is icky but I don't have a better way to do it than that. Probably less icky once we have tests that can actually confirm it did the right thing too.

PR #1766, opened 2021-08-31
Enable structured logs

jonnylangefeld:
This changes enables structured json logs and also writes to `stdout` rather than the default `stderr`.

Closes #[issue number]

Many monitoring systems understand structured json logs better than unstructured custom logs

gif

**If applicable**:
- [ ] this PR contains tests


matthchr:
Thanks for the PR @jonnylangefeld, but I think you want the ASO v2 main which is at https://github.com/Azure/azure-service-operator/blob/master/hack/generated/main.go#L41

jonnylangefeld:
Oh, didn't see there was a second main file. Hmm I see you switched from `zap` to `klogr`. Do you know the reason for that?

matthchr:
Yeah we're in the middle of reorganizing the repo to make it more understandable, but basically the top level is ASO v1 (handcrafted) and hack (right now) is V2 (autogenerated).

I don't have the history of zap vs klog but I can find out.

matthchr:
So it sounds like the history of zap vs klog is just that we're using klog as that's what upstream Kubernetes uses and some other large operator projects (Cluster API) - so not really that zap didn't support something we needed or anything like that.

I agree that the option of JSON formatting seems compelling. From what I can tell you can't do that directly with klog, Kubernetes own structured JSON format is using zap inside klog (which we could also do), or we could just move to zapr.

Is your objective here to:
1. Move logging to stdout
2. Move logs to JSON format?

theunrepentantgeek:
We agree that structured logging is a good idea. We'll merge this PR to update ASO v1, and I've logged Issue #1779 to make sure we don't lose track of doing the same for ASO v2.

theunrepentantgeek (approved):

theunrepentantgeek:
Thanks @jonnylangefeld.

PR #1773, opened 2021-09-02
Add GetStatus() and GetSpec() to all resources

theunrepentantgeek:

Adds the methods `GetSpec()` and `GetStatus()` to `KubernetesResource` ready for consumption by the generic controller, and generates implementations for all resources.

Also includes 

* Manual implementation for ResourceGroup
* Introduction of new package `interfaces` to further untangle `astmodel`

gif


hack/generated/apis/microsoft.resources/v1alpha1api20200601/resourcegroup_types.go
```diff
@@ -181,6 +240,42 @@ func (spec *ResourceGroupSpec) PopulateFromARM(owner genruntime.KnownResourceRef
 // SetAzureName sets the Azure name of the resource
 func (spec *ResourceGroupSpec) SetAzureName(azureName string) { spec.AzureName = azureName }
 
+func (spec *ResourceGroupSpec) ConvertSpecTo(destination genruntime.ConvertibleSpec) error {
```
babbageclunk, line 243: These are more like `AssignSpecTo` and `AssignSpecFrom` really, aren't they? In the non-handwritten versions of these methods, are `source` and `destination` guaranteed to be the same type as the receiver?

hack/generator/pkg/astmodel/object_function.go
```diff
@@ -0,0 +1,63 @@
...
+
+package astmodel
+
+import (
+	"github.com/dave/dst"
+)
+
+// objectFunction is a simple helper that implements the Function interface. It is intended for use for functions
+// that only need information about the object they are operating on
+// TODO: Reduce use of this in favour of the alternative in the functions package
```
babbageclunk, line 14: Ah, I was just coming back to ask about the difference between this one and the one in functions.

babbageclunk (approved):

hack/generated/apis/microsoft.resources/v1alpha1api20200601/resourcegroup_types.go
```diff
@@ -181,6 +240,42 @@ func (spec *ResourceGroupSpec) PopulateFromARM(owner genruntime.KnownResourceRef
...
+	dst, ok := destination.(*ResourceGroupSpec)
+	if !ok {
+		return errors.Errorf("cannot convert ResourceGroupSpec to %T", destination)
```
matthchr, line 246: Isn't this comment backwards? You're trying to confirm %T to ResourceGroupSpec, right?

hack/generated/apis/microsoft.batch/v1alpha1api20210101/batch_account_types_gen.go
```diff
@@ -76,6 +76,16 @@ func (batchAccount *BatchAccount) AzureName() string {
 	return batchAccount.Spec.AzureName
 }
 
+// GetSpec returns the specification of this resource
+func (batchAccount *BatchAccount) GetSpec() genruntime.ConvertibleSpec {
+	return &batchAccount.Spec
+}
+
+// GetStatus returns the current status of this resource
```
matthchr, line 84: minor: To match the comment above:
```suggestion
// GetStatus returns the status of this resource
```

Or change the above one to say "current"

hack/generated/apis/microsoft.resources/v1alpha1api20200601/resourcegroup_types.go
```diff
@@ -127,6 +137,54 @@ func (status *ResourceGroupStatus) PopulateFromARM(owner genruntime.KnownResourc
...
+		dst.Conditions = append(dst.Conditions, c.Copy())
+	}
+
+	return nil
+}
+
+func (status *ResourceGroupStatus) ConvertStatusFrom(source genruntime.ConvertibleStatus) error {
+	src, ok := source.(*ResourceGroupStatus)
+	if !ok {
+		return errors.Errorf("cannot convert ResourceGroupStatus from %T", source)
```
matthchr, line 167: Same comment as below about this log seeming backwards to me?

Should it instead read: `cannot convert from %T to ResourceGroupStatus`?

hack/generated/apis/microsoft.resources/v1alpha1api20200601/resourcegroup_types.go
```diff
@@ -70,6 +71,14 @@ func (rg *ResourceGroup) Owner() *genruntime.ResourceReference {
 	return nil
 }
 
+func (rg *ResourceGroup) GetSpec() genruntime.ConvertibleSpec {
```
matthchr, line 74: minor: Missing comments?

matthchr: below too

hack/generated/apis/microsoft.resources/v1alpha1api20200601/resourcegroup_types.go
```diff
@@ -127,6 +137,54 @@ func (status *ResourceGroupStatus) PopulateFromARM(owner genruntime.KnownResourc
 	return nil
 }
 
+func (status *ResourceGroupStatus) ConvertStatusTo(destination genruntime.ConvertibleStatus) error {
```
matthchr, line 140: minor: Missing comments here and below too?

hack/generated/pkg/genruntime/base_types.go
```diff
@@ -57,6 +57,12 @@ type KubernetesResource interface {
 
 	// TODO: I think we need this
 	// SetStatus(status interface{})
+
+	// GetSpec returns the specification of the resource
+	GetSpec() ConvertibleSpec
+
+	// GetStatus returns the current status of the resource
```
matthchr, line 64: minor: Same comment as above, one of these says "current X" and one says "X".

I think just `GetStatus returns the status of the resource` is good enough?

hack/generator/pkg/astmodel/conditioner_interface.go
```diff
@@ -49,7 +49,7 @@ func NewConditionerInterfaceImpl(
 
 // getConditionsFunction returns a function declaration containing the implementation of the GetConditions() function.
 //
-// func (r *<receiver>)GetConditions() genruntime.Conditions {
+// func (r *<receiver>) GetConditions() genruntime.Conditions {
```
matthchr, line 52: Good catch

hack/generator/pkg/interfaces/ReadMe.md
```diff
@@ -0,0 +1,3 @@
+# Interfaces Folder
```
matthchr, line 1: minor: The casing of this file should be `README.md` I believe?

hack/generator/pkg/astmodel/object_function.go
matthchr: Why not do that in this PR?

Seeing as structurally they seem identical it feels like a small reference update in a few usage locations? IIRC there are only 3 or 4 places this is used?

hack/generated/apis/microsoft.resources/v1alpha1api20200601/resourcegroup_types.go
theunrepentantgeek: In the generated versions, we have both `AssignProperties<To|From>()` and `ConvertSpec<To|From>()`. Given that we have exactly one version of ResourceGroup and that's not likely to change, I combined the functionality together here. 

Do you think I should maintain the same pattern and split them out?

To answer your specific question, the `AssignProperties<To|From>()` methods have strongly typed parameters of the exact type, but of the next version of the resource. So the pattern isn't exactly the same.

hack/generator/pkg/astmodel/object_function.go
theunrepentantgeek: It's far more than a few reference updates - the code consuming `astmodel.objectFunction` needs to be relocated out of `astmodel` before it can be modified to use `functions.ObjectFunction`. From discussion with @matthchr, this is possible but there may be fishhooks, so we agreed not to do it in this PR. I've created #1775 to track the required changes.

hack/generated/apis/microsoft.resources/v1alpha1api20200601/resourcegroup_types.go
theunrepentantgeek: No, I think the error is right - and matches the implementation.

We're converting from the current instance (`spec`) to the provided instance (`destination`), copying properties across.

hack/generated/apis/microsoft.batch/v1alpha1api20210101/batch_account_types_gen.go
theunrepentantgeek: Updated the comments to be consistent.

hack/generated/apis/microsoft.resources/v1alpha1api20200601/resourcegroup_types.go
theunrepentantgeek: Again, the error message matches the implementation.

We're converting from the provided instance (`source`) to the current one (`status`).

hack/generated/apis/microsoft.resources/v1alpha1api20200601/resourcegroup_types.go
theunrepentantgeek: Added.

hack/generated/apis/microsoft.resources/v1alpha1api20200601/resourcegroup_types.go
theunrepentantgeek: Added.

hack/generated/apis/microsoft.resources/v1alpha1api20200601/resourcegroup_types.go
babbageclunk: Ok, that makes sense - the part I was missing is that there's only one version of ResourceGroup and so this is the only reasonable way to implement `ConvertSpecTo` et al. Maybe add a comment to that effect?

hack/generated/apis/microsoft.resources/v1alpha1api20200601/resourcegroup_types.go
matthchr: I see what you are saying now. What threw me was that _this particular cast_ (which is what I was expecting the error message to be about) is from `destination` to `ResourceGroupSpec`, so I was expecting a comment that read something like: `expected destination to be ResourceGroupSpec, was: %T, destination` and what you're saying is more about what the overall method is trying to do.

You can feel free to ignore this as I see what you're going for now.

hack/generated/apis/microsoft.resources/v1alpha1api20200601/resourcegroup_types.go
matthchr: Did these get added?

matthchr (approved):

PR #1782, opened 2021-09-05
Code Gardening

theunrepentantgeek:

Minor code gardening, addressing three issues:

* Naming of types in `std_references` was inconsistent
* Factory method `NewGomegaWithT()` is deprecated in favour of `NewWithT()`
* We want to standardise on `TestGolden_` as a prefix for golden tests, to enable automation

This PR touches a lot of files, but the changes are individually very tedious and mechanical.

gif

**If applicable**:
- [x] this PR contains tests


hack/generator/pkg/astmodel/std_references.go
```diff
@@ -54,18 +54,18 @@ var (
 	KubernetesResourceType         = MakeTypeName(GenRuntimeReference, "KubernetesResource")
 	ConvertibleSpecInterfaceType   = MakeTypeName(GenRuntimeReference, "ConvertibleSpec")
 	ConvertibleStatusInterfaceType = MakeTypeName(GenRuntimeReference, "ConvertibleStatus")
-	ResourceReferenceTypeName      = MakeTypeName(GenRuntimeReference, "ResourceReference")
-	KnownResourceReferenceTypeName = MakeTypeName(GenRuntimeReference, "KnownResourceReference")
+	ResourceReferenceType          = MakeTypeName(GenRuntimeReference, "ResourceReference")
```
babbageclunk, line 57: I wrote a medium-length comment here about how this was confusing because the variable name suggested it was a type but it was a pointer to a type (the structure of the type was somewhere else). Then I realised that the type of this variable is `TypeName` so users can tell it's the name and not the thing itself. 
Now I think maybe you could drop the `-Type` suffix at the end as well? Although I guess the suffix here is the kind of thing the name refers to?

babbageclunk (approved):
Looks good - I got into a bit of a philosophical tangle about the renames in `std_references` but I think I've extricated myself now.

hack/generator/pkg/astmodel/std_references.go
theunrepentantgeek: I considered dropping the `Type` but decided not to, because the variable doesn't contain (for example) a `KnownResourceReference` (it's not a reference to any one resource, known or otherwise), it instead contains a reference to the `Type` of all of those references.

Issue #1771, opened 2021-09-01
Add fullyQualifiedUsername to MySQLUser secret


kyschouv:
**Describe the current behavior**
`MySQLServer` secrets include a `fullyQualifiedUsername` value. `MySQLUser` secrets do not, and the value must be created by combining the `username` with the `mySqlServerName`. It would be nice to not have to do that, especially as it can be difficult to compose and apply depending on the usage (e.g. a Helm chart may not be set up to define environment variables for username and server name and then later combine them into a fully qualified user name).

**Describe the improvement**
Add `fullyQualifiedUsername` to the secret generated for a `MySQLUser`.

theunrepentantgeek:
v1 is in maintenance mode, so we won't be making a change there.

However, we need to be aware of this issue for v2, as we need to avoid replicating the problem there.

matthchr:
Somewhat related to #2555

matthchr:
Closing this in favor of the more generic #2555. That issue may not be solved for a little while yet (we need to think about it), so if there's an explicit need for this field or one like it on an ASOv2 resource feel free to open a new issue for that and we may be able to action that specific one.

PR #1784, opened 2021-09-06
MySQL: support granting ALL privileges in a database to a user

### LINKED ISSUE #1770: Bug: MySQLUser v1alpha2 does support granting ALL privileges to a database
**Describe the bug**
`MySQLUser` resources at `v1alpha2` don't support granting `ALL` privileges. On `v1alpha1` this worked fine.

**To Reproduce**
Create a `MySQLServer` and `MySQLDatabase`, then deploy a user with a grant for all permissions on the database:
```
apiVersion: azure.microsoft.com/v1alpha2
kind: MySQLUser
metadata:
  name: myuser
  namespace: myns
spec:
  resourceGroup: myrg
  server: myserver
  username: myuser
  databaseRoles:
    mydb:
    - ALL
```

**Expected behavior**
The user is created and granted all privileges on the specified database.

**Additional context**
The user is created, but not granted any privileges on the database. If you change the privilege to `SELECT`, you can see that it correctly grants the user `SELECT` privilege on the database.

babbageclunk:
This allows specifying `ALL` (case insensitive) in `databaseRoles` for a `MySQLUser` or `MySQLAADUser` - in that case we will pass the `ALL` grant on to the database and won't revoke any privileges. Without special handling for this we could get into a confusing situation where if a user had select and update privileges, and then that user was updated to have the special ALL privilege, the operator would grant ALL and then revoke select and update privileges.

Closes #1770.

**If applicable**:
- [x] this PR contains tests


babbageclunk:
Hmm, I was updating the documentation and I realised that I hadn't tested ALL for global privileges. It doesn't work, it looks like the super user we use to connect to the database doesn't actually have all of the privileges (missing shutdown, file, super and create tablespace). So I'll change the code to avoid that.

babbageclunk:
It seems like the right thing to do is forbid ALL in spec.roles in a webhook.

pkg/helpers/sqlrole.go
```diff
@@ -5,18 +5,31 @@ Licensed under the MIT license.
...
+
 func DiffCurrentAndExpectedSQLRoles(currentRoles map[string]struct{}, expectedRoles map[string]struct{}) SQLRoleDelta {
 	result := SQLRoleDelta{
 		AddedRoles:   make(map[string]struct{}),
 		DeletedRoles: make(map[string]struct{}),
 	}
 
 	for role := range expectedRoles {
+		// Escape hatch - if they ask for ALL then we just grant ALL
+		// and don't delete any.
```
theunrepentantgeek, line 25: Maybe explain why we don't delete things in this comment?

pkg/helpers/sqlrole_test.go
```diff
@@ -10,57 +10,64 @@ import (
...
+	}, {
+		name:                "ALL is case insensitive",
+		currentRoles:        map[string]struct{}{"SELECT": {}, "INSERT": {}},
+		expectedRoles:       map[string]struct{}{"aLl": {}, "UPDATE": {}},
+		expectedRoleCreates: map[string]struct{}{"ALL": {}},
+		expectedRoleDeletes: make(map[string]struct{}),
+	}}
+	// There's no test for handling ALL in current roles because we
+	// don't see that in the database - it gets expanded into all
+	// permissions.
```
theunrepentantgeek, line 64: Ooo! There's a trap for the uninitiated!

theunrepentantgeek (approved):
One suggestion to add more details into a comment; looks good.

gif

docs/v1/services/mysql/mysql.md
```diff
@@ -81,13 +81,21 @@ Here is a [sample YAML](/config/samples/azure_v1alpha1_mysqlserveradministrator.
...
-Here is a [sample YAML](/config/samples/azure_v1alpha1_mysqluser.yaml) for MySQL user. 
+Here is a [sample YAML](/config/samples/azure_v1alpha2_mysqluser.yaml) for MySQL user. 
 
 The `resourceGroup` is the resource group of the MySQL server and MySQL database, provide the MySQL server name in `server` and MySQL database name in `dbName`. 
 
-The operator supports grant specified privileges using the concept of `roles`, and supports assigning one or more privileges from the list:
+The operator supports granting the user global privileges using the `roles` field, and supports assigning one or more privileges from the list:
 
-##### `SELECT, INSERT, UPDATE, DELETE, CREATE, DROP, RELOAD, PROCESS, REFERENCES, INDEX, ALTER, SHOW DATABASES, CREATE TEMPORARY TABLES, LOCK TABLES, EXECUTE, REPLICATION SLAVE, REPLICATION CLIENT, CREATE VIEW, SHOW VIEW, CREATE ROUTINE, ALTER ROUTINE, CREATE USER, EVENT, TRIGGER`.
+##### `CREATE TABLESPACE, CREATE USER, FILE, PROCESS, RELOAD, REPLICATION CLIENT, REPLICATION SLAVE, SHOW DATABASES, SHUTDOWN, SUPER`
```
theunrepentantgeek, line 90: Why do this as a heading?

```suggestion
* `CREATE TABLESPACE`
* `CREATE USER`
* `FILE`
* `PROCESS`
* `RELOAD`
* `REPLICATION CLIENT`
* `REPLICATION SLAVE`
* `SHOW DATABASES`
* `SHUTDOWN`
* `SUPER``
```

Same change for below.

theunrepentantgeek (approved):
One minor editorial comment. Looks good.

gif

docs/v1/services/mysql/mysql.md
babbageclunk: Yeah, fair point - I was just keeping the same format as previously without really thinking about it but a list is better.

kyschouv:
Thanks for the quick response here! When might we see this make it into a new release? The issue isn't fixed from my end until I can use it 😣.

babbageclunk:
Oh that's a good point - I'll unclose the issue until this fix is released, sorry about that.  I'll do #1771 as well before the v1.9 release. There are some other things I'm working on, but it should be mid next week.

Issue #1715, opened 2021-08-12
Autogenerate CRD docs


matthchr:
We should look into using a tool like https://github.com/ahmetb/gen-crd-api-reference-docs to generate API documentation for the CRDs.

matthchr:
There's also this: https://github.com/crdsdev/doc

matthchr:
Also see: https://github.com/kubernetes-sigs/controller-tools/issues/240

matthchr:
I tried to use gen-crd-api-reference-docs and wasn't above to get it to actually generate anything. It does seem to work for some projects (see [cert-manager](https://github.com/cert-manager/website/blob/172648ac48c09c2004b5ba3049346c84b5462722/scripts/gendocs/generate#L64) and [knative](https://github.com/knative/serving/blob/4ac36cef7b4603d54f5d4630a66a26bf75dfce8b/hack/update-reference-docs.sh), but it doesn't understand go modules I don't think, so there's some hoops you have to jump through to get it to actually output anything.

After fighting with it some I _was_ able to get it to run, it just didn't find any modules:
```go
~/work/github/azure-service-operator/hack/generated (master) $ /home/matthchr/go/bin/gen-crd-api-reference-docs -config ../../example-config.json -api-dir "./apis/microsoft.batch/v1alpha1api20210101" -template-dir ~/go/pkg/mod/github.com/ahmetb/gen-crd-api-reference-docs@v0.3.0/template/ -out-file docs.html -v=4
```

matthchr:
We could use the same tool that Golang itself uses and run it on bare types (generated from a custom pipeline stage) and then delete everything but the generated HTML...

PR #1791, opened 2021-09-08
Create pipeline stage to generate list of supported resources

theunrepentantgeek:

Generates a report that lists all the resources (and versions) included in the generation run.

gif

**If applicable**:
- [x] this PR contains tests


hack/generator/pkg/codegen/pipeline/report_resource_versions.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/pkg/errors"
+
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/astmodel"
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/config"
+)
+
+// ReportResourceVersionsStageID is the unique identifier of this stage
+const ReportResourceVersionsStageID = "reportResourceVersions"
+
+// ReportResourceVersions creates a pipeline stage that removes any wrapper types prior to actual code generation
```
matthchr, line 26: This comment doesn't seem right?

hack/generated/apis/resources.md
```diff
@@ -0,0 +1,58 @@
...
+
+### microsoft.managedidentity
+
+v1alpha1api20181130
+
+- UserAssignedIdentity
+
+### microsoft.network
+
+v1alpha1api20201101
```
matthchr, line 31: Consider bolding this, for emphasis?

hack/generator/pkg/codegen/pipeline/report_resource_versions.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func NewResourceVersionsReport() *ResourceVersionsReport {
+	return &ResourceVersionsReport{
+		lists: make(map[astmodel.PackageReference][]string),
+	}
+}
+
+// Summarize collates a list of all resources, grouped by package
+func (r *ResourceVersionsReport) Summarize(types astmodel.Types) {
```
matthchr, line 52: This is a style thing, but it feels sorta weird to me to first construct an empty one of these, and then call `Summarize` on it causing its immediate mutation. You could do all of this in the `NewResourceVersionsReport` call and then your type is immutable, which feels like a win. This fits especially nicely here because this iteration can't `error` anyway (which would be a reason to _not_ do it in the `New`).

hack/generator/pkg/codegen/pipeline/report_resource_versions.go
```diff
@@ -0,0 +1,121 @@
...
+	resources := astmodel.FindResourceTypes(types)
+	for _, rsrc := range resources {
+		name := rsrc.Name()
+		pkg := name.PackageReference
+		r.lists[pkg] = append(r.lists[pkg], name.Name())
+	}
+}
+
+func (r *ResourceVersionsReport) WriteTo(outputPath string) error {
+
```
matthchr, line 62: very minor: You have a newline here but not on the other methods

matthchr:
This is related to #1715

If we had #1715 would we want to keep this? In some ways it gives a nice summary of the resources in each API folder, which is neat. In other ways, it would be covered by #1715 from a "documentation for customers" perspective, because at that point we'd be pointing them at the generated docs and not here.

If you think this should be removed once we have #1715, possibly put a comment someplace indicating that.

matthchr (approved):

babbageclunk (approved):

hack/generator/pkg/codegen/pipeline/report_resource_versions.go
theunrepentantgeek: Fixed.

hack/generated/apis/resources.md
theunrepentantgeek: It's the only thing on the line, and it's followed by a bullet list, so I don't think it needs further emphasis.

hack/generator/pkg/codegen/pipeline/report_resource_versions.go
theunrepentantgeek: Good idea. Changed.

hack/generator/pkg/codegen/pipeline/report_resource_versions.go
theunrepentantgeek: Fixed.

theunrepentantgeek:
I don't think this would necessarily be supplanted by #1715; it would depend on what information is included in the output.

PR #1792, opened 2021-09-08
Update to go 1.17

Porges:
Update the (v2) devcontainer image and hence controller build to [go 1.17](https://golang.org/doc/go1.17).


hack/generator/go.mod
```diff
@@ -35,4 +35,17 @@ require (
...
+	github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578 // indirect
+	github.com/go-logr/logr v0.1.0 // indirect
+	github.com/go-openapi/swag v0.19.5 // indirect
+	github.com/inconshreveable/mousetrap v1.0.0 // indirect
+	github.com/pmezard/go-difflib v1.0.0 // indirect
+	github.com/sergi/go-diff v1.0.0 // indirect
+	golang.org/x/tools v0.0.0-20200509030707-2212a7e161a5 // indirect
+	golang.org/x/xerrors v0.0.0-20191204190536-9bdfabe68543 // indirect
+	gopkg.in/yaml.v2 v2.3.0 // indirect
+)
```
theunrepentantgeek, line 49: What's with all these new imports?

theunrepentantgeek (changes requested):
Blocking because something doesn't feel quite right - why does an upgrade of the Go compiler require now importing so many more libraries that we've not seen before.

A bunch of them have names as thoug they're XServer UX related, which is really odd.

Porges:
> Blocking because something doesn't feel quite right - why does an upgrade of the Go compiler require now importing so many more libraries that we've not seen before.

This: https://golang.org/doc/go1.17#go-command

hack/generator/go.mod
Porges: https://github.com/Azure/azure-service-operator/pull/1792#issuecomment-915566202

Porges:
> A bunch of them have names as though they're XServer UX related, which is really odd.

These `/x/*` ones are _experimental_ stdlib packages, I think.

theunrepentantgeek (approved):
Go mod changes in 1.17 are weird. #approved

PR #1804, opened 2021-09-14
Fix how we pass timeouts to ResourceGroup creating

matthchr:


hack/generated/pkg/testcommon/kube_per_test_context.go
```diff
@@ -191,26 +193,16 @@ func (tc KubePerTestContext) CreateTestResourceGroup(rg *resources.ResourceGroup
...
-		// that even though Kuberentes accepted our request to delete the resource, the
+		// that even though Kubernetes accepted our request to delete the resource, the
 		// controller running in envtest never got a chance to actually issue a request
 		// to Azure before the test is torn down (and envtest stops). We can't easily
 		// wait for just "Deleting" as that causes issues with determinism as the controller
 		// doesn't stop polling resources in "Deleting" state and so when running recordings
 		// different runs end up polling different numbers of times. Ensuring we reach a state
 		// the controller deems terminal (Deleted) resolves this issue.
-		tc.G.Eventually(rg, tc.DefaultTimeout(), tc.PollingInterval()).Should(tc.Match.BeDeleted())
+		tc.G.Eventually(rg, tc.RemainingTime(), tc.PollingInterval()).Should(tc.Match.BeDeleted())
```
Porges, line 203: In my other code I fudge this number a bit to give us “enough time to terminate happily”, reduce it by a second or even a bit more?

matthchr: `RemainingTime()` already includes that 1s fudge, so I don't think we need to do it again here?

Porges: bah, I thought this was the golang method, not ours

Porges (approved):

PR #1810, opened 2021-09-15
Modify return of CreateEmptyARMValue

theunrepentantgeek:

Move the implementation of `CreateEmptyARMValue` into the `functions` package and modify the implementation to use `functions.ObjectFunction`. The return type is now `genruntime.ARMResourceStatus` (instead of `interface{}`), and the value returned is a pointer instead of a value.

The controller is modified to use the new implementation.

Code changes are minor, but flow through a lot of generated code.

Also, `CreateEmptyARMValue` is only actually needed on Status types; the implementations on other objects aren't used by the controller. 

I believe we may be able to remove these implementations and reduce the surface area of the generated code - assuming we don't need the implementations for Spec diff'ing. That is why I didn't remove them as a part of this PR.

gif

**If applicable**:
- [x] this PR contains tests


Porges (approved):

hack/generated/apis/microsoft.resources/v1alpha1api20200601/resourcegroup_types.go
```diff
@@ -215,7 +215,7 @@ type ResourceGroupSpec struct {
 var _ genruntime.ARMTransformer = &ResourceGroupSpec{}
 var _ genruntime.ConvertibleSpec = &ResourceGroupSpec{}
 
-func (spec *ResourceGroupSpec) CreateEmptyARMValue() interface{} {
+func (spec *ResourceGroupSpec) CreateEmptyARMValue() genruntime.ARMResourceStatus {
 	return ResourceGroupSpecARM{}
```
matthchr, line 219: Should this not also return a ptr?

hack/generator/pkg/astmodel/std_references.go
```diff
@@ -58,9 +58,11 @@ var (
 	KnownResourceReferenceType     = MakeTypeName(GenRuntimeReference, "KnownResourceReference")
 	PropertyBagType                = MakeTypeName(GenRuntimeReference, "PropertyBag")
 	ToARMConverterInterfaceType    = MakeTypeName(GenRuntimeReference, "ToARMConverter")
-	ConditionType                  = MakeTypeName(GenRuntimeConditionsReference, "Condition")
-	ConditionsType                 = MakeTypeName(GenRuntimeConditionsReference, "Conditions")
-	ConditionerType                = MakeTypeName(GenRuntimeConditionsReference, "Conditioner")
+	ARMResourceStatusType          = MakeTypeName(GenRuntimeReference, "ARMResourceStatus")
+
```
matthchr, line 62: Why the newline here?

hack/generated/pkg/genruntime/arm_transformer.go
```diff
@@ -15,7 +15,7 @@ type ToARMConverter interface {
 }
 
 type FromARMConverter interface {
-	CreateEmptyARMValue() interface{}
+	CreateEmptyARMValue() ARMResourceStatus
```
matthchr, line 18: This type being an `XStatus` type suggests that this interface won't work for fields in the `Spec` graph, is that right? We'll need that eventually I think for diffing. IIRC `ARMResourceStatus` was basically an empty interface though so possibly we can just rename it to something and allow it to apply everywhere when we get to that point?

Just wondering if you've thought any about that.

hack/generator/pkg/codegen/testdata/ArmResource/Arm_oneof_resource_conversion_on_arm_type_only_azure.golden
```diff
@@ -243,8 +243,8 @@ func (fakeResourceSpec *FakeResource_Spec) ConvertToARM(name string, resolvedRef
 }
 
 // CreateEmptyARMValue returns an empty ARM value suitable for deserializing into
-func (fakeResourceSpec *FakeResource_Spec) CreateEmptyARMValue() interface{} {
-	return FakeResource_SpecARM{}
+func (fakeResourceSpec *FakeResource_Spec) CreateEmptyARMValue() genruntime.ARMResourceStatus {
```
matthchr, line 246: Like here, isn't the fact that the `_Spec` resource is impl-ing `ARMResourceStatus` sorta weird? It's not a `Status` right? related to my comment above about how maybe we need to rename this interface since possibly it's not actually representing "ARM Status's" anymore, and is really representing... any ARM type?

matthchr: Also, isn't there no practical different between `genruntime.ARMResourceStatus` and `interface{}` currently? Since `ARMResourceStatus` has no members? Probably some will get added in the future though (possibly in another one of your PRs?) What members are you adding? These non-status types will need them too obviously, so that they impl the interface

hack/generated/apis/microsoft.resources/v1alpha1api20200601/resourcegroup_types.go
theunrepentantgeek: Yes. The CI tests failed, and I fixed this with a later commit.

hack/generated/pkg/genruntime/arm_transformer.go
theunrepentantgeek: It's on my radar. 😀 

In my WIP branch I've got three different sets of types - one for Statusess (those directly referenced by resources), one for Specs  (ditto), and a generic set for other types. 

I suspected we'd need them on Spec types for diff'ing, thanks for confirming.

Given the three sets noted above, do we need `CreateEmptyARMValue()` on the other types? Looks (from what I've seen so far) as though we only need it as an entry point at the top of a structure.

hack/generator/pkg/codegen/testdata/ArmResource/Arm_oneof_resource_conversion_on_arm_type_only_azure.golden
theunrepentantgeek: Relates back to my WIP branch and my comments above.

PR #1811, opened 2021-09-15
Apply heuristic to update property documentation from swagger

### LINKED ISSUE #1757: Bug: Status types with properties overlapping with ResourceBase (name, type, etc) seem to have nondeterministic comments
I _think_ this only applies to Status types but I'm not 100% sure. An example was:
image

theunrepentantgeek:
Closes #[issue number]

I suspected that property augmentation from swagger was sometimes doing things in the wrong order, so I applied the same comment heuristic as used earlier.

Running `generator gen-types azure-crossplane.yml` 10 times in a row didn't create anything untoward, so maybe this resolves #1757? #maybe

The comments seem to be better though, which will pay off when we start generating documentation for the CRDs.

The actual code change is small, but the updates touch many of the generated files.


hack/generator/pkg/codegen/pipeline/status_augment.go
```diff
@@ -113,6 +113,14 @@ func flattenAugmenter(allTypes astmodel.ReadonlyTypes) augmenter {
 						changed = true
 					}
 
+					if len(swaggerProp.Description()) > len(specProp.Description()) {
```
Porges, line 116: if this is in here we should rename the flattenAugmenter to something like propertyAugmenter and pull out the bits that copy the flatten property and copy the documentation into an easily-extensible set

Porges (changes requested):
requesting changes

babbageclunk (approved):
The new comments are much better - @Porges' point about separating this and the flattening seems sensible though.

theunrepentantgeek:
Abandoning in favour of George's upcoming changes.

PR #1814, opened 2021-09-16
Write CRLF style source files on Windows

theunrepentantgeek:

Modified GoSourceWriter to generate Windows style files (CRLF end of lines) when run on Windows

**If applicable**:
- [x] this PR contains tests


Porges:
Alternately we could change the `gitattributes` so that the appropriate files are always checked out as LF?

matthchr (changes requested):
@theunrepentantgeek - Based on my reading [here](https://github.com/golang/go/issues/16355) I went and tested, and `go fmt` specifically reformats `.go` files to end in `\n`, even if they were previously formatted `\r\n`.

I think that this means that we shouldn't do this, as it will cause issues when we then run gofmt across things (it'll just undo this). It appears based on the issue above and [this](https://forum.golangbridge.org/t/gofmt-canonicalizes-line-endings/3615) that golangs official file format wants `\n` only, even on Windows.

theunrepentantgeek:
As discussed offline, abandoning because `go fmt` is hostile to Windows EoLn.

PR #1816, opened 2021-09-17
Code Gardening - astbuilder

theunrepentantgeek:

Some tidyup in areas that _should_ have been using helper methods from `astbuilder` (they were written recently) and which now _can_ use those helper methods (older code written before the helpers existed).

gif


hack/generator/pkg/astbuilder/builder_test.go
```diff
@@ -19,20 +18,20 @@ func asplode(l, r dst.Expr) dst.Expr {
...
-	g := gomega.NewWithT(t)
+	g := NewWithT(t)
 
 	expr := &dst.BadExpr{}
 	g.Expect(Reduce(asplode, expr)).To(BeIdenticalTo(expr))
 }
 
 func TestJoinOr(t *testing.T) {
-	g := gomega.NewWithT(t)
+	g := NewWithT(t)
```
Porges, line 34: I’d say we should use either `gomega.NewWithT` or `NewGomegaWithT` but not `NewWithT` or `gomega.NewGomegaWithT`

theunrepentantgeek: Can we use `gomega.NewWithT(t)` while still having the import in the current scope? Being able to write the assertions without the `gomega.` prefix is almost essential.

theunrepentantgeek: I've been moving to `NewWithT()` because the older `NewGomegaWithT()` because the older one is deprecated:

``` go
// GomegaWithT is deprecated in favor of gomega.WithT, which does not stutter.
type GomegaWithT = WithT
```

No, it looks like we can't. So, moving to `gomega.NewWithT()` also requires moving to

``` go
g.Expect(Reduce(asplode, expr)).To(gomega.BeIdenticalTo(expr))
```

Which is a bit suboptimal.

matthchr (approved):
LGTM so signing off. @porges did have that one comment about `gomega` though, interested to see how that discussion shakes out

hack/generator/pkg/astbuilder/builder_test.go
babbageclunk: Didn't we end up on using `NewGomegaWithT` in the meeting?

theunrepentantgeek: Yes, we did.

theunrepentantgeek: I've updated `builder_test.go` which is the only test file in this PR to use `NewGomegaWithT()`; I'll update other test files as I touch them (or, if I get a momement, I'll do a bulk cleanup PR.)

PR #1817, opened 2021-09-17
Add ability to generate scope from resource kind and name

matthchr:
- Add Scope() to ResourceHierarchy for Extension resources
- Remove scope in pipeline stage rather than config
  - This allows us to skip removing that property on extension resources. It also resolves a small possible bug where we would
have blindly removed a more deeply scoped "Scope" property. Now we're limited to just top level properties on the spec.

### DIFF
--- a/hack/generated/pkg/genruntime/resource_hierarchy.go
+++ b/hack/generated/pkg/genruntime/resource_hierarchy.go
@@ -56,28 +56,88 @@ func (h ResourceHierarchy) Location() (string, error) {
 }
 
 // FullAzureName returns the full Azure name for use in creating a resource.
-// This name is the full "path" to the resource being deployed. For example, a Virtual Network Subnet's
-// name might be: "myvnet/mysubnet"
+// For normal resources this name is the full "path" to the resource being deployed. For example, a Virtual Network Subnet's
+// name might be: "myvnet/mysubnet". For extension resources, this is the name of the last resource in the hierarchy (the extension
+// resource itself).
 func (h ResourceHierarchy) FullAzureName() string {
-	var azureNames []string
+	azureNames := h.getAzureNames()
 
-	rootKind := h.RootKind()
-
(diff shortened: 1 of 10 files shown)


hack/generated/pkg/genruntime/sample_resources_test.go
```diff
@@ -0,0 +1,116 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+// TODO: Should this go in its own package?
```
matthchr, line 6: TODO: remove this, it's in the _test package anyway so should be fine

hack/generated/pkg/genruntime/resource_hierarchy.go
```diff
@@ -56,28 +56,88 @@ func (h ResourceHierarchy) Location() (string, error) {
...
+	}
+
+	parentNames := h.getAzureNames()
+	parentNames = parentNames[:len(parentNames)-1]
+
+	// Note: This is actually impossible assuming we don't have a bug, because both of these sets are based on h
+	// which obviously has the same length as itself.
+	if len(parentNames) != len(resourceTypes) {
+		return "", errors.Errorf("hierarchy had %d parent resource types but %d parent names", len(resourceTypes), len(parentNames))
+	}
```
theunrepentantgeek, line 128: If this should never happen, should this instead be a panic?

theunrepentantgeek (approved):
Looks good.

gif

hack/generated/pkg/genruntime/resource_hierarchy.go
matthchr: This is called in the controller (not the generator). In the generator, a panic being a bug is ok (possibly desirable even), but in the controller because it's a single process hosting all of the controllers it feels like an antipattern to panic. I've been trying really hard to avoid any possible panics in the controller context because of that.

PR #1818, opened 2021-09-17
Add tests and samples for missed networking resources

### LINKED ISSUE #1789: Write tests for network resources
We're currently generating Microsoft.Network resources which we don't have tests for.

Resources missing tests:
- VirtualNetworksSubnet (although we use it for the VMSS test so it's kinda already covered)
- VirtualNetworksVirtualNetworkPeering
- VirtualNetworkGateways
- VirtualNetworkTap

These are all included because our config says:
```
  - action: include-transitive
    group: microsoft.network
    version: v*api20201101
    name: VirtualNetwork*
    because: "including all Virtual Network things"
```

matthchr:
Closes #1789

Adds tests for the following resources:
- Subnet (although we had some coverage already it wasn't in its own dedicated test)
- VirtualNetworkPeering

I wrote a test for VirtualNetworkGateway but then learned that it takes 45 minutes to allocate one... so I didn't include it here because that's _insanely_ long.

I removed VirutalNetworkTap because based on [their documentation](https://docs.microsoft.com/en-us/azure/virtual-network/virtual-network-tap-overview) you must have the feature enabled on your subscription to even use it and it's still in preview (even though it's in a non-preview API version...)

**If applicable**:
- [x] this PR contains tests


hack/generated/apis/resources.md
```diff
@@ -43,7 +43,6 @@ v1alpha1api20201101
 - PublicIPAddress
 - VirtualNetwork
 - VirtualNetworkGateway
-- VirtualNetworkTap
```
theunrepentantgeek, line 46: Why remove this resource?

theunrepentantgeek (commented):
Looks good. 

Do you need to sort out the TODO comments in `deployment_error_classifier.go` before approval?

gif

hack/generated/apis/resources.md
matthchr: See the comment in the PR description:
> I removed VirutalNetworkTap because based on [their documentation](https://docs.microsoft.com/en-us/azure/virtual-network/virtual-network-tap-overview) you must have the feature enabled on your subscription to even use it and it's still in preview (even though it's in a non-preview API version...)

From what I can tell, the feature looks pretty "dead" to me: Its documentation hasn't been updated since 2019, and there are no samples/examples that use it. The 2019 documentation says that it's in preview and your subscription must opt in and be approved (or something to that effect).

theunrepentantgeek (approved):

PR #1822, opened 2021-09-20
Ensure expected samples exist in CI

matthchr:
- Rename samples to match resource name exactly.
 - Add missing samples

**If applicable**:


scripts/check_samples.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def strip(text, suffix):
+    if suffix and text.endswith(suffix):
+        return text[:-len(suffix)]
+    return text
+
+def get_expected_samples(crd_path):
+    # For some reason it's 2-5x faster to do this with yq than it is to do it with pyyaml
+    result = subprocess.run(['yq', 'e', '{"name": .spec.names.singular, "group": .spec.group, "version": .spec.versions[] | select(.name != "*storage").name}', str(crd)], capture_output=True, text=True)
```
babbageclunk, line 18: It looks like this is using the global variable `crd` rather than the `crd_path` passed in?

babbageclunk: Also, could you add a comment here with an example of the output of the `yq` command? I'm not great with the syntax.

scripts/check_samples.py
```diff
@@ -0,0 +1,72 @@
...
+        sys.exit(1)
+
+    expected_samples = set()
+    lines = result.stdout.strip().splitlines()
+
+    # TODO: There's probably a way to do this in a list comprehension but my Python is clearly rusty
+    kind = None
+    group = None
+    version = None
+    for line in lines:
```
babbageclunk, line 31: So each gvk is on 3 lines? I'd use the `grouper` recipe on [this list](https://docs.python.org/3/library/itertools.html#itertools-recipes) and then iterate through them with `for kind, group, version in grouper(lines, 3):`.  

Or you could use `cut` to chop off the first part of the line and `paste` to stick them together in bash? But I guess you still have the checking below to do in Python.

babbageclunk (approved):
Looks pretty good - suggested some improvements to the Python, although it turns out mine is a  bit rusty as well!

scripts/check_samples.py
matthchr: Fixed and fixed.

I'm actually not sure how it worked before... lol.

scripts/check_samples.py
matthchr: Good suggestion - done.

matthchr: For clarification, did the `grouper` bit.

PR #1823, opened 2021-09-21
Change ConvertToARM function parameter and add scope

matthchr:
Working towards support for extension resources.


hack/generated/pkg/reflecthelpers/reflect_helpers.go
```diff
@@ -208,3 +201,40 @@ func FindResourceReferences(transformer genruntime.ARMTransformer) (map[genrunti
 
 	return result, nil
 }
+
+// TODO: This method is a bit busy...
```
babbageclunk, line 205: This function doesn't really hold together - each of the 4 sections of the function is using a different input, which is why it needs 5 args. Maybe just put a function around the scope handling at the end and merge the rest back into `ConvertResourceToDeployableResource`? 

If you want that function to be shorter, it would make sense to move the reflection code lines 26-45 into a method like `MetaObject.AsARMTransformer()` - then everything in `ConvertResourceToDeployableResource` would be working at the same abstraction level, I think.

babbageclunk (approved):
Looks good - just some reorganisation suggestions for the function that you weren't very happy with.

hack/generated/pkg/reflecthelpers/reflect_helpers.go
matthchr: but this function is building something from those 5 (really 4, because `ctx` can kinda be ignored) inputs. Those are the 4 inputs required to make a `genruntime.ConvertToARMResolvedDetails`. The awkward/annoying thing is that:
1. We start with `metaObject`.
2. We create `armTransformer` and `resourceHierarchy` from `metaObject`.
3. Neither `armTransformer` nor `resourceHierarchy` "contains" the original `metaObject` though, so we end up needing to pass all 3 around. Even in `CR2DR` all 3 of those types are used later on in the method (although `metaObject` is just for `metaObject.GetName()` used in logging which we could probably come up with an alternative way to do).

Pulling all of this back into `CR2DR` feels worse than having it pulled out (although I'll be the first to admit I don't love how I factored it out).

What we really have here is `metaObject` and then "data derived from or associated with `metaObject`" that I kinda want to pass around together...

Would this pattern be better (looking at what we do in CR2DR)?
1. Resolve what we need to resolve (`resourceHierarchy`, `resolved genruntime.ConvertToARMResolvedDetails`)
2. Extract `armTransformer`.
3. Call `armTransformer.ConvertToARM(resolved)`
4. Do the stuff at the bottom (under the comment): "// We have different deployment models for Subscription rooted vs ResourceGroup rooted resources"

I've sent an update to this PR that does this (in a fresh commit so you can just look at it if you like). Let me know if you think it's better.

babbageclunk: Taking a look - I'll do another review so I can tie comments to code lines.

hack/generated/pkg/reflecthelpers/reflect_helpers.go
```diff
@@ -23,45 +23,18 @@ func ConvertResourceToDeployableResource(
...
-	spec := specFieldPtr.Interface()
-
-	armTransformer, ok := spec.(genruntime.ARMTransformer)
-	if !ok {
-		return nil, errors.Errorf("spec was of type %T which doesn't implement genruntime.ArmTransformer", spec)
-	}
-
-	resourceHierarchy, err := resolver.ResolveResourceHierarchy(ctx, metaObject)
+	// TODO: This should be replaced by metaObject.GetSpec() eventually
+	armTransformer, err := extractARMTransformer(metaObject)
```
babbageclunk, line 27: This is way nicer!

hack/generated/pkg/reflecthelpers/reflect_helpers.go
```diff
@@ -208,3 +181,75 @@ func FindResourceReferences(transformer genruntime.ARMTransformer) (map[genrunti
 
 	return result, nil
 }
+
+type resolved struct {
+	ResolvedDetails   genruntime.ConvertToARMResolvedDetails
+	ResourceHierarchy genruntime.ResourceHierarchy
+}
+
+func resolve(
```
babbageclunk, line 190: This boundary definitely seems like a good one for a function since you don't need to pass too many things in. I'm trying to think of better names for things though. It seems like the `resolved` struct isn't buying you anything, and getting rid of it would free us from needing to come up with a better name for it! 

What about putting the resource hierarchy into `genruntime.ConvertToARMResolvedDetails` and just returning that? It doesn't seem like the struct is used for anything else where the hierarchy wouldn't be available, and it makes sense to have it in there.

If you're not keen on that then maybe just return details, hierarchy and error separately.

hack/generated/pkg/reflecthelpers/reflect_helpers.go
```diff
@@ -208,3 +181,75 @@ func FindResourceReferences(transformer genruntime.ARMTransformer) (map[genrunti
...
+	}
+
+	// Find all of the references
+	refs, err := FindResourceReferences(metaObject)
+	if err != nil {
+		return nil, errors.Wrapf(err, "finding references on %q", metaObject.GetName())
+	}
+
+	// resolve them
+	resolvedRefs, err := resolver.ResolveReferencesToARMIDs(ctx, refs)
```
babbageclunk, line 207: Minor: `resolver.ResolveReferencesToARMIDs` should return a `genruntime.ResolvedReferences`.

hack/generated/pkg/reflecthelpers/reflect_helpers.go
babbageclunk: Now that you've extracted this it seems like it should be a method on `genruntime.Resolver` - it ends up being the entrypoint that drives the rest of the resolver, turning a `MetaObject` into a `ConvertToARMResolvedDetails`.

babbageclunk: Still not sure on the name though - `Resolver.ResolveObject` seems a bit... plodding. `Resolver.FullyResolve` is ok though? I guess part of the problem is that the type is named for the thing it does, and we want it to do that thing.

babbageclunk (approved):
I like this much better - I think there are a couple more things that would help now that you've broken the resolve function out that way.

hack/generated/pkg/reflecthelpers/reflect_helpers.go
babbageclunk: `Resolver.DoItTo(object)`

matthchr: I came up with `Resolver.ResolveForARM`, which isn't terrible? I do agree that this method feels like it should be on `Resolver`. I end up running into a bit of a package dependency problem though:
1. `Resolver.ResolveForARM` uses the `FindResourceReferences` function, which in turn uses `ReflectVisitor`, both of which are defined in the `reflecthelpers` package.
2. `reflecthelpers` has a dependency on `genruntime`, partially for its use of `Resolver` and partially for `MetaObject`.

So there's a circular dependency here, possibly being caused by unclear package purpose (`genruntime` is so much of a catch-all that everything in `reflecthelpers` could in theory go into it too...)

I could move `reflect_helpers.go` into `genruntime`, and just leave `reflect_visitor` in `reflecthelpers`, which would solve this problem... and possibly also be closer to the admitted "reality" right now which is that `genruntime` and `reflecthelpers` are pretty tightly related

hack/generated/pkg/reflecthelpers/reflect_helpers.go
matthchr: Yeah you're right - fixed.

PR #1824, opened 2021-09-22
Move resources

Porges:
Step 1 of v2 reorganization: move v2 resources to v2 folder.

In passing I removed the automatic `format-code` invocation so that it is only invoked on the generated files.


babbageclunk (approved):

hack/generated/go.mod
```diff
@@ -79,4 +80,4 @@ require (
 	sigs.k8s.io/yaml v1.2.0 // indirect
 )
 
-replace github.com/Azure/azure-service-operator => ../../
+replace github.com/Azure/azure-service-operator/v2 => ../../v2
```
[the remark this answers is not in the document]
babbageclunk: (Or something else?)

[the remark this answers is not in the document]
Porges: Neither, stuff has moved and it wouldn't work without the replace, /shrug

The version here seems to be fake, for local replaced modules? But on the other hand the entry for _this_ module in the v2 `go.mod` _does_ have a real version. I have no idea why.

Taskfile.yml
```diff
@@ -64,20 +67,18 @@ tasks:
...
     - task: generator:unit-tests
     - task: basic-checks
-    # Lint is forced to the end because it expects the code is formatted
     - task: generator:lint
 
   generator:ci:
     cmds:
     - task: generator:unit-tests-cover
     - task: basic-checks
-    # lint must be after format-code
```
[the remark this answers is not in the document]
Porges: The lint checks generated code as well, so needs to be after that happens.

Taskfile.yml
```diff
@@ -258,23 +259,23 @@ tasks:
   controller:generate-crds:
     desc: Run controller-gen to generate {{.CONTROLLER_APP}} CRD files.
     deps: [controller:generate-types]
-    dir: "{{.CONTROLLER_ROOT}}"
     sources:
-    - "apis/**/*_gen.go" # depends on all generated types
-    - "pkg/genruntime/**/*.go" # Also depends on the genruntime types as they're embedded into the CRDs
+    - "v2/api/**/*_gen.go" # depends on all generated types
```
matthchr, line 263: Should we be using the `V2_ROOT` variable here?

matthchr: and below throughout

hack/generated/go.mod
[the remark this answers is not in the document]
matthchr: yeah I thought that this was so that you got stuff from the local environment rather than it trying to actually put what's hosted on GitHub?

hack/generated/pkg/testcommon/kube_test_context_envtest.go
```diff
@@ -32,11 +32,11 @@ func createEnvtestContext(perTestContext PerTestContext) (*KubeBaseTestContext,
 	environment := envtest.Environment{
 		ErrorIfCRDPathMissing: true,
 		CRDDirectoryPaths: []string{
-			"../config/crd/bases",
+			"../../../v2/config/crd/bases",
```
matthchr, line 35: Presumably this will also change further in an upcoming PR when this stuff moves too?

matthchr (approved):

matthchr:
> In passing I removed the automatic format-code invocation so that it is only invoked on the generated files.

FWIW I liked it running `format-code` during `task` runs locally, as that would shake out anything I missed formatting.

hack/generated/pkg/testcommon/kube_test_context_envtest.go
Porges: yes

hack/generated/go.mod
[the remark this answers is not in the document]
Porges: > yeah I thought that this was so that you got stuff from the local environment rather than it trying to actually put what's hosted on GitHub?

That's what I thought too. I don't know why it added a real version in the other direction.

Porges:
> > In passing I removed the automatic format-code invocation so that it is only invoked on the generated files.
> 
> FWIW I liked it running `format-code` during `task` runs locally, as that would shake out anything I missed formatting.

This will still happen.

Taskfile.yml
Porges: I thought about it but is `{{.V2_ROOT}}` really better than `v2/`? I may even remove the variable

matthchr: It seems weird to use V2_ROOT in some places and not others though. I'm fine with either (agree the variable seems unneeded), but you use it below.

PR #1825, opened 2021-09-22
Add ObservedGeneration to our Conditions

matthchr:
This gives us a way to pick winners when we're updating the same condition multiple times for a single resource "shape". It follows the KEP on conditions, I had originally not included it because I thought we didn't need it but it turns out we probably do.

This solves the problem of a Warning (of the same generation) immediately getting overwritten by the subsequent reconcile, while maintaining that overwrite in the case the resource had a warning _and was updated_ (since now the generation will be larger).

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/hack/generated/pkg/genruntime/conditions/conditions.go
+++ b/hack/generated/pkg/genruntime/conditions/conditions.go
@@ -96,6 +96,15 @@ type Condition struct {
 	// +kubebuilder:validation:Required
 	LastTransitionTime metav1.Time `json:"lastTransitionTime"`
 
+	// Note: see the https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1623-standardize-conditions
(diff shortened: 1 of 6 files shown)


Porges (approved):

hack/generated/pkg/genruntime/conditions/conditions.go
```diff
@@ -113,9 +122,70 @@ func (c Condition) IsEquivalent(other Condition) bool {
...
+	if c.priority() >= other.priority() {
+		return true
+	} else {
+		return false
+	}
+}
+
+// priority of the condition for overwrite purposes. Higher is more important.
+// The result of this is the following:
+//   1. Status == True conditions, and Status == False conditions with Severity == Warning or Error are all priority 0, meaning that
```
matthchr, line 157: Need to fix this doc

PR #1829, opened 2021-09-22
Fix CI to cleanup ALL resources not just 24h old ones

matthchr:


matthchr:
@porges I removed v2_root here based on the conversation we had on your other PR. If you dislike that let me know and I can put it back.

Porges (approved):

Porges:
Thoughts: We should probably also run `cleanup-azure-resources` _after_ the build so they aren’t lingering? Deleting beforehand could also cause CI errors due to the resources persisting in a deleting state for a long time afterwards.

Porges:
To do this we will need to invoke it in a separate Github Actions step because if the tests timeout the `Taskfile` invocation will fail.

Porges (approved):

PR #1837, opened 2021-09-24
Operator-per-namespace multitenancy in ASOv2

babbageclunk:
This adds the multitenancy features to the v2 operator:
* Restrict the operator to watch specific namespaces with AZURE_TARGET_NAMESPACES
* The operator will tag resources with its namespace and won't reconcile resources claimed by another operator
* Support AZURE_OPERATOR_MODE so the per-namespace operators only run reconcilers, with one central operator instance that handles all the webhooks.

It looks long but ~2500 lines are recorded requests for the tests.

gif

**If applicable**:
- [x] this PR contains tests


v2/internal/controller/controllers/targetnamespace_test.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+	tc.CreateResourceGroupAndWait(&rgWatched)
+	checkNamespaceAnnotation(tc, &rgWatched, podNamespace)
+
+	// But the unwatched namespace isn't...
+	unwatchedName := tc.Namer.GenerateName("rg")
+	rgUnwatched := resources.ResourceGroup{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      unwatchedName,
+			Namespace: "unwatched",
```
matthchr, line 81: minor: maybe `unwatchedNamespace` variable to go along with `unwatchedName`?

matthchr: Or just construct the `NamespacedName` from below earlier and then use that everywhere.

v2/internal/controller/controllers/targetnamespace_test.go
```diff
@@ -0,0 +1,258 @@
...
+	"github.com/Azure/azure-service-operator/v2/internal/controller/reconcilers"
+	"github.com/Azure/azure-service-operator/v2/internal/controller/testcommon"
+
+	. "github.com/onsi/gomega"
+)
+
+const finalizerName = reconcilers.GenericControllerFinalizer
+
+const (
+	timeoutFast = time.Minute * 3
```
matthchr, line 31: This seems like a really long fast timeout, especially for recording tests. We should craft the recording tests in such a way that they're _really_ fast, but here there are cases/modes where we're waiting for 3m no matter what?

Should this be making use of `tc.RemainingTime()`? There's also already a `tc.PollingInterval()` which is also optimized for speed in replaying.

v2/internal/controller/controllers/targetnamespace_test.go
```diff
@@ -0,0 +1,258 @@
...
+		g.Expect(err).NotTo(HaveOccurred())
+		res, err := meta.Accessor(&instance)
+		g.Expect(err).NotTo(HaveOccurred())
+		return res.GetAnnotations()[controllers.NamespaceAnnotation] == podNamespace
+	}
+
+	if configuredNamespaces == "" {
+		t.Log("**** all namespaces mode")
+		// The operator should be watching all namespaces.
+		tc.G.Eventually(
```
matthchr, line 103: Actually you should just use `tc.Eventually` and then you don't need to fuss with `RemainingTime()` and `PollingInterval()` that I mentioned above here, although you might need to for the `Consistently` below...

v2/internal/controller/controllers/targetnamespace_test.go
```diff
@@ -0,0 +1,258 @@
...
+			retry,
+		).Should(
+			BeTrue(),
+			"instance in some namespace never got an operator namespace annotation",
+		)
+	} else {
+		t.Log("**** restricted namespaces mode")
+		// We can tell that the resource isn't being reconciled if it
+		// never gets a finalizer.
+		tc.G.Consistently(
```
matthchr, line 124: So there isn't a friendly `tc.Consistently` that does the right thing with regard to record/replay and polling interval, but maybe you could write one and then use that? It can't wait for `RemainingTime()` as I don't think that'll work in the CI pass (`RemainingTime() is quite long then, I think?) - but it should be able to use `tc.PollingInterval()` and possibly you can make a nice `Consistently` rule

v2/internal/controller/controllers/targetnamespace_test.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+	_, err := tc.CreateResourceGroup(&notMine)
+	tc.Expect(err).NotTo(HaveOccurred())
+
+	name := types.NamespacedName{
+		Name:      notMine.ObjectMeta.Name,
+		Namespace: tc.Namespace(),
+	}
+
+	tc.G.Consistently(
```
matthchr, line 220: Same comments as above.

v2/internal/controller/testcommon/kube_per_test_context.go
```diff
@@ -278,7 +277,7 @@ func (tc *KubePerTestContext) Eventually(actual interface{}, intervals ...interf
 	return tc.G.Eventually(actual, tc.RemainingTime(), tc.PollingInterval())
 }
 
-func (tc *KubePerTestContext) CreateTestResourceGroupAndWait() *v1alpha1api20200601.ResourceGroup {
+func (tc *KubePerTestContext) CreateTestResourceGroupAndWait() *resources.ResourceGroup {
```
matthchr, line 280: Nice catch!

Taskfile.yml
```diff
@@ -146,6 +150,8 @@ tasks:
     cmds:
     - task: controller:verify-samples
     - task: controller:test-integration-envtest-cover
+    - task: controller:test-integration-envtest-target-namespace
```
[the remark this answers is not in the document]
babbageclunk: Ah, I'd intended to do that after this, but looking now I can see that it's easier than I realised! I'll do that now (with the config struct discussed below as well).

v2/internal/controller/controllers/generic_controller.go
```diff
@@ -179,6 +183,10 @@ func MakeResourceGVKLookup(mgr ctrl.Manager, objs []client.Object) (map[schema.G
 	return result, nil
 }
 
+// NamespaceAnnotation defines the annotation name to use when marking
+// a resource with the namespace of the managing operator.
+const NamespaceAnnotation = "azure.microsoft.com/operator-namespace"
```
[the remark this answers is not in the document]
babbageclunk: I think they're naturally disjoint because of the different API groups, right? I can't see that we'd have v1 and v2 operators both trying to manage the same resources.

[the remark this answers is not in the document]
matthchr: Yeah, I'm silly. I read this as "an annotation on the namespace" which I know is 100% not what it is. Ignore this =)

v2/internal/controller/config/operatormode.go
```diff
@@ -0,0 +1,61 @@
...
+		return "watchers-and-webhooks"
+	default:
+		panic(fmt.Sprintf("invalid operator mode value %d", m))
+	}
+}
+
+// ParseOperatorMode converts a string value into the corresponding
+// operator mode.
+func ParseOperatorMode(value string) (OperatorMode, error) {
+	switch value {
```
theunrepentantgeek, line 51: Should this include a `strings.ToLower()` call so that configuration is case-insensitive?

v2/internal/controller/config/operatormode_test.go
```diff
@@ -0,0 +1,62 @@
...
+func TestIncludesWatchers(t *testing.T) {
+	g := NewGomegaWithT(t)
+	g.Expect(config.OperatorModeWebhooks.IncludesWatchers()).To(BeFalse())
+	g.Expect(config.OperatorModeBoth.IncludesWatchers()).To(BeTrue())
+	g.Expect(config.OperatorModeWatchers.IncludesWatchers()).To(BeTrue())
+}
+
+func TestRoundtripString(t *testing.T) {
+	g := NewGomegaWithT(t)
+	values := []string{"watchers-and-webhooks", "webhooks", "watchers"}
```
theunrepentantgeek, line 30: Maybe test some mixed case strings as well?

theunrepentantgeek (approved):
Looks good. Nice and clean.

gif

v2/internal/controller/controllers/generic_controller.go
```diff
@@ -211,6 +219,27 @@ func (gr *GenericReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+		// so if we see another operator already has this object leave
+		// it alone. This will do the right thing in the case of two
+		// operators trying to manage the same namespace. It makes
+		// moving objects between namespaces or changing which
+		// operator owns a namespace fiddlier (since you'd need to
+		// remove the annotation) but those operations are likely to
+		// be rare.
+		message := fmt.Sprintf("Operators in %q and %q are both configured to manage this resource", gr.PodNamespace, reconcilerNamespace)
+		gr.Recorder.Event(obj, corev1.EventTypeWarning, "Overlap", message)
+		return ctrl.Result{}, gr.KubeClient.Client.Update(ctx, obj)
```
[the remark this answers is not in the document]
babbageclunk: Hmm, I might have been cargo-culting. Having a look now.
Ah, I foolishly thought that since Recorder.Event doesn't return an error it must just store something in the object that we still need to write out, but digging into the implementation it's because it's a fire-and-forget type thing, which makes a lot more sense. I'll remove the update here.

v2/internal/controller/controllers/operatormode_test.go
```diff
@@ -0,0 +1,143 @@
...
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	resources "github.com/Azure/azure-service-operator/v2/api/microsoft.resources/v1alpha1api20200601"
+	"github.com/Azure/azure-service-operator/v2/internal/controller/config"
+	"github.com/Azure/azure-service-operator/v2/internal/controller/testcommon"
+
+	. "github.com/onsi/gomega"
```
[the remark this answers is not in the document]
babbageclunk: Will check!

v2/internal/controller/controllers/operatormode_test.go
```diff
@@ -0,0 +1,143 @@
...
+)
+
+// Rerecording these tests should always be done with
+// AZURE_OPERATOR_MODE=watchers-and-webhooks to ensure that all needed
+// requests are in the recording.
+
+func TestOperatorModeWebhooks(t *testing.T) {
+	t.Parallel()
+	tc := globalTestContext.ForTest(t)
```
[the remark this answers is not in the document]
babbageclunk: Yeah, I like this much better - given how the per-test context works it'll be straightforward to do.

v2/internal/controller/controllers/targetnamespace_test.go
babbageclunk: Good point, I'll check this.

v2/internal/controller/controllers/operatormode_test.go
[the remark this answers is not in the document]
babbageclunk: Yup, fixed 'em.

v2/internal/controller/controllers/targetnamespace_test.go
babbageclunk: Rejigging the tests to pass in the config ended up getting rid of these, since I switched everything to just using `tc.Eventually`.

v2/internal/controller/controllers/targetnamespace_test.go
babbageclunk: Well, `Consistently` when it's successful ends up waiting for the the full time, so I don't think that can use `RemainingTime`. These aren't ever waiting for ARM, only ever to make sure that the object in question doesn't get a finalizer, so I might just pull it into a function rather than sticking it onto `KubePerTestContext`. Although one downside of that is the no call stack problem... Might chat with you about this one tomorrow.

babbageclunk: I ended up doing this - I think it's good.

v2/internal/controller/config/vars.go
```diff
@@ -0,0 +1,102 @@
...
+
+const (
+	subscriptionIDVar   = "AZURE_SUBSCRIPTION_ID"
+	targetNamespacesVar = "AZURE_TARGET_NAMESPACES"
+	operatorModeVar     = "AZURE_OPERATOR_MODE"
+	podNamespaceVar     = "POD_NAMESPACE"
+)
+
+// Values stores configuration values that are set for the operator.
+type Values struct {
```
matthchr, line 21: Nice - this looks really clean to me.

v2/internal/controller/config/vars.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	if !v.OperatorMode.IncludesWatchers() && len(v.TargetNamespaces) > 0 {
+		return errors.Errorf("%s must include watchers to specify target namespaces", targetNamespacesVar)
+	}
+	return nil
+}
+
+// ParseTargetNamespaces splits a comma-separated string into a slice
+// of strings with spaces trimmed.
+func ParseTargetNamespaces(fromEnv string) []string {
```
matthchr, line 92: Is this intentionally exported?

v2/internal/controller/controllers/operatormode_test.go
```diff
@@ -0,0 +1,108 @@
...
+			Tags:     testcommon.CreateTestResourceGroupDefaultTags(),
+		},
+	}
+	tc.Expect(rg.Spec.AzureName).To(Equal(""))
+
+	_, err := tc.CreateResourceGroup(&rg)
+	tc.Expect(err).ToNot(HaveOccurred())
+	// AzureName should have been defaulted on the group on the
+	// way in (it doesn't require waiting for a reconcile).
+	// TODO: this works here, but it's possibly because envtest
```
matthchr: Not 100% sure what you are getting at with this TODO, but I think that this assert would work even if APIServer was remote. Why do you think it wouldn't?

v2/internal/controller/controllers/operatormode_test.go
```diff
@@ -0,0 +1,108 @@
...
+		},
+	}
+	tc.Expect(rg.Spec.AzureName).To(Equal(""))
+
+	_, err := tc.CreateResourceGroup(&rg)
+	tc.Expect(err).NotTo(HaveOccurred())
+
+	// AzureName should have been defaulted on the group on the
+	// way in (it doesn't require waiting for a reconcile).
+	// TODO: this works here, but it's possibly because envtest
```
matthchr: Same comment as above

v2/internal/controller/controllers/operatormode_test.go
matthchr: As discussed on teams, delete this I think since it is guaranteed

v2/internal/controller/controllers/targetnamespace_test.go
```diff
@@ -0,0 +1,217 @@
...
+	"github.com/Azure/azure-service-operator/v2/internal/controller/reconcilers"
+	"github.com/Azure/azure-service-operator/v2/internal/controller/testcommon"
+)
+
+const finalizerName = reconcilers.GenericControllerFinalizer
+
+func TestTargetNamespaces(t *testing.T) {
+	t.Parallel()
+	podNamespace := "some-operator"
+	tc := globalTestContext.ForTestWithConfig(t, config.Values{
```
matthchr, line 31: This is very cool.

v2/internal/controller/config/vars.go
```diff
@@ -0,0 +1,102 @@
...
+		if err != nil {
+			return Values{}, err
+		}
+		result.OperatorMode = mode
+	}
+
+	result.SubscriptionID = os.Getenv(subscriptionIDVar)
+	result.PodNamespace = os.Getenv(podNamespaceVar)
+	result.TargetNamespaces = ParseTargetNamespaces(os.Getenv(targetNamespacesVar))
+	// Not calling validate here to support using from tests where we
```
matthchr, line 57: Are we actually violating the configs expectations in any of the tests that use `ReadFromEnvironment`? I would've thought that the expectations were only violated in the tests that passed a manually constructed config?

matthchr (approved):

v2/internal/controller/config/vars.go
babbageclunk: No, that's historical - I'll hide it.

v2/internal/controller/config/vars.go
babbageclunk: The main one is the check for subscription id - since we mostly run the tests in replay mode we don't want to require this to be set in the environment.

PR #1839, opened 2021-09-27
Avoid reusing variables incorrectly in property assignment functions

babbageclunk:

If a map of objects was named Keys or a slice of objects was named Indexes, the generated object conversion code `copyVar` would shadow the `key` or `index` loop variable - adding them to the locals before calling conversion prevents this.

[Generated code exhibiting the problem](https://github.com/babbageclunk/azure-service-operator/blob/7cde616141080bed136919e9191c61964ee9d0ab/v2/api/microsoft.documentdb/v1alpha1api20210515/database_accounts_mongodb_databases_collection_types_gen.go#L1149)

It looks like this could also cause a problem with a map of maps or slice of slices? Or possibly not - I think it needs the `copyVar` from the object handler to cause the issue.

Link to gif because embedding it doesn't seem to work: https://tenor.com/58KF.gif

**If applicable**:
- [x] this PR contains tests


hack/generator/pkg/codegen/testdata/ArmResource/Arm_test_simple_resource_array_properties_azure.golden
```diff
@@ -255,29 +255,29 @@ func (fakeResourceSpec *FakeResource_Spec) ConvertToARM(name string, resolvedRef
...
 	for _, item := range fakeResourceSpec.ArrayOfArraysOfArrays {
 		var itemTemp [][]FooARM
-		for _, item1 := range item {
-			var item1Temp []FooARM
-			for _, item2 := range item1 {
-				item2ARM, err := item2.ConvertToARM(name, resolvedReferences)
+		for _, itemItem := range item {
+			var itemItemTemp []FooARM
+			for _, itemItemItem := range itemItem {
+				itemItemItemARM, err := itemItemItem.ConvertToARM(name, resolvedReferences)
```
babbageclunk, line 274: This isn't great - any suggestion for fixing this @theunrepentantgeek?

hack/generator/pkg/codegen/testdata/ArmResource/Arm_test_simple_resource_array_properties_azure.golden
```diff
@@ -335,31 +335,31 @@ func (fakeResourceSpec *FakeResource_Spec) PopulateFromARM(owner genruntime.Know
...
 		var itemTemp [][]Foo
-		for _, item1 := range item {
-			var item1Temp []Foo
-			for _, item2 := range item1 {
-				var item3 Foo
-				err := item3.PopulateFromARM(owner, item2)
+		for _, itemItem := range item {
+			var itemItemTemp []Foo
+			for _, itemItemItem := range itemItem {
+				var itemItemItem1 Foo
```
babbageclunk, line 355: These ones haven't come out very well - any idea how to make it better @theunrepentantgeek?

babbageclunk:
Abandoning this PR - @theunrepentantgeek has an idea for a different solution that might work out better, and I've got a workaround in the meantime.

PR #1840, opened 2021-09-28
Finish support for extension resources

matthchr:
- Use UnknownOwnerReference rather than KnownResourceReference as KnownResourceReference doesn't allow multiple types to have ownership, whereas extension resources can root to anything.
- Allow extension resources to be owned by anything
  - Change generated Owner() function to work for extension resources
  - Change the type of the Owner property on extension resources to be ResourceReference rather than KnownResourceReference to reflect the reality that any GroupKind can own them.

**If applicable**:
- [x] this PR contains tests


theunrepentantgeek (approved):
Looks good.

Is `KnownReference` now used at all? Wondering if it needs-to/could be deleted.


gif

matthchr:
> Is KnownReference now used at all? Wondering if it needs-to/could be deleted.

It is used still, by the actual generated types that restrict an owner to a particular (known) type. The interface has to take the more generic of the two (Unknown) to cater for cases where the ref isn't known, but that's ok because a KnownReference can be converted to an UnknownReference easily.

matthchr:
I'm thinking I should change `UnknownReference` to `ArbitraryReference` or something... might send an update with that change

PR #1841, opened 2021-09-28
Swagger nondeterminism fixes

### LINKED ISSUE #1838: Nondeterminism in Swagger inputs
Swagger inputs are leading to nondeterministic outputs for a few reasons:

- Types in “common” files like `common-types/resource-management/v1/types.json` are pulled into the same group `Microsoft.Whatever` that the including types are in, for simplicity. However, they can clash with types that are already in that group. (This is most common with the name `Resource`.)
- Teams can have multiple files in the same Group+Version that have types with identical names. (e.g. `Microsoft.KeyVault` `keys.json` and `secrets.json` both have a type called `Attributes`).

Porges:
Fixes #1838.

We currently have two sources of non-determinism in Swagger imports:

1. If a file (say, `network.json`) refers to an external file (say, `common/types.json`), any types referred to in that file will be pulled into the same output package (same group & version) as the originating file. This is on purpose, so that shared types get copied into each output package (some shared files have no version or group). However, it can cause issues if the types in the external file share names that collide with types in the other package.
    - Fix: add the ability to provide namespace for common external files, which will be respected when referring to types from those files. Also add an error/check that will fail if this does happen so we can try to fix it via config.
      - Currently `Microsoft.Common` is specified for files under the  `common-types/resource-management` directory. Note that these types all disappear during generation due to allof/oneof collapsing, so we don’t actually see them in the output.
2. With Swagger there can be multiple `.json` files under each group/version, and they can share the same type names. When we combine them all into one package then types can get overwritten. We must not do this.
   -   Fix: Change to process each `.json` file individually and then merge the results by package. When typename collisions are found, we check if they are structurally identical. If not, they are resolved by prefixing each of the colliding names with the filename it was sourced from.
       - There is still a potential source of non-determinism here ~~to address before merge~~ which is also addressed:
          1. Type `X` (specified in multiple files) refers to type `Y`. `X` collides but is structurally identical, so it is okay to merge.
          2. However, type `Y` collides with others that are not structurally identical.
          3. Thus, all `Y`s get renamed.
          4. Types `X` still collide and are still structurally identical, but when generating the output we will pick an arbitrary exemplar. This means that the specific typename of `Y` that `X` refers to is random. We address this by sorting the files and always picking the first one.

This is a rather big PR as it reworks several core things. However, the output should not change!

### Ability to specify namespace for a directory

We previously had [overrides included from upstream in our config](https://github.com/Azure/azure-service-operator/blob/20832716956c8f3e3fa5da11cdda6abd075d815b/hack/generator/azure-arm.yaml#L375-L404), but only applied the `Suffix` override. We now apply the `Namespace` override as well. This is what is used to specify the `Microsoft.Common` namespace/group name.

### Testing structural equality

In order to detect structural equality, this expands the `type.Equals` function to take an additional parameter which can override the equality check for specific types (at the moment only `TypeName`). 

To prevent accidentally-not-passing the `overrides` argument, it is required on `type.Equals`. `TypeEquals` has been modified to take it as an optional varags argument. Callsites which invoked `type.Equals` directly have generally been updated to invoke `TypeEquals` instead (unless they are in a context where they have access to an `overrides` argument).

### `astmodel.DiffTypes`

This is added as I found it very useful when debugging. When `AddAllowDuplicates` fails, it will generate a structural diff between the types.


workspace.code-workspace
```diff
@@ -1,8 +1,5 @@
 {
 	"folders": [
-		{ 
-			"path": "."
```
Porges, line 4: undo

hack/generator/pkg/codegen/pipeline/status_from_swagger.go
```diff
@@ -57,7 +57,7 @@ func AddStatusFromSwagger(idFactory astmodel.IdentifierFactory, config *config.C
 				return nil, errors.Wrapf(err, "unable to load Swagger data")
 			}
 
-			klog.V(1).Infof("Loaded Swagger data (%d resources, %d other types)", len(swaggerTypes.resources), len(swaggerTypes.otherTypes))
+			klog.V(1).Infof("Loaded Swagger data (%d resources, %d other types)", len(swaggerTypes.ResourceTypes), len(swaggerTypes.OtherTypes))
```
Porges, line 60: Please hold off reviewing this file until I’ve refactored the large functions here.

hack/generator/pkg/jsonast/schema_abstraction_openapi.go
```diff
@@ -263,59 +261,100 @@ func (schema *OpenAPISchema) isRef() bool {
...
+		// make sure that pullling the type into the other package wouldn’t conflict with
+		// any definitions in that package
+		// note that this won’t detect collisions with another sibling file but it’s better than nothing
+		if absRefPath != schema.fileName {
+			referringPackage, err := schema.cache.fetchFileAbsolute(schema.fileName)
+			if err != nil {
+				panic(err) // assert, not error; should always have been loaded already
+			}
+
+			if _, ok := referringPackage.Swagger.Definitions[name]; ok {
```
Porges, line 293: Should add a test that exercises this error path to ensure it actually works.

hack/generator/pkg/jsonast/schema_abstraction.go
```diff
@@ -66,9 +68,7 @@ type Schema interface {
 
 	// ref things
 	isRef() bool
-	refGroupName() (string, error)
-	refObjectName() (string, error)
-	refVersion() (string, error)
+	refTypeName() (astmodel.TypeName, error)
```
Porges, line 71: Changed from separate parts to a combined `TypeName` as this makes it a lot easier to carry around `LocalPackaReference`s rather than separate `group`/`version` and combine them later. The latter was rather error-prone as well.

hack/generator/pkg/astmodel/type.go
```diff
@@ -35,7 +35,7 @@ type Type interface {
 	AsZero(types Types, ctx *CodeGenerationContext) dst.Expr
 
 	// Equals returns true if the passed type is the same as this one, false otherwise
-	Equals(t Type) bool
+	Equals(t Type, overrides EqualityOverrides) bool
```
theunrepentantgeek, line 38: It seems as though you want most (maybe all?) consumers to use `TypeEquals()` instead of calling this directly. Should we lower the visibility so that consumers outside this package have no choice?

theunrepentantgeek, line 38: It seems as though you want most (maybe all?) consumers to use `TypeEquals()` instead of calling this directly. Should we lower the visibility so that consumers outside this package have no choice?

hack/generator/pkg/astmodel/type.go
```diff
@@ -63,11 +67,22 @@ type DeclarationContext struct {
 	Validations []KubeBuilderValidation
 }
 
+func CombineOverrides(overrides ...EqualityOverrides) EqualityOverrides {
```
theunrepentantgeek, line 70: This isn't a combine - it's _use-the-last-one-provided-and-ignore-the-others_.

At minimum, needs a more accurate name ... better would be to have a composite override that really does combine.

hack/generator/pkg/codegen/pipeline/status_from_swagger_test.go
```diff
@@ -57,11 +59,77 @@ func Test_ShouldSkipDir_GivenPath_HasExpectedResult(t *testing.T) {
 		c := c
 		t.Run(c.name, func(t *testing.T) {
 			t.Parallel()
-			g := NewGomegaWithT(t)
+			g := gomega.NewWithT(t)
 
 			skipped := shouldSkipDir(c.path)
 
-			g.Expect(skipped).To(Equal(c.shouldSkip))
+			g.Expect(skipped).To(gomega.Equal(c.shouldSkip))
```
theunrepentantgeek, line 66: Didn't we standardise on using implicit+imports + `NewGomegaWithT()` to minimize stuttering through the exports?

theunrepentantgeek (commented):
Partial review, will come back to it.

gif

hack/generator/pkg/astmodel/type.go
Porges: @theunrepentantgeek That’s a good idea, let me see if it breaks anything…

hack/generator/pkg/astmodel/type.go
Porges: There’s no way that a Combine can work in general; some uses want to make equality stricter, some want to make it looser. I’ll rename this…

Porges: … actually removing as YAGNI.

hack/generator/pkg/astmodel/type.go
Porges: Cannot be done; `PropertyBagMemberType` is defined in another package.

theunrepentantgeek (approved):
LGTM.

gif

PR #1844, opened 2021-09-29
Move more of the controller out of hack and into v2

matthchr:
There are other problems we should solve (the crossplane directory and generator), but this at least removes the circular reference we had which was causing me a lot of issues
gif

**If applicable**:
- [x] this PR contains tests


matthchr:
Hmm I think I broke the Docker build for the controller image - fixing that...

matthchr:
Will be waiting for @porges approval on this (to ensure I didn't screw it up and didn't collide with anything he's doing)

matthchr:
Fixed the Dockerfile stuff

hack/generator/pkg/config/configuration.go
```diff
@@ -57,6 +57,12 @@ type Configuration struct {
 	// after init TypeTransformers is split into property and non-property transformers
 	typeTransformers     []*TypeTransformer
 	propertyTransformers []*TypeTransformer
+
+	// typesOutputModuleSuffix is the suffix to append to GoModulePath to build the actual go import path
+	// to include types with. For example if GoModulePath is "github.com/Azure/azure-service-operator/v2"
+	// and we locally put the generated types into an "api" folder then this suffix would be "api" so that
+	// when joined with GoModulePath the result is github.com/Azure/azure-service-operator/v2/api
+	typesOutputModuleSuffix string
```
Porges, line 65: this was added but not actually used in the config file?

hack/generator/azure-arm.yaml
```diff
@@ -2,9 +2,9 @@ schemaUrl: 'https://schema.management.azure.com/schemas/2019-04-01/deploymentTem
 schemaUrlRewrite:
   from: 'https://schema.management.azure.com/schemas/'
   to: 'specs/azure-resource-manager-schemas/schemas/'
-destinationGoModuleFile: ../generated/go.mod
+destinationGoModuleFile: ../../v2/go.mod
 typesOutputPath: ../../v2/api
```
Porges, line 6: one change I made when sketching this was to have the `typesOutputPath` be relative to the destination module file (since `typeRegistrationOutputFile` already is), but I can make this change later

Porges (approved):

hack/generator/pkg/config/configuration.go
matthchr: It's not for the config file - it's cached and set during `initialize` (mostly to avoid having to return err or panic where I need it)

Issue #1434, opened 2021-05-14
Simplify APIs by pulling types forward


theunrepentantgeek:
While most teams version their entire API, some teams only publish a subset of their types in each release of the API. 
(Currently we're concerned about the `microsoft.network` ARM APIs, but this may apply to others too.)

The problem here is that we might end up having to generate multiple versions of the API simply to ensure we have full coverage of the types required for the _latest_ version of the API.

We should consider the idea of "pulling up" types into later versions so that they are complete and can be used in isolation.

# Example

`DnsZone` and related types haven't been released in any version of the `microsoft.network` ARM API since `v20180501`. 

If these are needed by consumers of `v20200501`, they would have to import the earlier version just for access to``DnsZone`.

If we promoted all of the types related to `DnsZone` up to `v20200501` then consumers would get everything they need in the single API version, simplifying consumption.

# Approach

We'd need careful configuration to ensure that we only pull forward the types we need.

Each type would need to be tagged with the original ARM version (so that we can do the right thing when talking to the ARM API)

theunrepentantgeek:
We may be able to solve this just through config - say by explicitly importing `DnsZone` from the older version alongside the later imports of `microsoft.network`.

PR #1845, opened 2021-09-29
Modify ConversionGraph to support hub types not in the most recent package

theunrepentantgeek:

To this point we've made the unproven assumption that, at least for resources of interest, each version of an ARM service will include all the resources required for interaction with that service.

Though convenient, we've known that this assumption was unproven (see #1434) and that we would need to confirm or reject it at some point.

We've now how a concrete case that disproves the assumption: In the `Microsoft.Compute` namespace, API version `2020-09-30` includes the `Disk` resource, but version `2020-12-01`, despite being a later release, does not.

So, when identifying the hub type, we need to limit candidate packages to those that contain the specified resource, and select the latest available storage package from that reduced set instead of the complete set.

gif

**If applicable**:
- [x] this PR contains tests


hack/generator/pkg/codegen/pipeline/implement_convertible_interface.go
```diff
@@ -28,30 +27,38 @@ func ImplementConvertibleInterface(idFactory astmodel.IdentifierFactory) Stage {
...
+					for _, fn := range rsrc.Functions() {
+						if propertyAssignmentFn, ok := fn.(*functions.PropertyAssignmentFunction); ok {
+
+							conversionFn := functions.NewResourceConversionFunction(hub, propertyAssignmentFn, idFactory)
+							conversionFunctions = append(conversionFunctions, conversionFn)
+						}
+					}
+
+					// Create the interface implementation and inject into the rsrc
+					impl := astmodel.NewInterfaceImplementation(astmodel.ConvertibleInterface, conversionFunctions...)
```
babbageclunk, line 50: No check on the number of functions anymore - is that because they'll always exist?

hack/generator/pkg/codegen/storage/conversion_graph.go
```diff
@@ -35,19 +35,54 @@ func (graph *ConversionGraph) LookupTransition(start astmodel.PackageReference)
...
+		// We do this to allow for gaps where a type is dropped and then restored across versions
 		ref = r
 	}
+}
+
+// FindHub returns the type name of the hub resource, given the type name of one of the resources that is
+// persisted using that hub type. This is done by following links in the conversion graph until we either reach the end
+// or we find that a newer version of the type does not exist.
+// Returns the hub type and true if found; an empty name and false if not.
+// If the name passed in is for the hub type for the given resource, no hub type will be found.
```
babbageclunk, line 70: These return values are confusing. If you call `FindHub` with the hub, it should return `(hub, true)`.
If you call it with a non-hub name, it should return `(hub, false)`.

babbageclunk (approved):
Looks good - minor quibble about the return semantics of `FindHub`.

hack/generator/pkg/codegen/pipeline/implement_convertible_interface.go
theunrepentantgeek: The conversion functions should exist on every non-hub resource, and hubs are short circuited earlier in the function.

hack/generator/pkg/codegen/storage/conversion_graph.go
theunrepentantgeek: I wrote the function to make it easier to consume, but actually I agree with you. Will change.

theunrepentantgeek: Done.

PR #1847, opened 2021-09-30
Fix bug with name collisions in generated code

theunrepentantgeek:

While setting up code generation for a new resource type, @babbageclunk noticed a bug in the way we were generating local identifiers.

When a property name collided with one of the local variable suffixes used for conversion, the code was generated with a name collision that results in a compilation error, as can be seen in this test case:

``` go
// Keys
keyMap := make(map[string]Release)
for key, keyValue := range source.Keys {          // Loop variable is 'key'
    // Shadow the loop variable to avoid aliasing
    keyValue := keyValue
    var key Release                               // Staging variable is also 'key'
    err := key.AssignPropertiesFromRelease(&keyValue)
    if err != nil {
        return errors.Wrap(err, "populating Keys from Keys, calling AssignPropertiesFromRelease()")
    }

    keyMap[key] = key
}
person.Keys = keyMap
```

Note how both the loop and the staging variables are named `key`.

The root cause for this was due to having two different sets of known locals active at the same time - one in which the loop variable was defined, and another for the body of the loop.

After @babbageclunk wrote some additional tests to document the problems, he and I brainstormed a bunch of possible solutions, but each had significant edge cases resulting in issues with the generated code.

The solution in this PR was to go in the other direction - to simplify rather than introduce additional complexity. 

Local variables are now _only_ allocated during the code generation phase, not during the creation of conversion, ensuring that there is no possibilty of carryover between generation calls.

gif

**If applicable**:
- [x] this PR contains tests


hack/generator/pkg/astmodel/known_locals_set.go
```diff
@@ -89,36 +114,13 @@ func (locals *KnownLocalsSet) HasName(name string) bool {
 
 // Clone clones the KnownLocalsSet
 func (locals *KnownLocalsSet) Clone() *KnownLocalsSet {
-	return &KnownLocalsSet{
-		idFactory: locals.idFactory,
-		names:     locals.names.clone(),
+	names := make(map[string]struct{}, len(locals.names))
```
Porges, line 117: We should probably have a `StringSet` type since `map[string]struct{}` appears a bunch of times.

theunrepentantgeek: I'll add that to a CodeGardening TODO list that I have.

babbageclunk (approved):
This way is heaps simpler! Maybe it would make sense to include known locals in the code generation context? It fits with the other stuff in there pretty well, although I guess you'd need some more machinery around creation of nested scopes so possibly not worth doing.

PR #1848, opened 2021-09-30
Add new Role Assignment resource

matthchr:
This is the first extension resource we support. I've added some additional test helpers and various other things which came up when I was trying to write a test for it.


theunrepentantgeek (approved):
Looks relatively straightforward. :+1:

gif

PR #1849, opened 2021-09-30
Initial support for 'format' in specs

Porges:
Values of type `string` in JSON/OpenAPI schema can also have `format` which describes the format of the string. Add support for transforming this into a regex validation that Kubebuilder can use.

At the moment this doesn’t have a huge impact (since we are consuming the ARM schemata which are already preprocessed) but if we come to rely on Swagger more then this will be used more.

There are a few interesting formats: potentially `password` could be leveraged to indicate where we should handle secrets Kubernetes-natively.

### DIFF
--- a/hack/generator/pkg/astmodel/kubebuilder_validations.go
+++ b/hack/generator/pkg/astmodel/kubebuilder_validations.go
@@ -12,9 +12,10 @@ import (
 	"regexp"
 	"strings"
 
-	"github.com/Azure/azure-service-operator/hack/generator/pkg/astbuilder"
 	"github.com/dave/dst"
 	"k8s.io/klog/v2"
+
+	"github.com/Azure/azure-service-operator/hack/generator/pkg/astbuilder"
 )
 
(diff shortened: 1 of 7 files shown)


matthchr (approved):
LGTM

hack/generator/pkg/astmodel/kubebuilder_validations.go
```diff
@@ -130,7 +131,7 @@ func ValidateMaxLength(length int64) KubeBuilderValidation {
 	return KubeBuilderValidation{MaxLengthValidationName, length}
 }
 
-func ValidatePattern(pattern regexp.Regexp) KubeBuilderValidation {
+func ValidatePattern(pattern *regexp.Regexp) KubeBuilderValidation {
```
theunrepentantgeek, line 134: The current name `ValidatePattern` reads as though it's doing the validation, not creating a validation item using a pattern.

Suggest improving the name, maybe `AsValidationPattern()`?

theunrepentantgeek (approved):
One suggested rename, otherwise looks good.

gif

PR #1850, opened 2021-10-01
Code gardening of astmodel

theunrepentantgeek:

Minor cleanup.

* Update the implementations of `WriteDebugDescription` to not crash if called on a nil type reference
* Remove use of hard casts in favour of `As*()` utility methods
* Introduce `StringSet` and use it

gif


hack/generator/pkg/astmodel/string_set.go
```diff
@@ -0,0 +1,25 @@
...
+	return ok
+}
+
+func (set StringSet) Add(s string) {
+	set[s] = struct{}{}
+}
+
+func (set StringSet) Remove(s string) {
+	delete(set, s)
+}
```
matthchr, line 25: minor: missing newline at EoF

matthchr (approved):

hack/generator/pkg/astmodel/string_set.go
theunrepentantgeek: Fixed.

PR #1933, opened 2021-10-27
Apply heuristics to add line breaks to complex function calls

theunrepentantgeek:

Some of our generated function calls are difficult to read because of the number of parameters.

This PR adds two heuristics for when arguments to function calls should be broken onto separate lines:
* When the number of arguments is > 3
* When the number of nested function calls is > 1

gif

### DIFF
--- a/v2/tools/generator/internal/astbuilder/calls.go
+++ b/v2/tools/generator/internal/astbuilder/calls.go
@@ -14,10 +14,7 @@ import (
 // <funcName>(<arguments>...)
 //
 func CallFunc(funcName string, arguments ...dst.Expr) *dst.CallExpr {
-	return &dst.CallExpr{
-		Fun:  dst.NewIdent(funcName),
-		Args: Expressions(arguments),
-	}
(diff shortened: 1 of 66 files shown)


v2/tools/generator/internal/astbuilder/calls.go
```diff
@@ -26,27 +23,53 @@ func CallFunc(funcName string, arguments ...dst.Expr) *dst.CallExpr {
...
+	return createCallExpr(
+		&dst.SelectorExpr{
 			X:   expr,
 			Sel: dst.NewIdent(funcName),
 		},
+		arguments...)
+}
+
+func createCallExpr(expr dst.Expr, arguments ...dst.Expr) *dst.CallExpr {
+	// Check to see howe many of our arguments are nested function calls
```
babbageclunk, line 48:
```suggestion
	// Check to see how many of our arguments are nested function calls
```

babbageclunk (approved):

Issue #1941, opened 2021-11-01
Tests for managed identity in ASO v2


matthchr:
Need automated tests for managed identity. This can possibly be done in `kind` using `type: 1` and `aad-pod-identity`, but probably needs to be live tests (unlike our current mostly recorded, mostly envtest infrastructure).

PR #1942, opened 2021-11-01
Managed identity support

### LINKED ISSUE #1854: Ensure that AAD Pod Identity integration in ASOv2 works well
ASOv1 has an AAD Pod Identity binding named `aso_manager_binding`.

**Describe the improvement**
`aso_manager_binding` is a fine name when using AAD Pod Identity through its Helm installation, but when using it through the AKS addon, the name of the binding is by default the value used in the `selector`, which is `aso_manager_binding`, but since `_`'s are not allowed in Kubernetes resource names, that causes issues.

We should instead use a selector for ASO v2 that doesn't contain underscores (`aso-manager-binding`) or something.

### LINKED ISSUE #1869: Feature: Support managed identity in ASO v2
- [x] Support managed identity in ASO v2
- [ ] Tests for managed identity (they will need to run in Azure).
- [ ] Ensure that #1854 is dealt with.

matthchr:
Closes #1869 
Closes #1854

Adds Managed Identity support and documents how to use it. Testing is still an outstanding issue which we need to solve.
See #1941. I will be following up with a separate PR for this.

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/docs/hugo/content/docs/authentication.md
+++ b/docs/hugo/content/docs/authentication.md
@@ -1,0 +1,99 @@
+# Authentication in Azure Service Operator v2
+
+Azure Service Operator supports two different styles of authentication today.
+
+1. Service Principal
+2. aad-pod-identity authentication (managed identity)
+
+## Service Principal
+
+### Prerequisites
+1. An existing Azure Service Principal.
+
+To use Service Principal authentication, specify an `aso-controller-settings` secret with `AZURE_CLIENT_ID` and `AZURE_CLIENT_SECRET` set.
+
+* `AZURE_CLIENT_ID` must be set to the Service Principal client ID. This will be a GUID.
+* `AZURE_CLIENT_SECRET` must be set to the Service Principal client secret.
+
+For more information about Service Principals, see [creating an Azure Service Principal using the Azure CLI](https://docs.microsoft.com/cli/azure/create-an-azure-service-principal-azure-cli#password-based-authentication).
+The `AZURE_CLIENT_ID` is sometimes also called the App ID. The `AZURE_CLIENT_SECRET` is the "password" returned by the command in the previously linked documentation.
+
+```bash
+cat <<EOF | kubectl apply -f -
+apiVersion: v1
+kind: Secret
+metadata:
+ name: aso-controller-settings
+ namespace: azureserviceoperator-system
+stringData:
+ AZURE_SUBSCRIPTION_ID: "$AZURE_SUBSCRIPTION_ID"
+ AZURE_TENANT_ID: "$AZURE_TENANT_ID"
+ AZURE_CLIENT_ID: "$AZURE_CLIENT_ID"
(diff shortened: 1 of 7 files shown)


docs/hugo/content/docs/authentication.md
```diff
@@ -0,0 +1,95 @@
...
+2. aad-pod-identity authentication (managed identity)
+
+## Service Principal
+
+### Prerequisites
+1. An existing Azure Service Principal.
+
+To use Service Principal authentication, specify an `aso-controller-settings` secret with `AZURE_CLIENT_SECRET` set.
+`AZURE_CLIENT_ID` must be set to the Service Principal client ID.
+`AZURE_CLIENT_SECRET` must be set to the Service Principal client secret.
```
theunrepentantgeek, line 15:
```suggestion
To use Service Principal authentication, specify an `aso-controller-settings` secret with `AZURE_CLIENT_ID` and `AZURE_CLIENT_SECRET` set.

* `AZURE_CLIENT_ID` must be set to the Service Principal client ID.
* `AZURE_CLIENT_SECRET` must be set to the Service Principal client secret.
```

Maybe give an example of each, so that people can recognize when they've found the right value to include?

docs/hugo/content/docs/authentication.md
```diff
@@ -0,0 +1,95 @@
...
+1. An existing Azure Managed Identity.
+2. [aad-pod-identity](https://github.com/Azure/aad-pod-identity) installed into your cluster. If you are running ASO on an Azure Kubernetes Service (AKS) cluster, you can instead use the
+   [integrated aad-pod-identity](https://docs.microsoft.com/azure/aks/use-azure-ad-pod-identity).
+
+First, set the following environment variables:
+```bash
+export IDENTITY_RESOURCE_GROUP="..."
+export IDENTITY_NAME="..."
+export AZURE_SUBSCRIPTION_ID="..."
+export AZURE_TENANT_ID="..."
```
theunrepentantgeek, line 44: Same comment here - one of the regular trip hazards I find is when documentation is unclear about what's expected. 

E.g. is `IDENTIY_RESOURCE_GROUP` just the name of the group (is it case sensitive?), or a fully qualified path including the subscription ID, or maybe an ARM URL?

theunrepentantgeek (approved):
Some minor comments, but nothing substantial - great stuff.

docs/hugo/content/docs/authentication.md
matthchr: Giving an example here is a bit awkward because both fields are pretty inscrutable (CLIENT_ID is a GUID and the password is some black-box base64'ed string that Azure gives to them when they create the SP). What I've done is call out that client ID is a GUID and also link to the documentation in the Azure CLI that shows how to create an SP (which will return these values).

docs/hugo/content/docs/authentication.md
matthchr: I've provided examples here and also added some comments at the end of these to explain what they are.

PR #1946, opened 2021-11-02
Update v2 README to use `kubectl create`

babbageclunk:
The VM CRD is too big to be applied with `kubectl apply` because it can't fit into the `last-applied-configuration` annotation. Running with `create` instead works though.

### DIFF
--- a/v2/README.md
+++ b/v2/README.md
@@ -61,7 +61,7 @@ Sample YAMLs for creating each of these resources can be found in the [samples d
    ```
 3. Download [the latest **v2+** release](https://github.com/Azure/azure-service-operator/releases) of Azure Service Operator and install it into your cluster.
    ```bash
-   kubectl apply -f azureserviceoperator_v2.0.0-alpha.0.yaml
+   kubectl create -f azureserviceoperator_v2.0.0-alpha.0.yaml
    ```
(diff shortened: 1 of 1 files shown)


babbageclunk:
Example of the problem here: https://stackoverflow.com/a/62409266

v2/README.md
```diff
@@ -61,7 +61,7 @@ Sample YAMLs for creating each of these resources can be found in the [samples d
    ```
 3. Download [the latest **v2+** release](https://github.com/Azure/azure-service-operator/releases) of Azure Service Operator and install it into your cluster.
    ```bash
-   kubectl apply -f azureserviceoperator_v2.0.0-alpha.0.yaml
+   kubectl create -f azureserviceoperator_v2.0.0-alpha.0.yaml
```
theunrepentantgeek, line 64: Do you need to explain why to use `create` and not `apply`?

theunrepentantgeek (approved):
A one word change with one question. Good ratio!

v2/README.md
babbageclunk: I don't think so here - it would be a distraction in the installation instructions. If a user who's a bit more knowledgeable tries using `apply` instead they'll see why we're saying to use `create` - in that case if they then go back to using `create` things will work fine (although they'll get errors that some things already exist).

babbageclunk:
Smash-merged rather than waiting for the build since this was just a readme change.

PR #1949, opened 2021-11-02
Improve the generated versions report to show ARM API versions

theunrepentantgeek:

The supported resources list was showing only our internal version reference, not the matching ARM resource version, which may have been confusing for those new to ASO. 

I've changed it to show the ARM api version (which can therefore be searched for) as well as the CRD version as a note for those who need it.

gif

**If applicable**:

### DIFF
--- a/v2/tools/generator/internal/codegen/pipeline/report_resource_versions.go
+++ b/v2/tools/generator/internal/codegen/pipeline/report_resource_versions.go
@@ -37,12 +37,12 @@ func ReportResourceVersions(configuration *config.Configuration) Stage {
 
 type ResourceVersionsReport struct {
 	// A separate list of resources for each package
-	lists map[astmodel.PackageReference][]string
+	lists map[astmodel.PackageReference][]astmodel.TypeDefinition
 }
 
 func NewResourceVersionsReport(types astmodel.Types) *ResourceVersionsReport {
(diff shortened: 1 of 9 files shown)


v2/tools/generator/internal/codegen/pipeline/report_resource_versions.go
```diff
@@ -98,19 +98,37 @@ func (r *ResourceVersionsReport) WriteToBuffer(buffer *strings.Builder, _ string
...
+		buffer.WriteString(
+			fmt.Sprintf(
+				"\n#### ARM version %s\n\n",
+				strings.Trim(firstResource.APIVersionEnumValue().Value, "\"")))
+
+		// write an alphabetical list of resources
 
 		for _, rsrc := range resources {
 			// Note: These links are guaranteed to work because of the Taskfile 'controller:verify-samples' target
 			path := "github.com/Azure/azure-service-operator/blob/main/v2/"
```
matthchr, line 120: Note: This will conflict with my other PR so we should just be aware.

matthchr (approved):

v2/tools/generator/internal/codegen/pipeline/report_resource_versions.go
theunrepentantgeek: Thanks for the heads up.

PR #1951, opened 2021-11-03
Update docs to use `kubectl apply --server-side=true`

babbageclunk:
We can't use client-side apply because the VirtualMachines CRD is too big to fit in the `last-applied-configuration` annotation kubectl uses, but using `kubectl create` is a bit unpleasant (since it won't update existing resources).

Also note to wait for cert-manager to be started before installing ASO - if we don't do this the ASO webhook-service certificate creation fails because it can't talk to the cert-manager webhook, and the ASO pod won't start. (It bit me while I was testing the various options.)

### DIFF
--- a/v2/README.md
+++ b/v2/README.md
@@ -37,8 +37,10 @@ Sample YAMLs for creating each of these resources can be found in the [samples d
    cert-manager-webhook-c4b5687dc-x66bj      1/1     Running   0          1m
    ```
 
+   (Alternatively, you can wait for cert-manager to be ready with `cmctl check api --wait=2m` - see the [cert-manager documentation](https://cert-manager.io/docs/usage/cmctl/) for more information about `cmctl`.)
+
 2. Create an Azure Service Principal. You'll need this to grant Azure Service Operator permissions to create resources in your subscription.
-   
+
    First, set the following environment variables to your Azure Tenant ID and Subscription ID with your values:
    ```yaml
    AZURE_TENANT_ID=<your-tenant-id-goes-here>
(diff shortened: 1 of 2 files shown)


matthchr:
> Also note to wait for cert-manager to be started before installing ASO - if we don't do this the ASO webhook-service certificate creation fails because it can't talk to the cert-manager webhook, and the ASO pod won't start. (It bit me while I was testing the various options.)

I have a "fix" for this in a branch - will try to get the PR out today.
Fix is to use the `cmctl` tool to wait for cert manager to be up.

v2/README.md
```diff
@@ -27,6 +27,15 @@ Sample YAMLs for creating each of these resources can be found in the [samples d
     ```bash
     kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.1.0/cert-manager.yaml
     ```
+   Check that the cert-manager pods have started successfully before continuing.
```
matthchr, line 30: As mentioned in my other comment you can use `cmctl` here if you want as well: `cmctl check api --wait=2m`

matthchr (approved):

v2/README.md
babbageclunk: Do we want to depend on `cmctl` in our installation instructions? I wouldn't have thought we'd want people to need to install it just for this?

matthchr: I wouldn't want to depend on it, no - but if we want we could mention it as an alternative way to check? I was just thinking something like: "Optionally instead use `cmctl check api --wait=2m`, see X for more details about cmctl"

PR #1953, opened 2021-11-03
Add E2E test infrastructure

### LINKED ISSUE #1941: Tests for managed identity in ASO v2

matthchr:
- Add a single test to this infrastructure to showcase its use.
- E2E tests (in Kind) are run as part of CI on each PR.

Closes #1941.

### DIFF
--- a/.github/workflows/pr-validation-with-secrets.yml
+++ b/.github/workflows/pr-validation-with-secrets.yml
@@ -93,7 +93,7 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      # Note: Changes to this step must also be mirror into pr-validation.yaml
+      # Note: Changes to this step must also be mirrored into pr-validation.yaml
       - name: Build devcontainer image
         # We must issue a manual pull before the build so the image gets copied locally, because
         # docker.pkg.github.com is not a valid Docker registry and doesn't work with --cache-from,
--- a/.github/workflows/pr-validation-with-secrets.yml
+++ b/.github/workflows/pr-validation-with-secrets.yml
@@ -1,0 +1,162 @@
+# This was adapted from https://github.com/imjohnbo/ok-to-test/blob/master/.github/workflows/integration.yml
+name: Validate Pull Request (Integration Tests)
+on:
+  pull_request:
+    branches:
+      - main
+  repository_dispatch:
+    types: [ok-to-test-command]
+
+jobs:
+
+  # Runs on a branch-based pull request
+  integration-trusted:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository
+    permissions:
+      packages: read
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # required to access tags
+          submodules: 'true'
+
+      - name: Log in to GitHub Docker Registry
+        uses: docker/login-action@v1
+        with:
+          registry: docker.pkg.github.com # ghcr.io not yet enabled for Azure org
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      # Note: Changes to this step must also be mirror into pr-validation.yaml
+      - name: Build devcontainer image
+        # We must issue a manual pull before the build so the image gets copied locally, because
+        # docker.pkg.github.com is not a valid Docker registry and doesn't work with --cache-from,
+        # however, `docker pull` will fall back to other methods that do work and get the image loaded.
+        #
+        # This message comes from "docker pull":
+        #
+        # Run docker pull docker.pkg.github.com/azure/azure-service-operator/aso-devcontainer:latest
+        # WARNING: ⚠️ Failed to pull manifest by the resolved digest. This registry does not
+        #  appear to conform to the distribution registry specification; falling back to
+        #  pull by tag.  This fallback is DEPRECATED, and will be removed in a future
+        #  release.  Please contact admins of https://docker.pkg.github.com. ⚠️
+        #
+        # See: https://github.com/moby/moby/issues/41687#issuecomment-733826074 and related issues
+        run: |
+          docker pull docker.pkg.github.com/azure/azure-service-operator/aso-devcontainer:latest
+          docker build --cache-from docker.pkg.github.com/azure/azure-service-operator/aso-devcontainer:latest --tag devcontainer:latest .devcontainer
+        env:
+          DOCKER_BUILDKIT: 1
+
+      - name: Run devcontainer image
+        id: devcontainer
+        run: |
+          container_id=$(docker create -w /workspace -v $GITHUB_WORKSPACE:/workspace -v /var/run/docker.sock:/var/run/docker.sock --network=host devcontainer:latest)
+          docker start "$container_id"
+          echo "::set-output name=id::$container_id"
+
+      - name: Run integration tests
+        run: |
+          container_id=${{steps.devcontainer.outputs.id}}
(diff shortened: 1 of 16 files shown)


matthchr:
I will abandon this after using it to check some things

matthchr:
It looks like this has to merge before I can start testing the forks workflow with "ok-to-test"

.github/workflows/pr-validation-with-secrets.yml
```diff
@@ -0,0 +1,162 @@
...
+          ref: 'refs/pull/${{ github.event.client_payload.pull_request.number }}/merge'
+
+      - name: Log in to GitHub Docker Registry
+        uses: docker/login-action@v1
+        with:
+          registry: docker.pkg.github.com # ghcr.io not yet enabled for Azure org
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      # Note: Changes to this step must also be mirror into pr-validation.yaml
```
babbageclunk, line 96:
```suggestion
      # Note: Changes to this step must also be mirrored into pr-validation.yaml
```

Taskfile.yml
```diff
@@ -261,6 +272,30 @@ tasks:
...
+    dir: "{{.CONTROLLER_ROOT}}"
+    deps: [controller:kind-create-with-podidentity]
+    cmds:
+      # This timeout is purposefully low - if we find that this job is taking a long time we may need to think of other ways
+      # to keep CI fast
+      # TODO: Below pattern was taken from https://github.com/go-task/task/pull/474 to ensure cleanup runs after the test.
+      - |
+        (go test -timeout 15m -run '{{default ".*" .TEST_FILTER}}' ./test); # Run the tests in a subshell
+        TEST_EXIT=$?;
+        kind delete cluster --name=asov2
```
babbageclunk, line 286: Maybe split the cluster deletion out so that it's possible to run `test-integration-kind` without tearing down an existing kind cluster? Although I guess this way prevents having old CRDs accidentally hanging around.

scripts/delete-old-resourcegroups.sh
```diff
@@ -0,0 +1,54 @@
...
+done
+
+
+if [ ! $PREFIX ]; then
+  echo "-p flag is required"
+  print_usage
+  exit 1
+fi
+
+# Verify that the KEEP_AGE is an int
```
babbageclunk, line 35: It might have been easier to write this script in Python? I don't think it's worth rewriting it now though.

v2/internal/testcommon/resource_namer.go
```diff
@@ -18,36 +19,65 @@ type ResourceNameConfig struct {
 	prefix      string
 	randomChars int
 	separator   string
+	mode        ResourceNamerMode
 }
 
+type ResourceNamerMode string
```
babbageclunk, line 25: This seems familiar - was it in the other PR as well?

v2/internal/testcommon/test_context.go
```diff
@@ -54,14 +54,28 @@ type PerTestContext struct {
 	Namespace           string
 }
 
-// If you modify this make sure to modify the cleanup-test-azure-resources target in the Makefile too
+// There are two prefixes here because each represents a resource kind with a distinct lifecycle.
+// If you modify these, make sure to modify the cleanup-azure-resources target in the Taskfile too
+
+// ResourcePrefix is for resources which are used in the record/replay test infrastructure. These tests are not expected to
+// be run live in parallel. In parallel runs should be done from the recordings. As such, for cleanup, we can delete any resources
```
babbageclunk, line 61: Is this going to cause problems with recording (against the ASO-CI sub)? Or do we just need to check that the live envtest action isn't running?

v2/internal/testcommon/test_context.go
```diff
@@ -54,14 +54,28 @@ type PerTestContext struct {
 	Namespace           string
 }
 
-// If you modify this make sure to modify the cleanup-test-azure-resources target in the Makefile too
+// There are two prefixes here because each represents a resource kind with a distinct lifecycle.
```
babbageclunk, line 57: It might be useful to clarify that the former tests are run in an envtest-based cluster while the latter ones are in a real cluster?

babbageclunk (approved):
Looks good, although I'm not at all familiar with github actions so I'm less clear on those.

Taskfile.yml
```diff
@@ -367,6 +413,31 @@ tasks:
...
+    desc: Creates a local kind cluster with ASO using service principal auth.
+    cmds:
+      - task: controller:kind-create
+      - task: controller:install-cert-manager
+      - task: controller:docker-push-local
+      # We need the below to wait until cert-manager is up, otherwise the subsequent installation of webhooks fails. See https://cert-manager.io/next-docs/installation/verify/
+      - "cmctl check api --wait=2m"
+      - task: controller:install
+      - task: controller:make-sp-secret
+      - "kubectl wait --for=condition=ready --timeout=2m pod -n azureserviceoperator-system -l control-plane=controller-manager"
```
theunrepentantgeek, line 426: Does `kubectl wait` return a non-zero exit code if the timeout expires?

theunrepentantgeek (approved):

scripts/delete-old-resourcegroups.sh
matthchr: I was thinking that as I finished writing it... but yea didn't seem worth rewriting.

v2/internal/testcommon/test_context.go
matthchr: This has always been the case, this is just documenting it.

Taskfile.yml
matthchr: yes

.github/workflows/pr-validation-with-secrets.yml
matthchr: Fixed

Taskfile.yml
matthchr: The problem is that if the tests fail, there is no way to write a task in such a way that other tasks "run afterwards". So you have to do the tests + delete in the same task in order to deal with the fail case. In our other tests that's not an issue because there's not a long-running thing (the kind cluster) we want to avoid leaking.

I think at least for the purposes of this PR, keeping it as is makes sense for this target.

I added a separate target `controller:test-integration-kind` and renamed this target to `controller:test-integration-kind-ci`. Also added a note on both to explain what they're for a bit more clearly. I think that accomplishes what you want while maintaining the cleanup step in CI where we need it.

v2/internal/testcommon/resource_namer.go
matthchr: Yes, this PR and the other PR are "the same" except this one isn't from my fork it's from a branch of the Azure repo. Once this is merged I'll open a new PR from my fork to make sure the fork workflow works.

v2/internal/testcommon/test_context.go
matthchr: Added that clarification in the comments onto each `Prefix`

v2/internal/testcommon/test_context.go
matthchr: To answer the actual question, as long as the live envtest isn't running you should be fine - and we moved that to run only at like 4am PST or something so it should always be ok during working hours. We could collide with one another I suppose.

Taskfile.yml
babbageclunk: Yup, that makes sense, thanks!

v2/internal/testcommon/test_context.go
babbageclunk: Ah, I guess the specific thing I was wondering about was for new tests, which would never collide with existing live test runs since the group would be different, but the backstop delete job could remove them while the test was running. But you're right, if they're running in the middle of the night it's not really a worry.

PR #1955, opened 2021-11-03
Improve timeout behaviour in tests

Porges:
At the moment we allow the tests to run as long as possible governed by the overall test timeout (deadline).

Instead, we should have a lower per-operation timeout that can be overridden by the deadline if we are close to hitting it.


v2/internal/testcommon/kube_per_test_context.go
```diff
@@ -246,19 +247,19 @@ func (tc *KubePerTestContext) Subtest(t *testing.T) *KubePerTestContext {
 	return result
 }
 
-// DefaultTimeoutReplaying is the default timeout for a single operation when replaying.
-var DefaultTimeoutReplaying = 2 * time.Minute
+// OperationTimeoutReplaying is the default timeout for a single operation when replaying.
+var OperationTimeoutReplaying = 30 * time.Second
```
matthchr, line 251: We might need to leave this at like 1m (I'm not sure how I picked 2m). If it passes at 30s though then awesome let's use that.

matthchr (approved):

babbageclunk (approved):

v2/internal/testcommon/kube_per_test_context.go
Porges: If it takes 30s to achieve the desired state I think something has definitely gone wrong?

theunrepentantgeek (approved):

v2/internal/testcommon/kube_per_test_context.go
Porges:
```suggestion
var OperationTimeoutReplaying = 2 * time.Minute
```

Porges: Restored to 2 minutes because apparently the Mongo/SQL ones do take a while.

PR #1959, opened 2021-11-04
Improve names for one-of options

Porges:
Remove any common suffix from OneOf property names so that instead of (e.g.):

```yaml
backupPolicy:
    continuousModeBackupPolicy:
        …
```

We now have:

```yaml
backupPolicy:
    continuous:
        …
```


v2/tools/generator/internal/codegen/pipeline/convert_allof_and_oneof_to_objects.go
```diff
@@ -160,7 +160,65 @@ func (s synthesizer) getOneOfPropNames(oneOf *astmodel.OneOfType) ([]propertyNam
 		return err
 	})
 
-	return result, err
+	return simplifyPropNames(result), err
+}
+
+// simplifyPropNames makes the OneOf names better by trimming any common suffix
```
matthchr, line 166: It trims prefixes (finds common suffixes) right?

matthchr (approved):

v2/tools/generator/internal/codegen/pipeline/convert_allof_and_oneof_to_objects.go
Porges: it finds the common suffixes and then trims them off

babbageclunk (approved):

PR #1960, opened 2021-11-05
Improve test reliability

matthchr:
* Use a KubeClient in EnvTest that always hits APIServer to avoid cache latency and inconsistency issues that can cause races and intermittent test failures. See kubernetes-sigs/controller-runtime#1464 and kubernetes-sigs/controller-runtime#343 for details.
 * Write Status after Spec. This ensures that tests waiting for a status update cannot possibly see it so fast that they go on   to perform a write that conflicts with the Spec write of the controller.
 * Improve log messages to be clearer (aids in test debugging).
 * Prior to performing any deletion actions (either monitoring or deletion), verify that the finalizer is still set. If it's not set,
immediately allow the k8s resource to be deleted. Don't take any action on the Azure resource in this instance.

**How does this PR make you feel:**
gif

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/internal/reconcilers/azure_generic_arm_reconciler.go
+++ b/v2/internal/reconcilers/azure_generic_arm_reconciler.go
@@ -378,8 +378,9 @@ func (r *AzureDeploymentReconciler) StartDeleteOfResource(ctx context.Context) (
 	r.log.V(Status).Info(msg)
 	r.recorder.Event(r.obj, v1.EventTypeNormal, string(DeleteActionBeginDelete), msg)
 
-	// If we have no resourceID to begin with, the Azure resource was never created
-	if genruntime.GetResourceIDOrDefault(r.obj) == "" {
+	// If we have no resourceID to begin with, or no finalizer, the Azure resource was never created
+	hasFinalizer := controllerutil.ContainsFinalizer(r.obj, GenericControllerFinalizer)
+	if genruntime.GetResourceIDOrDefault(r.obj) == "" || !hasFinalizer {
 		return ctrl.Result{}, r.deleteResourceSucceeded(ctx)
 	}
 
@@ -427,6 +428,12 @@ func (r *AzureDeploymentReconciler) StartDeleteOfResource(ctx context.Context) (
 // MonitorDelete will call Azure to check if the resource still exists. If so, it will requeue, else,
 // the finalizer will be removed.
 func (r *AzureDeploymentReconciler) MonitorDelete(ctx context.Context) (ctrl.Result, error) {
+	hasFinalizer := controllerutil.ContainsFinalizer(r.obj, GenericControllerFinalizer)
+	if !hasFinalizer {
+		r.log.V(Status).Info("Resource no longer has finalizer, moving deletion to success")
+		return ctrl.Result{}, r.deleteResourceSucceeded(ctx)
+	}
+
 	msg := "Continue monitoring deletion"
 	r.log.V(Verbose).Info(msg)
 	r.recorder.Event(r.obj, v1.EventTypeNormal, string(DeleteActionMonitorDelete), msg)
--- a/v2/internal/reconcilers/azure_generic_arm_reconciler.go
+++ b/v2/internal/reconcilers/azure_generic_arm_reconciler.go
@@ -53,8 +53,8 @@ type CreateOrUpdateAction string
 const (
 	CreateOrUpdateActionNoAction        = CreateOrUpdateAction("NoAction")
 	CreateOrUpdateActionManageOwnership = CreateOrUpdateAction("ManageOwnership")
-	CreateOrUpdateActionBeginCreation   = CreateOrUpdateAction("BeginCreation")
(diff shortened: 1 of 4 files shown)


matthchr:
I ran the full test suite locally 8 times to test this. Tests all passed each of the 8 times, with runtime ranging from 60 to 100s.

matthchr:
So, somehow the recorded tests take 10m to pass in CI... That's still faster than they were but feels slow compared to what I'm seeing locally...
@porges any idea what spec machines we get in actions?

v2/internal/reconcilers/azure_generic_arm_reconciler.go
```diff
@@ -137,7 +137,7 @@ func (r *AzureDeploymentReconciler) Reconcile(ctx context.Context) (ctrl.Result,
 }
 
 func (r *AzureDeploymentReconciler) CreateOrUpdate(ctx context.Context) (ctrl.Result, error) {
-	r.logObj("reconciling resource")
+	r.logObj("reconciling resource", r.obj)
```
theunrepentantgeek, line 140: 😃

v2/internal/reconcilers/azure_generic_arm_reconciler.go
```diff
@@ -553,8 +539,7 @@ func (r *AzureDeploymentReconciler) handlePollerFailed(ctx context.Context, err
 	}
 
 	// We probably just modified spec so don't need to requeue this
-	needRequeue := !hasResumeToken
-	return ctrl.Result{Requeue: needRequeue}, nil
+	return ctrl.Result{Requeue: true}, nil
```
theunrepentantgeek, line 542: Comment says "probably don't need to", but then we always do. Is that just in case?  Perhaps clarify the comment?

v2/internal/testcommon/direct_client.go
```diff
@@ -0,0 +1,86 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+)
+
+type directClient struct {
+	inner client.Client
+}
+
+// NewTestClient is a thin wrapper around controller-runtime client.New, except that we
+// repopulate the objects GVK since for some reason they do that in the cached client and not
+// in the direct one...
```
theunrepentantgeek, line 24: 😲 Only in the cached client! 😲

theunrepentantgeek (commented):
I'm not sure if the GVK changes are safe to have only when running tests; they might make tests run that _should fail_.

v2/internal/reconcilers/azure_generic_arm_reconciler.go
matthchr: I should remove this comment - it no longer applies

v2/internal/testcommon/direct_client.go
matthchr: Yep - no idea why the behavioral different but this direct client is just a thin wrapper that makes the behave the same.

matthchr:
@theunrepentantgeek, you said:
> I'm not sure if the GVK changes are safe to have only when running tests; they might make tests run that should fail.

I'm not 100% sure I follow. AFAIK, the GVK "additions" in the test client are just to make the direct to APIServer (no cache) client behave the same as the other client already does. So it makes tests pass that should pass (without it, they fail with the direct client but pass with the cached client due to the behavior difference).

Can you expand upon what scenarios it might make something pass that wasn't before (with the cached client?)

theunrepentantgeek:
> Can you expand upon what scenarios it might make something pass that wasn't before (with the cached client?)

I had a concern that code dependent on having the GVK would pass in test but fail in production because GVK was only reliably present during testing. If the client used during production runs will also reliably provide the GVK then my concerns are moot.

theunrepentantgeek (approved):

babbageclunk (approved):
This looks good, but I'm worried that the finalizer change might make the tests more stable at the risk of the occasional lost resource.

v2/internal/reconcilers/azure_generic_arm_reconciler.go
```diff
@@ -464,14 +465,14 @@ func (r *AzureDeploymentReconciler) MonitorDelete(ctx context.Context) (ctrl.Res
...
+func (r *AzureDeploymentReconciler) BeginCreateOrUpdateResource(ctx context.Context) (ctrl.Result, error) {
 	armResource, err := r.ConvertResourceToARMResource(ctx)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
-	r.log.V(Status).Info("Deploying new resource to Azure")
+	r.log.V(Status).Info("About to send resource to Azure")
 
-	changed, err := r.AddInitialResourceState(armResource.GetID())
+	err = r.AddInitialResourceState(armResource.GetID())
```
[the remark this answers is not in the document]
matthchr: Nope, you're right. This is probably a longstanding bug we've had. I'll fix this but do it in a different PR as it's pretty unrelated to tests.

PR #1964, opened 2021-11-08
Move finalizer

matthchr:
Avoids possible situation where we have created an Azure resource but not set the finalizer.


babbageclunk (approved):
Nice.

PR #1965, opened 2021-11-09
postgresql user database permissions v1alpha2

morelina:
refs #1346 

The new version of PostgreSQLUser enables to define per-database owner role. It adds the conversions from v1alpha1 to v1alpha2 and Documentation with an example. 

This implements or partially implements #1346 for PostgreSQL. Before, it was only possible to define a server-level permission and it was not possible to grant access to a single database in the server. The program was tested solely for our own use cases, which might differ from yours. 

**If applicable**:

**Legal notice**
Javier Moreno Molina <javier.moreno_molina@daimler.com> on behalf of Mercedes-Benz Connectivity Services GmbH, [Provider Information](https://github.com/Daimler/daimler-foss/blob/master/PROVIDER_INFORMATION.md)


api/v1alpha2/postgresqluser_types.go
```diff
@@ -0,0 +1,76 @@
...
+	// +kubebuilder:validation:Required
+	ResourceGroup string `json:"resourceGroup"`
+
+	// optional
+
+	// The server-level roles assigned ot the user.
+	Roles []string `json:"roles,omitempty"`
+
+	// The databases of which this user should inherit the ownership role.
+	OwnedDatabases []string `json:"ownedDatabases,omitempty"`
```
babbageclunk, line 33: Hi @morelina - thanks for your work on this PR! 

Unfortunately, the way you're handling `ownedDatabases` here isn't compatible with the  way we were intending to do it in #1346, so if this lands as-is we'd eventually need to add another version to do it like that. (Unfortunately we've been focused on the code-generation work for v2 so we haven't gotten to this.)

Would you be willing to change it to work the way the `MySQLUser` type does? Rather than `OwnedDatabases []string`  there'd be `DatabaseRoles map[string][]string` of database name -> roles. You should be able to reuse the diffing code the MySQL reconciler uses to get the set of roles that have been added/removed, although the SQL to read and update those will be Postgres-specific.

babbageclunk (changes requested):

atlslscsrv-app (approved):
signed-off by Laura Vanlaningham

atlslscsrv-app (approved):

theunrepentantgeek:
Are you still interested in progressing this PR? If so, we're willing to merge it into ASO v1 even though that version is technically in maintenance mode as we're putting all our efforts into ASO v2. 

If you don't want to progress this, we'll abandon this PR.

PR #1973, opened 2021-11-15
Generate patches for Kustomize to enable storage versioning

theunrepentantgeek:

Enabling versioning requires the yaml files generated by `controller-gen` to be modified by Kustomize to inject the required settings.

We modify the existing `gen-kustomize` mode of our generator to write the required files.

gif

**Prerequisites**:

- [ ] Merge PR #1985

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -186,7 +186,7 @@ tasks:
   controller:test:
     desc: Run {{.CONTROLLER_APP}} unit tests.
     dir: "{{.CONTROLLER_ROOT}}"
-    deps: [controller:generate-crds]
+    deps: [controller:generate-kustomize]
     cmds:
     - go test -short -tags=noexit -timeout 10m -run '{{default ".*" .TEST_FILTER}}' ./...
 
@@ -257,7 +257,7 @@ tasks:
   controller:test-integration-envtest-cover:
     desc: Run integration tests with envtest using record/replay and output coverage.
     dir: "{{.CONTROLLER_ROOT}}"
-    deps: [controller:generate-kustomize]
+    deps: [controller:pack-crd]
     cmds:
     # -race fails at the moment in controller-runtime
     - go test -covermode atomic -coverprofile=coverage-integration-envtest.out -coverpkg="./..." -timeout 15m -run '{{default ".*" .TEST_FILTER}}' ./internal/controllers
@@ -265,7 +265,7 @@ tasks:
   controller:test-integration-envtest-live:
     desc: Run integration tests with envtest against live data and output coverage.
     dir: "{{.CONTROLLER_ROOT}}"
-    deps: [controller:generate-kustomize, cleanup-azure-resources]
+    deps: [controller:pack-crd, cleanup-azure-resources]
     cmds:
     # -race fails at the moment in controller-runtime
     - go test -timeout {{.LIVE_TEST_TIMEOUT}} -covermode atomic -coverprofile=coverage-integration-envtest-live.out -coverpkg="./..." -run '{{default ".*" .TEST_FILTER}}' ./internal/controllers
@@ -328,6 +328,7 @@ tasks:
     - mkdir -p v2/api
     - find v2/api -type f -name "zz_generated.*" -delete
     - if [ -d "{{.OUTPUT}}/crd/bases" ]; then find "{{.OUTPUT}}/crd/bases" -type f -delete; fi
+    - if [ -d "{{.OUTPUT}}/crd/patches" ]; then find "{{.OUTPUT}}/crd/patches" -type f -delete; fi
     - cd v2/api && controller-gen {{.OBJECT_OPTIONS}} paths=./... 
     - cd v2/api && controller-gen {{.CRD_OPTIONS}} {{.WEBHOOK_OPTIONS}} {{.RBAC_OPTIONS}} paths=./...
     - cd v2/api && gofmt -l -s -w . # format all generated code
@@ -373,8 +374,16 @@ tasks:
     deps: [controller:generate-kustomize]
     dir: "v2/"
     cmds:
+    - mkdir -p config/crd/packed
+    - kustomize build config/default > config/crd/packed/azureserviceoperator.yaml
+
+  controller:pack-crd:
+    desc: Packs the configuration bundle
+    deps: [controller:run-kustomize]
+    dir: "v2/"
+    cmds:
     - mkdir -p bin # in case it doesn’t exist
-    - kustomize build config/default | sed -e 's@localhost:5000/azureserviceoperator:latest@{{.PUBLIC_REGISTRY}}{{.CONTROLLER_DOCKER_IMAGE}}@g' > bin/azureserviceoperator_{{.VERSION}}.yaml
+    - sed config/crd/packed/azureserviceoperator.yaml -e 's@localhost:5000/azureserviceoperator:latest@{{.PUBLIC_REGISTRY}}{{.CONTROLLER_DOCKER_IMAGE}}@g' > bin/azureserviceoperator_{{.VERSION}}.yaml
 
   controller:kind-delete:
     desc: Deletes a kind cluster
@@ -485,6 +494,7 @@ tasks:
       - mkdir -p ./apis
       - find ./apis -type f -name "zz_generated.*" -delete
       - if [ -d "{{.OUTPUT}}/crd/bases" ]; then find "{{.OUTPUT}}/crd/bases" -type f -delete; fi
+      - if [ -d "{{.OUTPUT}}/crd/patches" ]; then find "{{.OUTPUT}}/crd/patches" -type f -delete; fi
       - cd apis && controller-gen {{.OBJECT_OPTIONS}} paths=./...
       - cd apis && controller-gen {{.CRD_OPTIONS}} {{.WEBHOOK_OPTIONS}} {{.RBAC_OPTIONS}} paths=./...
       - cd apis && gofmt -l -s -w . # format all generated code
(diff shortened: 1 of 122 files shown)


Taskfile.yml
```diff
@@ -326,8 +327,16 @@ tasks:
     deps: [controller:generate-kustomize]
     dir: "v2/"
     cmds:
+    - mkdir -p config/crd/packed
+    - kustomize build config/default > config/crd/packed/azureserviceoperator.yaml
```
babbageclunk, line 331: This directory seems like a strange place to put this file - it's not just CRDs, if you're building `config/default` the output will include the deployment, webhook service, RBAC, webhook configurations, etc.

It might make more sense just to run `kustomize build config/crd`, then the output would only be the patched CRDs. (You wouldn't be able to use this file as a precursor to building the full release yaml, but I think that's ok.)

Also, I'm fairly sure building `config/default` will mean that the conversion sections are all pointing at the `azureserviceoperator-system` namespace, which doesn't seem right - the mutating/validating webhook configurations that get included from `../config/webhook` won't have those prefixes.

So maybe leave the `run-kustomize` task the way it was, and change `pack-crds` to just build `config/crd`?

v2/internal/testcommon/kube_test_context_envtest.go
```diff
@@ -37,7 +37,7 @@ func createSharedEnvTest(cfg testConfig, namespaceResources *namespaceResources)
 	environment := envtest.Environment{
 		ErrorIfCRDPathMissing: true,
 		CRDDirectoryPaths: []string{
-			"../../config/crd/bases",
+			"../../config/crd/packed",
```
babbageclunk, line 40: Oh, I guess the benefit of putting the patched CRD output somewhere under `config/crd` is that we can hardcode it here rather than trying to feed it in as a parameter somehow. That makes sense.

v2/tools/generator/gen_kustomize.go
```diff
@@ -32,39 +34,68 @@ func NewGenKustomizeCommand() (*cobra.Command, error) {
 			crdPath := args[0]
 
 			bases := "bases"
-			// We have an expectation that the folder structure is: .../config/crd/bases
+			patches := "patches"
```
babbageclunk, line 37: Make these consts?

v2/tools/generator/internal/kustomization/crd_kustomization_file.go
```diff
@@ -0,0 +1,50 @@
...
+
+// AddResource adds a filepath specifying another CRD definition to include
+func (k *CRDKustomizeFile) AddResource(resourceFilePath string) {
+	k.Resources = append(k.Resources, resourceFilePath)
+}
+
+// AddPatch adds a filepath specifying another CRD patch to include
+func (k *CRDKustomizeFile) AddPatch(patchFilePath string) {
+	k.Patches = append(k.Patches, patchFilePath)
+}
```
babbageclunk, line 35: I was thinking maybe merge these into one method since we always want to add a resource and patch at the same time, but thinking about it I guess that would tie this type to this first use case too much.

babbageclunk (changes requested):
Looks mostly great! I think you need to build `config/crd` rather than `config/default` as it does too much (in particular name/namespace mangling), and then that has knock-on consequences for the tasks.

Taskfile.yml
theunrepentantgeek: I need the output as a file so that it can be picked up by our tests, which expect to point at an existing file.

Should I move this file to a different location or change the target I'm building?

v2/tools/generator/gen_kustomize.go
theunrepentantgeek: Good idea.

v2/tools/generator/internal/kustomization/crd_kustomization_file.go
theunrepentantgeek: Even though it's a single use type, keeping coupling between methods down seems like a good practice.

v2/internal/testcommon/kube_test_context_envtest.go
matthchr, line 40: I'm just curious -- why is this called `packed`? What's packed about it?

Taskfile.yml
matthchr: @babbageclunk just trying to make sure I understand what you are saying... you are saying:

1. Change `run-kustomize` back to the way it was before. This target will run kustomize at the `config/default` level and produce a file containing _everything_ (webhooks, crds, deployment, etc). Primary usage for this target is when producing the official release and/or creating a deployment to deploy into envtest (or a real cluster).
2. Create a new target (@theunrepentantgeek you have `pack-crd` now but maybe it could be called `produce-envtest-crds` or `run-kustomize-envtest` or something). This target would _just_ produce a file containing CRDs + conversion webhooks for use in envtest. Doing this involves running kustomize on just the `config/crd` directory rather than `config/default`?

I did look into the envtest code and confirm that if envtest is passed a file with stuff in it other than CRDs for the `CRDDirectoryPaths` parameter it just ignores everything that isn't a CRD (at least, that was my read of their code).

v2/tools/generator/internal/kustomization/conversion_patch_file.go
```diff
@@ -0,0 +1,92 @@
...
+package kustomization
+
+import (
+	"io/ioutil"
+
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v3"
+)
+
+type ConversionPatchFile struct {
```
matthchr, line 15: Might be worth leaving a comment in here with what the actual generated file looks like (I think it's pretty short?). Same as for our generated code it might help for people to be able to see an example or templatized output

Taskfile.yml
matthchr: If that is what you're saying, I think I like that idea? If we don't do that I think we should rename this `pack-crd` command as all it's really doing is replacing an already generated YAML's docker image reference, so it feels like `pack-crd` should be called something like `prepare-for-release` or somesuch.

babbageclunk: @matthchr Yup, that's what I was saying - sorry it wasn't too clear! 

> I did look into the envtest code and confirm that if envtest is passed a file with stuff in it other than CRDs for the CRDDirectoryPaths parameter it just ignores everything that isn't a CRD (at least, that was my read of their code).

I'm sure this is right, but I think the namespace mangling that is done by the `config/default` kustomize file would still mean the conversion webhooks pointed at the wrong place.

babbageclunk:
I've just realised - we'll also need a `v2/config/crd/kustomizeconfig.yaml`, which needs to be referenced from `v2/config/crd/kustomization.yaml`.  The content of the file can just be the same as v1's `config/crd/kustomizeconfig.yaml`.

This tells kustomize how to find names and namespaces that need to be updated when it applies the mangling (to put `azureserviceoperator-` at the start).

Some documentation links about this:
* https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/
* https://github.com/kubernetes-sigs/kustomize/tree/master/examples/transformerconfigs

v2/internal/testcommon/kube_test_context_envtest.go
theunrepentantgeek: Because all the CRDs are packed into one file, vs the originals where each resource is in a separate file.

I'm renaming the folder to `out` though, to better reflect it's an output of the compilation process.

v2/tools/generator/internal/kustomization/conversion_patch_file.go
theunrepentantgeek: Done.

theunrepentantgeek: Done.

Taskfile.yml
theunrepentantgeek: Moved output into `crd/out` for clarity and made the suggested change.

matthchr:
@theunrepentantgeek I don't think you should need to ok-to-test this PR as it's not from a fork. Is that not what you experienced?

v2/tools/generator/internal/codegen/pipeline/testdata/TestGolden_InjectConvertibleInterface/microsoft.person-v20200101.golden
```diff
@@ -98,6 +107,9 @@
  	}
  	destination.Status = status
  
+ 	// TypeMeta
+ 	destination.TypeMeta = person.TypeMeta
```
babbageclunk, line 111: I thought you needed to avoid doing this, since it ends up copying the GVK from one version to another?

babbageclunk (approved):
This looks good, but I'm not sure about copying the TypeMeta field - isn't that always determined by the package+type that it's on?

Taskfile.yml
```diff
@@ -373,8 +374,17 @@ tasks:
     deps: [controller:generate-kustomize]
     dir: "v2/"
     cmds:
-    - mkdir -p bin # in case it doesn’t exist
-    - kustomize build config/default | sed -e 's@localhost:5000/azureserviceoperator:latest@{{.PUBLIC_REGISTRY}}{{.CONTROLLER_DOCKER_IMAGE}}@g' > bin/azureserviceoperator_{{.VERSION}}.yaml
+      - mkdir -p bing # in case it doesn't exist
```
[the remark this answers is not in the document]
theunrepentantgeek: I think I fixed this with my force push. 😊

v2/tools/generator/internal/codegen/pipeline/testdata/TestGolden_InjectConvertibleInterface/microsoft.person-v20200101.golden
theunrepentantgeek: Yeah, I had changes that I hadn't pushed. Now fixed.

Taskfile.yml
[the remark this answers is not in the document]
matthchr: This doesn't look fixed?

PR #1985, opened 2021-11-17
Copy both ObjectMeta and TypeMeta when doing version conversion

### LINKED ISSUE #1565: Should we make it harder to add properties to Resources?

In PR #1558 we added the ability for Resources to have arbitary properties, over and above the `Spec` and `Status` properties that are standard. 

Adding more properties should be a rare event - we believe that it goes against the usual good-practice convensions for Kubernetes custom resources - but nothing in the code as it stands makes this difficult. 

**Should we add a speed-bump to encourage future maintainers (which includes ourselves) to reconsider before additional properties are included?**

**Describe the improvement**

We could, for example, add a check that only properties of specific names are permitted - with documentation in the code about why extending the list is almost certainly not what you want to do.

theunrepentantgeek:

Includes embedded properties when doing intra-version conversion of resources; copying of `ObjectMeta` is required to maintain the name of the resource.

Also simplifies ResourceType by removing support for arbitrary properties - we've found that we don't need it.

Resolves #1565

gif

### DIFF
--- a/v2/api/microsoft.authorization/v1alpha1api20200801preview/role_assignment_types_gen.go
+++ b/v2/api/microsoft.authorization/v1alpha1api20200801preview/role_assignment_types_gen.go
@@ -222,6 +222,9 @@ func (roleAssignment *RoleAssignment) validateResourceReferences() error {
 // AssignPropertiesFromRoleAssignment populates our RoleAssignment from the provided source RoleAssignment
 func (roleAssignment *RoleAssignment) AssignPropertiesFromRoleAssignment(source *v1alpha1api20200801previewstorage.RoleAssignment) error {
 
+	// ObjectMeta
+	roleAssignment.ObjectMeta = *source.ObjectMeta.DeepCopy()
+
 	// Spec
 	var spec RoleAssignments_Spec
 	err := spec.AssignPropertiesFromRoleAssignmentsSpec(&source.Spec)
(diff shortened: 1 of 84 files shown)


v2/api/microsoft.authorization/v1alpha1api20200801preview/role_assignment_types_gen.go
```diff
@@ -261,6 +270,9 @@ func (roleAssignment *RoleAssignment) AssignPropertiesToRoleAssignment(destinati
 	}
 	destination.Status = status
 
+	// TypeMeta
+	destination.TypeMeta = roleAssignment.TypeMeta
```
matthchr, line 274: `TypeMeta` contains the `Kind` _and_ the `APIVersion`. Is it safe to copy this to a type with a different version? Obviously the `Kind` copy is safe, but the `APIVersion` copy seems suspect since the `APIVersion` of the destination is not the same as the source?

v2/tools/generator/internal/astmodel/object_type.go
```diff
@@ -528,7 +528,7 @@ func (objectType *ObjectType) TestCases() []TestCase {
 	return result
 }
 
-// IsObjectDefinition returns true if the passed definition is for a Arm type; false otherwise.
+// IsObjectDefinition returns true if the passed definition is for an Arm type; false otherwise.
```
matthchr, line 531:
```suggestion
// IsObjectDefinition returns true if the passed definition is for an ARM type; false otherwise.
```
While you're at it?

v2/tools/generator/internal/astmodel/resource_type.go
```diff
@@ -360,38 +358,9 @@ func (resource *ResourceType) EmbeddedProperties() []*PropertyDefinition {
 	}
 }
 
-// WithProperty creates a new ResourceType with another property attached to it
```
matthchr, line 363: Yay deleting code is always awesome

v2/api/microsoft.authorization/v1alpha1api20200801preview/role_assignment_types_gen.go
theunrepentantgeek: Indeed! I had changes to only copy `ObjectMeta` that I hadn't pushed. I've rebased and updated.

v2/tools/generator/internal/astmodel/object_type.go
theunrepentantgeek: Done.

matthchr (approved):

PR #1995, opened 2021-11-23
Remove namespace from ResourceReference

matthchr:
As per https://github.com/kubernetes/community/pull/5455, it is not a recommended pattern.

### DIFF
--- a/v2/pkg/genruntime/resource_reference.go
+++ b/v2/pkg/genruntime/resource_reference.go
@@ -55,8 +55,6 @@ type ResourceReference struct {
 	Group string `json:"group,omitempty"`
 	// Kind is the Kubernetes kind of the resource.
 	Kind string `json:"kind,omitempty"`
-	// Namespace is the Kubernetes namespace of the resource.
-	Namespace string `json:"namespace,omitempty"`
 	// Name is the Kubernetes name of the resource.
 	Name string `json:"name,omitempty"`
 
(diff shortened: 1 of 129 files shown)


v2/pkg/genruntime/resource_reference.go
```diff
@@ -111,6 +109,13 @@ func (ref ResourceReference) Validate() error {
 	return nil
 }
 
+func (ref ResourceReference) ToNamespacedRef(namespace string) NamespacedResourceReference {
```
babbageclunk, line 112: Maybe add a test for this? I guess it's covered by other things though.

v2/tools/generator/internal/interfaces/kubernetes_resource_interface.go
```diff
@@ -400,7 +400,7 @@ func createResourceReference(
 			"ResourceReference"))
 	compositeLitDetails.AddField("Group", group)
 	compositeLitDetails.AddField("Kind", kind)
-	compositeLitDetails.AddField("Namespace", astbuilder.Selector(dst.NewIdent(receiverIdent), "Namespace"))
+	// compositeLitDetails.AddField("Namespace", astbuilder.Selector(dst.NewIdent(receiverIdent), "Namespace"))
```
babbageclunk, line 403: This can go now - presumably left from initial steps.

babbageclunk (approved):
Nice!

v2/tools/generator/internal/interfaces/kubernetes_resource_interface.go
matthchr: Oh, yep - good catch

v2/pkg/genruntime/resource_reference.go
matthchr: Good prodding to add test, found some awkward behavior (not really a bug but confusing). Clarified it, added comment, and added test.

PR #2002, opened 2021-11-29
Don’t generate empty names

Porges:
Trimming common suffixes can generate an empty name if one name is the suffix of all others. Detect and handle this case.


theunrepentantgeek (approved):
Good tweak. 

I'm wondering whether we should also trim any common prefix from the property names?

matthchr (approved):

PR #2006, opened 2021-12-01
Fix bug with float enums

matthchr:
Previously we would generate code that didn't compile for this case (can't cast string to float).


v2/tools/generator/internal/jsonast/jsonast.go
```diff
@@ -451,7 +451,7 @@ func enumHandler(ctx context.Context, scanner *SchemaScanner, schema Schema) (as
 		vTrimmed := strings.Trim(v, "\"")
 
 		// Some specs include boolean enums with quotes around the boolean literals
-		if baseType == astmodel.BoolType {
+		if baseType == astmodel.BoolType || baseType == astmodel.FloatType {
```
theunrepentantgeek, line 454: Suggest always trimming quotes _unless_ the baseType is string (e.g. int enums would also need trimming). 
Also need to update the comment.

theunrepentantgeek (approved):

v2/tools/generator/internal/jsonast/jsonast.go
matthchr: Oops, I thought about the comment but forgot to fix it. Will do as you say.

matthchr: Fixed!

Issue #1400, opened 2021-03-29
Feature: Azure Cache for Redis,  Enterprise Tiers support


dashaun:
Basic, Standard and Premium are currently supported.  Redis Enterprise tiers were public preview October 2020, and GA in March 2021.

The enterprise tiers are represented as:
```      "type": "Microsoft.Cache/redisEnterprise"```

Active-Active feature is currently in public preview.

matthchr:
We'll make sure to add this to our roadmap for codegenerated resources (see #1192).

PR #2008, opened 2021-12-06
Add support for Redis cache resources

### LINKED ISSUE #1225: Add support for Redis update schedule
Redis is a supported CRD but it is missing the update schedule property

**Describe the improvement**
Add support for the update schedule property.

babbageclunk:
Part of #1400
Closes #1225 

This adds the following resources:
* Redis
* RedisFirewallRule
* RedisLinkedServer
* RedisPatchSchedule

The Redis Enterprise resources will be added in a followup PR.

The RedisConfiguration field is converted to a `map[string]string` - in the resource manager JSON it's an object with some specified fields and `additionalProperties: {type: "object"}` which we don't support.
The RP responds with Conflict errors with "try again later" in the message - I've added some special-case handling that treats those as retryable errors rather than dropping the resource into a terminal state.

gif

**If applicable**:
- [x] this PR contains tests


Porges (approved):

v2/internal/reconcilers/error_classifier.go
```diff
@@ -86,10 +88,20 @@ func classifyInnerCloudError(err *genericarmclient.ErrorResponse) CloudErrorClas
 		"ResourceQuotaExceeded",
 		"SubscriptionNotRegistered":
 		return CloudErrorRetryable
+	case "Conflict":
+		// Conflict should really be fatal, but there are some
+		// services that include a message that we should try again
+		// later. In that case classify it as retryable.
```
theunrepentantgeek, line 94: This is one of those things that _works_, but really leaves a sour taste in the back of the throat.

theunrepentantgeek (approved):

v2/internal/reconcilers/error_classifier.go
babbageclunk: Yeah definitely - once we change things allow customising classification at the resource level this will be a bit better contained.

PR #2010, opened 2021-12-08
Improve error reporting

theunrepentantgeek:

Allows for property conversions to report a fatal error when they find something that just can't be handled. 
Adds error reporting to the object model configuration when desired configuration is not found.

Both of these are to give end users more information about problems so they can fix things up.

Also tweaks our report output for clarity.

These changes are prerequisites for my WIP branch to for type renaming.

gif

**If applicable**:
- [x] this PR contains tests


v2/tools/generator/internal/codegen/pipeline/add_cross_resource_references.go
```diff
@@ -36,14 +36,18 @@ func AddCrossResourceReferences(configuration *config.Configuration, idFactory a
 			var isCrossResourceReferenceErrs []error
 
 			isCrossResourceReference := func(typeName astmodel.TypeName, prop *astmodel.PropertyDefinition) bool {
-				isReference, found := configuration.ARMReference(typeName, prop.PropertyName())
-				if DoesPropertyLookLikeARMReference(prop) && !found {
+				isReference, err := configuration.ARMReference(typeName, prop.PropertyName())
+				if DoesPropertyLookLikeARMReference(prop) && err != nil {
```
babbageclunk, line 40: I realise the old version of this had no way to return errors anyway so we know that any error is actually going to be a not-found error, but for future-proofing it might be better to have this check for a specific not-found error type rather than just not-nil. Really just so the wrapped context won't be misleading if a different error case is introduced somewhere underneath `configuration.ARMReference`.

v2/tools/generator/internal/config/group_configuration.go
```diff
@@ -91,10 +108,25 @@ func (gc *GroupConfiguration) Add(version *VersionConfiguration) *GroupConfigura
...
+		// as all our configuration is based on API versions, not storage versions
+		ref = s.Local()
+	}
+
+	v := strings.ToLower(ref.PackageName())
+	if version, ok := gc.versions[v]; ok {
+		return version, nil
+	}
+
+	return nil, errors.Errorf(
```
babbageclunk, line 124: This is the error I'd change to be a specific NotFoundError type.

v2/tools/generator/internal/config/group_configuration.go
```diff
@@ -91,10 +108,25 @@ func (gc *GroupConfiguration) Add(version *VersionConfiguration) *GroupConfigura
...
+		"configuration of group %s has no detail for version %s (configured versions are: %s)",
+		gc.name,
+		ref.PackageName(),
+		strings.Join(gc.configuredVersions(), ", "))
```
babbageclunk, line 128: If you're a bit worried about the performance impact of doing this repeatedly, I think the name and package name that it's trying to look up are the most useful parts here. The configured versions are easy enough to see in the file, it's the name of the type (as it is in this specific part of the pipeline between transformations) that's hard to determine when running the generator.

v2/tools/generator/internal/config/group_configuration_test.go
```diff
@@ -51,8 +51,8 @@ func TestGroupConfiguration_TypeRename_WhenTypeFound_ReturnsExpectedResult(t *te
 	group := NewGroupConfiguration(test.Group).Add(version2015)
 
 	typeName := astmodel.MakeTypeName(test.Pkg2020, "Person")
-	name, ok := group.TypeRename(typeName)
-	g.Expect(ok).To(BeTrue())
+	name, err := group.TypeRename(typeName)
+	g.Expect(err).To(Succeed())
```
babbageclunk, line 55: Ooh, I should be using `Succeed()` rather than `.[Not]To(BeNil())`.

v2/tools/generator/internal/config/object_model_configuration.go
```diff
@@ -81,18 +81,22 @@ func (omc *ObjectModelConfiguration) Add(group *GroupConfiguration) *ObjectModel
...
-		return nil, false
+	if g, ok := omc.groups[group]; ok {
+		return g, nil
 	}
 
-	return g, true
+	return nil, errors.Errorf(
+		"no configuration for group %s (configured groups are: %s)",
+		group,
+		strings.Join(omc.configuredGroups(), ", "))
```
babbageclunk, line 99: Same points here.

v2/tools/generator/internal/config/type_configuration.go
```diff
@@ -79,18 +88,26 @@ func (tc *TypeConfiguration) FindUnusedARMReferences() []string {
...
 	// Store the property id using lowercase,
 	// so we can do case-insensitive lookups later
 	p := strings.ToLower(string(property))
-	pc, ok := tc.properties[p]
-	return pc, ok
+	if pc, ok := tc.properties[p]; ok {
+		return pc, nil
+	}
+
+	return nil, errors.Errorf(
```
babbageclunk, line 106: Same here.

babbageclunk (approved):
Looks good - I think it would be worth adding a specific not-found error type to avoid misleading wrapped context further up the stack,

v2/tools/generator/internal/codegen/pipeline/add_cross_resource_references.go
theunrepentantgeek: Done.

v2/tools/generator/internal/config/group_configuration.go
theunrepentantgeek: Done.

v2/tools/generator/internal/config/group_configuration.go
theunrepentantgeek: I found the list of options very useful when debugging my WIP branch. What I've done with the custom error type is to defer the sort/join until the error is rendered, which mitigates much of the cost.

v2/tools/generator/internal/config/object_model_configuration.go
theunrepentantgeek: See `NotConfiguredError`.

v2/tools/generator/internal/config/type_configuration.go
theunrepentantgeek: See `NotConfiguredError`.

PR #2018, opened 2021-12-19
Add `inject-ca-from` annotation to generated CRD patches

babbageclunk:
The `cert-manager.io/inject-ca-from` annotation on a CRD tells the cert-manager webhooks to add the CA bundle to the conversion webhook client config. Without this the conversion webhooks will fail and the only resource that will be able to be created will be resource groups (since they don't have multiple versions). This was missed because our integration test and manual post-release check only tested resource groups 😢. This has now been rectified! 

Envtests were working because they manage the CA bundle themselves when the CRD is being installed.

gif

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -296,7 +296,7 @@ tasks:
     cmds:
       # This timeout is purposefully low - if we find that this job is taking a long time we may need to think of other ways
       # to keep CI fast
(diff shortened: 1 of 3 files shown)


theunrepentantgeek (approved):
Looks good to me.

gif

PR #2027, opened 2022-01-10
Create default merger helper method for status merging

matthchr:
The default merger deals with the common cases so that augmenters can
focus on the specific cases they want to handle.


matthchr:
Doing this as a prerequisite step before adding an augmenter that injects `x-ms-secret` and `format: secret` handling

Porges (approved):

PR #2034, opened 2022-01-11
Simplify createARMTypes method

matthchr:
Simplified by breaking out property specific handling
into their own handlers. This paves the way for future handlers as
well, which will be required for supporting secrets.

### DIFF
--- a/v2/tools/generator/internal/codegen/pipeline/create_arm_types.go
+++ b/v2/tools/generator/internal/codegen/pipeline/create_arm_types.go
@@ -39,6 +39,8 @@ func CreateARMTypes(idFactory astmodel.IdentifierFactory) Stage {
 		})
 }
 
+type armPropertyTypeConversionHandler func(prop *astmodel.PropertyDefinition, isSpec bool) (*astmodel.PropertyDefinition, error)
+
 type armTypeCreator struct {
 	definitions astmodel.Types
 	idFactory   astmodel.IdentifierFactory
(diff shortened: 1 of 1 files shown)


v2/tools/generator/internal/codegen/pipeline/create_arm_types.go
```diff
@@ -225,75 +227,106 @@ func (c *armTypeCreator) convertARMPropertyTypeIfNeeded(t astmodel.Type) (astmod
...
-			continue
+		for _, handler := range embeddedPropertyHandlers {
+			newProp, err := handler(prop, isSpecType)
+			if err != nil {
+				errs = append(errs, err)
+				break // Stop calling handlers and proceed to the next property
+			}
+
+			if newProp != nil {
+				result, err = result.WithEmbeddedProperty(newProp)
```
theunrepentantgeek, line 324: Wouldn't it be more efficient to update the property outside the loop, after all the property handlers have made their change?

theunrepentantgeek: Actually, every handler is called with `prop` (initialized on line 315) - but that's never updated from `newProp`, so if there are multiple handlers the results of the earlier ones are discarded and overwritten.

theunrepentantgeek (changes requested):
I think I found a bug.

v2/tools/generator/internal/codegen/pipeline/create_arm_types.go
matthchr: Hmm, that doesn't make it a ton clearer either at least for me.

Maybe I should just put a comment on the break inside `newProp != nil` that says:
`// Once the first property has been matched, stop.`

We have similar code to this elsewhere and that's the pattern it follows - for example see [`renamer.go` line 177](https://github.com/Azure/azure-service-operator/blob/1f708876c2d9d906c93c788781e2a4ea739a92e6/v2/tools/generator/internal/codegen/embeddedresources/renamer.go#L177), or [`property_conversions.go` line 158](https://github.com/Azure/azure-service-operator/blob/main/v2/tools/generator/internal/conversions/property_conversions.go#L163) (which has a return rather than a break but otherwise is similar).

matthchr: I've put the comments, see what you think

theunrepentantgeek: Works for me. :+1:

theunrepentantgeek (approved):

PR #2035, opened 2022-01-12
Specification for Reconciler Extension points

theunrepentantgeek:

We are discovering inconsistencies in the way different Azure Resource Providers behave. 

This ADR proposes a way we can provide extension points allowing resource providers to customize behaviour in a controlled fashion.

@matthchr suggested using the approach of extending resource registration, which is a lot cleaner than the `init` based approach I had in mind originally.

gif

**If applicable**:

### DIFF
--- a/docs/hugo/content/architecture-decision-records/ADR-2022-01-Reconciler-Extensions.md
+++ b/docs/hugo/content/architecture-decision-records/ADR-2022-01-Reconciler-Extensions.md
@@ -1,0 +1,106 @@
+# Reconciler Extensions
+
+## Context
+
+We are discovering inconsistencies in the way different Azure Resource Providers behave.
+
+These inconsistencies are anticipated. Given the number of disparate teams independently implementing their providers and given the way the ARM guidance has changed over time, differences in behaviour are to be expected.
+
+For example, the Azure Redis Resource Provider can return a HTTP conflict (409) with a retry message indicating that it's not a fatal error but a transient one.
+
+Currently we handle this with special case behaviour in `error_classifier.go` (see [PR#2008](https://github.com/Azure/azure-service-operator/pull/2008) for details).
+
+However, directly changing our generic reconciler is a problem. The growing complexity makes this unsustainable, and there is a risk that changes made to support a new resource type will break an existing one.
+
+We need a way to customize its behaviour on a per-resource basis, allowing us to maintain consistent functionality within the operator.
+
+When a new version of a resource is introduced, we don't want to lose any existing customizations. This rules out hosting the extensions directly on the current hub type.
+
+It's quite likely that some extension points will need to import our generated resource types to allow per-version manipulation. This rules out referencing extensions from the generated resource types.
+
+We don't want extension points to reimplement functionality already present in the generic reconciler as this runs the risk of inconsistencies (especially if we change the generic reconciler itself), and introduces a significant overhead to implementation. Instead, the extension points will provide an opportunity to modify the information already held. An extension point that does nothing should have the same effect as one that hasn't been implemented.
+
+## Decision
+
+To provide a stable implementation location for extensions, we'll introduce a dedicated package alongside the existing versioned packages for each supported service. This package will provide extension support for all supported resource versions. For example, alongside the existing **batch** packages `v1alpha1api20210101`, and `v1alpha1api20210101storage`, we will generate `extensions`.
+
+Into the `extensions` package we will codegen a skeleton framework. For each resource there will be a separate extension type, suffixed with `Extensions`. For example, for the **compute** resources `Disk`, `VirtualMachine`, and `VirtualMachineScaleSet` we will create the extension types `DiskExtensions`, `VirtualMachineExtensions`, and `VirtualMachineScaleSetExtensions`.
+
+We'll enhance the existing registration file (`controller_resources_gen.go`) to also provide registration of extensions. They will be indexed by the GVK of each resource, allowing easy lookup by the generic reconciler.
+
+For each supported extension, we'll define an interface. Implementers will assert type compatibility with this extension interface to ensure they have the correct method signature. The generic reconciler will identify available extensions by testing for the presence of the interface.
+
+### Example
+
+To allow customization of error handling for Azure Redis, we'll introduce an extension point, which would look something like this:
+
+We create the required extension interface in `genruntime`:
+
+``` go
+type ErrorClassifierExtension interface {
+    // Classify the provided error.
+    // cloudError is the error returned from ARM.
+    // details contains the initial analysis by the ASO reconciler, and may be changed if desired.
+    // apiVersion is the ARM API version used for the request
+    ClassifyError(cloudError *genericarmclient.CloudError, details *CloudErrorDetails, apiVersion string)
+}
+```
+
+In the `cache/extensions` package, the new extension point will be implemented:
+
+``` go
+var _ ErrorClassifierExtension = &RedisExtensions{}
+
+func (e *RedisExtensions) ClassifyError(
+    cloudError *genericarmclient.CloudError, 
+    details *CloudErrorDetails, 
+    apiVersion string) {
+
+    inner := cloudError.InnerError
+    if inner == nil {
+        return
+    }
+
+    code := to.String(inner.Code)
+    if code == "Conflict" {
+        // For Redis, Conflict can be retried as it can occur because something doesn't yet exist
+        details.Classification = CloudErrorRetryable
+    }
+}
+```
+
+This code goes in a different file so that it's not obliterated next time we rerun the generator.
+
+In the generic reconciler, we make use of the extension:
+
+``` go
+func (r *AzureDeploymentReconciler) makeReadyConditionFromError(
+    cloudError *genericarmclient.CloudError) conditions.Condition {
+
+	var severity conditions.ConditionSeverity
+	errorDetails := ClassifyCloudError(cloudError)
+
+    ext := r.FindErrorClassifierExtension()
+    ext.ClassifyError(cloudError, &errorDetails, apiVersion /* found where? */)
+
+    // ... existing code elided ...
+}
+```
+
+The helper method `FindErrorClassifierExtension()` does the lookup for the extension, and returns a null implementation if not found, so the consuming code doesn't need to worry about null checking.
+
+## Status
+
+Proposed.
+
+## Consequences
+
+TBC.
+
+## Experience Report
+
+TBC.
+
+## References
+
+
--- a/docs/hugo/content/architecture-decision-records/ADR-2022-01-Reconciler-Extensions.md
+++ b/docs/hugo/content/architecture-decision-records/ADR-2022-01-Reconciler-Extensions.md
@@ -83,6 +83,8 @@ func (e *RedisExtensions) ClassifyError(
 }
 ```
 
+Passing in the default behavour as `next` allows extensions to augmenet and/or supplant the behaviour with a great deal of flexibility.
+
 In the generic reconciler, we make use of the extension:
 
 ``` go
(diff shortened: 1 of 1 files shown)


docs/hugo/content/architecture-decision-records/ADR-2022-01-Reconciler-Extensions.md
```diff
@@ -0,0 +1,106 @@
...
+
+We are discovering inconsistencies in the way different Azure Resource Providers behave.
+
+These inconsistencies are anticipated. Given the number of disparate teams independently implementing their providers and given the way the ARM guidance has changed over time, differences in behaviour are to be expected.
+
+For example, the Azure Redis Resource Provider can return a HTTP conflict (409) with a retry message indicating that it's not a fatal error but a transient one.
+
+Currently we handle this with special case behaviour in `error_classifier.go` (see [PR#2008](https://github.com/Azure/azure-service-operator/pull/2008) for details).
+
+However, directly changing our generic reconciler is a problem. The growing complexity makes this unsustainable, and there is a risk that changes made to support a new resource type will break an existing one.
```
matthchr, line 13: minor: complexity and number of resources

docs/hugo/content/architecture-decision-records/ADR-2022-01-Reconciler-Extensions.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Decision
+
+To provide a stable implementation location for extensions, we'll introduce a dedicated package alongside the existing versioned packages for each supported service. This package will provide extension support for all supported resource versions. For example, alongside the existing **batch** packages `v1alpha1api20210101`, and `v1alpha1api20210101storage`, we will generate `extensions`.
+
+Into the `extensions` package we will codegen a skeleton framework. For each resource there will be a separate extension type, suffixed with `Extensions`. For example, for the **compute** resources `Disk`, `VirtualMachine`, and `VirtualMachineScaleSet` we will create the extension types `DiskExtensions`, `VirtualMachineExtensions`, and `VirtualMachineScaleSetExtensions`.
```
matthchr, line 27: Is the advantage of pre-generating some things into `extensions` that we don't have to manually register things in the controller when an extension is added?

If we don't want to actually do the code generator work, we could instead just say that users must implement an `extensions` package (in that location) when they want to include an extension, and then register the extension in the `controller_resources.go` file (as opposed to the `controller_resources_gen.go` which is the generated one).

This has a few advantages in my head:
1. It avoids us having to generate templates, we could instead just write an example one manually, which is easier.
2. It avoids producing a bunch of code that does nothing - because an empty extension is the same as no extension at all. Much of the generated boilerplate will actually not be used in lots of packages where there is no need for an extension.
3. It avoids the "some of this package is generated and some of it is handcrafted" problem - it would mean that _everything_ in `extensions` is handcrafted. If we do generate templates we'll need to be careful about which files are generated and make sure to leave handcrafted stuff alone. That's fine but it places some restrictions on file layout.

docs/hugo/content/architecture-decision-records/ADR-2022-01-Reconciler-Extensions.md
```diff
@@ -0,0 +1,106 @@
...
+    ClassifyError(cloudError *genericarmclient.CloudError, details *CloudErrorDetails, apiVersion string)
+}
+```
+
+In the `cache/extensions` package, the new extension point will be implemented:
+
+``` go
+var _ ErrorClassifierExtension = &RedisExtensions{}
+
+func (e *RedisExtensions) ClassifyError(
```
matthchr, line 54: What's the actual shape of `RedisExtensions` - does it have any properties? If not it does seem to lend itself quite well to what I was talking about above of just handcrafting the whole `extensions` - since the bit that's being autogenerated is basically an empty type anyway.

docs/hugo/content/architecture-decision-records/ADR-2022-01-Reconciler-Extensions.md
```diff
@@ -0,0 +1,106 @@
...
+In the generic reconciler, we make use of the extension:
+
+``` go
+func (r *AzureDeploymentReconciler) makeReadyConditionFromError(
+    cloudError *genericarmclient.CloudError) conditions.Condition {
+
+	var severity conditions.ConditionSeverity
+	errorDetails := ClassifyCloudError(cloudError)
+
+    ext := r.FindErrorClassifierExtension()
```
matthchr, line 83: I like this not having to deal with "didn't find" or `nil`. Just a no-op method if no extension found makes usage clean.

docs/hugo/content/architecture-decision-records/ADR-2022-01-Reconciler-Extensions.md
theunrepentantgeek: I want to make the overhead of writing an extenstion as low as possible, primarily so that it's easier for future maintainers to do the right thing than to start hacking around in the generic reconciler. Reducing the number of things that need to be known (or discovered) to correctly add an extension is key.

Generating the framework also allows us to constrain things. For example, by generating a host type, we can prevent it from having any state - which neatly sidesteps any issues caused by concurrent calls or extensions sharing state.

Doing less work is an attractive option though!

Leaving handcrafted stuff alone is easy - we already have the protocol of generating `_gen.go` files, and we only delete them if they have our header line in place. We'd generate a `VirtualMachineExtensions_gen.go` file, and any manually written code would go into `VirtualMachineExtensions.go`.

docs/hugo/content/architecture-decision-records/ADR-2022-01-Reconciler-Extensions.md
```diff
@@ -0,0 +1,106 @@
...
+
+We'll enhance the existing registration file (`controller_resources_gen.go`) to also provide registration of extensions. They will be indexed by the GVK of each resource, allowing easy lookup by the generic reconciler.
+
+For each supported extension, we'll define an interface. Implementers will assert type compatibility with this extension interface to ensure they have the correct method signature. The generic reconciler will identify available extensions by testing for the presence of the interface.
+
+### Example
+
+To allow customization of error handling for Azure Redis, we'll introduce an extension point, which would look something like this:
+
+We create the required extension interface in `genruntime`:
```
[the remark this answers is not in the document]
theunrepentantgeek: I like the idea of having the extension interfaces in a separate namespace, as it improves discovery. If we call it `genruntime/extensions` though, we probably don't want to call the other packages `extensions` too.

docs/hugo/content/architecture-decision-records/ADR-2022-01-Reconciler-Extensions.md
matthchr: > Leaving handcrafted stuff alone is easy

Yeah fair enough.

I guess what I am getting at is that we should look at what generating is giving us, and unless there's a lot of required boilerplate to get an extension working consider avoiding generating them. At least with my current understanding of the proposal I don't really see a ton of boilerplate, there's two parts:
1. Must create a type that implements the proposed extension interface
2. Must register that type with the controller

The actual implementation for item 1 has to be handcrafted (I think?), so code generation would just be making 2 easier - but I'm imagining 2 is basically a one-liner saying: `extensions.Add(&MySQLFlexibleServer{}, &MySQLFlexibleServerExtension{})`?

matthchr (approved):

docs/hugo/content/architecture-decision-records/ADR-2022-01-Reconciler-Extensions.md
```diff
@@ -0,0 +1,123 @@
...
+}
+```
+
+In the `cache/extensions` package, a host type for the extension point will be generated in `redis_extensions_gen.go`:
+
+```
+type RedisExtensions struct{}
+```
+
+**TODO**: Should this have any members?
```
matthchr, line 58: Probably not? They're really just containers for extension methods right? Probably clearer to pass whatever data we need in the method signatures than through here, I would think.

docs/hugo/content/architecture-decision-records/ADR-2022-01-Reconciler-Extensions.md
```diff
@@ -0,0 +1,123 @@
...
+        // handle error
+    }
+
+    // ... existing code elided ...
+}
+```
+
+The helper method `FindErrorClassifierExtension()` does the lookup for the extension, and returns a null implementation if not found, so the consuming code doesn't need to worry about null checking.
+
+**TODO**: Should we actually return a wrapper around the extension that provides logging and other useful features?
```
matthchr, line 107: Might need to pass some "standard stuff" to the extensions, for example logging, Kubernetes client, Azure client, etc.

Not sure if that's a "wrapper" per-se, but some sort of common `ExtensionEssentials` datatype holding `log`, `armClient`, `kubeClient`, etc?

docs/hugo/content/architecture-decision-records/ADR-2022-01-Reconciler-Extensions.md
matthchr: Unless we have some sort of `MakeExtensionHandler` interface that produces these and _thats_ what we're registering?

docs/hugo/content/architecture-decision-records/ADR-2022-01-Reconciler-Extensions.md
theunrepentantgeek: I was more thinking that the implementation of `ErrorClassifierExtension` returned would be one that does logging itself, with the `RedisExtensions` instance wrapped inside it; that way individual extensions don't need to worry about logging.

Would it ever be sensible for clients to be passed to an extension? They're supposed to tweak the reconciler's behaviour, not provide an opportunity for doing something completely new.

docs/hugo/content/architecture-decision-records/ADR-2022-01-Reconciler-Extensions.md
theunrepentantgeek: I'm not seeing any need for members at this point. By generating the type ourselves, we preclude extensions from having persistent state, which I think is a useful restriction to avoid subtle and hard to diagnose failure modes.

docs/hugo/content/architecture-decision-records/ADR-2022-01-Reconciler-Extensions.md
theunrepentantgeek: Agreed.

docs/hugo/content/architecture-decision-records/ADR-2022-01-Reconciler-Extensions.md
matthchr: If we wanna do secrets via the extension mechanism they'll need both an Azure client and a k8s client, because they will have to:

1. Call the Azure `GetSecrets API`
2. Write the k8s (or KeyVault or whatever) secret.

We could do the secrets stuff in its own special extension though, and deny others access to the clients. I'm not sure if there are other situations where we'll want access to one or the other clients in extensions - do we have a catalog of the customizations we know we'd like to perform? We could look through it and see if they're all simple or if some are going to require calls to Azure.

docs/hugo/content/architecture-decision-records/ADR-2022-01-Reconciler-Extensions.md
theunrepentantgeek: Reworded.

docs/hugo/content/architecture-decision-records/ADR-2022-01-Reconciler-Extensions.md
theunrepentantgeek: Each extension point has a custom set of parameters, so we can pass those clients through for that extension point without offering them anywhere else. I've clarified the spec to make this clearer.

theunrepentantgeek: Makes sense. I've clarified in the spec that each extension is custom designed.

PR #2040, opened 2022-01-19
Add preamble to report

theunrepentantgeek:

Adds a preamble to our supported resources report to make it work better as a page in our documentation.

Also tweaks formatting; previously we were using level 3 & 4 headers, now uses level 2 & 3.

gif

The result looks like this:
image

### DIFF
--- a/v2/tools/generator/internal/codegen/pipeline/report_resource_versions.go
+++ b/v2/tools/generator/internal/codegen/pipeline/report_resource_versions.go
@@ -77,6 +77,11 @@ func (r *ResourceVersionsReport) WriteTo(outputPath string, samplesURL string) e
 
 // WriteToBuffer creates the report in the provided buffer
 func (r *ResourceVersionsReport) WriteToBuffer(buffer *strings.Builder, samplesURL string) {
+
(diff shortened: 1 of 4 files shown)


v2/tools/generator/internal/codegen/pipeline/report_resource_versions.go
```diff
@@ -77,6 +77,11 @@ func (r *ResourceVersionsReport) WriteTo(outputPath string, samplesURL string) e
 
 // WriteToBuffer creates the report in the provided buffer
 func (r *ResourceVersionsReport) WriteToBuffer(buffer *strings.Builder, samplesURL string) {
+
+	buffer.WriteString("# Supported Resources\n\n")
+	buffer.WriteString("This page lists all resources currently supported by Azure Service Operator, grouped by the originating ARM service.\n\n")
```
matthchr, line 82: Should we clarify what exactly `current` means? In this context it's actually what is supported in `master` rather than at a particular release - so it's possible there are items in this list that aren't really usable yet (because we haven't released since they were added?)

matthchr (approved):

PR #2042, opened 2022-01-19
Use SecretReference instead of string for secrets

matthchr:
* Swagger data is used to determine which fields are secrets.
 * New SecretResolver resolves SecretReference's into secret values.
 * Those secret values are passed to ConvertToARM. The result is
   that on the Kubernetes type the user specifies a SecretReference
   and on the ARM type the field is still a string which is the resolved
   secret value.

**If applicable**:
- [x] this PR contains tests


v2/tools/generator/internal/armconversion/convert_to_arm_function_builder.go
```diff
@@ -387,6 +388,40 @@ func (builder *convertToARMBuilder) convertReferenceProperty(_ *astmodel.Convers
...
+	if !isString {
+		return nil
+	}
+
+	isReference := astmodel.TypeEquals(params.SourceType, astmodel.SecretReferenceType)
+	if !isReference {
+		return nil
+	}
+
+	localVarName := builder.idFactory.CreateIdentifier(params.NameHint+"Secret", astmodel.NotExported)
```
theunrepentantgeek, line 409: With my recently merged changes to receiver names, there's a new intention-revealing method on `IdentifierFactory` for this:

```suggestion
	localVarName := builder.idFactory.CreateLocal(params.NameHint+"Secret")
```

v2/tools/generator/internal/codegen/pipeline/add_secrets.go
```diff
@@ -0,0 +1,134 @@
...
+
+	"github.com/Azure/azure-service-operator/v2/tools/generator/internal/astmodel"
+	"github.com/Azure/azure-service-operator/v2/tools/generator/internal/config"
+)
+
+// AddSecretsStageID is the unique identifier for this pipeline stage
+const AddSecretsStageID = "addSecrets"
+
+// AddSecrets replaces properties flagged as secret with genruntime.SecretReference
+func AddSecrets(_ *config.Configuration) Stage { // TODO: Will need this configuration parameter eventually to allow config overrides
```
theunrepentantgeek, line 21: Doesn't this stage have pre/posrequisites that should be explicit?

At the very least, the stages creating the ARM types must run second, yes?

v2/tools/generator/internal/codegen/pipeline/add_secrets.go
```diff
@@ -0,0 +1,134 @@
...
+	return MakeStage(
+		AddSecretsStageID,
+		"Replace properties flagged as secret with genruntime.SecretReference",
+		func(ctx context.Context, state *State) (*State, error) {
+
+			updatedSpecs, err := transformSpecSecrets(state.Types())
+			if err != nil {
+				return nil, err
```
theunrepentantgeek, line 29:
```suggestion
				return nil, errors.Wrap(err, "transforming spec secrets")
```

v2/tools/generator/internal/codegen/pipeline/add_secrets.go
```diff
@@ -0,0 +1,134 @@
...
+			}
+
+			updatedStatuses, err := removeStatusSecrets(state.Types())
+			if err != nil {
+				return nil, err
```
theunrepentantgeek, line 34:
```suggestion
				return nil, errors.Wrap(err, "removing status secrets")
```

v2/tools/generator/internal/codegen/pipeline/add_secrets.go
```diff
@@ -0,0 +1,134 @@
...
+		result.Add(updatedDef)
+	}
+
+	return result, nil
+}
+
+func isTypeString(t astmodel.Type) bool {
+	primType, ok := astmodel.AsPrimitiveType(t)
+	return ok && primType == astmodel.StringType
+}
```
theunrepentantgeek, line 93: Add a predicate to `PropertyDefinition` called `HasType()` and use that instead of creating a standalone function.

v2/tools/generator/internal/codegen/pipeline/add_secrets.go
```diff
@@ -0,0 +1,134 @@
...
+
+func transformSecretProperties(_ *astmodel.TypeVisitor, it *astmodel.ObjectType, _ interface{}) (astmodel.Type, error) {
+	for _, prop := range it.Properties() {
+		if prop.IsSecret() {
+			// The expectation is that this is a string
+			propType := prop.PropertyType()
+			if !isTypeString(propType) {
+				return nil, errors.Errorf("expected property %q to be a string, but was: %T", prop.PropertyName(), propType)
+			}
```
theunrepentantgeek, line 118: With `PropertyDefinition.HasType()`, this would become

```suggestion
			if !prop.HasType(astmodel.StringType) {
				return nil, errors.Errorf("expected property %q to be a string, but was: %T", prop.PropertyName(), prop.PropertyType())
			}
```

v2/tools/generator/internal/armconversion/convert_to_arm_function_builder.go
```diff
@@ -387,6 +388,40 @@ func (builder *convertToARMBuilder) convertReferenceProperty(_ *astmodel.Convers
...
+//		return nil, err
+//	}
+//	<destination> = <namehint>Secret
+func (builder *convertToARMBuilder) convertSecretProperty(_ *astmodel.ConversionFunctionBuilder, params astmodel.ConversionParameters) []dst.Stmt {
+	isString := astmodel.TypeEquals(params.DestinationType, astmodel.StringType)
+	if !isString {
+		return nil
+	}
+
+	isReference := astmodel.TypeEquals(params.SourceType, astmodel.SecretReferenceType)
```
theunrepentantgeek, line 404: I originally read _reference_ as _pointer_, but it's actually whether it is a reference to a secret.
```suggestion
	isSecretReference := astmodel.TypeEquals(params.SourceType, astmodel.SecretReferenceType)
```

v2/api/dbformysql/v1alpha1api20210501/flexible_server_types_gen.go
```diff
@@ -393,7 +393,11 @@ func (servers *FlexibleServers_Spec) ConvertToARM(resolved genruntime.ConvertToA
 		result.Properties.AdministratorLogin = &administratorLogin
 	}
 	if servers.AdministratorLoginPassword != nil {
-		administratorLoginPassword := *servers.AdministratorLoginPassword
+		administratorLoginPasswordSecret, err := resolved.ResolvedSecrets.SecretOrErr(*servers.AdministratorLoginPassword)
+		if err != nil {
+			return nil, err
```
theunrepentantgeek, line 398: Suggest generating call here so that it's explicit which property was trying to use the secret, for example

```suggestion
			return nil, errors.Wrap(err, "looking up secret for AdministratorLoginPassword")
```

theunrepentantgeek (changes requested):

v2/internal/reconcilers/azure_generic_arm_reconciler.go
```diff
@@ -324,6 +324,8 @@ func (r *AzureDeploymentReconciler) StartDeleteOfResource(ctx context.Context) (
 		// means that the owner was deleted in Kubernetes. The current
 		// assumption is that that deletion has been propagated to Azure
 		// and so the child resource is already deleted.
+		// TODO: We should confirm that this is actuall the owner missing and not some
+		// TODO: other reference.
```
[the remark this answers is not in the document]
matthchr: Fixed

v2/pkg/genruntime/resolver_test.go
```diff
@@ -190,6 +193,26 @@ func createExtensionResourceOnDeepHierarchyInResourceGroup(rgName string, parent
...
+//		Spec: mysql.FlexibleServers_Spec{
+//			Owner: genruntime.KnownResourceReference{
+//				Name: rg.Name,
+//			},
+//			AzureName: name, // defaulter webhook will copy Name to AzureName
+//			AdministratorLoginPassword: &secretRef,
+//		},
+//	}
+//}
+
```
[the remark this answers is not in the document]
matthchr: Done

v2/pkg/genruntime/resolver_test.go
```diff
@@ -381,11 +404,83 @@ func Test_ResolveReferenceToARMID_ARMResource_ReturnsExpectedID(t *testing.T) {
...
+		},
+		// Needed to avoid nil map error
+		Data: map[string][]byte{
+			secretKey: []byte(secretValue),
+		},
+		Type: "Opaque",
+	}
+
+	err = client.Client.Create(ctx, secret)
+	g.Expect(err).ToNot(HaveOccurred())
```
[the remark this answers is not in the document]
matthchr: The pattern I try to follow is:

`g.Expect(err).ToNot(HaveOccurred())`
`g.Expect(myFunc()).To(Succeed())`

It reads weird to me to be talking about an error "succeeding" (or a function "Not Occurring").

v2/pkg/genruntime/secret_resolver.go
```diff
@@ -0,0 +1,108 @@
...
+	valueBytes, ok := secret.Data[ref.Key]
+	if !ok {
+		// TODO: Do we want a specific error type for this?
+		return "", errors.Errorf("Secret %q does not contain key %q", refNamespacedName.String(), ref.Key)
+	}
+
+	//value, err := base64.StdEncoding.DecodeString(string(valueBytes))
+	//if err != nil {
+	//	return "", errors.Wrapf(err, "SecretReference %q value was not valid base64", ref.String())
+	//}
```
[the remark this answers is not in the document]
matthchr: Hah, no. Was from an old implementation, will fix.

v2/tools/generator/internal/codegen/pipeline/add_secrets.go
matthchr: Done

v2/tools/generator/internal/codegen/pipeline/add_secrets.go
matthchr: Done

v2/tools/generator/internal/armconversion/convert_to_arm_function_builder.go
matthchr: Yeah good catch! Fixed.

v2/tools/generator/internal/armconversion/convert_to_arm_function_builder.go
matthchr: Done

v2/pkg/genruntime/secret_resolver.go
```diff
@@ -0,0 +1,108 @@
...
+
+// MakeResolvedSecrets creates a ResolvedSecrets
+func MakeResolvedSecrets(secrets map[SecretReference]string) ResolvedSecrets {
+	return ResolvedSecrets{
+		secrets: secrets,
+	}
+}
+
+// SecretOrErr looks up the secret value for the given reference. If it cannot be found, an error is returned.
+func (r ResolvedSecrets) SecretOrErr(ref SecretReference) (string, error) {
```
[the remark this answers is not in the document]
matthchr: `TryGet` makes me think the signature would be `func (ref SecretReference) (string, bool)` - `LookupSecret` is ok though, will swap to that.

v2/api/dbformysql/v1alpha1api20210501/flexible_server_types_gen.go
matthchr: Good idea.

v2/tools/generator/internal/codegen/pipeline/add_secrets.go
matthchr: Did this but then realized that all that was doing was calling `prop.PropertyType().Equals`. It felt silly to have a method on property which just proxied a call to `PropertyType()` (which is already exported).

I just moved to do `prop.PropertyType().Equals` instead.

matthchr: The problem here is that we need to call `AsPrimitiveType` to perform `MetaType` unwrapping, which feels like a very powerful/risky operation. The name `HasType` doesn't really match the behavior because it sounds like it's checking if the type equals something (also accomplishable by `prop.PropertyType().Equals()`), but in reality in order to work here it would actually have to be doing something like `astmodel.Unwrap(prop.PropertyType()).Equals()`.

We previously didn't have a generic `astmodel.Unwrap` though - I've added one and done `astmodel.Unwrap(prop.PropertyType()).Equals()` directly here instead, which avoids the custom function but also doesn't promote the pattern into a method on `PropertyDefinition`.

Let me know what you think of that approach.

v2/tools/generator/internal/codegen/pipeline/add_secrets.go
matthchr: See above

v2/tools/generator/internal/codegen/pipeline/add_secrets.go
matthchr: Yeah it has to run between the SpecWithStatus stage and the ARM types stage. I've added:
```
RequiresPrerequisiteStages(AugmentSpecWithStatusStageID).RequiresPostrequisiteStages(CreateARMTypesStageID)
```

theunrepentantgeek (approved):

PR #2044, opened 2022-01-20
Update Extensions Spec and example

theunrepentantgeek:

Updates the spec with our recent decisions about package names and clarifies some language.

gif

**If applicable**:


docs/hugo/content/architecture-decision-records/ADR-2022-01-Reconciler-Extensions.md
```diff
@@ -72,26 +74,26 @@ type ErrorClassifierExtension interface {
...
 // redis_extensions.go
 
-package extensions
+package customization
 
 type RedisExtensions struct{}
 ```
 
-This type doesn't contain any members as we require extension points to be pure functions; if they retain mutable state, we run the risk of introducing concurrency and sequence errors that would be hard to diagnose and fix.
+This type doesn't contain any members as we require extension points to be pure functions; if they contain mutable state, we run the risk of introducing concurrency and sequence errors that would be hard to diagnose and fix.
```
babbageclunk, line 87: For something like the vnet/subnet problem we're going to need to look up the subnets and put them into the vnet before sending to ARM, aren't we? That wouldn't be doable with a pure function.

babbageclunk (approved):
Looks good, just a quibble about whether the customisation methods are pure functions.

docs/hugo/content/architecture-decision-records/ADR-2022-01-Reconciler-Extensions.md
theunrepentantgeek: As long as the VNet is passed into the function, and back out again afterwards, I think it can.

PR #2049, opened 2022-01-24
astbuilder improvements

matthchr:
* Rename CompositeLiteralDetails to CompositeLiteralBuilder (because it's a builder).
* Add SliceLiteralBuilder for slices.


v2/tools/generator/internal/astbuilder/builder.go
```diff
@@ -293,6 +293,14 @@ func Dereference(expr dst.Expr) *dst.StarExpr {
 	}
 }
 
+// Ptr returns a statement including a star, indicating the inner expression is a pointer
```
Porges, line 296: This seems like a bad name? Is it mean to be that the `expr` is a type?

Porges (approved):

v2/tools/generator/internal/astbuilder/builder.go
matthchr: I had a use for this that was different than `Dereference` (which this does exactly), but I deleted it and agree this name is bad. I'll just remove this.

PR #2050, opened 2022-01-24
gofmt -s -w .

Porges:
Run gofmt on v1 code; this helps our "Go Report Card" score.


matthchr (approved):
Would be cool to have a check in the CI for V1 too but this is a good start.

PR #2059, opened 2022-01-28
Support secret rollover

matthchr:
* Add index in client-go for resources that support secrets, indexed on
   the property containing the secret.
 * Trigger reconcile for resources that have a secret reference by watching
   secret modifications and using the resource specific indexes to
   see if that secret is in use by a resource.
 * Add tests to ensure that this whole dance works. The test cannot run
   in the recording tests because they do not use a cached client and
   the custom indexing is all client-side.

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/internal/controllers/controller_resources.go
+++ b/v2/internal/controllers/controller_resources.go
@@ -6,16 +6,31 @@ Licensed under the MIT license.
 package controllers
 
 import (
+	"context"
+	"fmt"
+	"reflect"
+	"time"
+
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	resources "github.com/Azure/azure-service-operator/v2/api/resources/v1alpha1api20200601"
+	. "github.com/Azure/azure-service-operator/v2/internal/logging"
+	"github.com/Azure/azure-service-operator/v2/internal/reflecthelpers"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/registration"
 )
 
-func GetKnownStorageTypes() []client.Object {
+func GetKnownStorageTypes() []registration.StorageType {
 	knownTypes := getKnownStorageTypes()
 
-	knownTypes = append(knownTypes, new(resources.ResourceGroup))
+	knownTypes = append(
+		knownTypes,
+		registration.MakeStorageType(new(resources.ResourceGroup)))
 
 	return knownTypes
 }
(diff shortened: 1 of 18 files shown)


v2/internal/controllers/controller_resources.go
```diff
@@ -6,16 +6,33 @@ Licensed under the MIT license.
...
-func GetKnownStorageTypes() []client.Object {
+func GetKnownStorageTypes() []registration.StorageTypeInfo {
 	knownTypes := getKnownStorageTypes()
 
-	knownTypes = append(knownTypes, new(resources.ResourceGroup))
+	knownTypes = append(
+		knownTypes,
+		registration.StorageTypeInfo{
+			Obj: new(resources.ResourceGroup),
+		})
```
theunrepentantgeek, line 35: Suggest creating helper method `registration.NewStorageType()`

v2/internal/controllers/controller_resources.go
```diff
@@ -6,16 +6,33 @@ Licensed under the MIT license.
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	resources "github.com/Azure/azure-service-operator/v2/api/resources/v1alpha1api20200601"
+	. "github.com/Azure/azure-service-operator/v2/internal/logging"
+	"github.com/Azure/azure-service-operator/v2/internal/reflecthelpers"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/registration"
 )
 
-func GetKnownStorageTypes() []client.Object {
+func GetKnownStorageTypes() []registration.StorageTypeInfo {
```
theunrepentantgeek, line 28: Since this is in the `registration` package, consider renamign this to `registration.StorageType`.

v2/internal/controllers/controller_resources.go
```diff
@@ -6,16 +6,33 @@ Licensed under the MIT license.
...
 	knownTypes := getKnownStorageTypes()
 
-	knownTypes = append(knownTypes, new(resources.ResourceGroup))
+	knownTypes = append(
+		knownTypes,
+		registration.StorageTypeInfo{
+			Obj: new(resources.ResourceGroup),
```
super-harsh, line 34: Would be better if we rename `Obj` to something which makes more sense.

v2/internal/controllers/controller_resources.go
matthchr: Did this, except called it `MakeStorageType` instead of `New` because it doesn't return a pointer.

v2/internal/controllers/controller_resources.go
matthchr: Dropped the `Registration` suffix from the other two types too, as they were repeating themselves (`registration.WatchRegistration`)

v2/internal/controllers/controller_resources.go
matthchr: Agree that the `Obj` naming is pretty "generic", but it is what we call it in other places, such as the [generic controller](https://github.com/Azure/azure-service-operator/blob/main/v2/internal/controllers/generic_controller.go#L101). The generic controller took the abbreviation from controller-runtime, where it's [object](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/builder/controller.go#L81), which I don't think is tons clearer.

I've added some documentation onto the `StorageType` that describes what Obj is, which maybe will help clarify things, but left it named as is due to it matching parameters elsewhere.

theunrepentantgeek (approved):

PR #2060, opened 2022-01-28
Add support for reconcile policy

### LINKED ISSUE #1633: Feature: Add a no-delete annotation
Users have requested an annotation similar to the `skip-reconcile` annotation that doesn't skip all reconciliation but skips deletion.
This is especially attractive to users using tools like Flux, who may want optional additional safeguards to prevent the accidental deletion of resources in Azure. Kubernetes RBAC doesn't really solve this problem because:

* Tools like Flux often have to have sufficient permissions to issue a delete, and denying them that capability universally (or at a per-namespace level) is awkward/difficult.
* RBAC doesn't allow sub-namespace granularity, nor does it universally prevent deletes from all identities (just from the ones denied RBAC permissions).

In discussion with some users, we brainstormed a couple of ideas:
1. A `skip-delete` style annotation which allows the deletion of the resource in Kubernetes, but doesn't actually do anything to the backing Azure resource.
2. A `no-delete` style annotation, paired with a per-resource validating webhook. Resources with the `no-delete` annotation would have their DELETE requests rejected by the validating webhook with a corresponding error.

There are some existing projects that already exist and seem to provide a `no-delete` annotation style experience, such as:
https://github.com/petrkotas/k8s-object-lock and https://github.com/k14s/k8s-resource-protection. (Note: I am not sure how production ready these projects are)
This feature has been more broadly requested in core k8s [here](https://github.com/kubernetes/kubernetes/issues/10179). It does seem like if we were to take the `no-delete` route it could be made generically and applicable to all kinds of resources.

matthchr:
This closes #1633.

Adds support for a new reconcile-policy annotation that allows users to
prevent all modifications to Azure resources (PUT, DELETE), or just all
DELETE's.

**If applicable**:
- [x] this PR contains tests


v2/internal/reconcilers/reconcile_policy.go
```diff
@@ -0,0 +1,58 @@
...
+
+package reconcilers
+
+import "github.com/pkg/errors"
+
+// ReconcilePolicyAnnotation describes the reconcile policy for the resource in question.
+// A reconcile policy describes what action (if any) the operator is allowed to take when
+// reconciling the resource.
+// If no reconcile policy is specified, the default is "run"
+const ReconcilePolicyAnnotation = "serviceoperator.azure.com/reconcile-policy"
```
matthchr, line 14: I'm not _totally_ sold on this name. Open to better ones if people have them. Other ideas I had:
`azure-interaction-policy`, `reconcile-mode`?

Similarly I am not totally sold on the values. Would `reconcile-policy: act|observe|skip-delete` be better?

docs/hugo/content/introduction/annotations.md
```diff
@@ -0,0 +1,27 @@
...
+
+Note that unless otherwise specified, allowed values are _case sensitive_ and should be provided in lower case.
+
+1. `serviceoperator.azure.com/reconcile-policy`: Specifies the reconcile policy to use. Allowed values are:
+    - `run`: The operator performs all actions as normal. This is the default if no annotation is specified.
+    - `skip`: All modification actions on the backing Azure resource are skipped. GETs are still issued to ensure that the resource
+    exists. If the resource doesn't exist, the `Ready` condition will show a `Warning` state with the details of the missing resource
+    until the resource is created. If the resource is deleted in Kubernetes, it is _not_ deleted in Azure. In REST API terminology, 
+    PUT and DELETE are skipped while GET is allowed.
+    - `skip-delete`: Modifications are pushed to the backing Azure resource, but if the resource is deleted in Kubernetes,
```
jonnylangefeld, line 13: I think `abandon` or `detach` might be good terms?

docs/hugo/content/introduction/annotations.md
```diff
@@ -0,0 +1,27 @@
+# Annotations understood by the operator
+
+## Annotations specified by the user
+
+Note that unless otherwise specified, allowed values are _case sensitive_ and should be provided in lower case.
+
+1. `serviceoperator.azure.com/reconcile-policy`: Specifies the reconcile policy to use. Allowed values are:
+    - `run`: The operator performs all actions as normal. This is the default if no annotation is specified.
```
jonnylangefeld, line 8: I think this is technically not needed. Generally I understand that one might want to be explicit, but in this case that's what anyone using the operator would expect anyway.

v2/internal/controllers/reconcile_policy_test.go
```diff
@@ -0,0 +1,132 @@
...
+	resources "github.com/Azure/azure-service-operator/v2/api/resources/v1alpha1api20200601"
+)
+
+func Test_ReconcilePolicy_SkipReconcileAddedAlongWithTagsChange_ReconcileIsSkipped(t *testing.T) {
+	testModificationSkipped(t, "skip")
+}
+
+func Test_UnknownReconcilePolicy_SkipsAllUpdates(t *testing.T) {
+	testModificationSkipped(t, "UNKNOWN")
+}
```
jonnylangefeld, line 24: hmm should the reconciliation be skipped if there is an unknown string in the policy?

v2/internal/reconcilers/reconcile_policy.go
```diff
@@ -0,0 +1,58 @@
...
+	case string(ReconcilePolicyRun):
+		return ReconcilePolicyRun, nil
+	case string(ReconcilePolicySkip):
+		return ReconcilePolicySkip, nil
+	case string(ReconcilePolicySkipDelete):
+		return ReconcilePolicySkipDelete, nil
+	default:
+		// Defaulting to skip. The user is attempting to configure policy but has done it wrong,
+		// if we default to Run we may inadvertently modify their object
+		return ReconcilePolicySkip, errors.Errorf("%q is not a known reconcile policy", policy)
```
jonnylangefeld, line 46: However, this also makes so that as soon as the `serviceoperator.azure.com/reconcile-policy` annotation is present we skip, no matter the string value. That also makes the verb `skip` unnecessary (as if it doesn't matter if I put `skip` or `sk1p` or `piks` in there). I think this has great implications that as soon as the annotation is present, the operator skips reconcile.

That is fine, I would just be more explicit in the docs about that. Essentially there's three states:
1. I don't set anything and the operator reconciles
2. I set the `serviceoperator.azure.com/reconcile-policy` annotation with any value and the operator doesn't reconcile.
3. I set `serviceoperator.azure.com/reconcile-policy: skip-delete` and the operator reconciles but skips delete.

Given the name of the annotation however, I probably wouldn't expect that the operator skips reconciling as soon as the annotation is set. I'd probably expect that if the annotation was called `serviceoperator.azure.com/skip-reconcile`. But since it's called `serviceoperator.azure.com/reconcile-policy`, I'd probably expect that I have to set the correct policy to achieve the behavior I want (either `skip` or `skip-delete`).

jonnylangefeld (commented):
Thanks for coming around so quickly with this PR!

v2/internal/reconcilers/reconcile_policy.go
matthchr: Or `abandon` as @jonnylangefeld suggests below for the `skip-delete` case?

docs/hugo/content/introduction/annotations.md
matthchr, line 13: @theunrepentantgeek brought up a concern regarding the "delete-by-default even for resources we may not have created" which this implies. He asked if there was a way to know if we created the resource and base the default behavior on that instead. So for resources which ASO created it would default to deleting them but for resources it adopted it would default to abandoning them instead (not issuing a DELETE to Azure).

Just making this comment here as a place for discussion, I'll be doing some thinking about this and reply later.

docs/hugo/content/introduction/annotations.md
theunrepentantgeek: I like the idea of enabling users to be explicit _even if_ it's the default behaviour.

docs/hugo/content/introduction/annotations.md
theunrepentantgeek: I'm thinking there's a particular risk here for stateful resources that can't simply be recreated from their spec - the most obvious of these being various kinds of database.

docs/hugo/content/introduction/annotations.md
theunrepentantgeek, line 8: I'm not sure that `run` is an obivous term that someone would understand without reference to documentation. 

How about `fully-managed`?

docs/hugo/content/introduction/annotations.md
theunrepentantgeek: Reusing the term `skip` seems to leave things open for misinterpretation (not that we can ever completely avoid that).

In the context of reconciler behaviour, how about `never-delete` ?

docs/hugo/content/introduction/annotations.md
```diff
@@ -0,0 +1,27 @@
...
+    - `skip`: All modification actions on the backing Azure resource are skipped. GETs are still issued to ensure that the resource
+    exists. If the resource doesn't exist, the `Ready` condition will show a `Warning` state with the details of the missing resource
+    until the resource is created. If the resource is deleted in Kubernetes, it is _not_ deleted in Azure. In REST API terminology, 
+    PUT and DELETE are skipped while GET is allowed.
+    - `skip-delete`: Modifications are pushed to the backing Azure resource, but if the resource is deleted in Kubernetes, 
+    it is _not_ deleted in Azure. In REST API terminology, PUT and GET are allowed while DELETE is skipped.
+    
+    If a value other than one of the above is specified, an error is written to the operator logs and the most restrictive 
+    option (`skip`) is chosen. To allow modifications, either correct the malformed annotation on the resource or remove the 
+    annotation entirely.
```
theunrepentantgeek, line 18: I like this safe default.

docs/hugo/content/introduction/annotations.md
```diff
@@ -0,0 +1,27 @@
+# Annotations understood by the operator
+
+## Annotations specified by the user
+
+Note that unless otherwise specified, allowed values are _case sensitive_ and should be provided in lower case.
```
theunrepentantgeek, line 5: Maybe we apply Postel's law and accept with a case-insensitive match?

matthchr: I believe users will expect these to be case sensitive as that is the behavior elsewhere. For example see:
[Kubernetes labels are case sensitive](https://downey.io/notes/dev/kubernetes-labels-are-case-sensitive/).

Or from the documentation [names in Kubernetes are case sensitive](https://kubernetes.io/docs/reference/kubectl/_print/)
> NAME: Specifies the name of the resource. Names are case-sensitive. 

Since the type used to store the annotations by k8s is `map[string]string`, the key is by definition case sensitive. If we tried to make the keys case insensitive we introduce the possibility of multiple keys differing only by case but with conflicting values. We could make the value case insensitive but I argue that it is the name of a "mode" and so should match the label/name casing rules and also be case sensitive.

docs/hugo/content/introduction/annotations.md
matthchr: is `fully` adding a lot here? Maybe just `managed`?

v2/internal/reconcilers/azure_generic_arm_reconciler.go
```diff
@@ -226,6 +226,27 @@ func (r *AzureDeploymentReconciler) SetPollerResumeToken(id string, token string
 	genruntime.AddAnnotation(r.obj, PollerResumeIDAnnotation, id)
 }
 
+// ClearPollerResumeToken clears the poller resume token and ID annotations
+func (r *AzureDeploymentReconciler) ClearPollerResumeToken() {
+	genruntime.AddAnnotation(r.obj, PollerResumeTokenAnnotation, "")
+	genruntime.AddAnnotation(r.obj, PollerResumeIDAnnotation, "")
```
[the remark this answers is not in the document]
matthchr: Done

v2/internal/reconcilers/reconcile_policy.go
matthchr: Ok I like: `reconcile-policy`: `manage|skip|detach-on-delete` I think. Any thoughts?

docs/hugo/content/introduction/annotations.md
matthchr: Ok I like: `reconcile-policy:` `manage|skip|detach-on-delete` I think. Thoughts?

v2/internal/controllers/reconcile_policy_test.go
matthchr: Resolving this and moving discussion into [this comment](https://github.com/Azure/azure-service-operator/pull/2060#discussion_r794972922) which talks about it more in-depth.

v2/internal/reconcilers/reconcile_policy.go
matthchr: Agree that an empty `reconcile-policy` doing `skip` probably doesn't make sense. I'm going to change this behavior and default to `manage` (aka behave normally). That seems to be the standard behavior if you get an annotation like this "wrong".

If we really want to enforce that people don't set a "wrong" value here we could prevent it at the webhook level (which has the advantage of being able to actually tell them they did it wrong as opposed to just doing something they didn't expect), but that seems overkill for now. We could investigate doing that in the future if there's actual desire for it from users.

docs/hugo/content/introduction/annotations.md
matthchr: I've written a little design up [here](https://github.com/Azure/azure-service-operator/pull/2067). I think it makes sense to decouple that from this PR though, as it's really two things:

1. Support for `reconcile-policy` at all.
2. The default behavior of `reconcile-policy` for adopted resources (as well as the ability to actually detect if we _are_ adopting a resource, which will take additional controller changes).

Once 1 exists we can do 2 (if we decide that we want to).

docs/hugo/content/introduction/annotations.md
matthchr: I changed this to `manage`. Let me know what you think.

v2/internal/reconcilers/reconcile_policy.go
dhilipkumars: How about ? mimicking Azure action [types](https://docs.microsoft.com/en-us/azure/role-based-access-control/role-definitions#actions-format)
```
"serviceoperator.azure.com/object-access-type": "*/read/write/delete"
```

docs/hugo/content/introduction/annotations.md
matthchr: I've changed this (see the conversation [here](https://github.com/Azure/azure-service-operator/pull/2060#discussion_r794972922)). I'm thinking if we want this behavior we need to rename the annotation itself to something that more strongly conveys "this will skip no matter what you set" such as `skip-reconcile` and `skip-delete`.

I'd be OK doing that, although the fact that the two annotations overlap some and you can set them both at the same time (`skip-reconcile` is basically `skip-delete` + `skip-modifications`) could be confusing. It feels a bit like two coupled booleans to me.

As I mentioned in the other comment chain, we could also do a webhook that rejects this annotation with unknown values. I'm not sure that's really a pattern that Kubernetes users are used to but it's a much nicer failure mode than the way I had this (which allows an invalid value and then makes you debug through the operator logs to determine that your value was bad).

docs/hugo/content/introduction/annotations.md
matthchr: Agree this is technically not needed, but given the name of the property `reconcile-policy` it felt like we should have a name for the "normal" policy, even if people practically speaking aren't going to bother setting it explicitly.

docs/hugo/content/introduction/annotations.md
matthchr: Resolving this: Let's move conversation on this topic to the design around it [here](https://github.com/Azure/azure-service-operator/pull/2067).

docs/hugo/content/introduction/annotations.md
theunrepentantgeek: I'm not convinced, but I'll go with whatever you choose. 

Failing to do what the user wanted because they typed `Skip` instead of `skip` seems needlessly pedantic.

v2/internal/reconcilers/reconcile_policy.go
theunrepentantgeek: I'm not sure we need this level of flexibility - what would it mean for ASO to have `read|delete` as the policy?

v2/internal/reconcilers/reconcile_policy.go
```diff
@@ -0,0 +1,81 @@
...
+// A reconcile policy describes what action (if any) the operator is allowed to take when
+// reconciling the resource.
+// If no reconcile policy is specified, the default is "run"
+const ReconcilePolicyAnnotation = "serviceoperator.azure.com/reconcile-policy"
+
+type ReconcilePolicy string
+
+const (
+	// ReconcilePolicyManage instructs the operator to manage the resource in question.
+	// This includes issuing PUTs to update it and DELETE's to delete it if deleted in Kuberentes.
```
theunrepentantgeek, line 25:
```suggestion
	// This includes issuing PUTs to update it and DELETE's to delete it from Azure if deleted in Kuberentes.
```

v2/internal/reconcilers/reconcile_policy.go
```diff
@@ -0,0 +1,81 @@
...
+	case string(ReconcilePolicyDetachOnDelete):
+		return ReconcilePolicyDetachOnDelete, nil
+	default:
+		// Defaulting to manage.
+		return ReconcilePolicyManage, errors.Errorf("%q is not a known reconcile policy", policy)
+	}
+}
+
+// ShouldDelete determines if the policy allows deletion of the backing Azure resource
+func (r ReconcilePolicy) ShouldDelete() bool {
```
theunrepentantgeek, line 56: There's a disconnect between the naming of the method and the comment. I suspect the comment is easier to understand, so I suggest renaming these methods as `AllowsXYZ()`.

```suggestion
func (r ReconcilePolicy) AllowsDeletion() bool {
```

theunrepentantgeek (approved):
Looks good. A couple of minor suggestions.

gif

docs/hugo/content/introduction/annotations.md
theunrepentantgeek: Works for me.

docs/hugo/content/introduction/annotations.md
theunrepentantgeek: Fair enough.

v2/internal/reconcilers/azure_generic_arm_reconciler.go
```diff
@@ -476,53 +508,73 @@ func (r *AzureDeploymentReconciler) handlePollerFailed(ctx context.Context, err
...
 	}
 
-	// TODO: I think if we're really we can avoid this as the HTTP payload in poller should have all we need already?
-	// TODO: for now doing this though because my quick read of the Go SDK is that it does a final GET on the URL too...
-	// TODO: Maybe some Azure services or types of async operations actually don't return the latest resource?
-	status, _, err := r.getStatus(ctx, resourceID)
+	r.ClearPollerResumeToken()
+	conditions.SetCondition(r.obj, r.PositiveConditions.Ready.Succeeded(r.obj.GetGeneration()))
+	err = r.CommitUpdate(ctx)
 	if err != nil {
```
super-harsh, line 549: Here we can use something like `if err = r.CommitUpdate(ctx) ; err != nil {`

v2/internal/reconcilers/azure_generic_arm_reconciler.go
```diff
@@ -476,53 +508,73 @@ func (r *AzureDeploymentReconciler) handlePollerFailed(ctx context.Context, err
...
+	return r.makeSuccessResult(), nil
+}
+
+func (r *AzureDeploymentReconciler) handlePollerSuccess(ctx context.Context) (ctrl.Result, error) {
+	r.log.V(Status).Info(
+		"Resource successfully created",
+		"resourceID", genruntime.GetResourceIDOrDefault(r.obj))
+
+	err := r.updateStatus(ctx)
+	if err != nil {
```
super-harsh, line 564: same here, since we're not using the lhs anywhere but inside the if block, below syntax would avoid the extra variable assignment lines.
`if err = r.updateStatus(ctx); err != nil {`

v2/internal/reconcilers/azure_generic_arm_reconciler.go
```diff
@@ -659,6 +711,39 @@ func (r *AzureDeploymentReconciler) getStatus(ctx context.Context, id string) (g
...
+		return errors.Errorf("resource has no resource id")
+	}
+
+	status, _, err := r.getStatus(ctx, resourceID)
+	if err != nil {
+		return errors.Errorf("error getting status for resource ID %q", resourceID)
+	}
+
+	err = r.setStatus(status)
+	if err != nil {
```
super-harsh, line 740: same here, can avoid extra lines by having one liner assignment and if statement.

v2/internal/reconcilers/reconcile_policy.go
matthchr: Agree @theunrepentantgeek - this is part of the reason why it feels better IMO to have a single `policy` rather than a `modification-policy` and a `delete-policy`, because it'd be weird to allow deletes but not writes. The 3 values that seem to make sense are:
`read|write|delete` == my proposed `manage`
`read|write` == my proposed `detach-on-delete`
`read` == my proposed `skip`.

v2/internal/reconcilers/reconcile_policy.go
matthchr: Done

v2/internal/reconcilers/reconcile_policy.go
matthchr: Done (and for AllowsModify)

v2/internal/reconcilers/azure_generic_arm_reconciler.go
matthchr: Done

v2/internal/reconcilers/azure_generic_arm_reconciler.go
matthchr: Done

PR #2063, opened 2022-02-02
Tidy up configuration code

theunrepentantgeek:

We were ending up with a large amount of code duplication across the configuration hierarchy, and this was going to get worse as the number of configuration options increased.

This PR uses the visitor pattern to abstract away the task of traversing the configuration structure, allowing new configuration options to be added without a lot of boilerplate.

(Next step will be to roll ExportFilters into the configuration hierarchy, which should reduce the size of the configuration file somewhat, making it easier for others to contribute to.)

gif

**If applicable**:
- [x] this PR contains tests


v2/tools/generator/internal/codegen/pipeline/add_cross_resource_references.go
```diff
@@ -83,20 +83,20 @@ func AddCrossResourceReferences(configuration *config.Configuration, idFactory a
 				// TODO: Properties collapsing work for this.
 			}
 
-			err := kerrors.NewAggregate(crossResourceReferenceErrs)
-			if err != nil {
-				return nil, err
+			if len(crossResourceReferenceErrs) > 0 {
```
babbageclunk, line 86: It looks like `kerrors.NewAggregate` returns nil if it's passed an empty slice - do you need the guard?

v2/tools/generator/internal/config/configuration_visitor.go
```diff
@@ -0,0 +1,142 @@
...
+ */
+
+package config
+
+import (
+	"github.com/Azure/azure-service-operator/v2/tools/generator/internal/astmodel"
+)
+
+// configurationVisitor is used to facilitate easy walking of the ObjectModelConfiguration hierarchy, abstracting
+// away traversal logic so that new uses of the heirarchy can concentrate on their specific functionality.
```
babbageclunk, line 13:
```suggestion
// away traversal logic so that new uses of the hierarchy can concentrate on their specific functionality.
```

v2/tools/generator/internal/config/group_configuration.go
```diff
@@ -37,101 +38,53 @@ func NewGroupConfiguration(name string) *GroupConfiguration {
 	}
 }
 
-// TypeRename looks up a rename for the specified type, returning the new name and true if found, or empty string
-// and false if not.
-func (gc *GroupConfiguration) TypeRename(name astmodel.TypeName) (string, error) {
```
babbageclunk, line 42: Mmm-mmm, delicious red

babbageclunk (approved):
This came out really nicely! Love to see all that red.

v2/tools/generator/internal/codegen/pipeline/add_cross_resource_references.go
theunrepentantgeek: I wrote it because `err` was being declared as a `kerrors.Aggregate` instead of an `error`; have reworked.

Issue #1978, opened 2021-11-16
Feature: Support handcrafted overrides for reconcile


matthchr:
This probably needs a basic design as it's involved, but the idea is to support things like:
* Before sending this resource to Azure...
* After this resource has been successfully created...
* After this resource has failed to create...

We also probably need hooks at the ARM serialization/payload level to be able to hook in and deal with inconsistencies, etc.

The general idea here is that we could manually implement an interface on the code generated resource, that lets us craft some specialized handlers to deal with specific cases that aren't possible to handle generically. Things like:
* VNETs PUT deleting its child resources if they aren't present in the PUT.
* Certain resources needing to create a secret by calling a `GetKeys` API or similar.

matthchr:
See #1471 for a feature that will probably make use of this

matthchr:
One thing we're going to need to figure out for this is how it interactions with storage versions. There are sorta two options that I see:

1. Customizations are put on the storage version, this means that the same customization will be run regardless of what API version the customer used.
2. Customizations are put on the customer facing API version. This means that different customizations can be run for the same `Kind` of resource on different API versions but also means that we may need to implement the customization on the storage version (which will then perform the conversion and delegate?) This is a bit awkward though because then we'll run the same conversion twice (we're already doing in inside the storage versions `ConvertToARM`).

theunrepentantgeek:
We need to also think about the upgrading process - if someone injects custom behaviour for version `v2020-06-01` of `microsoft.foobar` then what happens when we introduce support for a later version of that resource, say `v2021-05-01`.

* If the customization is implemented on the API version `v2020-06-01` then it will need to be duplicated on the new version.

* If the customization is implemented on the Hub/Storage version `v2020-06-01-storage` then it will need to be moved/updated as the new version will change the Hub/Storage version to `v2021-05-01-storage`.

Neither approach is entirely simple.

matthchr:
Support for this was added in #2064.

PR #2064, opened 2022-02-02
Add support for resource extensions generation 

super-harsh:
Links to : #1978 

We are discovering inconsistencies in the way different Azure Resource Providers behave. These inconsistencies are anticipated. Given the number of disparate teams independently implementing their providers and given the way the ARM guidance has changed over time, differences in behaviour are to be expected.

This PR adds support to generate the resource extensions for each Azure resources, registers them and returns a map of **resourceOwnerRef:ResourceExtension**. Which will help us in the future to extend the behaviour of the reconciler for each resource. 

gif

**If applicable**:


matthchr:
I don't think this is related to #1977?

I think this "closes" #1978 - if you say that in the PR it should auto-link it to the issue and when this PR merges close that issue.

v2/tools/generator/internal/codegen/pipeline/create_resource_extension_types.go
```diff
@@ -0,0 +1,55 @@
...
+			extendedResourceTypes := make(astmodel.Types)
+			extendedResourceTypesMapping := make(map[astmodel.TypeName][]astmodel.TypeName)
+			resourcePackageRef := make(map[astmodel.TypeName][]astmodel.PackageReference)
+			resourceTypes := astmodel.FindResourceTypes(types)
+
+			//iterate through resource types and aggregate the similar resource types and package refs in a map.
+			for _, typeDef := range resourceTypes {
+				group, _, _ := typeDef.Name().PackageReference.GroupVersion()
+				packageRef := astmodel.MakeLocalPackageReference(localPath, group, "extensions")
+				name := astmodel.MakeTypeName(packageRef, typeDef.Name().Name()+"Extension")
```
matthchr, line 32: minor: Maybe use a more descriptive name here like `extensionTypeName`?

v2/tools/generator/internal/codegen/pipeline/create_resource_extension_types.go
```diff
@@ -0,0 +1,55 @@
...
+
+func CreateResourceExtensions(localPath string, idFactory astmodel.IdentifierFactory) Stage {
+	return MakeLegacyStage(
+		CreateResourceExtensionsStageID,
+		"create Resource Extensions for each resource type",
+		func(ctx context.Context, types astmodel.Types) (astmodel.Types, error) {
+
+			extendedResourceTypes := make(astmodel.Types)
+			extendedResourceTypesMapping := make(map[astmodel.TypeName][]astmodel.TypeName)
+			resourcePackageRef := make(map[astmodel.TypeName][]astmodel.PackageReference)
```
matthchr, line 25: Don't you already have this information in the `extendedResourceTypesMapping`? You do:
```
extendedResourceTypesMapping[name] = append(extendedResourceTypesMapping[name], typeDef.Name())

//putting package references in map for each extension
resourcePackageRef[name] = append(resourcePackageRef[name], typeDef.Name().PackageReference)
```

but instead of using `resourcePackageRef` I think you can get at the same data by just iterating over the `[]astmodel.TypeName` (the value of `extendedResourceTypesMapping`) and calling `.PackageReference` on the `TypeName`?

I think you can just delete this map entirely

v2/tools/generator/internal/codegen/pipeline/create_resource_extension_types.go
```diff
@@ -0,0 +1,55 @@
...
+			//iterate through the extendedResources map and create a ResourceExtension type
+			for extensionName, extendedResources := range extendedResourceTypesMapping {
+				fn := functions.NewGetExtendedResourcesFunction(idFactory, extendedResources, resourcePackageRef[extensionName])
+
+				newType := astmodel.MakeTypeDefinition(
+					extensionName,
+					astmodel.NewObjectType().WithFunction(fn))
+
+				if err := extendedResourceTypes.AddAllowDuplicates(newType); err != nil {
+					return types, err
```
matthchr, line 48: minor: IMO return `nil, err` here, since in the error case we shouldn't be using the LHS of this return anyway.

v2/tools/generator/internal/codegen/pipeline/create_resource_extension_types.go
```diff
@@ -0,0 +1,55 @@
...
+				name := astmodel.MakeTypeName(packageRef, typeDef.Name().Name()+"Extension")
+
+				extendedResourceTypesMapping[name] = append(extendedResourceTypesMapping[name], typeDef.Name())
+				//putting package references in map for each extension
+				resourcePackageRef[name] = append(resourcePackageRef[name], typeDef.Name().PackageReference)
+			}
+
+			//iterate through the extendedResources map and create a ResourceExtension type
+			for extensionName, extendedResources := range extendedResourceTypesMapping {
+				fn := functions.NewGetExtendedResourcesFunction(idFactory, extendedResources, resourcePackageRef[extensionName])
```
matthchr, line 41: As mentioned above I think you can just pass `extendedResources` here and get the refs from that by calling `.PackageReference` on each item.

v2/tools/generator/internal/codegen/pipeline/create_resource_extension_types.go
```diff
@@ -0,0 +1,55 @@
...
+const CreateResourceExtensionsStageID = "createResourceExtensions"
+
+func CreateResourceExtensions(localPath string, idFactory astmodel.IdentifierFactory) Stage {
+	return MakeLegacyStage(
+		CreateResourceExtensionsStageID,
+		"create Resource Extensions for each resource type",
+		func(ctx context.Context, types astmodel.Types) (astmodel.Types, error) {
+
+			extendedResourceTypes := make(astmodel.Types)
+			extendedResourceTypesMapping := make(map[astmodel.TypeName][]astmodel.TypeName)
```
matthchr, line 24: minor: A comment describing what this map is between would be helpful.

I think (from reading the code) it's a map of the new extension type name, to all of the resource type names (across versions) which the extension applies to.

v2/tools/generator/internal/codegen/pipeline/resource_registration_file.go
```diff
@@ -132,6 +143,8 @@ func (r *ResourceRegistrationFile) generateImports() *astmodel.PackageImportSet
 
 	// We require these imports
 	clientGoSchemeImport := astmodel.NewPackageImport(astmodel.ClientGoSchemeReference).WithName("clientgoscheme")
+	genRuntimeReferenceImport := astmodel.NewPackageImport(astmodel.GenRuntimeReference)
```
matthchr, line 146: minor, style: I'd just do these 2 lines all in 1 like the other lines above/below are doing

v2/tools/generator/internal/codegen/pipeline/resource_registration_file.go
```diff
@@ -378,6 +391,75 @@ func createGetKnownStorageTypesFunc(
...
 }
 
+func (r *ResourceRegistrationFile) createGetResourceExtensions(context *astmodel.CodeGenerationContext) (dst.Decl, error) {
+
+	return createGetResourceExtensionsImpl(
+		context,
+		r.resourceExtensions,
+		"getResourceExtensions",
+		"returns a list of Resource Extension types")
+
```
matthchr, line 401: minor: Extra newline

v2/tools/generator/internal/codegen/pipeline/resource_registration_file.go
```diff
@@ -378,6 +391,75 @@ func createGetKnownStorageTypesFunc(
...
+	sort.Slice(resources, orderByImportedTypeName(codeGenerationContext, resources))
+	packageRef, err := codeGenerationContext.GetImportedPackageName(astmodel.GenRuntimeReference)
+	if err != nil {
+		return nil, err
+	}
+	resultIdent := dst.NewIdent("result")
+	resultVar := astbuilder.LocalVariableDeclaration(
+		resultIdent.String(),
+		&dst.ArrayType{
+			Elt: &dst.SelectorExpr{
```
matthchr, line 415: So instead of building this `SelectorExpr` yourself, it's actually cleaner to define a `TypeName` in `astmodel` in the `std_references.go` file. Like this: `SecretReferenceType             = MakeTypeName(GenRuntimeReference, "SecretReference")`

Then you can just use the type by doing `astmodel.SecretReferenceType.AsType(codeGenerationContext)` rather than having to import the pkgRef and build the selector yourself.

matthchr: I think this will save you from having to do this above too:
```
	packageRef, err := codeGenerationContext.GetImportedPackageName(astmodel.GenRuntimeReference)
	if err != nil {
		return nil, err
	}
```

v2/tools/generator/internal/codegen/pipeline/resource_registration_file.go
```diff
@@ -378,6 +391,75 @@ func createGetKnownStorageTypesFunc(
...
+	var resourceAppendStatements []dst.Stmt
+	for _, typeName := range resources {
+		appendStmt := astbuilder.AppendSlice(
+			resultIdent,
+			astbuilder.AddrOf(astbuilder.NewCompositeLiteralBuilder(typeName.AsType(codeGenerationContext)).Build()),
+		)
+		resourceAppendStatements = append(resourceAppendStatements, appendStmt)
+	}
+
+	returnStmt := &dst.ReturnStmt{
```
matthchr, line 431: minor: I think we have an `astbuilder.Returns` to make this a bit cleaner.

v2/tools/generator/internal/codegen/pipeline/create_resource_extension_types.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/Azure/azure-service-operator/v2/tools/generator/internal/functions"
+)
+
+// CreateResourceExtensionsStageID is the unique identifier of this stage
+const CreateResourceExtensionsStageID = "createResourceExtensions"
+
+func CreateResourceExtensions(localPath string, idFactory astmodel.IdentifierFactory) Stage {
+	return MakeLegacyStage(
+		CreateResourceExtensionsStageID,
+		"create Resource Extensions for each resource type",
```
matthchr, line 20: minor:
```suggestion
		"Create Resource Extensions for each resource type",
```

v2/tools/generator/internal/functions/get_extended_resources_function.go
```diff
@@ -0,0 +1,108 @@
...
+	idFactory  astmodel.IdentifierFactory
+	resources  []astmodel.TypeName
+	packageRef *astmodel.PackageReferenceSet
+}
+
+// Ensure GetExtendedResources properly implements Function
+var _ astmodel.Function = &GetExtendedResourcesFunction{}
+
+// NewGetExtendedResourcesFunction creates a new GetExtendedResources
+func NewGetExtendedResourcesFunction(idFactory astmodel.IdentifierFactory, resources []astmodel.TypeName, references []astmodel.PackageReference) *GetExtendedResourcesFunction {
```
matthchr, line 29: As I mentioned elsewhere, I think you can build the []`astmodel.PackageReference` from the `[]astmodel.TypeName`'s in this case, so I don't think you need both

matthchr (changes requested):
Looks pretty good. Had a few minor cleanup things, especially about what is getting generated: I think we can reduce the amount of generated code quite a bit, as the DeepCopy stuff and the `scheme` stuff we don't need (just need the actual extension types themselves)

v2/api/authorization/extensions/groupversion_info_gen.go
```diff
@@ -0,0 +1,31 @@
+/*
```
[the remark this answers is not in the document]
super-harsh: In the feature spec document by @theunrepentantgeek, the package name suggested was `extensions` and it made more sense to have the package being called as `extensions` rather than `customizations`.

v2/internal/controllers/controller_resources.go
```diff
@@ -50,6 +51,18 @@ func CreateScheme() *runtime.Scheme {
 	return scheme
 }
 
+//GetResourceExtensions returns a map between resource and resource extension
+func GetResourceExtensions() map[string]genruntime.ResourceExtension {
```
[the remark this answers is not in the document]
theunrepentantgeek: The difference in return types comes from the `1-n` relationship between a single extension type and the (potentially many) resource types it extends. Generating the code to return a slice of extended resources is simple, but the reconciler is going to need a map from resource -> extension to look things up.

I think the dictionary needs to be keyed by the GVK of the resource, not a string.

v2/tools/generator/internal/codegen/code_generator.go
```diff
@@ -179,6 +179,9 @@ func createAllPipelineStages(idFactory astmodel.IdentifierFactory, configuration
 		pipeline.ImplementConvertibleStatusInterface(idFactory).UsedFor(pipeline.ARMTarget),
 		pipeline.InjectOriginalGVKFunction(idFactory).UsedFor(pipeline.ARMTarget),
 
+		//Create Resource Extensions
```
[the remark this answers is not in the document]
theunrepentantgeek: Our convention is to use a space and leading capitalization in our comments.

General Go convention is to do that, _unless_ the comment is a compiler directive, in which case _no space_.

v2/tools/generator/internal/codegen/pipeline/create_resource_extension_types.go
theunrepentantgeek: I agree - and deriving the set of package references from the set of types guarantees consistency.

v2/tools/generator/internal/codegen/pipeline/resource_registration_file.go
theunrepentantgeek: Yes, it'll become a (short!) one liner:

``` go
returnStmt := astbuilder.Returns(resultIdent)
```

v2/tools/generator/internal/codegen/pipeline/resource_registration_file.go
```diff
@@ -378,6 +391,75 @@ func createGetKnownStorageTypesFunc(
...
+			{
+				Type: &dst.ArrayType{
+					Elt: &dst.SelectorExpr{
+						X:   dst.NewIdent(packageRef),
+						Sel: dst.NewIdent("ResourceExtension"),
+					},
+				},
+			},
+		},
+	}
```
theunrepentantgeek, line 456: Lean in to the helper methods, they reduce boilerplate. 
```suggestion
	f := &astbuilder.FuncDetails{
		Name:   funcName,
		Body:   body,
	}
	f.AddReturn(astbuilder.QualifiedTypeName(packageRef, "ResourceExtension"))
```

v2/tools/generator/internal/functions/get_extended_resources_function.go
theunrepentantgeek: I'd suggest removing `references` from the factory func, and you don't need to keep it on the struct either - just create the `PackageReferenceSet` on the fly inside `RequiredPackageReferences()`.

theunrepentantgeek (changes requested):

v2/api/authorization/extensions/groupversion_info_gen.go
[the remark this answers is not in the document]
theunrepentantgeek: The subpackage for the interface definitions was spec'd as `genruntime/extensions`, but to avoid name colllisions and other issues, the non-versioned packages were spec'd as `batch/customization`.

So, you're both right. :grin:

v2/internal/controllers/controller_resources.go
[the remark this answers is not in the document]
matthchr: +1 to keyed by `schema.GroupVersionKind`. That's what I ended up doing for my hacky secrets work locally, as you can't use `client.Object` as a lookup due to ptr equality rules.

v2/pkg/genruntime/kubernetes_resource.go
```diff
@@ -37,6 +38,9 @@ type KubernetesResource interface {
 	// 	SetAzureName(name string)
 	// They do not if the name must be a fixed value (like 'default').
 
+	//GroupVersionKind returns a GroupVersionKind of the resource
+	GroupVersionKind() schema.GroupVersionKind
```
matthchr, line 42: Where is the actual implementation of this method?

I _think_ this isn't going to do what you want because of the way that k8s populates the GVK. Basically it strips it from the object when it deserializes it, and it's also empty for empty objects, so you'll get an empty struct result. You should be able to test it in a debugger and check.

The way to get the GVK from a given `client.Object` is using the `scheme`, like this: `gvk, err := apiutil.GVKForObject(obj, scheme)`.

I _think_ that's what you'll end up having to do. This is all basically because the GVK is just a 3-tuple to identify the type, so they assume that if you already have the type you basically don't need it. In our case we do for things like this so we have to jump through a few hoops to actually get it.

v2/tools/generator/internal/codegen/pipeline/export_controller_type_registrations.go
```diff
@@ -30,32 +30,38 @@ func ExportControllerResourceRegistrations(idFactory astmodel.IdentifierFactory,
...
+						resourceIndexFunctions, resourceSecretPropertyKeys := handleSecretPropertyChains(chains, idFactory, def)
+						indexFunctions[def.Name()] = resourceIndexFunctions
+						secretPropertyKeys[def.Name()] = resourceSecretPropertyKeys
 					}
 
-					resourceIndexFunctions, resourceSecretPropertyKeys := handleSecretPropertyChains(chains, idFactory, def)
-					indexFunctions[def.Name()] = resourceIndexFunctions
-					secretPropertyKeys[def.Name()] = resourceSecretPropertyKeys
+					resources = append(resources, def.Name())
+				} else if resourceExtension, ok := astmodel.AsObjectType(def.Type()); ok {
```
matthchr, line 56: minor: you don't know it's a `resourceExtension` at this point, it's just an `objectType`?

v2/tools/generator/internal/functions/get_extended_resources_function.go
```diff
@@ -0,0 +1,110 @@
...
+func NewGetExtendedResourcesFunction(idFactory astmodel.IdentifierFactory, resources []astmodel.TypeName) *GetExtendedResourcesFunction {
+
+	result := &GetExtendedResourcesFunction{
+		idFactory: idFactory,
+		resources: sortResources(resources),
+	}
+	return result
+}
+
+//getPackageRefs iterates through the resources and returns package references
```
matthchr, line 37: minor, style: spaces before comments here too (and below -- won't repeat though).

v2/tools/generator/internal/functions/get_extended_resources_function.go
```diff
@@ -0,0 +1,110 @@
...
+	funcDetails.AddComments("Returns the KubernetesResource slice for Resource versions")
+	funcDetails.AddReturn(krType)
+
+	return funcDetails.DefineFunc()
+}
+
+// Equals returns true if the passed function is equal textus, or false otherwise
+func (ext *GetExtendedResourcesFunction) Equals(f astmodel.Function, _ astmodel.EqualityOverrides) bool {
+	_, ok := f.(*GetExtendedResourcesFunction)
+	// Equality is just based on Type for now
```
matthchr, line 108: minor: Not that this is really used, but might not hurt to at least check that the `ext.resources` are the same?

v2/tools/generator/internal/test/resource.go
```diff
@@ -65,3 +78,9 @@ func CreateSimpleResource(
 	status := CreateStatus(pkg, name)
 	return CreateResource(pkg, name, spec, status)
 }
+
+func CreateTypeName(
```
matthchr, line 82: This method doesn't seem to add a ton, since it's just calling `astmodel.MakeTypeName`? Why not just call `astmodel.MakeTypeName` wherever you're using this and delete this method?

matthchr (commented):
Great job @super-harsh. I've left a bunch more nits, but looks good to me overall.

I held off on approval _just_ because of the one comment about `scheme` + `GroupVersionKind` which I _think_ is an issue, but you're basically there I think.

v2/tools/generator/internal/astmodel/package_definition.go
```diff
@@ -228,6 +231,7 @@ var (
 `))
 
 func emitGroupVersionFile(pkgDef *PackageDefinition, outputDir string) error {
+
```
theunrepentantgeek, line 234: Extra blank line

v2/tools/generator/internal/codegen/pipeline/create_resource_extension_types.go
```diff
@@ -0,0 +1,56 @@
...
+		CreateResourceExtensionsStageID,
+		"Create Resource Extensions for each resource type",
+		func(ctx context.Context, state *State) (*State, error) {
+
+			// Map of the new extension types, to all the resource types names on which the extension applies to
+			extendedResourceTypesMapping := make(map[astmodel.TypeName][]astmodel.TypeName)
+			extendedResourceTypes := make(astmodel.Types)
+			resourceTypes := astmodel.FindResourceTypes(state.types)
+
+			// Iterate through resource types and aggregate the resource types that share the same extension in a map.
```
theunrepentantgeek, line 30: When I first read this I thought "extension" meant a common suffix on the type names. 
```suggestion
			// Iterate through resource types and aggregate the resource types that share the same extension type in a map.
```

v2/tools/generator/internal/codegen/pipeline/create_resource_extension_types.go
```diff
@@ -0,0 +1,56 @@
...
+			for _, typeDef := range resourceTypes {
+				group, _, _ := typeDef.Name().PackageReference.GroupVersion()
+				packageRef := astmodel.MakeLocalPackageReference(localPath, group, "extensions")
+				extensionTypeName := astmodel.MakeTypeName(packageRef, typeDef.Name().Name()+"Extension")
+				extendedResourceTypesMapping[extensionTypeName] = append(extendedResourceTypesMapping[extensionTypeName], typeDef.Name())
+			}
+
+			// Iterate through the extendedResources map and create a ResourceExtension type
```
theunrepentantgeek, line 38:
```suggestion
			// For each resource extension type, create a registration function
```

v2/tools/generator/internal/codegen/pipeline/create_resource_extension_types_test.go
```diff
@@ -0,0 +1,67 @@
...
+	types.AddAll(resourceV1, specV1, statusV1, resourceV2, specV2, statusV2)
+
+	initialState, err := RunTestPipeline(
+		NewState().WithTypes(types))
+	g.Expect(err).To(Succeed())
+
+	finalState, err := RunTestPipeline(initialState, CreateResourceExtensions("testPath", astmodel.NewIdentifierFactory()))
+	g.Expect(err).To(Succeed())
+
+	test.AssertPackagesGenerateExpectedCode(t, finalState.Types())
```
theunrepentantgeek, line 56: ❤️

theunrepentantgeek: Have you committed the Golden files generated by this assert?

v2/tools/generator/internal/codegen/pipeline/json_serialization_test_cases.go
```diff
@@ -45,7 +45,7 @@ func InjectJsonSerializationTests(idFactory astmodel.IdentifierFactory) Stage {
 			return state.WithTypes(state.Types().OverlayWith(modifiedTypes)), nil
 		})
 
-	return stage.RequiresPostrequisiteStages("simplifyDefinitions" /* needs flags */)
+	return stage.RequiresPostrequisiteStages("simplifyDefinitions", CreateResourceExtensionsStageID /* needs flags */)
```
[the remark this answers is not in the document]
theunrepentantgeek: The reason we have both prerequisite and postrequisite stages is so that we can put the restriction on the stage that cares. In this case, it's the `CreateResourceExtensionsStage` that cares about not having tests; the test stages don't give two hoots.

v2/tools/generator/internal/codegen/pipeline/create_resource_extension_types_test.go
theunrepentantgeek: Yes, you have - I just found them. 😁

v2/tools/generator/internal/functions/get_extended_resources_function.go
```diff
@@ -0,0 +1,114 @@
...
+		resources: sortResources(resources),
+	}
+	return result
+}
+
+// getPackageRefs iterates through the resources and returns package references
+func getPackageRefs(resources []astmodel.TypeName) []astmodel.PackageReference {
+	var packageRefs []astmodel.PackageReference
+	// Package reference for return type
+	packageRefs = append(packageRefs, astmodel.KubernetesResourceType.PackageReference)
```
theunrepentantgeek, line 42: Nice. This guarantees consistency even if we later move things around.

v2/tools/generator/internal/functions/get_extended_resources_function.go
```diff
@@ -0,0 +1,114 @@
...
+
+	for _, typeDef := range resources {
+		packageRefs = append(packageRefs, typeDef.PackageReference)
+	}
+	return packageRefs
```
theunrepentantgeek, line 47: Nitpit: I like to have blank lines between blocks and before return statements so the return statement is easier to spot.

theunrepentantgeek (commented):
This is looking really good.

v2/tools/generator/internal/astmodel/package_definition.go
super-harsh: Thats `gofmt` doing its job :/

v2/api/authorization/extensions/role_assignment_extension_types_gen.go
```diff
@@ -0,0 +1,20 @@
...
+	authorization "github.com/Azure/azure-service-operator/v2/api/authorization/v1alpha1api20200801preview"
+	"github.com/Azure/azure-service-operator/v2/api/authorization/v1alpha1api20200801previewstorage"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+)
+
+type RoleAssignmentExtension struct {
+}
+
+// GetExtendedResources Returns the KubernetesResource slice for Resource versions
+func (extension *RoleAssignmentExtension) GetExtendedResources() []genruntime.KubernetesResource {
```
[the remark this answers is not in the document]
super-harsh: I have those tests in `functions/get_extended_resources_function_test.go` and `codegen/pipeline/create_resource_extension_types_test.go`

v2/tools/generator/internal/codegen/code_generator.go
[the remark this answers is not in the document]
super-harsh: `gofmt` doesn't check for spacing in comments I guess?

v2/tools/generator/internal/codegen/pipeline/resource_registration_file.go
```diff
@@ -378,6 +389,52 @@ func createGetKnownStorageTypesFunc(
 	return f.DefineFunc()
 }
 
+func (r *ResourceRegistrationFile) createGetResourceExtensions(context *astmodel.CodeGenerationContext) (dst.Decl, error) {
+
```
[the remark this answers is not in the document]
super-harsh: I've merged `createGetResourceExtensionsImpl` into the method above.

v2/tools/generator/internal/codegen/pipeline/resource_registration_file.go
```diff
@@ -378,6 +389,52 @@ func createGetKnownStorageTypesFunc(
...
+	return createGetResourceExtensionsImpl(
+		context,
+		r.resourceExtensions,
+		"getResourceExtensions",
+		"returns a list of resource extensions")
+}
+
+func createGetResourceExtensionsImpl(codeGenerationContext *astmodel.CodeGenerationContext, resources []astmodel.TypeName, funcName string, funcComment string) (dst.Decl, error) {
```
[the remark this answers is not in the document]
super-harsh: This method no longer exists!

v2/api/authorization/extensions/groupversion_info_gen.go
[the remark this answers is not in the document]
super-harsh: Have changed the package to be called `customizations`.

v2/internal/controllers/controller_resources.go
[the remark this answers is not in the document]
super-harsh: Have pushed changes regarding this one. The map returned now is  `typeGVK:typeExtension`

v2/tools/generator/internal/codegen/code_generator.go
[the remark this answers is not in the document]
super-harsh: Done!

v2/tools/generator/internal/codegen/pipeline/create_resource_extension_types.go
super-harsh: have made the changes above! now looks cleaner and better :)

v2/pkg/genruntime/kubernetes_resource.go
super-harsh: have removed the above! Thought it might work with getting the GVK, however, yes you're right - it doesn't have any implementation hence gonna be empty always.

v2/tools/generator/internal/functions/get_extended_resources_function.go
super-harsh: Implemented the deep compare on this one.. since resources are sorted beforehand being assigned into the struct, I'm comparing them using `reflect.DeepEqual` in this block.

v2/tools/generator/internal/test/resource.go
super-harsh: Removed this method, thought we have to have helper methods in the test package. Have made the changed to use `astmodel.MakeTypeName` .

v2/tools/generator/internal/codegen/pipeline/resource_registration_file.go
```diff
@@ -378,6 +389,55 @@ func createGetKnownStorageTypesFunc(
...
 
+func (r *ResourceRegistrationFile) createGetResourceExtensions(context *astmodel.CodeGenerationContext) (dst.Decl, error) {
+	return createGetResourceExtensionsImpl(
+		context,
+		r.resourceExtensions,
+		"getResourceExtensions",
+		"returns a list of resource extensions")
+}
+
+func createGetResourceExtensionsImpl(
```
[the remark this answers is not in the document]
super-harsh: have merged the above methods!

v2/tools/generator/internal/codegen/pipeline/export_controller_type_registrations.go
super-harsh: Yep, you're right. Made it to be called `object`.

v2/tools/generator/internal/codegen/pipeline/property_assignment_test_cases.go
```diff
@@ -49,7 +49,7 @@ func InjectPropertyAssignmentTests(idFactory astmodel.IdentifierFactory) Stage {
 		InjectJsonSerializationTestsID,           // We reuse the generators from the JSON tests
 	)
 
-	return stage
+	return stage.RequiresPostrequisiteStages(CreateResourceExtensionsStageID)
```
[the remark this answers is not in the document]
super-harsh: Have removed the above, doing `RequiresPrerequesiteStages` check inside CreateResourceExtensions stage.

v2/pkg/genruntime/kubernetes_resource.go
```diff
@@ -6,11 +6,9 @@
 package genruntime
 
 import (
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/conditions"
```
matthchr, line 9: minor: Fix ordering, this should be sorted to the bottom.

I think there's an option in goland to do this automatically for you.

v2/tools/generator/internal/codegen/pipeline/create_resource_extension_types.go
```diff
@@ -17,38 +17,40 @@ const (
...
 				}
 
 			}
-			types.AddTypes(extendedResourceTypes)
-			return types, nil
+			state.types.AddTypes(extendedResourceTypes)
+			return state, nil
 		})
+
+	return stage.RequiresPrerequisiteStages(InjectJsonSerializationTestsID, InjectPropertyAssignmentTestsID)
```
matthchr, line 55: minor: Consider leaving a comment explaining why this ordering is required.

v2/internal/controllers/controller_resources.go
```diff
@@ -52,16 +53,31 @@ func CreateScheme() *runtime.Scheme {
...
 	extensionMapping := make(map[schema.GroupVersionKind]genruntime.ResourceExtension)
 
 	for _, extension := range getResourceExtensions() {
 		for _, resource := range extension.GetExtendedResources() {
-			extensionMapping[resource.GroupVersionKind()] = extension
+
+			// Make sure the type casting goes well, and we can extract the GVK successfully.
+			resourceObj, ok := resource.(runtime.Object)
+			if !ok {
+				log.V(Status).Info("Unexpected resource type", "resource", resource.AzureName(), "actual", fmt.Sprintf("%T", resource))
```
matthchr, line 68: Instead of logging here, just have this return `(map[schema.GroupVersionKind]genruntime.ResourceExtension, error)`.

Instead of logging, use `errors.Errorf` and create an error with the same message you were logging here. This gives the caller the opportunity to choose the logger to direct this to, which we need because these methods are called in different contexts (both in our tests w/ the test logger and in our actual controller with its logger).

v2/tools/generator/internal/codegen/pipeline/resource_registration_file.go
```diff
@@ -428,22 +422,12 @@ func createGetResourceExtensionsImpl(codeGenerationContext *astmodel.CodeGenerat
...
-	f.AddReturn(astmodel.NewArrayType(astmodel.ResourceExtensionType).AsType(codeGenerationContext))
+	f.AddReturn(astmodel.NewArrayType(astmodel.ResourceExtensionType).AsType(context))
 	f.AddComments(funcComment)
 
-	return f.DefineFunc(), nil
-
+	return f.DefineFunc()
 }
 
-// createCreateSchemeFunc creates a createScheme() function like:
```
matthchr, line 381: why delete this comment?

super-harsh: Thats my bad. Apologies!

v2/pkg/genruntime/kubernetes_resource.go
super-harsh: ah! this is what goland did it for me 😆. I'll make it align to our coding convention.

v2/tools/generator/internal/codegen/pipeline/resource_registration_file.go
super-harsh: Might've been the result of merge from `main` I guess. I remember having a massive merge conflict in this file.

matthchr (approved):
Nice job!

theunrepentantgeek (approved):

PR #2065, opened 2022-02-02
Add documentation to config file

theunrepentantgeek:

Adds some documentation to our configuration file to help someone new to understand what they're reading. 

To Do

- [x] Review and revise wording
- [x] Update `azure-crossplane.yaml` with the same details

gif

**If applicable**:

### DIFF
--- a/v2/azure-arm.yaml
+++ b/v2/azure-arm.yaml
@@ -805,6 +805,9 @@ status:
 #     type:           // A resource or subtype (similar to kubernetes 'kind')
 #       property:     // An attribute found on an object type
 #
+# Please add new elements to this configuration in ALPHABETICAL order to make 
+# it easier for future maintainers to scan for the item they want.
+# 
 # At specific levels, optional modifiers may be injected.
 # All modifiers start with `$` in order to avoid clashes with ARM object and 
 # property names
--- a/v2/azure-arm.yaml
+++ b/v2/azure-arm.yaml
@@ -3,11 +3,22 @@ schemaUrlRewrite:
   from: 'https://schema.management.azure.com/schemas/'
   to: 'specs/azure-resource-manager-schemas/schemas/'
 destinationGoModuleFile: go.mod
-# These two paths are relative to the module path above.
+
+# typesOutputPath specifies the output folder name, relative to the directory containing the mod file above.
 typesOutputPath: api
+
+# typeRegistrationOutputFile specifies the output file name for registration code, relative to the directory 
+# containing the mod file above.
 typeRegistrationOutputFile: internal/controllers/controller_resources_gen.go
+
+# samplesUrl is the base URL used for linking to samples when we generate a list of supported resources
 samplesUrl: https://github.com/Azure/azure-service-operator/tree/main/v2/config/samples
+
 pipeline: azure
+
+#
+# These filters are used early in the processing pipeline to simplify 
+#
 typeFilters:
   - action: prune
     group: definitions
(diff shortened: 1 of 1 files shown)


matthchr (approved):

v2/azure-arm.yaml
```diff
@@ -3,11 +3,22 @@ schemaUrlRewrite:
...
+# containing the mod file above.
 typeRegistrationOutputFile: internal/controllers/controller_resources_gen.go
+
+# samplesUrl is the base URL used for linking to samples when we generate a list of supported resources
 samplesUrl: https://github.com/Azure/azure-service-operator/tree/main/v2/config/samples
+
 pipeline: azure
+
+#
+# These filters are used early in the processing pipeline to simplify
```
matthchr, line 20: to simplify what?

v2/azure-arm.yaml
```diff
@@ -814,22 +826,35 @@ status:
 #
 # Available 'type' modifiers:
 #
-#   $renamedTo: <string>  // Gives the name this type will have in the next 
-#                         // version of the resource
+# $export: <bool>
+#     Requests that support for this resource type be generated.
+#     Automatically includes all other types required for this resource
+#     If not specified, false is assumed.
```
matthchr, line 832: minor: "assumed" here feels more wishy-washy to me than you possibly intend?

maybe: "If not specified, the default is false."

v2/azure-arm.yaml
```diff
@@ -814,22 +826,35 @@ status:
...
+#     Set to 'true' to confirm the property is an ARM reference.
+#     Set to 'false' to disable our heuristics if a property is incorrectly 
+#     identified as an ARM reference.
 #
-#   $isSecret: <bool>     // Specifies that the property references a secret
-#                         // Set to `true` to integrate with ASO custom secret
-#                         // management.
+# $isSecret: <bool>
+#     Specifies that the property references a secret.
+#     Set to `true` to integrate with ASO custom secret management.
```
matthchr, line 854: Rather than "to integrate with ASO custom secret management" maybe say "to flag this property as a secret. Secrets are string values read from a Kubernetes secret via a SecretReference" (or something?)

Basically just saying a bit more about what exactly "ASO custom secret management" is. Alternatively could leave as is and link to a doc.

v2/azure-arm.yaml
```diff
@@ -814,22 +826,35 @@ status:
...
 #
-#   $renamedTo: <string>  // Gives the name this property will have in the 
-#                         // next version of the resource
+# $nameInNextVersion: <string>
+#     Gives the name this property will have in the next version of the resource
```
matthchr, line 857: Will have, or _has_? Is it a directive to change the name or a directive that the name _is_ a particular value? It sorta sounds like the former but I think you actually mean the latter.

matthchr (approved):
Few little suggestions

v2/azure-arm.yaml
theunrepentantgeek: _Everything_. 😁 

I'd made changes but not saved them. Will update.

v2/azure-arm.yaml
theunrepentantgeek: Fixed.

v2/azure-arm.yaml
theunrepentantgeek: I don't want this comment to be too specific about how secrets are managed; I've made it more explicit without going into details.

v2/azure-arm.yaml
theunrepentantgeek: Clarified.

Issue #1465, opened 2020-02-09
Enable bring your own resource


devigned:
A user should be able to create a K8s resource of any kind. If that resource already exists, then the resource should be owned by the user, not by the controller. Ownership of a resource implies the owner is responsible for the lifecycle of the resource. For example, an owner of a resource is responsible for deletion of the resource, and if a user brings their own resource, the controller should not delete the Azure resource when the K8s resource is deleted.

Transitively, a resource group should not be deleted if it contains resources which are not owned by the owner of the resource group. These resources need not be represented by K8s resources, but do exist in Azure. This implies that there may need to be a mechanism for a K8s resource to check in Azure for child resources owned by other entities.

matthchr:
See the above linked adoption design.

theunrepentantgeek:
We need to finish the adoption design and make this possible.

matthchr:
The adoption design mentioned above still covers this, although technically it's doable today with `reconcile-policy: skip` already.

matthchr:
We're still working on this, we're tracking this through the Adoption PR so closing this

PR #2067, opened 2022-02-04
Adoption design

matthchr:
A design for how the operator will "adopt" resources that existed in Azure before they were created in Kubernetes.

### DIFF
--- a/docs/hugo/content/design/ADR-2023-02-Adoption-Policy.md
+++ b/docs/hugo/content/design/ADR-2023-02-Adoption-Policy.md
@@ -1,0 +1,275 @@
+# Adoption Policy
+
+## Context
+
+Resources which exist in Azure prior to being created in Kubernetes are "adopted" by the operator when the resource
+is created in Kubernetes. Today this adoption process is automatic and implicit. As a user there isn't an easy way
+to tell if a resource you're about to `kubectl apply` already exists in Azure (will be adopted) or does not exist in
+Azure (will be created).
+
+We already have support for [reconcile-policy](https://github.com/Azure/azure-service-operator/pull/2060) which
+is related to adoption. Some of `reconcile-policy`'s primary use cases are for when the user does not want ASO to
+fully manage an adopted resource.
+
+The following are principles we should follow related to resource adoption:
+
+1. **The operator should adopt resources by default:** This is standard goal-seeking/desired-state behavior. 
+   If you create a resource in Kubernetes that says: "I expect there to be a Virtual Machine that looks like this
+   at location X" and that Virtual Machine already exists, then we are in the desired state.
+   The alternative would be to error if the resource already exists, which goes against the Kubernetes model.
+
+2. **Users must be able to instruct the operator to stop making changes to the backing Azure resource:**
+   This is enabled by `reconcile-policy`.
+
+   This is useful for the following scenarios:
+   - The operator is doing something wrong and the user needs a way to stop the operator from acting.
+   - Pausing the operator temporarily while they make some manual fix via the Azure portal.
+   - Creating a resource in ASO purely to get ownership references to work. The resource in question is not managed
+     by ASO at all. It was created and managed some other way (portal, terraform, az-cli, etc).
+
+3. **Users must be able to instruct the operator to skip deletion of the backing resource:**
+   This is useful for scenarios such as moving ownership of resources from ASO in cluster A to ASO in cluster B,
+   or as a safety net to protect stateful resources such as databases from accidental deletion.
+   This is enabled by `reconcile-policy`.
+
+### Open question: what should the default `reconcile-policy` be for resources which are adopted?
+
+There are two main options.
+
+#### Option 1: The default `reconcile-policy` for adopted resources is `manage`
+
+`manage` is also the default `reconcile-policy` for resources ASO creates, so there is no difference between
+creation and adoption with this option.
+This means ASO automatically adopts, updates, and will delete backing Azure resources as normal unless the user
+explicitly sets `reconcile-policy`.
+
+**Advantages:**
+- It's easy to explain what the behavior is. Simplistic users don't have to understand `reconcile-policy` (they won't
+  ever see it unless they set it themselves).
+- Matches ARM/BICEP, raw REST, and SDK behavior, as well as some of our competitors (Crossplane and Google Config Connector)
+  as spelled out below.
+- Is the current default
+
+**Disadvantages:**
+- Users have no good way to know their resource existed previously. They may `kubectl delete` it without realizing that
+  other things depend on this resource. Deletion (and some mutation) is destructive in Azure so lost data, etc cannot 
+  easily be recovered.
+
+**Thoughts/comments:**
+
+This sort-of matches the behavior of ARM templates and SDKs. If you call the `DELETE` API the resource is gone.
+
+Some discussion of this topic has happened over in [ACK](https://github.com/aws-controllers-k8s/community/issues/82).
+
+[Crossplane](https://github.com/crossplane/crossplane/issues/2590) doesn't support any notion of `reconcile-policy` at all.
+`kubectl delete` always deletes the backing resource no matter what.
+
+[GCP](https://cloud.google.com/config-connector/docs/how-to/managing-deleting-resources#creating_a_resource) does it this way:
+> When you create a resource, Config Connector creates the resource if it doesn't exist. If a Google Cloud resource 
+> already exists with the same name, then Config Connector acquires the resource and manages it.
+
+GCP also [supports](https://cloud.google.com/config-connector/docs/how-to/managing-deleting-resources) a 
+`reconcile-policy`-like annotation, so doing it like this would basically be doing it exactly like GCP.
+
+Kubernetes has a similar concept for `StorageClasses`/`PersistentVolumes` called 
+[reclaimPolicy](https://kubernetes.io/docs/concepts/storage/storage-classes/#reclaim-policy), whose default
+is `Delete` for dynamically provisioned PVs, but `Retain` for manually created PVs.
+
+#### Option 2: The default `reconcile-policy` for adopted resource is `abandon-on-delete`.
+
+ASO detects that a resource previously existed and sets `reconcile-policy` automatically to `abandon-on-delete`.
+
+**Advantages:**
+- Defaults to a more conservative `reconcile-policy` to protect users from accidentally deleting things they may not
+  have wanted to delete.
+
+**Disadvantages:**
+- Higher complexity. Users will be exposed to `reconcile-policy` as set by the operator (when a resource is adopted)
+  so need to understand it.
+- Bias towards safety may cause resource user thinks was deleted to continue existing (and billing) in Azure. Not a data
+  loss though.
+- `reconcile-policy` would have split ownership between user and operator, which means deciding if
+  we should trigger reconciliation events on its change is tricky (we don't want our defaulting of it to trigger an event).
+
+**Thoughts/comments:**
+
+There's some discussion of a similar feature over in [ACK](https://github.com/aws-controllers-k8s/community/pull/1148#discussion_r794605821).
+Jay Pipes had an interesting comment:
+> In the case of ACK and the user experience of resource deletion, I posit that the behaviour of our interface has 
+> no surprises and is actually hard to misuse. It doesn't have surprises because our interface to delete a resource 
+> is to call kubectl delete on that resource. This is as non-surprising as it gets. It is hard to misuse because 
+> calling kubectl delete is an explicit action the user has to take and it's not a confusing verb or semantic. 
+> "delete" means to remove the resource.
+
+It's important to note that ACK does adoption differently than ASO though (see 
+[example](https://github.com/aws-controllers-k8s/sagemaker-controller#60-adopt-resources)), so adoption is much more
+explicit for them. We don't want to have such explicit adoption though as covered in 
+principle #1: The operator should adopt resources by default
+
+CAPZ does something sort-of like this using Azure `tags`, but only for a select set of resources which they support
+"BYO-resource" for. Things like VNET. Since some child resources don't support tags they apply the parent tag to
+all of the children automatically, which makes sense for their use-case but doesn't make as much sense for ASO.
+
+### Open question: Should the adoption behavior be configurable at the operator (global) or per-resource level?
+
+Some users (such as CAPZ) may want adoption semantics different than the defaults we choose. Even on 
+[the PR that added this design proposal](https://github.com/Azure/azure-service-operator/pull/2067) different users disagreed
+on the behavior they wanted.
+
+#### Option 3: No configuration at the operator or per-resource level
+
+**Advantages:**
+- Simple - value of this cannot be understated. The more configuration users _have_ to understand the harder the operator
+  becomes to adopt or understand.
+
+**Disadvantages:**
+- Fails to cater to some legitimate use-cases.
+
+#### Option 4: Allow configuration at the operator or per-resource level
+
+This option would just be configuring the behavior of the operator in case it found an existing resource
+in Azure. The behavior if ASO created the resource is unchanged.
+
+The 3 scenarios here are, if the resource already exists in Azure when ASO goes to create it:
+1. Fully adopt the resource, mutate it if its configuration is different in ASO than in Azure, and delete 
+   it when `kubectl delete` is run.
+2. Partially adopt the resource. Mutations of the resource would be performed by ASO but `kubectl delete` will only
+   delete the resource in Kubernetes and _not_ in Azure.
+3. Use the existing resource and don't make any changes to it to make it match the ASO shape, and do not delete it
+   when `kubectl delete` is run.
+
+This is discussed in [#2703](https://github.com/Azure/azure-service-operator/issues/2703)
+
+##### Option 4.1: adoption-policy annotation
+
+A new property `ADOPTION_POLICY` in the operator global secret, and an annotation on each
+resource `serviceoperator.azure.com/adoption-policy` would allow users to control this behavior.
+As with our other configurations, the more specific configuration would win.
+
+Possible `adoption-policy`'s are:
+- `adopt-if-exists`: Corresponding to scenario 1 above.
+- `detach-on-delete-if-exists`: Corresponding to scenario 2 above.
+- `skip-if-exists`: Corresponding to scenario 3 above.
+
+This would exist alongside the existing `reconcile-policy` annotation. The most restrictive of the two options would
+be used.
+The full matrix of possibilities (where there is overlap) is:
+
+| `reconcile-policy` | `adoption-policy`          | Resource already exists in Azure? | Behavior                                 |
+|--------------------|----------------------------|-----------------------------------|------------------------------------------|
+| Unspecified        | Unspecified                | Yes                               | Fully manages the resource¹              |
+| Unspecified        | skip-if-exists             | Yes                               | Skips managing the resource              |
+| Unspecified        | skip-if-exists             | No                                | Fully manages the resource               |
+| detach-on-delete   | skip-if-exists             | Yes                               | Skips managing the resource              |
+| detach-on-delete   | skip-if-exists             | No                                | Manages the resource, detaches on delete |
+| skip               | skip-if-exists             | Yes                               | Skips managing the resource              |
+| skip               | skip-if-exists             | No                                | Skips managing the resource              |
+| Unspecified        | detach-on-delete-if-exists | Yes                               | Manages the resource, detaches on delete |
+| Unspecified        | detach-on-delete-if-exists | No                                | Fully manages the resource               |
+| detach-on-delete   | detach-on-delete-if-exists | Yes                               | Manages the resource, detaches on delete |
+| detach-on-delete   | detach-on-delete-if-exists | No                                | Manages the resource, detaches on delete |
+| skip               | detach-on-delete-if-exists | Yes                               | Skips managing the resource              |
+| skip               | detach-on-delete-if-exists | No                                | Skips managing the resource              |
+
+¹: Actual behavior depends on earlier sections of this design
+
+**Advantages:**
+- ???
+
+**Disadvantages:**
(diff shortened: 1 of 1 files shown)


matthchr:
I'm honestly pretty torn on this one. @sakthi-vetrivel curious about your thoughts on this as well.

docs/hugo/content/architecture-decision-records/ADR-2022-02-Adoption-Policy.md
```diff
@@ -0,0 +1,127 @@
...
+seems to be `Delete`.
+
+#### Option 2: The default `reconcile-policy` for adopted resource is `abandon-on-delete`.
+
+ASO detects that a resource previously existed and sets `reconcile-policy` automatically to `abandon-on-delete`.
+A new option in the ASO config `adoption-policy` would control this default so users could opt into the Option 1 style
+discussed above if they wanted (but it's not the default behavior).
+
+**Advantages:**
+- More flexibility. Users can choose which behavior they want when deploying ASO.
```
jonnylangefeld, line 77: Hmm correct me if I'm wrong, but I don't see this as a distinctive advantage to this approach. Users should be just as flexible with the other approach as they can actively choose and set the reconcile-policy. This would just be to attempt a better default, an attempt in guessing what the user actually wants without them actively setting it.

docs/hugo/content/architecture-decision-records/ADR-2022-02-Adoption-Policy.md
```diff
@@ -0,0 +1,127 @@
...
+ASO detects that a resource previously existed and sets `reconcile-policy` automatically to `abandon-on-delete`.
+A new option in the ASO config `adoption-policy` would control this default so users could opt into the Option 1 style
+discussed above if they wanted (but it's not the default behavior).
+
+**Advantages:**
+- More flexibility. Users can choose which behavior they want when deploying ASO. 
+- Defaults to a more conservative `reconcile-policy` to protect them from accidentally deleting things they may not
+  have wanted to delete.
+
+**Disadvantages:**
```
jonnylangefeld, line 81: I think we can add one more to this list which is potential inconsistency across resources.
Imagine you have a helm chart of 10 ASO resources, 5 of which were already present on Azure, and the 5 others are just introduced through the helm chart. Now half is 'owned' by the ASO controller and the other half is not. You forget about that and a long time later you run `helm uninstall` because you intend to get rid of all the resources in the chart. And in fact even all the resources get deleted from the cluster and you won't see them anymore. But half of those resources would still continue to remain in Azure.

docs/hugo/content/architecture-decision-records/ADR-2022-02-Adoption-Policy.md
```diff
@@ -0,0 +1,127 @@
...
+
+**Thoughts/comments:**
+
+There's some discussion of a similar feature over in [ACK](https://github.com/aws-controllers-k8s/community/pull/1148#discussion_r794605821).
+Jay Pipes had an interesting comment:
+> In the case of ACK and the user experience of resource deletion, I posit that the behaviour of our interface has 
+> no surprises and is actually hard to misuse. It doesn't have surprises because our interface to delete a resource 
+> is to call kubectl delete on that resource. This is as non-surprising as it gets. It is hard to misuse because 
+> calling kubectl delete is an explicit action the user has to take and it's not a confusing verb or semantic. 
+> "delete" means to remove the resource.
```
jonnylangefeld, line 100: fully agree with this.

docs/hugo/content/architecture-decision-records/ADR-2022-02-Adoption-Policy.md
```diff
@@ -0,0 +1,127 @@
...
+explicitly sets `reconcile-policy`.
+
+**Advantages:**
+- It's easy to explain what the behavior is. Simplistic users don't have to understand `reconcile-policy` (they won't
+  ever see it unless they set it themselves).
+
+**Disadvantages:**
+- Users have no good way to know their resource existed previously. They may `kubectl delete` it without realizing that
+  other things depend on this resource. Deletion (and some mutation) is destructive in Azure so lost data, etc cannot 
+  easily be recovered.
```
jonnylangefeld, line 48: I think this is a minor issue. Once I decide as a user to use ASO as the controller over my resources, I expect ASO to own the resources and don't care so much which already existed before and which didn't.
If I do care than I'll be explicit about it and set the `reconcile-policy` label, but then at least it's obvious in the source code of the resource (whether helm chart or bare yaml) that I for instance wanted to abandon a resource or skip the reconcile on it.

jonnylangefeld (commented):
If would have a vote on this I'd vote for option 1. That's what I would expect from a controller and I guess the whole point of a "smart guess for deletion policy" is to match what a user would expect.
I think it makes it more homogenous across all resources in a set of resources you apply and also doesn't leave room for surprises. I don't think we have to protect the users from deletion, because it is equivalent to ARM templates or SDKs as you said. Since there's no UI involved there's not going to be a "are you sure you want to delete?" pop-up, but someone intentionally calls the delete API endpoint. In my opinion it's the same whether I call the ARM API, use the SDK or cal a k8s API endpoint to delete a resource.

matthchr:
This is related to #1465

docs/hugo/content/architecture-decision-records/ADR-2022-02-Adoption-Policy.md
matthchr: I covered this I think with this line:
> Bias towards safety may cause resource user thinks was deleted to continue existing (and billing) in Azure. Not a data
  loss though.

docs/hugo/content/architecture-decision-records/ADR-2022-02-Adoption-Policy.md
matthchr: What I was getting at with flexibility here was that, if users wanted the operator to autodetect adopted resources, it could. If they didn't want that, they can just set `adoption-policy` and get the option1 behavior. We'd basically be letting them choose which behavior they want.

Whereas if we just go with option1 and no `adoption-policy` at the operator level then the users can't have the operator "know" for them, they'll have to know and annotate things accordingly.

docs/hugo/content/architecture-decision-records/ADR-2022-02-Adoption-Policy.md
```diff
@@ -0,0 +1,127 @@
...
+is related to adoption. Some of `reconcile-policy`'s primary use cases are for when the user does not want ASO to
+fully manage an adopted resource.
+
+The following are principles we should follow related to resource adoption:
+
+1. **The operator should adopt resources by default:** This is standard goal-seeking/desired-state behavior. 
+   If you create a resource in Kubernetes that says: "I expect there to be a Virtual Machine that looks like this
+   at location X" and that Virtual Machine already exists, then we are in the desired state.
+   The alternative would be to error if the resource already exists, which goes against the Kubernetes model.
+2. **Users must be able to instruct the operator to stop making changes to the backing Azure resource:**
```
theunrepentantgeek, line 20: Suggest a blank line between each principle, since they're paragraphs.

```suggestion
   The alternative would be to error if the resource already exists, which goes against the Kubernetes model.
   
2. **Users must be able to instruct the operator to stop making changes to the backing Azure resource:**
```

docs/hugo/content/architecture-decision-records/ADR-2022-02-Adoption-Policy.md
```diff
@@ -0,0 +1,127 @@
...
+   This is enabled by `reconcile-policy`.
+
+   This is useful for the following scenarios:
+   - The operator is doing something wrong and the user needs an out.
```
theunrepentantgeek, line 24: Not sure of the phrasing here, but feels too informal at the moment.

```suggestion
   - The operator is doing something wrong and the user needs a way to stop it from proceeding
```

docs/hugo/content/architecture-decision-records/ADR-2022-02-Adoption-Policy.md
```diff
@@ -0,0 +1,127 @@
...
+   - Pausing the operator temporarily while they make some manual fix via the Azure portal.
+   - Creating a resource in ASO purely to get ownership references to work. The resource in question is not managed
+     by ASO at all. It was created and managed some other way (portal, terraform, az-cli, etc).
+3. **Users must be able to instruct the operator to skip deletion of the backing resource:**
+   This is useful for scenarios such as moving ownership of resources from ASO in cluster A to ASO in cluster B.
```
theunrepentantgeek, line 29:
```suggestion
   This is useful for scenarios such as moving ownership of resources from ASO in cluster A to ASO in cluster B,
   or as a safety net to protect stateful resources such as databases from accidental deletion.
```

docs/hugo/content/architecture-decision-records/ADR-2022-02-Adoption-Policy.md
theunrepentantgeek: I agree this is the case for users who understand the service operator model.

I have a concern about users who aren't experienced/knowledgable.

Consider what happens if a potential user starts auditioning ASO to see if it meets their needs and then inadvertently deletes something in Azure that they really shouldn't have? The most likely case would be a database or similar resource in their Dev/Test subscription. Worst case, they delete something from prod.

Is it really useful at that point for us to say "You should have understood what you were doing before you did it?"

While this is a minor issue overall, it wouldn't feel minor to that particular user.

I'm interested to see if there's a safe default that we can reasonably adopt, so that data-loss is less likely.

docs/hugo/content/architecture-decision-records/ADR-2022-02-Adoption-Policy.md
theunrepentantgeek: I fully agree that there's no ambiguity that `kubectl delete` means remove the resource _from the cluster_.

That said, I do think that different users will have different expectations about whether deletion from the cluster necessarily requires deletion from Azure as a natural consequence.

docs/hugo/content/architecture-decision-records/ADR-2022-02-Adoption-Policy.md
CecileRobertMichon: Another important use case to consider is "update". For example, what happens if I create a LB with the name of a LB that already exists (without realizing it) but specify properties (eg. LB rules) that are different from the existing LB? Will ASO silently update my LB, potentially breaking other applications that rely on that resource? I can see how a "kubectl delete" is an explicit action but creating a resource with a name collision might be more surprising. 

A way to look at it another way is that performing a PUT in Azure on an existing resource does result in the resource being updated in the same pattern.

docs/hugo/content/architecture-decision-records/ADR-2022-02-Adoption-Policy.md
```diff
@@ -0,0 +1,127 @@
...
+`kubectl delete` always deletes the backing resource no matter what.
+
+[GCP](https://cloud.google.com/config-connector/docs/how-to/managing-deleting-resources#creating_a_resource) does it this way:
+> When you create a resource, Config Connector creates the resource if it doesn't exist. If a Google Cloud resource 
+> already exists with the same name, then Config Connector acquires the resource and manages it.
+
+GCP also [supports](https://cloud.google.com/config-connector/docs/how-to/managing-deleting-resources) a 
+`reconcile-policy`-like annotation, so doing it like this would basically be doing it exactly like GCP.
+
+Kuberentes has a similar concept for `StorageClasses`/`PersistentVolumes` called
```
CecileRobertMichon, line 66:
```suggestion
Kubernetes has a similar concept for `StorageClasses`/`PersistentVolumes` called 
```

docs/hugo/content/architecture-decision-records/ADR-2022-02-Adoption-Policy.md
matthchr: > For example, what happens if I create a LB with the name of a LB that already exists (without realizing it) but specify properties (eg. LB rules) that are different from the existing LB?

Agree that this is definitely an important case to consider. There's nothing currently in this proposal that would automatically block mutation of an existing resource. In all cases that I could find, the mutations are reversible. I was thinking that something like shrinking PostgreSQL DB size would be non-reversible, but they disallow that operation (possibly because it's destructive). In your example, assuming you know what your old LB rules are, you can always put them back later, so at least there's an "out". The focus was on the delete case because there is a possibility that there's no recovery if you delete your database.

Option 2 outlined here would help with the delete case but does nothing for the update case. I don't see a way to satisfy this goal:
> **The operator should adopt resources by default:** This is standard goal-seeking/desired-state behavior. 
   If you create a resource in Kubernetes that says: "I expect there to be a Virtual Machine that looks like this
   at location X" and that Virtual Machine already exists, then we are in the desired state.
   The alternative would be to error if the resource already exists, which goes against the Kubernetes model.

and _also_ make adoption truly "safe". As you say, it's like submitting an ARM template for an existing resource. If you do it, it will modify the resource.

docs/hugo/content/architecture-decision-records/ADR-2022-02-Adoption-Policy.md
matthchr: Added

docs/hugo/content/architecture-decision-records/ADR-2022-02-Adoption-Policy.md
```diff
@@ -0,0 +1,127 @@
...
+
+[GCP](https://cloud.google.com/config-connector/docs/how-to/managing-deleting-resources#creating_a_resource) does it this way:
+> When you create a resource, Config Connector creates the resource if it doesn't exist. If a Google Cloud resource 
+> already exists with the same name, then Config Connector acquires the resource and manages it.
+
+GCP also [supports](https://cloud.google.com/config-connector/docs/how-to/managing-deleting-resources) a 
+`reconcile-policy`-like annotation, so doing it like this would basically be doing it exactly like GCP.
+
+Kuberentes has a similar concept for `StorageClasses`/`PersistentVolumes` called 
+[reclaimPolicy](https://kubernetes.io/docs/concepts/storage/storage-classes/#reclaim-policy), whose default
```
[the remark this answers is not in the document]
matthchr: Clarified this in the docs

docs/hugo/content/architecture-decision-records/ADR-2022-02-Adoption-Policy.md
matthchr: I have amended this design to talk about two different questions here:
1. What is the default behavior?
2. Can users override that behavior?

My proposed choice is that the default behavior is to do what we're doing today (adopt resources if they already existed and manage them), _but_ unlike the original design (from nearly a year ago!), I also suggest we allow per-resource and operator-level overrides to this behavior so that if there are particular resources where users want a different behavior they can select that behavior.

Resolving this conversation as most of the points brought up in it have been incorporated into the design.

docs/hugo/content/design/ADR-2023-02-Adoption-Policy.md
```diff
@@ -0,0 +1,258 @@
...
+when reconciliation ran for the first time. The operator would check if the resource already exists and write the corresponding 
+`reconcile-policy` annotation to the resource if the resource existed.
+
+**Advantages:**
+- Uses a concept we already have.
+
+**Disadvantages:**
+- Possibility of collision between user specified `reconcile-policy` and the `reconcile-policy` written by the operator because
+  of `reconcile-policy-if-exists`. 
+  - We can mitigate this issue by rejecting some configurations with a webhook, although we have to be careful how we craft
```
nojnhuh, line 206: Would something like `reconcile-policy: skip`, `reconcile-policy-if-not-exists: manage` be possible to express using this method? What would happen in that case for a resource that does not already exist? 

(Asking generally, I don't think we'd do something like that in CAPZ)

docs/hugo/content/design/ADR-2023-02-Adoption-Policy.md
```diff
@@ -0,0 +1,258 @@
...
+    that webhook because the operator itself needs to be able to set both. We can either just block both on create but allow
+    both on update, or we can block both on create/update but only as long as the finalizer hasn't been set. Once the finalizer 
+    has been set, changing `reconcile-policy-if-exists` has no effect (we've already applied whatever we needed to) and all changes
+    to the policy should flow via the `reconcile-policy` annotation.
+- Excess events generated by us setting `reconcile-policy`, which is an annotation we watch for updates. We can deal with this: 
+  - If the `reconcile-policy-if-exists` is `skip` this is safe as we never re-reconcile when `reconcile-policy` is being set to skip.
```
nojnhuh, line 212: > we never re-reconcile when `reconcile-policy` is being set to skip.

Clarified by @matthchr on Slack, but this wasn't exactly clear to me that this referred to only the reconciliation where `skip` is set. Future periodic reconciliations still happen for `skip`-ped resources.

docs/hugo/content/design/ADR-2023-02-Adoption-Policy.md
```diff
@@ -0,0 +1,258 @@
...
+
+## Decision
+
+I propose we choose [option 1: Default `reconcile-policy` is `manage`](#option-1--the-default-reconcile-policy-for-adopted-resources-is-manage).
+As mentioned in the discussion of that topic, it's the default we currently have and so is non-breaking to adopt, and is how
+ARM templates/BICEP, raw REST and the SDKs work today already. Extra protection against adoption/updating a resource that
+already exists can be achieved with some of those technologies, but it's not the default and requires the user to opt-in
+to it.
+
+Alongside that, I propose we choose [option 4.2: `reconcile-policy-if-exists` annotation](#option-42--reconcile-policy-if-exists-annotation)
```
nojnhuh, line 241: This would be great to have for CAPZ and is basically exactly what I had in mind. Agreed that this approach seems better than `adoption-policy`.

docs/hugo/content/design/ADR-2023-02-Adoption-Policy.md
```diff
@@ -0,0 +1,258 @@
...
+on the behavior they wanted.
+
+#### Option 3: No configuration at the operator or per-resource level
+
+**Advantages:**
+- Simple - value of this cannot be understated. The more configuration users _have_ to understand the harder the operator
+  becomes to adopt or understand.
+
+**Disadvantages:**
+- Fails to cater to some legitimate use-cases.
```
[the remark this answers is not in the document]
matthchr: Yeah this is pretty awkward. I think it makes sense to support something like this natively in the operator.

docs/hugo/content/design/ADR-2023-02-Adoption-Policy.md
```diff
@@ -0,0 +1,258 @@
...
+#### Option 4: Allow configuration at the operator or per-resource level
+
+This option would just be configuring the behavior of the operator in case it found an existing resource
+in Azure. The behavior if ASO created the resource is unchanged.
+
+The 3 scenarios here are, if the resource already exists in Azure when ASO goes to create it:
+1. Fully adopt the resource, mutate it if its configuration is different in ASO than in Azure, and delete 
+   it when `kubectl delete` is run.
+2. Partially adopt the resource. Mutations of the resource would be performed by ASO but `kubectl delete` will only
+   delete the resource in Kuberentes and _not_ in Azure.
```
theunrepentantgeek, line 137:
```suggestion
   delete the resource in Kubernetes and _not_ in Azure.
```

theunrepentantgeek (approved):
I agree with your conclusions.

docs/hugo/content/design/ADR-2023-02-Adoption-Policy.md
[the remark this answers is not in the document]
nojnhuh: Once scenario I forgot to mention here is that for the first CAPZ upgrade that introduces ASO for a particular resource type, we _will_ want to adopt resources that already exist as long as they have a CAPZ-specific tag which communicates CAPZ's ownership of the resource pre-ASO. So I think we would still need a similar flow to what I outlined above:

1. Create a resource with `reconcile-policy-if-exists: skip`
2. Requeue the corresponding CAPZ resource
3. Check if the `reconcile-policy` was set to `skip` and the CAPZ "owned" tag is set
4. If so, update `reconcile-policy` to `manage`

I imagine this scenario is probably unique enough that ASO would never have a native way to handle that entirely declaratively.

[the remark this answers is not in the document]
matthchr: Yeah, that seems like something that CAPZ can manage either as you describe or using the SDK (before moving off of the SDK entirely).

docs/hugo/content/design/ADR-2023-02-Adoption-Policy.md
matthchr: Have clarified this in the design by rephrasing to:

> If the `reconcile-policy-if-exists` is `skip` this is safe as we don't have to trigger reconciliation if the only modification to the resource is that the `reconcile-policy` is being set to `skip`. This helps avoid a situation where we reconcile once, issue a GET to Azure, see that the resource exists and update the `reconcile-policy` annotation to be `skip`, which then triggers another reconciliation and another GET to Azure.

docs/hugo/content/design/ADR-2023-02-Adoption-Policy.md
matthchr: Great question. I've expanded the disadvantages section to talk about two types of possible mitigation here (up from 1).

The second option is to allow both and have `reconcile-policy-if-exists` always overwrite `reconcile-policy`. The only downside I see to this is that it could cause some gitops flows issue as we might overwrite the annotation they provided with the annotation in `reconcile-policy-if-exists`, causing their checked-in spec to differ from the active spec (annotation mismatch). It's a minor issue but at the same time I am not clear on what exact scenarios users would really want to set both in practice, for example your example scenario feels pretty contrived. One that seems possibly valid is: `reconcile-policy: detach-on-delete` and  `reconcile-policy-if-exists: skip`...

I've updated the decision section to lean in the direction of the first option (webhook prevention) as we can always expand it later but we can't undo letting users set both if we do it now, but I'm open to other opinions on that.

super-harsh (approved):

PR #2071, opened 2022-02-07
Simplify configuration of which resources to export

theunrepentantgeek:

Moves configuration for which resources are exported into the new object model section of the configuration file, which is more compact and easier to understand.

We previously had a model where all resources would be exported by default, and we've moved to a new model where we specify the resources of interest; this change updates the configuration file to match that policy.

gif

**If applicable**:
- [x] this PR contains tests


v2/azure-arm.yaml
```diff
@@ -984,18 +1032,32 @@ objectModelConfiguration:
...
+      DatabaseAccountsMongodbDatabase:
+        $exportAs: MongodbDatabase
+      DatabaseAccountsMongodbDatabasesCollection:
+        $exportAs: MongodbDatabaseCollection
+      DatabaseAccountsMongodbDatabasesCollectionsThroughputSetting:
+        $exportAs: MongodbDatabaseCollectionThroughputSetting
+      DatabaseAccountsMongodbDatabasesThroughputSetting:
+        $exportAs: MongodbDatabaseThroughputSetting
+      DatabaseAccountsSqlDatabase:
+        $exportAs: SqlDatabase
```
babbageclunk, line 1048: This is really nice.

v2/tools/generator/internal/config/configuration.go
```diff
@@ -344,91 +305,6 @@ func absDirectoryPathToURL(path string) *url.URL {
 	return result
 }
 
-type ExportFilterFunc func(astmodel.TypeName) (result ShouldExportResult, newName *astmodel.TypeName, because string)
```
babbageclunk, line 347: Love to see this.

v2/tools/generator/internal/config/type_configuration.go
```diff
@@ -27,9 +27,10 @@ import (
 //
 type TypeConfiguration struct {
 	name              string
-	renamedTo         *string
-	renamedToConsumed bool
 	properties        map[string]*PropertyConfiguration
+	nameInNextVersion configurableString
```
babbageclunk, line 31: The configurableX types are nice - I guess they'll be even nicer once we can use generics!

babbageclunk (approved):
I think I must be too hot, didn't find anything to complain about! Looks great!

v2/azure-arm.yaml
```diff
@@ -869,64 +869,87 @@ status:
 objectModelConfiguration: 
   authorization:
     2020-08-01-preview:
+      RoleAssignment:
```
matthchr, line 872: If we're doing this, shouldn't the `include-transitive` section of this document be deleted? They aren't both needed, right?

v2/tools/generator/internal/codegen/pipeline/apply_export_filters.go
```diff
@@ -32,47 +31,69 @@ func filterTypes(
...
-		case config.Export:
-			if reason == "" {
-				klog.V(3).Infof("Exporting %s", defName)
-			} else {
-				klog.V(2).Infof("Exporting %s because %s", defName, reason)
+		export, err := configuration.ObjectModelConfiguration.LookupExport(defName)
+		if err != nil {
+			if config.IsNotConfiguredError(err) {
+				export = false
 			}
```
matthchr, line 44: There is no path in this if statement that deals with an `err` that is _not_ a `config.NotConfiguredError`. That seems like an oversight, as an error would be silently ignored then?

v2/tools/generator/internal/codegen/pipeline/apply_export_filters.go
```diff
@@ -32,47 +31,69 @@ func filterTypes(
 	configuration *config.Configuration,
 	state *State) (*State, error) {
 
-	newDefinitions := make(astmodel.Types)
-
-	filterer := configuration.BuildExportFilterer(state.types)
```
matthchr, line 37: I don't have any issue with us deleting this, BUT this `filterTypes` is doing something quite different than it was before.

Before: It applied filters to all types, and could manually exclude things _or_ include things.
Now: It applies filters only to resource types and excludes everything by default unless it was manually included. There's no way to include everything in a package. I'm ok with this (we can expand it back out to support that in the future if we want).

I think it's clearer now, but we have a bunch of rules that weren't `include-transitive`, like:
```
  - action: exclude
    group: definitions
    name: '*ResourceBase*'
    because: Not actually a resource, used for inheritance only.
  - action: exclude
    group: deploymenttemplate
    name: '*ResourceBase*'
    because: Not actually a resource, used for inheritance only.
```

these are still in the YAML but as far as I can tell they aren't doing anything?

v2/tools/generator/internal/codegen/pipeline/apply_export_filters.go
```diff
@@ -32,47 +31,69 @@ func filterTypes(
...
-			} else {
-				klog.V(2).Infof("Exporting %s because %s", defName, reason)
+		export, err := configuration.ObjectModelConfiguration.LookupExport(defName)
+		if err != nil {
+			if config.IsNotConfiguredError(err) {
+				export = false
 			}
+		}
+
+		if _, err = configuration.ObjectModelConfiguration.LookupExportAs(defName); err == nil {
```
matthchr, line 47: minor: Wondering if it would be slightly clearer if this second bit was also done by `LookupExport` (possibly with a changed name), so that the contract w/ the configuration was just `config.IsThisThingSupposedToBeExported()`, as opposed to having to stitch together a couple properties that actually each mean that in their own way

v2/tools/generator/internal/codegen/pipeline/apply_export_filters.go
```diff
@@ -32,47 +31,69 @@ func filterTypes(
...
+	renamingVisitor := astmodel.NewRenamingVisitor(renames)
+	result, err := renamingVisitor.RenameAll(defs)
 	if err != nil {
 		return nil, err
 	}
 
 	return state.WithTypes(result), nil
 }
+
+func findReferencedTypes(root astmodel.TypeName, types astmodel.Types) astmodel.TypeNameSet {
```
matthchr, line 86: You can already do this in two ways:
1. `astmodel.FindConnectedTypes` returns an `astmodel.Types` collection
2. `reference_graph.go` `graph.Connected()` returns a collection of `astmodel.TypeName`'s

IMO should use one of the existing methods

v2/tools/generator/internal/config/configurable_bool.go
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package config
+
+type configurableBool struct {
```
matthchr, line 8: minor: comment onto this as well?

v2/tools/generator/internal/config/configurable_bool.go
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package config
```
matthchr, line 6: I like this type!

v2/tools/generator/internal/config/configurable_bool_test.go
```diff
@@ -0,0 +1,61 @@
...
+	var s configurableBool
+	s.write(expected)
+
+	actual, ok := s.read()
+	g.Expect(actual).To(Equal(expected))
+	g.Expect(ok).To(BeTrue())
+	g.Expect(s.isUnconsumed()).To(BeFalse())
+}
+
+func TestConfigurableBool_WhenValueWritten_MarkedAsUnConsumed(t *testing.T) {
```
matthchr, line 53: minor: `Unconsumed` not `UnConsumed` here

v2/tools/generator/internal/config/configuration.go
```diff
@@ -46,8 +46,6 @@ type Configuration struct {
 	// AnyTypePackages lists packages which we expect to generate
 	// interface{} fields.
 	AnyTypePackages []string `yaml:"anyTypePackages"`
-	// Filters used to control which types are exported
-	ExportFilters []*ExportFilter `yaml:"exportFilters"`
```
matthchr, line 50: Oh, you've removed this and the YAML has this section still (I think you forgot to remove it), but there's no error when that happens?

Can we pass a "error on unknown fields" to the YAML deserializer?

matthchr: Happy to deal with this in a separate PR though

v2/tools/generator/internal/config/object_model_configuration.go
```diff
@@ -222,3 +283,34 @@ func (omc *ObjectModelConfiguration) configuredGroups() []string {
 
 	return result
 }
+
+// CreateTestObjectModelConfiguration builds up a new configuration for a particular type, returning both the top level
+// configuration and the type configuration.
+// While intended only for test use, this isn't in a _test.go file as we want to use it from tests in multiple packages.
```
matthchr, line 289: IMO it's awkward having code only for test alongside "production" code. Would putting this into the `testcommon` package work? That package already exists, is used in multiple test packages, and isn't next to the production code.

There's quite a bit of shared stuff there such as the test gomega matchers, test resoruce namer, and test context stuff which feels like it goes well with this?

v2/tools/generator/internal/codegen/pipeline/apply_export_filters.go
theunrepentantgeek: Yes, it's an oversight. At the moment the only possible error is a `NotConfiguredError`, but that might change in future. Thanks.

v2/tools/generator/internal/codegen/pipeline/apply_export_filters.go
theunrepentantgeek: I thought about that, but decided the `config` package should be about configuration - not so much about the business rules. I'll move the logic into new function, `shouldExport()` which should make things clearer.

v2/tools/generator/internal/codegen/pipeline/apply_export_filters.go
theunrepentantgeek: This method was just relocated from the (now deleted ExportFilters); happy to simplify.

v2/tools/generator/internal/config/configurable_bool.go
theunrepentantgeek: Done.

v2/tools/generator/internal/config/configurable_bool_test.go
theunrepentantgeek: Fixed in both test files.

v2/tools/generator/internal/config/configuration.go
theunrepentantgeek: Created issue #2081 to track.

v2/tools/generator/internal/config/object_model_configuration.go
theunrepentantgeek: I agree with your concerns. That said, I put this in the same package so it could have access to private fields, not sure about moving it. Will consider, but not in this PR.

v2/tools/generator/internal/codegen/pipeline/apply_export_filters.go
matthchr: Hah! I missed that it was just relocated. Yeah give one of those other methods a try. I think they should basically just work for you out of the box.

v2/azure-arm.yaml
```diff
@@ -99,359 +99,6 @@ typeFilters:
...
-    group: datafactory
-    version: v*api20180601
-    because: ExecuteSSISPackageActivityTypeProperties.PackageConnectionManagers is a map[string]map[string]SSISExecutionParameter, which controller-gen doesn't know how to handle
-
-  # Include just the types we want to generate.
-  # 'include-transitive' includes their dependencies recursively and makes sure everything that is needed is included:
-  - action: include-transitive
-    group: storage
-    version: v*api20210401
-    name: 'StorageAccount'
```
matthchr, line 128: Now that's what I like to see.

v2/tools/generator/internal/codegen/pipeline/apply_export_filters.go
```diff
@@ -32,47 +32,79 @@ func filterTypes(
...
+	result, err := renamingVisitor.RenameAll(typesToExport)
 	if err != nil {
 		return nil, err
 	}
 
 	return state.WithTypes(result), nil
 }
+
+// shouldExport works out whether the specified Resource should be exported or not
+func shouldExport(defName astmodel.TypeName, configuration *config.Configuration) (bool, error) {
```
matthchr, line 85: Yeah this helps with clarity quite a bit.

matthchr (approved):

PR #2074, opened 2022-02-08
Add ContainerRegistry resource

vaspahomov:

Add Container Registry Resource

This is my first PR in ASO. For now IDK the right way to add new resource, and I've tried to follow [Add new generated resource guide](https://azure.github.io/azure-service-operator/contributing/add-a-new-code-generated-resource/).

I've: 
1. Add ContainerRegistry in `exportFilters` 
2. Call `./tool/generator`
3. Fix generator errors with patches in `objectModelConfiguration`

gif

**If applicable**:
- [ ] this PR contains tests


v2/azure-arm.yaml
```diff
@@ -449,6 +449,11 @@ exportFilters:
     version: v*api20210301
     name: RedisEnterpriseDatabase
     because: "including redis enterprise resources"
+  - action: include-transitive
+    group: containerregistry
+    version: v*api20190501
```
matthchr, line 454: Can you try using the new 2021-09-01 API instead? It's significantly newer.

To do this we will have to update our submodule so that we can see it. I've just opened a PR to do that: #2075

Once that PR merges you can merge with it (or rebase onto it) and should be able to use `20210901` here (and below) instead.

matthchr:
@vaspahomov thanks for this! You've done exactly the right thing.

Can you also follow the [instructions](https://azure.github.io/azure-service-operator/contributing/add-a-new-code-generated-resource/#write-a-crud-test-for-the-resource) for writing a CRUD test for the new resource, to make sure it works? Once you've written the test you can run it by following [these instructions](https://azure.github.io/azure-service-operator/contributing/contributing/#running-integration-tests).

The tests have a "recording" and a "replay" mode. Running in record mode (which you will have to do initially) requires an Azure subscription and actually creates the resource in question. It should delete it when the test is done though and save the HTTP request/response payloads (secrets and actual subscription IDs are redacted automatically). You can the include the test and the recording file as part of this PR.

Lastly, you will also need a sample in order to pass CI. Usually the easiest thing to do is after you've written a working simple test, use that same set of parameters and craft a sample as well. There are instructions on that [here](https://azure.github.io/azure-service-operator/contributing/add-a-new-code-generated-resource/#add-a-new-sample)

v2/azure-arm.yaml
matthchr: That PR merged so you should be able to try the newer API version

vaspahomov: I've just update api on newer 2021-09-01 and now I have this error:

```
E0209 10:26:50.957979   23472 gen_kustomize.go:111] Error during code generation:
failed during pipeline stage 23/56: Verify there are no ErroredType's containing errors: "github.com/Azure/azure-service-operator/v2/api/containerregistry/v1alpha1api20210901/Registry" has property "Status" with errors: "missing status information for github.com/Azure/azure-service-operator/v2/api/containerregistry/v1alpha1api20210901/Registries"
```

IDK what I'am doing wrong. Can you help me and explain the reason of this error?
May be this related to that Registry ARM resource [does not have status property](https://github.com/Azure/azure-resource-manager-schemas/blob/60b772a297ef5dff7feb57761dd26557bd600bf3/schemas/2021-09-01/Microsoft.ContainerRegistry.json#L9)?

v2/azure-arm.yaml
```diff
@@ -449,6 +449,11 @@ exportFilters:
...
+    name: Registry
+    because: "including container registry resource"
```
theunrepentantgeek, line 456: When I merge PR #2071, you'll end up with a conflict here in the configuration file - we're removing the export filters section in favour of a simpler technique.

You'll instead need to add this resource to the `ObjectModelConfiguration` section; it'll look something like this:

``` yaml
ContainerRegistry:
  2021-09-01:
    Registry:
      $export: true
```

See PR #2080 for the matching updates to _Adding a new code generated resource to ASO v2_.

v2/azure-arm.yaml
matthchr: Ah, this is my fault! I updated one of our submodules but not the other. Here's a PR that fixes it: #2079

I've also confirmed locally that with that PR merged your changes here should work.
You will need to replace your section on line 951 with this, I believe:
```
  containerregistry:
    2021-09-01:
      IdentityProperties:
        PrincipalId:
          $armReference: false
```

Mark that as `$armReference: false` because that field isn't an ARM reference (ARM ID) it's the principal ID of the identity in AAD.

vaspahomov: Thanks!

v2/azure-arm.yaml
vaspahomov: Thanks!

matthchr (approved):
LGTM!

theunrepentantgeek (approved):

Porges:
Nice job @vaspahomov :+1:

matthchr:
I think you need to re-run the generator tool because I merged #2082 and so the format of `controller_resourges_gen.go` changed a bit. That's causing the CI to fail. Once you do that we'll run the rest of the tests and merge.

matthchr:
Thanks for your contribution @vaspahomov! This will be included in the next ASO release, which will be in the next ~1 week (give or take a bit).

PR #2079, opened 2022-02-09
Update Swagger submodule

matthchr:
This enables ongoing work for 2021-09-01 ContainerRegistry as well as
2021-07-01 Compute/images.


theunrepentantgeek (approved):
gif

Issue #1944, opened 2021-11-01
Bug: updating a virtual network will (try to) remove any subnets it has


babbageclunk:
**Version of Azure Service Operator**
v2.0.0-alpha.2

**Describe the bug**
Subnet is both a standalone resource *and* and embedded in VirtualNetwork. If with ASO v2 you create a vnet, then add a subnet, then update some field on the vnet, the subnet will be deleted in Azure and the resource in k8s will be orphaned.  Or if you create the vnet and the subnet at the same time in k8s, the LRO to create the vnet will be cancelled and ASO will reconcile the vnet again. The subnet won't ever be successfully created.

This is happening because the semantics for PUT on vnets is different from that of most resources in Azure. https://github.com/Azure/azure-quickstart-templates/issues/2786
ASO doesn't send the subnets embedded in the vnet when it sends the vnet update, so ARM deletes the subnets.

This also affects other resources with this same relationship in Network (for example [RouteTable/Route](https://github.com/Azure/azure-service-operator/pull/1925)).

**To Reproduce**
* Create a vnet and a subnet in the vnet at the same time. The vnet will be created in Azure with no subnets, and the subnet will be stuck reconciling and not be created.

**Expected behavior**
The subnet should be created successfully.

PR #2084, opened 2022-02-10
Swagger only

### LINKED ISSUE #1460: Outstanding items with resource status information
- [x] Get rid of all `unable to produce type` errors and [make it fatal](https://github.com/Azure/k8s-infra/pull/205#discussion_r471845144). Fixed in Azure/k8s-infra#229.
- [x] Need to handle enum parameters in URLs. e.g. `Dnszones/{type}` expands to `DnszonesPTR`, `DnszonesSOA`, … Fixed by Azure/k8s-infra#230.
- [x] Investigate instances of “resource already defined differently” (13). Fixed by Azure/k8s-infra#231.
- [x] Add `controller-gen` status annotations.
- [x] Validate some example status types against real ARM (with test/demo controller).
- [x] #1838
- [x] Investigate missing status information (currently 289/1885 types). Some explained by items above. Azure/k8s-infra#231 reduces this to 199.
  - Note that this is addressed for all currently-used resource types; we currently error out if any types would be generated without status information.
- [ ] Get rid of all unable-to-extract-resource-name errors and [make it fatal](https://github.com/Azure/k8s-infra/pull/205#discussion_r471845030). (Some ideas here: we are matching things as resources that aren't really resources, we can be more strict on the checks by investigating the types to see if they are tagged as ARM resources or not.)

### LINKED ISSUE #1554: Generated packages should define an accessible ARMAPIVersion variable
Today, we have per-resource ARM API versions, for example:
```
// +kubebuilder:validation:Enum={"2020-11-01"}
type LoadBalancersSpecApiVersion string

const LoadBalancersSpecApiVersion20201101 = LoadBalancersSpecApiVersion("2020-11-01")
```

This is used like so:
```
// ConvertToARM converts from a Kubernetes CRD object to an ARM object
func (loadBalancersSpec *LoadBalancers_Spec) ConvertToARM(name string, resolvedReferences genruntime.ResolvedReferences) (interface{}, error) {
        ...
	result.ApiVersion = LoadBalancersSpecApiVersion20201101
        ...
```

**Describe the improvement**
Since our packages are per-API version anyway, I think we can do away with the per-resource APIVersion and instead just use a package global one, maybe called `ARMAPIVersion`.

Then the tests could also use this via:
```
network "github.com/Azure/azure-service-operator/hack/generated/_apis/microsoft.network/v1alpha1api20201101"
...

network.ARMAPIVersion
```

### LINKED ISSUE #1730: Not all resources implement ResourceBase
The resource manager schema implies that all resources implement `resourceBase`, which has `tags` and a few other properties. This is not correct though, there are multiple "kinds" of ARM resources and some of the (Proxy) don't have to support `tags`. See [this](https://armwiki.azurewebsites.net/rp_onboarding/tracked_vs_proxy_resources.html) for details.

**Describe the improvement**
The Swagger specifications seem to correctly encode this information, see for example the [ASK AgentPool](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/containerservice/resource-manager/Microsoft.ContainerService/stable/2021-07-01/managedClusters.json#L2593) resource which doesn't support `tags` according to Swagger, but which incorrectly is marked as supporting `tags` according to the resource manager schema JSON document.

We should stop forcing all types to inherit these `ResourceBase` properties, as not all of them do.

Porges:
DRAFT for diffing purposes only

---

Bonus change: when removing validations from Status types, replace validated types with the underlying type. This means we use plain strings instead of unvalidated enums for Status values.

Fixes #1730, fixes #1554, fixes #1460


v2/api/batch/versions_matrix.md
```diff
@@ -1,35 +1,23 @@
...
+| BatchAccount_Status                               | v1alpha1api20210101 |
+| EncryptionProperties                              | v1alpha1api20210101 |
+| EncryptionPropertiesKeySource                     | v1alpha1api20210101 |
+| EncryptionProperties_Status                       | v1alpha1api20210101 |
+| KeyVaultProperties                                | v1alpha1api20210101 |
+| KeyVaultProperties_Status                         | v1alpha1api20210101 |
+| KeyVaultReference                                 | v1alpha1api20210101 |
+| KeyVaultReference_Status                          | v1alpha1api20210101 |
+| PoolAllocationMode                                | v1alpha1api20210101 |
+| PublicNetworkAccessType                           | v1alpha1api20210101 |
```
theunrepentantgeek, line 23: How come the list of generated types is so much shorter? 33 vs 21 is a large decrease.

v2/tools/generator/internal/testcases/json_serialization_test_case.go
```diff
@@ -646,7 +646,36 @@ func (o *JSONSerializationTestCase) createIndependentGenerator(
...
+								Names: []*dst.Ident{dst.NewIdent("result")},
+								Type:  dst.NewIdent(vt.ElementType().String()),
+							},
+						}},
+						Results: &dst.FieldList{List: []*dst.Field{
+							{Type: dst.NewIdent(def.Name().Name())},
+						}},
+					},
+					Body: astbuilder.StatementBlock(astbuilder.Returns(astbuilder.CallFunc(def.Name().Name(), dst.NewIdent("result")))),
+				})
```
theunrepentantgeek: Pretty7 sure we have helpers in `astbuilder` to simplify this.

v2/tools/generator/internal/jsonast/swagger_type_extractor_test.go
```diff
@@ -50,9 +49,10 @@ func Test_InferNameFromURLPath_SkipsDefault(t *testing.T) {
 	group, name, err := inferNameFromURLPath("Microsoft.Storage/storageAccounts/{accountName}/blobServices/default/containers/{containerName}")
 	g.Expect(err).To(BeNil())
 	g.Expect(group).To(Equal("Microsoft.Storage"))
-	g.Expect(name).To(Equal("StorageAccountsBlobServicesContainers"))
+	g.Expect(name).To(Equal("StorageAccountsBlobServicesContainer")) // should be singularized
 }
 
+/*
```
theunrepentantgeek: Should this test still be disabled?

v2/tools/generator/internal/jsonast/swagger_type_extractor.go
```diff
@@ -49,7 +48,17 @@ func NewSwaggerTypeExtractor(
 }
 
 type SwaggerTypes struct {
-	ResourceTypes, OtherTypes astmodel.Types
+	ResourceTypes ResourceTypes
+	OtherTypes    astmodel.Types
+}
```
theunrepentantgeek: 😃

v2/tools/generator/internal/jsonast/swagger_type_extractor.go
```diff
@@ -345,13 +490,19 @@ func inferNameFromURLPath(operationPath string) (string, string, error) {
...
 
+	// Uppercase first letter in each part:
+	for ix := range nameParts {
+		nameParts[ix] = strings.ToUpper(nameParts[ix][0:1]) + nameParts[ix][1:]
+	}
+
+	// Singularize last part of name:
+	nameParts[len(nameParts)-1] = astmodel.Singularize(nameParts[len(nameParts)-1])
+
+	name := strings.Join(nameParts, "")
```
theunrepentantgeek: I think there's already this functionality in IdentifierFactory - maybe reuse?

v2/api/network/v1alpha1api20201101/virtual_network_types_gen.go
```diff
@@ -335,112 +340,165 @@ type VirtualNetwork_Status struct {
...
-
-	//ProvisioningState: The provisioning state of the virtual network resource.
-	ProvisioningState *ProvisioningState_Status `json:"provisioningState,omitempty"`
+	// +kubebuilder:validation:Required
+	Owner genruntime.KnownResourceReference `group:"resources.azure.com" json:"owner" kind:"ResourceGroup"`
 
-	//ResourceGuid: The resourceGuid property of the Virtual Network resource.
-	ResourceGuid *string `json:"resourceGuid,omitempty"`
+	//Reference: Resource ID.
+	Reference *genruntime.ResourceReference `armReference:"Id" json:"reference,omitempty"`
```
matthchr, line 356: Not 100% sure this is right. This looks like the ID field (which should be GET only, I think?) is being turned into a field called `Reference`. I think this just shouldn't be on the spec at all?

v2/api/network/v1alpha1api20201101/virtual_network_types_gen.go
```diff
@@ -335,112 +340,165 @@ type VirtualNetwork_Status struct {
...
 
 	//Subnets: A list of subnets in a Virtual Network.
-	Subnets []Subnet_Status_VirtualNetwork_SubResourceEmbedded `json:"subnets,omitempty"`
+	Subnets []genruntime.ResourceReference `json:"subnets,omitempty"`
```
matthchr, line 359: I _think_ we actually just want to remove this entirely. Might be worth syncing up with me on this though. My current thinking is that these shouldn't exist on the k8s type but should exist on the ARM type, and then we'll write an extension that populates them on the ARM type just before sending to Azure, to deal with #1944.

Although, doing this might solve the problem (by forcing the users to manage their vnets + subnets as one). I am actually not sure what ARM does with a payload that looks like this though, as presumably this ends up being just a collection of ARM IDs rather than a full subnet specification. It's possible that they then clear out other fields on the subnet because they were "unset" here?

v2/api/network/v1alpha1api20201101/virtual_network_types_gen.go
Porges: yes this stuff is still messed up, it's the final keystone piece :D

v2/tools/generator/internal/astmodel/package_import.go
```diff
@@ -24,6 +24,10 @@ var _ fmt.Stringer = &PackageImport{}
 
 // NewPackageImport creates a new package import from a reference
 func NewPackageImport(packageReference PackageReference) PackageImport {
+	if packageReference == nil {
+		panic("nil")
```
theunrepentantgeek, line 28: Suggest improving this message

v2/tools/generator/internal/astmodel/package_reference_set.go
```diff
@@ -27,6 +27,10 @@ func NewPackageReferenceSet(refs ...PackageReference) *PackageReferenceSet {
 
 // AddReference ensures the set includes an specified Reference
 func (set *PackageReferenceSet) AddReference(ref PackageReference) {
+	if ref == nil {
+		panic("nil reference passed")
```
theunrepentantgeek, line 31: Suggest improving this message

v2/tools/generator/internal/codegen/code_generator.go
```diff
@@ -82,50 +82,56 @@ func NewCodeGeneratorFromConfig(configuration *config.Configuration, idFactory a
...
-		pipeline.LoadSchemaIntoTypes(idFactory, configuration, pipeline.DefaultSchemaLoader),
-
-		// Import status info from Swagger:
-		pipeline.AddStatusFromSwagger(idFactory, configuration),
+		// Import data from Swagger:
+		pipeline.LoadTypesFromSwagger(idFactory, configuration),
+		pipeline.SimplifySwaggerNames(idFactory, configuration),
+		// De-pluralize resource types
+		// (Must come after type aliases are resolved)
+		// pipeline.ImproveResourcePluralization(),
```
theunrepentantgeek, line 90: Is this prerequisite declared using pre- or post-requisites?

v2/tools/generator/internal/codegen/code_generator.go
```diff
@@ -82,50 +82,56 @@ func NewCodeGeneratorFromConfig(configuration *config.Configuration, idFactory a
...
 		pipeline.ConvertAllOfAndOneOfToObjects(idFactory),
+		pipeline.StripUnreferencedTypeDefinitions(),
 
 		// Flatten out any nested resources created by allOf, etc. we want to do this before naming types or things
 		// get named with names like Resource_Spec_Spec_Spec:
 		pipeline.FlattenResources(),
 
 		// Copy additional swagger-derived information from status into spec
-		pipeline.AugmentSpecWithStatus().RequiresPrerequisiteStages("allof-anyof-objects", "addStatusFromSwagger"),
+		// pipeline.AugmentSpecWithStatus().RequiresPrerequisiteStages("allof-anyof-objects", "addStatusFromSwagger"),
```
theunrepentantgeek, line 101: Do we need to keep this commented out code? If so, move the pre-requisite into the stage definition.

theunrepentantgeek:
With #2323 merged, we no longer need to keep this open for reference.

PR #2087, opened 2022-02-11
Register extensions with reconciler

matthchr:


super-harsh:
Changes looking good 😄

v2/internal/controllers/generic_controller.go
```diff
@@ -176,7 +187,8 @@ func register(
 		RequeueDelayOverride: options.RequeueDelay,
 		PositiveConditions:   conditions.NewPositiveConditionBuilder(clock.New()),
 		//nolint:gosec // do not want cryptographic randomness here
-		rand: rand.New(lockedrand.NewSource(time.Now().UnixNano())),
+		rand:       rand.New(lockedrand.NewSource(time.Now().UnixNano())),
```
babbageclunk, line 190: What's the rationale for `rand` and `extensions` to be unexported when everything else on `GenericReconciler` is exported? It's a bit odd to have a mixture, particularly on an exported struct for fields that are required (I think?) - it means that some code in another package can never construct a valid instance.

v2/internal/controllers/generic_controller.go
```diff
@@ -289,7 +301,8 @@ func (gr *GenericReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		gr.ResourceResolver,
 		gr.PositiveConditions,
 		gr.Config,
-		gr.rand)
+		gr.rand,
```
babbageclunk, line 304: This constructor pulls a lot of things off gr to pass in - could it just take a reference to gr and use the fields directly? I guess that would couple them, but it kind of seems like they're coupled anyway.

v2/internal/reconcilers/azure_generic_arm_reconciler.go
```diff
@@ -80,6 +80,7 @@ type AzureDeploymentReconciler struct {
 	PositiveConditions *conditions.PositiveConditionBuilder
 	config             config.Values
```
babbageclunk, line 81: Ha, I guess this is me doing the exported/unexported thing on `AzureDeploymentReconciler` that I'm complaining about above, but I think passing the generic reconciler would avoid that?

v2/cmd/controller/main.go
```diff
@@ -86,7 +87,20 @@ func main() {
 	log.V(Status).Info("Configuration details", "config", cfg.String())
 
 	if cfg.OperatorMode.IncludesWatchers() {
-		if errs := controllers.RegisterAll(mgr, clientFactory, controllers.GetKnownStorageTypes(), makeControllerOptions(log, cfg)); errs != nil {
+		var extensions map[schema.GroupVersionKind]genruntime.ResourceExtension
+		extensions, err = controllers.GetResourceExtensions(scheme)
```
babbageclunk, line 91: Why not just ` extensions, err := ...` here? It seems like contortions to avoid shadowing `err` from outside the `if`, but is there a reason to do that?

babbageclunk (approved):
This looks good but a few questions/suggestions about struct fields and long parameter lists.

v2/cmd/controller/main.go
theunrepentantgeek: The linting pass in our CI complains bitterly about shadowing variables, even just for `err`. I've had to do this dance a couple of times myself. I understand _why_ it complains, and believe it's a good check to have. But just occasionally, it forces a bit of break-dancing.

gif

v2/internal/controllers/generic_controller.go
```diff
@@ -100,7 +101,13 @@ func registerWebhook(mgr ctrl.Manager, obj client.Object) error {
 	return ctrl.NewWebhookManagedBy(mgr).For(obj).Complete()
 }
 
-func RegisterAll(mgr ctrl.Manager, clientFactory ARMClientFactory, objs []registration.StorageType, options Options) error {
+func RegisterAll(
+	mgr ctrl.Manager,
+	clientFactory ARMClientFactory,
+	objs []registration.StorageType,
+	extensionMap map[schema.GroupVersionKind]genruntime.ResourceExtension,
```
theunrepentantgeek, line 108: Suggest simply naming this for what it contains, not the type.
```suggestion
	extensions map[schema.GroupVersionKind]genruntime.ResourceExtension,
```

v2/internal/controllers/generic_controller.go
```diff
@@ -134,6 +143,7 @@ func register(
 	reconciledResourceLookup map[schema.GroupKind]schema.GroupVersionKind,
 	clientFactory ARMClientFactory,
 	info registration.StorageType,
+	extensionsMap map[schema.GroupVersionKind]genruntime.ResourceExtension,
```
theunrepentantgeek, line 146: Similarly, perhaps call this just `extensions`.
```suggestion
	extensions map[schema.GroupVersionKind]genruntime.ResourceExtension,
```

v2/internal/controllers/generic_controller.go
theunrepentantgeek: I agree, it does kinda feel like the object-envy code smell.

v2/internal/reconcilers/azure_generic_arm_reconciler.go
```diff
@@ -93,7 +94,8 @@ func NewAzureDeploymentReconciler(
 	resourceResolver *genruntime.Resolver,
 	positiveConditions *conditions.PositiveConditionBuilder,
 	cfg config.Values,
-	rand *rand.Rand) *AzureDeploymentReconciler {
+	rand *rand.Rand,
+	extensions genruntime.ResourceExtension) *AzureDeploymentReconciler {
```
theunrepentantgeek, line 98: Why does this have a plural name for a single instance of the interface?

theunrepentantgeek (changes requested):

theunrepentantgeek:
Not so much "Request changes" as "I have a couple of suggestions I'd like considered before I approve". 😁

v2/cmd/controller/main.go
matthchr: Yeah this is to make the linter happy. Happy to do something else but it's not clear what's better than this. I could suppress the linter here (seems like if I'm doing that I should just turn it off?)

v2/internal/controllers/generic_controller.go
matthchr: We can't pass `GenericReconciler` here because that would introduce a cyclic dependency, where `GenericReconciler` depends on `AzureDeploymentReconciler` (like it does already), but `AzureDeploymentReconciler` depends on `GenericReconciler`.

The intent/goal (that has never been fully realized) is:
`GenericReconciler`: Implements `reconcile.Reconciler` (the controller-runtime interface). Performs some logging setup, gets the initial object and ensures it implements `genruntime.MetaObject` (our base interface all types must implement). Defers actual reconciliation to a separate interface `genruntime.Reconciler`.

`AzureDeploymentReconciler`: Generic reconciler that uses ARM PUT to reconcile resources. Implements `genruntime.Reconciler`. Is provided with logging and a `genruntime.MetaObject`.

There could be other implementers of `genruntime.Reconciler` that took a `genruntime.MetaObject` but reconciled them differently (for say, handcrafted resources that cannot use the ARM API such as SQL users). These would still be a `genruntime.MetaObject` and have the same logging, etc setups, but the actual reconcile logic would differ.

The way it probably _ought_ to work is that when we register resource types we also have an affiliated reconciler type that we infer somehow, and when we construct the `GenericReconciler` we would also include the associated `genruntime.Reconciler`. This way you have an outer reconciler that's the same for all resources, but an inner one that can differ. The outer ensures uniform logging setup, etc while the inner is free to do as it pleases.

This is part of what the comment above saying `TODO: We need some factory-lookup here` is getting at. Really `GenericReconciler` probably should not be new-ing up the `AzureDeploymentReconciler` because that pins it to only a single inner impl when we really want many possible.

v2/internal/reconcilers/azure_generic_arm_reconciler.go
matthchr: As mentioned above, can't pass it as it'd created a circular reference, and passing it goes against the actual intent of the relationship of these objects. The real solution is to construct `AzureDeploymentReconciler` first and pass _that_ to `GenericReconciler`. Then we can delete almost all of the `GenericReconciler` parameters as it doesn't need them directly (it only needs them to give them to `AzureDeploymentReconciler`.

matthchr: Also, all of these should probably be unexported. I've made a note to fix that but will do it in a separate PR

v2/cmd/controller/main.go
theunrepentantgeek: The only case where shadowing might be acceptable is for `err` - being unable to shadow that seems a small price to ensure we catch other cases.

v2/internal/reconcilers/azure_generic_arm_reconciler.go
theunrepentantgeek: Ah, I was thinking that there was only one _instance_ being passed in, so it should have a singular name, regardless of how many extension-points it implements.

theunrepentantgeek (approved):

PR #2090, opened 2022-02-14
Add support for Compute/Image and Compute/Snapshot

### LINKED ISSUE #2057: Feature: Add support for Microsoft.Compute/images and Microsoft.Compute/snapshots
These resources would pair well with existing `Microsoft.Compute/disks` and `Microsoft.Compute/virtualMachines` support.

This was requested by a customer.

theunrepentantgeek:

Adds support for two new resources - Compute/Image and Compute/Snapshot.

Closes #2057 

gif

**If applicable**:
- [x] this PR contains tests


v2/internal/controllers/crd_compute_image_test.go
```diff
@@ -0,0 +1,81 @@
...
+						},
+					},
+				},
+			},
+		},
+	}
+
+	tc.CreateResourceAndWait(image)
+	tc.Expect(image.Status.Id).ToNot(BeNil())
+	imageARMId := *image.Status.Id
```
matthchr, line 70: Possibly worth asserting a bit about the structure as well? For example that the disk size is set to what it should be, etc?

Don't need to exhaustively check everything, but 1-2 key properties is usually good.

v2/internal/controllers/crd_compute_image_test.go
```diff
@@ -0,0 +1,81 @@
...
+
```
matthchr, line 71: minor: Most tests we try to do a  patch/update section too (just to ensure patch/update works and catch any possible issues in the RP like what @babbageclunk found where they didn't accept PUT for update). See for example the `crd_networking_publicip_test.go`

v2/internal/controllers/crd_compute_snapshot_test.go
```diff
@@ -0,0 +1,60 @@
...
+				CreateOption: compute.CreationDataCreateOptionEmpty,
+			},
+			DiskSizeGB: to.IntPtr(32),
+			Location:   tc.AzureRegion,
+			Owner:      testcommon.AsOwner(rg),
+		},
+	}
+
+	tc.CreateResourceAndWait(snapshot)
+	tc.Expect(snapshot.Status.Id).ToNot(BeNil())
```
matthchr, line 39: Same comment, assert on a few key properties (disk size?)

matthchr (approved):

v2/internal/controllers/crd_compute_image_test.go
theunrepentantgeek: I didn't find any properties that could be patched after the initial create.

v2/config/samples/compute/v1alpha1api20210701_image.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: compute.azure.com/v1alpha1api20201201
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

v2/internal/controllers/crd_compute_image_test.go
theunrepentantgeek: Added more asserts.

v2/internal/controllers/crd_compute_snapshot_test.go
theunrepentantgeek: Added asserts.

v2/internal/controllers/crd_compute_image_test.go
matthchr: You can almost always patch `Tags`, but it's not all that interesting since we use that trick in lots of other places already so it's unlikely it's going to be broken for just this resource.

PR #2093, opened 2022-02-15
Add SecretRetrieval extension hooks and first implementation

### LINKED ISSUE #1471: Determine plan for managing secrets
Once an ARM service has been created there are some which have keys or secrets associated with them.

For example, a storage account has storage keys, a Redis account has keys, etc. These keys need to be stored and access has to be given to the right users.

ASO historically had two approaches for this, originally storing secrets using Kubernetes secrets, and now optionally supports storing secrets in a KV instead (value for this is controlled via env variable):
`AZURE_OPERATOR_KEYVAULT`

Take a look at ASO rediscache_types.go for example:
```
	SecretName             string               `json:"secretName,omitempty"`
	KeyVaultToStoreSecrets string               `json:"keyVaultToStoreSecrets,omitempty"`
```

matthchr:
* Add extensions.SecretRetrieval extension interface.
* Add hook in azure_generic_arm_reconciler where the SecretRetrieval
  interface is called (if implemented).
* Add first implementation of SecretRetrieval interface on storage
  account.
* Add tests focused on generic secret write logic in controller
  via secrets_from_azure_test.go.
* Add tests focused on secret retrieval of storage account
  in crd_storage-storageaccount_test.go.

Closes #1471

**If applicable**:
- [x] this PR contains tests


matthchr:
Will hold off on merging this release until alpha-6 ships

v2/api/storage/customizations/storage_account_extensions.go
```diff
@@ -0,0 +1,136 @@
...
+
+func (ext *StorageAccountExtension) RetrieveSecrets(
+	ctx context.Context,
+	obj genruntime.MetaObject,
+	armClient *genericarmclient.GenericClient,
+	log logr.Logger) ([]*v1.Secret, error) {
+
+	// This has to be the current storage version. It will need to be updated
+	// if the storage version changes.
+	typedObj, ok := obj.(*storage.StorageAccount)
```
theunrepentantgeek, line 35: Add in a type assertion that `storage.StorageAccount` implements `Hub` - that way we'll get a compilation error here if/when the storage version changes.

v2/internal/controllers/secrets_from_azure_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+	// Create a storage account
+	storageKeysSecret := "storagekeys"
+	acct := makeSimpleStorageAccountWithOperatorSpecSecrets(tc, rg, storageKeysSecret, "key1")
+
+	tc.CreateResourceAndWait(acct)
+
+	// Expect that the secret exists and has an ownerReference set. We can't actually ensure
+	// that the resource is deleted in envtest because the controller that performs ownerReference
+	// resource deletion doesn't run in envtest.
```
theunrepentantgeek, line 37: Is this "doesn't run" == "doesn't work" or "doesn't run" == "isn't run (but we could)" ?

theunrepentantgeek (approved):

v2/internal/controllers/secrets_from_azure_test.go
matthchr: doesn't work. Will clarify. That component isn't actually part of apiserver and envtest only emulates apiserver, so some of the "background processes" that exist in a normal cluster don't work in envtest

v2/api/storage/customizations/storage_account_extensions.go
matthchr: Did this. Great idea.

PR #2098, opened 2022-02-16
Fix bug where Condition is not set in some error cases

matthchr:
- Sets a Ready Condition with severity=Warning if a secret reference
  or cross-resource reference could not be found.
- Adds tests to ensure that the error is set and delete of the resource
  can still go through.

**If applicable**:
- [x] this PR contains tests


theunrepentantgeek (approved):
LGTM.

gif

PR #2099, opened 2022-02-16
Create an extension point for customization of error classification

### LINKED ISSUE #1997: Feature: ASOv2 should support per resource type error handling overrides
There are some error codes returned from Azure that have somewhat ambiguous meaning. One of them is `BadRequest`. It often means that a request is fatally formed and cannot ever succeed, but there are certain services where it doesn't always mean that, for example in CosmosDB that code is returned when attempting to create a database at the same time as the parent DatabaseAccount, which _should_ be retried.

The solution is to have customizable error handling for these special cases.

theunrepentantgeek:

Creates an extension point where a resource can augment the the normal error classification process of the generic reconciler and provide its own rules.

Closes #1997

gif

### DIFF
--- a/v2/api/cache/customizations/redis_extension_types.go
+++ b/v2/api/cache/customizations/redis_extension_types.go
@@ -1,0 +1,47 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+package customizations
+
+import (
+	"strings"
+
+	"github.com/Azure/go-autorest/autorest/to"
+	"github.com/go-logr/logr"
+
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/core"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/extensions"
+)
+
+var _ extensions.ErrorClassifier = &RedisExtension{}
+
+// ClassifyError evaluates the provided error, returning including whether it is fatal or can be retried.
+// A conflict error (409) is normally fatal, but Redis resources may return 409 whilst a dependency is being created,
+// so we override for that case.
+// cloudError is the error returned from ARM.
+// apiVersion is the ARM API version used for the request.
+// log is a logger than can be used for telemetry.
+// next is the next implementation to call.
+func (e *RedisExtension) ClassifyError(
+	cloudError *core.CloudError,
+	apiVersion string,
+	log logr.Logger,
+	next extensions.ErrorClassifierFunc) (core.CloudErrorDetails, error) {
+	details, err := next(cloudError)
+	if err != nil {
+		return details, err
+	}
+
+	// Override is to treat Conflict as retryable for Redis, if the message contains "try again later"
+	//TODO: Do we need to check the message, or was that just a discriminator for when the code was generic?
+	inner := cloudError.InnerError
+	if details.Classification == core.ErrorFatal && to.String(inner.Code) == "Conflict" {
+		if inner.Message != nil && strings.Contains(strings.ToLower(*inner.Message), "try again later") {
+			details.Classification = core.ErrorRetryable
+		}
+	}
+
+	return details, nil
+}
--- a/v2/api/cache/customizations/redis_extension_types.go
+++ b/v2/api/cache/customizations/redis_extension_types.go
@@ -12,6 +12,7 @@ import (
 	"github.com/go-logr/logr"
 
 	"github.com/Azure/azure-service-operator/v2/internal/genericarmclient"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/core"
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/extensions"
 )
 
@@ -28,10 +29,10 @@ func (e *RedisExtension) ClassifyError(
 	cloudError *genericarmclient.CloudError,
 	apiVersion string,
 	log logr.Logger,
-	next extensions.ErrorClassifierFunc) (genericarmclient.CloudErrorDetails, error) {
+	next extensions.ErrorClassifierFunc) (core.CloudErrorDetails, error) {
 	details, err := next(cloudError)
 	if err != nil {
-		return genericarmclient.CloudErrorDetails{}, err
+		return core.CloudErrorDetails{}, err
 	}
 
 	// Override is to treat Conflict as retryable for Redis, if the message contains "try again later"
@@ -40,7 +41,7 @@ func (e *RedisExtension) ClassifyError(
 		inner := cloudError.InnerError
 		if to.String(inner.Code) == "Conflict" &&
 			strings.Contains(strings.ToLower(*inner.Message), "try again later") {
-			details.Classification = genericarmclient.ErrorRetryable
+			details.Classification = core.ErrorRetryable
 		}
 	}
 
--- a/v2/api/cache/customizations/redis_extension_types.go
+++ b/v2/api/cache/customizations/redis_extension_types.go
@@ -35,10 +35,13 @@ func (e *RedisExtension) ClassifyError(
 	}
 
 	// Override is to treat Conflict as retryable for Redis, if the message contains "try again later"
-	//TODO: Do we need to check the message, or was that just a discriminator for when the code was generic?
-	inner := cloudError.InnerError
-	if details.Classification == genericarmclient.ErrorFatal && to.String(inner.Code) == "Conflict" {
-		if inner.Message != nil && strings.Contains(strings.ToLower(*inner.Message), "try again later") {
+	// TODO: Do we need to check the message, or was that just a discriminator for when the code was generic?
+	if details.Classification == genericarmclient.ErrorFatal &&
+		cloudError.InnerError != nil &&
+		cloudError.InnerError.Message != nil {
+		inner := cloudError.InnerError
+		if to.String(cloudError.InnerError.Code) == "Conflict" &&
+			strings.Contains(strings.ToLower(*inner.Message), "try again later") {
 			details.Classification = genericarmclient.ErrorRetryable
 		}
 	}
--- a/v2/api/cache/customizations/redis_extension_types.go
+++ b/v2/api/cache/customizations/redis_extension_types.go
@@ -18,7 +18,7 @@ import (
 
 var _ extensions.ErrorClassifier = &RedisExtension{}
 
-// ClassifyError evaluates the provided error, returning including whether it is fatal or can be retried.
+// ClassifyError evaluates the provided error, returning whether it is fatal or can be retried.
 // A conflict error (409) is normally fatal, but Redis resources may return 409 whilst a dependency is being created,
 // so we override for that case.
 // cloudError is the error returned from ARM.
@@ -36,14 +36,19 @@ func (e *RedisExtension) ClassifyError(
 	}
 
 	// Override is to treat Conflict as retryable for Redis, if the message contains "try again later"
-	if cloudError.InnerError != nil &&
-		cloudError.InnerError.Message != nil {
-		inner := cloudError.InnerError
-		if to.String(inner.Code) == "Conflict" &&
-			strings.Contains(strings.ToLower(*inner.Message), "try again later") {
-			details.Classification = core.ErrorRetryable
-		}
+	if isRetryableConflict(cloudError.InnerError) {
+		details.Classification = core.ErrorRetryable
 	}
 
 	return details, nil
+}
+
+// isRetryableConflict checks the passed error to see if it is a retryable conflict, returning true if it is.
+func isRetryableConflict(err *genericarmclient.ErrorResponse) bool {
+	if err == nil || err.Message == nil {
+		return false
+	}
+
+	return to.String(err.Code) == "Conflict" &&
+		strings.Contains(strings.ToLower(*err.Message), "try again later")
 }
--- a/v2/api/cache/customizations/redis_extension_types.go
+++ b/v2/api/cache/customizations/redis_extension_types.go
@@ -11,7 +11,7 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/go-logr/logr"
 
-	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/core"
+	"github.com/Azure/azure-service-operator/v2/internal/genericarmclient"
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/extensions"
 )
 
@@ -25,10 +25,10 @@ var _ extensions.ErrorClassifier = &RedisExtension{}
 // log is a logger than can be used for telemetry.
 // next is the next implementation to call.
 func (e *RedisExtension) ClassifyError(
-	cloudError *core.CloudError,
+	cloudError *genericarmclient.CloudError,
 	apiVersion string,
 	log logr.Logger,
-	next extensions.ErrorClassifierFunc) (core.CloudErrorDetails, error) {
+	next extensions.ErrorClassifierFunc) (genericarmclient.CloudErrorDetails, error) {
 	details, err := next(cloudError)
 	if err != nil {
 		return details, err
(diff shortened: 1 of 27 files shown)


v2/pkg/genruntime/extensions/error_classifier.go
```diff
@@ -0,0 +1,59 @@
...
+		cloudError *core.CloudError,
+		apiVersion string,
+		log logr.Logger,
+		next ErrorClassifierFunc) (core.CloudErrorDetails, error)
+}
+
+// ErrorClassifierFunc is the signature of a function that can be used to create a DefaultErrorClassifier
+type ErrorClassifierFunc func(cloudError *core.CloudError) (core.CloudErrorDetails, error)
+
+func CreateErrorClassifier(
```
matthchr, line 33: I like this pattern of the wrapper. I'll adopt it for my secrets extension in a future PR.

v2/internal/genericarmclient/errors.go
```diff
@@ -1,50 +0,0 @@
-/*
-Copyright (c) Microsoft Corporation.
-Licensed under the MIT license.
-*/
-
-package genericarmclient
```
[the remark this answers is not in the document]
matthchr: Was there a need for these error types to move because of a circular reference? Can you expand on what that circular reference was? At least at first blush I don't see the issue with the extension for customizing error classifications from Azure importing `genericarmclient` which defines what the known Azure payloads are.

v2/internal/reconcilers/azure_generic_arm_reconciler.go
```diff
@@ -250,21 +253,31 @@ func (r *AzureDeploymentReconciler) GetReconcilePolicy() ReconcilePolicy {
...
+	case core.ErrorRetryable:
 		severity = conditions.ConditionSeverityWarning
-	case CloudErrorFatal:
+	case core.ErrorFatal:
 		severity = conditions.ConditionSeverityError
 		// This case purposefully does nothing as the fatal provisioning state was already set above
 	default:
 		// TODO: Is panic OK here?
-		panic(fmt.Sprintf("Unknown error classification %q", errorDetails.Classification))
+		panic(fmt.Sprintf("Unknown error classification %q", details.Classification))
```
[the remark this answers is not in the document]
matthchr: (I think I was trying to avoid an `err` return here and this case is a bug in our code so panic felt ok-ish, but if we have an error return now anyway we might as well use it for this too)

v2/api/cache/customizations/redis_extension_types.go
```diff
@@ -0,0 +1,47 @@
...
+// so we override for that case.
+// cloudError is the error returned from ARM.
+// apiVersion is the ARM API version used for the request.
+// log is a logger than can be used for telemetry.
+// next is the next implementation to call.
+func (e *RedisExtension) ClassifyError(
+	cloudError *core.CloudError,
+	apiVersion string,
+	log logr.Logger,
+	next extensions.ErrorClassifierFunc) (core.CloudErrorDetails, error) {
```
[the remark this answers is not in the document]
theunrepentantgeek: @Porges suggested the model that I adopted here. 

Passing the default logic as `next` allows extensions to work in four different ways:

* Add custom behavior only _before_ the default
* Add custom behaviour only _after_ the default 
* Wrap the default behaviour in custom logic
* Completely replace the functionality (ignoring `next`)

The first two are probably going to cover most cases (though not all) and failing to provide easy access to the default behaviour risks extensions reinventing the wheel. If they do so in a way accidentally inconsistent with the default behaviour, we might get some surpsing edge cases that are hard to diagnose. We end up with the same situation if they _copy_ the default behaviour (violating DRY), and then we fix a bug in the original.

> since the purpose of the extension is to override the default

That's the fourth case listed above; I think it's going to be far more common for extensions to _augment_ the default.

v2/api/cache/customizations/redis_extension_types.go
```diff
@@ -0,0 +1,47 @@
...
+	details, err := next(cloudError)
+	if err != nil {
+		return details, err
+	}
+
+	// Override is to treat Conflict as retryable for Redis, if the message contains "try again later"
+	//TODO: Do we need to check the message, or was that just a discriminator for when the code was generic?
+	inner := cloudError.InnerError
```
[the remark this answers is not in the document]
theunrepentantgeek: Possibly. Worth being cautious.

v2/api/cache/customizations/redis_extension_types.go
```diff
@@ -0,0 +1,47 @@
...
+// log is a logger than can be used for telemetry.
+// next is the next implementation to call.
+func (e *RedisExtension) ClassifyError(
+	cloudError *core.CloudError,
+	apiVersion string,
+	log logr.Logger,
+	next extensions.ErrorClassifierFunc) (core.CloudErrorDetails, error) {
+	details, err := next(cloudError)
+	if err != nil {
+		return details, err
```
super-harsh, line 34: do we still need to return details here or return nil, err?

v2/api/cache/customizations/redis_extension_types.go
super-harsh, line 39: Would be worth checking for `nil` here.

v2/api/cache/customizations/redis_extension_types.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	// Override is to treat Conflict as retryable for Redis, if the message contains "try again later"
+	//TODO: Do we need to check the message, or was that just a discriminator for when the code was generic?
+	inner := cloudError.InnerError
+	if details.Classification == core.ErrorFatal && to.String(inner.Code) == "Conflict" {
+		if inner.Message != nil && strings.Contains(strings.ToLower(*inner.Message), "try again later") {
```
super-harsh, line 41: Would it be better to check error code AND message here to handle just the special case.

v2/internal/reconcilers/azure_generic_arm_reconciler.go
super-harsh, line 276: `Panic` here would stop the execution and kill the thread, rather we can return/set error here.

v2/internal/reconcilers/error_classifier.go
```diff
@@ -39,33 +24,35 @@ func stringOrDefault(str string, def string) string {
 	return str
 }
 
-func ClassifyCloudError(err *genericarmclient.CloudError) CloudErrorDetails {
+func ClassifyCloudError(err *core.CloudError) (core.CloudErrorDetails, error) {
```
super-harsh, line 27: In this function we're always returning `error` as `nil`, do we still need to return an `error`? I think linter won't pass this.

v2/internal/genericarmclient/errors.go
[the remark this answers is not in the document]
theunrepentantgeek: Done. I think this works.

v2/internal/reconcilers/azure_generic_arm_reconciler.go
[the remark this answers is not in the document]
theunrepentantgeek: Done.

v2/api/cache/customizations/redis_extension_types.go
```diff
@@ -0,0 +1,47 @@
...
+	apiVersion string,
+	log logr.Logger,
+	next extensions.ErrorClassifierFunc) (core.CloudErrorDetails, error) {
+	details, err := next(cloudError)
+	if err != nil {
+		return details, err
+	}
+
+	// Override is to treat Conflict as retryable for Redis, if the message contains "try again later"
+	//TODO: Do we need to check the message, or was that just a discriminator for when the code was generic?
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

v2/api/cache/customizations/redis_extension_types.go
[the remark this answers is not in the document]
theunrepentantgeek: Done.

v2/api/cache/customizations/redis_extension_types.go
theunrepentantgeek: Returning just `err` makes sense; can't return **nil**, it's a struct.

v2/api/cache/customizations/redis_extension_types.go
theunrepentantgeek: Fixed.

v2/api/cache/customizations/redis_extension_types.go
theunrepentantgeek: We _do_ check both: error code must be `"Conflict"` and message must contain "try again later".

v2/internal/reconcilers/azure_generic_arm_reconciler.go
theunrepentantgeek: Done.

v2/internal/reconcilers/error_classifier.go
theunrepentantgeek: We need to match the signature of the extension point, even though this version can never error. 

I _hope_ it doesn't trigger the linter!

v2/internal/genericarmclient/error_classification.go
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package genericarmclient
+
+type ErrorClassification string
```
matthchr, line 8: minor: So originally this was in `reconcilers`, which probably caused issues with circular references?

I'm cool with _this_ being in `core` if you want, as _this_ is the reconcilers classification of an error as retryable or fatal. (It's not as about the wire format as the other stuff).

I'm also OK with it here though, but just wanted to clarify that if you felt this made more sense in `core` I could get behind that, since it's not wire-format stuff but rather instructions for the reconciler(s).

v2/internal/genericarmclient/cloud_error_details.go
```diff
@@ -0,0 +1,12 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+package genericarmclient
+
+type CloudErrorDetails struct {
```
matthchr, line 8: Same as below, I'd be OK with this in `core` if you prefer, as it's not a wire-format thing (it's the controllers classification of a wire-format thing). Sorry I wasn't clear before.

v2/api/cache/customizations/redis_extension_types.go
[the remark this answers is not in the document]
matthchr: I was expecting that the extension for error handling would always make a decision based on the response from Azure. There would be no need for "before or after", because the extension isn't making it's choice based on what the default classifier does, it's overriding _part_ of the classifier (your augment case). Then if the extension didn't make any classification, the default would apply.

Something like this:

```
type ErrorClassifier interface {
     func ClassifyError(
         cloudError *genericarmclient.CloudError,
        apiVersion string,
        log logr.Logger) (*genericarmclient.CloudErrorDetails, error)
}

func (e *RedisExtension) ClassifyError(
    cloudError *genericarmclient.CloudError,
    apiVersion string,
    log logr.Logger) (genericarmclient.CloudErrorDetails, error) {

    if cloudError.Inner != nil && cloudError.Inner.Code == "Conflict" && inner.Message != nil && strings.Contains(strings.ToLower(*inner.Message), "try again later") {
        return core.MakeCloudErrorDetails(cloudError, classification) // This is a new method that extracts the code/message to their respective fields
    }

    return nil
}

func CreateErrorClassifier(
    host genruntime.ResourceExtension,
    classifier ErrorClassifierFunc,
    apiVersion string,
    log logr.Logger) ErrorClassifierFunc {
    impl, ok := host.(ErrorClassifier)
    if !ok {
        return classifier
    }

    return func(cloudError *genericarmclient.CloudError) (genericarmclient.CloudErrorDetails, error) {
        log.V(Info).Info(
            "Classifying cloud error",
            "Message", cloudError.InnerError.Message,
            "Code", cloudError.InnerError.Code,
            "Target", cloudError.InnerError.Target)

        result, err := impl.ClassifyError(cloudError, apiVersion, log, classifier)
        if err != nil {
            return genericarmclient.CloudErroDetails{}, err
        }
        if result == nil {
            result, err = classifier(cloudError)
        }

        log.V(Info).Info(
            "Cloud error classified",
            "Classification", result.Classification,
            "Code", result.Code,
            "Message", result.Message)
        return result, err
    }
}
```

I'm ok with it being more powerful, I just think it could lead people to write code like you did where you're checking "what did the default classifier do": [if details.Classification == genericarmclient.ErrorFatal](https://github.com/Azure/azure-service-operator/pull/2099/files#diff-420d9866ee41deef2e288d2da9accce7f98f9fed82823b633f68695db21c3829R38), which feels weird to me. Why does the extension care what the default classifier classified this error as? It _knows_ that this error is nonfatal in this context, why does it matter how the `next` function classified it?

With all that said, this is pretty minor! Maybe I'm just not seeing the cases where the actual extensions will care about what the default classifier thought and only want to make modifications based on that.

v2/internal/genericarmclient/cloud_error_details.go
theunrepentantgeek: All good. I've moved `CloudErrorDetails` and `ErrorClassification` back into `core`, leaving the others in `genericarmclient`.

v2/internal/genericarmclient/error_classification.go
theunrepentantgeek: > which probably caused issues with circular references?

Yes! I'd forgotten the details. Will move this back to `core`.

v2/internal/reconcilers/error_classifier.go
```diff
@@ -87,19 +74,9 @@ func classifyInnerCloudError(err *genericarmclient.ErrorResponse) CloudErrorClas
...
-		}
-		if strings.Contains(strings.ToLower(*err.Message), "try again later") {
-			return CloudErrorRetryable
-		}
-		return CloudErrorFatal
+		return genericarmclient.ErrorRetryable
 	case "BadRequestFormat",
+		"Conflict",
 		// TODO: See https://github.com/Azure/azure-service-operator/issues/1997 for why this is commented out
 		// "BadRequest",
```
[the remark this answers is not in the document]
theunrepentantgeek: I will fix this - but will do so in a separate PR; logged as #2106 to make sure I don't forget.

v2/pkg/genruntime/core/error_classification.go
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package core
```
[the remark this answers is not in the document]
theunrepentantgeek: I don't think we need to do so just yet - but we should consider breaking up `core` if it does get large.

v2/api/cache/customizations/redis_extension_types.go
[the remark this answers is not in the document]
theunrepentantgeek: I took out the check to see what the default classifier did as it was redundant.

> I'm just not seeing the cases where the actual extensions will care about what the default classifier thought and only want to make modifications based on that.

There are two cases. The first is when the error classification extension needs to do something we haven't anticipated. The other is that I'd like most of the extension points to have parallel structure, and we can foresee extensions where they might want to do that (or might want to _sometimes_ skip the default behaviour completely).

v2/api/cache/customizations/redis_firewall_rule_extension_types.go.go
```diff
@@ -0,0 +1,49 @@
...
+var _ extensions.ErrorClassifier = &RedisFirewallRuleExtension{}
+
+// ClassifyError evaluates the provided error, returning including whether it is fatal or can be retried.
+// A conflict error (409) is normally fatal, but RedisFirewallRule resources may return 409 whilst a dependency is being
+// created, so we override for that case.
+// cloudError is the error returned from ARM.
+// apiVersion is the ARM API version used for the request.
+// log is a logger than can be used for telemetry.
+// next is the next implementation to call.
+func (e *RedisFirewallRuleExtension) ClassifyError(
```
babbageclunk, line 28: I think this method is the same between all 4 types - any reason the methods can't all call the same helper function?

babbageclunk (approved):
This looks good - I think the 4 ClassifyError methods could call one helper function though.

v2/api/cache/customizations/redis_firewall_rule_extension_types.go.go
theunrepentantgeek: Good idea. I factored out a common predicate function for the test.

PR #2100, opened 2022-02-17
Feature/upgrade cert manager resource versions

mkretz:
We need to deploy v1 of the operator via Helm on an AKS cluster where a more recent version of cert-manager is already deployed. Hence, we require an option to configure the API version of the cert-manager custom resources used in the Helm chart.

The change is backward compatible as the API versions default to the old values. We also added a note in the v1 README.

gif

**If applicable**:
- [ x ] this PR contains documentation
- [ ] this PR contains tests


matthchr:
Thanks for this @mkretz! One issue is that the files in `charts/azure-service-operator` are generated, and you changed the generated files, which will end up getting deleted if/when we regenerate the chart via the `Makefile`.

Rather than try to explain what changes you need to make in our build process (which is a smidge messy), I've just gone ahead and adopted your changes and created a fresh PR that modifies the build process so that the generated files don't need to be changed manually. The PR is #2103. 

Let me know if that looks good to you and we can close this in favor of that and merge it.

Thanks!

mkretz:
Looks good to me @matthchr! Please go ahead with #2103 and yes, I will close this one. Sorry, I didn't pay attention to the fact that I was changing generated code. Thanks for correcting that!

mkretz:
Closing this in favour of #2103. Thanks @matthchr!

PR #2101, opened 2022-02-17
Add Redis secrets extension

matthchr:
**If applicable**:
- [x] this PR contains tests


v2/api/cache/customizations/redis_extensions.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	if secrets.HostName != nil ||
+		secrets.Port != nil ||
+		secrets.SSLPort != nil {
+		hasEndpoints = true
+	}
+
+	return hasSecrets, hasEndpoints
+}
```
theunrepentantgeek, line 99: Not for this PR, but perhaps we should consider generating this, adding as a function on `Spec` at the same time we add the `Secrets` property.

theunrepentantgeek (approved):
Looks great!

v2/api/cache/customizations/redis_extensions.go
matthchr: We could, although we need to make the `azure-arm.yaml` config a bit more intelligent then in order to do it, as we have to classify secrets that come from the ListSecrets call vs "secrets" that come from the Status and are just there for convenience.

This gets _extra_ complicated with Cosmos where there are actually 2 list secrets calls and 3 categories (secrets, readonly secrets, and endpoints).

I'll make a note to look into it though

PR #2103, opened 2022-02-18
Support newer versions of cert manager

### LINKED ISSUE #2021: Bug: cert-manager v1.6 support
mcr.microsoft.com/k8s/azureserviceoperator:1.0.22275

**Describe the bug**
When you have cert-manager v1.6 installed on your cluster, it will upgrade existing certificates from [deprecated](https://cert-manager.io/docs/installation/upgrading/upgrading-1.5-1.6/) `v1alpha2` to `v1` apiVersion. But as `v1alpha2` is not supported anymore, you can not upgrade your ASO to latest version as it returns an error:
```
Error: [unable to recognize "": no matches for kind "Certificate" in version "cert-manager.io/v1alpha2", unable to recognize "": no matches for kind "Issuer" in version "cert-manager.io/v1alpha2"]
```

**To Reproduce**
1. Install cert-manager via Helm v1.5
2. Install ASO v1 via Helm 1.2.0
3. Upgrade cert-manager to v1.6
4. Try to update ASO v1 to latest 1.8.0 via Helm

**Expected behavior**
ASO should be able to support latest API Version `cert-manager.io/v1`

matthchr:
**If applicable**:

Built with changes based on #2100 (Thanks @mkretz).
Fixes #2021.


Makefile
```diff
@@ -229,6 +229,8 @@ helm-chart-manifests: generate
 	# Ensure that what we're about to try to replace actually exists (if it doesn't we want to fail)
 	grep -E $(KUBE_RBAC_PROXY) ./charts/azure-service-operator/templates/generated/*_deployment_*
 	perl -pi -e s,$(KUBE_RBAC_PROXY),"{{ .Values.image.kubeRBACProxy }}",g ./charts/azure-service-operator/templates/generated/*_deployment_*
+	# replace hard coded cert-manager version with templating
+	perl -pi -e s,cert-manager.io/v1alpha2,"{{ .Values.certManagerResourcesAPIVersion }}",g ./charts/azure-service-operator/templates/generated/*cert-manager.io*
```
theunrepentantgeek, line 233: I'm reluctant to expand the dependency on perl (even though this is just for v1) - is there a simpler way?

theunrepentantgeek (approved):

Makefile
super-harsh, line 233: I think here the best practice would be to match with `cert-manager.io/<REGEX_MATCH>` and then replace. The reason behind is that it'll reduce an extra step and complexity to change in places in the future if we're changing versions.

super-harsh (changes requested):

Makefile
matthchr: Fixed, although changed to only replace `apiVersion: cert-manager.io/.*` just in case there's something else in the files that matches `cert-manager.io/` that we don't want to replace (just a little extra safety that we're only replacing the APIVersion)

matthchr: Did this with sed instead.

super-harsh (approved):
Looks good! 👍

theunrepentantgeek:
Test failure seems unrelated to the change here, so go ahead and merge.

Issue #1744, opened 2021-08-24
Don't keep "status-like" fields on spec


matthchr:
**Describe the current behavior**
The spec is modified to contain the `deployment-id` and `deployment-name` annotations in annotation to the `resource-sig` annotation. All 3 of these annotations are really representing "status-like" concepts and so should be on status.

Modifying the spec in the control loop creates multiple problems:
1. A modification triggers a new reconciliation. This isn't a problem by itself but can easily throw off deterministic execution for things like recording tests if you aren't careful to mark the current request as "done".
2. A multi-write problem in cases where there is a status update and a spec update that both need to be done. This happens after a successful deployment for example, where we get a new status and clear the `deployment-id` and `deployment-name` in the same stage.

**Describe the improvement**
The spec shouldn't be modified by the control loop, except for cases where it's really required (ownership comes to mind).

**Additional context**
In many ways this is related to #1491, #1685 and #1612

matthchr:
Closing this as these annotations are no-longer really a problem. They don't trigger fresh reconciliations and it's easier to use them than it is to use something in status. Moreover, since these are primarily used by the reconciler itself, leaving them as annotations leaves them in the "loosely-typed, probably don't depend on them" territory as opposed to `Status` which is versioned and clearly has an API contract associated with it.

PR #2104, opened 2022-02-19
Refactor generic_controller and arm_reconciler some

matthchr:
- Break azure_generic_arm_reconciler up some.
- Move azure_generic_arm_reconciler and associated types into
  their own package reconcilers/arm. There may be other reconcilers
  in the future such as for Database users, etc.
- Get rid of FatalReconciliationError in favor of new
  conditions.ReadyConditionImpactingError.
- GenericReconciler now operates on a genruntime.Reconciler interface
  instance.
- Error conditions are now mostly set at the top level by returning
  a conditions.ReadyConditionImpactingError. This has the advantage
  of making the code in our if err != nil blocks quite a bit cleaner
  as they don't need to CommitObject (which can fail).
- Fixed bug where Warning conditions were inconsistent on if they
  returned an error to the GenericReconciler or not. Now they all do,
  which ensures the reconcile will be retried.
- Fixed bug in from-azure secrets where we were setting a Warning
  condition but not returning an error or ctrl.Result{requeue: true}.
  Moved this error condition to a fatal one anyway as the user must
  take some action to correct it, so it fits our fatal error definition.

gif

**If applicable**:
- [x] this PR contains tests


matthchr:
Note that this still isn't _super_ clean, but I think it's cleaner than it was.

v2/internal/reconcilers/arm/client_helpers.go
```diff
@@ -0,0 +1,61 @@
...
+	// otherwise harmless in an actual deployment.
+	// We update the spec first to avoid the above problem.
+
+	// We must clone here because the result of this update could contain
+	// fields such as status.location that may not be set but are not omitempty.
+	// This will cause the contents we have in Status.Location to be overwritten.
+	clone := obj.DeepCopyObject().(client.Object)
+
+	// TODO: We should stop updating spec at all, except maybe for the finalizer.
+	// TODO: See: https://github.com/Azure/azure-service-operator/issues/1744
```
theunrepentantgeek, line 37: Issue #1744 is closed - is there still work to do here?

v2/internal/reconcilers/arm/client_helpers.go
```diff
@@ -0,0 +1,61 @@
...
+	err := kubeClient.Client.Update(ctx, clone)
+	if err != nil {
+		return errors.Wrap(err, "updating resource")
```
theunrepentantgeek, line 40: Should this error identify the resource so we have some way of tracing from the error to the affected resource?

v2/pkg/genruntime/secrets/errors.go
```diff
@@ -0,0 +1,50 @@
...
+// SecretNotOwnedError indicates the secret is not owned by the resource attempting to write it
+type SecretNotOwnedError struct {
+	Namespace  string
+	SecretName string
+	ObjectName string
+}
+
+var _ error = &SecretNotOwnedError{}
+
+func AsSecretNotOwnedError(err error) (*SecretNotOwnedError, bool) {
```
theunrepentantgeek, line 24: Does this need to deal with wrapped errors too? I think there's a standard `Unwrap()` method you can check for.

v2/pkg/genruntime/secrets/secrets.go
```diff
@@ -124,12 +123,11 @@ func checkSecretOwner(obj client.Object, secret *v1.Secret) error {
...
-			"cannot overwrite secret %s/%s which is not owned by %s/%s",
-			secret.GetNamespace(),
-			secret.GetName(),
-			secret.GetNamespace(),
-			obj.GetName())
+		return &SecretNotOwnedError{
+			Namespace:  secret.GetNamespace(),
+			SecretName: secret.GetName(),
+			ObjectName: obj.GetName(),
+		}
```
theunrepentantgeek, line 130: Should this be a factory method that takes just `secret` and `obj`, extracting the information desired?

theunrepentantgeek (approved):
A few comments that you may want to address, but overall looks good.

v2/internal/reconcilers/arm/client_helpers.go
matthchr: There is not really, removed this comment. Good catch

v2/internal/reconcilers/arm/client_helpers.go
matthchr: Generally speaking I think that the lower level functions usually add this, but I've gone ahead and included it here too just in case they dont.

v2/pkg/genruntime/secrets/errors.go
matthchr: `errors.As` calls that internally, so this already deals with wrapped errors.

v2/pkg/genruntime/secrets/secrets.go
matthchr: I've moved to use a `NewSecretNotOwnedError` method here, but still taking the 3 individual parameters. This is the only usage of the constructor though so this method is effectively the factory method you're talking about. I don't want to tightly couple the `SecretNotOwnedError` to always expecting those inputs though as in the future it's likely to be expanded to cover the KeyVault case too (whenever we implement that).

PR #2108, opened 2022-02-22
Custom error classification for CosmosDB

### LINKED ISSUE #2106: Fix special casing of `BadRequest` for CosmosDB

At the moment we have special behaviour in `error_classifier.go` to handle the way CosmosDB returns `BadRequest` - it normally should be a fatal error, but we're currently treating it as retryable for all resources.

**Describe the improvement**

We should move this behaviour into a per-resource extension, now that we have support for doing so.

theunrepentantgeek:
Closes #[issue number]

Moves CosmosDB and MongoDB specific error handling into specific extensions.

Closes #2106 

Update 24 Feb: Now that PR #2112 has been merged, I was able to make our CI fail by disabling the extensions, which gives me much more confidence that this works.

gif

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/api/documentdb/customizations/sql_database_extension_types.go
+++ b/v2/api/documentdb/customizations/sql_database_extension_types.go
@@ -1,0 +1,50 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+package customizations
+
+import (
+	"github.com/go-logr/logr"
+
+	"github.com/Azure/azure-service-operator/v2/internal/genericarmclient"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/core"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/extensions"
+)
+
+var _ extensions.ErrorClassifier = &SqlDatabaseExtension{}
+
+// ClassifyError evaluates the provided error, returning whether it is fatal or can be retried.
+// A BadRequest (400) is normally fatal, but CosmosDB Databases may return 400 if database creation is attempted while
+// the parent account is still being created, so we make BadRequest retryable for this case.
+// cloudError is the error returned from ARM.
+// apiVersion is the ARM API version used for the request.
+// log is a logger than can be used for telemetry.
+// next is the next implementation to call.
+func (extension *SqlDatabaseExtension) ClassifyError(
+	cloudError *genericarmclient.CloudError,
+	_ string,
+	_ logr.Logger,
+	next extensions.ErrorClassifierFunc) (core.CloudErrorDetails, error) {
+	details, err := next(cloudError)
+	if err != nil {
+		return core.CloudErrorDetails{}, err
+	}
+
+	// Override is to treat BadRequest as retryable for SqlDatabases
+	if isRetryableBadRequest(cloudError) {
+		details.Classification = core.ErrorRetryable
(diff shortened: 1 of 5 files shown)


v2/pkg/genruntime/extensions/error_classifier.go
```diff
@@ -42,11 +42,18 @@ func CreateErrorClassifier(
...
+				"Message", cloudError.InnerError.Message,
+				"Code", cloudError.InnerError.Code,
+				"Target", cloudError.InnerError.Target)
+		} else {
+			log.V(Info).Info(
+				"Classifying cloud error",
+				"Error", cloudError.Error())
+		}
 
 		result, err := impl.ClassifyError(cloudError, apiVersion, log, classifier)
```
matthchr, line 59: minor: I can't actually comment below here but I think you might want to check `err != nil` and either log a separate log for that, or just return before the log below, as if you don't you'll log "Cloud error classified Classification= "", Code="",  Message=""" which seems like it has a high probability of being confusing (since actually the error was _not_ classified and there was an error).

v2/pkg/genruntime/extensions/error_classifier.go
```diff
@@ -42,11 +42,18 @@ func CreateErrorClassifier(
 	}
 
 	return func(cloudError *genericarmclient.CloudError) (core.CloudErrorDetails, error) {
-		log.V(Info).Info(
-			"Classifying cloud error",
-			"Message", cloudError.InnerError.Message,
-			"Code", cloudError.InnerError.Code,
-			"Target", cloudError.InnerError.Target)
+		if cloudError.InnerError != nil {
```
matthchr, line 45: Maybe just add a `String()` method to `CloudError` type and do this there?

v2/api/documentdb/customizations/sql_database_extension_types.go
```diff
@@ -0,0 +1,51 @@
...
+	if isRetryableBadRequest(cloudError.InnerError) {
+		details.Classification = core.ErrorRetryable
+	}
+
+	return details, nil
+}
+
+// isRetryableConflict checks the passed error to see if it is a retryable conflict, returning true if it is.
+func isRetryableBadRequest(err *genericarmclient.ErrorResponse) bool {
+	if err == nil || err.Message == nil {
```
matthchr, line 46: A bit weird you check `err.Message` here but then use the `Code` below? What if they sent a `Code` but no `Message`?

I think just the `Code` check below is sufficient and you could remove this `Message == nil` bit.

matthchr (approved):
Looks good but had a few small comments.

v2/pkg/genruntime/extensions/error_classifier.go
theunrepentantgeek: Good idea. Done.

v2/pkg/genruntime/extensions/error_classifier.go
theunrepentantgeek: The changes from PR #2112 simplified this.

v2/api/documentdb/customizations/sql_database_extension_types.go
theunrepentantgeek: Agreed.

matthchr (approved):

PR #2111, opened 2022-02-22
CodeGenerationContext implements ReadonlyTypes too

matthchr:
This is useful for an upcoming PR where I wanted to have a single method that could operator on a `Types` and a `CodeGenerationContext`.

Once early in the process to pick out what types to add a function to, and then in the actual implementation of that function in order to examine some of the types in more detail.

gif

### DIFF
--- a/v2/tools/generator/internal/astmodel/code_generation_context.go
+++ b/v2/tools/generator/internal/astmodel/code_generation_context.go
@@ -23,6 +23,8 @@ type CodeGenerationContext struct {
 	generatedPackages map[PackageReference]*PackageDefinition
 }
 
+var _ ReadonlyTypes = &CodeGenerationContext{}
+
 // NewCodeGenerationContext creates a new immutable code generation context
 func NewCodeGenerationContext(
 	currentPackage PackageReference,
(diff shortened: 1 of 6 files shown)


v2/tools/generator/internal/astmodel/types.go
```diff
@@ -361,25 +416,6 @@ func (set TypeDefinitionSet) ResolveResourceSpecAndStatus(resourceDef TypeDefini
 	}, nil
 }
 
-// Process applies a func to transform all members of this set of type definitions, returning a new set of type
-// definitions containing the results of the transformation, or possibly an error
-// Only definitions returned by the func will be included in the results of the function. The func may return a nil
-// TypeDefinition if it doesn't want to include anything in the output set.
-func (set TypeDefinitionSet) Process(transformation func(definition TypeDefinition) (*TypeDefinition, error)) (TypeDefinitionSet, error) {
```
matthchr, line 368: I removed this because I observed there seemed to be no usage of it. If folks think I should keep it happy to though.

v2/tools/generator/internal/astmodel/code_generation_context.go
```diff
@@ -83,34 +85,62 @@ func (codeGenContext *CodeGenerationContext) MustGetImportedPackageName(referenc
...
 		return nil, errors.Errorf("%s not imported", reference)
 	}
 	return packageDef, nil
 }
 
-// GetImportedDefinition looks up a particular type definition in a package available in this context
-func (codeGenContext *CodeGenerationContext) GetImportedDefinition(typeName TypeName) (TypeDefinition, error) {
-	pkg, err := codeGenContext.GetGeneratedPackage(typeName.PackageReference)
+// GetOrErr looks up a particular type definition in a package available in this context
+func (ctx *CodeGenerationContext) GetOrErr(name TypeName) (TypeDefinition, error) {
```
theunrepentantgeek, line 103: I strongly dislike the name `GetOrErr()` - we don't use `OrErr` as a suffix on any other methods. 

What was wrong with the previous name `GetImportedDefinition()`? If it must be renamed, consider taking inspiration from the comment: `LookupDefinition()`.

v2/tools/generator/internal/astmodel/code_generation_context.go
```diff
@@ -83,34 +85,62 @@ func (codeGenContext *CodeGenerationContext) MustGetImportedPackageName(referenc
...
 	if err != nil {
 		return TypeDefinition{}, err
 	}
 
-	return pkg.GetDefinition(typeName)
+	return pkg.GetDefinition(name)
+}
+
+// Get looks up a particular type definition in a package available in this context
+func (ctx *CodeGenerationContext) Get(name TypeName) TypeDefinition {
```
theunrepentantgeek, line 114: This one would then be renamed `MustGetImportedDefinition()` or `MustLookupDefinition()` - and probably should panic if there is an error, not ignore it.

v2/tools/generator/internal/astmodel/types.go
```diff
@@ -18,11 +18,13 @@ type TypeDefinitionSet map[TypeName]TypeDefinition
 // A restricted interface to indicate that the
 // consumer won’t modify the contained types.
 type ReadonlyTypes interface {
-	FullyResolve(t Type) (Type, error)
-	Get(t TypeName) TypeDefinition
-	TryGet(t TypeName) (TypeDefinition, bool)
+	Get(name TypeName) TypeDefinition
+	TryGet(name TypeName) (TypeDefinition, bool)
+	GetOrErr(name TypeName) (TypeDefinition, error)
```
theunrepentantgeek, line 23: FWIW, I once worked on a team that had this very useful convention:

* Find() - method looked for something that might not be there, returned item or nil
* Get() - method looked for something that must be there, returning the item or throwing

Something like that might be better than the current method names.

v2/tools/generator/internal/astmodel/types.go
theunrepentantgeek: We moved to a different idiom; happy for this to go.

theunrepentantgeek (changes requested):
Like the update - dislike some of the naming. 😉

v2/tools/generator/internal/astmodel/code_generation_context.go
matthchr: > What was wrong with the previous name `GetImportedDefinition()`?

The issue is that since this is an interface `ReadonlyTypes`, that not all collections of `ReadonlyTypes` have a notion of an "imported definition" - for example `TypeDefinitionSet` doesn't really. It has definitions but no notion of "imported vs not imported"

What about: 

`GetDefinition (TypeDefinition, error)`
`MustGetDefinition TypeDefinition`

and I just delete `TryGet` entirely since you can do the same thing with `GetDefinition` and checking the error and it has very minimal usage (2 calls).

matthchr: Did the above (as discussed over Teams)

theunrepentantgeek (approved):

PR #2112, opened 2022-02-23
Flexible error formats

theunrepentantgeek:

We have to support two different formats for the error as some services do things differently.

The [ARM spec](https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/common-api-details.md#error-response-content) says that error details should be nested inside an `error` element.

However, some services put the code & message at the top level instead.
This is common enough that the [Azure Python SDK](https://github.com/Azure/azure-sdk-for-python/blob/9791fb5bc4cb6001768e6e1fb986b8d8f8326c43/sdk/core/azure-core/azure/core/exceptions.py#L153) has specific handling to promote a nested error to the top level.

gif

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/internal/genericarmclient/cloud_error.go
+++ b/v2/internal/genericarmclient/cloud_error.go
@@ -90,6 +90,10 @@ func (e CloudError) ErrorTarget() string {
 		return *e.InnerError.Target
 	}
 
+	if e.Target != nil && *e.Target != "" {
+		return *e.Target
+	}
+
 	return ""
 }
 
--- a/v2/internal/genericarmclient/cloud_error.go
+++ b/v2/internal/genericarmclient/cloud_error.go
@@ -5,12 +5,34 @@ Licensed under the MIT license.
 
 package genericarmclient
 
-// CloudError - An error response for a resource management request.
+import (
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/core"
+)
+
+// CloudError - An error response for a resource management request
+// We have to support two different formats for the error as some services do things differently.
+//
+// The ARM spec says that error details should be nested inside an `error` element.
+// See https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/common-api-details.md#error-response-content
+//
+// However, some services put the code & message at the top level instead.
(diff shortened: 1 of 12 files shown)


v2/internal/genericarmclient/cloud_error.go
```diff
@@ -27,4 +49,54 @@ func (e CloudError) Error() string {
 	return e.error.Error()
 }
 
-func (e CloudError) Unwrap() error { return e.error }
+// ErrorCode returns the error code from the message.
+// If InnerError is present, we have an ARM compliant error (our preferred kind), so we use that code if present.
+// If InnerError is not present, we return the outer error code if present.
+// If neither is present, we return UnknownErrorCode
+func (e CloudError) ErrorCode() string {
```
matthchr, line 56: Did you consider overriding the `UnmarshalJSON` handler and deserializing both flavors into a single structure? The advantage would be:
- No duplicate fields
- No need to write accessor methods (which are fine but awkward when the fields they are exposing are public and anybody can decide to read from them, which might work most of the time but be wrong)

v2/internal/genericarmclient/cloud_error.go
```diff
@@ -5,12 +5,34 @@ Licensed under the MIT license.
...
-	// format.)
+	// READ-ONLY; The error code.
+	Code *string `json:"code,omitempty" azure:"ro"`
+
+	// READ-ONLY; The error message.
+	Message *string `json:"message,omitempty" azure:"ro"`
+
+	// READ-ONLY; The error target.
+	Target *string `json:"target,omitempty" azure:"ro"`
+
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed by flattening the struct

v2/internal/genericarmclient/cloud_error.go
theunrepentantgeek: I implemented `UnmarshalJSON()` but kept the accessor methods as they do useful things (like handling **nil** values) that make consumption of the values simpler.

matthchr (approved):

PR #2113, opened 2022-02-23
Reduce test delay times [test]

Porges:
See if this makes much difference…


Porges:
No overall improvement, abandoning.

PR #2115, opened 2022-02-24
Simplify kubernetes_resource_interface

matthchr:
Split validation and defaulting into their own methods added by a different stage.
Conceptually these two interfaces are not directly related to the `KubernetesResource` interface.


v2/tools/generator/internal/codegen/pipeline/apply_kubernetes_resource_interface.go
```diff
@@ -14,43 +14,36 @@ import (
 	"github.com/Azure/azure-service-operator/v2/tools/generator/internal/interfaces"
 )
 
+const ApplyKubernetesResourceInterfaceStageID = "applyKubernetesResourceInterface"
+
 // ApplyKubernetesResourceInterface ensures that every Resource implements the KubernetesResource interface
 func ApplyKubernetesResourceInterface(idFactory astmodel.IdentifierFactory) Stage {
 	return MakeLegacyStage(
```
theunrepentantgeek, line 21: Update this to use `MakeStage()` at the same time - we want to eventually remove `MakeLegacyStage()` completely.

v2/tools/generator/internal/codegen/pipeline/apply_kubernetes_resource_interface.go
```diff
@@ -14,43 +14,36 @@ import (
...
-				} else {
-					result.Add(typeDef)
+				// this is really very ugly; a better way?
+				if _, ok := newResource.SpecType().(astmodel.TypeName); !ok {
+					// the resource Spec was replaced with a new definition; update it
+					// by replacing the named definition:
+					specName := resource.SpecType().(astmodel.TypeName)
+					updatedDefs.Add(astmodel.MakeTypeDefinition(specName, newResource.SpecType()))
+					newResource = newResource.WithSpec(specName)
 				}
```
theunrepentantgeek, line 41: Do we need something like a `Redefine(TypeName, Type)` method on `TypeDefinitionSet` that overwrites any existing definition by that name with a new one? 

The resulting code would resemble this:

```
				// this is really very ugly; a better way?
				if newType, ok := newResource.SpecType().(astmodel.TypeName); !ok {
					// the resource Spec was replaced with a new definition; update it
					// by replacing the named definition:
					specName := resource.SpecType().(astmodel.TypeName)
					updatedDefs.Redefine(specName, newType)
					newResource = newResource.WithSpec(specName)
				}
```

theunrepentantgeek (changes requested):

v2/tools/generator/internal/codegen/pipeline/apply_kubernetes_resource_interface.go
matthchr: Done

v2/tools/generator/internal/codegen/pipeline/apply_kubernetes_resource_interface.go
matthchr: At least in this context though, the `updatedDefs` doesn't have this type defined at all so it's not clear what `Redefine` would do in that case, or how it's different than `Add`?

theunrepentantgeek: I was thinking that maybe a helper method would reduce the number of steps; if we had overloads, I'd suggest `Add(TypeName, Type)` as an overload for `Add(TypeDefinition)`.

theunrepentantgeek (approved):
Looks good.

PR #2116, opened 2022-02-25
Autogenerate CRD docs

### LINKED ISSUE #1715: Autogenerate CRD docs

super-harsh:
Closes #1715 

This PR adds support to generate API docs for ASOv2 CRDs for the end-users. API docs will help the end-users to get a better understanding of the fields and values for CRDs.

This PR includes the generated API docs, task and script to generate the API docs. Docs are generated separately for each group and version. The tool we're using to generate the docs is [gen-crd-api-reference-docs](https://github.com/ahmetb/gen-crd-api-reference-docs). 

gif

**If applicable**:


v2/tools/generator/internal/astmodel/package_definition.go
```diff
@@ -231,17 +241,40 @@ var (
...
+func emitDocFile(pkgDef *PackageDefinition, outputDir string) error {
+	docFile := filepath.Join(outputDir, "doc.go")
+	return emitTemplateFile(pkgDef, docFileTemplate, docFile)
+
+}
+
+func emitTemplateFile(pkgDef *PackageDefinition, template *template.Template, fileRef string) error {
 	buf := &bytes.Buffer{}
-	err := groupVersionFileTemplate.Execute(buf, pkgDef)
+	defer buf.Reset()
```
matthchr, line 269: What's the point of doing this given that buf goes out of scope when the function ends?

v2/tools/generator/internal/astmodel/package_definition.go
```diff
@@ -58,12 +59,21 @@ func (p *PackageDefinition) EmitDefinitions(outputDir string, generatedPackages
...
 	if resources := FindResourceDefinitions(p.definitions); len(resources) > 0 {
+
+		// If package contains resources, then we generate GroupVersion file for the package
 		err = emitGroupVersionFile(p, outputDir)
 		if err != nil {
 			return 0, err
 		}
+
+		// If package path does not include crossplane api dir and is not Storage package, then we generate doc file for the package
+		if !strings.HasSuffix(p.PackageName, StoragePackageSuffix) && !strings.Contains(outputDir, "apis") {
```
matthchr, line 72: Maybe rather than specifically checking for crossplane here, you can make a config option in `azure-arm.yaml` that sets `emitdocs` to true, and then check that here? That way crossplane's YAML can set it to false (or not set it at all and false is the default) and `azure-arm.yaml` can set it to true, and we don't need to hardcode understanding that there are two into the actual tool.

scripts/gen-api-docs.sh
```diff
@@ -0,0 +1,50 @@
...
+mkdir $OUTPUTDIR
+
+#Iterate through the directories
+for package in $(find "$APIROOT" -type d); 
+do
+    PACKAGE_VERSION=$(basename $package)
+    GROUPNAME=$(basename $(dirname $package))
+
+    #Filter the main CRD packages
+    if [[ $PACKAGE_VERSION =~ $PATTERN ]] && [[ "$PACKAGE_VERSION" != *"storage" ]]
```
matthchr, line 37: Is this actually needed now given the storage packages also do not have doc.go? (Not entirely sure what doc.go gets the tool so just checking)

docs/api/template/config.json
```diff
@@ -0,0 +1,28 @@
+{
+    "hideMemberFields": [
+        "TypeMeta"
+    ],
+    "hideTypePatterns": [
+        "ParseError$",
+        "List$"
+    ],
```
matthchr, line 8: If these were copied from gen-crd-api-reference-docs, consider sticking a README.md or something alongside these that explains what they are and where they came from (in case we want to fetch updated ones from upstream later we remember where we got them).

docs/hugo/config.toml
```diff
@@ -5,4 +5,4 @@ theme = 'book'
 
 # theme-specific settings
 [params]
-BookSection = "*"
+BookSection = "*"
```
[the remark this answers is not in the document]
super-harsh: There's no change here, I tried to add some config to work out with 'hugo plaintext HTML error' which turned out to be useless so reverted the changes.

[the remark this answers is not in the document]
matthchr: You deleted a newline, I believe (which is why this is showing up as a change at all)

docs/api/template/config.json
```diff
@@ -0,0 +1,28 @@
+{
```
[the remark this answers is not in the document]
super-harsh: This is not specific to v1 neither is the script. So thought would be a better place to put here. Can put it under `docs/v2/api..` if you suggest is the better place for templates.

scripts/gen-api-docs.sh
```diff
@@ -0,0 +1,50 @@
...
+
+APIROOT=$1
+OUTPUTDIR=$2
+PATTERN='^v[0-9]((alpha|beta)[a-z0-9]+)?$'
+
+GOPATH=/Users/$USER/go
+mkdir -p $GOPATH/bin
+GOBIN="$GOPATH/bin"
+
+#Check if 'gen-crd-api-reference-docs' tool is installed, if not, then install.
```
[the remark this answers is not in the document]
super-harsh: Yep, this sounds better. Will make the changes to put it under `.devcontainer/install-dependencies.sh`.

scripts/gen-api-docs.sh
super-harsh: This checks saves us time in generating docs, if the tool goes through the storage packages then it tries to iterate through all the elements inside the package and then come up with an error `no API packages found in ./api/batch/v1alpha1api20210101storage` which takes a few seconds.

v2/tools/generator/internal/astmodel/package_definition.go
super-harsh: It will dealloc the resources and garbage collection efficiency will be better.

.devcontainer/install-dependencies.sh
```diff
@@ -62,6 +62,9 @@ go get \
 go install -tags extended github.com/gohugoio/hugo@v0.88.1
 go install github.com/wjdp/htmltest@v0.15.0
 
+# for api docs 
+go install github.com/ahmetb/gen-crd-api-reference-docs@v0.3.1-0.20220223025230-af7c5e0048a3
```
matthchr, line 66: Minor: Missing TODO/note to update this when a new release is tagged?

scripts/gen-api-docs.sh
matthchr: Maybe leave a comment explaining that here?

v2/tools/generator/internal/astmodel/package_definition.go
```diff
@@ -50,20 +51,29 @@ func (p *PackageDefinition) AddDefinition(def TypeDefinition) {
...
+	// Check if current package contains resources
 	if resources := FindResourceDefinitions(p.definitions); len(resources) > 0 {
+
+		// If package contains resources, then we generate GroupVersion file for the package
 		err = emitGroupVersionFile(p, outputDir)
 		if err != nil {
 			return 0, err
 		}
+
+		// If package path does not include crossplane api dir and is not Storage package, then we generate doc file for the package
```
matthchr, line 71: minor: Fix comment as crossplane is not longer explicitly checked here

v2/tools/generator/internal/astmodel/package_definition.go
matthchr: Really? If you look at the [`Reset()` implementation](https://go.dev/src/bytes/buffer.go):

```
// Reset resets the buffer to be empty,
// but it retains the underlying storage for use by future writes.
// Reset is the same as Truncate(0).
func (b *Buffer) Reset() {
	b.buf = b.buf[:0]
	b.off = 0
	b.lastRead = opInvalid
}
```

You can see there's no actual space savings here: https://go.dev/play/p/zXGZkI8C3Rv. Even if this did actually deallocate space, I don't think there's anything that makes a difference between waiting til the GC comes to clean up the slice whose space you deallocated, vs waiting for the GC to come clean up the `buf` itself. AFAIK both will take the same amount of time. @Porges might know something I don't there though.

v2/tools/generator/internal/config/configuration.go
```diff
@@ -53,6 +53,8 @@ type Configuration struct {
 	// SamplesURL is the URL the samples are accessible at. Paths will be appended to the end of this to
 	// build full sample links. If this is not specified, no samples links are generated.
 	SamplesURL string `yaml:"samplesUrl"`
+	// EmitDocFiles is used as a signal to create doc.go files for packages
```
matthchr, line 56: minor: Consider saying "the default is false" (even though it's sorta obvious because this is a non-ptr boolean).

matthchr (approved):
This looks good to me. There are still some rough edges, for example the fact we generate documentation for the ARM types. It might be worth filing an item to stop doing that (we should just document the k8s types), but I think this is a good starting point so going ahead and approving.

PR #2126, opened 2022-03-02
Make kubebuilder:validation:Required work

### LINKED ISSUE #1999: kubebuilder:validation:Required is broken without json:omitEmpty
Fields with `kubebuilder:validation:Required` and without `json:omitEmpty` are "present" on deserialization, meaning they look like they've been specified. This causes the APIServer to let them through and then the request fails when going to ARM.

We either need to also automatically include a length requirement, or put `omitempty` on these fields.

matthchr:
This is required in order to make Go JSON serialization behave the way that we need it to. All types must be optional (regardless of if the property is required or not) because of how non-optional types (int, string, MyType, etc) interact with omitempty.

If a field is `json:omitempty` but its type is not optional (not a ptr) then the default value for that type will be omitted from the JSON payload. For example 0 would be omitted for ints. On the other hand if a field is NOT `json:omitempty` then the type is always serialized in the payload which causes issues for `kubebuilder:validation:Required` (how can we tell
the user didn't specify that value?)

Note: This doesn't currently make slices and maps ptrs. We may end up needing to do that as well. If they aren't ptrs (but they are omitempty) then there isn't a good way to differentiate between unspecified and empty, which may be a problem for some RP APIs. Doing this is significantly more horrible than wrapping other types in ptrs so I haven't done it for now.

Closes #1999.

**If applicable**:
- [x] this PR contains tests


matthchr:
Recommend looking at the first and last commit first, and ignoring the middle one until the end. The middle commit is just regenerating things.

v2/tools/generator/internal/astmodel/property_definition.go
```diff
@@ -283,70 +270,107 @@ func (property *PropertyDefinition) JSONName() (string, bool) {
...
-}
-
-func (property *PropertyDefinition) withJsonOmitEmpty() *PropertyDefinition {
-	return property.WithTag("json", "omitempty")
-}
-
-func (property *PropertyDefinition) withoutJsonOmitEmpty() *PropertyDefinition {
-	return property.WithoutTag("json", "omitempty")
-}
+//func (property *PropertyDefinition) hasJsonOmitEmpty() bool {
```
matthchr, line 273: Fix

v2/tools/generator/internal/astmodel/property_definition.go
```diff
@@ -283,70 +270,107 @@ func (property *PropertyDefinition) JSONName() (string, bool) {
...
 		// No change required
 		return property
 	}
 
 	result := property.copy()
 
-	// Note that this function uses isTypeOptional while MakeRequired uses property.hasOptionalType
-	// because in this direction we care if the type behaves optionally already (Map, Array included),
-	// whereas in the MakeRequired direction we only care if the type is specifically *astmodel.Optional
+	//// Note that this function uses isTypeOptional while MakeRequired uses property.hasOptionalType
```
matthchr, line 313: Fix

v2/tools/generator/internal/astmodel/property_definition.go
```diff
@@ -283,70 +270,107 @@ func (property *PropertyDefinition) JSONName() (string, bool) {
...
 		return property
 	}
 	result := property.copy()
 
-	if property.hasOptionalType() {
-		// Need to remove the optionality
-		ot := property.propertyType.(*OptionalType)
-		result.propertyType = ot.BaseType()
-	}
+	//if property.hasOptionalType() {
```
matthchr, line 292: Fix

v2/tools/generator/internal/astmodel/property_definition_test.go
```diff
@@ -271,50 +271,6 @@ func Test_PropertyDefinitionWithType_GivenSameType_ReturnsExistingReference(t *t
 	g.Expect(updated).To(BeIdenticalTo(original))
 }
 
-/*
```
matthchr, line 274: TODO: Maybe should add a test or two for my new methods.

v2/tools/generator/internal/codegen/code_generator.go
```diff
@@ -152,6 +152,7 @@ func createAllPipelineStages(idFactory astmodel.IdentifierFactory, configuration
 		pipeline.ReportOnTypesAndVersions(configuration).UsedFor(pipeline.ARMTarget), // TODO: For now only used for ARM
 
 		pipeline.CreateARMTypes(idFactory).UsedFor(pipeline.ARMTarget),
+		pipeline.FixOneOfTypes().UsedFor(pipeline.ARMTarget),
```
matthchr, line 155: Possibly improve the name of this pipeline stage.

v2/tools/generator/internal/astmodel/property_definition_test.go
matthchr: Did this.

v2/api/cache/v1alpha1api20201201/redis_types_gen.go
```diff
@@ -368,6 +368,7 @@ type RedisResource_Status struct {
 	//ShardCount: The number of shards to be created on a Premium Cluster Cache.
 	ShardCount *int `json:"shardCount,omitempty"`
 
+	// +kubebuilder:validation:Required
```
matthchr, line 371: This shouldn't be here.

v2/tools/generator/internal/codegen/pipeline/add_arm_conversion_interface.go
```diff
@@ -198,8 +198,12 @@ func (c *armConversionApplier) transformSpec(resourceType *astmodel.ResourceType
 			return t, nil
 		}
 
-		// rename Name to AzureName
-		azureNameProp := armconversion.GetAzureNameProperty(c.idFactory).WithType(nameProp.PropertyType())
+		// rename Name to AzureName and promote type if needed
+		namePropType := nameProp.PropertyType()
+		if optional, ok := namePropType.(*astmodel.OptionalType); ok {
+			namePropType = optional.Element()
+		}
```
theunrepentantgeek, line 205: Might this ever be wrapped in another type?

v2/tools/generator/internal/codegen/pipeline/add_arm_conversion_interface.go
```diff
@@ -263,5 +267,5 @@ func (c *armConversionApplier) createExtensionResourceOwnerProperty() *astmodel.
 	return astmodel.NewPropertyDefinition(
 		c.idFactory.CreatePropertyName(astmodel.OwnerProperty, astmodel.Exported),
 		c.idFactory.CreateIdentifier(astmodel.OwnerProperty, astmodel.NotExported),
-		astmodel.ArbitraryOwnerReference).MakeRequired()
+		astmodel.NewOptionalType(astmodel.ArbitraryOwnerReference)).MakeRequired()
```
theunrepentantgeek, line 270: Should "MakeRequired()" enforce the optional wrapper?

v2/tools/generator/internal/codegen/pipeline/fix_oneof_types.go
```diff
@@ -0,0 +1,101 @@
...
+
+	"github.com/Azure/azure-service-operator/v2/tools/generator/internal/astmodel"
+)
+
+// MakeOneOfDiscriminantRequiredStageId is the unique identifier for this pipeline stage
+const MakeOneOfDiscriminantRequiredStageId = "makeOneOfDiscriminantRequired"
+
+// MakeOneOfDiscriminantRequired walks the type graph and builds new types for communicating
+// with ARM
+func MakeOneOfDiscriminantRequired() Stage {
```
theunrepentantgeek, line 22: Maybe this file should be called `make_one+_of_discriminant_required.go`.

Though, actually, why can't you fix this right where the discriminant is created? Why does this need a new stage?

theunrepentantgeek (approved):

v2/tools/generator/internal/codegen/pipeline/fix_oneof_types.go
matthchr: Will doublecheck but I think the issue was identifying it as a discriminant there.

v2/internal/controllers/edge_case_test.go
```diff
@@ -188,3 +188,47 @@ func Test_CreateStorageAccountThatAlreadyExists_ReconcilesSuccessfully(t *testin
...
+func Test_CreateStorageAccountWithoutRequiredProperties_Rejected(t *testing.T) {
+	t.Parallel()
+
+	tc := globalTestContext.ForTest(t)
+
+	rg := tc.CreateTestResourceGroupAndWait()
+
+	// Create another resource group that points to the same Azure resource
+	// Custom namer because storage accounts have strict names
+	namer := tc.Namer.WithSeparator("")
```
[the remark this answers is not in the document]
matthchr: Fixed! This is what I get for rebasing my stuff 😄

v2/internal/testcommon/kube_per_test_context.go
```diff
@@ -346,10 +341,10 @@ func (tc *KubePerTestContext) CreateResourceUntracked(obj client.Object) {
 	tc.G.Expect(tc.kubeClient.Create(tc.Ctx, obj)).To(gomega.Succeed())
 }
 
-// CreateResourceExpectFailure attempts to create a resource an asserts that the resource
+// CreateResourceExpectRequestFailure attempts to create a resource an asserts that the resource
```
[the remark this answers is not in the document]
matthchr: Fixed

v2/tools/generator/internal/codegen/pipeline/fix_oneof_types.go
matthchr: Yeah, so the issue is that we get a type like this:
```
    "NumberInAdvancedFilter": {
      "type": "object",
      "properties": {
        "operatorType": {
          "type": "string",
          "enum": [
            "NumberIn"
          ]
        },
        "values": {
          "oneOf": [
            {
              "type": "array",
              "items": {
                "type": "number"
              }
            },
            {
              "$ref": "https://schema.management.azure.com/schemas/common/definitions.json#/definitions/expression"
            }
          ],
          "description": "The set of filter values."
        }
      },
      "required": [
        "operatorType"
      ],
      "description": "NumberIn Advanced Filter."
    },
```

There's no field that says "This is the discriminant". The way to determine discriminant is by looking at all of the types and finding the one property they have in common (which in each case should be an enum with a single allowed value).
If we had a type that had 2 enums that satisfied those criteria we could still find the discriminant provided we have other types to compare it to, but at the point the type is being created we dont have access to those other types, so it needs to be done after the fact.

I _think_ in the Swagger there's a field that identifies the discriminant so we could maybe change this when @porges stuff lands.

v2/tools/generator/internal/codegen/pipeline/add_arm_conversion_interface.go
matthchr: At least currently, no.
If it were, we would need to make this modification with a visitor to maintain the wrapping. I've updated the code to leave a note to this effect but left it as is for now.

v2/tools/generator/internal/astmodel/optional_type.go
```diff
@@ -48,16 +48,33 @@ func NewOptionalType(element Type) Type {
...
+	switch typ := t.(type) {
 	case *ArrayType:
 		return true
 	case *MapType:
 		return true
 	case *OptionalType:
 		return true
+	case *ValidatedType:
+		return isTypeOptional(typ.ElementType())
 	default:
```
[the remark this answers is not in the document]
matthchr: Good idea. I've changed the `ValidatedType` handling to instead be `MetaType` handling (since `ValidatedType` is a `MetaType`)

v2/tools/generator/internal/codegen/pipeline/fix_oneof_types.go
matthchr: Renamed this file. Good catch.

v2/tools/generator/internal/astmodel/optional_type.go
```diff
@@ -48,16 +48,33 @@ func NewOptionalType(element Type) Type {
...
+	switch typ := t.(type) {
+	case *ArrayType:
+		return false
+	case *MapType:
+		return false
+	case *OptionalType:
+		return true
+	case *ValidatedType:
+		return canTypeBeMadeRequired(typ.ElementType())
+	default:
```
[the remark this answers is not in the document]
matthchr: Fixed this.

v2/tools/generator/internal/codegen/pipeline/add_arm_conversion_interface.go
matthchr: I don't want to have `MakeRequired` return an `err`, but agree this is a good idea. I think it makes sense to make `MakeRequired` panic in this case, as really the problem is a programming error (you have asked the property to move into a nonsensical state).

Thoughts?

theunrepentantgeek: Oh, I was thinking that `MakeRequired()` could proactively wrap the property type if it's not already optional, so the call site would simplify:

```suggestion
		astmodel.ArbitraryOwnerReference).MakeRequired()
```

matthchr: Hmm, we could but it feels a bit strange to have `MakeRequired` actually do `MakeTypeOptional` internally? It would also make `MakeRequired` and `MakeOptional` non-symmetric as one would internally do `MakeTypeOptional` but I don't think that `MakeOptional` could do `MakeTypeRequired`, which feels weird?

theunrepentantgeek: Fair enough.

PR #2127, opened 2022-03-02
Fix up stage pre/post-requisite enforcement

theunrepentantgeek:

Discovered that many of the pre- and post-requisites declared on our pipeline stages were being dropped and thus not enforced, which is highly likely to result in problems as our pipeline is modified.

* Changes `pipeline.Stage` to be used as a reference type instead of a value type
* Renames factory methods to `NewStage()` and `NewLegacyStage()`
* Fixes pipeline creation to create stages in the correct order
* Adds enforcement so that `NewLegacyStage()` isn't used for new stages
* Added `V(3)` logging of requisite checks

gif

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/tools/generator/internal/codegen/golden_files_test.go
+++ b/v2/tools/generator/internal/codegen/golden_files_test.go
@@ -201,10 +201,11 @@ func NewTestCodeGenerator(
 		pipeline.ApplyExportFiltersStageID, // Don't want any filtering of resources during tests
 	)
 
+	//TODO: Enable this check once we fix up the test pipeline so that all the dependencies pass
 	//if err := codegen.verifyPipeline(); err != nil {
 	//	return nil, err
 	//}
-	//
+
 	return codegen, nil
 }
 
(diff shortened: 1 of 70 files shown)


Porges (approved):

v2/tools/generator/internal/codegen/golden_files_test.go
```diff
@@ -196,19 +201,23 @@ func NewTestCodeGenerator(testName string, path string, t *testing.T, testConfig
 		pipeline.ApplyExportFiltersStageID, // Don't want any filtering of resources during tests
 	)
 
+	//if err := codegen.verifyPipeline(); err != nil {
```
matthchr, line 204: fix?

v2/tools/generator/internal/codegen/pipeline/stage.go
```diff
@@ -65,13 +74,54 @@ func MakeLegacyStage(
...
+	"nameTypes",
+	"pluralizeNames",
+	"propertyRewrites",
+	"removeAliases",
+	"removeEmbeddedResources",
+	"replaceAnyTypeWithJSON",
+	"reportTypesAndVersions",
+	"rogueCheck",
+	"simplifyDefinitions",
+	"stripUnreferenced",
```
matthchr, line 113: is there a reason this is here 4 times?

matthchr (approved):

PR #2129, opened 2022-03-02
Remove sed -i ''

matthchr:
This is a mac-ism which doesn't work on Linux. Changing to just sed -i
(no subsequent quotes) so that the script can run in the .devcontainer.


theunrepentantgeek (approved):

matthchr:
Abandoning this in favor of #2132

PR #2131, opened 2022-03-03
Add extension wrapper for Secrets

matthchr:
Makes calling a resource with the extension a bit cleaner

### DIFF
--- a/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
+++ b/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
@@ -383,15 +383,13 @@ func (r *azureDeploymentReconcilerInstance) handlePollerSuccess(ctx context.Cont
 		return ctrl.Result{}, errors.Wrapf(err, "error updating status")
 	}
 
-	if retrieveSecrets, ok := r.Extension.(extensions.SecretsRetriever); ok {
-		err = r.writeSecrets(ctx, retrieveSecrets)
-		if err != nil {
-			if _, ok := secrets.AsSecretNotOwnedError(err); ok {
-				err = conditions.NewReadyConditionImpactingError(err, conditions.ConditionSeverityError, conditions.ReasonSecretWriteFailure)
-			}
-
-			return ctrl.Result{}, err
+	err = r.saveAzureSecrets(ctx)
+	if err != nil {
+		if _, ok := secrets.AsSecretNotOwnedError(err); ok {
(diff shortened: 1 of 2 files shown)


v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -632,8 +630,9 @@ func (r *azureDeploymentReconcilerInstance) deleteResourceSucceeded(ctx context.
 	return nil
 }
 
-func (r *azureDeploymentReconcilerInstance) writeSecrets(ctx context.Context, extension extensions.SecretsRetriever) error {
-	secretSlice, err := extension.RetrieveSecrets(ctx, r.Obj, r.ARMClient, r.Log)
+func (r *azureDeploymentReconcilerInstance) retrieveAndWriteSecretsIfNeeded(ctx context.Context) error {
```
theunrepentantgeek, line 633: Suggest just calling this `RetrieveSecrets()` - cause it doesn't write them (that's below), and because methods with *And* in the name are worrisome.

theunrepentantgeek (approved):

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
matthchr: `results, err := secrets.ApplySecretsAndEnsureOwner(ctx, r.KubeClient.Client, r.Obj, secretSlice)` is writing the secrets, and this bit is retrieving them:
```
	retriever := extensions.CreateSecretRetriever(ctx, r.Extension, r.ARMClient, r.Log)
	secretSlice, err := retriever(r.Obj)
```

So this method does do both (at least indirectly). IIRC you were the one who I talked with about method taxonomy and this is a coordination method: It stitches 2 things together. If we want to remove the "and" in the name IMO it's more correct to call it `writeSecretsIfNeeded`. I would personally expect a `RetrieveSecrets` method to return a collection of secrets (which this doesn't do). The end result of this method being called is that the secrets are written (or an error is returned).

Thoughts?

theunrepentantgeek: As agreed offline, change this to `saveAzureSecrets`.

matthchr: We decided on `saveAzureSecrets`. I also added a little comment explaining that this method is a noop if there are no secrets to save.

PR #2132, opened 2022-03-03
Add support for generating ASOv2 helm chart packages and fix helm manifest files

### LINKED ISSUE #2122: Bug: No HTTP resource was found that matches the request URI
mcr.microsoft.com/k8s/azureserviceoperator:v2.0.0-alpha.5

**Describe the bug**
We receive an error when following the v2 README to create the azure sample resource group.

azure_generic_arm_reconciler.go:518] controllers/ResourceGroupController "msg"="Resource creation failure" "azureName"="" "name"="asosamplerg" "namespace"="default" "error"="{\"message\":\"No HTTP resource was found that matches the request URI 'https://management.azure.com/subscriptions/XXXXXX/resourceGroups/?api-version=2020-06-01'.\"}" "resourceID"="/subscriptions/XXXXXX/resourceGroups/"

**To Reproduce**
Create a sample resource group as described in the V2 README

**Expected behavior**
We expect the resource group to be created without error.

**Additional context**
We have verified that the service principle being used by the operator has access to the subscription where the resource group is supposed to be created.

### LINKED ISSUE #2128: Hide the stuff used to produce Helm charts
the CRDs and other things are visible in the v2 charts repo. According to the [Helm documentation](https://helm.sh/docs/topics/chart_repository/) we really just need the .tar.gz's.

There's some stuff (like the handcrafted values.yaml) we ought to keep here but wondering if we should .gitignore the rest so that people don't go trying to install the "chart" at `master` (see #2122).

super-harsh:
Closes #2128 

- Add index.yaml for helm chart
- Add azure-service-operator-v2.0.0-alpha.6.tgz helm package
- Update generate-helm-manifest.sh script to generate index file and helm package
- Remove helm generated manifest files 
- Add helm manifest dirs into .gitignore


scripts/generate-helm-manifest.sh
```diff
@@ -11,17 +11,18 @@ VERSION=$4
...
-sed -i '' "s@$LOCAL_REGISTRY_CONTROLLER_DOCKER_IMAGE@{{.Values.image.repository}}@g" "$DIR"charts/azure-service-operator/templates/generated/*_deployment_* # Replace hardcoded ASO image
+sed -i "s@$LOCAL_REGISTRY_CONTROLLER_DOCKER_IMAGE@{{.Values.image.repository}}@g" "$DIR"charts/azure-service-operator/templates/generated/*_deployment_* # Replace hardcoded ASO image
 grep -E $KUBE_RBAC_PROXY "$DIR"charts/azure-service-operator/templates/generated/*_deployment_* # Ensure that what we're about to try to replace actually exists (if it doesn't we want to fail)
-sed -i '' "s@$KUBE_RBAC_PROXY.*@{{.Values.image.kubeRBACProxy}}@g" "$DIR"charts/azure-service-operator/templates/generated/*_deployment_*
-sed -i '' "s@cert-manager.io/.*@{{.Values.certManagerResourcesAPIVersion}}@g" "$DIR"charts/azure-service-operator/templates/generated/*cert-manager.io*
+sed -i "s@$KUBE_RBAC_PROXY.*@{{.Values.image.kubeRBACProxy}}@g" "$DIR"charts/azure-service-operator/templates/generated/*_deployment_*
 find "$DIR"charts/azure-service-operator/templates/generated/ -type f -exec sed -i '' "s@azureserviceoperator-system@{{.Release.Namespace}}@g" {} \;
```
matthchr, line 25: I think you missed a `sed -i ''` here

v2/.gitignore
```diff
@@ -4,3 +4,7 @@ config/crd/out
 config/crd/kustomization.yaml
 config/webhook/manifests.yaml
 config/rbac/role.yaml
+
+# Helm chart manifest
+charts/azure-service-operator/templates/generated
```
matthchr, line 9: minor: You can nest `.gitignore` so you could put this closer to the actual files being ignored (like in `charts/azure-service-operator`?). Is the reason to avoid that that then that file would get included in the actual helm chart? If so this makes sense and is fine, but if that's not an issue then might wanna put this closer to what it's ignoring?

matthchr (approved):

scripts/generate-helm-manifest.sh
super-harsh: Was intended to remove it anyway.

PR #2139, opened 2022-03-06
Refactoring and cleanup in preparation for changing namespaces

theunrepentantgeek:

Refactoring and cleanup in preparation for changing our namespace prefixes from from `v1alpha1api` to `v1beta`.

To avoid having to mess around with string manipulation when creating compatibility types (coming in a future PR), changed `LocalPackageReference` to explicitly receive the prefix as a separate parameter.

* Updated the config file to use versions that aren't prefix specific
* Updated most tests to use the prefix `v` so they won't change when the production prefix is changed
* Updated `StoragePackageReference` to more consistently defer to the wrapped `LocalPackageReference` 
* No longer need to special case `containsPreviewVersionLabel()` to skip our fixed prefix 
* Consistently use `test.MakeLocalPackageReference()` from tests 

Conversion graph generation was prevously driven by the API types - but we don't have those available for generation of compatibility types. 

* generate the conversion graph from the storage types (since each `StoragePackageReference` contains within it the linked LocalPackagReference`). 

* Remove dependency of `CreateStorageTypes` on the conversion graph

Moving `CreateStorageTypes` ahead of `CreateConversionGraph` in the pipeline unearthed a number of problems with our pipeline and it's testing. Verification of pre- and post-requisities wasn't being done for tests, resulting in some tests having incorrect object structures

* Moved verification of pre- and post-requisites into `pipeline.Stage`, including tests.
* Moved tracking of seen/expected stages into `pipeline.State`.
* Fix code generation panic caused by presence of types from multiple packages when creating golden test file output 
* Fix dependencies of pipeline stages during testing.
* Use `RunTestPipeline()` more consistently when testing pipeline stages

Code gardening and diagnostics

* For both ResourceType and ObjectType, added context to any panic that occurs during code generation of a **func** so we know which one failed
* Strip functions and interface implementations from golden file output so the data structures are more easily seen

gif

**If applicable**:
- [x] this PR contains tests


v2/tools/generator/internal/astmodel/resource_type.go
```diff
@@ -596,13 +605,31 @@ func (resource *ResourceType) generateMethodDecls(codeGenerationContext *CodeGen
...
 	for _, f := range resource.Functions() {
-		funcDef := f.AsFunc(codeGenerationContext, typeName)
+		funcDef := resource.generateMethodDeclForFunction(codeGenerationContext, typeName, f)
 		result = append(result, funcDef)
 	}
 
 	return result
 }
 
+func (resource *ResourceType) generateMethodDeclForFunction(
```
super-harsh, line 615: Since we're not using any values from the receiver type, can consider putting this method somewhere in common to be re-used by ObjectType as well.

v2/tools/generator/internal/codegen/pipeline/create_resource_extension_types.go
```diff
@@ -53,10 +57,5 @@ func CreateResourceExtensions(localPath string, idFactory astmodel.IdentifierFac
 			return state, nil
 		})
 
-	// We don't want tests to be generated for resourceExtensions, since these are not the actual resource types.
-	// Therefore, we want to make sure that 'createResourceExtensions' stage only runs when these prerequisite
-	// stages have completed.
-	stage.RequiresPrerequisiteStages(InjectJsonSerializationTestsID, InjectPropertyAssignmentTestsID)
```
super-harsh, line 59: Don't we need this anymore?

v2/tools/generator/internal/codegen/pipeline/state.go
```diff
@@ -51,10 +57,31 @@ func (s *State) WithConversionGraph(graph *storage.ConversionGraph) *State {
...
-		definitions:     s.definitions.Copy(),
-		conversionGraph: graph,
+	result := s.copy()
+	result.conversionGraph = graph
+	return result
+}
+
+// WithSeenStage records that the passed stage has been seen
+func (s *State) WithSeenStage(id string) *State {
+	result := s.copy()
```
super-harsh, line 67: Do we need copy of the whole struct here? We can only get `stagesSeen` and `stagesExpected` into vars independently and do the manipulation?

v2/tools/generator/internal/codegen/pipeline/state.go
```diff
@@ -51,10 +57,31 @@ func (s *State) WithConversionGraph(graph *storage.ConversionGraph) *State {
...
+	result.stagesSeen.Add(id)         // Record that we saw this stage
+	delete(result.stagesExpected, id) // Discard expectations as they are satisfied
+	return result
+}
+
+// WithExpectation records our expectation that the later stage is coming
+func (s *State) WithExpectation(earlierStage string, laterStage string) *State {
+	result := s.copy()
```
super-harsh, line 75: Same as above!

v2/tools/generator/internal/codegen/pipeline/create_resource_extension_types.go
theunrepentantgeek: We do need the run-order dependency - I moved it to the other end, using `RequiresPostrequisiteStages()`; this allows us to test this stage _without_ having the test injection stages present.

v2/tools/generator/internal/codegen/pipeline/state.go
theunrepentantgeek: We used to have `stagesSeen` and `stagesExpected` in local variables; this is why we ended up with tests that weren't configured properly. Moving them into `State` guarantees they're carried forward correctly.

We return a copy because `State` is (effectively) an immutable type that should never be changed after instantiation; this allows (for example) comparision of pre- and post- stage values to see what changed.

v2/tools/generator/internal/functions/testdata/TestGolden_GetExtendedResourceFunction_moreThanOneVersion_GeneratesExpectedCode/GetExtendedResources.golden
```diff
@@ -1,7 +1,7 @@
 // Code generated by azure-service-operator-codegen. DO NOT EDIT.
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
-package customizations
+package vcustomizations
```
super-harsh, line 4: That doesn't look ideal - customizations should not get `v` prefixed.

v2/tools/generator/internal/functions/testdata/TestGolden_GetExtendedResourceFunction_oneVersion_GeneratesExpectedCode/GetExtendedResources.golden
```diff
@@ -1,7 +1,7 @@
 // Code generated by azure-service-operator-codegen. DO NOT EDIT.
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
-package customizations
+package vcustomizations
```
super-harsh, line 4: same as above!

v2/tools/generator/internal/config/group_configuration.go
```diff
@@ -43,15 +43,20 @@ func NewGroupConfiguration(name string) *GroupConfiguration {
 // of the version, so we can do lookups via TypeName. All indexing is lower-case to allow case-insensitive lookups (this
 // makes our configuration more forgiving).
 func (gc *GroupConfiguration) add(version *VersionConfiguration) {
-	pkg := astmodel.CreateLocalPackageNameFromVersion(version.name)
+	// Convert version.name into a package version
+	// We do this by constructing a local package reference because this avoids replicating the logic here and risking
+	// inconsistency if things are changed in the future.
+	local := astmodel.MakeLocalPackageReference("prefix", "group", astmodel.GeneratorVersionPrefix, version.name)
+	_, lv, _ := local.GroupVersion()
```
matthchr, line 50: Should we panic rather than discard here? And below?

v2/tools/generator/internal/astmodel/resource_type.go
theunrepentantgeek: Done.

v2/tools/generator/internal/codegen/pipeline/state.go
theunrepentantgeek: As discussed above - `State` is an immutable object, we always want to get a new copy.

v2/tools/generator/internal/functions/testdata/TestGolden_GetExtendedResourceFunction_moreThanOneVersion_GeneratesExpectedCode/GetExtendedResources.golden
theunrepentantgeek: Fixed.

v2/tools/generator/internal/functions/testdata/TestGolden_GetExtendedResourceFunction_oneVersion_GeneratesExpectedCode/GetExtendedResources.golden
theunrepentantgeek: Fixed.

v2/tools/generator/internal/astmodel/resource_type.go
```diff
@@ -596,13 +605,31 @@ func (resource *ResourceType) generateMethodDecls(codeGenerationContext *CodeGen
...
+	codeGenerationContext *CodeGenerationContext,
+	typeName TypeName,
+	f Function) *dst.FuncDecl {
+
+	defer func() {
```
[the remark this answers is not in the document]
theunrepentantgeek: I factored this out into a common routine (it's in `resource_type.go`) with a good comment.

v2/tools/generator/internal/astmodel/type_visitor_builder.go
```diff
@@ -14,7 +14,11 @@ import "fmt"
 // func(it TypeName) (Type, error)
 // func(it TypeName) Type
 //
-// These examples assume TypeName but can be generalized across any of the supported types.
+// These examples assume TypeName but can be generalized across any of the supported types, e.g. ResourceType:
+//
+// func(this *TypeVisitor, it *ResourceType, ctx interface{}) (Type, error)
+// func(it *ResourceType) (Type, error)
+// func(it *ResourceType) Type
```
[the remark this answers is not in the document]
theunrepentantgeek: I got confused by the comment and used the wrong kind of signature; will try another way to clarify.

v2/tools/generator/internal/astmodel/types.go
```diff
@@ -293,6 +293,16 @@ func (set TypeDefinitionSet) ResolveResourceStatusDefinition(resourceType *Resou
 	return ResolveResourceStatusDefinition(set, resourceType)
 }
 
+// AsSlice creates a new slice containing all the definitions
+func (set TypeDefinitionSet) AsSlice() []TypeDefinition {
```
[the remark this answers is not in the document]
theunrepentantgeek: Not in this usage - it doesn't care. We normally use `AsSortedSlice()` if we guarantee return order, so I think we're find.

v2/tools/generator/internal/codegen/code_generator.go
```diff
@@ -232,6 +226,11 @@ func (generator *CodeGenerator) Generate(ctx context.Context) error {
 			return errors.Wrapf(err, "failed during pipeline stage %d/%d: %s", i+1, len(generator.pipeline), stage.Description())
 		}
 
+		// Fail fast if something goes awry
+		if len(stateOut.Definitions()) == 0 {
+			return errors.Errorf("All type definitions removed by stage %s", stage.Id())
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

v2/tools/generator/internal/codegen/golden_files_test.go
```diff
@@ -257,43 +259,51 @@ func exportPackagesTestPipelineStage(t *testing.T, testName string) *pipeline.St
...
 			for _, def := range state.Definitions() {
-				ds = append(ds, def)
 				ref := def.Name().PackageReference
-				if !found {
-					pr = ref
-					group, version, found = ref.GroupVersion()
+				pkg, ok := pkgs[ref]
+				if !ok {
+					// expected to always be ok because we don't use external package references for type definitions
+					g, v, _ := ref.GroupVersion()
```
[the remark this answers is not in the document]
theunrepentantgeek: Good idea.

v2/tools/generator/internal/codegen/golden_files_test.go
```diff
@@ -318,6 +325,36 @@ func stripUnusedTypesPipelineStage() *pipeline.Stage {
 		})
 }
 
+// stripFunctionsPipelineStage creates a stage to remove functions and interface implementations (which contain
+// functions) from the golden test output.
+// These golden tests are looking at the structure of the generated structs; including all the code for functions in
+// the output is just noise that makes the golden files hard to review.
+func stripFunctionsPipelineStage() *pipeline.Stage {
```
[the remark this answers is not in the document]
theunrepentantgeek: We can't just pull those stages.

For one, they're often prerequisites of other stages that we do want to test.

For another, a lot of our stages are ONLY tested by these golden tests (they don't have isolated unit tests of their own). The more our test pipeline resembles the actual one we use for code generation the better, because we rule out unexpected interactions between the stages.

As discussed offline, I'll remove the filtering for now, and have logged #2150 so we remember to circle back and make things better.

v2/tools/generator/internal/config/group_configuration.go
theunrepentantgeek: Don't need to - the only reason it would fail is if it was an external package reference; and we just created it, so we know that's not the case.

theunrepentantgeek: Added, though I don't see how it would ever fire.

v2/tools/generator/internal/astmodel/type_visitor_builder.go
```diff
@@ -10,11 +10,18 @@ import "fmt"
 // TypeVisitorBuilder provides a flexible way to create a TypeVisitor. Fields should be initialized
 // with funcs matching one of the following forms:
 //
-// func(this *TypeVisitor, it TypeName, ctx interface{}) (Type, error)
-// func(it TypeName) (Type, error)
-// func(it TypeName) Type
+// func(this *TypeVisitor, it <sometype>, ctx interface{}) (Type, error)
```
babbageclunk, line 13: Ha, I had a comment suggesting this!

babbageclunk (approved):
Looks good - slight complaining about the loss of fail-fast checking notwithstanding! But I definitely take your point about the benefit of having the check on all pipelines, rather than just the production one.

v2/tools/generator/internal/astmodel/object_type.go
```diff
@@ -81,13 +82,31 @@ func (objectType *ObjectType) generateMethodDecls(codeGenerationContext *CodeGen
...
 
+func (objectType *ObjectType) generateMethodDeclForFunction(
+	codeGenerationContext *CodeGenerationContext,
+	typeName TypeName,
+	f Function) *dst.FuncDecl {
+
+	defer func() {
+		if err := recover(); err != nil {
+			panic(fmt.Sprintf(
+				"generating method declaration for %s.%s: %s",
```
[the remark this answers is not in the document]
theunrepentantgeek: You're not wrong. 😁

v2/tools/generator/internal/codegen/pipeline/stage.go
```diff
@@ -181,22 +184,61 @@ func (stage *Stage) Description() string {
 
 // Run is used to execute the action associated with this stage
 func (stage *Stage) Run(ctx context.Context, state *State) (*State, error) {
-	return stage.action(ctx, state)
+	if err := stage.checkPreconditions(state); err != nil {
```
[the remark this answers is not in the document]
theunrepentantgeek: As discussed, there's a minor downside to having pre/post-requisites checkec during execution instead of up front, but I think that's well mitigated by having them always checked.

v2/tools/generator/internal/config/group_configuration.go
matthchr: Yeah agreed, neither do I. It pays to be paranoid sometimes though - who knows in the future when "That cannot possibly happen" can actually start happening because of some other refactor.

Thanks for adding!

v2/tools/generator/internal/codegen/pipeline/stage.go
```diff
@@ -131,6 +133,7 @@ func (stage *Stage) RequiresPrerequisiteStages(prerequisites ...string) {
 }
 
 // RequiresPostrequisiteStages declares which stages must be executed after this one has completed
+// Use postrequisites when it is necessary for that later stage to act on the results of this one.
 // This is not completely isomorphic with RequiresPrerequisiteStages as there may be supporting stages that are
 // sometimes omitted from execution when targeting different outcomes. Having both pre- and post-requisites allows the
```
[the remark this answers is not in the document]
theunrepentantgeek: Added.

v2/tools/generator/internal/codegen/pipeline/stage.go
```diff
@@ -181,22 +184,61 @@ func (stage *Stage) Description() string {
...
+	}
+
+	return kerrors.NewAggregate(errs)
+}
+
+// checkPostrequisites returns an error if the postrequisites of this stage have been satisfied early
+func (stage *Stage) checkPostrequisites(state *State) error {
+	var errs []error
+	for _, postreq := range stage.postrequisites {
+		early := state.stagesSeen.Contains(postreq)
```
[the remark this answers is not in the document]
theunrepentantgeek: Discussed offline; we check `stagesExpected` at the end of the pipeline, to check there are no further stages pending.

hack/crossplane/azure-crossplane.yaml
```diff
@@ -133,7 +133,7 @@ typeTransformers:
     property: Copy
     ifType:
       group: deploymenttemplate
-      version: v1alpha1api20190401
+      version: "2019-04-01"
```
matthchr, line 136: minor: Can this just be `2019-04-01`? Does it need the quotes?

v2/tools/generator/internal/astmodel/type_visitor_builder.go
```diff
@@ -10,11 +10,18 @@ import "fmt"
...
+// func(it <sometype>) (Type, error)
+// func(it <sometype>) Type
 //
-// These examples assume TypeName but can be generalized across any of the supported types.
+// o  Must always return Type, and optionally an error
+// o  <sometype> must match the type for the field being initialized
+//
+// Some examples:
+//
+// VisitTypeName = func(it TypeName) Type      // Works
```
matthchr, line 22: minor: Maybe include 1 more `// Works` example too at the bottom? Possibly that's one of the larger format ones like 
`func(it *ObjectType) (Type, error)` or `func(this *TypeVisitor, it *ObjectType, ctx interface{}) (Type, error)`. I think that helps accomplished what you were going for before of showing examples for `TypeName` _and_ other fields. One that is a ptr (since `TypeName` isn't) probably?

v2/tools/generator/internal/codegen/pipeline/stage.go
```diff
@@ -131,9 +133,18 @@ func (stage *Stage) RequiresPrerequisiteStages(prerequisites ...string) {
 }
 
 // RequiresPostrequisiteStages declares which stages must be executed after this one has completed
-// This is not completely isomorphic with RequiresPrerequisiteStages as there may be supporting stages that are
+// Use postrequisites when it is necessary for that later stage to act on the results of this one.
+//
+// For example, InjectJsonSerializationTests creates round trip serialization tests for any object types that have
+// properties. It's not correct to give InjectJsonSerializationTests a prerequisite on every earlier stage that creates
+// new object types becauses it isn't concerned with where those object came from. However, but those earlier stages DO
```
matthchr, line 140: Minor/wording: `However, but` on this line is phrased strangely? maybe just:
```suggestion
// new object types becauses it isn't concerned with where those object came from. However, those earlier stages DO
```

v2/tools/generator/internal/config/group_configuration.go
matthchr: Reactivating as I don't think you actually did this at this location? (You did it below though I see)

matthchr (approved):

hack/crossplane/azure-crossplane.yaml
theunrepentantgeek: Needs the quotes, else it seems to get interpreted as a date.

v2/tools/generator/internal/astmodel/type_visitor_builder.go
theunrepentantgeek: Added.

v2/tools/generator/internal/codegen/pipeline/stage.go
theunrepentantgeek: Updated.

PR #2148, opened 2022-03-08
Actually remove all status validations

matthchr:


v2/tools/generator/internal/codegen/pipeline/remove_status_property_validations.go
```diff
@@ -114,3 +115,12 @@ func removeValidatedType(this *astmodel.TypeVisitor, vt *astmodel.ValidatedType,
 func removeEnumValidations(this *astmodel.TypeVisitor, et *astmodel.EnumType, _ interface{}) (astmodel.Type, error) {
 	return et.WithoutValidation(), nil
 }
+
+// removeKubebuilderRequired removes kubebuilder:validation:Required from all properties
+func removeKubebuilderRequired(this *astmodel.TypeVisitor, ot *astmodel.ObjectType, ctx interface{}) (astmodel.Type, error) {
```
theunrepentantgeek, line 120: Isn't this going to remove it from ALL `ObjectType` instances - don't you only want to do this for Status types?

theunrepentantgeek (changes requested):
Mostly LGTM, but I have a question about blast radius.

v2/tools/generator/internal/codegen/pipeline/remove_status_property_validations.go
matthchr: If you look up in this file on line 57, we only run this on status types.

theunrepentantgeek: All good then. 😁 #approved

theunrepentantgeek (approved):

PR #2155, opened 2022-03-13
Prepositioning before prefix change

theunrepentantgeek:

A couple of final changes necessary to make the change of prefix from `v1alpha1api` to `v1beta` as smooth as possible.

Creates a new stage to inject storage type clones using the older `v1alpha1api` prefix, enabling users of the older ASO to transparently upgrade. This stage is currently disabled.

Creation of the Conversion Graph now accepts the current generator prefix as a parameter. The global constant as our tests use the hard coded prefix `v` instead.

Updated the skipping properties detector to only fire if the properties have different types. Properties of the same exact type can deserialise from the property bag just fine. The introduction of compatibility storage types triggered the detector in the `network` and `compute` groups.

When only preview ARM versions are being generated, creation of the conversion graph was incorrectly using the earliest compatibility storage version as the hub version, because of the existing rule that preview versions always convert back to the prior version. Rebuilt the conversion graph builder to use a declarative approach, ensuring the compatibilty types always convert forward.

Following PR will change the prefix and regenerate all the things.

**Prerequisites**

- [ ] Merge PR #2177

gif

**If applicable**:
- [x] this PR contains tests


v2/tools/generator/internal/codegen/code_generator.go
```diff
@@ -177,7 +177,9 @@ func createAllPipelineStages(idFactory astmodel.IdentifierFactory, configuration
 		// TODO: For now only used for ARM
 		pipeline.InjectOriginalVersionFunction(idFactory).UsedFor(pipeline.ARMTarget),
 		pipeline.CreateStorageTypes().UsedFor(pipeline.ARMTarget),
-		pipeline.CreateConversionGraph(configuration).UsedFor(pipeline.ARMTarget),
+		// Enable this in the next PR
```
matthchr, line 180: super minor:
```suggestion
		// TODO: Enable this in the next PR
```

v2/tools/generator/internal/codegen/pipeline/create_backwards_compatibility_types.go
```diff
@@ -0,0 +1,101 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/pkg/errors"
+
+	"github.com/Azure/azure-service-operator/v2/tools/generator/internal/astmodel"
+)
+
+const CreateBackwardCompatibilityTypesStageID = "createStorageTypes"
```
matthchr, line 17: minor: Wrong stage ID?

matthchr:
```suggestion
const CreateBackwardCompatibilityTypesStageID = "createBackwardCompabilityTypes"
```

matthchr: also minor: Wondering about the name of this. Would it be more grammatically correct to say `createBackwardsCompatibleStorageTypes` or something?

The "create backwards compatibility" for some reason feels like it should read "create backwards compatible" to me

v2/tools/generator/internal/codegen/pipeline/create_backwards_compatibility_types.go
```diff
@@ -0,0 +1,101 @@
...
+
+// CreateBackwardCompatibilityTypes returns a pipeline stage that creates copies of storage types in other packages to
+// provide backward compatibility with previous releases of Azure Service Operator.
+// Backward compatibility versions are created for all storage versions to allow users of older versions of the operator
+// to easily upgrade. We lean on the fact that storage versions have already been modified to increase compatibility -
+// everything is optional, all enums have been rendered into their underlying primitive types, and so on.
+func CreateBackwardCompatibilityTypes(prefix string) *Stage {
```
matthchr, line 24: This is only creating the storage types. Do we also need to create the non-storage types too? If we want to allow people to continue submitting v1alpha1 types I think we would, although maybe we don't need that...

There's also the reality that they have the CRDs in their cluster already (since they installed them previously), and AFAIK they won't get removed by anything, so it might be ok for us to omit them?

If they did provision a new cluster and install `beta0` on it, and then attempt to port over their `v1alpha1` YAML, it would be rejected without the corresponding API types, right? Is that the desired behavior or no? (I can see arguments for both ways).

v2/tools/generator/internal/codegen/pipeline/detect_skipping_properties_test.go
```diff
@@ -138,8 +138,40 @@ func Test_DetectSkippingProperties(t *testing.T) {
...
+	_, err = RunTestPipeline(
+		initialState,
+		DetectSkippingProperties(), // and then we get to run the stage we're testing
+	)
+
+	g.Expect(err).To(Succeed())
+}
+
+// Test_DetectSkippingProperties checks that the pipeline stage correctly detects when properties of different types skip versions.
+func Test_DetectSkippingProperties_WhenPropertyTypesDiffer_ReturnsError(t *testing.T) {
```
matthchr, line 155: Nice

v2/tools/generator/internal/codegen/storage/group_conversion_graph_builder.go
```diff
@@ -33,57 +41,109 @@ func (b *GroupConversionGraphBuilder) Add(refs ...astmodel.PackageReference) {
...
-	var gaRelease astmodel.PackageReference
-	for index, ref := range storageReferences {
-		if index == 0 {
-			// Always treat the very earliest version as GA
-			gaRelease = storageReferences[0]
+// nonPreviewReferencesConvertForward links each version with the immediately following version.
+// By the time we run this stage, we should only have non-preview (aka GA) releases left
+func (b *GroupConversionGraphBuilder) nonPreviewReferencesConvertForward(refs []astmodel.PackageReference) {
+	for i, ref := range refs {
+		if i == 0 {
```
matthchr, line 113: What's special about `i==0` here? Maybe leave a comment explaining?

matthchr (approved):
Approved, but with one big question about eventual user experience

v2/tools/generator/internal/codegen/pipeline/create_backwards_compatibility_types.go
theunrepentantgeek: Updated naming for clarity.

v2/tools/generator/internal/codegen/pipeline/create_backwards_compatibility_types.go
theunrepentantgeek: Discussed offline and spec changed; we'll keep the API types for the alpha versions to avoid breaking existing devops scenarios.

v2/tools/generator/internal/codegen/storage/group_conversion_graph_builder.go
theunrepentantgeek: Reworked this method for clarity; included a comment too.

v2/api/resources/v1beta20200601/resourcegroup_types.go
```diff
@@ -179,7 +179,7 @@ func (status *ResourceGroupStatus) PopulateFromARM(owner genruntime.ArbitraryOwn
 	status.ManagedBy = typedInput.ManagedBy
 	status.Name = typedInput.Name
 	status.Tags = typedInput.Tags
-
+	// Set property ‘AccessTier’:
```
matthchr, line 182: Were you getting rid of this?

v2/internal/controllers/controller_resources.go
```diff
@@ -18,7 +18,7 @@ import (
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
 
 	"github.com/go-logr/logr"
-	corev1 "k8s.io/api/core/v1"
+	coreV1 "k8s.io/api/core/v1"
```
matthchr, line 21: minor: Was this fixed in the other commit? Or was it reverted here?

v2/tools/generator/internal/codegen/pipeline/create_types_for_backwards_compatibility.go
```diff
@@ -0,0 +1,146 @@
...
+			// Add the new types into our state
+			defs := state.Definitions()
+			defs.AddTypes(renamed)
+
+			return state.WithDefinitions(defs), nil
+		})
+
+	//
+	// Our pre-requisite support can't currently handle our needs because the earlier stage MUST be present, else the
+	// pipeline fails to run. These requirements are more of the "if the stage is present it must run before me" variety
```
matthchr, line 55: Is this in contrast to "the stage MUST EXIST and run before me"? It's weaker as it allows the stage to not exist?

v2/tools/generator/internal/codegen/pipeline/create_types_for_backwards_compatibility.go
```diff
@@ -0,0 +1,146 @@
...
+	}.Build()
+
+	result := make(astmodel.TypeDefinitionSet)
+
+	var errs []error
+	for _, def := range defs {
+		name := def.Name()
+		desc := []string{
+			fmt.Sprintf(
+				"Backward compatibility type for %s.%s",
```
matthchr, line 88: minor: Were you going to change this comment? I thought it had come up in the discussion we had on the Monday (Tuesday for you) sync up

matthchr (approved):

v2/api/resources/v1beta20200601/resourcegroup_types.go
theunrepentantgeek: Fixed in my ResourceGroup PR - possibly reverted here by my rebase. Will double check.

v2/internal/controllers/controller_resources.go
theunrepentantgeek: Was reverted by rebase; will fix.

v2/api/resources/v1beta20200601/resourcegroup_types.go
theunrepentantgeek: Was reverted by rebase; will fix.

v2/tools/generator/internal/codegen/pipeline/create_types_for_backwards_compatibility.go
theunrepentantgeek: Yeah; if the stage is present at all, it must run first, but it's ok for it to be missing.

v2/tools/generator/internal/codegen/pipeline/create_types_for_backwards_compatibility.go
theunrepentantgeek: Yes, I was - thanks for the nudge. Changed to this format:

``` go
//Deprecated version of Person_Spec. Use v1beta20200101.Person_Spec instead
```

PR #2157, opened 2022-03-14
Add support to handle validated 'number' type 

super-harsh:

As code-gen can't handle validated float64 types, validated `number` from swagger spec gets transformed into validated int. When we receive float value inside ARM status for that type, the reconciler runs into un-marshalling error as it's expecting an int type. 

To avoid the above scenario, changes in this PR make the number type as validated float for `Spec` and `Status`, then, further in a pipeline stage we change the validated float for Spec to a validated int. Which makes the number type as float64 type in `Status` and a validated int in `Spec`


v2/tools/generator/internal/codegen/pipeline/handle_float_values.go
```diff
@@ -0,0 +1,89 @@
...
+
+// HandleFloatValuesStageID is the unique identifier of this stage
+const (
+	HandleFloatValuesStageID = "handleFloatValues"
+)
+
+func HandleFloatValues() *Stage {
+	stage := NewStage(
+		HandleFloatValuesStageID,
+		"Transform Float type values to Integer types",
```
matthchr, line 24: It's not _all_ floats, right? Clarify a little bit in the description that it's `Validated` float values? And maybe leave a comment linking to the issue upstream explaining why they aren't supported

v2/tools/generator/internal/codegen/pipeline/handle_float_values.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/pkg/errors"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+)
+
+// HandleFloatValuesStageID is the unique identifier of this stage
+const (
+	HandleFloatValuesStageID = "handleFloatValues"
+)
+
+func HandleFloatValues() *Stage {
```
matthchr, line 21: Similarly to below, improve the name of this a bit to be a bit more clear about the specific float values being handled. It's not actually all of them, it's all of them that are validated? So `HandleValidatedFloats` or something?

v2/tools/generator/internal/codegen/pipeline/handle_float_values.go
matthchr: And also possibly note just dealing with `spec` floats because `status` types have all their validation stripped later?

v2/tools/generator/internal/codegen/pipeline/handle_float_values.go
```diff
@@ -0,0 +1,89 @@
...
+			if validatedType, ok := prop.PropertyType().(*astmodel.ValidatedType); ok {
+				ot = ot.WithProperty(prop.WithType(validatedType.WithType(astmodel.IntType)).MakeTypeOptional())
+			}
+		}
+	}
+
+	return astmodel.IdentityVisitOfObjectType(this, ot, ctx)
+}
+
+func walkTypes(definitions astmodel.TypeDefinitionSet, walker *astmodel.TypeWalker) (astmodel.TypeDefinitionSet, []error) {
```
matthchr, line 72: minor: Instead of using a `TypeWalker`, you can call `astmodel.FindSpecConnectedDefinitions` and just process each definition individually with a standard `astmodel.TypeVisitor`. `FindSpecConnectedDefinitions` uses a `TypeWalker` internally and saves you the pain of having to do it yourself.

v2/tools/generator/internal/codegen/pipeline/handle_float_values.go
```diff
@@ -0,0 +1,89 @@
...
+	stage.RequiresPrerequisiteStages(RemoveStatusPropertyValidationsStageID)
+
+	return stage
+}
+
+func getFloatTransformations(definitions astmodel.TypeDefinitionSet, validatedDefinitions astmodel.TypeDefinitionSet) (astmodel.TypeDefinitionSet, []error) {
+	walker := astmodel.NewTypeWalker(
+		definitions,
+		astmodel.TypeVisitorBuilder{
+			VisitObjectType: transformFloatToInt,
```
matthchr, line 50: Instead of doing this on `VisitObjectType` where you have to iterate the properties, etc, I think you can just write this:

```
VisitValidatedType: transformFloatToInt,
```

and then:

```
func transformFloatToInt(this *astmodel.TypeVisitor, validated *astmodel.ValidateType, ctx interface{}) (astmodel.Type, error) {
	if result, ok := astmodel.AsPrimitiveType(validated.ElementType()); ok && result == astmodel.FloatType {
		validated = validated.WithElement(astmodel.IntType)
	}
	astmodel.IdentityVisitOfValidatedType(this, validated, ctx)
}
```

I also don't think you need the `MakeTypeOptional` bit?

Actually you probably ought to set a context instead of changing the type right in the `VisitValidatedType`, so do:
```
func visitValidatedType(this *astmodel.TypeVisitor, validated *astmodel.ValidateType, ctx interface{}) (astmodel.Type, error) {
	astmodel.IdentityVisitOfValidatedType(this, validated, true) // Pass ctx so that transformFloatToInt can use it
}

func transformFloatToInt(this *astmodel.TypeVisitor, prim *astmodel.PrimitiveType, ctx interface{}) (astmodel.Type, error) {
	validated, ok := ctx.(bool)
	if prim == astmodel.FloatType && validated {
		return astmodel.IntType
	}
	astmodel.IdentityVisitOfPrimitiveType(this, validated, ctx)
}
```

v2/api/storage/v1alpha1api20210401/management_policy__status_arm_types_gen.go
```diff
@@ -134,14 +134,14 @@ type TagFilter_StatusARM struct {
 
 type DateAfterCreation_StatusARM struct {
 	//DaysAfterCreationGreaterThan: Value indicating the age in days after creation
-	DaysAfterCreationGreaterThan float64 `json:"daysAfterCreationGreaterThan,omitempty"`
+	DaysAfterCreationGreaterThan *float64 `json:"daysAfterCreationGreaterThan,omitempty"`
```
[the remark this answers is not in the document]
matthchr: See my comment about `MakeTypeOptional`

matthchr (changes requested):

v2/tools/generator/internal/codegen/pipeline/handle_float_values.go
theunrepentantgeek: I agree that the name is too generic - even just `TransformValidatedFloats` would be clearer.

v2/tools/generator/internal/codegen/pipeline/handle_float_values.go
theunrepentantgeek, line 24: We're doing this because of a limitation in `controller-gen` so we should mention that.

```suggestion
		"Transform validated float type values to validated integer types for compatibility with controller-gen",
```

v2/tools/generator/internal/codegen/pipeline/handle_float_values.go
```diff
@@ -0,0 +1,89 @@
...
+		}.Build())
+
+	return walkTypes(validatedDefinitions, walker)
+}
+
+// transformFloatToInt transforms all the FloatTypes to IntergerTypes
+func transformFloatToInt(this *astmodel.TypeVisitor, ot *astmodel.ObjectType, ctx interface{}) (astmodel.Type, error) {
+	for _, prop := range ot.Properties() {
+		// check if the base type is Float64 Type
+		if result, ok := astmodel.AsPrimitiveType(prop.PropertyType()); ok && result == astmodel.FloatType {
```
theunrepentantgeek, line 60: The `AsX()` methods are great for querying types but shouldn't be used for transforming them - you end up throwing away any/all wrappers around the types (these properties might have `OptionalType`, `FlaggedType`, etc around them; you want to change the interior of the onion, not blow it away).

The TypeVisitor can do this iteration/unwrap/rewrap for you: 
* Add a handler for `VisitValidatedType` that passes in a context object to the recursive call
* Add a handler for `VisitPrimitive` that checks for context + float and changes returns an int instead

v2/api/storage/v1alpha1api20210401/management_policy__status_arm_types_gen.go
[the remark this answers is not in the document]
theunrepentantgeek: Agreed; should only change the type, not optionality. This is happening because `AsPrimitiveType()` is throwing away layers of the composite type. See my other comment for how to use `TypeVisitor` to make the change while preserving all the layers.

theunrepentantgeek (changes requested):
Nice surgical use of a pipeline stage to make the required change.

v2/api/storage/v1alpha1api20210401/management_policy__status_arm_types_gen.go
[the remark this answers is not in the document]
super-harsh: This is the behaviour I also noticed. So having pointer value is the right thing. When I added transformation using azure-arm.yaml spec by using `  target:  name: float`, it just replaced with a non pointer type. As of the changes made by @matthchr in #2126 this is right.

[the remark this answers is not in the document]
super-harsh: Even with all the suggested changes, I'm getting the similar results for this one.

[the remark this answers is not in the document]
matthchr: That makes sense actually. Agreed that this is getting generated as expected. I was thinking it was right before but as you say actually it was not.

v2/tools/generator/internal/codegen/pipeline/transform_validated_floats.go
```diff
@@ -0,0 +1,63 @@
...
+	return visitor.VisitDefinitions(astmodel.FindSpecConnectedDefinitions(definitions))
+}
+
+func visitValidatedType(this *astmodel.TypeVisitor, validated *astmodel.ValidatedType, ctx interface{}) (astmodel.Type, error) {
+	return astmodel.IdentityVisitOfValidatedType(this, validated, true) // Pass ctx so that transformFloatToInt can use it
+}
+
+// transformFloatToInt transforms all the validated FloatTypes to IntegerTypes
+func transformFloatToInt(this *astmodel.TypeVisitor, prim *astmodel.PrimitiveType, ctx interface{}) (astmodel.Type, error) {
+	validated, _ := ctx.(bool)
```
theunrepentantgeek, line 58: Nitpick - need to check to see whether `ctx` really was a **bool**, more to avoid issues with future maintenance than anything else. 

Can easily be done by including a local `ok` here and then checking that as well in the condition just below.

v2/api/storage/v1alpha1api20210401/management_policy__status_arm_types_gen.go
```diff
@@ -134,14 +134,14 @@ type TagFilter_StatusARM struct {
...
 
 type DateAfterModification_StatusARM struct {
 	//DaysAfterLastAccessTimeGreaterThan: Value indicating the age in days after last blob access. This property can only be
 	//used in conjunction with last access time tracking policy
-	DaysAfterLastAccessTimeGreaterThan *int `json:"daysAfterLastAccessTimeGreaterThan,omitempty"`
+	DaysAfterLastAccessTimeGreaterThan *float64 `json:"daysAfterLastAccessTimeGreaterThan,omitempty"`
 
 	//DaysAfterModificationGreaterThan: Value indicating the age in days after last modification
-	DaysAfterModificationGreaterThan *int `json:"daysAfterModificationGreaterThan,omitempty"`
+	DaysAfterModificationGreaterThan *float64 `json:"daysAfterModificationGreaterThan,omitempty"`
```
theunrepentantgeek, line 146: This change looks good.

theunrepentantgeek (approved):

v2/tools/generator/internal/codegen/pipeline/transform_validated_floats.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"github.com/Azure/azure-service-operator/v2/tools/generator/internal/astmodel"
+)
+
+// TransformValidatedFloatsStageID is the unique identifier of this stage
+const (
+	TransformValidatedFloatsStageID = "transformValidatedFloats"
+)
+
+func TransformValidatedFloats() *Stage {
```
matthchr, line 19: Nice. This is quite clean.

matthchr (approved):

PR #2167, opened 2022-03-17
Update our decision doc on how we transition to beta

theunrepentantgeek:

We've identified some issues with the approach we were going to take for our transition to beta. Updating the spec to outline the options and consequences to promote a fuller discussion.

gif

**If applicable**:


docs/hugo/content/architecture-decision-records/ADR-2022-02-Backward-Compatibility.md
```diff
@@ -1,17 +1,88 @@
...
+Change the generated prefix for all resources
+
+Pros
+
+* Easy to implement (change the prefix, update affected tests, and regenerate all files), and quick to deliver
+
+Cons
+
+* Existing clusters with custom resources (CRs) reliant on ASO would break because ASO would no longer support the older version.  
+  Mitigation would require deleting the existing resource and creating a new CR with an updated version for the resource.
```
davefellows, line 29:
```suggestion
  Migration would require deleting the existing resource and creating a new CR with an updated version for the resource.
```

davefellows: Not sure if that was meant to read "Migration"? I guess "Mitigation" also works :)

docs/hugo/content/architecture-decision-records/ADR-2022-02-Backward-Compatibility.md
```diff
@@ -1,17 +1,88 @@
...
+Pros
+
+* Moderate effort to implement (requires a new pipeline stage to create the alpha storage variants for compatibility with older versions).
+* Existing clusters with custom resources reliant on ASO would continue to run because ASO would support conversion from the old hub version to the new.
+
+Cons
+
+* Existing automated deployments (e.g. as used by teams with a GitOpts model or other forms of continuous deployment) would break because the version specified in the existing YAML files would no longer exist.  
+  Mitigation would require modification of the deployment pipelines to change the version of the resources being deployed.
+* Existing manual deployments would also break because the version specified in the existing YAML files would no longer exist.
```
matthchr, line 48: Might be clearer to say users attempting to update existing CRs (by reapplying an updated YAML). I think the word "deployments" is ambiguous because there is a Kubernetes resource Deployments (that doesn't come into play here at all). 

Since this is basically the same thing as the first bullet you could also combine them into something like:
"Any attempt to apply old v1alpha1 YAML will fail, as it no longer exists. This includes attempting to manually apply a YAML using `kubectl apply -f` or through some other mechanism where YAMLs are continuously applied, as in GitOps style workflows"

docs/hugo/content/architecture-decision-records/ADR-2022-02-Backward-Compatibility.md
```diff
@@ -1,17 +1,88 @@
...
+
+* Moderate effort to implement (requires pipeline and configuration changes to generate multiple versions of some resources).  
+  Could be implemented in two parts: generate all resources with both prefixes for `codegen-beta-0`, and introduce configuration for newer resources in `codegen-beta-1`.
+* Existing clusters with custom resources reliant on ASO would continue to run because ASO would support conversion from the old hub version to the new.
+* Existing automated deployments (e.g. as used by teams with a GitOpts model or other forms of continuous deployment) would continue to run because ASO would still support the version.
+* Existing manual deployments would also continue to run because ASO would still support the version.
+
+Cons
+
+* The size of our CRDs would double - and they're already large.
```
matthchr, line 65: Is this actually a problem? I know there's a limit but it's only 25% more content than Option 3b, not double 3b?

docs/hugo/content/architecture-decision-records/ADR-2022-02-Backward-Compatibility.md
theunrepentantgeek: Good phrasing, I'll use that.

docs/hugo/content/architecture-decision-records/ADR-2022-02-Backward-Compatibility.md
theunrepentantgeek: I've clarified the context. I don't think we're too close to the limit at the moment, but we can't afford to double very many times!

matthchr (approved):

PR #2180, opened 2022-03-22
Add support to record azure based metrics 

### LINKED ISSUE #1464: Enable metrics and tracing
Wire up OpenCensus metrics and tracing.

- Main should take in parameters for configuring / exposing / relaying metrics and traces.
- Zips and the Azure SDK for Go should be wired up as well (Azure SDK for Go has tracing / metrics support)

Reasoning:
- Provide insight to where time is spent
- Help to diagnose optimal time for deployment polling / requeue
- Enable better debugging

super-harsh:
Closes #1464 

In this PR, we're making changes to add support for registering new Azure based metrics below which would help us with better debugging and diagnosability towards ARM requests from reconciler. 

Metrics Added:
- azure_requests_total (PUT and DELETE)
- azure_failed_requests_total (PUT and DELETE)
- azure_requests_time_seconds (PUT and DELETE)
- azure_response_code (PUT and DELETE)

Not including tracing atm, this is treated as MVM for now. Excluding tracing from the plan as it does not add much value for the end user, we already have rich metrics and logging which gives us enough information regarding time spent and the flow.


v2/internal/metrics/metrics.go
```diff
@@ -0,0 +1,102 @@
...
+		Name: "requests_requeue_total",
+		Help: "Total number of request requeue(s) for resource",
+	}, []string{"name", "requestType"})
+)
+
+func init() {
+	metrics.Registry.MustRegister(azureRequestsTotal, azureFailedRequestsTotal, azureRequestsTime, azureResponseCode)
+}
+
+func RecordRequeueTotalPUT(resourceName string) {
```
matthchr, line 64: Rather than putting the name of the HTTP verb in the method, could you just take an enum that defines the verb as a parameter? There's already one defined in `http`: `http.MethodPut`, which you could use (although I guess that's not actually an enum it's just a string).

v2/internal/genericarmclient/generic_client.go
```diff
@@ -115,16 +116,25 @@ func (client *GenericClient) BeginCreateOrUpdateByID(ctx context.Context, resour
 
 func (client *GenericClient) createOrUpdateByID(
 	ctx context.Context,
-	resourceID string,
+	resourceID,
+	resourceName,
```
matthchr, line 120: Do you mean this to be the resource type `ResourceGroup`, or the name of the resource? What you have here is the name of the resource, like "myFoo", which I don't think you want?

matthchr: It's also possible to parse this from the `resourceID` so you might not need a separate parameter.

v2/internal/metrics/metrics.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+const (
+	putLabel    = "PUT"
+	deleteLabel = "DELETE"
+)
+
+// TODO: Add more status codes here if there are more expected codes.
+// Basic http API request status codes
+var statusCodes = []float64{
```
matthchr, line 23: Are these actually floats or are they ints?

v2/internal/metrics/metrics.go
```diff
@@ -0,0 +1,102 @@
...
+	}, []string{"name", "requestType"})
+
+	requeueTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "requests_requeue_total",
+		Help: "Total number of request requeue(s) for resource",
+	}, []string{"name", "requestType"})
+)
+
+func init() {
+	metrics.Registry.MustRegister(azureRequestsTotal, azureFailedRequestsTotal, azureRequestsTime, azureResponseCode)
```
matthchr, line 61: Is this the standard way to do this? Wondering if it would be cleaner to do this not as a package global but instead initialized in main and passed down to the ARM client? I believe the ARM client is initialized right there in main as well so it'd be easy to pass it around.

Package globals are a bit of a smell IMO.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -65,13 +66,18 @@ func (r *azureDeploymentReconcilerInstance) CreateOrUpdate(ctx context.Context)
...
+	//TODO: check result here and record requeue and move RecordAzureRequestsTotalPUT inside where making put request
+
 	if err != nil {
 		r.Log.Error(err, "Error during CreateOrUpdate", "action", action)
 		r.Recorder.Event(r.Obj, v1.EventTypeWarning, "CreateOrUpdateActionError", err.Error())
-
 		return ctrl.Result{}, err
 	}
 
+	if result.Requeue || result.RequeueAfter > 0 {
```
matthchr, line 77: Why not record Total PUTs and delete's in the generic client too? I think with this here, we will overcount PUTs because the fact there's a requeue doesn't mean there was a PUT

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -89,13 +95,17 @@ func (r *azureDeploymentReconcilerInstance) Delete(ctx context.Context) (ctrl.Re
...
 	result, err := actionFunc(ctx)
+
 	if err != nil {
 		r.Log.Error(err, "Error during Delete", "action", action)
 		r.Recorder.Event(r.Obj, v1.EventTypeWarning, "DeleteActionError", err.Error())
-
 		return ctrl.Result{}, err
 	}
 
+	if result.Requeue || result.RequeueAfter > 0 {
```
matthchr, line 105: Same comment here about, why not do this in the client

v2/charts/azure-service-operator/values.yaml
```diff
@@ -45,3 +45,8 @@ image:
   repository: mcr.microsoft.com/k8s/azureserviceoperator:v2.0.0-alpha.6
   kubeRBACProxy: gcr.io/kubebuilder/kube-rbac-proxy:v0.4.1 
 
+# 'metrics' define settings for the metrics from controller.
+# 'address' field defines the metrics binding address on which metrics
+metrics:
+  enable: true
```
matthchr, line 51: Where does this flag end up getting passed?

v2/cmd/controller/main.go
```diff
@@ -39,7 +39,7 @@ func main() {
 
 	var metricsAddr string
 	var enableLeaderElection bool
-	flagSet.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
+	flagSet.StringVar(&metricsAddr, "metrics-addr", "0", "The address the metric endpoint binds to.")
```
matthchr, line 42: The default of `0` doesn't seem to make sense? Maybe `:8080` didn't either? does 0 here mean "don't run metrics at all"? That's sorta surprising because it's a string so I'd expect that empty string did that?

v2/charts/azure-service-operator/values.yaml
super-harsh: Basically this doesn't end up being a flag.. turning this to false omits the `metrics-addr` flag which results in a default `disable`

v2/cmd/controller/main.go
super-harsh: Yes, you are right. "0" is for disabling the metrics (standard prometheus setting). 

I agree on this being a bit unambiguous, should add a flag `enable-metrics` which then forces `metricsAddr` to be "0"?

v2/internal/metrics/metrics.go
super-harsh: They are ints, they get parsed as floats. However a histogram doesn't make sense here, I'm actually planning to remove this and merge status code as a label with `request_counter`

v2/internal/genericarmclient/generic_client.go
super-harsh: True, `kind` would make more sense here?

v2/internal/metrics/metrics.go
super-harsh: this is referring to `sigs.k8s.io/controller-runtime/pkg/metrics` and registering with the existing manager's prometheus server.

v2/internal/genericarmclient/generic_client.go
super-harsh: can surely parse `type` from resourceID but would that be a good idea? Would resourceID format be changing from resource to resource ?

matthchr: There's a defined format and a parse method. See [here](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/arm/resource_identifier.go#L64). We use this as `arm.ParseResourceID` in a few places already. I _think_ the string you want is:

```
parsedID, err := arm.ParseResourceID(id)
if err != nil {
    // Handle error
}

fullIdentifier := parsedID.ResourceType.Namespace + "/" + parsedID.ResourceType.Type
```

v2/internal/genericarmclient/generic_client.go
```diff
@@ -93,7 +96,7 @@ func (client *GenericClient) ClientOptions() *arm.ClientOptions {
 	return client.opts
 }
 
-func (client *GenericClient) BeginCreateOrUpdateByID(ctx context.Context, resourceID string, apiVersion string, resource interface{}) (*PollerResponse, error) {
+func (client *GenericClient) BeginCreateOrUpdateByID(ctx context.Context, resourceID, apiVersion string, resource interface{}) (*PollerResponse, error) {
```
theunrepentantgeek, line 99: Better to be explicit with types in function signatures - and given there are now several, consider putting them across several lines:

```suggestion
func (client *GenericClient) BeginCreateOrUpdateByID(
    ctx context.Context, 
    resourceID string, 
    apiVersion string, 
    resource interface{}) (*PollerResponse, error) {
```

v2/internal/genericarmclient/generic_client.go
```diff
@@ -115,16 +118,26 @@ func (client *GenericClient) BeginCreateOrUpdateByID(ctx context.Context, resour
 
 func (client *GenericClient) createOrUpdateByID(
 	ctx context.Context,
-	resourceID string,
+	resourceID,
```
theunrepentantgeek, line 121: Same here - cascading several parameter types together invites misinterpretation. Be explicit.

v2/internal/genericarmclient/generic_client.go
```diff
@@ -115,16 +118,26 @@ func (client *GenericClient) BeginCreateOrUpdateByID(ctx context.Context, resour
...
 	req, err := client.createOrUpdateByIDCreateRequest(ctx, resourceID, apiVersion, resource)
 	if err != nil {
 		return nil, err
 	}
+
+	requestStartTime := time.Now()
 	resp, err := client.pl.Do(req)
+
+	// Ignoring error here as resourceID would always be correct at this stage.
+	resourceType, _ := metrics.GetTypeFromResourceID(resourceID)
```
theunrepentantgeek, line 134: Maybe panic? I've made similar assertions in comments a few times and have been encouraged to turn them into **panic** calls so that the assumption is verified.

v2/internal/genericarmclient/generic_client.go
```diff
@@ -212,7 +225,7 @@ func (client *GenericClient) getByIDHandleResponse(resp *http.Response, resource
 
 // DeleteByID - Deletes a resource by ID.
 // If the operation fails it returns the *CloudError error type.
-func (client *GenericClient) DeleteByID(ctx context.Context, resourceID string, apiVersion string) (time.Duration, error) {
+func (client *GenericClient) DeleteByID(ctx context.Context, resourceID, apiVersion string) (time.Duration, error) {
```
theunrepentantgeek, line 228:
```suggestion
func (client *GenericClient) DeleteByID(ctx context.Context, resourceID string, apiVersion string) (time.Duration, error) {
```

v2/internal/genericarmclient/generic_client.go
```diff
@@ -224,15 +237,26 @@ func (client *GenericClient) DeleteByID(ctx context.Context, resourceID string,
 
 // DeleteByID - Deletes a resource by ID.
 // If the operation fails it returns the *CloudError error type.
-func (client *GenericClient) deleteByID(ctx context.Context, resourceID string, apiVersion string) (*http.Response, error) {
+func (client *GenericClient) deleteByID(ctx context.Context, resourceID, apiVersion string) (*http.Response, error) {
```
theunrepentantgeek, line 240:
```suggestion
func (client *GenericClient) deleteByID(ctx context.Context, resourceID string, apiVersion string) (*http.Response, error) {
```

v2/internal/metrics/arm_client_metrics.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+type HttpMethod string
```
theunrepentantgeek, line 17: Is there not already an enum for this defined somewhere else we can reuse?

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -65,10 +65,10 @@ func (r *azureDeploymentReconcilerInstance) CreateOrUpdate(ctx context.Context)
 	r.Log.V(Verbose).Info("Reconciling resource", "action", action)
 
 	result, err := actionFunc(ctx)
+
```
theunrepentantgeek, line 68: We usually have the error check (`if err != nil`) on the immediately following line (no gap) because it's a part of checking the statement worked.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -65,10 +65,10 @@ func (r *azureDeploymentReconcilerInstance) CreateOrUpdate(ctx context.Context)
...
 	if err != nil {
 		r.Log.Error(err, "Error during CreateOrUpdate", "action", action)
 		r.Recorder.Event(r.Obj, v1.EventTypeWarning, "CreateOrUpdateActionError", err.Error())
-
```
theunrepentantgeek, line 71: I like to see a blank line before **return** - makes it easier to see what's returned.

theunrepentantgeek (approved):
A few minor comments, mostly about coding style and being consistent with the rest of the ASO project. 

gif

v2/internal/genericarmclient/generic_client.go
super-harsh: I think we don;t wanna go into the panic state if resourceID is not parsed properly. If there's something wrong with the resourceID, we'll get an error from ARM client anyway. We're just recording metrics here. Rather just thinking of it now, if resourceID is not parsed properly, we can just log the error and have the whole resourceID recorded in the metrics?

What do you think about above?

v2/internal/metrics/arm_client_metrics.go
super-harsh: AFAIK, there's nothing in `net/http` and other standard go libs. Have a look at the link below:

https://pkg.go.dev/search?limit=25&m=symbol&q=httpmethod

PR #2181, opened 2022-03-22
Document why we are committing our generated code

### LINKED ISSUE #1778: Documentation: Why check in the code generated files?
We need to document why we made the decision to check in our generated files - while we have good reasons for doing so, it does go against usual practice, and it's reasonable to expect new devs will question why we did it.

Suggest creating a file under `docs/architecture`.

Needs to cover
* Why we need a history
* Why we want to be aware of changes
* Backward compatibility

theunrepentantgeek:

Documents the motivations behind our unusual decision to commit our generated code. 

Closes #1778 

gif

**If applicable**:


docs/hugo/content/architecture-decision-records/ADR-2021-08-Committing-Generated-Code.md
```diff
@@ -0,0 +1,59 @@
+# Committing Generated Code
+
+## Context
+
+We’re aware that common practice avoids committing generated code into source control, treating it as an ephemeral resource that can be regenerated on demand.
+
+Even so, we are concluding that we want to check in our generated code.
```
matthchr, line 7: minor: Not sure of tense here. Maybe, "we concluded" or "we have concluded"? Since we've made that conclusion and this is documenting how we came to it?

matthchr (approved):

docs/hugo/content/architecture-decision-records/ADR-2021-08-Committing-Generated-Code.md
theunrepentantgeek: The Context section of an ADR is supposed to be written in present tense to describe the context as at that time.

PR #2186, opened 2022-03-24
Fix bug where the conversion graph wouldn't bridge for resources that skip versions

### LINKED ISSUE #2185: Incorrect conversions generated when resources imported from non-adjacent versions
Instead of importing all resources from each version, we have adopted a surgical approach where we import specified resources only. These may be selected from different versions (the reasons for this are outside the scope of this issue).

For example, returning to the hypothetical CRM service used for many of our specifications, we may import **Person** from one version, **Location** from a second, and **Event** from a third, giving rise to the following conversion graph between versions:

adr-2022-02-crm-version-graph

From this, we generate a set of available conversions between resources (note that each is a subset of the graph above):

adr-2022-02-crm-resource-graphs

When introducing a second version of Person, the current code works correctly if the newly introduced version is adjacent to the previous one:

adr-2022-02-crm-person-graph-valid

But we encounter a previously unanticipated edge case when the newly introduced version is not adjacent:

adr-2022-02-crm-person-graph-invalid

We end up with two hub versions of Person, which is neither correct nor supported.

What we want is for the **Person** conversion graph to correctly link the two available storage versions:

adr-2022-02-crm-person-graph-fixed

theunrepentantgeek:

Creation of the conversion graph was incorrectly identifying multiple versions of the same resource as Hub types because it was failing to skip versions where the resource does not exist. See #2185 for more details.

Closes #2185 

gif

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/tools/generator/internal/codegen/storage/conversion_graph.go
+++ b/v2/tools/generator/internal/codegen/storage/conversion_graph.go
@@ -39,77 +39,45 @@ func (graph *ConversionGraph) LookupTransition(ref astmodel.PackageReference) (a
 }
 
 // FindNextType returns the type name of the next closest type on the path to the hub type.
-// Returns the type name and no error if the next type is found; nil and no error if not; nil and an error if something
-// goes wrong.
+// Returns the type name and no error if the next type is found; empty name and no error if not; empty name and an error if
+// something goes wrong.
 // If the name passed in is for the hub type for the given resource, no next type will be found.
 // This is used to identify the next type needed for property assignment functions, and is a building block for
 // identification of hub definitions.
 func (graph *ConversionGraph) FindNextType(name astmodel.TypeName, definitions astmodel.TypeDefinitionSet) (astmodel.TypeName, error) {
 
-	// Find the next package to consider
-	nextPackage, ok := graph.LookupTransition(name.PackageReference)
-	if !ok {
-		// No next package, return nil. This is not an error condition.
-		return astmodel.EmptyTypeName, nil
+	// Look for a next type with the same name
+	nextType, nextSteps := graph.searchForMatchingType(name, definitions)
+
+	// Look for a renamed type with the same name
+	renamedType, renameSteps, err := graph.searchForRenamedType(name, definitions)
+	if err != nil {
+		return astmodel.EmptyTypeName, errors.Wrapf(err, "searching for type renamed from %s", name)
 	}
 
-	// Look to see if we have a type-rename that's directing us to use a different type
-	//
-	// We only look for type renames if we're starting from a storage package - no renames are needed when converting
-	// from an api package to a storage package because the storage versions are always synthesized with an exact match
-	// on type names.
-	//
-	haveRename := false
-	rename := ""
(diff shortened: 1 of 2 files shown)


v2/tools/generator/internal/codegen/storage/conversion_graph.go
```diff
@@ -39,77 +39,45 @@ func (graph *ConversionGraph) LookupTransition(ref astmodel.PackageReference) (a
 }
 
 // FindNextType returns the type name of the next closest type on the path to the hub type.
-// Returns the type name and no error if the next type is found; nil and no error if not; nil and an error if something
-// goes wrong.
+// Returns the type name and no error if the next type is found; empty name and no error if not; nil and an error if
```
babbageclunk, line 42:
```suggestion
// Returns the type name and no error if the next type is found; empty name and no error if not; empty name and an error if
```

babbageclunk (approved):
Looks good to me!

v2/tools/generator/internal/codegen/storage/conversion_graph.go
```diff
@@ -174,3 +142,83 @@ func (graph *ConversionGraph) FindNextProperty(
...
+	typeName astmodel.TypeName,
+	definitions astmodel.TypeDefinitionSet) (astmodel.TypeName, int) {
+	return graph.searchForMatchingTypeImpl(typeName, definitions, 1)
+}
+
+// searchForMatchingTypeImpl is the recursive implementation of searchForMatchingType.
+func (graph *ConversionGraph) searchForMatchingTypeImpl(
+	typeName astmodel.TypeName,
+	definitions astmodel.TypeDefinitionSet,
+	steps int) (astmodel.TypeName, int) {
```
matthchr, line 209: really minor: `steps` sorta sounds like it's an instruction on how many steps to use, rather than a "current state". Wondering if it might be clearer as `currentSteps` or something. Obviously this is super minor though.

v2/tools/generator/internal/codegen/storage/conversion_graph.go
```diff
@@ -174,3 +142,83 @@ func (graph *ConversionGraph) FindNextProperty(
...
+	// If we have any error other than a NotConfiguredError, something went wrong, and we must abort
+	if err != nil {
+		return astmodel.EmptyTypeName, -1, errors.Wrapf(
+			err,
+			"finding next type after %s",
+			name)
+	}
+
+	newType := name.WithName(rename)
+	result, steps := graph.searchForMatchingTypeImpl(newType, definitions, 1)
```
matthchr, line 176: minor: isn't this just `searchForMatchingType(newType, definitions)`?

matthchr (approved):
LGTM

v2/tools/generator/internal/codegen/storage/conversion_graph.go
theunrepentantgeek: Renamed to `stepsSoFar`

v2/tools/generator/internal/codegen/storage/conversion_graph.go
theunrepentantgeek: Yup. Fixed.

PR #2188, opened 2022-03-24
Generate multitenant deployment yaml files automatically

babbageclunk:
The multitenant deployment files can be generated by running the `controller:make-multitenant-files` task. They'll be generated and attached to new releases.

Also adds an integration test to validate that we can create resources in a multitenant-deployed cluster that is run in the PR validation CI.

gif

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -34,6 +34,11 @@ jobs:
           container_id=${{steps.devcontainer.outputs.id}}
           docker exec "$container_id" task controller:run-kustomize
 
+      - name: Make multitenant files
+        run: |
+          container_id=${{steps.devcontainer.outputs.id}}
+          docker exec "$container_id" task controller:make-multitenant-files
+
       - name: Build & tag Docker image
         run: |
           container_id=${{steps.devcontainer.outputs.id}}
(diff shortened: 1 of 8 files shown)


.github/workflows/create-release.yml
```diff
@@ -46,7 +51,7 @@ jobs:
         with:
           repo_token: ${{ secrets.GITHUB_TOKEN }}
           tag: ${{ github.ref }}
-          file: v2/bin/azureserviceoperator_*.yaml
+          file: "v2/bin/{azureserviceoperator,multitenant-cluster,multitenant-tenant}_*.yaml"
```
babbageclunk, line 54: I got a bit worried about whether the braces would be respected in the pattern but it [seems like they should be](https://www.npmjs.com/package/glob). 
(I haven't tested the release action.)

scripts/make-multitenant-cluster.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+set -euxo pipefail
```
babbageclunk, line 2: I had the `x` option turned on for development but I've left it on since these scripts will mostly be running in CI or during a release so having the extra logging seems useful.

theunrepentantgeek (approved):

PR #2189, opened 2022-03-25
Fix bug where Version configuration wasn't used for some packages

theunrepentantgeek:

Discovered during work to move our generator prefix to `v1beta`, our configuration wasn't being picked up by some packages, resulting in needed modifications not being made to the object model.

This PR changes the way we look up version-configuration within the object-model-configuration hierarchy so that it's independent of the generator version.

We also enable configuration based on the generator version as well (which is needed by some tests), ensuring some forward compatibility.

image

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/tools/generator/internal/astmodel/local_package_reference.go
+++ b/v2/tools/generator/internal/astmodel/local_package_reference.go
@@ -13,10 +13,10 @@ import (
 
 // LocalPackageReference specifies a local package name or reference
 type LocalPackageReference struct {
-	localPathPrefix string
-	group           string
-	version         string
-	versionPrefix   string
+	localPathPrefix  string
+	group            string
+	apiVersion       string
+	generatorVersion string
 }
 
 var (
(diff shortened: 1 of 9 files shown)


babbageclunk (approved):

v2/tools/generator/internal/astmodel/local_package_reference.go
```diff
@@ -87,18 +87,26 @@ func (pr LocalPackageReference) String() string {
...
 	return pr
 }
 
 // HasVersionPrefix returns true if we have the specified version prefix, false otherwise.
 func (pr LocalPackageReference) HasVersionPrefix(prefix string) bool {
-	return pr.versionPrefix == prefix
+	return pr.generatorVersion == prefix
+}
+
+func (pr LocalPackageReference) GeneratorVersion() string {
```
matthchr, line 104: minor: Missing some comments here, and below

v2/tools/generator/internal/astmodel/local_package_reference.go
```diff
@@ -87,18 +87,26 @@ func (pr LocalPackageReference) String() string {
...
+	return pr.generatorVersion
+}
+
+func (pr LocalPackageReference) ApiVersion() string {
```
matthchr, line 108: minor: [this](https://github.com/golang/go/wiki/CodeReviewComments#initialisms) suggests that we should say `APIVersion`, not `ApiVersion`.

I think we violate this rule in places (I know I've fixed a few that I named that way), but should try to honor it going forward IMO

PR #2196, opened 2022-03-28
Fix helm chart generation and API docs creation workflows

### LINKED ISSUE #2160: Auto-create PR for ASOv2 helm chart and API docs

Currently we have to manually generate helm chart and API docs and create a PR for them for each release. 

**Describe the improvement**

To improve the above issue we can automate the process to generate helm manifest and API docs, then get a PR created. We have a few options below to tackle this:

1. Create a new workflow for above which runs after create-release workflow
2. Add a step in create-release workflow to automate above.
3. Create a new scheduled workflow to update the above weekly
4. 1 AND 3
5. 2 AND 3

super-harsh:
Closes #2160 

This PR contains changes to the helm chart and API docs workflow to make them functional as expected. For now the trigger is set to manual for testing purposes. Later, will change the trigger to be :
```
on:
  # run when a release is published,
  # which creates a new tag
  push:
    tags:
    - v2*
```


.github/workflows/api-docs-repo.yaml
```diff
@@ -64,19 +65,19 @@ jobs:
...
 
+      - name: Commit and push changes
+        uses: devops-infra/action-commit-push@v0.9.0
+        with:
+          github_token: ${{ secrets.GH_PAT }}
+          commit_message: Add API Docs
+
       - name: Create Pull Request
-        uses: peter-evans/create-pull-request@7380612b49221684fefa025244f2ef4008ae50ad # v3.10.1
+        uses: devops-infra/action-pull-request@v0.4.2
```
matthchr, line 75: Should we be using this everywhere?

matthchr (approved):

PR #2200, opened 2022-03-29
Fix to make helm and API docs auto-PR functional on new release push

super-harsh:

- Small one here to change the trigger for helm and api docs workflows. Now will be triggered once a new release tag is pushed.
- Removed mac-ism from `gen-api-docs.sh` script.


scripts/gen-api-docs.sh
```diff
@@ -35,4 +35,4 @@ do
 done
 
 # Hacky way to get through the "plain text html not allowed" hugo error
-find $OUTPUTDIR -type f -exec sed -i '' '1 s/^/---\n---\n/' {} \;
+find $OUTPUTDIR -type f -exec sed -i '1 s/^/---\n---\n/' {} \;
```
matthchr, line 38: Ah, did we do this instead of turning on full HTML rendering in Hugo?

matthchr (approved):

.github/workflows/api-docs-repo.yaml
```diff
@@ -1,7 +1,10 @@
 name: Publish API docs
 on:
-  workflow_dispatch:
-  # no content, allows manual triggering
+  # run when a release is published,
+  # which creates a new tag
+  push:
+    tags:
+      - v2*
```
theunrepentantgeek, line 7: This disallows manual triggering - do we need to consider that case (say, for when we change the template or CSS and don't retag)? 

Suggest keeping the `workflow_dispatch` option so that we can.

.github/workflows/helm-chart-repo.yaml
```diff
@@ -1,7 +1,10 @@
 name: Publish helm charts
 on:
-  workflow_dispatch:
-  # no content, allows manual triggering
+  # run when a release is published,
+  # which creates a new tag
+  push:
+    tags:
+      - v2*
```
theunrepentantgeek, line 7: This one might be ok without manual triggering.

theunrepentantgeek (approved):
A couple minor comments about triggering, but otherwise looks good.

scripts/gen-api-docs.sh
super-harsh: We need both, or else it doesn't work.

.github/workflows/api-docs-repo.yaml
super-harsh: Done!

PR #2201, opened 2022-03-29
Fix code generation for object conversion with shared sub-objects

### LINKED ISSUE #2193: Bug: Incorrect code generated for object conversion with shared sub-objects
**Describe the bug**

We're generating incorrect code when a shared object is present across more versions than the resource being handled.

To illustrate, consider the following:

adr-2022-02-shared-resources-initial dot

While the resource `Person` exists in only versions **v1** and **v3**, the shared object `Location` _also_ exists in **v2**.

When generating the `AssignTo()` and `AssignFrom()` methods for `v1.Person` we correctly use `v3.Person` as the destination and source, allowing the conversion to work. (Issue #2185 documents the problems we encountered when conversions wouldn't jump unused versions).

However, those conversions are currently assuming that the `AssignTo()` and `AssignFrom()` methods on `v1.Location` should be passed the `Residence` from `v3.Person` - but that's a `v3.Location`.

The generated code doesn't compile, because the `AssignTo()` and `AssignFrom()` methods on `v1.Location` expect a `v2.Location`, not a `v3.Location`.

This was discovered when generating code for both `Redis` and `RedisEnterprise` from the `cache` group.

Another way to look at this is that we're currently assuming that both `Person` and `Location` have the same set of versions in their conversion graph. When this isn't the case, incorrect code is generated.

**Expected behavior**

Conversion between `v1.Location` and `v3.Location` should happen via the intermediate type `v2.Location`.

theunrepentantgeek:

Fixes the incorrect code generation that happens when there are intermediate versions of a nested object that exist in between versions of a parent resource. See #2193 for a full description.

Closes #2193 

Look at `SharedObject.golden` to see how the extra versions are bridged.

I separated `assignObjectToObject` into two separate versions (one for _ConvertTo_, one for _ConvertFrom_) to make it easier to read what they were doing.

Needed to move `IsKubernetesResourceProperty` into `astmodel` to defuse a circular reference.

I also did some cleanup to reduce the volume of boilerplate in `property_conversions.go`, and to remove **nil** parameters being supplied to `PropertyConversionContext` during testing.

gif

**If applicable**:
- [x] this PR contains tests


v2/tools/generator/internal/conversions/property_conversions.go
```diff
@@ -1374,73 +1391,162 @@ func assignObjectFromObject(
...
+//
+// Note the actual steps are generated by nested conversions; this handler works by finding the two conversions needed
+// given our intermediate type and chaining them together.
+//
+func assignObjectsViaIntermediateObject(
+	sourceEndpoint *TypedConversionEndpoint,
+	destinationEndpoint *TypedConversionEndpoint,
+	conversionContext *PropertyConversionContext) (PropertyConversion, error) {
+
+	// Require both source and destination to not be bag items
```
super-harsh, line 1440: Consider putting these checks in a function to be re-used by other methods(assignObjectDirectlyFromObject) as well?

v2/tools/generator/internal/functions/property_assignment_function_test.go
```diff
@@ -375,3 +376,58 @@ func TestGolden_PropertyAssignmentFunction_WhenTypeRenamed(t *testing.T) {
...
+
+	cfg := config.NewObjectModelConfiguration()
+	builder := storage.NewConversionGraphBuilder(cfg, "v")
+	builder.Add(test.Pkg2020, test.Pkg2021, test.Pkg2022)
+	graph, err := builder.Build()
+	g.Expect(err).To(BeNil())
+
+	conversionContext := conversions.NewPropertyConversionContext(definitions, idFactory).WithConversionGraph(graph)
+
+	assignFrom, err := NewPropertyAssignmentFunction(person2020, person2022, conversionContext, conversions.ConvertFrom)
```
super-harsh, line 423: Would it make sense to create `NewPropertyAssignmentFunction` for `Location` as well here?

v2/tools/generator/internal/conversions/property_conversions.go
```diff
@@ -1242,49 +1187,145 @@ func assignPrimitiveFromEnum(
 	}, nil
 }
 
-// assignObjectFromObject will generate a conversion if both properties are TypeNames
-// referencing ObjectType definitions and neither property is optional
-//
-// For ConvertFrom:
+// assignObjectDirectlyFromObject will generate a conversion if both properties are TypeNames referencing ObjectType
+//// definitions, neither property is optional, and the types are adjacent in our storage conversion graph.
```
matthchr, line 1191:
```suggestion
// definitions, neither property is optional, and the types are adjacent in our storage conversion graph.
```

matthchr (approved):

v2/tools/generator/internal/conversions/property_conversions.go
```diff
@@ -262,7 +264,7 @@ func assignToOptional(
 	conversionContext *PropertyConversionContext) (PropertyConversion, error) {
 
 	// Require destination to not be a bag item
-	if _, destinationIsBagItem := AsPropertyBagMemberType(destinationEndpoint.Type()); destinationIsBagItem {
+	if destinationEndpoint.IsBagItem() {
```
babbageclunk, line 267: These predicates are nice!

v2/tools/generator/internal/functions/testdata/TestGolden_PropertyAssignmentFunction_WhenSharedObjectVersion/SharedObject.golden
```diff
@@ -0,0 +1,54 @@
...
+	err := locationStash.AssignPropertiesFromLocation(&source.Residence)
+	if err != nil {
+		return errors.Wrap(err, "calling AssignPropertiesFromLocation() to populate field LocationStash from Residence")
+	}
+	var residence Location
+	err = residence.AssignPropertiesFromLocation(&locationStash)
+	if err != nil {
+		return errors.Wrap(err, "calling AssignPropertiesFromLocation() to populate field Residence from LocationStash")
+	}
+	person.Residence = residence
```
babbageclunk, line 30: Is it worth making this scenario have more than one intermediate version to check that the conversion generation can deal with that?

babbageclunk (approved):
Looks great, some minor suggestions.

v2/tools/generator/internal/conversions/property_conversions.go
theunrepentantgeek: I don't know that we can make these much smaller - we'd still need to call the function and then return `nil, nil` if it returned true, and possibly at the risk of obscuring which prerequisites are checked.  Moving the older checks into the `IsBagItem()` and `IsOptional()` predicates seemed like a good middle path.

v2/tools/generator/internal/astmodel/std_properties.go
```diff
@@ -18,3 +18,8 @@ const (
 	OperatorSpecSecretsProperty = "Secrets"
 	ConditionsProperty          = "Conditions"
 )
+
+// IsKubernetesResourceProperty returns true if the supplied property name is one of our "magical" names
+func IsKubernetesResourceProperty(name PropertyName) bool {
```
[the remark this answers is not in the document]
theunrepentantgeek: Comment updated.

v2/tools/generator/internal/codegen/pipeline/inject_property_assignment_functions.go
```diff
@@ -109,14 +109,18 @@ func (f propertyAssignmentFunctionsFactory) injectBetween(
...
+		WithConfiguration(f.configuration.ObjectModelConfiguration).
+		WithConversionGraph(f.graph)
 	assignFromFn, err := functions.NewPropertyAssignmentFunction(upstreamDef, downstreamDef, assignFromContext, conversions.ConvertFrom)
 	upstreamName := upstreamDef.Name()
 	if err != nil {
 		return astmodel.TypeDefinition{}, errors.Wrapf(err, "creating AssignFrom() function for %q", upstreamName)
 	}
 
-	assignToContext := conversions.NewPropertyConversionContext(f.definitions, f.idFactory, f.configuration.ObjectModelConfiguration)
+	assignToContext := conversions.NewPropertyConversionContext(f.definitions, f.idFactory).
```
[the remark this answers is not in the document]
theunrepentantgeek: They used to be quite different but have converged over time; will change to reuse a single context as that's clearer.

v2/tools/generator/internal/conversions/property_conversion_context.go
```diff
@@ -112,5 +157,6 @@ func (c *PropertyConversionContext) clone() *PropertyConversionContext {
 		propertyBagName: c.propertyBagName,
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

v2/tools/generator/internal/functions/testdata/TestGolden_PropertyAssignmentFunction_WhenSharedObjectVersion/SharedObject.golden
theunrepentantgeek: Good idea. Done.

PR #2205, opened 2022-03-31
Include error in panic

theunrepentantgeek:

Includes the underlying error in the panic when waiting for the webhook server times out.

Also increases the timeout to 15s - given this is failing on a PR that doubles the number of resource versions we have, it's a reasonable hypothesis that it's failing due to workload.

gif

### DIFF
--- a/v2/internal/testcommon/kube_test_context_envtest.go
+++ b/v2/internal/testcommon/kube_test_context_envtest.go
@@ -198,7 +198,7 @@ func createSharedEnvTest(cfg testConfig, namespaceResources *namespaceResources)
 
 			if time.Now().After(timeoutAt) {
 				err = errors.Wrap(err, "timed out waiting for webhook server to start")
-				panic(err)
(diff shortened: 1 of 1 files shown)


v2/internal/testcommon/kube_test_context_envtest.go
```diff
@@ -189,19 +189,21 @@ func createSharedEnvTest(cfg testConfig, namespaceResources *namespaceResources)
...
 		for {
 			err = chk(nil)
 			if err == nil {
 				break
 			}
 
 			if time.Now().After(timeoutAt) {
-				panic("timed out waiting for webhook server to start")
+				err = errors.Wrap(err, "timed out waiting for webhook server to start")
+				panic(err)
```
babbageclunk, line 201: I think just reporting the error message rather than the full dump of the error struct would be better here.
```suggestion
				panic(err.Error())
```

babbageclunk (approved):

PR #2208, opened 2022-04-01
Add helm test installation workflow and tasks

### LINKED ISSUE #1772: Add a test that installs a newly-built helm chart
The current helm-install test installs the built tgz chart. This is useful for cases where someone has forgotten to run `make helm-chart-manifests` after changing the chart files (I did this yesterday), but it means that if there's an error in one of the templates in the chart they'll only be found when someone updates the helm chart after a release.

**Describe the improvement**
We need another test that builds and installs the chart from the current chart files so that errors in templates can be found in that PR, rather than potentially much later.

super-harsh:
Closes #1772 

This PR includes: 
- Workflow for testing the helm chart installation, which gets triggered on a pull-request.
- Tasks to install, delete and test helm chart.


.github/workflows/helm-ci-test.yml
```diff
@@ -0,0 +1,101 @@
...
+          # however, since we require branches to always be up-to-date (GitHub setting),
+          # using the head of the PR branch should provide equivalent behaviour
+          sha='${{ github.event.pull_request.head.sha }}'
+          repo='${{ github.event.pull_request.head.repo.html_url}}'
+          repo=${repo#"https://"} # trim prefix
+
+          if [ "$repo" != "github.com/Azure/azure-service-operator" ]; then
+            echo "Skipping 'go get' check for fork…" 
+            exit 0
+          fi
```
theunrepentantgeek, line 30: Under which circumstances would this skip the check? 

That is, when would the workflow be running and the branch NOT be ours?
(Suggest adding a comment to explain)

theunrepentantgeek (approved):

Taskfile.yml
```diff
@@ -407,6 +408,33 @@ tasks:
...
+    cmds:
+    - "helm install --set azureSubscriptionID=$AZURE_SUBSCRIPTION_ID \
+      --set azureTenantID=$AZURE_TENANT_ID \
+      --set azureClientSecret=$AZURE_CLIENT_SECRET \
+      --set azureClientID=$AZURE_CLIENT_ID \
+      --set image.repository={{.LOCAL_REGISTRY_CONTROLLER_DOCKER_IMAGE}}
+      asov2 -n {{.ASO_NAMESPACE}} --create-namespace ./charts/azure-service-operator/"
+    - "kubectl rollout status -n {{.ASO_NAMESPACE}} deployment/azureserviceoperator-controller-manager"
+
+
```
matthchr, line 423: minor: There's an extra newline here that doesn't need to be.

Taskfile.yml
```diff
@@ -407,6 +408,33 @@ tasks:
...
+    cmds:
+      - "helm delete asov2 -n {{.ASO_NAMESPACE}}"
+
+  controller:test-helm-installation:
+    desc: Test helm chart installation and deletion
+    cmds:
+      - task: controller:install-cert-manager
+      - task: controller:build
+      - task: controller:docker-push-local
+      - task: controller:gen-helm-manifest
```
matthchr, line 435: After this you should put 
```
      # We need the below to wait until cert-manager is up, otherwise the subsequent installation of webhooks fails. See https://cert-manager.io/next-docs/installation/verify/
      - "cmctl check api --wait=2m"
```

See the `kind-create-with-podidentity` target

Taskfile.yml
```diff
@@ -407,6 +408,33 @@ tasks:
...
+    desc: Generate and install helm chart on cluster
+    dir: "{{.CONTROLLER_ROOT}}"
+    cmds:
+    - "helm install --set azureSubscriptionID=$AZURE_SUBSCRIPTION_ID \
+      --set azureTenantID=$AZURE_TENANT_ID \
+      --set azureClientSecret=$AZURE_CLIENT_SECRET \
+      --set azureClientID=$AZURE_CLIENT_ID \
+      --set image.repository={{.LOCAL_REGISTRY_CONTROLLER_DOCKER_IMAGE}}
+      asov2 -n {{.ASO_NAMESPACE}} --create-namespace ./charts/azure-service-operator/"
+    - "kubectl rollout status -n {{.ASO_NAMESPACE}} deployment/azureserviceoperator-controller-manager"
```
matthchr, line 421: Does this fail if the rollout fails?
Do we also want to do
`- "kubectl wait --for=condition=ready --timeout=2m pod -n azureserviceoperator-system -l control-plane=controller-manager"` to make sure that things come up?

Taskfile.yml
```diff
@@ -407,6 +408,33 @@ tasks:
...
 
+  controller:install-helm:
+    desc: Generate and install helm chart on cluster
+    dir: "{{.CONTROLLER_ROOT}}"
+    cmds:
+    - "helm install --set azureSubscriptionID=$AZURE_SUBSCRIPTION_ID \
+      --set azureTenantID=$AZURE_TENANT_ID \
+      --set azureClientSecret=$AZURE_CLIENT_SECRET \
+      --set azureClientID=$AZURE_CLIENT_ID \
+      --set image.repository={{.LOCAL_REGISTRY_CONTROLLER_DOCKER_IMAGE}}
```
matthchr, line 419: Is this line missing a `\`?

.github/workflows/helm-ci-test.yml
```diff
@@ -0,0 +1,101 @@
...
+          DOCKER_BUILDKIT: 1
+
+      - name: Run devcontainer image
+        id: devcontainer
+        run: |
+          container_id=$(docker create -w /workspace -v $GITHUB_WORKSPACE:/workspace -v /var/run/docker.sock:/var/run/docker.sock --network=host devcontainer:latest)
+          docker start "$container_id"
+          echo "::set-output name=id::$container_id"
+
+      - name: Spin up a kind cluster
```
matthchr, line 87: I think you should combine these two steps (start `kind` + `test helm` into a single command in the Taskfile so that it's a one-liner to do this locally as well.

See for example what we do with `controller:test-integration-kind`.

.github/workflows/helm-ci-test.yml
```diff
@@ -0,0 +1,101 @@
...
+        run: |
+          container_id=${{steps.devcontainer.outputs.id}}
+          docker exec "$container_id" task controller:kind-create
+
+      - name: Test Helm Installation
```
matthchr, line 92: This isn't going to work for forks, unless you set up the integration to make it work for forks.

I am wondering if it makes more sense to, instead of having a separate CI job for Helm, to just use the `test-integration-kind` job, which I've already set up to work for fork PRs. You could do this by creating a new `controller:kind-create-helm-podidentity`, and then substitute that into the `controller:test-integration-kind` and `controller:test-integration-kind-ci` targets

.github/workflows/helm-ci-test.yml
```diff
@@ -0,0 +1,101 @@
+name: Test helm installation
+on:
+  push:
+    branches:
+      - main
```
matthchr, line 5: I don't think we need trigger on push to main (especially if we combine this with the existing `kind` test as I mention below.

matthchr (changes requested):

.github/workflows/helm-ci-test.yml
matthchr: Then we'd have 3 sets of validation we do (rather than 4 as you've proposed):

1. envtest with recordings. Needs no Azure secrets and runs pretty fast.
2. kind cluster with Helm chart installed on it (using the latest master container image and YAMLs)
3. kind cluster in multitenant mode (installed manually)

Otherwise if we do not combine things, we'd have:

item 2 == kind cluster with latest operator installed manually
item 4 == kind cluster with latest Helm chart

and I'm not sure there's a lot of practical difference between these... especially because we already test installation of the standalone operator in the multitenant deployment as well (just added by @babbageclunk)

super-harsh: Yes, sounds very clean. I've made changes according to the suggestion.

Taskfile.yml
super-harsh: Yes it does fail when the rollout fails. Also we can add wait on this which I missed here. 
However, I think its better to do it the way you suggested by waiting for the pod to be ready.. As deployment "rolled out" is not necessarily without errors.

Taskfile.yml
```diff
@@ -456,6 +457,41 @@ tasks:
     cmds:
      - "{{.SCRIPTS_ROOT}}/generate-helm-manifest.sh {{.KUBE_RBAC_PROXY}} {{.LOCAL_REGISTRY_CONTROLLER_DOCKER_IMAGE}} {{.PUBLIC_REGISTRY}} {{.LATEST_VERSION_TAG}} `pwd`/{{.CONTROLLER_ROOT}}"
 
+  controller:install-helm:
```
matthchr, line 460: You can (and should) remove this line: `--set azureClientSecret=$AZURE_CLIENT_SECRET \`. It's not needed because later you run `controller:make-aadpodidentity-secret` which makes the secret simulating managed identity. If you don't delete this line we might have access to both secrets and I am not sure what happens then. We want to make sure we're using the AAD Pod Identity path. Deleting the `azureClientSecret` should still allow everything to pass.

matthchr (commented):
Had one more comment, then it looks good.

matthchr:
Hmm, this might be a problem: https://github.com/helm/helm/issues/1996

> Error: INSTALLATION FAILED: create: failed to create: Secret "sh.helm.release.v1.asov2.v1" is invalid: data: Too long: must have at most 1048576 bytes

.github/workflows/helm-ci-test.yml
super-harsh: This was to prevent check for forks. I've removed this one and as suggested by @matthchr, and have done it the other way to replace `controller:test-integration-kind` and `controller:test-integration-kind-ci` targets. Which resulted in not having any seperate workflow.

Taskfile.yml
```diff
@@ -290,7 +291,7 @@ tasks:
   controller:test-integration-kind-ci:
     desc: Run live integration tests in kind and deletes the kind cluster afterwards.
     dir: "{{.CONTROLLER_ROOT}}"
-    deps: [controller:kind-create-with-podidentity]
+    deps: [controller:kind-create-helm-podidentity]
```
matthchr, line 294: Minor: it might not hurt to take a page from @babbageclunk's book and stick a `-v` on the `go test -timeout 15m` line down below, to make it: `go test -timeout 15m -run -v ...`

matthchr: No need to do that in this PR though necessarily. Feel free to merge this!

matthchr (approved):
LGTM.

PR #2210, opened 2022-04-02
Create README for Property Conversions

theunrepentantgeek:

The recursive nature of the property conversions is powerful, flexible - and hard to initially understand.

I thought a walkthrough document that gets a reader up to speed would be a useful thing to have.

gif

**If applicable**:


v2/tools/generator/internal/conversions/README.md
```diff
@@ -0,0 +1,568 @@
...
+
+* `List` is the final list we're creating
+* `Item` is a single item from the source list
+* `Index` is the integer index of the item we're converting
+
+Clones of `knownLocals` are used for each nested scope (one inside our **if** statement, see below; the other inside our loop) so that definitions don't leak out.
+
+Our declaration initializes a new list of the correct final size.
+
+``` go
```
matthchr, line 464: FWIW I think the bits that are the hardest to understand are the recursive calls and how the endpoints are crafted/work with that process. (Including `knownLocals`, etc). The actual generation of the for loop code once you have the inner conversion isn't all that complicated, so I feel like if you wanted to shorten this document you could remove some of the specificity here and focus on the content you have up above.

It's pretty minor though and you've also already written this so it's probably more work to rejigger it at this point.

matthchr (approved):

v2/tools/generator/internal/conversions/README.md
```diff
@@ -0,0 +1,568 @@
...
+        return nil, nil
+    }
+
+    // Require both source and destination to be non-optional
+    if sourceEndpoint.IsOptional() || destinationEndpoint.IsOptional() {
+        return nil, nil
+    }
+```
+
+Neither of our endpoints is a bag item, and neither is optional, so our sample data passes these checks and the handler continues.
```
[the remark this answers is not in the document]
theunrepentantgeek: Good idea. Noted for a followup PR.

v2/tools/generator/internal/conversions/README.md
theunrepentantgeek: Another good idea - also noted for a followup PR.

v2/tools/generator/internal/conversions/README.md
```diff
@@ -0,0 +1,568 @@
...
+    destinationEndpoint *TypedConversionEndpoint,
+    conversionContext *PropertyConversionContext) (PropertyConversion, error)
+```
+
+* `sourceEndpoint` captures the name and type of the source value that needs to be transferred;
+* `destinationEndpoint` similarly captures the name and type of the destination that needs to be populated;
+* `conversionContext` contains other information about the conversion being generated.
+
+
+The core algorithm works by scanning through the handlers listed in [`propertyConversionFactories`](https://github.com/Azure/azure-service-operator/blob/main/v2/tools/generator/internal/conversions/property_conversions.go#L51) to find one that can handle the current conversion. Many of those conversions recursively call `CreateTypeConversion` with a simpler conversion, breaking the problem down into simpler steps.
```
super-harsh, line 21: `propertyConversionFactories` is the link pointing to the right line? Should it be line 57?

v2/tools/generator/internal/conversions/README.md
```diff
@@ -0,0 +1,568 @@
...
+                    &dst.IndexExpr{
+                        X:     dst.NewIdent(tempId),
+                        Index: dst.NewIdent(indexId),
+                    },
+                    expr),
+            }
+        }
+```
+
+Within our loop we create a copy of the item variable, to avoid aliasing between loops. We know for our sample case that an alias isn't necessary - but we don't know generally know what an conversions might do, so this is a safety precaution.
```
super-harsh, line 487:
```suggestion
Within our loop we create a copy of the item variable, to avoid aliasing between loops. We know for our sample case that an alias isn't necessary - but we don't know generally know what conversions might do, so this is a safety precaution.
```

v2/tools/generator/internal/conversions/README.md
```diff
@@ -0,0 +1,568 @@
...
+        knownLocals *astmodel.KnownLocalsSet, 
+        generationContext *astmodel.CodeGenerationContext) []dst.Stmt {
+        return writer(reader)
+    }, nil
+}
+```
+
+* `reader` is an expression giving access to the original value
+* `writer` is a function takes an expression and writes that to the final destination
+* `knownLocals` is a set that contains all local variables declared so far (not used in this handler)
```
super-harsh, line 123: This is a bit unclear as this puts a question in my mind regarding - local variables declared so far.. where? how?

super-harsh (approved):
Legendary document! Loved how the doc is walking through the code and explaining the process step by step. Definitely a very helpful doc.

v2/tools/generator/internal/conversions/README.md
theunrepentantgeek: I've clarified.

PR #2213, opened 2022-04-03
Avoid metrics panic if HTTP request returns an error

theunrepentantgeek:

We'll currently experience a panic if an HTTP request returns an error as the response will be **nil**. 

Fixed by moving the metrics line after the `err` check, at which point we expect to have a response.

gif

### DIFF
--- a/v2/internal/genericarmclient/generic_client.go
+++ b/v2/internal/genericarmclient/generic_client.go
@@ -140,16 +140,21 @@ func (client *GenericClient) createOrUpdateByID(
 
 	client.metrics.RecordAzureRequestsTime(resourceType, time.Since(requestStartTime), metrics.HttpPut)
 
+	statusCode := 0
+	if resp != nil {
+		statusCode = resp.StatusCode
+	}
+	client.metrics.RecordAzureRequestsTotal(resourceType, statusCode, metrics.HttpPut)
+
 	if err != nil {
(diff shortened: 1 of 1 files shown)


v2/internal/genericarmclient/generic_client.go
```diff
@@ -138,12 +139,14 @@ func (client *GenericClient) createOrUpdateByID(
...
 
 	client.metrics.RecordAzureRequestsTime(resourceType, time.Since(requestStartTime), metrics.HttpPut)
-	client.metrics.RecordAzureRequestsTotal(resourceType, resp.StatusCode, metrics.HttpPut)
 
 	if err != nil {
 		client.metrics.RecordAzureFailedRequestsTotal(resourceType, metrics.HttpPut)
 		return nil, err
 	}
+
+	client.metrics.RecordAzureRequestsTotal(resourceType, resp.StatusCode, metrics.HttpPut)
```
super-harsh, line 148: Might be worth updating the comment as well for the method inside `arm_client_metrics.go`. With this change it will not longer be `total(failed + successful)` and just the successful requests to the ARM server

super-harsh (approved):
Minor: Might be worth updating the comment as well for the method inside arm_client_metrics.go. With this change it will not longer be total(failed + successful) and just the successful requests to the ARM server.

matthchr (approved):

PR #2224, opened 2022-04-08
Improve aliasing used for imports

### LINKED ISSUE #2147: Improve names used for imports

When selecting the import names used for generated code, our current approach gives names that are not ideal:

``` go
import (
	storage "github.com/Azure/azure-service-operator/v2/api/storage/v1alpha1api20210401"
	"github.com/Azure/azure-service-operator/v2/api/storage/v1alpha1api20210401storage"
)
```

* Not clear why the first import gets named when there's no conflict.
* Looks as though the first import is somehow special (spoiler: it's not).
* When these names are consumed, it's not clear which version is which (while some imports use the version, others don't)

**Describe the improvement**

We should proactively select names that reflect the version being imported, keeping only the group (and the generator prefix) where required to resolve ambiguity.

``` go
import (
	storage20210401 "github.com/Azure/azure-service-operator/v2/api/storage/v1alpha1api20210401"
	storage20210401storage "github.com/Azure/azure-service-operator/v2/api/storage/v1alpha1api20210401storage"
)
```

For clarity: the prefix `storage` reflects _Azure.Storage_ - the group of the resource; the suffix `storage` refers to the persistence type synthesized for version compatibility.

theunrepentantgeek:

Our import naming is a bit messy, with some names hard to read. 

This PR improves things by applying one of three different rules, as follows.

If all imports come from the same group, use a simplified form of the version:
``` go
v20210301  "github.com/Azure/azure-service-operator/v2/api/cache/v1beta20210301"
v20210301s "github.com/Azure/azure-service-operator/v2/api/cache/v1beta20210301storage"
```

If an import is the only one from a specific group, use just the group:
``` go 
cache "github.com/Azure/azure-service-operator/v2/api/cache/v1alpha1api20210301"
```

If there are imports of multiple groups and multiple versions, use both group and version:
``` go
cache_customizations "github.com/Azure/azure-service-operator/v2/api/cache/customizations"
cache_v20201201      "github.com/Azure/azure-service-operator/v2/api/cache/v1beta20201201"
cache_v20201201s     "github.com/Azure/azure-service-operator/v2/api/cache/v1beta20201201storage"
cache_v20210301      "github.com/Azure/azure-service-operator/v2/api/cache/v1beta20210301"
cache_v20210301s     "github.com/Azure/azure-service-operator/v2/api/cache/v1beta20210301storage"
```

Imports of compatibility packages swap out a different prefix for the `v` of each version:

``` go
cache_alpha20201201  "github.com/Azure/azure-service-operator/v2/api/cache/v1alpha1api20201201"
cache_alpha20201201s "github.com/Azure/azure-service-operator/v2/api/cache/v1alpha1api20201201storage"
cache_alpha20210301  "github.com/Azure/azure-service-operator/v2/api/cache/v1alpha1api20210301"
cache_alpha20210301s "github.com/Azure/azure-service-operator/v2/api/cache/v1alpha1api20210301storage"
```

Closes #2147 

I know using an `_` in the import alias is not very go-ish, but it does dramatically improve readability when there are a large number of imports - such as in `controller_resources_gen.go` - so I think the style violation is worth it.

Changes to the generated code are in a separate PR (see PR #2225) which will need to be merged in here to make the build pass.

gif

**If applicable**:
- [x] this PR contains tests


v2/tools/generator/internal/astmodel/package_import_set.go
```diff
@@ -9,15 +9,12 @@ import (
...
-	"k8s.io/klog/v2"
-
-	kerrors "k8s.io/apimachinery/pkg/util/errors"
 )
 
 // PackageImportSet represents a set of distinct PackageImport references
 type PackageImportSet struct {
-	imports map[PackageReference]PackageImport
+	imports         map[PackageReference]PackageImport
+	aliasesAssigned bool
```
matthchr, line 17: minor: Maybe leave a comment here explaining that this is for optimization purposes? (I think it is?).

v2/tools/generator/internal/astmodel/local_package_reference.go
```diff
@@ -109,6 +109,20 @@ func (pr LocalPackageReference) ApiVersion() string {
 	return pr.apiVersion
 }
 
+// CreateImportAlias creates a custom alias for importing this reference
+func (pr LocalPackageReference) CreateImportAlias(style PackageImportStyle) string {
```
matthchr, line 113: It's a bit weird that this is a function on `LocalPackageReference`, because all of the `PackageReference`'s are really just the _reference_ part of things, not the name (the name is `PackageImport`). In many contexts that a `PackageReference` is used (such as for describing the object graph), this method doesn't really make sense.

Would it be cleaner to instead put this on `PackageImport`? I suppose that'd also involve moving the helper methods for `SimplifiedXyz`, but I think they could be written in terms of the Getters on this type anyway.

Only suggesting this because it feels like the name generation/simplification logic should be on the type with the name (`PackageImport`) and not on the type that's just the "link"

v2/tools/generator/internal/astmodel/package_import.go
```diff
@@ -101,3 +101,9 @@ func (pi PackageImport) VersionedNameForImport() string {
 	service := pi.ServiceNameForImport()
 	return service + pi.packageReference.PackageName()
 }
+
+// WithImportAlias creates a copy of this import with a name following the specified rules
+func (pi PackageImport) WithImportAlias(style PackageImportStyle) PackageImport {
```
matthchr, line 106: Do we need the `ServiceNameForImport` and `VersionedNameForImport` above? It feels like they are trying to do something very similar to this?

v2/tools/generator/internal/astmodel/package_reference.go
```diff
@@ -31,6 +31,8 @@ type PackageReference interface {
 	// GroupVersion returns the group and version of this reference.
 	// Returns true if the reference has a group and version, false otherwise.
 	GroupVersion() (string, string, bool)
+	// CreateImportAlias creates a custom alias for importing this reference
+	CreateImportAlias(style PackageImportStyle) string
```
matthchr, line 35: See comment above about question if this makes more sense on `PackageImport` rather than `PackageReference`

v2/tools/generator/internal/astmodel/package_import_set.go
theunrepentantgeek: Added comment.

v2/tools/generator/internal/astmodel/local_package_reference.go
theunrepentantgeek: I was 50/50 on putting the logic there originally - it was largely to leverage polymorphism, but you've convinced me to move them.

v2/tools/generator/internal/astmodel/package_import.go
theunrepentantgeek: Oh, thanks - I meant to remove those but neglected to. Will remove them now.

matthchr (approved):

PR #2229, opened 2022-04-11
Upgrade v2 deps

Porges:
Corrected version of #2217 to actually update v2 go.mod.


v2/go.mod
```diff
@@ -11,69 +11,79 @@ require (
...
+	github.com/fsnotify/fsnotify v1.5.1 // indirect
+	github.com/go-openapi/jsonpointer v0.19.5 // indirect
+	github.com/go-openapi/jsonreference v0.19.5 // indirect
+	github.com/go-openapi/swag v0.19.14 // indirect
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/golang-jwt/jwt v3.2.2+incompatible // indirect
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
 	github.com/golang/protobuf v1.5.2 // indirect
-	github.com/google/gofuzz v1.1.0 // indirect
-	github.com/googleapis/gnostic v0.5.5 // indirect
```
Porges, line 50: had to go to beta versions of k8s.io deps to get rid of this problematic dependency

matthchr: Should we leave a comment to that effect up by the beta dependencies?

v2/internal/controllers/generic_controller.go
```diff
@@ -142,8 +142,8 @@ func register(
 	clientFactory arm.ARMClientFactory,
 	info *registration.StorageType,
 	extensions map[schema.GroupVersionKind]genruntime.ResourceExtension,
-	options Options) error {
-
+	options Options,
```
matthchr, line 145: Is this the normal style we use? For some reason I thought we left `error` on the same line as the last line. Was this gofmt "fixing" things or you?

matthchr (approved):

v2/go.mod
Porges: I get the feeling that `go mod` would remove it

v2/internal/controllers/generic_controller.go
Porges: yes, this was "fixed" 🙂

Issue #2227, opened 2022-04-11
Bug: Helm chart - data: Too long: must have at most 1048576 bytes


super-harsh:
**Describe the current behaviour**

Since, we have a chunky collection of CRDs (which can increase in the future as well). We are running into the helm [issue](https://github.com/helm/helm/issues/9788) of `invalid: data: Too long: must have at most 1048576 bytes`. 

**Describe the improvement**

We can improve/avoid this by having two separate charts for `azure-service-operator` and `azure-service-operator-resources` and use the later one in `azure-service-operator` chart as a dependency. Which would then allow us with flexibility of the following: 

- Allow end-user to control if they want to install CRDs or not
- Different namespace installation
- Removing the CRDs with `helm delete` on the parent chart
- Allow us(in the future) to support more resources without getting blocked

PR #2230, opened 2022-04-12
Split helm charts

### LINKED ISSUE #2190: Bug: namspace some how harcoded
ASO V2: mcr.microsoft.com/k8s/azureserviceoperator:v2.0.0-alpha.6

**Describe the bug**
I deployed ASO in a different namespace, when trying to apply a Storage Account it fails with the error below. It works for resource group resources, but not for storage account. It is trying to call the webhook from the namespace: `azureserviceoperator-system`

PS I deployed ASO with helm.
PS2: if i change the namespace of ASO to `azureserviceoperator-system` it works

**To Reproduce**
```
$ cat sa.yml
apiVersion: resources.azure.com/v1alpha1api20200601
kind: ResourceGroup
metadata:
  name: mth-aso-sa-test
  namespace: test
spec:
  location: westeurope
---
apiVersion: storage.azure.com/v1alpha1api20210401
kind: StorageAccount
metadata:
  name: mthasosatest
  namespace: test
spec:
  location: westeurope
  owner:
    name: mth-aso-sa-test
  kind: StorageV2
  sku:
    name: Standard_LRS

$ k apply -f sa.yaml
```

**Expected behavior**
The namespace should not be hardcoded in the code or config

**Error**
```
Error from server: error when creating "examples/02-storage-account.yaml": conversion webhook for storage.azure.com/v1alpha1api20210401, Kind=StorageAccount failed: Post "https://azureserviceoperator-webhook-service.azureserviceoperator-system.svc:443/convert?timeout=30s": service "azureserviceoperator-webhook-service" not found
```

super-harsh:
Closes #2190 & #2227 

Since we are facing problems with the helm chart containing both CRDs and templates, with this PR we can have more flexibility around adding more resources, handling CRDs and having custom namespaces.

Changes with this PR:
- `Azure-service-operator` chart doesn't anymore have crds, however is replaced with a dependency chart for CRDS
- Add new parameter `installCRDs` into `Azure-service-operator` values.yaml to control CRD installation
- A new chart for CRDs called `azure-service-operator-resources` 
- Modify the `generate-helm-manifest.sh` to keep the changes consistent and package both charts
- New `.tgz` packages for both charts


v2/charts/azure-service-operator/Chart.yaml
```diff
@@ -5,3 +5,7 @@ description: Deploy components and dependencies of azure-service-operator
 home: https://github.com/Azure/azure-service-operator/v2
 sources:
   - https://github.com/Azure/azure-service-operator/v2
+dependencies:
+  - name: azure-service-operator-resources
+    version: v2.0.0-alpha.6
+    condition: installCRDs
```
theunrepentantgeek, line 11: Missing end of line at end of file.

theunrepentantgeek (commented):
Looks good - as far as I understand what's going on. Going to leave approval for someone else though.

matthchr (commented):
You mentioned that this should resolve #2190, but I don't actually see how it does. The CRDs will still have the webhook entry:

```
    webhook:
      clientConfig:
        service:
          name: azureserviceoperator-webhook-service
          namespace: azureserviceoperator-system
          path: /convert
```

If we allow users to configure the namespace of the deployment (which includes the pod serving the webhook for conversion), then that path is wrong. So I don't think that cross-namespace installation works even here?

I think, given Helms limitations in this space, we should just remove the namespace parameter as it doesn't seem like it's going to work?

v2/charts/index.yaml
```diff
@@ -2,14 +2,37 @@ apiVersion: v1
 entries:
   azure-service-operator:
   - apiVersion: v2
-    created: "2022-03-03T16:41:18.238942+13:00"
+    created: "2022-04-13T10:31:56.855826+12:00"
+    dependencies:
+    - condition: installCRDs
```
matthchr, line 7: minor: Consider rebasing and rerunning to instead update (add) a beta chart?

matthchr (approved):
Approved, but I think rebase and run against latest `main` so that you generate beta instead of another alpha.6?

PR #2245, opened 2022-04-21
Make theme wider

theunrepentantgeek:

Our current docs site theme ([book](https://github.com/alex-shpak/hugo-book)) is a bit narrow for our purposes, with many of our tables ending up awkwardly cropped or squashed:

image
image

By making the central column wider, we can avoid most issues:

image
image

Our theme comes from a submodule, so we can't tweak it directly; I ended up creating the required change as a separate file and copying it into place via the Taskfile.

gif

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -70,6 +70,7 @@ tasks:
   build-docs-site:
     dir: docs/hugo
     cmds:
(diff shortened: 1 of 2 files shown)


theunrepentantgeek:
This approach doesn't work. Will try something else.

PR #2246, opened 2022-04-24
Add Frontmatter for Hugo

theunrepentantgeek:

Adds frontmatter to many files to improve our documentation site

gif

**If applicable**:


super-harsh (approved):
Love this!

Taskfile.yml
```diff
@@ -693,6 +693,13 @@ tasks:
     cmds:
       - "{{.SCRIPTS_ROOT}}/check_headers.py"
 
+  frontmatter-check:
+    desc: Ensure all Markdown documentation has frontmatter.
+    dir: docs/hugo/content
+    cmds:
+      - echo "Foo"
```
super-harsh, line 700: Do we need this here?

docs/hugo/content/contributing/create-a-new-release.md
```diff
@@ -1,3 +1,6 @@
+---
+title: Creating a new release of ASO v2
+---
 # Creating a new release of ASO v2
```
Porges, line 4: I haven't looked at the output yet but it might be that the title will be appearing in two locations on the page?

Porges: … or maybe we should have the template do that already if it doesn’t, to make the source files easier to manage.

Porges (approved):

scripts/check_frontmatter.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
+
+import itertools
+import os
+import re
+import sys
+
+
+number_of_lines = 8
```
super-harsh, line 9: Maybe a comment here would be useful?

theunrepentantgeek: Added.

docs/hugo/content/contributing/create-a-new-release.md
theunrepentantgeek: Current theme (Book) leaves the page to do it's own thing - the FrontMatter is only used for sidebar/etc.

It seems more common to do things the other way around - Docsy injects the title at the top of the page, and has a separate FrontMatter item for the name in the sidebar.

Taskfile.yml
theunrepentantgeek: Whoops! Removed.

PR #2254, opened 2022-04-27
Liveness probe

### LINKED ISSUE #1201: Task: Add liveness/readiness probes
The initial health checking probe was removed in #1198 to avoid issues where the Kubelet could not connect to the health endpoint on some clusters. The initial check was only doing a naive check anyway. We should regroup and investigate what these endpoints should actually do in order to ensure the operator is functioning and the reintroduce the code changes.

super-harsh:
Closes #1201 

This PR adds liveness probe for ASOv2 which makes sure ASO is started successfully, then only pod shows `Ready` state or else goes into `Error` state.


v2/config/manager/manager_auth_proxy_patch.yaml
```diff
@@ -22,5 +22,6 @@ spec:
       - name: manager
         args:
         - "--metrics-addr=127.0.0.1:8080"
+        - "--health-addr=:8081"
```
matthchr, line 25: Does this addr need to be exposed in the service too? I assume no but just checking

matthchr (approved):

v2/config/manager/manager_auth_proxy_patch.yaml
super-harsh: Nope! it gets reached through the container port

theunrepentantgeek (commented):
Looks good, but one question: How long does it wait for the health probe to succeed before signaling an error?

theunrepentantgeek:
Those defaults seem reasonable. 😸

PR #2255, opened 2022-04-27
Add pod annotationfor ASOv2 helm chart

### LINKED ISSUE #1367: Feature: Add pod annotations to aso helm charts
<Describe the feature request>

- the current helm charts don't support the ability to add pod annotations.
- ex: pod annotations are useful where we want our aso creds to be accessed from vault etc.

super-harsh:
Closes #1367 

This PR adds support for having pod annotations for ASOv2 helm chart. 

Didn't include new helm `.tgz` bundle in this PR, considering including it in the beta.1 release helm chart.

### DIFF
--- a/v2/charts/azure-service-operator/values.yaml
+++ b/v2/charts/azure-service-operator/values.yaml
@@ -53,3 +53,6 @@ metrics:
 
 # installCRDs determines if the CRDs subchart should be installed.
 installCRDs: true
+
+#podAnnotations contain the pod annotations for Azure Service Operator
+podAnnotations: {}
--- a/v2/charts/azure-service-operator/values.yaml
+++ b/v2/charts/azure-service-operator/values.yaml
@@ -54,5 +54,5 @@ metrics:
 # installCRDs determines if the CRDs subchart should be installed.
 installCRDs: true
 
-#podAnnotations contain the pod annotations for Azure Service Operator
+# podAnnotations contain the pod annotations for Azure Service Operator
 podAnnotations: {}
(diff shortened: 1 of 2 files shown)


matthchr (approved):
LGTM. Agree not to include a new chart with this PR. We can make an updated chart for beta.1

v2/charts/azure-service-operator/values.yaml
```diff
@@ -53,3 +53,6 @@ metrics:
 
 # installCRDs determines if the CRDs subchart should be installed.
 installCRDs: true
+
+#podAnnotations contain the pod annotations for Azure Service Operator
```
matthchr, line 57: minor: 
```suggestion
# podAnnotations contain the pod annotations for Azure Service Operator
```

OleksandrBrodskiy:
Hi , seems like in this change you have an error with inserting data, ` {{ toYaml .Values.podAnnotations | indent 6 }}` , indent not needed here. Now it works like this : 
```
  template:
    metadata:
      annotations:
              testAnnotation: test
        kubectl.kubernetes.io/default-container: manager
```

matthchr:
Thanks for the heads up @OleksandrBrodskiy, @super-harsh could you help take a look?

super-harsh:
@OleksandrBrodskiy Thanks for raising this up. I'm getting a similar result, will send a PR shortly with the fix.

OleksandrBrodskiy:
Thanks for the quick fix, is it possible to update the chart with this change?

OleksandrBrodskiy:
And one more thing, for me it works only with this configuration :
```
        {{- if .Values.podAnnotations }}
        {{- toYaml .Values.podAnnotations| nindent 8 -}}
        {{- end }}
```
without indent, my annotation was also misplaced in the file, values example: 
```
podAnnotations:
  test.annotation: |
        [{
          "key": "value"
        }]
```

super-harsh:
@OleksandrBrodskiy The change would be in the `beta.4` release. Which we're planning to publish soon. Also, for trying out your case, follow [helm documentation on multiline strings](https://helm.sh/docs/chart_template_guide/yaml_techniques/#strings-in-yaml).

PR #2256, opened 2022-04-27
Change site theme

theunrepentantgeek:

Our existing website theme is fairly plain and has some issues

* Not wide enough for some content to render properly
* Navigation sidebar is very large and dominates the content

Switching to the [Docsy](https://www.docsy.dev/) theme gives us a nicer look & feel, and addresses the above issues nicely.

image

**Prerequisites**

- [x] Merge #2252 so that our API docs have proper FrontMatter

Our existing theme (Book) relies on each document having it's own level one header; Docsy injects one based on the FrontMatter, so I needed to remove a bunch to avoid having duplicate headers showing on pages.

Also related is #2251 which should ensure tables can wrap within the (wider) margins of this theme.

gif

**If applicable**:


super-harsh (approved):

docs/hugo/content/architecture-decision-records/ADR-2020-04-Code-Generation.md
```diff
@@ -2,8 +2,6 @@
 title: '2020-04: Why Code Generation?'
```
matthchr, line 2: Does this actually get rendered in the page? Just making sure it makes sense to nuke the title below.

matthchr (approved):

docs/hugo/content/architecture-decision-records/ADR-2020-04-Code-Generation.md
theunrepentantgeek: Different themes do different things with the frontmatter - with Book, it was used only for menus; with Docsy it's rendered in the page as well.

Issue #1443, opened 2021-05-14
What to do about write-once properties?


Porges:
Some ARM templates have properties that can be set on creation but not modified later. What do we do about these? Does Kubernetes have support/concepts that match this?

bergerx:
New place for the kep is here: https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1101-immutable-fields#suggested-marker-syntax

matthchr:
Still something we should consider

matthchr:
Still tracking this but as far as I can tell no movement upstream, so we'll sit on this some more. It's not an issue in practice for us right now.

matthchr:
No status change from August when we visited this issue last.

theunrepentantgeek:
No change. Still something we need to sort.

matthchr:
No change from above. Looks like the KEP is stalled.

matthchr:
See https://kubernetes.io/blog/2022/09/23/crd-validation-rules-beta/ and https://github.com/google/cel-spec

matthchr:
still interesting

matthchr:
This is still interesting but not really that critical (which is why it keeps lingering in this list).

theunrepentantgeek:
Still of interest.

matthchr:
Still interesting but not urgent

matthchr:
Still interested in this.

theunrepentantgeek:
Still of interest.

PR #2260, opened 2022-05-03
Add support for validation on AzureName and ResourceGroup immutability

### LINKED ISSUE #2134: We should have a webhook that rejects updates to AzureName
`AzureName` can be updated, which will cause a new resource to be created in Azure and orphan the old one.

**Describe the improvement**
We should block changes to `AzureName` on updates.

super-harsh:
Closes #2134 

We had an issue where changing AzureName and Resource groups would abandon the old resource and creates a new one. Which is solved using validation web-hook on UPDATE for immutable fields in this PR. 

This PR includes:

- Validation for AzureName immutability
- Validation for Resource Group immutability
- Helper method to handle more immutable fields in future
- Edge case test for AzureName and ResourceGroup


v2/tools/generator/internal/codegen/testdata/ArmResource/Arm_test_dependent_resource_and_ownership_azure.golden
```diff
@@ -193,7 +193,32 @@ func (a *A) updateValidations() []func(old runtime.Object) error {
 		func(old runtime.Object) error {
 			return a.validateResourceReferences()
 		},
+		a.validateImmutableProperties}
+}
+
+// validateImmutableProperties validates all immutable properties
+func (a *A) validateImmutableProperties(old runtime.Object) error {
+
+	resourceID := genruntime.GetResourceIDOrDefault(a)
```
Porges, line 202: rather than inlining the code here, can you generate a call to a function in `genruntime` instead? we should be able to reuse the same code for all resource types, which will save on code size (binary and textual)

Porges (changes requested):
see comment

v2/tools/generator/internal/codegen/testdata/ArmResource/Arm_test_dependent_resource_and_ownership_azure.golden
super-harsh: Done!

v2/api/authorization/v1alpha1api20200801preview/role_assignment_types_gen.go
```diff
@@ -239,7 +239,16 @@ func (assignment *RoleAssignment) updateValidations() []func(old runtime.Object)
 		func(old runtime.Object) error {
 			return assignment.validateResourceReferences()
 		},
+		assignment.validateImmutableProperties}
```
matthchr, line 242: I think this file isn't gofmt'ed? We might need to fix the `astbuilder` bit that builds this list because I think the `{` should be on the next line? Or maybe I'm crazy.

v2/internal/controllers/edge_case_test.go
```diff
@@ -221,3 +176,69 @@ func Test_CreateStorageAccountWithoutRequiredProperties_Rejected(t *testing.T) {
 	// Create it again
 	tc.CreateResourcesAndWait(acctCopy)
 }
+
+func Test_AzureNameImmutability(t *testing.T) {
```
matthchr, line 180: minor: Name of the test should include what it's testing. Possibly `Test_AzureName_IsImmutableOnceSuccessfullyCreated`?

matthchr: Same below for the owner one.

matthchr: You should also cover the cases I mentioned in Teams where the object isn't successfully created yet. We should allow those mutations if we can.

v2/internal/controllers/edge_case_test.go
```diff
@@ -221,3 +176,69 @@ func Test_CreateStorageAccountWithoutRequiredProperties_Rejected(t *testing.T) {
...
+	rg := tc.CreateTestResourceGroupAndWait()
+
+	acct := createStorageAccount(tc, rg)
+	tc.CreateResourcesAndWait(acct)
+
+	// Patch the account to change AzureName
+	newAzureName := "test123"
+	old := acct.DeepCopy()
+	acct.Spec.AzureName = newAzureName
+	tc.PatchAndExpectError(old, acct)
```
matthchr, line 194: Maybe instead of doing `PatchAndExpectError` we should just have a `Patch` that returns the error. Then you can actually assert that it has the right shape as well?

Porges (approved):

v2/tools/generator/internal/functions/kubernetes_admissions_validations.go
```diff
@@ -77,3 +85,59 @@ func validateResourceReferencesBody(codeGenerationContext *astmodel.CodeGenerati
...
+	fn.AddParameter("old", astbuilder.QualifiedTypeName(runtimePackage, "Object"))
+	fn.AddComments("validates all immutable properties")
+
+	return fn.DefineFunc()
+}
+
+// validateImmutablePropertiesFunctionBody helps generate the body of the validateImmutablePropertiesFunctionBody function:
+// oldObj, ok := old.(*Receiver)
+// if !ok {
+// return nil
```
matthchr, line 119: minor: Fix indentation in the comment

v2/pkg/genruntime/admissions.go
```diff
@@ -26,3 +28,22 @@ type Defaulter interface {
...
+// ValidateImmutableProperties function validates the update on immutable properties.
+func ValidateImmutableProperties(oldObj MetaObject, newObj MetaObject) error {
+	var errs []error
+
+	if GetResourceIDOrDefault(newObj) == "" {
+		return nil
+	}
+
+	if oldObj.AzureName() != newObj.AzureName() {
+		errs = append(errs, errors.Errorf("updating 'AzureName()' is not allowed for '%s", oldObj.GetName()))
```
theunrepentantgeek, line 41: Why is `AzureName` written as though it's a method call?

```suggestion
		errs = append(errs, errors.Errorf("updating 'AzureName( is not allowed for '%s", oldObj.GetName()))
```

theunrepentantgeek: Also, does `oldObj.GetName()` return a fully qualified GVK or only a partial name?

v2/tools/generator/internal/codegen/testdata/ArmResource/Arm_oneof_resource_conversion_on_arm_type_only_azure.golden
```diff
@@ -196,7 +196,16 @@ func (resource *FakeResource) updateValidations() []func(old runtime.Object) err
 		func(old runtime.Object) error {
 			return resource.validateResourceReferences()
 		},
+		resource.validateImmutableProperties}
+}
+
+// validateImmutableProperties validates all immutable properties
+func (resource *FakeResource) validateImmutableProperties(old runtime.Object) error {
```
theunrepentantgeek, line 203: We suppress some pipeline stages for these golden tests as they focus on property structure and the functions are noise - perhaps validation should be suppressed too? #unsure

v2/pkg/genruntime/admissions.go
```diff
@@ -26,3 +28,22 @@ type Defaulter interface {
 	// CustomDefault performs custom defaults that are run in addition to the code generated defaults.
 	CustomDefault()
 }
+
+// ValidateImmutableProperties function validates the update on immutable properties.
+func ValidateImmutableProperties(oldObj MetaObject, newObj MetaObject) error {
```
theunrepentantgeek, line 33: I'm not sure "Immutable" is quite the right term for these properties, but I'm having trouble coming up with a reasonable alternative.

v2/api/authorization/v1alpha1api20200801preview/role_assignment_types_gen.go
super-harsh: Its the `astbuilder` .. I'll double check and fix it

v2/tools/generator/internal/codegen/testdata/ArmResource/Arm_oneof_resource_conversion_on_arm_type_only_azure.golden
super-harsh: We already have `validateResourceReferences` in the golden tests though and its been a while..

v2/internal/controllers/edge_case_test.go
super-harsh: We already have `Patch` which expects error to be nil. Can make this one return `error` and then assert on if error is not nil

v2/pkg/genruntime/admissions.go
super-harsh: Just the partial name.

v2/tools/generator/internal/codegen/testdata/ArmResource/Arm_oneof_resource_conversion_on_arm_type_only_azure.golden
super-harsh: We already have validateResourceReferences in the golden tests though

v2/internal/controllers/edge_case_test.go
```diff
@@ -221,3 +176,97 @@ func Test_CreateStorageAccountWithoutRequiredProperties_Rejected(t *testing.T) {
...
+	acct.Spec.AzureName = newAzureName
+	
+	err := tc.PatchAndExpectError(old, acct)
+
+	tc.Expect(err).ToNot(BeNil())
+	tc.Expect(old.Spec.AzureName).ToNot(BeIdenticalTo(newAzureName))
+
+	// Delete the account
+	tc.DeleteResourceAndWait(acct)
+
```
matthchr, line 202: minor: extra line here

v2/internal/controllers/edge_case_test.go
```diff
@@ -221,3 +176,97 @@ func Test_CreateStorageAccountWithoutRequiredProperties_Rejected(t *testing.T) {
...
+	acct.Spec.Owner = testcommon.AsOwner(rg2)
+
+	err := tc.PatchAndExpectError(old, acct)
+
+	tc.Expect(err).ToNot(BeNil())
+	tc.Expect(old.Owner().Name).ToNot(BeIdenticalTo(rg2.Name))
+
+	// Delete the account
+	tc.DeleteResourceAndWait(acct)
+
```
matthchr, line 227: minor: extra line here

v2/internal/controllers/edge_case_test.go
```diff
@@ -221,3 +176,97 @@ func Test_CreateStorageAccountWithoutRequiredProperties_Rejected(t *testing.T) {
...
+	tc := globalTestContext.ForTest(t)
+
+	rg := tc.CreateTestResourceGroupAndWait()
+
+	acct := createStorageAccount(tc, rg)
+	tc.CreateResourcesAndWait(acct)
+
+	rg2 := tc.CreateTestResourceGroupAndWait()
+	// Patch the account to change AzureName
```
matthchr, line 216: You're changing the Owner here rather than the `AzureName` - I think fix the comment

v2/internal/controllers/edge_case_test.go
```diff
@@ -221,3 +176,97 @@ func Test_CreateStorageAccountWithoutRequiredProperties_Rejected(t *testing.T) {
...
+
+	tc.Expect(err).ToNot(BeNil())
+	tc.Expect(old.Owner().Name).ToNot(BeIdenticalTo(rg2.Name))
+
+	// Delete the account
+	tc.DeleteResourceAndWait(acct)
+
+}
+
+func Test_Owner_IsNotImmutableIfSuccessfullyNotCreated(t *testing.T) {
```
matthchr, line 230: minor/naming clarity: Might be clearer if rather than double (triple?) negative we do: `IsMutableIfNotSuccessfullyCreated`

v2/internal/controllers/edge_case_test.go
matthchr: same in below tests

v2/internal/controllers/edge_case_test.go
```diff
@@ -221,3 +176,97 @@ func Test_CreateStorageAccountWithoutRequiredProperties_Rejected(t *testing.T) {
...
+	actualOwnerName := rg.Name
+	rg.Name = invalidOwnerName
+
+	acct := createStorageAccount(tc, rg)
+	tc.CreateResource(acct)
+
+	// Patch the account to change AzureName
+	old := acct.DeepCopy()
+	acct.Spec.Owner.Name = actualOwnerName
+	tc.Patch(old, acct)
```
matthchr, line 246: Should this be `PatchResourceAndWait` to ensure that the patched resource actually provisions successfully too?

v2/internal/controllers/edge_case_test.go
matthchr, line 230: You should be able to test this for `AzureName` too by first attempting to use an `AzureName` that's invalid and then correct it? For example if you try to make a storage account named `-------` I think you'll get rejected. You can then patch it to have a different name which should work.

matthchr (approved):
Approved but take a look at my test comments as I think there's 1 more test you should have.

v2/internal/controllers/edge_case_test.go
super-harsh: Done!

v2/internal/controllers/edge_case_test.go
matthchr: Totally agree that's the right approach

v2/tools/generator/internal/functions/kubernetes_admissions_validations.go
```diff
@@ -77,3 +85,59 @@ func validateResourceReferencesBody(codeGenerationContext *astmodel.CodeGenerati
...
+	receiverType := receiver.AsType(codeGenerationContext)
+
+	runtimePackage := codeGenerationContext.MustGetImportedPackageName(astmodel.APIMachineryRuntimeReference)
+
+	fn := &astbuilder.FuncDetails{
+		Name:          methodName,
+		ReceiverIdent: receiverIdent,
+		ReceiverType: &dst.StarExpr{
+			X: receiverType,
+		},
```
theunrepentantgeek, line 101: There's a helper function in `astbuilder` for this.

v2/tools/generator/internal/functions/kubernetes_admissions_validations.go
```diff
@@ -77,3 +85,59 @@ func validateResourceReferencesBody(codeGenerationContext *astmodel.CodeGenerati
...
+	return fn.DefineFunc()
+}
+
+// validateImmutablePropertiesFunctionBody helps generate the body of the validateImmutablePropertiesFunctionBody function:
+//  oldObj, ok := old.(*Receiver)
+//  if !ok {
+//		return nil
+//  }
+//
+// 	return genruntime.ValidateImmutableProperties(oldObj, <receiverIndent>)
```
theunrepentantgeek, line 122: Use spaces for indentation within comments, and add a blank line.

```suggestion
// validateImmutablePropertiesFunctionBody helps generate the body of the validateImmutablePropertiesFunctionBody function:
//
//  oldObj, ok := old.(*Receiver)
//  if !ok {
//      return nil
//  }
//
// return genruntime.ValidateImmutableProperties(oldObj, <receiverIndent>)
```

v2/tools/generator/internal/functions/kubernetes_admissions_validations.go
```diff
@@ -77,3 +85,59 @@ func validateResourceReferencesBody(codeGenerationContext *astmodel.CodeGenerati
...
+func validateImmutablePropertiesFunctionBody(receiver astmodel.TypeName, codeGenerationContext *astmodel.CodeGenerationContext, receiverIdent string) []dst.Stmt {
+
+	genRuntime := codeGenerationContext.MustGetImportedPackageName(astmodel.GenRuntimeReference)
+
+	obj := dst.NewIdent("oldObj")
+
+	cast := astbuilder.TypeAssert(obj, dst.NewIdent("old"), astbuilder.Dereference(receiver.AsType(codeGenerationContext)))
+	checkAssert := astbuilder.ReturnIfNotOk(astbuilder.Nil())
+
+	returnStmt := astbuilder.Returns(
```
theunrepentantgeek, line 132: Suggest configuring this to have a blank line before the **return** statement.

v2/tools/generator/internal/functions/kubernetes_admissions_validations.go
super-harsh: Updated to use `astbuilder.Deference`

v2/tools/generator/internal/functions/kubernetes_admissions_validations.go
super-harsh: Added!

v2/pkg/genruntime/admissions.go
matthchr: We probably dont need version as this error is for all versions always I think? But G + K is good.

v2/internal/controllers/edge_case_test.go
```diff
@@ -221,3 +176,115 @@ func Test_CreateStorageAccountWithoutRequiredProperties_Rejected(t *testing.T) {
...
+
+	invalidAzureName := "==--039+/"
+
+	acct := createStorageAccount(tc, rg)
+	acct.Spec.AzureName = invalidAzureName
+	tc.CreateResource(acct)
+
+	// Patch the account to change AzureName
+	old := acct.DeepCopy()
+	acct.Spec.AzureName = "storagetestname"
```
matthchr, line 241: Instead of a hardcoded name here you should use the name helper (or just the original name from the `acct` when returned from `createStorageAccount`.

Name helper is: `tc.MakeObjectMetaWithName(tc.NoSpaceNamer.GenerateName("stor"))`

Porges:
Somewhat fixes #1443, but we still need support for more field types there.

PR #2270, opened 2022-05-10
Permit lookup of type configuration after rename

theunrepentantgeek:

We support renaming a resource (or other type) using a `$exportAs` directive in our configuration YAML. 

When this is used, we can get some odd errors when looking up other configuration for that type:

``` 
configuration of group documentdb: 
configuration of version 2021-05-15 has no detail for type MongodbDatabase 
(20 available types are ... elided ...)
```

The problem stems from using the new name to complete the lookup, rather than the original one, and not being able to find the expected config.

With this PR, retrieving the new name automatically creates an alias so that the configuration can still be found when required.

I also found a better way to allow injection of configuration for test purposes (see the new methods `ModifyGroup()`, `ModifyVersion()`, `ModifyType()` and `ModifyProperty()` on `ObjectModelConfiguration`) and migrated existing tests to the new technique.

I considered adding metadata to the type to carry the old name forward allowing lookup to fall back to the original name but that seemed to be very unsimple.

I also considered requiring two separate configuration sections, one under the old name, one under the new, but that would require users to have knowledge of our pipeline in order to put the configuration in the right place and that didn't seem very viable.

gif

**If applicable**:
- [x] this PR contains tests


v2/tools/generator/internal/config/version_configuration.go
```diff
@@ -90,6 +88,34 @@ func (vc *VersionConfiguration) findType(name string) (*TypeConfiguration, error
...
+	}
+
+	// Make sure we don't already have configuration for 'alias'
+	other, err := vc.findType(alias)
+	if other != nil && other != tc {
+		return errors.Errorf(
+			"unable to create type alias %s for %s because that would conflict with existing configuration",
+			alias,
+			name)
+	}
```
Porges, line 108: 💄 I always like empty lines after blocks 😄

v2/tools/generator/internal/config/version_configuration.go
```diff
@@ -90,6 +88,34 @@ func (vc *VersionConfiguration) findType(name string) (*TypeConfiguration, error
...
+	if err != nil && !IsNotConfiguredError(err) {
+		return errors.Wrapf(err, "unable to create type alias %s", alias)
+	}
+
+	// Add the alias as another route to the existing configuration
+	a := strings.ToLower(alias)
```
Porges, line 114: would prefer an `addType` that bundles the `ToLower` with adding it, so we don't make any mistakes

Porges:
> I considered adding metadata to the type to carry the old name forward allowing lookup to fall back to the original name but that seemed to be very unsimple.

Another argument for per-type metadata in the `Types` set 😄

Porges (approved):

v2/tools/generator/internal/config/version_configuration.go
theunrepentantgeek: Done.

PR #2278, opened 2022-05-12
Try setting a higher limit on GOMAXPROCS

Porges:
Attempt to speed up CI run with 🪄magic🔮.


Porges:
Doesn't seem to have made a material difference. Closing.

PR #2279, opened 2022-05-12
Grouped taskfile output

Porges:
This will make things easier to read in CI. GitHub has special syntax for specifying groups which we use here.

Note that this disables log streaming, even on local runs. As far as I can tell there's not yet a way to conditionally switch between grouped/prefixed output.


Porges:
The output here isn't great.

PR #2285, opened 2022-05-17
Use real API Version enums

### LINKED ISSUE #1554: Generated packages should define an accessible ARMAPIVersion variable

Porges:
1. Add a new pipeline stage to create (singleton) enum types for API versions
2. Update GetAPIVersion function generation to return this enum value
3. Correct TypeVisitor so it visits the API Version typename (if any) on resource types
4. Add a small hack into the Storage type creation to also copy API versions explicitly

This fixes #1554


v2/tools/generator/internal/armconversion/arm_spec_interface.go
```diff
@@ -30,8 +30,8 @@ func checkPropertyPresence(o *astmodel.ObjectType, name astmodel.PropertyName) e
 func NewARMSpecInterfaceImpl(
 	idFactory astmodel.IdentifierFactory,
 	resource *astmodel.ResourceType,
-	spec *astmodel.ObjectType) (*astmodel.InterfaceImplementation, error) {
-
+	spec *astmodel.ObjectType,
+) (*astmodel.InterfaceImplementation, error) {
```
theunrepentantgeek, line 34: I still dislike this style.

Reads line. Sees `,`. Aha! There's more on the next line. `)`. Nope! Fooled you.

v2/tools/generator/internal/astmodel/resource_type.go
```diff
@@ -453,9 +469,10 @@ func (resource *ResourceType) References() TypeNameSet {
 	}
 
 	result := SetUnion(spec, status)
+
 	// It's a bit awkward to have to do this, but it doesn't exist as a reference
 	// anywhere else
-	if !resource.APIVersionTypeName().Equals(TypeName{}, EqualityOverrides{}) {
+	if resource.HasAPIVersion() {
```
theunrepentantgeek, line 475: :+1: Declarative intention revealing code.

v2/tools/generator/internal/codegen/code_generator.go
```diff
@@ -137,6 +137,7 @@ func createAllPipelineStages(idFactory astmodel.IdentifierFactory, configuration
 		// TODO: These should be removed if/when we move to Swagger as the single source of truth
 		pipeline.RemoveTypeProperty(),
 		pipeline.RemoveAPIVersionProperty(),
+		pipeline.AddAPIVersionEnums(),
```
theunrepentantgeek, line 140: Are these two lines removing something and then adding it right back again? 
If so, does this need to happen further upstream?

v2/tools/generator/internal/codegen/pipeline/create_storage_types.go
```diff
@@ -39,6 +39,20 @@ func CreateStorageTypes() *Stage {
...
+			for _, tdef := range typesToConvert {
+				if rt, ok := astmodel.AsResourceType(tdef.Type()); ok && rt.HasAPIVersion() {
+					storageAPIVersions.Add(rt.APIVersionTypeName())
+				}
+			}
+
+			for name := range storageAPIVersions {
+				typesToConvert.Add(state.Definitions()[name])
+			}
+
```
theunrepentantgeek, line 55: Could this be done by modifying (and renaming) the predicate `isResourceOrObject` instead?

theunrepentantgeek (approved):
A couple of minor suggestions, but otherwise looks good.

v2/tools/generator/internal/astmodel/resource_type.go
```diff
@@ -225,6 +225,15 @@ func (resource *ResourceType) WithFunction(function Function) *ResourceType {
 	return result
 }
 
+// WithFunction returns a new Resource without the specific function
```
matthchr, line 228:
```suggestion
// WithoutFunction returns a new Resource without the specific function
```

v2/tools/generator/internal/armconversion/arm_spec_interface.go
matthchr: For whatever it's worth I agree with @theunrepentantgeek.

v2/tools/generator/internal/astmodel/type_visitor.go
```diff
@@ -236,7 +236,26 @@ func IdentityVisitOfResourceType(this *TypeVisitor, it *ResourceType, ctx interf
 		return nil, errors.Wrapf(err, "failed to visit resource status type %q", it.status)
 	}
 
-	if visitedSpec == it.spec && visitedStatus == it.status {
+	changedAPIVersionName := false
+	if it.HasAPIVersion() {
+		originalAPIVersionTypeName := it.APIVersionTypeName()
+		newAPIVersion, err := this.visitTypeName(this, originalAPIVersionTypeName, ctx)
```
matthchr, line 242: Ah, this is nice.

v2/tools/generator/internal/codegen/code_generator.go
matthchr: I think the distinction is between the property (on the resource) containing an APIVersion field, and the actual API versions themselves (in the package, but not as a field on the resource)

v2/tools/generator/internal/codegen/testdata/ArmResource/Arm_test_dependent_resource_and_ownership_crossplane.golden
```diff
@@ -89,40 +89,25 @@ type DList struct {
 }
 
 // +kubebuilder:validation:Enum={"2020-06-01"}
-type ASpecAPIVersion string
+type APIVersion string
 
-const ASpecAPIVersion20200601 = ASpecAPIVersion("2020-06-01")
+const APIVersionValue = APIVersion("2020-06-01")
```
matthchr, line 94: I think that this issue ends up fixing #1554

matthchr: I've updated your PR summary to include that (so on merge this should close the issue)

matthchr (approved):

v2/tools/generator/internal/codegen/code_generator.go
Porges: Precisely.

v2/tools/generator/internal/codegen/pipeline/create_storage_types.go
Porges: Not very easily, it would require another pass. I think it's good to call this out as the HACK that it is 😁 

Ideally we wouldn't have these at all in the storage types, so this is a reminder that we would like to remove them in future.

v2/tools/generator/internal/armconversion/arm_spec_interface.go
Porges: I think the default formatter I'm getting is slightly more aggressive than `go fmt`…

v2/internal/controllers/crd_dbformysql_flexibleserver_test.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/kr/pretty"
 	. "github.com/onsi/gomega"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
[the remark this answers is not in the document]
Porges: Not sure! Formatter must have done it 🤔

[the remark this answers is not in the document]
Porges: Changed it to `core`

v2/cmd/controller/main.go
```diff
@@ -23,6 +22,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 
+	asometrics "github.com/Azure/azure-service-operator/v2/internal/metrics"
+
```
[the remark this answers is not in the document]
Porges: Done

PR #2287, opened 2022-05-17
Treat failures to flatten properties as warnings, not errors

theunrepentantgeek:

Sometimes we can't flatten properties (for example, when the object to be flattened implements a One-Of restriction); instead of aborting the generator for such cases, log a warning and proceed.

gif


Porges (approved):
Ideally we'd have a test for this as well, but approving.

PR #2288, opened 2022-05-17
Simply access to Group and Version

theunrepentantgeek:

Most of the time when accessing the `Group` and `Version` from a `PackageReference` we either know that we won't have an `ExternalPackageReference` or want to panic if we do have one. We had boilerplate doing this over and over, so I've added a helper to simplify consumption.

We now have
* `TryGroupVersion()` as it may not succeed (returning a **false**)
* `GroupVersion()` which either succeeds or panics

gif

### DIFF
--- a/v2/tools/generator/internal/astmodel/external_package_reference.go
+++ b/v2/tools/generator/internal/astmodel/external_package_reference.go
@@ -54,12 +54,12 @@ func (pr ExternalPackageReference) String() string {
 	return pr.packagePath
 }
 
-// GroupVersion returns the group and version of this local reference.
+// TryGroupVersion returns the group and version of this external reference.
 func (pr ExternalPackageReference) TryGroupVersion() (string, string, bool) {
 	return "", "", false
 }
 
-// MustGroupVersion triggers a panic because external references don't have a group or version
+// GroupVersion triggers a panic because external references don't have a group or version
 func (pr ExternalPackageReference) GroupVersion() (string, string) {
 	panic(fmt.Sprintf("external package reference %s doesn't have a group or version", pr))
 }
--- a/v2/tools/generator/internal/astmodel/external_package_reference.go
+++ b/v2/tools/generator/internal/astmodel/external_package_reference.go
@@ -55,6 +55,6 @@ func (pr ExternalPackageReference) String() string {
 }
 
 // GroupVersion returns the group and version of this local reference.
-func (pr ExternalPackageReference) GroupVersion() (string, string, bool) {
+func (pr ExternalPackageReference) TryGroupVersion() (string, string, bool) {
 	return "", "", false
 }
--- a/v2/tools/generator/internal/astmodel/external_package_reference.go
+++ b/v2/tools/generator/internal/astmodel/external_package_reference.go
@@ -58,3 +58,8 @@ func (pr ExternalPackageReference) String() string {
 func (pr ExternalPackageReference) TryGroupVersion() (string, string, bool) {
 	return "", "", false
 }
+
+// MustGroupVersion triggers a panic because external references don't have a group or version
+func (pr ExternalPackageReference) GroupVersion() (string, string) {
+	panic(fmt.Sprintf("external package reference %s doesn't have a group or version", pr))
+}
--- a/v2/tools/generator/internal/astmodel/local_package_reference.go
+++ b/v2/tools/generator/internal/astmodel/local_package_reference.go
@@ -116,7 +116,7 @@ func IsLocalPackageReference(ref PackageReference) bool {
 }
 
 // GroupVersion returns the group and version of this local reference.
-func (pr LocalPackageReference) GroupVersion() (string, string, bool) {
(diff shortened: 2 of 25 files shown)


v2/tools/generator/internal/config/type_matcher.go
```diff
@@ -78,26 +78,25 @@ func (t *TypeMatcher) matches(glob string, regex **regexp.Regexp, name string) b
...
-		if result {
-			if t.matchedTypes == nil {
-				t.matchedTypes = astmodel.NewTypeNameSet(typeName)
-			} else {
-				t.matchedTypes.Add(typeName)
-			}
-		}
+	group, version, ok := typeName.PackageReference.TryGroupVersion()
+	if !ok {
+		// Never match external references
```
matthchr, line 83: This code seems like it has to be a bug?

If `!ok` we enter this if and then just proceed? Is this supposed to be a `return false`?

matthchr (changes requested):

v2/tools/generator/internal/astmodel/external_package_reference.go
```diff
@@ -55,6 +55,11 @@ func (pr ExternalPackageReference) String() string {
 }
 
 // GroupVersion returns the group and version of this local reference.
```
super-harsh, line 57:
```suggestion
// TryGroupVersion returns the group and version of this local reference.
```

v2/tools/generator/internal/astmodel/external_package_reference.go
```diff
@@ -55,6 +55,11 @@ func (pr ExternalPackageReference) String() string {
...
-func (pr ExternalPackageReference) GroupVersion() (string, string, bool) {
+func (pr ExternalPackageReference) TryGroupVersion() (string, string, bool) {
 	return "", "", false
 }
+
+// MustGroupVersion triggers a panic because external references don't have a group or version
```
super-harsh, line 62:
```suggestion
// GroupVersion triggers a panic because external references don't have a group or version
```

v2/tools/generator/internal/astmodel/local_package_reference.go
```diff
@@ -116,10 +116,15 @@ func IsLocalPackageReference(ref PackageReference) bool {
 }
 
 // GroupVersion returns the group and version of this local reference.
```
super-harsh, line 118:
```suggestion
// TryGroupVersion returns the group and version of this local reference.
```

v2/tools/generator/internal/astmodel/local_package_reference.go
```diff
@@ -116,10 +116,15 @@ func IsLocalPackageReference(ref PackageReference) bool {
...
-func (pr LocalPackageReference) GroupVersion() (string, string, bool) {
+func (pr LocalPackageReference) TryGroupVersion() (string, string, bool) {
 	return pr.group, pr.Version(), true
 }
 
+// MustGroupVersion returns the group and version of this local reference.
```
super-harsh, line 123:
```suggestion
// GroupVersion returns the group and version of this local reference.
```

v2/tools/generator/internal/astmodel/package_reference.go
```diff
@@ -30,7 +30,9 @@ type PackageReference interface {
 	IsPreview() bool
 	// GroupVersion returns the group and version of this reference.
```
super-harsh, line 31: Comments here need some alteration?

v2/tools/generator/internal/astmodel/property_reference.go
```diff
@@ -43,6 +43,6 @@ func (ref PropertyReference) IsEmpty() bool {
 
 // String returns a string representation of this property reference
 func (ref PropertyReference) String() string {
-	g, v, _ := ref.declaringType.PackageReference.GroupVersion()
+	g, v := ref.declaringType.PackageReference.GroupVersion()
```
super-harsh, line 46: Not sure if we should use `TryGroupVersion()` here? Or would `PackageReference` always be local reference?

v2/tools/generator/internal/astmodel/external_package_reference.go
super-harsh: Same for some below.

v2/tools/generator/internal/config/type_matcher.go
theunrepentantgeek: Oh, yes - good catch.

super-harsh (approved):

v2/tools/generator/internal/astmodel/property_reference.go
theunrepentantgeek: A `PropertyReference` is a reference to a property on one of the types we decare - which are never in external packages, so the reference will always be either local or storage.

v2/tools/generator/internal/astmodel/package_reference.go
theunrepentantgeek: Fixed.

matthchr (approved):

PR #2290, opened 2022-05-17
Refactor reconciler to further cater to non-ARM resources

matthchr:
This is moving us in a positive direction but doesn't get us all the way there.
There will be further changes needed, but rather than build them all up and get it perfect, 
I'm instead sending pieces so that it's merged faster and also is somewhat more digestible.

### DIFF
--- a/v2/internal/reconcilers/arm/azure_generic_arm_reconciler.go
+++ b/v2/internal/reconcilers/arm/azure_generic_arm_reconciler.go
@@ -99,12 +99,9 @@ func (r *AzureDeploymentReconciler) Reconcile(
 		result, err = instance.CreateOrUpdate(ctx)
 	}
 
-	if readyErr, ok := conditions.AsReadyConditionImpactingError(err); ok {
-		return ctrl.Result{}, instance.WriteReadyConditionError(ctx, obj, readyErr)
-	}
 	if err != nil {
 		return ctrl.Result{}, err
 	}
 
-	return result, err
+	return result, nil
 }
--- a/v2/internal/reconcilers/arm/azure_generic_arm_reconciler.go
+++ b/v2/internal/reconcilers/arm/azure_generic_arm_reconciler.go
@@ -7,17 +7,13 @@ package arm
 
 import (
 	"context"
-	"fmt"
-	"strings"
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/Azure/azure-service-operator/v2/internal/config"
-	. "github.com/Azure/azure-service-operator/v2/internal/logging"
 	"github.com/Azure/azure-service-operator/v2/internal/resolver"
 	"github.com/Azure/azure-service-operator/v2/internal/util/kubeclient"
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
@@ -26,9 +22,6 @@ import (
 
 // TODO: I think we will want to pull some of this back into the Generic Controller so that it happens
 // TODO: for all resources
-
-// TODO: Do we actually want this at the controller level or this level?
-const GenericControllerFinalizer = "serviceoperator.azure.com/finalizer"
 
 type CreateOrUpdateAction string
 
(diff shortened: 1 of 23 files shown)


v2/internal/reconcilers/arm/azure_generic_arm_reconciler.go
```diff
@@ -46,6 +39,7 @@ const (
 	DeleteActionMonitorDelete = DeleteAction("MonitorDelete")
 )
 
+// TODO: Stopped here... noodling about if this should be a generic interface?
```
matthchr, line 42: Remove this

v2/pkg/genruntime/resolved_secrets.go
```diff
@@ -30,3 +30,14 @@ func (r ResolvedSecrets) LookupSecret(ref SecretReference) (string, error) {
 	}
 	return result, nil
 }
+
+// MustLookupSecret looks up the secret value for the given reference. If the reference is nil, an error is returned.
+// If the secret cannot be found, an error is returned
+// TODO: Wondering if there's a better name for this than what I've got?
```
matthchr, line 36: Remove this note... but also if anybody has a better name feel free to suggest it

theunrepentantgeek: Why is this not just `LookupSecret()`?

I thought the Go ecosystem convention for the prefix `Must` was that such methods would panic, and not return errors.

Given that this one _can_ fail, maybe `TryLookupSecret()`?

theunrepentantgeek (approved):
Looks good - I think I see where you're heading. Might be an idea to create some class diagrams to show the relationships between interfaces.

v2/internal/reconcilers/common.go
```diff
@@ -0,0 +1,171 @@
...
+			"resourceVersion", obj.GetResourceVersion(),
+			"generation", obj.GetGeneration(),
+			"uid", obj.GetUID(),
+			"ownerReferences", obj.GetOwnerReferences(),
+			"creationTimestamp", obj.GetCreationTimestamp(),
+			"finalizers", obj.GetFinalizers(),
+			"annotations", ourAnnotations,
+			// Use fmt here to ensure the output uses the String() method, which log.Info doesn't seem to do by default
+			"conditions", fmt.Sprintf("%s", obj.GetConditions()),
+		}
```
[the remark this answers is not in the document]
matthchr: The problem is that the actual `log` call below which uses this value has the signature:
`func (l Logger) Info(msg string, keysAndValues ...interface{})`

So if I make this `map[string]interface{}`, then I can't just pass it in to the log function (which just expects a slice of interfaces).

We're no worse off here than we were before though. Prior to refactoring we called `Info` directly which lets you do whatever you want.

v2/pkg/genruntime/resolved_secrets.go
matthchr: Yes you're right, `Must` is not the right prefix. The reason I'm not using `LookupSecret` is that already exists. The difference is that this ones signature accepts a ptr `*SecretReference`.

matthchr: Maybe just `LookupSecretPtr` (even though it's a bit horrible it's clearer than `Must`)

matthchr: Have gone with `LookupSecretPtr` - let me know if you think there's a better name.

matthchr:
> Looks good - I think I see where you're heading. Might be an idea to create some class diagrams to show the relationships between interfaces.

Have filed #2295 this for tracking the relationships. They're still in flux a bit so for now I think it makes sense to evolve them in code and then once we've gotten them solidified we can draw something up.

v2/pkg/genruntime/resolved_secrets.go
theunrepentantgeek: Ah, my kingdom for some method overloads. 😁

matthchr: Will do `LookupSecretFromPtr` after discussion for extra clarity.

v2/internal/reconcilers/common.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+}
+
+type ARMOwnedResourceReconcilerCommon struct {
+	ReconcilerCommon
+	ResourceResolver *resolver.Resolver
+}
+
+// IsOwnerReady returns true if the owner is ready or if there is no owner required
+func (r *ARMOwnedResourceReconcilerCommon) IsOwnerReady(ctx context.Context, log logr.Logger, obj genruntime.ARMOwnedMetaObject) (bool, error) {
```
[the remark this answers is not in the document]
matthchr: Done

PR #2292, opened 2022-05-18
Change ConvertToARM to return a ptr

matthchr:
`ConvertToARM` returning a struct rather than a ptr to a struct originally came about because the contract in the ARM conversion code generation is that handlers should return value types. This contract makes sense for primitive types. It also makes sense when assigning to slices or maps, as usually there the type is `map[string]Foo` rather than `map[string]*Foo`.

The unfortunate result of this pattern though is that the final type returned by `ConvertToARM` is a struct rather than a ptr. Until now that hasn't been a problem, but for `RouteTable` we need the ability to modify that struct by assigning a new `Properties` field to it, if one doesn't already exist. That isn't possible if a struct is returned and wrapped in an interface like we do.

This change rectifies the issue by having `ConvertToARM` return a pointer. This impacts both the top level result but also the "inner" results as we walk the type graph and convert everything. This has a few consequences:
- The `ConvertToARM` code generation now must dereference the returned pointer in order to continue meeting the contract of always returning value types. This sometimes results in unnatural code that looks like:
    ```
    identity := *identityARM.(*BatchAccountIdentityARM)
    result.Identity = &identity
    ```
- The callers of `ConvertToARM` can now easily modify the returned types.

In an effort to avoid the unnatural `*` + `&` combo above, I put a timeboxed effort towards doing a better job. It turns out that while the contract of the inner handler returning a struct is awkward in this case, it's NOT awkward in other cases such as map and slice handling. There's no trivial way to avoid this as the conversion generation exists now, and given the complexity of that code I'm worried about making it _more_ complicated to solve this "problem", so for now I'm going to leave this as is.

### DIFF
--- a/v2/api/storage/v1beta20210401/storage_accounts_management_policy_types_gen.go
+++ b/v2/api/storage/v1beta20210401/storage_accounts_management_policy_types_gen.go
@@ -543,7 +543,7 @@ func (policies *StorageAccountsManagementPolicies_Spec) ConvertToARM(resolved ge
 	if policies == nil {
 		return nil, nil
 	}
-	var result StorageAccountsManagementPolicies_SpecARM
+	result := &StorageAccountsManagementPolicies_SpecARM{}
 
 	// Set property ‘Name’:
 	result.Name = resolved.Name
@@ -557,7 +557,7 @@ func (policies *StorageAccountsManagementPolicies_Spec) ConvertToARM(resolved ge
 		if err != nil {
 			return nil, err
 		}
-		policy := policyARM.(ManagementPolicySchemaARM)
+		policy := *policyARM.(*ManagementPolicySchemaARM)
 		result.Properties.Policy = &policy
 	}
 
@@ -756,7 +756,7 @@ func (schema *ManagementPolicySchema) ConvertToARM(resolved genruntime.ConvertTo
 	if schema == nil {
 		return nil, nil
 	}
-	var result ManagementPolicySchemaARM
+	result := &ManagementPolicySchemaARM{}
 
 	// Set property ‘Rules’:
 	for _, item := range schema.Rules {
@@ -764,7 +764,7 @@ func (schema *ManagementPolicySchema) ConvertToARM(resolved genruntime.ConvertTo
 		if err != nil {
 			return nil, err
 		}
-		result.Rules = append(result.Rules, itemARM.(ManagementPolicyRuleARM))
+		result.Rules = append(result.Rules, *itemARM.(*ManagementPolicyRuleARM))
 	}
 	return result, nil
 }
@@ -973,7 +973,7 @@ func (rule *ManagementPolicyRule) ConvertToARM(resolved genruntime.ConvertToARMR
 	if rule == nil {
 		return nil, nil
 	}
-	var result ManagementPolicyRuleARM
+	result := &ManagementPolicyRuleARM{}
 
 	// Set property ‘Definition’:
 	if rule.Definition != nil {
@@ -981,7 +981,7 @@ func (rule *ManagementPolicyRule) ConvertToARM(resolved genruntime.ConvertToARMR
 		if err != nil {
 			return nil, err
 		}
-		definition := definitionARM.(ManagementPolicyDefinitionARM)
+		definition := *definitionARM.(*ManagementPolicyDefinitionARM)
 		result.Definition = &definition
 	}
 
@@ -1300,7 +1300,7 @@ func (definition *ManagementPolicyDefinition) ConvertToARM(resolved genruntime.C
 	if definition == nil {
 		return nil, nil
 	}
-	var result ManagementPolicyDefinitionARM
+	result := &ManagementPolicyDefinitionARM{}
 
 	// Set property ‘Actions’:
 	if definition.Actions != nil {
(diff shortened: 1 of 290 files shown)


v2/tools/generator/internal/armconversion/convert_to_arm_function_builder.go
```diff
@@ -100,7 +100,9 @@ func (builder *convertToARMBuilder) functionBodyStatements() []dst.Stmt {
 	result = append(
 		result,
 		astbuilder.ReturnIfNil(dst.NewIdent(builder.receiverIdent), astbuilder.Nil(), astbuilder.Nil()))
-	result = append(result, astbuilder.NewVariable(builder.resultIdent, builder.armTypeIdent))
+
+	result = append(result, astbuilder.ShortDeclaration(builder.resultIdent, astbuilder.AddrOf(astbuilder.NewCompositeLiteralBuilder(dst.NewIdent(builder.armTypeIdent)).Build())))
```
theunrepentantgeek, line 104: This coule be simplified by using `astbuilder.Statements()` to build the final `[]dst.Stmt`.

v2/tools/generator/internal/codegen/testdata/ArmResource/Arm_oneof_resource_conversion_on_arm_type_only_azure.golden
```diff
@@ -616,7 +616,7 @@ func (properties *Properties) ConvertToARM(resolved genruntime.ConvertToARMResol
 		if err != nil {
 			return nil, err
 		}
-		foo := fooARM.(FooARM)
+		foo := *fooARM.(*FooARM)
```
theunrepentantgeek, line 619: I wonder if we might be able to simplify things to return the expected type directly using generics - in a future PR.

v2/api/storage/v1beta20210401/storage_accounts_management_policy_types_gen.go
```diff
@@ -562,7 +562,7 @@ func (policies *StorageAccountsManagementPolicies_Spec) ConvertToARM(resolved ge
 		if err != nil {
 			return nil, err
 		}
-		policy := policyARM.(ManagementPolicySchemaARM)
+		policy := *policyARM.(*ManagementPolicySchemaARM)
 		result.Properties.Policy = &policy
```
theunrepentantgeek, line 566: Trying to understand the purpose of this - is it to ensure that the original and the ARM variant are completely independent in memory?

theunrepentantgeek (approved):

v2/tools/generator/internal/armconversion/convert_to_arm_function_builder.go
```diff
@@ -482,10 +484,19 @@ func (builder *convertToARMBuilder) convertComplexTypeNameProperty(conversionBui
 			panic(err)
 		}
 
-		typeAssertExpr.Type = astbuilder.Selector(dst.NewIdent(packageName), destinationType.Name())
+		typeAssertExpr.Type = astbuilder.Dereference(astbuilder.Selector(dst.NewIdent(packageName), destinationType.Name()))
 	}
 
-	results = append(results, params.AssignmentHandlerOrDefault()(params.GetDestination(), typeAssertExpr))
+	// TODO: This results in code that isn't very "human-like". Today the contract of these methods is that they
```
matthchr, line 490: Remove this TODO/clarify it

v2/api/storage/v1beta20210401/storage_accounts_management_policy_types_gen.go
matthchr: No it's an artefact of the code generation. I describe the exact "why" of this in the summary of this PR, but the tl;dr is that this code block is generated by a set of handlers calling themselves recursively. So there's two handlers at play here, the inner handler, whose contract is to return a non-ptr, and an outer handler which is the `Optional` handler whose contract is to return a pointer.

In this actual code, this bit is generated by the inner handler:
```
	policyARM, err := (*policies.Policy).ConvertToARM(resolved)
	if err != nil {
		return nil, err
	}
	policy := policyARM.(ManagementPolicySchemaARM)
	policy := *policyARM.(*ManagementPolicySchemaARM) // Result is a non-ptr, as per contract
```

and this bit is generated by the outer (optional) handler:
```
	if policies.Policy != nil {
		<inner>
		result.Properties.Policy = &<inner result> // result is a ptr
	}
```

The only way I see to fix this is that we change the target type the inner handler is trying to "get" to to be a ptr rather than a struct. This means delegating the `&<result>` to that inner handler. The problem is there's a lot of possible inner handlers that can get called which means I need to increase each of their complexity to deal with this case. I can't see an obvious clean way to make sure to generate code that is more human-like that doesn't also increase the complexity of this already complicated code, so I punted on it for now, favoring readability of the generator over readability of the generated code.

It'd be nice to have both though.

v2/tools/generator/internal/armconversion/convert_to_arm_function_builder.go
matthchr: Agreed that might help some, but have kept this as is for now as it matches the pattern of the rest of the file (which was written before `astbuilder.Statements` existed). I think it's probably more important to match the local style of the file, especially in this code which is pretty complex. The alternative would be to fix the whole file, but that'd be a change for another PR

PR #2300, opened 2022-05-20
Tighten yaml config checks

theunrepentantgeek:

Check kind of yaml node when parsing `$azureGeneratedSecrets` to catch possible configuration errors quickly.

Sample error (formatted):
```
E0520 18:49:13.388756   11352 gen_types.go:33] 
Error creating code generator: configuration file loaded from ".\\azure-arm.yaml" is not valid YAML: 
decoding yaml for "storage": 
decoding yaml for "2021-04-01": 
decoding yaml for "StorageAccount": 
type configuration, unexpected configuration for $azureGeneratedSecrets: 
unexpected yaml (line 1046 col 13)
```

This is tightening validation in an area where I tend to make mistakes. 😁 

gif


v2/tools/generator/internal/config/type_configuration.go
```diff
@@ -295,9 +295,24 @@ func (tc *TypeConfiguration) UnmarshalYAML(value *yaml.Node) error {
 		// - secret2
 		if strings.EqualFold(lastId, azureGeneratedSecretsTag) && c.Kind == yaml.SequenceNode {
 			var azureGeneratedSecrets []string
+			var errs []error
```
matthchr, line 298: minor: I'm not totally sold that a collection of errors is actually necessary. For parsing errors, I'd sorta expect the error to come from the beginning of the problem, not all following lines/elements.

matthchr (approved):

v2/tools/generator/internal/config/type_configuration.go
theunrepentantgeek: Fair call, have simplified to return on the first error.

PR #2304, opened 2022-05-20
Improve error handling and fix test race

matthchr:
**If applicable**:
- [x] this PR contains tests


v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -302,6 +302,15 @@ func (r *azureDeploymentReconcilerInstance) BeginCreateOrUpdateResource(ctx cont
...
 	r.Log.V(Status).Info("About to send resource to Azure")
 
+	// TODO: This is a bit tricky. The goal here is, before we send the request, we know we haven't managed this resource before,
+	// TODO: since it doesn't have an ID yet. Then, later IF the first request to the request fails (the actual request, not the
+	// TODO: polling of the long-running operation) with an error we classify as fatal, we can know that the resource wasn't successfully
+	// TODO: created and so we can safely unset the finalizer and ID which we set below in AddInitialResourceState..., which frees the
+	// TODO: user up to make modifications (to owner, AzureName, etc) that they otherwise would've been blocked from
+	// TODO: Adoption of existing resources is a problem here though. What's the expected behavior behavior if we attempt to adopt (and modify)
+	// TODO: a resource and the request is rejected. Have we adopted that resource, or no? If no... then what we're doing below is OK...
+	// resourceAlreadyExisted := genruntime.GetResourceIDOrDefault(r.Obj) != ""
```
matthchr, line 312: Curious as to peoples thoughts on this...

I see two main options:

**We just keep things simple and don't let you change `AzureName`, even if the name is clearly invalid.**
The consequence of this is if you create a resource that gets rejected from Azure because of its `AzureName`, you need to delete and recreate it, rather than just updating the `AzureName`. This matches with the standard `.metadata.name` behavior (at least in ASO), so there's some consistency here.

**We try to be smart about if we've successfully created or updated the resource and allow `AzureName` changes when we "know" we haven't successfully managed the resource ever before.**
The advantage of this is that it's a bit nicer ergonomically to let users to a `k edit` to fix their bad name, rather than forcing a delete and recreate. There are lots of disadvantages though:
- It increases complexity. We'll still need to write the finalizer and ID first, for failover purposes. Then, if we got a fatal error from the request we could unset them.
- It's not clear that _every_ fatal error means the resource wasn't created. We might have to actually look at HTTP status code at this level rather than just errors, which would be awkward (and further increase complexity).
- It doesn't deal with cases where the resource wasn't created but also didn't reject in the response, but instead rejected in the async operation poll, which some services do (I think cosmosdb does that and there are probably others).

I tend to lean towards just cutting this, deleting that commented out test, and moving on with our life... but figured I'd check with folks.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -302,6 +302,15 @@ func (r *azureDeploymentReconcilerInstance) BeginCreateOrUpdateResource(ctx cont
 	}
 	r.Log.V(Status).Info("About to send resource to Azure")
 
+	// TODO: This is a bit tricky. The goal here is, before we send the request, we know we haven't managed this resource before,
+	// TODO: since it doesn't have an ID yet. Then, later IF the first request to the request fails (the actual request, not the
+	// TODO: polling of the long-running operation) with an error we classify as fatal, we can know that the resource wasn't successfully
+	// TODO: created and so we can safely unset the finalizer and ID which we set below in AddInitialResourceState..., which frees the
+	// TODO: user up to make modifications (to owner, AzureName, etc) that they otherwise would've been blocked from
+	// TODO: Adoption of existing resources is a problem here though. What's the expected behavior behavior if we attempt to adopt (and modify)
```
super-harsh, line 310:
```suggestion
	// TODO: Adoption of existing resources is a problem here though. What's the expected behaviour if we attempt to adopt (and modify)
```

v2/internal/controllers/edge_case_test.go
```diff
@@ -225,24 +225,64 @@ func Test_Owner_IsImmutableOnceSuccessfullyCreated(t *testing.T) {
...
+//
+//	// Patch the account to change AzureName
+//	old := acct.DeepCopy()
+//	acct.Spec.AzureName = tc.NoSpaceNamer.GenerateName("stor")
+//	tc.PatchResourceAndWait(old, acct)
+//
+//	tc.Expect(acct.Owner().Name).ToNot(BeIdenticalTo(invalidAzureName))
+//	// Delete the account
+//	tc.DeleteResourceAndWait(acct)
+//}
```
theunrepentantgeek, line 248: Should either uncomment or delete this test.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
theunrepentantgeek: I'm generally in favour of simplicity - trying to be smart about when the name can be updated, and when it cannot, seems like something that only ever gets more complicated, something that acrues edge cases.

How tight can we make the Kubernetes validation of AzureName - can we make it tight enough to enusre that the AzureName must be valid? 

Since the likelihood of any naming error is very low, and we can catch most (if not all) of them with validation, requiring users to delete & recreate doesn't seem to be too onerous.

Porges: > How tight can we make the Kubernetes validation of AzureName - can we make it tight enough to enusre that the AzureName must be valid?

Not at the moment. Some resources have name restrictions that we can't apply since they are not valid Golang regexps.

Porges: I'm in favour of rejecting changes even if the name is invalid. If this happens someone has only just authored the CRI and tried to submit it, so they are (hopefully) still in the stage where the name can be changed and resubmitted.

v2/internal/controllers/edge_case_test.go
matthchr: Fixed

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
matthchr: As discussed, will keep complexity out of here at least for now. We will just reject all changes to AzureName after we've communicated with Azure at all. You _can_ still change it before the owner has been created (because that's how @super-harsh's webhook works), but I think that makes sense and matches how we deal with `Owner`.

babbageclunk (approved):

theunrepentantgeek (approved):

Porges (approved):
🚀

PR #2314, opened 2022-05-24
Remove status enums entirely

Porges:
Simplify the Status types output by removing intermediary enum type.


theunrepentantgeek (approved):

Porges:
`await matthchr;`

matthchr:
I'm not totally sold on this.

**Pros I see:**
- It makes the generated code slightly cleaner
- It conveys the reality that these are just strings and there is a (small) possibility that the service returns something that isn't actually in the Enum set. Technically speaking I think that's a service/ARM bug, but it can happen and we don't prevent it.

**Cons I see:**
- It removes a symmetry between the spec/status types.
- It makes some status comparisons slightly harder to do.
- It makes the hardcoded set of possible enum strings slightly harder to find, because you don't have a type to help guide you.

The argument to remove I _think_ basically boils down to: "We don't validate these so there's no guarantee that they're actually the set of enum values"... which is true, but it's sorta also true for all golang "enums" so in a lot of ways that's a feature of the language and not our fault. Or am I missing something and there's another argument to remove as well?

theunrepentantgeek:
> It conveys the reality that these are just strings and there is a (small) possibility that the service returns something that isn't actually in the Enum set. Technically speaking I think that's a service/ARM bug, but it can happen and we don't prevent it.

Is "adding a new enum value" one of the permitted ARM changes that doesn't require updating the API version?

I'm wondering if one of the more likely changes of this ilk would be when the relevant service adds a new value, say to support a new VM SKU or something like that; the sort of change that might sneak in without us noticing until we regenerate.

Porges:
> > It conveys the reality that these are just strings and there is a (small) possibility that the service returns something that isn't actually in the Enum set. Technically speaking I think that's a service/ARM bug, but it can happen and we don't prevent it.
> 
> Is "adding a new enum value" one of the permitted ARM changes that doesn't require updating the API version?
> 
> I'm wondering if one of the more likely changes of this ilk would be when the relevant service adds a new value, say to support a new VM SKU or something like that; the sort of change that might sneak in without us noticing until we regenerate.

I think that's quite likely, additional VM types might not show up in Swagger specs…

matthchr:
> Is "adding a new enum value" one of the permitted ARM changes that doesn't require updating the API version?

It depends on if the enum is a `model-as-string` enum or not. Most are, so technically teams can add values at any time. Here's what aka.ms/bcforapi says:

> Table 1: Changes not requiring a new api-version of the service: Add an extensible enum (string) 
> Docs for this field must indicate that previously-undocumented values might be returned someday so customer code is written properly to deal with the unexpected value 

> I think that's quite likely, additional VM types might not show up in Swagger specs…

Yeah but doesn't this just leave us in exactly the same place that say, the Golang SDK for Azure is in? They don't republish it once it's out, so new enum fields don't "show up" there. The enum is modelled as a string but in Go that just means "a normal Go enum"

See [here](https://github.com/Azure/azure-sdk-for-go/blob/89aa033db0255021d04ce0dd5bc8200b9810b255/profiles/2020-09-01/compute/mgmt/compute/enums.go#L752) for example

Porges:
Really it would be nice to share the same enum between spec and status, as long as it wasn't validated on the status side…

Porges:
Closing based on @matthchr’s points

PR #2316, opened 2022-05-25
Generator performance fixes

Porges:
We've had a bit of a regression recently in generator performance, which leads to increased time spent doing the edit-test loop.

Here's a grab-bag of fixes driven by `pprof` analysis.

Note that a lot of time is being spent in GC due to the number of object copies we make, so much of the changes below are about reducing the number of allocations.

# Summary

Time spent in memory/GC improved a lot (cumulative times below):

- malloc from 4.59s to 2.63s
- gcDrain from 18.91s to 10.73s

Overall time spent in TypeVisitor (one of the main hotspots) from 11.56s to 6.55s.

Biggest win in terms of allocations is that `ObjectType.copy` used to allocate >2GB of data and now allocates <150MB.

The main thing left to do would be somehow speeding up `TypeDefinitionSet`, as there is a lot of time spent in hashing `TypeNames` and copying maps and so on. A nice fix would be to replace `TypeName` with some lightweight standin containing just an integer value. 

JSON loading also takes a significant part of the runtime but there's not a lot we can do about that.

# Changes made

## Add short-circuiting to `MakeIdentityVisitOfObjectType`, and `WithProperties` on `ObjectType`. 

Improves runtimes of type visitors:
| Before | After |
|-|-|
| image | image |

Memory/GC churn is also greatly reduced:
| Before | After |
|-|-|
| image GC | image |

This reduces runtime by quite a bit for pipeline stages that visit lots of types, for example:
- `Remove validation from all status properties, completed in 2.796s`
- `Remove validation from all status properties, completed in 1.887s`

## Right-size allocations

Enable the `prealloc` linter and follow its suggestions. Also right-size various maps (not detected by `prealloc`) and the map allocations in generated code.

## Reduce memory usage of `TypeMatcher`

`TypeMatcher` was holding onto 14 MB of strings. Now it holds onto none, since it only needs to track `count > 0`.

This reduced snapshot memory usage from 77MB to 59MB.

image

## Make `LocalPackageReference.Version` non-allocating

This was responsible for a lot of garbage:
image

## Make `MakeIdentityVisitOfObjectType` not invoke `PropertySet.AsSlice`

There's no need for this as the function does not require the properties to be in sorted order. Removes all these allocations:

image

(NB: this is improved again by a later change, below, to remove `PropertySet` copies.)

## Shortcut for `IdentityVisitOfTypeName`

When we are using a `TypeVisitor` to visit a `TypeName` and the visitor for `TypeName`s is `IdentityVisitOfTypeName`, skip calling the visitor and return the `Type` directly. This avoids reboxing the `TypeName` into a new `Type`.

This removes most of these allocations:
image

## Introduce `ReadOnlyPropertySet` interface to reduce defensive copies

This only allows readonly operations on the property set. A `PropertySet` can be obtained (to retain preëxisting behaviour) by invoking `Copy()`.

```go
type ReadOnlyPropertySet interface {
	AsSlice() []*PropertyDefinition
	ContainsProperty(name PropertyName) bool
	Get(name PropertyName) (*PropertyDefinition, bool)
	Find(predicate func(*PropertyDefinition) bool) (*PropertyDefinition, bool)
	Copy() PropertySet
	Equals(other ReadOnlyPropertySet, overrides EqualityOverrides) bool
	First() *PropertyDefinition
	ForEach(func(def *PropertyDefinition))
	IsEmpty() bool
	Len() int
}
```

This reduces the number of property copies by a lot:

| Before | After |
|-|-|
| MicrosoftTeams-image (5) | image  | 

## Make `ObjectType` use readonly maps internally to reduce defensive copies

This is the logical conclusion of the above. By using readonly maps internally, we only make copies when needed.

Total allocations from maps decreases a lot:
| Before | After |
|-|-|
| image | image |

## Make `PropertyDefinition` use readonly maps internally

Same idea as previous, and the last change.

| Before | After |
|-|-|
| image | image |


v2/tools/generator/internal/astmodel/type_visitor.go
```diff
@@ -155,7 +155,11 @@ func MakeIdentityVisitOfObjectType(makeCtx MakePerPropertyContext) func(this *Ty
 			if err != nil {
 				errs = append(errs, err)
 			} else {
-				newProps = append(newProps, prop.WithType(p))
+				// only replace property if the type was changed;
+				// this allows short-circuiting below
+				if !TypeEquals(p, prop.propertyType) {
```
theunrepentantgeek, line 160: Only checking `propertyType` runs the risk of discarding required changes to other properties; suggest using `PropertyDefinition.Equals()` instead to check for changes.

theunrepentantgeek (changes requested):
One potential bug, but otherwise looks good.

Porges:
More improvements to come before merging.

v2/tools/generator/internal/astmodel/type_visitor.go
Porges: Only the type of the property is changed by the visitor?

theunrepentantgeek: Apologies, I didn't read the code above - was sure that other characteristics of a property could be modified by the visitor. As you were, all good.

theunrepentantgeek (commented):
I figure this is still a work in progress, so just commenting for now.

v2/tools/generator/internal/codegen/code_generator.go
```diff
@@ -224,6 +228,11 @@ func createAllPipelineStages(idFactory astmodel.IdentifierFactory, configuration
 func (generator *CodeGenerator) Generate(ctx context.Context) error {
 	klog.V(1).Infof("Generator version: %s", version.BuildVersion)
 
+	f, err := os.Create("cpuprofile.prof")
+	if err != nil {
+		log.Fatal(err)
+	}
+
```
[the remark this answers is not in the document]
Porges: undo! undo!

Porges:
hmm, one ordering/nondeterminism issue to chase down before merge

v2/api/authorization/v1alpha1api20200801preview/role_assignment_types_gen.go
```diff
@@ -771,7 +771,7 @@ func (assignments *RoleAssignments_Spec) ConvertToARM(resolved genruntime.Conver
 
 	// Set property ‘Tags’:
 	if assignments.Tags != nil {
-		result.Tags = make(map[string]string)
+		result.Tags = make(map[string]string, len(assignments.Tags))
```
theunrepentantgeek, line 774: Nice - updating the generated code too.

theunrepentantgeek (approved):
Looks good to me. 

BTW: I think you can delete `v2/tools/generator/internal/codegen/cpuprofile.prof`

v2/tools/generator/internal/astmodel/type_visitor_builder.go
```diff
@@ -66,20 +69,20 @@ func (b TypeVisitorBuilder) Build() TypeVisitor {
 // buildVisitTypeName returns a function to use in the TypeVisitor
 // If the field VisitTypeName is nil, we return an identity visitor. Otherwise we attempt to
 // convert the func found in the field, triggering a panic if no suitable func is found.
-func (b *TypeVisitorBuilder) buildVisitTypeName() func(*TypeVisitor, TypeName, interface{}) (Type, error) {
+func (b *TypeVisitorBuilder) buildVisitTypeName() (bool, func(*TypeVisitor, TypeName, interface{}) (Type, error)) {
```
matthchr, line 72: minor: maybe document the first bool params meaning in the functions // comments

matthchr, line 72: minor: maybe document the first bool return meaning in the functions // comments

v2/tools/generator/internal/codegen/pipeline/add_secrets.go
```diff
@@ -54,7 +53,7 @@ func applyConfigSecretOverrides(config *config.Configuration, definitions astmod
 
 	applyConfigSecrets := func(_ *astmodel.TypeVisitor, it *astmodel.ObjectType, ctx interface{}) (astmodel.Type, error) {
 		typeName := ctx.(astmodel.TypeName)
-		for _, prop := range it.Properties() {
+		for _, prop := range it.Properties().Copy() {
```
matthchr, line 56: why do we need `.Copy()` here?

v2/tools/generator/internal/codegen/pipeline/create_arm_types.go
```diff
@@ -123,7 +123,7 @@ func (c *armTypeCreator) createARMResourceSpecDefinition(
 }
 
 func removeValidations(t *astmodel.ObjectType) (*astmodel.ObjectType, error) {
-	for _, p := range t.Properties() {
+	for _, p := range t.Properties().Copy() {
```
matthchr, line 126: Same question on `Copy` here

matthchr (approved):

v2/tools/generator/internal/codegen/pipeline/add_secrets.go
Porges: `Properties` is a `ReadOnlyPropertySet` which you cannot iterate over. `Copy`ing it is pessimistic but also what the existing behaviour was, so it doesn’t make performance _worse_.

v2/tools/generator/internal/codegen/pipeline/create_arm_types.go
Porges: same answer 😄

v2/tools/generator/internal/armconversion/convert_from_arm_function_builder.go
```diff
@@ -301,13 +301,13 @@ func (builder *convertFromARMBuilder) operatorSpecPropertyHandler(
 // to the type being converted.
 func (builder *convertFromARMBuilder) flattenedPropertyHandler(
 	toProp *astmodel.PropertyDefinition,
-	fromType *astmodel.ObjectType) ([]dst.Stmt, bool) {
-
+	fromType *astmodel.ObjectType,
+) ([]dst.Stmt, bool) {
```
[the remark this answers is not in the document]
theunrepentantgeek: While I like having the return values on a separate line, the dangling comma required on the previous line makes this cure considerably worse (in my view) than the disease. My vote is **revert** (but I suspect I'm outvoted here).

v2/internal/resolver/secret_resolver.go
```diff
@@ -9,7 +9,7 @@ import (
 	"context"
 
 	"github.com/pkg/errors"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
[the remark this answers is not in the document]
Porges: Yep. I'm not sure what the intent is here…

v2/tools/generator/internal/codegen/pipeline/add_secrets.go
matthchr: Ahh gotcha, I missed that. That's fine then carry on.

v2/tools/generator/internal/armconversion/convert_from_arm_function_builder.go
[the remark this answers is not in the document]
babbageclunk: Just commenting to cancel out @theunrepentantgeek's vote 😁 - this is my preferred style.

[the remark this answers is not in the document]
babbageclunk: (Although I'm not sure my vote necessarily has the same weight as his!)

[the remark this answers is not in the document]
Porges: The more I look at it the more I like it, 😬

v2/tools/generator/internal/astmodel/type_visitor.go
```diff
@@ -217,6 +218,80 @@ func MakeIdentityVisitOfObjectType(makeCtx MakePerPropertyContext) func(this *Ty
 	}
 }
 
+// This is identical to MakeIdentityVisitOfObjectType except that it iterates proeprties in order
```
theunrepentantgeek, line 221: #pedantry

```suggestion
// This is identical to MakeIdentityVisitOfObjectType except that it iterates properties in alphabetical order
```

theunrepentantgeek (approved):
Looking good.

PR #2322, opened 2022-05-26
Improve diagnostics when TypeMatcher or TypeTransformer matches no types

theunrepentantgeek:

Improves our TypeMacher so that it gives useful diagnostics when misconfigured; previously it would only flag that nothing had matched, leaving a user in the dark about what was wrong.

Motivated by my recent experience where I had to debug things to try and work out what was going wrong.

**Prerequisites**

- [ ] Merge PR #2316 

* When matching a literal string, I've used the `TypoAdvisor` to give the closest matching result.
* When matching with a wildcard, I just list all the available matches.
* When matching with multiple strings (semicolon ';' separated), each is checked in turn so that adding an ineffective `;*.foo` to the end of a matcher gets properly flagged.

Both `literalMatcher` and `globMatcher` stop collecting candidates once a match has happened; this helps to keep memory consumption down, avoiding GC delays.

Some examples showing the results.

When **group** doesn't match, using a literal:
```
Error during code generation:
failed during pipeline stage 1/64: 
Load and walk schema: type filter action: "prune": 
type matcher [Group: "comicbooks"; Version: "2022"; Name: "Superman"; Because: "comic books are cool"] matched no types; 
every group was excluded: no match for "comicbooks" (did you mean communication?)
```

When **version** doesn't match, using a literal:
```
E0526 14:20:23.900005   13692 gen_kustomize.go:111] Error during code generation:
failed during pipeline stage 1/64: Load and walk schema: type filter action: "prune": 
type matcher [Group: "compute"; Version: "2022"; Name: "Superman"; Because: "comic books are cool"] matched no types; 
groups matched, but every version was excluded: no match for "2022" (did you mean v1beta20201201?)
```

When **name** doesn't match, using a literal
```
E0526 14:22:01.974545   37420 gen_kustomize.go:111] Error during code generation:
failed during pipeline stage 1/64: Load and walk schema: type filter action: "prune": 
type matcher [Group: "compute"; Version: "v1beta20201201"; Name: "Superman"; Because: "comic books are cool"] matched no types; 
groups and versions matched, but every type was excluded: no match for "Superman" (did you mean PurchasePlan?)
```

When **group** doesn't match, using a wildcard:
```
E0526 14:31:37.219774   11712 gen_kustomize.go:111] Error during code generation:
failed during pipeline stage 1/64: Load and walk schema: type filter action: "prune": 
Type matcher [Group: "Foo*"; Version: "2030*"; Name: "Virtual*"; Because: "comic books are cool"] matched no types; 
every group was excluded: no match for "Foo*" (available candidates were healthbot, storsimple, dbformysql, appplatform, datafactory, delegatednetwork, devspaces, extendedlocation, resources, edgeorder, healthcareapis, streamanalytics, ...
```

When **version** doesn't match, using a wildcard:
```
E0526 14:22:47.352777   29256 gen_kustomize.go:111] Error during code generation:
failed during pipeline stage 1/64: Load and walk schema: type filter action: "prune": 
Type matcher [Group: "net*"; Version: "2030*"; Name: "Virtual*"; Because: "comic books are cool"] matched no types; 
groups matched, but every version was excluded: no match for "2030*" (available candidates were v1beta20200501, v1beta20200601, v1beta20200801, v1beta20201101, v1beta20191001, v1beta20210201, v1beta20201201, v1beta20210801, v1beta20190501, v1beta20190601, v1beta20190701, v1beta20191101, v1beta20200701, v1beta20200901, v1beta20200301, v1beta20200201, v1beta20200101, v1beta20200401, v1beta20170815, v1beta20190801, v1beta20210401, v1beta20210601)
```

When **name** doesn't match, using a wildcard:
```
E0526 14:34:49.196514   10152 gen_kustomize.go:111] Error during code generation:
failed during pipeline stage 1/64: Load and walk schema: type filter action: "prune": 
Type matcher [Group: "net*"; Version: "v1beta2020*"; Name: "Fomo*"; Because: "comic books are cool"] matched no types; 
groups and versions matched, but every type was excluded: no match for "Fomo*" (available candidates were VolumePropertiesExportPolicy, RulesEngineMatchCondition, FirewallPolicyRuleCondition, FirewallPolicyCertificateAuthority, WebApplicationFirewallPolicyProperties, ApplicationGatewayIPConfiguration,...
```

gif

**If applicable**:
- [x] this PR contains tests


v2/tools/generator/internal/config/glob_matcher.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	return &globMatcher{
+		glob:       glob,
+		candidates: make(set.Set[string]),
+	}
+}
+
+func (gm *globMatcher) Matches(term string) bool {
+	gm.once.Do(gm.createRegex)
```
babbageclunk, line 44: Why defer creating the regex until use, rather than creating it in `newGlobMatcher`? It doesn't seem like there's going to be a situation where the regex isn't needed, and creating it later makes it _slightly_ harder to track down failures if there's a panic from `MustCompile`.

v2/tools/generator/internal/config/glob_matcher.go
```diff
@@ -0,0 +1,92 @@
...
+	return false
+}
+
+func (gm *globMatcher) WasMatched() error {
+	if gm.matched {
+		return nil
+	}
+
+	choices := set.AsSortedSlice(gm.candidates)
+	return errors.Errorf(
```
babbageclunk, line 70: Is this going to get ridiculously long in some cases?

v2/tools/generator/internal/config/glob_matcher.go
```diff
@@ -0,0 +1,92 @@
...
+		gm.glob,
+		strings.Join(choices, ", "))
+}
+
+func (gm *globMatcher) createRegex() {
+	g := regexp.QuoteMeta(gm.glob)
+	g = strings.ReplaceAll(g, "\\*", ".*")
+	g = strings.ReplaceAll(g, "\\?", ".")
+	g = "(?i)(^" + g + "$)"
+	gm.regex = regexp.MustCompile(g)
```
babbageclunk, line 81: I was a bit concerned about the panicking on bad user input here, but I guess the thinking is that this should never fail no matter what the glob pattern is, right?

v2/tools/generator/internal/config/literal_matcher_test.go
```diff
@@ -0,0 +1,43 @@
...
+		value    string
+		expected bool
+	}{
+		{"Exact match", "Foo", "Foo", true},
+		{"Uppercase literal matches lowercase value", "FOO", "foo", true},
+		{"Lowercase literal matches uppercase value", "foo", "FOO", true},
+		{"Mixed case literal matches different mixed case value", "Foo", "foO", true},
+		{"Different strings do not match", "Foo", "Bar", false},
+		{"Leading whitespace does not match", "  Foo", "Foo", false},
+		{"Trailing whitespace does not match", "Foo  ", "Foo", false},
```
babbageclunk, line 29: These are a bit surprising - wouldn't it be nicer to handle these cases and succeed? Or is there some reason a literal matcher shouldn't match them?

v2/tools/generator/internal/config/multi_matcher.go
```diff
@@ -0,0 +1,85 @@
...
+			count++
+		}
+	}
+
+	err := kerrors.NewAggregate(errs)
+	if err != nil {
+		return errors.Wrapf(err, "%d of %d did not match", count, len(mm.matchers))
+	}
+
+	return nil
```
babbageclunk, line 65: `kerrors.NewAggregate` and `errors.Wrapf` both return nil if there are no errors (and `Wrapf` avoids formatting in that case as well), so you _could_ do:

```suggestion
	return errors.Wrapf(kerrors.NewAggregate(errs), "%d of %d did not match", count, len(mm.matchers))
```

v2/tools/generator/internal/config/multi_matcher_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+package config
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+)
+
+func TestMultiMatcher_GivenDefinition_MatchesExpectedStrings(t *testing.T) {
```
babbageclunk, line 14: Add another test for the WasMatched behaviour? I realise that requires indirectly testing another of the matches, but otherwise I don't think there's a test checking for the non-short-circuiting behaviour - it might just disappear unless someone reads the comment.

v2/tools/generator/internal/config/string_matcher.go
```diff
@@ -0,0 +1,39 @@
...
+
+	if HasMultipleMatchers(matcher) {
+		return newMultiMatcher(matcher)
+	}
+
+	if HasWildCards(matcher) {
+		return newGlobMatcher(matcher)
+	}
+
+	return newLiteralMatcher(matcher)
```
babbageclunk, line 38: Maybe make this a `switch`?

v2/tools/generator/internal/config/glob_matcher.go
babbageclunk: Ooh, I just realised - is this to do with making a zero `globMatcher` work like in a new `TypeMatcher`? That's fair enough then.

babbageclunk (approved):
Looks great! Some small comments and questions.

The output examples in the PR description are super-nice, thanks!

v2/tools/generator/internal/config/string_matcher.go
theunrepentantgeek: We're checking for characteristics of the string, not on type values, so I don't think a switch is applicable.

v2/tools/generator/internal/config/multi_matcher_test.go
theunrepentantgeek: Added test. Good idea.

v2/tools/generator/internal/config/multi_matcher.go
theunrepentantgeek: Good suggestion; adopted with formatting changes.

v2/tools/generator/internal/config/literal_matcher_test.go
theunrepentantgeek: I don't really have much of an opinion, included the tests as documentation. Changed to be whitespace tolerant.

v2/tools/generator/internal/config/glob_matcher.go
theunrepentantgeek: Correct; the regex returned by `QuoteMeta()` should be valid and the transformations are safe ones.

v2/tools/generator/internal/config/glob_matcher.go
theunrepentantgeek: Potentially, yes - but I didn't see a good way to prune the list of candidates without potentially omitting the one the user wants.

v2/tools/generator/internal/config/glob_matcher.go
theunrepentantgeek: Yup. If the value is never set, the zero value needs to work.

v2/tools/generator/internal/config/field_matcher.go
```diff
@@ -0,0 +1,63 @@
...
+var _ StringMatcher = &FieldMatcher{}
+
+func NewFieldMatcher(field string) FieldMatcher {
+	return FieldMatcher{
+		actual: NewStringMatcher(field),
+	}
+}
+
+func (dm *FieldMatcher) String() string {
+	if dm.actual == nil {
```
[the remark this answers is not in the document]
theunrepentantgeek: Yup. That's the wall I ran into.

PR #2323, opened 2022-05-26
Use Swagger as sole input

Porges:
nothing to see here, WIP

----

## Changes

Note that this time around I reduced scope to try to do the minimal amount to change over to Swagger-only. We can always improve the naming conventions in a fresh PR.

### Naming changes

- Now allow `_` in names. This is contra established Golang guidelines, but they don't appear in CRDs and being able to separate names makes a significant difference in readability when we generate long names due to multiple layers of unnamed nested objects. This also avoids generating clashes when RP Swagger uses `_` in some types.
- Use `_` to separate enum name from enum value in enum value identifiers. (e.g. `SkuName_Standard_LRS`)
- Use `_STATUS` as status-type suffix to avoid generating collisions with `_Status` names that are generated due to properties etc.
- When generating names based upon `_STATUS` names, ensure that the suffix is kept at the end of the name, instead of being embedded when adding more parts onto the end.
- Spec types are now correctly singularized, matching the resource type. 

### Improvements

- `Location` property disappears from several places that it was not needed. (`RoleAssignment`, global `MariaDB`, …)
- Some other properties that were documented as "doing nothing" have disappeared.
- Most names have gotten better, a few have gotten worse. `VirtualMachineScaleSets_Spec_Properties_VirtualMachineProfile_NetworkProfile_NetworkInterfaceConfigurations_Properties_IpConfigurations` is now `VirtualMachineScaleSetIPConfiguration`

----

### Remaining work

- [x] add `Id` property to all status types (?)
- [x] figure out why extension types aren't triggering/working properly
- [ ] restore cross-resource-references functionality
- [ ] review all `donotmerge`, including config file


Porges:
Getting close now, a few compilation errors in tests.

v2/tools/generator/internal/codegen/pipeline/determine_resource_ownership.go
```diff
@@ -26,32 +26,52 @@ func DetermineResourceOwnership(configuration *config.Configuration) *Stage {
...
+	}
+}
+
+func isNewer(left, right astmodel.TypeDefinition) bool {
+	leftVersion := left.Name().PackageReference.(astmodel.LocalPackageReference).ApiVersion()
+	rightVersion := right.Name().PackageReference.(astmodel.LocalPackageReference).ApiVersion()
+
+	// versions are ASCIIbetically ordered
+	return leftVersion > rightVersion
+}
```
theunrepentantgeek, line 48: Maybe use the existing `astmodel.ComparePathAndVersion` instead?

theunrepentantgeek: Might need to copy/relocate it.

v2/tools/generator/internal/codegen/pipeline/determine_resource_ownership.go
```diff
@@ -26,32 +26,52 @@ func DetermineResourceOwnership(configuration *config.Configuration) *Stage {
...
-			def.Name(),
-			resolved.SpecDef.Name(),
-			resolved.SpecType)
-		if err != nil {
-			return nil, err
-		}
-		if childResourcePropertyTypeDef == nil {
-			continue // This just means skip
-		}
+		me := def.Type().(*astmodel.ResourceType)
```
theunrepentantgeek, line 72: Consider using `astmodel.AsResourcetype()` to avoid issues with wrapper meta-types.

Porges:
Currently stuck awaiting a workaround for go compiler bug... 🤦

hack/crossplane/apis/resources.md
```diff
@@ -7,14 +7,14 @@ These are the Crossplane resources committed to our **main** branch, grouped by
...
 ## Sql
 
-| Resource        | ARM Version        | CRD Version           | Supported From | Sample |
-|-----------------|--------------------|-----------------------|----------------|--------|
-| Server          | 2020-11-01-preview | v1beta20201101preview | -              | -      |
-| ServersDatabase | 2020-11-01-preview | v1beta20201101preview | -              | -      |
+| Resource        | ARM Version           | CRD Version           | Supported From | Sample |
+|-----------------|-----------------------|-----------------------|----------------|--------|
+| Server          | 2020--1-1--01-preview | v1beta20201101preview | -              | -      |
+| ServersDatabase | 2020--1-1--01-preview | v1beta20201101preview | -              | -      |
```
theunrepentantgeek, line 19: This ARM version is definitely amiss.

theunrepentantgeek (approved):
LGTM. After so much effort, will be pleased to land this!

PR #2325, opened 2022-06-01
Support wildcards with transformation targets

theunrepentantgeek:

Despite the appearance of fragments like this in our configuration

``` yaml
  - name: "*"
    property: Copy
    ifType:
      group: definitions
      version: "*"
      name: ResourceCopy
      optional: true
    remove: true
    because: ... elided... 
```

We did not support wildcards within the `ifType` section; the matching on version `*` was accidentally working, due to the way our package references filtered out non-alphanumeric characters during creation.

In PR #2294 I'm removing this filtering to improve our diagnostics, breaking this match.

This PR fixes the matching to properly work with wildcards. We could do some further work to improve diagnostics, etc, but I figured this was change enough.

I've included caching to improve the performance of the transforms, and to reduce the amount of heap debris, hopefully lowering the amount of time spent doing garbage collection. Crudely measures, this seems to help:

Without caching:
```
8/64: Modify property types using configured transforms, completed in 2.162s
8/64: Modify property types using configured transforms, completed in 2.328s
```

With caching
```
8/64: Modify property types using configured transforms, completed in 1.221s
8/64: Modify property types using configured transforms, completed in 1.099s
```

I've left in tests I wrote during troubleshooting.

gif

**If applicable**:
- [x] this PR contains tests


v2/tools/generator/internal/config/type_transformer.go
```diff
@@ -55,37 +56,163 @@ type TypeTransformer struct {
...
+	t, err := target.produceTargetType(descriptor, makeLocalPackageReferenceFunc)
+	if err != nil {
+		return err
+	}
+
+	target.actualType = t
+	return nil
+}
+
+func (target TransformTarget) produceTargetType(
```
matthchr, line 186: be consistent with ptr or non-ptr. This one is non-ptr while all the rest are ptr

v2/tools/generator/internal/config/type_transformer.go
```diff
@@ -55,37 +56,163 @@ type TypeTransformer struct {
...
+		return result
+	}
+
+	result := target.appliesToType(t)
+	target.appliesCache[t] = result
+	return result
+}
+
+func (target *TransformTarget) appliesToType(t astmodel.Type) bool {
+	if target == nil {
```
matthchr, line 74: Does this ever actually happen? Do we actually call `appliesToType` on a `nil` `target`? Guarding against this is sorta weird but it'd be weird if this ever happened too?

v2/tools/generator/internal/config/type_transformer.go
```diff
@@ -55,37 +56,163 @@ type TypeTransformer struct {
...
+		return true
+	}
+
+	inspect := t
+	if target.Optional {
+		// Need optional
+		opt, ok := astmodel.AsOptionalType(inspect)
```
matthchr, line 81: I don't think this is what `target.Optional` meant in the old code? Looking at the original `TransformTarget`, optional was used like this in `produceTargetType`:

```
if target.Optional {
	result = astmodel.NewOptionalType(result)
}
```

That's the only place it was used. Maybe you're fixing a bug here where it just was ignored when used on `ifType`? I think that's what's happening but just wanna confirm.

v2/tools/generator/internal/config/type_transformer.go
```diff
@@ -55,37 +56,163 @@ type TypeTransformer struct {
...
+		if !ok {
+			// but don't have optional
+			return false
```
matthchr, line 84: This could be a bit weird though, because what error is going to result when you set `Optional: True`, then match a non-optional type. Is it going to say "I couldn't find that property"? That's not actually correct though because it can find it, it just didn't match the optional requirement.

matthchr (approved):

v2/tools/generator/internal/astmodel/local_package_reference.go
```diff
@@ -110,14 +110,23 @@ func (pr LocalPackageReference) HasVersionPrefix(prefix string) bool {
...
 }
 
+// ApiVersion returns the API version of this reference, separate from the generator version
 func (pr LocalPackageReference) ApiVersion() string {
 	return pr.apiVersion
 }
 
+// HasApiVersion returns true if this reference has the specified API version
+func (pr LocalPackageReference) HasApiVersion(ver string) bool {
+	// TODO: When we start preserving the API version properly switch this to
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed in #2294, but this PR needs to merge first.

v2/tools/generator/internal/config/type_transformer.go
theunrepentantgeek: That result type was used with a `Equals()` to see if we had the type that needed to be modified; this approach doesn't work with wildcards. Instead, we check to see if we need optional, and if we do, whether we have it.

v2/tools/generator/internal/config/string_matcher.go
```diff
@@ -16,6 +16,8 @@ type StringMatcher interface {
 	Matches(value string) bool
 	// WasMatched returns nil if the matcher had a match, otherwise returning a diagnostic error
 	WasMatched() error
+	// IsRestrictive returns true if the matcher is populated and will restrict matches
+	IsRestrictive() bool
```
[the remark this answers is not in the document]
theunrepentantgeek: I don't think those are better - they imply that tha match is required, whereas the semantics I'm looking for are "is there a matcher here that will reduce the number of matches?"

v2/tools/generator/internal/config/type_transformer.go
theunrepentantgeek: Fixed.

v2/tools/generator/internal/config/type_transformer.go
theunrepentantgeek: It seemed like an easy way to remove boilerplate from calling methods. A **nil** value can happen if a matcher is never populated (say, if it is not used in the YAML).

v2/tools/generator/internal/config/type_transformer.go
theunrepentantgeek: I'll tweak the error message to say "matched types but all properties were excluded **by name or type**" (the bold bit is new)

v2/tools/generator/internal/config/type_transformer.go
```diff
@@ -36,11 +39,9 @@ type TypeTransformer struct {
 
 	// IfType only performs the transform if the original type matches (only usable with Property at the moment)
 	IfType *TransformTarget `yaml:"ifType,omitempty"`
-	ifType astmodel.Type    // cache the astmodel type
```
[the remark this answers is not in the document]
theunrepentantgeek: Excactly.

PR #2331, opened 2022-06-07
bump azcore, azidentity and azure-sdk to v1.0.0

### LINKED ISSUE #2324: Bump azcore and azidentity packages to v1.0.0

Currently, we're using `azcore v0.21.0` and `azidentity v0.13.0` which is refraining us from using newer versions from `azure-sdk-for-go/sdk/resourcemanager`.

**Describe the improvement**

Bump versions `azcore v1.0.0` and `azidentity v1.0.0` and refactor `internal/genericarmclient` to support newer versions.

super-harsh:
Closes #2324 

This PR supports update to azcore, azidentity and azure-sdk packages to v1.0.0 . 

We need this PR to support extensions for Machine Learning resources which are supported from azure-sdk v1.0.0 .


v2/internal/genericarmclient/poller.go
```diff
@@ -10,15 +10,16 @@ import (
...
 	"github.com/pkg/errors"
 )
 
 // PollerResponse is the response from issuing a PUT to Azure. It contains a poller (for polling the long-running
 // operation URL) and a RawResponse containing the raw HTTP response.
 type PollerResponse struct {
 	// Poller contains an initialized poller.
-	Poller *azcore.Poller
+	//TODO: Keeping the generic type as string here.. Should we narrow it down to a type?
+	Poller *azcoreruntime.Poller[string]
```
matthchr, line 22: I don't think we want this to be `string`. I based `generic_client.go` on https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/resourcemanager/resources/armresources/zz_generated_client.go

If you look at that, the poller is using `GenericResource` https://github.com/Azure/azure-sdk-for-go/blob/09a0280a389ea06855ce9c7e78afd1fdb3bf2d99/sdk/resourcemanager/resources/armresources/zz_generated_models.go#L922 as its generic type. I had deleted the `GenericResource` because originally it was only used for the input type and we're instead using `interface{}`.

Possibly we should have a generic type we're deserializing responses into, as subset of `GenericResource` that looks something like:
```
type GenericResource struct {
	// The kind of the resource.
	Kind *string `json:"kind,omitempty"`

	// READ-ONLY; Resource ID
	ID *string `json:"id,omitempty"

	// READ-ONLY; Resource name
	Name *string `json:"name,omitempty"

	// READ-ONLY; Resource type
	Type *string `json:"type,omitempty"
}
```

then we could use that here?

matthchr: It doesn't really matter that much right now because nothing uses the well-typed response from the poller... but in the future it might be useful.

v2/api/cache/customizations/redis_extensions.go
```diff
@@ -60,7 +60,11 @@ func (ext *RedisExtension) RetrieveSecrets(
 		subscription := armClient.SubscriptionID()
 		// Using armClient.ClientOptions() here ensures we share the same HTTP connection, so this is not opening a new
 		// connection each time through
-		redisClient := armredis.NewClient(subscription, armClient.Creds(), armClient.ClientOptions())
+		var redisClient *armredis.Client
+		redisClient, err = armredis.NewClient(subscription, armClient.Creds(), armClient.ClientOptions())
```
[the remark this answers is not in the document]
super-harsh: re-using the `err` variable here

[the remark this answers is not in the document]
super-harsh: Otherwise using `:=` would initialise another scoped var?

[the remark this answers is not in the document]
theunrepentantgeek: I think one of our linters complains, forcing the slightly clunky approach.

theunrepentantgeek (approved):

v2/internal/controllers/recordings/Test_Authorization_RoleAssignment_OnResourceGroup_CRUD.yaml
```diff
@@ -108,818 +102,21 @@ interactions:
...
-      Strict-Transport-Security:
-      - max-age=31536000; includeSubDomains
-      X-Content-Type-Options:
-      - nosniff
-    status: 200 OK
-    code: 200
-    duration: ""
-- request:
-    body: ""
+    body: '{"location":"westus2","name":"asotest-mi-tbbith"}'
```
matthchr, line 105: Our recordings shouldn't have had to change when we updated the SDK version... because we should be using fixed/pinned API versions for everything. Do you have some example errors you got after this update? We should investigate them because AFAIK we shouldn't need to change recordings to do this upgrade.

v2/internal/genericarmclient/poller.go
```diff
@@ -10,15 +10,16 @@ import (
...
+	azcoreruntime "github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 	"github.com/pkg/errors"
 )
 
 // PollerResponse is the response from issuing a PUT to Azure. It contains a poller (for polling the long-running
 // operation URL) and a RawResponse containing the raw HTTP response.
 type PollerResponse struct {
 	// Poller contains an initialized poller.
-	Poller *azcore.Poller
+	//TODO: Keeping the generic type as string here.. Should we narrow it down to a type?
```
matthchr, line 21: minor: You fixed this TODO

matthchr (changes requested):

matthchr (approved):
Switching to approve now that I understand why we had to re-record. Thanks for doing that @super-harsh. Once https://github.com/Azure/azure-sdk-for-go/pull/18353 merges and we update to it unfortunately we'll have to re-record again, but oh well.

PR #2333, opened 2022-06-08
Fix makeUniqueIndexMethodName to be unique always

### LINKED ISSUE #2332: Index methods generating with duplicate names

Currently, `makeUniqueIndexMethodName` function does not exactly makes unique name for function in the case where a resource has multiple similar named secret property

**Describe the improvement**

Have some sort of map to check the names for a property, if already exists, add the index(number in order) 

**Additional context**
Encountered this while adding MachineLearning/WorkspacesCompute

super-harsh:
Closes #2332 

This PR adds some counter index logic to make sure that `IndexMethods` always have a unique name by appending counter index to the method names.

### DIFF
--- a/v2/tools/generator/internal/codegen/pipeline/export_controller_type_registrations.go
+++ b/v2/tools/generator/internal/codegen/pipeline/export_controller_type_registrations.go
@@ -75,13 +75,18 @@ func handleSecretPropertyChains(
 	idFactory astmodel.IdentifierFactory,
 	def astmodel.TypeDefinition,
 ) ([]*functions.IndexRegistrationFunction, []string) {
+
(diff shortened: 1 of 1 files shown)


theunrepentantgeek (changes requested):
There's already a PR somewhere fixing this in a different way - @matthchr may remember details.

If memory serves, instead of adding index numbers (which may not be stable from generation to generation), we were going to use the last _**n**_ steps of the chain to disambiguate.

That is, if using just step _**n**_ is insufficient to be unique, use steps _**n-1**_ + _**n**_ for those particular index methods. If that's not unique, introduce _**n-2**_, and so on.

matthchr:
I think we talked about fixing this but I dont think I ever actually wrote it up. At least, I dont see it in a branch locally.

PR #2342, opened 2022-06-13
Add verifier for IsSecret property modifier 

### LINKED ISSUE #2341: IsSecret modifier is not verified by the generator
Currently, our generator does not care about if `isSecret` is consumed or not before returning the result. Which results in getting no errors(skips) when the field is not correct, or cannot be a secret. 

**Current Behaviour:** 
Generator skips such unknown fields 

**Fix:** 
Add a verifier which verifies is `isSecret` is consumed or not before returning the result

super-harsh:
Closes #2341 

Added fix for verifying if the secret is consumed or not before adding secrets in generator

### DIFF
--- a/v2/azure-arm.yaml
+++ b/v2/azure-arm.yaml
@@ -728,12 +728,6 @@ objectModelConfiguration:
       ManagedDiskParameters:
         Id:
           $armReference: true
-      OSProfile:
-        AdminPassword:
-          $isSecret: true
-      VirtualMachineScaleSetOSProfile:
-        AdminPassword:
-          $isSecret: true
     2021-07-01:
       DiskEncryptionSetParameters:
         Id: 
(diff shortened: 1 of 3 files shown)


matthchr (approved):

v2/azure-arm.yaml
```diff
@@ -728,12 +728,6 @@ objectModelConfiguration:
       ManagedDiskParameters:
         Id:
           $armReference: true
-      OSProfile:
-        AdminPassword:
-          $isSecret: true
-      VirtualMachineScaleSetOSProfile:
-        AdminPassword:
-          $isSecret: true
```
theunrepentantgeek, line 736: Do these properties exist? If so, they should be treated as secrets.

super-harsh: No, they don't exist. These are the redundant ones.

Issue #83, opened 2019-08-14
Submit Azure-Sample example for Go SDK API Management


frodopwns:
Once we are confident in our usage of the API Management client in the Azure SDK for Go, create and submit an example of using the SDK for deploying an API to https://github.com/Azure-Samples/azure-sdk-for-go-samples

Use the ACR or one of the other existing samples as a model for your work.

Acceptance Criteria: Users wanting to interact with Azure API Management services via Go, can achieve a minimal deployment by cloning/running our example

PR #2351, opened 2022-06-21
Add samples test suite 

### LINKED ISSUE #1458: Do a better job testing the samples we have

super-harsh:
Closes #1458 

This PR adds support for testing samples we have to make the samples we provide are correct and working as is. 

We can't use random names for some resources due to name being used in armIDs and tags. Figuring the tags and armIDs in each file would be too much of processing cost. So for these resources using as is names in the samples. 

Does not include tests for the following groups:
- dbformysql ( Flexible server has a bug )
- dbforpostfresql ( Flexible server has a bug 
- cache (Resource refs are too complicated and nested. We might need improvisation to the test suite to support this)
- network (Resource refs are too complicated and nested. We might need improvisation to the test suite to support this)

**If applicable**:
- [x] this PR contains tests


v2/config/samples/storage/v1alpha1api20210401_storageaccountsmanagementpolicy.yaml
```diff
@@ -1,26 +0,0 @@
-apiVersion: storage.azure.com/v1alpha1api20210401
```
super-harsh, line 1: Deleted this file as this resource was not supported in `alpha` versions

v2/internal/testcommon/samples_tester.go
```diff
@@ -0,0 +1,226 @@
...
+	// We need unstructured object here to fetch the correct GVK for the object
+	unstructuredObj := unstructured.Unstructured{Object: jsonMap}
+
+	obj, err := genruntime.NewObjectFromExemplar(&unstructuredObj, t.tc.GetScheme())
+	t.tc.Expect(err).To(BeNil())
+
+	byteData, err = json.Marshal(unstructuredObj.Object)
+	t.tc.Expect(err).To(BeNil())
+
+	err = runtime.DecodeInto(t.decoder, byteData, obj)
```
theunrepentantgeek, line 132: Why do you need to remarshal from `unstructuredObj` back to `byteData` instead of just reusing `byteData` since it's already populated, presumably with the same content?

v2/internal/testcommon/samples_tester.go
```diff
@@ -0,0 +1,226 @@
...
+	handleChildObjects(t.tc, &depSamples)
+
+	return &samples, &depSamples
+}
+
+func (t SamplesTester) getObjectFromFile(path string) genruntime.ARMMetaObject {
+	jsonMap := make(map[string]interface{})
+
+	byteData, err := ioutil.ReadFile(path)
+	t.tc.Expect(err).To(BeNil())
```
theunrepentantgeek, line 115: These `err` checks should be more idiomatic, with the method returning `(genruntime.ARMMetaObject, error)` and the calling code (line #83) using `t.tc.Expect()` to do the check in one place.

v2/internal/testcommon/samples_tester.go
```diff
@@ -0,0 +1,226 @@
...
+	for gvk, sample := range samples.childObjects {
+
+		specField := reflect.ValueOf(sample.GetSpec())
+		originalVersion := specField.MethodByName("OriginalVersion").Call([]reflect.Value{})
+
+		tc.Expect(originalVersion[0]).ToNot(BeNil())
+		tc.Expect(originalVersion).ToNot(HaveLen(0))
+		tc.Expect(originalVersion[0].String()).ToNot(BeEmpty())
+
+		version := originalVersion[0].String()
```
theunrepentantgeek, line 155: No need to monkey around with reflection; all our code generated resources meet the `genruntime.GroupVersionKindAware` interface by implementing `OriginalGVK()`, like this:

``` go
// OriginalGVK returns a GroupValueKind for the original API version used to create the resource
func (machine *VirtualMachine) OriginalGVK() *schema.GroupVersionKind {
	return &schema.GroupVersionKind{
		Group:   GroupVersion.Group,
		Version: machine.Spec.OriginalVersion(),
		Kind:    "VirtualMachine",
	}
}
```

So you should be able to do a simple typecast to `genruntime.GroupVersionKindAware` and call it directly.

theunrepentantgeek: Using the interface is also safer - not every resource has `OriginalVersion` as a method; some have a field with the same name.

v2/internal/testcommon/samples_tester.go
```diff
@@ -0,0 +1,226 @@
...
+		}
+	}
+	return false
+}
+
+func getGroupPath(group string) string {
+	return path.Join(samplesPath, group)
+}
+
+func constructGVK(group, version, kind string) string {
```
theunrepentantgeek, line 223: Prefer to be explicit about the type of each parameter:

```suggestion
func constructGVK(group string, version string, kind string) string {
```

theunrepentantgeek (commented):
First pass of review, just a few comments.

I need to come back and understand what this is doing, and why it's doing it - it's different to what I thought we'd discussed, so I need to understand it better before coming back with additional questions.

v2/config/samples/cache/v1alphaapi/refs/v1alpha1api20201201_redis.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: cache.azure.com/v1alpha1api20201201
 kind: Redis
 metadata:
-  name: sampleredis
+  name: sampleredis2
```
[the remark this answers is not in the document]
super-harsh: I know this looks weird. Only did the splitting for versions for `cache` as recording them as a whole took around 2.5 hours. So this split is just to record versions separately to reduce the time if someone wants to re-record in the future.

v2/internal/controllers/samples_test.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+func Test_DbforMariaDB_Samples_CreationAndDeletion(t *testing.T) {
+	t.Parallel()
+	tc := globalTestContext.ForTest(t)
+	runGroupTest(tc, "dbformariadb", true)
+}
+
+// TODO: FlexiblServer is not being able to be created. Hitting RESPONSE 404 as ARM is sending a success
+// and still controller can't find the created resource TODO: update me
```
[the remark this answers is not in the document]
super-harsh: ah! my bad. Had to update the comment here so left a comment. Will remove this

v2/internal/testcommon/samples_tester.go
super-harsh: True! I had some mangling around with the byte data before parsing to set the owners. For which I found solution using `reflect` and have missed out on removing this. My bad

v2/internal/testcommon/samples_tester.go
```diff
@@ -0,0 +1,230 @@
...
+func NewSampleObject() SampleObject {
+	return SampleObject{
+		SamplesMap:   make(map[string]*linkedhashmap.Map),
+		childObjects: make(map[string]genruntime.ARMMetaObject),
+	}
+}
+
+func NewSamplesTester(tc *KubePerTestContext, useRandomName bool) *SamplesTester {
+	return &SamplesTester{
+		tc:            tc,
```
matthchr, line 65: I think it might be better if you wrote this "like a library". What I mean is that, you didn't use `tc` to do assertions (`tc.Expect`) in the methods. Just return error and bubble up errors to the caller and then do the `tc.Expect(err).ToNot(HaveOccurred())` at the top level in the test helper (`runGroupTest`).

This has a number of advantages:
1. You can just peel off the `NoSpaceNamer` and `Namespace` from `tc` rather than keeping the whole `tc` on `SamplesTester`.
2. It becomes clearer which functions can fail in this file (the ones that return error).
3. It makes this code more unit testable (not that you're necessarily going to write unit tests for your test helper, but it's good practice to be in IMO).

Basically, don't _assume_ this library is going to be used in a test context, with Gomega assertions. Ours _is_ but it's easier to just leave that up to the caller of this class.

v2/internal/testcommon/samples_tester.go
```diff
@@ -0,0 +1,230 @@
...
+
+				if !strings.Contains(filePath, RefsPackage) {
+					if t.UseRandomName {
+						sample.SetName(t.tc.NoSpaceNamer.GenerateName(""))
+					}
+					handleObject(sample, rg, &samples)
+				} else {
+					// We don't change name for dependencies as they have fixed refs which we can't access inside the object
+					// need to make sure we have right names for the refs in samples
+					handleObject(sample, rg, &depSamples)
```
matthchr, line 99: Technically speaking I think you could access and change the refs, but we'd need to write a reflect helper to do it and it'd be a bit of a pain. This is OK for now but might come back to bite us if things aren't uniquely named.

v2/internal/testcommon/samples_tester.go
```diff
@@ -0,0 +1,230 @@
...
+				return err
+			}
+			if !info.IsDir() && !isExclusion(filePath) {
+				//if !info.IsDir() && !isExclusion(filePath) {
+
+				sample := t.getObjectFromFile(filePath)
+
+				sample.SetNamespace(t.tc.Namespace)
+
+				if !strings.Contains(filePath, RefsPackage) {
```
matthchr, line 91: This assumes that no samples directory has `refs` in it, which I am not sure is a valid assumption. Better to use some `filepath` primitive here to do an actual check on the paths rather than treating them all as string?

matthchr (commented):
Left a bunch of comments, not quite done reviewing all of the samples_tester code but figured I'd give you the comments I had so far.

v2/internal/testcommon/samples_tester.go
```diff
@@ -0,0 +1,230 @@
...
+const samplesPath = "../../config/samples"
+
+// exclusions slice contains resources to exclude from test as these resources run into
+// "runtime.notRegisteredErr" error from the Scheme.
+var exclusions = [...]string{
+        "batchaccountpool", 
+        "snapshot",
+	"sqldatabasecontainerthroughputsetting", 
+	"sqldatabasethroughputsetting",
+	"storageaccountsqueueservice",
```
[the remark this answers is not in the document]
super-harsh: Thanks for pointing this out. The apiVersion for this was `microsoft.storage.azure.com/v1alpha1api20210401` where it should be `storage.azure.com/v1alpha1api20210401`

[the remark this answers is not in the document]
super-harsh: Fixing it!

v2/config/samples/dbformariadb/v1beta20180601_configuration.yaml
```diff
@@ -1,34 +1,10 @@
...
-    name: GP_Gen5_2
-    tier: GeneralPurpose
----
-apiVersion: dbformariadb.azure.com/v1beta20180601
 kind: Configuration
 metadata:
   name: query_cache_size
   namespace: default
 spec:
+  azureName: query_cache_size
```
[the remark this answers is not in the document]
super-harsh: Actually, this is need, as we use random name for this sample, which then ends up being the `AzureName` and the config gets rejected as it should be one of the settings name.

v2/config/samples/eventgrid/v1alpha1api20200601_eventsubscription.yaml
```diff
@@ -6,12 +6,14 @@ metadata:
...
+    kind: ResourceGroup
   destination:
     storageQueue:
       endpointType: StorageQueue
       properties: 
         resourceReference: 
           group: storage.azure.com
           kind: StorageAccount
-          name: samplekubestorage
+          name: samplekubestoragealpha
```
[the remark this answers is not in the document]
super-harsh: Have to leave the changed naming for refs, as they get deployed before the resources, so its just to avoid the conflict between alpha and beta resources

v2/config/samples/storage/v1beta20210401_storageaccountsmanagementpolicy.yaml
```diff
@@ -22,5 +22,5 @@ spec:
           - blockBlob
           prefixMatch:
           - sample-container/blob1
-      enabled: true
+      enabled: false
```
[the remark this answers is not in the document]
super-harsh: turned back to true

v2/go.mod
```diff
@@ -79,7 +116,7 @@ require (
 	google.golang.org/protobuf v1.28.0 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
-	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
```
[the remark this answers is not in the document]
super-harsh: ah! its just `go mod tidy` did some playing with indirect refs

v2/internal/testcommon/samples_tester.go
super-harsh: Yes, we could? The only thing I can think of is as each resource has uniquely named field for refs so for example: For EventSubscription it looks like `destination/storageQueueEventSubscriptionDestination/properties/resourceReference` and for VirtualMachine its `networkProfile/networkInterfaces/reference`. 

Which gives a hint that we'll have to iterate through each field using reflect and find out where we want to put the refs?

v2/config/samples/storage/v1alpha1api20210401_storageaccountsmanagementpolicy.yaml
matthchr: I brought this back, so I think you'll need to un-delete this now

v2/config/samples/eventgrid/v1alpha1api20200601_eventsubscription.yaml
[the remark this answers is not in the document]
matthchr: Gotcha. That's something we can maybe iterate on in the tool (to generate ref names too)

v2/internal/testcommon/samples_tester.go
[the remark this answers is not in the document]
matthchr: Wow you caught a lot of sample issues!

v2/internal/testcommon/samples_tester.go
matthchr: Yeah I think this is something to maybe file an item for, not something to do right now

v2/go.mod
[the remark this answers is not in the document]
matthchr: Don't we want the "other" ref though? Or you're saying they're the same...?

v2/internal/controllers/samples_test.go
```diff
@@ -0,0 +1,230 @@
...
+	"github.com/Azure/azure-service-operator/v2/internal/testcommon"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+)
+
+var cases = []struct {
+	Group          string
+	testName       string
+	UseRandomName  bool
+	DeleteChildren bool
+}{
```
matthchr, line 27: Still need to move this to a directory-scan based approach?

v2/internal/testcommon/samples_tester.go
```diff
@@ -0,0 +1,247 @@
...
+
+	resources "github.com/Azure/azure-service-operator/v2/api/resources/v1beta20200601"
+	"github.com/Azure/azure-service-operator/v2/internal/resolver"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+)
+
+const refsPackage = "refs"
+const samplesPath = "../../config/samples"
+
+// exclusions slice contains resources to exclude from test as these are either not an arm resource
```
matthchr, line 32: I think the webtest is the hidden link reference one, I might put that part of this comment down on (or near) the actual webtest line and explain it there, because "hidden link references" is something webtest specific, not actually some general or broad concept.

v2/internal/controllers/samples_test.go
super-harsh: With the directory scan approach we can't set the settings flags per test group as we can right now.

v2/go.mod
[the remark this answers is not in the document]
matthchr: Checking on this as it still seems to be like this vuln is present in v3.0.0?

v2/internal/controllers/samples_test.go
matthchr: Is there a reason that we can't just set `UseRandomName` to true for everybody?

matthchr: For the `DeleteChildren` flag, I think we could maybe get rid of it entirely and just always delete the RG when we're done, rather than exercising delete on each thing individually?

The intent of these tests is to test the samples themselves more than the reconciler behavior. We already cover this in the handcrafted tests so for the samples a simpler deletion approach seems ok since we're more interested in the `kubectl apply` anyway?

That (I think) removes both of these flags and enables you to do a directory scan instead?

v2/internal/controllers/samples_test.go
```diff
@@ -0,0 +1,230 @@
...
+Licensed under the MIT license.
+*/
+
+package controllers_test
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/emirpasic/gods/maps/linkedhashmap"
```
matthchr, line 12: I'm not sure that I follow why we need `linkedhashmap`. I get the impression it's for:
> // Using recursion here to maintain the order of creation and deletion of resources

but I think if you (as discussed down below) just simplify the deletion case into deleting the RG at the end, that you can also simplify the creation case by just applying everything in whatever order you want. Apply order shouldn't matter because the reconciler should work it all out. Then you can just drop the usage of `linkedhashmap` and simplify your `createAndDeleteResourceTree` function some?

Were there specific cases you remember where the creation order really mattered? AFAIK it shouldn't matter for anything.

v2/internal/controllers/samples_test.go
super-harsh: For a few with references present, we can't set random name for all of them

v2/internal/controllers/samples_test.go
super-harsh: We're testing each by using `CreateResourceAndWait`, So if we apply everything in whatever order, and if owner does not already exist for that resource, the test won't proceed. Linkedhashmap is used here to make sure the resources and refs are created in order made sure everything gets created successfully.

matthchr: Hmmm, wondering if we should just split the resources into alpha/beta and then `CreateResourcesAndWait` the whole alpha/beta set? AFAIK order shouldn't matter, although maybe it does for secrets right now as reported in #2385

super-harsh: Also, we can't be using random name for all the tests, as we've got armIDs, refs and links within the resource or child resources which should match the name of resource.

matthchr:
Seems some test failures in CI?

```
[controller:test-integration-kind-ci] --- FAIL: Test_StorageAccount_Secret_CRUD (70.98s)
[controller:test-integration-kind-ci]     kube_per_test_context.go:201: Creating test resource group "asolivetest-rg-wuzygf"
[controller:test-integration-kind-ci]     kube_per_test_context.go:472: 
[controller:test-integration-kind-ci]         Expected an error to have occurred.  Got:
[controller:test-integration-kind-ci]             <nil>: nil
[controller:test-integration-kind-ci]     kube_per_test_context.go:171: Deleting resources before test completes. Resources: [aso-test-storageaccount-secret-crud/asolivetest-rg-wuzygf aso-test-storageaccount-secret-crud/asoliveteststorevxqhw]...
[controller:test-integration-kind-ci]     kube_per_test_context.go:173: All resources deleted
[controller:test-integration-kind-ci] FAIL
```

I also see that the unit tests pass failed here:
```
[produce-markdown-summary] W0725 23:51:58.362287   85348 reflector.go:442] pkg/mod/k8s.io/client-go@v0.24.0-beta.0/tools/cache/reflector.go:167: watch of *v1beta20201201storage.RedisPatchSchedule ended with: an error on the server ("unable to decode an event from the watch stream: http2: client connection lost") has prevented the request from succeeding
[produce-markdown-summary] W0725 23:51:58.365014   85348 reflector.go:442] pkg/mod/k8s.io/client-go@v0.24.0-beta.0/tools/cache/reflector.go:167: watch of *v1beta20180601storage.Server ended with: an error on the server ("unable to decode an event from the watch stream: http2: client connection lost") has prevented the request from succeeding
[produce-markdown-summary] W0725 23:51:58.408065   85348 reflector.go:442] pkg/mod/k8s.io/client-go@v0.24.0-beta.0/tools/cache/reflector.go:167: watch of *v1beta20201101storage.PublicIPAddress ended with: an error on the server ("unable to decode an event from the watch stream: http2: client connection lost") has prevented the request from succeeding
```


I checked out this branch and tried running the tests but got:
```
[controller:test-integration-envtest]         kube_per_test_context.go:342: 
[controller:test-integration-envtest]             Expected success, but got an error:
[controller:test-integration-envtest]                 <*errors.errorString | 0xc0079c86f0>: {
[controller:test-integration-envtest]                     s: "expected pointer, but got invalid kind",
[controller:test-integration-envtest]                 }
[controller:test-integration-envtest]                 expected pointer, but got invalid kind
```

I debugged it some and seem to see that `processSamples` is incorrectly being created like `samplesSlice := make([]client.Object, len(samples))`, but you're using `append`, so you end up with a list of length `2* len(samples)` with the first 2 items `nil`, which causes problems. **I pushed a fix for that**. Will see if that's related to the test failures or if they are just perf problems...

matthchr (approved):

PR #2354, opened 2022-06-21
Merge the design and ADR folders

theunrepentantgeek:

As discussed, offline, having separate folders called _design_ and _architecture-decision-records_ is confusing because it's unclear what goes where. 

This PR merges the two folders with some overview guidance that an ADR should always be present, with a more extensive design document where necessary.

image

gif

**If applicable**:


docs/hugo/content/design/_index.md
```diff
@@ -8,3 +8,17 @@ menu:
...
+Significant changes to Azure Service Operator should be captured by an ADR (architecture design record) in this folder. ADRs are listed chronologically by month, and should contain the following sections:
+
+* Context
+* Decision
+* Status
+* Consequences
+* Experience Report
+* References
+
+Ideally, each ADR should run to one or two pages in length; if more detail is necessary, create a separate design document in this folder and link it from the ADR.
```
matthchr, line 20: So making sure I understand the goal here:

If _planning_ a change: Make an ADR.
If that ADR is getting long, split out into separate design document (more free-form structure) and keep the ADR that links to it. ADR would then just describe what we decided to do and a very brief summary of each option considered, whereas the design document might outline (in detail) multiple proposals, complex diagrams, etc?

Assuming that's correct, it might be useful to clarify here that (if there is a longer design document) the ADR should focus only at a high level on the options and the choice (and then the after-impacts of that choice), but not duplicate the content from the design doc?

matthchr (approved):
Had 1 clarifying question, but this looks good to me otherwise.

super-harsh (approved):

docs/hugo/content/design/_index.md
theunrepentantgeek: That's correct; I'll amend the paragraph to try and capture that.

PR #2358, opened 2022-06-24
Helm updates and upgrade documentation updates

### LINKED ISSUE #2338: Bug: Helm doesn't upgrade CRDs
This basically makes the Helm chart of questionable usefulness. Possibly we should do like cert-manager does: https://cert-manager.io/docs/installation/helm

They're able to make this work, although they also support standalone CRD installation as well.

matthchr:
Closes #2338

### DIFF
--- a/docs/hugo/content/introduction/upgrading.md
+++ b/docs/hugo/content/introduction/upgrading.md
@@ -1,0 +1,50 @@
+---
+title: "Upgrading"
+---
+
+## Before you upgrade
+
+Ensure that you have carefully reviewed the upgrade instructions included in the [release notes](https://github.com/Azure/azure-service-operator/releases) for the release
+you are upgrading to.
+
+## Recommended upgrade pattern
+
+We recommend that you upgrade ASO one minor version at a time. 
+
+| Old version      | New version     |  Recommended  |
+|------------------|-----------------|:-------------:|
+| `v2.0.0-beta.0`  | `v2.0.0-beta.1` |       ✔       |
+| `v2.0.0-alpha.6` | `v2.0.0-beta.1` |       ❌       |
+
+If you need to upgrade Azure Service Operator more than 1 minor version, we recommend that you do so one minor version at a time.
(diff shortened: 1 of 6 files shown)


docs/hugo/content/introduction/upgrading.md
```diff
@@ -0,0 +1,50 @@
...
+
+Ensure that you have carefully reviewed the upgrade instructions included in the [release notes](https://github.com/Azure/azure-service-operator/releases) for the release
+you are upgrading to.
+
+## Recommended upgrade pattern
+
+We recommend that you upgrade ASO one minor version at a time. 
+
+| Old version      | New version     | Recommended |
+|------------------|-----------------|-------------|
```
theunrepentantgeek, line 15: Let's centre the Recommended column.

```suggestion
|------------------|-----------------|:-----------:|
```

matthchr: TIL that centers things.

scripts/generate-helm-manifest.sh
```diff
@@ -14,8 +14,8 @@ DIR=$5
 echo "Generating helm chart manifest"
 sed -i "s@\($PUBLIC_REGISTRY\)\(.*\)@\1azureserviceoperator:$VERSION@g" "$DIR"charts/azure-service-operator/values.yaml
 rm -rf "$DIR"charts/azure-service-operator/templates/generated # remove generated files
-rm -rf "$DIR"charts/azure-service-operator/charts/azure-service-operator-crds/crds
-mkdir "$DIR"charts/azure-service-operator/charts/azure-service-operator-crds/crds # create dirs for generated files
+rm -rf "$DIR"charts/azure-service-operator/charts/azure-service-operator-crds/templates
```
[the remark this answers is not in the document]
matthchr: yeah, that's a good point. The issue is if we don't put CRDs into the templates directory, they won't be overwritten. So possibly we need to do like what cert-manager is doing and document the uninstall process too and clearly call out you must delete the resources before uninstalling ASO. I'll update the docs.

super-harsh (approved):

PR #2365, opened 2022-06-28
Add leases access to leader role

matthchr:


scripts/make-multitenant-cluster.sh
```diff
@@ -26,6 +26,9 @@ yq eval "$query" "$source" > "$target"
 # Remove the aadpodidbinding label - this is only needed for communicating to ARM
 yq eval -i "del(select($deployment) | .spec.template.metadata.labels.aadpodidbinding)" "$target"
 
+# Edit the deployment to turn off leader election for the webhook only pod, as the webhooks don't wait for leader election anyway.
```
babbageclunk, line 29: I'm pretty sure that the parameter was only carried across for inertia/superstition reasons!

babbageclunk (approved):

PR #2366, opened 2022-06-29
Clearer error if AzureName is not set

### LINKED ISSUE #2362: Bad error message when AzureName is not set
If a webhook is misconfigured, `AzureName` can be empty. This results in something like:
> https://management.azure.com/subscriptions/sub/resourceGroups/rg/providers/Microsoft.Network/virtualNetworks/vnet/subnets/
      ERROR CODE UNAVAILABLE
      {
        "Message": "The requested resource does not support http method 'PUT'."
      }
      ---------------------------------------------------------------------

**Describe the improvement**
We should have an explicit check in the ARM reconciler and give a clear error if this is not set. It basically means that the webhooks aren't working right.

**Additional context**

matthchr:
Fixes #2362

No tests because there's not a great way to force the webhooks to not run and cause this to happen.


v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -296,6 +296,11 @@ func (r *azureDeploymentReconcilerInstance) MonitorDelete(ctx context.Context) (
 }
 
 func (r *azureDeploymentReconcilerInstance) BeginCreateOrUpdateResource(ctx context.Context) (ctrl.Result, error) {
+	if r.Obj.AzureName() == "" {
+		err := errors.New("AzureName was not set. A webhook should default this to .metadata.name if it was omitted. Is the ASO webhook service running?")
+		return ctrl.Result{}, conditions.NewReadyConditionImpactingError(err, conditions.ConditionSeverityWarning, core.UnknownErrorCode)
```
matthchr, line 301: Ah this is wrong, it should be an error with a known error code. WIll fix

theunrepentantgeek (approved):

PR #2378, opened 2022-07-08
Update submodules

theunrepentantgeek:

Updates our upstream submodules to a more recent version and modifies our configuration to match.

**cdn**
Breaking: Identity removed from cdn.Profile; according to PR[#18187](https://github.com/Azure/azure-rest-api-specs/pull/18187) it's not ready for GA and will return in a later preview API version.

**containerinstance**
Password is now correctly flagged with x-ms-secret and there are flow on effects, including type names and local variable changes.

**costmanagement**
We now find nothing when scanning the JSON schema; deferring further investigation until we're swagger only for our source data.

**dbformysql**
New optional properties DataEncryption and Identity added to FlexibleServer.

**operationalinsights**
Rename of property `eTag` to `etag` and documentation to match.

gif

**If applicable**:
- [x] this PR contains tests


v2/api/dbformysql/v1alpha1api20210501/flexible_server_types_gen.go
```diff
@@ -2380,6 +2653,132 @@ func (availability *HighAvailability_Status) AssignPropertiesToHighAvailabilityS
 	return nil
 }
 
+// Deprecated version of Identity. Use v1beta20210501.Identity instead
+type Identity struct {
+	Type                   *IdentityType      `json:"type,omitempty"`
+	UserAssignedIdentities map[string]v1.JSON `json:"userAssignedIdentities,omitempty"`
```
matthchr, line 2659: I think this is readonyl and should be removed in our config.

matthchr:
> Rename of ETag to Etag and documentation to match.

I think really (in the serialized payload) it's a rename from `eTag` to `etag`. Our users are primarily using YAML where they would've specified it as `eTag` (except it probably didn't work as it had the wrong casing). Nothing really for you to change here though other than possibly your note in the PR description

v2/api/dbformysql/v1alpha1api20210501/flexible_server_types_gen.go
theunrepentantgeek: Done.

Issue #2375, opened 2022-07-05
Improve performance of gopter tests


theunrepentantgeek:
**Describe the current behavior**

Our generated `gopter` tests are useful, having caught an occasional generator bug, but are taking too long to run. 

**Describe the improvement**

* Reduce `MaxSize` to 2 (was 10)  
  We only need to test the [] [x] and [x,y] cases for arrays & similar for maps, as we test individual objects manually

* Reduce `MinSuccessfulTests` to 50 for resource, spec, and status types
  Every test of a resource also tests spec and status, so we can reduce these counts while maintaining good coverage

theunrepentantgeek:
Closed by #2379

PR #2379, opened 2022-07-08
Limit number of tests for resources, spec and status types

theunrepentantgeek:

As discussed in #2375, we have scope to improve the performance of our gopter test suite by reducing the number of times we test particular types.

In local testing, this change reduced the time to test the single package `network\v1beta20201101` from ~15s per run to ~6s per run.

The generator code change is in the first commit.
The second commit includes only changes to generated code.

gif

**If applicable**:
- [ ] this PR contains tests


v2/tools/generator/internal/codegen/pipeline/json_serialization_test_cases.go
```diff
@@ -104,5 +103,13 @@ func (s *objectSerializationTestCaseFactory) AddTestTo(def astmodel.TypeDefiniti
 	isOneOf := astmodel.OneOfFlag.IsOn(def.Type()) // this is ugly but can’t do much better right now
 
 	testcase := testcases.NewJSONSerializationTestCase(def.Name(), container, isOneOf, s.idFactory)
+	if _, isResource := astmodel.AsResourceType(def.Type()); isResource {
+		// Don't need to test resources many times, the spec and status types are tested independently
+		testcase.SetMinSuccessfulTests(20)
```
super-harsh, line 108: are 20 and 80 below arbitrary numbers? or is there such a reason to use these specific numbers?

theunrepentantgeek: Relatively arbitrary - default is to run 100 trials, so 20xResource + 80xSped/Status adds to the default; Other than that, wanted to heavily bias towards testing Spec/Status in isolation.

v2/tools/generator/internal/codegen/pipeline/json_serialization_test_cases.go
```diff
@@ -104,5 +103,13 @@ func (s *objectSerializationTestCaseFactory) AddTestTo(def astmodel.TypeDefiniti
...
+	} else if astmodel.IsSpec(def.Name()) || astmodel.IsStatus(def.Name()) {
+		// Reduce count of Spec and Status tests to reflect those done by the resource tests
```
super-harsh, line 110: Maybe worth adding here about the comparison of `Reduce`. Maybe something to tell about the default being 100 and we're reducing to 80

super-harsh (approved):

v2/tools/generator/internal/codegen/pipeline/json_serialization_test_cases.go
super-harsh: Makes sense! Not sure if we should add a comment mentioning that?

v2/tools/generator/internal/codegen/pipeline/json_serialization_test_cases.go
theunrepentantgeek: I'll add some constants and comments.

matthchr:
Hmm, this doesn't actually seem to have reduced the overall test runtime very much? It's still about 15m in CI, which is _really_ long considering that the recording tests run in 3-4m in CI.




> Fri, 08 Jul 2022 05:04:45 GMT
task: [controller:test-cover] go test -short -tags=noexit -timeout 10m -covermode atomic -coverprofile=controller-coverage.out -json -coverpkg="./..." -run '.*' ./... > '/workspace/reports/controller-unit-tests.json'
Fri, 08 Jul 2022 05:04:46 GMT
[controller:test-cover] go: downloading github.com/kr/pretty v0.2.0
Fri, 08 Jul 2022 05:04:46 GMT
[controller:test-cover] go: downloading github.com/leanovate/gopter v0.2.8
Fri, 08 Jul 2022 05:04:46 GMT
[controller:test-cover] go: downloading github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.0.0
Fri, 08 Jul 2022 05:04:46 GMT
[controller:test-cover] go: downloading github.com/go-sql-driver/mysql v1.6.0
Fri, 08 Jul 2022 05:04:46 GMT
[controller:test-cover] go: downloading github.com/dnaeon/go-vcr v1.2.0
Fri, 08 Jul 2022 05:04:46 GMT
[controller:test-cover] go: downloading golang.org/x/crypto v0.0.0-20220511200225-c6db032c6c88
Fri, 08 Jul 2022 05:04:47 GMT
[controller:test-cover] go: downloading github.com/kr/text v0.2.0
Fri, 08 Jul 2022 05:04:47 GMT
[controller:test-cover] go: downloading github.com/AzureAD/microsoft-authentication-library-for-go v0.4.0
Fri, 08 Jul 2022 05:04:47 GMT
[controller:test-cover] go: downloading cloud.google.com/go v0.81.0
Fri, 08 Jul 2022 05:04:47 GMT
[controller:test-cover] go: downloading github.com/pkg/browser v0.0.0-20210911075715-681adbf594b8
Fri, 08 Jul 2022 05:04:47 GMT
[controller:test-cover] go: downloading github.com/golang-jwt/jwt v3.2.2+incompatible
Fri, 08 Jul 2022 05:21:38 GMT
task: [produce-markdown-summary] go run v2/tools/mangle-test-json/main.go "/workspace/reports/controller-unit-tests.json" > "/workspace/reports/controller-unit-tests.md"

It feels like there's a lot of duplication in the tests. For example we test:
`Test_ManagedCluster_WhenSerializedToJson_DeserializesAsEqual`
and
`Test_ManagedClusters_Spec_WhenSerializedToJson_DeserializesAsEqual`


but the first one does:
```
// AddRelatedPropertyGeneratorsForManagedCluster is a factory method for creating gopter generators
func AddRelatedPropertyGeneratorsForManagedCluster(gens map[string]gopter.Gen) {
	gens["Spec"] = ManagedClustersSpecGenerator()
	gens["Status"] = ManagedClusterStatusGenerator()
}
```

which if I understand correctly means that we generate ~100 specs/status for the first test, and then 100 more for the second? So that particular `Spec` is covered 200 times, right? (100 after your change?) _but_, the child properties of the `Spec` were covered even more times (another 100 for the first level of nesting, 200 for the second, etc), as we also test each of them individually. The more nested a property is, the more tests it has covering it, right?

On top of that, we also have 4 copies of these tests, for alpha, alphastorage, beta, and betastorage.

**My questions are:**
1. Would it make more sense to instead just run a single test on the resource with high-ish iteration count (200 or 500 or something, or a variable number based on the total number of transitive properties?), but no `Spec`, `Status` (or their properties) specific tests? Then you don't get the cascading effect where types that have been nested deeply at the 4th or 5th level have actually been covered 500 times each?
2. Or we could leave it as is and just set anything at lower nesting levels (below Spec/Status) to some low number like 10 or 20, and assume that they're covered sufficiently by their 20 + the Spec/Status's 80?
3. Do we need the JSON roundtrip tests for the storage types? Given that they're generated exactly from the non-storage type except everything is optional is there really any risk they can't JSON roundtrip?

I tend to lean towards 1 or 2 being at least a partial solution. We can also investigate turning off code coverage for these tests as based on my testing that at least doubles their runtime.

My 2c though is that we should try to drive them down to the 3-5m range in CI (in-line with the recording tests and other tests), which would in theory get our overall PR CI time down to under 30m again.

v2/tools/generator/internal/codegen/pipeline/json_serialization_test_cases.go
theunrepentantgeek: I've updated the code to use some constants.

theunrepentantgeek:
> It feels like there's a lot of duplication in the tests. 

Yes, there's a lot of duplication in the tests - but these are random fuzz tests, duplication is kind-of the point, as each iteration will use different data.

> So that particular Spec is covered 200 times, right? (100 after your change?) but, the child properties of the Spec were covered even more times (another 100 for the first level of nesting, 200 for the second, etc), as we also test each of them individually.

We make all our spec and status properties into pointers so that CRD validation correctly kicks in - and `gopter` will only generate a value for a pointer with a 50% probability (it generates a random bool), so we're not testing nested properties as much as you might think.

For resources, where the properties are values (not pointers), we _were_ testing Spec and Status 200x each, which is too much; this PR reduces the count to 100.  The resource, spec and status types are the largest types too, so reducing the counts on those should have the biggest effect.

But when testing the Spec, any nested objects will only be tested half the time, so approximately 50 times; any properties nested in those only ~25 times, and so on.

> Would it make more sense to instead just run a single test on the resource with high-ish iteration count

Because `gopter` doesn't always drill down through pointers, we'd get better coverage by testing everything with a low iteration count.

> Do we need the JSON roundtrip tests for the storage types?

Yes, for the same reason we need them for the API types - they're our line of defense against changes to the generator creating an edge case where things don't round trip as we expect - which would cause a bunch of odd errors in the operator that would only manifest for certain configurations of particular resources and would be really hard to diagnose.

PR #2387, opened 2022-07-15
Add recording target to the Taskfile

### LINKED ISSUE #2386: Add target for record-tests in taskfile

Currently we have a single target for running env tests which has very early timeout and not suitable for recording most of the tests. 

**Describe the improvement**

Should add a target in `taskfile` to record-tests which takes in a timeout parameter when running. Default would be set to 60m

matthchr:
This fixes #2386

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -289,10 +289,18 @@ tasks:
     deps: [controller:run-kustomize-for-envtest]
     cmds:
     # -race fails at the moment in controller-runtime
-    - go test -timeout 15m -run '{{default ".*" .TEST_FILTER}}' {{.VERBOSE}} ./internal/controllers
+    - go test -timeout {{.TIMEOUT}} -run '{{default ".*" .TEST_FILTER}}' {{.VERBOSE}} ./internal/controllers
     vars:
       VERBOSE:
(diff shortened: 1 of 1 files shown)


theunrepentantgeek (approved):

Taskfile.yml
```diff
@@ -294,6 +294,18 @@ tasks:
       VERBOSE:
         sh: if [ $TEST_FILTER ];  then echo "-v"; fi
 
+  # This target is intended for local use only
+  controller:test-integration-envtest-record:
```
super-harsh, line 298: I think would be a good idea to add TIMEOUT setting like `{{default "15m" .TIMEOUT}}` for `controller:test-integration-envtest` and then use something like below for this target. That'll make the exiting `controller:test-integration-envtest` configurable and avoid duplication here. 

```
   cmds:
      - task: controller:test-integration-envtest
        vars: {TIMEOUT: 1h}
```

super-harsh (changes requested):

Taskfile.yml
theunrepentantgeek: I like this idea - nice and DRY.

super-harsh (approved):

PR #2391, opened 2022-07-18
Fix property bag use for maps and slices

### LINKED ISSUE #2388: Bug: Asymetry in generated code results in round trip failures with PropertyBag
**Describe the bug**

We're not treating **nil** consistently between the two directions (`To`/`From`) of the `AssignProperties()` functions, resulting in some problems testing adjacent conversions in our conversion graph.

Discovered while testing the addition of a new version of `compute`, two adjacent versions of `VirtualMachineScaleSetExtension_Status` have a minor different that should be handled via a `PropertyBag`.

Version `v1beta20220301storage` of `VirtualMachineScaleSetExtension_Status`  has additional properties, including `ProtectedSettingsFromKeyVault`:

``` go
type VirtualMachineScaleSetExtension_Status struct {
	AutoUpgradeMinorVersion       *bool                  `json:"autoUpgradeMinorVersion,omitempty"`
	EnableAutomaticUpgrade        *bool                  `json:"enableAutomaticUpgrade,omitempty"`
	ForceUpdateTag                *string                `json:"forceUpdateTag,omitempty"`
	Id                            *string                `json:"id,omitempty"`
	Name                          *string                `json:"name,omitempty"`
	PropertiesType                *string                `json:"properties_type,omitempty"`
	PropertyBag                   genruntime.PropertyBag `json:"$propertyBag,omitempty"`
	ProtectedSettings             map[string]v1.JSON     `json:"protectedSettings,omitempty"`
	ProtectedSettingsFromKeyVault map[string]v1.JSON     `json:"protectedSettingsFromKeyVault,omitempty"`
	ProvisionAfterExtensions      []string               `json:"provisionAfterExtensions,omitempty"`
	ProvisioningState             *string                `json:"provisioningState,omitempty"`
	Publisher                     *string                `json:"publisher,omitempty"`
	Settings                      map[string]v1.JSON     `json:"settings,omitempty"`
	SuppressFailures              *bool                  `json:"suppressFailures,omitempty"`
	Type                          *string                `json:"type,omitempty"`
	TypeHandlerVersion            *string                `json:"typeHandlerVersion,omitempty"`
}
```

For comparison, version `v1beta20201201storage` of `VirtualMachineScaleSetExtension_Status`:

``` go
type VirtualMachineScaleSetExtension_Status struct {
	AutoUpgradeMinorVersion  *bool                  `json:"autoUpgradeMinorVersion,omitempty"`
	EnableAutomaticUpgrade   *bool                  `json:"enableAutomaticUpgrade,omitempty"`
	ForceUpdateTag           *string                `json:"forceUpdateTag,omitempty"`
	Id                       *string                `json:"id,omitempty"`
	Name                     *string                `json:"name,omitempty"`
	PropertiesType           *string                `json:"properties_type,omitempty"`
	PropertyBag              genruntime.PropertyBag `json:"$propertyBag,omitempty"`
	ProtectedSettings        map[string]v1.JSON     `json:"protectedSettings,omitempty"`
	ProvisionAfterExtensions []string               `json:"provisionAfterExtensions,omitempty"`
	ProvisioningState        *string                `json:"provisioningState,omitempty"`
	Publisher                *string                `json:"publisher,omitempty"`
	Settings                 map[string]v1.JSON     `json:"settings,omitempty"`
	Type                     *string                `json:"type,omitempty"`
	TypeHandlerVersion       *string                `json:"typeHandlerVersion,omitempty"`
}
```

When reading from the property bag, we check for the presence of a value, and assign **nil** if one isn't found:

``` go
// ProtectedSettingsFromKeyVault
if propertyBag.Contains("ProtectedSettingsFromKeyVault") {
	var protectedSettingsFromKeyVault map[string]v1.JSON
	err := propertyBag.Pull("ProtectedSettingsFromKeyVault", &protectedSettingsFromKeyVault)
	if err != nil {
		return errors.Wrap(err, "pulling 'ProtectedSettingsFromKeyVault' from propertyBag")
	}

	destination.ProtectedSettingsFromKeyVault = protectedSettingsFromKeyVault
} else {
	destination.ProtectedSettingsFromKeyVault = nil
}
```

But when writing to the property bag, we always write, regardless.

``` go
// ProtectedSettingsFromKeyVault
popertyBag.Add("ProtectedSettingsFromKeyVault", source.ProtectedSettingsFromKeyVault)
```

Always writing a **nil** works when we test conversion from `v1beta20220301storage` to the earlier version, and back again, as the value in the property bag gets consumed.

But when we test conversion from `v1beta20201201storage` to the later version, and back again, we end up with an extra value in the property bag representing the **nil** value for `ProtectedSettingsFromKeyVault`.

Essentially, we currently have two ways to represent a **nil** value with the property bag - by explicitly storing a **nil**, or by omitting the value entirely. 

**Expected behavior**

We should always store a **nil** value the same way when using a property bag - by omitting the value entirely.

Doing this requires us to generate slightly different code when storing the value in the bag:

``` go
// ProtectedSettingsFromKeyVault
if source.ProtectedSettingsFromKeyVault != nil {
	popertyBag.Add("ProtectedSettingsFromKeyVault", source.ProtectedSettingsFromKeyVault)
} else {
    propertyBag.Remove("ProtectedSettingsFromKeyVault")
}
```

The `Remove()` is necessary to ensure no other value is present (we may have that from chained conversions originating elsewhere).

theunrepentantgeek:

We already had checks to ensure we don't store explicit **nil** values in a `PropertyBag` when converting versions of a resource (we represent missing values by omitting them from the bag).

This PR extends this practice to empty maps and slices.

Also fixes a bug where an existing value would be left in the bag if our value was nil; now we explicitly remove it.

Closes #2388

The updates to the golden tests show all the new cases.

gif

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/tools/generator/internal/conversions/property_conversions.go
+++ b/v2/tools/generator/internal/conversions/property_conversions.go
@@ -195,10 +195,20 @@ func NameOfPropertyAssignmentFunction(
 //
 // <propertyBag>.Add(<propertyName>, <source>)
 //
-// For optional sources, the value is only added if non-nil
+// For optional sources, the value is only added if non-nil; if nil, we remove any existing item
 //
 // if <source> != nil {
 //     <propertyBag>.Add(<propertyName>, *<source>)
+// } else {
+//    <propertyBag>.Remove(<propertyName>)
+// }
+//
+// For slice and slice sources, the value is only added if it is non-empty; if empty we remove any existing item
+//
+// if len(<source>) > 0 {
+//	   <propertyBag>.Add(<propertyName>, <source>)
+// } else {
+//	   <propertyBag>.Remove(<propertyName>)
 // }
 //
 func writeToBagItem(
@@ -229,6 +239,9 @@ func writeToBagItem(
 		return nil, nil
 	}
 
+	_, sourceIsMap := astmodel.AsMapType(actualSourceType)
+	_, sourceIsSlice := astmodel.AsArrayType(actualSourceType)
+
 	return func(reader dst.Expr, _ func(dst.Expr) []dst.Stmt, knownLocals *astmodel.KnownLocalsSet, generationContext *astmodel.CodeGenerationContext) []dst.Stmt {
 		createAddToBag := func(expr dst.Expr) dst.Stmt {
 			addToBag := astbuilder.InvokeQualifiedFunc(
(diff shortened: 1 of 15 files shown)


v2/tools/generator/internal/conversions/property_conversions.go
```diff
@@ -195,10 +195,20 @@ func NameOfPropertyAssignmentFunction(
 //
 // <propertyBag>.Add(<propertyName>, <source>)
 //
-// For optional sources, the value is only added if non-nil
+// For optional sources, the value is only added if non-nil; if nil, we remove any existing item
 //
 // if <source> != nil {
 //     <propertyBag>.Add(<propertyName>, *<source>)
+// } else {
```
matthchr, line 202: I'm surprised that this change only shows up in 3 places in the actual generated types. Is that expected? I would have thought we had more optional sources than that.

matthchr (approved):
LGTM, just had one question (not even a comment, just trying to understand)

v2/tools/generator/internal/conversions/property_conversions.go
theunrepentantgeek: It's not just that the source is optional - it's that the field is present in one storage version of the resource and not in another storage version; this is pretty rare at the moment because we only have one version of most resources.

PR #2402, opened 2022-07-20
Improve the readability of our resource registration function

theunrepentantgeek:

Our resource registration file (`controller_resources_gen.go`) was harder to read than it could be.

* Batch addition of resources from the same package into a single `append()` call
* Remove initialization of the Indexes field if the slice is empty, reducing the overall size of the file
* Remove initialization of the Watches field if the slice is empty, also reducing the size of the file
* Lean into helper methods from `astbuilder` more

gif


v2/internal/controllers/controller_resources_gen.go
```diff
@@ -628,52 +484,46 @@ func getKnownTypes() []client.Object {
...
-	result = append(result, new(cdn_v20210601s.ProfilesEndpoint))
-	result = append(result, new(compute_alpha20200930.Disk))
-	result = append(result, new(compute_alpha20200930.Snapshot))
-	result = append(result, new(compute_alpha20200930s.Disk))
-	result = append(result, new(compute_alpha20200930s.Snapshot))
-	result = append(result, new(compute_alpha20201201.VirtualMachine))
-	result = append(result, new(compute_alpha20201201.VirtualMachineScaleSet))
-	result = append(result, new(compute_alpha20201201s.VirtualMachine))
-	result = append(result, new(compute_alpha20201201s.VirtualMachineScaleSet))
+	result = append(
```
babbageclunk, line 487: This grouping is nice.

babbageclunk (approved):

PR #2403, opened 2022-07-20
Add option to list namespaced secrets

### LINKED ISSUE #2398: Bug: Secrets can trigger reconciliation events for resources in a different namespace
**Version of Azure Service Operator**: ASOv1

**Describe the bug**
We watch secrets in order to trigger a reconciliation when a secret is updated (for key rollover events). If a secret exists in two different namespaces `ns1` and `ns2` with the same name `s1`, and we have a resource in `ns1` associated with secret `s1` and a separate resource in `ns2` associated with `s1` (in `ns2`), changes in either secret trigger events for both resources.

The [code in question](https://github.com/Azure/azure-service-operator/blob/main/v2/internal/controllers/controller_resources.go#L291) lacks a guard for namespace.

In envtest we wrote our own indexer, and [that code](https://github.com/Azure/azure-service-operator/blob/main/v2/internal/testcommon/kube_client_no_cache.go#L106) will also need to be fixed so that we can actually query by namespace. See [the upstream code](https://github.com/kubernetes-sigs/controller-runtime/blob/f46919744bee01060c9084a285e049afffd38c9d/pkg/cache/informer_cache.go#L180)

The fix for this is as follows:
1. Update the `watchSecrets` code to do `err := c.List(ctx, matchingResources, client.MatchingFields{key: o.GetName()}, client.InNamespace(o.GetNamespace()))`
2. Update our no_cache client to either generate an index key for "AllNamespaces" and also for the particular namespace. I am not actually sure this is required though given we do seem to pass the other filtering options to `filteredOpts` in our `noCachingList` test helper...

So possibly the fix is just `err := c.List(ctx, matchingResources, client.MatchingFields{key: o.GetName()}, client.InNamespace(o.GetNamespace()))`

super-harsh:
Closes #2398


theunrepentantgeek (approved):

matthchr:
We should probably have a test that ensures that this behaves correctly? Happy to merge this now as it's correct I think, but maybe file an issue for it so we don't forget to write a test that covers this case in a follow-on PR.

PR #2409, opened 2022-07-22
Update version of Compute resources to 2022-03-01

### LINKED ISSUE #2368: Add ability to delete osDisk when VMs are deleted

The imageReference property on a virtualMachine allows a VM to be created with a specific image and OS disk. 

**Describe the improvement**

VMs support having their osDisk deleted when they themselves are deleted. The property is deleteOption under osDisk - https://docs.microsoft.com/en-us/azure/virtual-machines/delete?tabs=portal2%2Cportal3%2Cportal4.

Please can this be supported. Thanks

theunrepentantgeek:

Updates the available versions of several compute resources to version 2022-03-01.

Tests for existing versions were renamed to reflect the API version, to allow the new tests to sit alongside.

Added a helper method for exporting a Sample YAML file from a test - while this file requires further editing, it's easier than hand crafting a sample with the bonus that it's known to work.

Resolves #2368 

Most of the changes are to generated files and test recordings:

* `api/compute` (excluding new packages)  
  41 files changed, 25034 insertions(+), 3820 deletions(-) 
* `api/compute/v1beta20220301`  
  21 files changed, 93646 insertions(+)
* `api/compute/v1beta20220301storage`  
  8 files changed, 23925 insertions(+)
* `internal/controllers/recordings`  
  6 files changed, 5557 insertions(+), 746 deletions(-)

gif

**If applicable**:
- [x] this PR contains tests


v2/config/samples/compute/v1beta20220301_image.yaml
```diff
@@ -0,0 +1,30 @@
...
+            bloburi: null
+            caching: null
+            diskencryptionset: null
+            disksizegb: 32
+            manageddisk: null
+            osstate: Generalized
+            ostype: Linux
+            snapshot:
+                reference:
+                    armid: /subscriptions/82acd5bb-4206-47d4-9c12-a65db028483d/resourceGroups/asotest-rg-hitcxz/providers/Microsoft.Compute/snapshots/asotest-snapshot-pcngac
```
super-harsh, line 26: maybe you don't want to keep subID here.. better to replace it with `00000000-0000-0000-0000-000000000000`

v2/config/samples/compute/v1beta20220301_image.yaml
```diff
@@ -0,0 +1,30 @@
...
+            storageaccounttype: null
+        zoneresilient: null
+    tags: {}
```
super-harsh, line 29: do we need the nulls and empty tags?

v2/config/samples/compute/v1beta20220301_virtualmachine.yaml
```diff
@@ -0,0 +1,73 @@
+apiVersion: compute.azure.com/v1beta20220301
+kind: VirtualMachine
+metadata:
+  name: samplevm
+  namespace: default
+spec:
+    additionalcapabilities: null
+    applicationprofile: null
+    availabilityset: null
+    azurename: asotest-vm-ljhvpw
```
super-harsh, line 10: maybe we want to get rid of this?

v2/config/samples/compute/v1beta20220301_virtualmachine.yaml
```diff
@@ -0,0 +1,73 @@
...
+    networkprofile:
+        networkapiversion: null
+        networkinterfaceconfigurations: []
+        networkinterfaces:
+            - deleteoption: null
+              primary: null
+              reference:
+                group: network.azure.com
+                kind: NetworkInterface
+                name: asotest-nic-dajxvb
```
super-harsh, line 37: would be better to match this with name in `NetworkInterface` sample

v2/config/samples/compute/v1beta20220301_virtualmachine.yaml
```diff
@@ -0,0 +1,73 @@
...
+                armid: ""
+    osprofile:
+        adminpassword:
+            name: asotest-vmsecret-jnczcw
```
super-harsh, line 41: Same as above for this one and all `asotest` names below.

v2/config/samples/compute/v1beta20220301_virtualmachinescaleset.yaml
```diff
@@ -0,0 +1,46 @@
...
+kind: VirtualMachineScaleSet
+metadata:
+  name: samplevmss
+  namespace: default
+spec:
+  location: westcentralus
+  owner:
+    name: aso-sample-rg
+  sku:
+    name: standard_d1_v2
```
super-harsh, line 11: Not sure if this would go through in lower case.. Ideally should be ``STANDARD_D1_v2`

v2/config/samples/compute/v1beta20220301_image.yaml
```diff
@@ -0,0 +1,30 @@
...
+metadata:
+  name: aso-sample-image
+  namespace: default
+spec:
+    azurename: asotest-image-hkobbe
+    extendedlocation: null
+    hypervgeneration: V2
+    location: westus2
+    owner:
+        name: asotest-rg-hitcxz
```
super-harsh, line 12: replace with `aso-sample-rg` to align with other samples

v2/config/samples/compute/v1beta20220301_image.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: compute.azure.com/v1beta20220301
+kind: Image
+metadata:
+  name: aso-sample-image
+  namespace: default
+spec:
+    azurename: asotest-image-hkobbe
```
super-harsh, line 7: maybe we don't want `azurename` ?

v2/internal/controllers/export_sample_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v3"
+	"k8s.io/apimachinery/pkg/runtime"
+	"os"
+	"path"
+)
+
+// ExportAsSample exports the provided resource to a YAML file as the basis for a sample
+// This is used to generate the samples for the repo
+func ExportAsSample(tc *testcommon.KubePerTestContext, resource runtime.Object) {
```
super-harsh, line 20: Love this!

super-harsh (changes requested):
Have left a few comments for the added samples, otherwise LGTM.

v2/config/samples/compute/v1beta20220301_image.yaml
theunrepentantgeek: Done.

v2/config/samples/compute/v1beta20220301_image.yaml
theunrepentantgeek: I like them because they show the full list of available properties.

v2/config/samples/compute/v1beta20220301_virtualmachine.yaml
theunrepentantgeek: Done.

v2/config/samples/compute/v1beta20220301_virtualmachinescaleset.yaml
theunrepentantgeek: That's the casing that comes back from the service, so I'm pretty sure it works.

v2/config/samples/compute/v1beta20220301_virtualmachine.yaml
theunrepentantgeek: Changed all `asotest` to `asosample`.

v2/config/samples/compute/v1beta20220301_image.yaml
theunrepentantgeek: Done.

v2/config/samples/compute/v1beta20220301_virtualmachinescaleset.yaml
matthchr: I think they're case insensitive on this one (it's not an enum like many other fields, it's just a raw string)

v2/config/samples/compute/v1beta20220301_image.yaml
```diff
@@ -0,0 +1,29 @@
...
+    location: westus2
+    owner:
+        name: aso-sample-rg
+    sourcevirtualmachine: null
+    storageprofile:
+        datadisks: []
+        osdisk:
+            bloburi: null
+            caching: null
+            diskencryptionset: null
```
matthchr, line 18: Is this casing really correct? Shouldn't it be `diskEncryptionSet`?

v2/config/samples/compute/v1beta20220301_virtualmachine.yaml
```diff
@@ -0,0 +1,72 @@
...
+              reference:
+                group: network.azure.com
+                kind: NetworkInterface
+                name: samplenic
+    osprofile:
+        adminpassword:
+            name: asosample-vmsecret
+            key: password
+        adminusername: bloom
+        allowextensionoperations: null
```
matthchr, line 43: Casing here also looks wrong to me

v2/internal/controllers/crd_compute_vm_20201201_test.go
```diff
@@ -98,10 +98,12 @@ func newVMNetworkInterface(tc *testcommon.KubePerTestContext, owner *genruntime.
 	}
 }
 
-func Test_Compute_VM_CRUD(t *testing.T) {
+func Test_Compute_VM_20201201_CRUD(t *testing.T) {
 	t.Parallel()
 
 	tc := globalTestContext.ForTest(t)
+	tc.AzureRegion = to.StringPtr("westeurope")
```
matthchr, line 105: Leave a note as to why this was done, and/or that it's temporary and presumably capacity related?

v2/internal/controllers/crd_compute_vmss_20201201_test.go
```diff
@@ -247,10 +250,13 @@ func Test_Compute_VMSS_CRUD(t *testing.T) {
 	tc.Expect(found).To(BeTrue())
 
 	// Delete VMSS
+	// We do this in stages to avoid race conditions between owner/owned resources
+	// that can make tests non-deterministic
 	tc.DeleteResourceAndWait(vmss)
+	tc.DeleteResourceAndWait(rg)
```
matthchr, line 256: `rg` wasn't being deleted here before, so you don't actually need to delete it here at all? It'll get cleaned up automatically by the test GC process so you can just not delete the rg in the test (like before)

v2/internal/controllers/crd_compute_vmss_20220301_test.go
```diff
@@ -0,0 +1,170 @@
...
+			found = true
+		}
+	}
+	tc.Expect(found).To(BeTrue())
+
+	// Delete VMSS
+	// We do this in stages to avoid race conditions between owner/owned resources
+	// that can make tests non-deterministic
+	tc.DeleteResourceAndWait(vmss)
+	tc.DeleteResourceAndWait(rg)
```
matthchr, line 163: Same here, no need to delete the RG directly here I dont think.

v2/internal/controllers/export_sample_test.go
matthchr: Where is this actually called though?

v2/internal/testcommon/kube_per_test_context.go
```diff
@@ -143,6 +144,8 @@ func (ctx KubeGlobalContext) ForTestWithConfig(t *testing.T, cfg config.Values)
 	context := context.Background() // we could consider using context.WithTimeout(OperationTimeout()) here
 	match := NewKubeMatcher(verify, context)
 
+	format.MaxLength = 0 // Disable output truncation
```
matthchr, line 147: Yeah this has annoyed me in the past too. Happy to keep it set like this as it's better IMO

v2/config/samples/compute/v1beta20220301_image.yaml
super-harsh: I agree, casing does not look alright for most of the fields in the sample.

v2/config/samples/compute/v1beta20220301_image.yaml
theunrepentantgeek: Done.

v2/internal/controllers/crd_compute_vmss_20201201_test.go
theunrepentantgeek: I added this to be consistent with other tests that explicitly delete the resource group already.

matthchr: Today we're not all that consistent (some tests delete at the end, some don't. Some use defer, some don't). Im actually going to be removing the extra deletes in a future PR where I fix the parallel delete problem, which is why I mentioned it here (because these will end up removed in that PR). The test framework itself always issues these deletes when the test completes, so there's no need to do it explicitly here.

The intent is that the tests can just be focused on the actions applicable to the test at hand, and the ancillary stuff like creating a namespace, deleting RG when done, etc are taken care of for free by the framework.

v2/config/samples/compute/v1beta20220301_virtualmachinescaleset.yaml
```diff
@@ -0,0 +1,55 @@
...
+        primary: true
+    osProfile:
+      adminUsername: adminUser
+      computerNamePrefix: computer
+      linuxConfiguration:
+        disablePasswordAuthentication: true
+        ssh:
+          publicKeys:
+          - keyData: |
+              ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDknXTIQcOeg3Fg/j9ZYK45Sl6PbBdxZrhm7CwOgGMqi6txxhPYXUKJahHvtPBDQFN0cyt0B8jfiey2zU7R0WgXGgBxL9cJYuke1o5DpjdCb4c0hmuqZI8kCfxh5kCaMMn+0JCoAyipCkYQzzhiNoF7/19aW9NZrPa/ybEl1aUy4dkkRkNQIzlwg9zr8PSHHFRUof7+EOtK6s3CHo/n8gWMR912Bg3kNn/Lq1kct+iMTA0S7da0qWXlDNIU0R3fTqCt6vkHVhkOBZYbhC2ZbwwI0u6rkUddy6ju6tnLrzq4GUNOh/Qp3ofItDew7Uwgb2DQKcyGb79fq04LbGpLRRxz
```
super-harsh, line 48: Minor: Is it a good idea to include the ssh public key here? 

or we could have something like `ssh-rsa AAAAB  # Specify your SSH public key here` <-- this is from the existing sample we have.

super-harsh (approved):

PR #2415, opened 2022-07-25
Add SecretWatcher namespace test

### LINKED ISSUE #2408: Add tests for secretWatcher
**Describe the improvement**
This is related to #2398. As suggested by @matthchr, we need to add tests for secretWatcher which we currently don't have.

super-harsh:
Closes #2408 

Added test for user secret when secret exist in different namespace and secret watcher can not list it. 

The test added is for issue [#2398](https://github.com/Azure/azure-service-operator/issues/2398)


v2/internal/controllers/to_azure_secrets_test.go
```diff
@@ -66,10 +71,68 @@ func Test_MissingSecretKey_ReturnsError(t *testing.T) {
...
-	tc.Expect(vm.Status.Conditions[0].Reason).To(Equal("SecretNotFound"))
+	tc.Expect(vm.Status.Conditions[0].Reason).To(Equal(conditions.ReasonSecretNotFound))
 	tc.Expect(vm.Status.Conditions[0].Message).To(
 		ContainSubstring("Secret \"%s/%s\" does not contain key \"%s\"", tc.Namespace, secret.Name, secret.Key))
 
 	// Delete VM and resources.
 	tc.DeleteResourcesAndWait(vm, networkInterface, subnet, vnet, rg)
 }
+
+func Test_UserSecretInDifferentNamespace_SecretNotFound(t *testing.T) {
```
matthchr, line 82: I don't think that this test will detect #2398 (though it's a fine test to have).

To detect the issue with #2398, what we need is 2 of the same kind of resource, each in different namespaces, where each is looking at a secret with the same name. Something like:

1. Create vm1 in ns1, with secret named "mysecret" also in ns1. Allow this to be created fully and get to successfully provisioned.
2. Create vm2 in ns2, with secret ALSO named "mysecret" but in ns2. Allow this to be created fully and to get successfully provisioned.
3. Ensure that step 2 doesn't trigger a reconcile of vm1.

Item 3 is the "hard part" because we don't have great visibility into what reconciliations have happened in the tests, and it's hard to get that visibility in a way that will work in both envtest and real k8s clusters. You can easily see from logs, but that's not that helpful. Because nothing has changed you won't be able to tell by looking at the Ready condition of vm1.

The best idea I have to deal with item 3 is: use the azure client to go behind the back and delete vm1 before creating vm2. The operator won't realize it's been deleted. Then create vm2. Then confirm vm1 is still deleted in Azure. If vm1 was recreated in Azure, that means that reconcile was triggered which is unexpected: test fails.

matthchr (changes requested):

v2/internal/controllers/crd_dbformysql_flexibleserver_test.go
```diff
@@ -22,50 +21,12 @@ func Test_DBForMySQL_FlexibleServer_CRUD(t *testing.T) {
 	t.Parallel()
 	tc := globalTestContext.ForTest(t)
 
-	//location := tc.AzureRegion Capacity crunch in West US 2 makes this not work when live
-	location := "westcentralus"
-
 	rg := tc.CreateTestResourceGroupAndWait()
-
+	secrertName := "mysqlsecret"
```
matthchr, line 25:
```suggestion
	secretName := "mysqlsecret"
```

v2/internal/controllers/to_azure_secrets_test.go
```diff
@@ -66,10 +69,121 @@ func Test_MissingSecretKey_ReturnsError(t *testing.T) {
...
+	t.Parallel()
+	ns1 := "ns-1"
+	ns2 := "ns-2"
+
+	cfg, err := config.ReadFromEnvironment()
+	if err != nil {
+		t.Fatal(err)
+	}
+	cfg.TargetNamespaces = []string{ns1, ns2}
+	cfg.SyncPeriod = nil
```
matthchr, line 122: Rather than putting this here, can we do one of two things?

1. Change `ForTestWithConfig` to force this parameter to be `nil` always? `ForTest` does it right now so you could just move that.
2. Make a test specific `ReadFromEnvironmentForTest()` that forces it to be nil. 

The small advantage of 2 is that it means if you really wanted to leave this value set you can actually do it by calling `ReadFromEnvironment` instead of the test one. As part of doing 2 you'd wanna move all the calls to `ReadFromEnvironment` to `ReadFromEnvironmentForTest` in the test package.

v2/internal/controllers/to_azure_secrets_test.go
```diff
@@ -66,10 +69,121 @@ func Test_MissingSecretKey_ReturnsError(t *testing.T) {
...
+	tc.Expect(vm.Status.Conditions[0].Message).To(
+		ContainSubstring("failed resolving secret references: %s/%s does not exist", tc.Namespace, secretInDiffNamespace.Name))
+
+	// Delete VM and resources.
+	tc.DeleteResourcesAndWait(vm, networkInterface, subnet, vnet, rg)
+}
+
+func Test_UserSecretInDifferentNamespace_ShouldNotTriggerReconcile(t *testing.T) {
+	t.Parallel()
+	ns1 := "ns-1"
```
matthchr, line 114: minor: I think name these namespaces something more closely affiliated with this test... maybe: `secret-in-different-namespace-ns-1` and `secret-in-different-namespace-ns-2`?

v2/internal/controllers/crd_dbformysql_flexibleserver_test.go
```diff
@@ -112,6 +73,37 @@ func Test_DBForMySQL_FlexibleServer_CRUD(t *testing.T) {
 	tc.Expect(exists).To(BeFalse())
 }
 
+func newFlexibleServer(tc *testcommon.KubePerTestContext, owner *genruntime.KnownResourceReference, secretRef genruntime.SecretReference) (*mysql.FlexibleServer, string) {
```
matthchr, line 76: minor: I think call the `secretRef` parameter something like `adminPasswordSecretRef` so it's clear what it's going to be used for.

matthchr: Also instead of taking `*genruntime.KnownResourceReference` maybe consider taking a `ResourceGroup` and then calling the `testcommon.AsOwner(rg)` in this helper? That way it's a bit clearer to the caller what the owner actually _is_.

v2/internal/controllers/to_azure_secrets_test.go
```diff
@@ -66,10 +69,121 @@ func Test_MissingSecretKey_ReturnsError(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	cfg.TargetNamespaces = []string{ns1, ns2}
+	cfg.SyncPeriod = nil
+	tc := globalTestContext.ForTestWithConfig(t, cfg)
+
+	createNamespaces(tc, ns1, ns2)
+
+	secretName := tc.Namer.GenerateName("server-secret")
+	secretInCurrentNamespace := createNamespacedSecret(tc, ns1, secretName)
```
matthchr, line 128: minor: `ns1Secret` and `ns2Secret` maybe?

v2/internal/controllers/to_azure_secrets_test.go
```diff
@@ -66,10 +68,120 @@ func Test_MissingSecretKey_ReturnsError(t *testing.T) {
...
+	tc := globalTestContext.ForTestWithConfig(t, cfg)
+
+	createNamespaces(tc, ns1, ns2)
+
+	secretName := tc.Namer.GenerateName("server-secret")
+	ns1Secret := createNamespacedSecret(tc, ns1, secretName)
+	ns2Secret := createNamespacedSecret(tc, ns2, secretName)
+
+	rg := tc.NewTestResourceGroup()
+	rg.Namespace = ns1
```
theunrepentantgeek, line 130: Consider modifying `NewResourceGroup()` so you can optionally pass a namespace.

v2/internal/controllers/to_azure_secrets_test.go
```diff
@@ -66,10 +68,120 @@ func Test_MissingSecretKey_ReturnsError(t *testing.T) {
...
+
+	server2, _ := newFlexibleServer(tc, rg2, ns2Secret)
+	server2.ObjectMeta = metav1.ObjectMeta{
+		Name:      tc.Namer.GenerateName("mysql2"),
+		Namespace: ns2,
+	}
+
+	tc.CreateResourceAndWait(server2)
+
+	// Here we want to make sure that the server1 we deleted in Azure should not be created again with reconcile of secret
```
theunrepentantgeek, line 161:
```suggestion
	// Here we want to make sure that the server1 we deleted from Azure was not created again by a reconcile triggered when the second secret was created
```

v2/internal/controllers/to_azure_secrets_test.go
super-harsh: done!

v2/internal/controllers/to_azure_secrets_test.go
super-harsh: done for this and all below!

v2/internal/controllers/to_azure_secrets_test.go
```diff
@@ -79,10 +80,118 @@ func Test_MissingSecretKey_ReturnsError(t *testing.T) {
...
 }
+
+func Test_UserSecretInDifferentNamespace_SecretNotFound(t *testing.T) {
+	t.Parallel()
+
+	tc := globalTestContext.ForTest(t)
+	secretName := tc.Namer.GenerateName("admin-pw")
+	nameSpaceName := "secret-namespace"
+
+	createNamespaces(tc, nameSpaceName)
```
matthchr, line 98: Consider refactoring this to be a method on `tc`. Can probably reuse or mostly reuse `createTestNamespace` and then have that call this new `CreateTestNamespaces` helper?

matthchr: can also use that helper below in the other new test method

v2/internal/controllers/to_azure_secrets_test.go
```diff
@@ -79,10 +80,118 @@ func Test_MissingSecretKey_ReturnsError(t *testing.T) {
...
+		Name:      tc.Namer.GenerateName("mysql2"),
+		Namespace: ns2,
+	}
+
+	tc.CreateResourceAndWait(server2)
+
+	// Here we want to make sure that the server1 we deleted from Azure was not created again by a reconcile triggered when the second secret was created
+	// in other namespace
+	_, err = tc.AzureClient.GetByID(tc.Ctx, resourceID, server1.GetAPIVersion(), server1)
+	tc.Expect(err).ToNot(BeNil())
```
matthchr, line 174: Test for a specific `NotFound` error?

matthchr: Also consider issuing a fresh GET on `server1` in k8s and checking that its Ready condition is the same as before as well. Otherwise there could be a race where the operator is in the process of creating that new server but it's not yet done creating in Azure so this GET fails but really it did reconcile. If you also check Ready condition and that condition is _still_ Ready (and not `Reconciling` or something like that, then you know we're good.

v2/internal/controllers/to_azure_secrets_test.go
```diff
@@ -79,10 +80,118 @@ func Test_MissingSecretKey_ReturnsError(t *testing.T) {
...
+	server1.ObjectMeta = metav1.ObjectMeta{
+		Name:      tc.Namer.GenerateName("mysql"),
+		Namespace: ns1,
+	}
+	tc.CreateResourceAndWait(server1)
+
+	resourceID := genruntime.GetResourceIDOrDefault(server1)
+	tc.Expect(resourceID).ToNot(BeEmpty())
+
+	// Deleting server1 here using AzureClient so that Operator does not know about the deletion
```
matthchr, line 157: Minor: suggest expanding this to: "Deleting server1 here using AzureClient so that Operator does not know about the deletion. If operator reconciles that resource again it will be recreated and we will notice"

v2/internal/controllers/to_azure_secrets_test.go
```diff
@@ -79,10 +80,118 @@ func Test_MissingSecretKey_ReturnsError(t *testing.T) {
...
+	_, err = tc.AzureClient.DeleteByID(tc.Ctx, resourceID, server1.GetAPIVersion())
+	if err != nil {
+		return
+	}
+
+	server2, _ := newFlexibleServer(tc, rg2, ns2Secret)
+	server2.ObjectMeta = metav1.ObjectMeta{
```
matthchr, line 164: minor but might make our life easier when re-recording this: I recall that flexible servers can be a bit slow to create and delete. It might make more sense to do this with VMs instead if they are faster to provision/delete?

If you don't wanna do that now you could also just leave a TODO in the code that if we have trouble with this test runtime in the future we can swap it to VMs.

v2/internal/testcommon/kube_per_test_context.go
```diff
@@ -95,8 +95,15 @@ func (tc KubePerTestContext) MakeReferenceFromResource(resource client.Object) *
 }
 
 func (tc KubePerTestContext) NewTestResourceGroup() *resources.ResourceGroup {
+	return tc.NewTestResourceGroupWithNamespace(tc.Namespace)
+}
+
+func (tc KubePerTestContext) NewTestResourceGroupWithNamespace(namespace string) *resources.ResourceGroup {
```
matthchr, line 101: I don't think we really need this: Instead you can just modify the returned `resources.ResourceGroup` before creating it.

```
rg := tc.NewTestResourceGroup()
rg.Namespace = ns1
```

v2/internal/controllers/to_azure_secrets_test.go
```diff
@@ -79,10 +80,118 @@ func Test_MissingSecretKey_ReturnsError(t *testing.T) {
...
+
+	rg := tc.CreateTestResourceGroupAndWait()
+
+	vnet := newVMVirtualNetwork(tc, testcommon.AsOwner(rg))
+	subnet := newVMSubnet(tc, testcommon.AsOwner(vnet))
+	networkInterface := newVMNetworkInterface(tc, testcommon.AsOwner(rg), subnet)
+
+	tc.CreateResourceAndWait(vnet)
+	tc.CreateResourceAndWait(subnet)
+	tc.CreateResourceAndWait(networkInterface)
```
matthchr, line 109: minor: You should be able to create all 3 of these in parallel and have it work

v2/internal/controllers/to_azure_secrets_test.go
```diff
@@ -79,10 +80,118 @@ func Test_MissingSecretKey_ReturnsError(t *testing.T) {
...
+func Test_UserSecretInDifferentNamespace_ShouldNotTriggerReconcile(t *testing.T) {
+	t.Parallel()
+	ns1 := "secret-in-different-namespace-ns-1"
+	ns2 := "secret-in-different-namespace-ns-2"
+
+	cfg, err := testcommon.ReadFromEnvironmentForTest()
+	if err != nil {
+		t.Fatal(err)
+	}
+	cfg.TargetNamespaces = []string{ns1, ns2}
```
matthchr, line 132: Oh, wait. You don't actually have to set this/want to set this. This option is for the multitenancy option where it says "only watch these namespaces". So you should be able to just use a normal environment and make your extra namespaces without registering them?

matthchr: You were having an issue when you did that... with logger lookup being namespace-based I think? We might need to resolve that in order to get these merged

matthchr, line 132: So technically, this `TargetNamespaces` setting is for the multitenancy stuff that @babbageclunk added. The reason that adding this here helps/works is that we add logger-mappings for these namespaces too.

There's not currently another way to get logger mappings added for the new namespaces except doing this, but we should probably leave a comment here explaining that this is just for the logger mapping stuff and we're not actually doing any multitenancy.

super-harsh: Done!

v2/internal/controllers/to_azure_secrets_test.go
super-harsh: Done!

v2/internal/testcommon/kube_per_test_context.go
super-harsh: Done!

v2/internal/controllers/to_azure_secrets_test.go
super-harsh: Done!

v2/internal/controllers/to_azure_secrets_test.go
super-harsh: Done!

v2/internal/controllers/to_azure_secrets_test.go
super-harsh: Done!

v2/internal/controllers/to_azure_secrets_test.go
super-harsh: Done!

v2/internal/controllers/to_azure_secrets_test.go
```diff
@@ -95,7 +99,7 @@ func Test_UserSecretInDifferentNamespace_SecretNotFound(t *testing.T) {
 	secretName := tc.Namer.GenerateName("admin-pw")
 	nameSpaceName := "secret-namespace"
 
-	createNamespaces(tc, nameSpaceName)
+	tc.CreateTestNamespaces(nameSpaceName)
```
matthchr, line 102: minor: I know you didn't write this but `namespace` is all one word so this should be `namespaceName`

v2/internal/testcommon/kube_per_test_context.go
```diff
@@ -561,6 +554,16 @@ func (tc *KubePerTestContext) ExpectSecretHasKeys(name string, expectedKeys ...s
 	}
 }
 
+func (tc *KubePerTestContext) CreateTestNamespaces(names ...string) {
+	for _, name := range names {
+		tc.CreateResourceUntracked(&corev1.Namespace{
```
matthchr, line 559: Call `createTestNamespace` in here rather than duplicating its functionality?

matthchr: Though actually looking at `createTestNamespace` it's written a little weird. Still think it's worth reusing it rather than not though

v2/internal/controllers/to_azure_secrets_test.go
```diff
@@ -79,10 +84,131 @@ func Test_MissingSecretKey_ReturnsError(t *testing.T) {
...
+	_ = createNamespacedVM(tc, rg2, ns2, ns2Secret)
+
+	// Here we want to make sure that the server1 we deleted from Azure was not created again by a reconcile triggered when the second secret was created
+	// in other namespace
+	_, err = tc.AzureClient.GetByID(tc.Ctx, resourceID, vm1.GetAPIVersion(), vm1)
+	tc.Expect(err).To(HaveOccurred())
+	tc.Expect(genericarmclient.IsNotFoundError(err)).To(BeTrue())
+
+	tc.GetResource(client.ObjectKeyFromObject(vm1), vm1)
+	tc.Expect(vm1.Status.Conditions[0].Type).To(BeEquivalentTo(conditions.ConditionTypeReady))
```
matthchr, line 176: You also need to assert that `Conditions` has at least 1 item (or else `[0]` access here will panic).
Also need to assert that the status of the condition is `True`. The `Type` of the condition will always be ready, but it might be `ready = false`, which would indicate a failure of the test.

matthchr (approved):
Nice! Thanks for doing this. I just merged main into this which hopefully will let it pass with Bevan's CI fixes.

PR #2419, opened 2022-07-26
Add helper method GetAPIVersion to genruntime

### LINKED ISSUE #2416: API version should be consistent when interacting with ARM
**Describe the bug**

Observed while debugging PR #2409, some of the URLs showing in our recordings are using the wrong API version for the resource.

One example is the test `Test_Compute_VM_20201201_CRUD` which should exclusively use the `2020-12-01` API version for VirtualMachine resources. In the recording for this test, we find this URL at line #623:

``` 
https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/
resourceGroups/asotest-rg-ibybct/
providers/Microsoft.Compute/
virtualMachines/asotest-vm-guihjc?api-version=2022-03-01
```

The API version `2022-03-01` is unexpected, and reflects the latest version of that resource.

**Expected behavior**

All interaction with ARM should specify the API version the resource was created with, in order to ensure we get the expected behaviour.

theunrepentantgeek:
Closes #[issue number]

The API version associated with a resource is the one that should always be used when making ARM REST API calls. This wasn't the case, as documented in #2416

Fixes #2416

Thanks to @matthchr for helping to track down the issue.

gif

### DIFF
--- a/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
+++ b/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
@@ -116,7 +116,12 @@ func (r *azureDeploymentReconcilerInstance) MakeReadyConditionImpactingErrorFrom
 		return conditions.NewReadyConditionImpactingError(azureErr, conditions.ConditionSeverityWarning, core.UnknownErrorCode)
 	}
 
-	classifier := extensions.CreateErrorClassifier(r.Extension, ClassifyCloudError, r.MustGetAPIVersion(), r.Log)
+	apiVersion, verr := r.GetAPIVersion()
+	if verr != nil {
+		return errors.Wrapf(verr, "error getting api version for resource %s while making Ready condition", r.Obj.GetName())
+	}
+
+	classifier := extensions.CreateErrorClassifier(r.Extension, ClassifyCloudError, apiVersion, r.Log)
 	details, err := classifier(cloudError)
 	if err != nil {
 		return errors.Wrapf(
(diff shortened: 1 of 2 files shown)


v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -542,3 +544,16 @@ func ConvertToARMResourceImpl(
...
 }
+
+// MustGetAPIVersion returns the ARM API version for the resource we're reconciling
+func (r *azureDeploymentReconcilerInstance) MustGetAPIVersion() string {
+	metaObject := r.Obj
+	scheme := r.ResourceResolver.Scheme()
+
+	apiVersion, err := genruntime.GetAPIVersion(metaObject, scheme)
+	if err != nil {
+		panic(err)
```
matthchr, line 555: I think we should avoid `panic` in the reconciler codepaths as much as we can? I know it makes it more awkward to use but can we return `err` here instead?

matthchr (changes requested):

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
super-harsh: Agree with @matthchr here.

theunrepentantgeek: I was on the fence; originally wrote it with errors and then changed to use a panic to reduce boilerplate.

Changed it back.

matthchr (approved):
LGTM! Thanks for making that err change

super-harsh (approved):

PR #2422, opened 2022-07-29
Ensure Condition is correctly updated when secret created after resource

### LINKED ISSUE #2385: Bug: Input Secrets must be created before Azure Resources
**Version of Azure Service Operator**: mcr.microsoft.com/k8s/azureserviceoperator:v2.0.0-beta.1

**Describe the bug**
Azure Operator does not re-reconcile on missing Secret. This means the Secret *Must* exist before the Resource, this makes doing things like gitops hard where ordering is not guaranteed.

**To Reproduce**
1. Create a flexible server
```yaml
cat <<EOF | kubectl apply -f -
apiVersion: dbforpostgresql.azure.com/v1alpha1api20210601
kind: FlexibleServer
metadata:
  name: aso-psql-bug
spec:
  location: centralus
  owner:
    name: aso-psql-rg
  version: "13"
  sku:
    name: Standard_D4s_v3
    tier: GeneralPurpose
  administratorLogin: myAdmin
  administratorLoginPassword:
    name: aso-psql-bug
    key: password
  storage:
    storageSizeGB: 128
EOF
```

2. Observe the correct error in status
```yaml
status:
  conditions:
  - lastTransitionTime: "2022-07-14T21:35:07Z"
    message: 'failed resolving secret references: default/aso-psql-bug does not exist
      (Secret "aso-psql-bug" not found)'
    observedGeneration: 1
    reason: SecretNotFound
    severity: Warning
    status: "False"
    type: Ready
```

3. Now create the secret
```yaml
cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: Secret
metadata:
  annotations:
  name: aso-psql-bug
  namespace: default
data:
  password: Zm9vCg==
type: Opaque
EOF
---
```

4. Watch the FlexibleServer and observe its status never changes from the error.

**Expected behavior**
A missing Secret should be treated as an error and the reconciler should retry with some backoff and it eventual succeed.

matthchr:
Previously the Warning condition would win out over the Info condition of "Reconciling", which would cause the resource to continue to complain about missing secrets even after the problem was solved. This didn't impact actual resource creation but is confusing for users.

Closes #2385

**If applicable**:
- [x] this PR contains tests


v2/internal/controllers/to_azure_secrets_test.go
```diff
@@ -31,20 +32,32 @@ func Test_MissingUserSecret_ReturnsError(t *testing.T) {
...
+		},
+	}
+	tc.CreateResource(secret)
+
+	// Wait for the VM to make it to a steady state
+	tc.Eventually(vm).Should(tc.Match.BeProvisioned(0))
 
 	// Delete VM and resources.
-	tc.DeleteResourcesAndWait(vm, networkInterface, subnet, vnet, rg)
+	tc.DeleteResourcesAndWait(rg)
```
theunrepentantgeek, line 60: Shouldn't this be deleting the virtual machine and leaving deletion of the resource group to our test framework?

v2/pkg/genruntime/conditions/conditions.go
```diff
@@ -207,17 +207,50 @@ func (c Condition) String() string {
...
+// error from Azure can take a long time, and workflows like: submit -> warning -> fix warning -> call Azure -> wait -> success
+// otherwise would continue reporting the Warning Condition until the final success step (possibly many minutes after the
+// warning was resolved).
+func SetConditionReasonAware(o Conditioner, new Condition) {
+	shouldOverwrite := func(new Condition, old Condition) bool {
+		reasonPriority := map[string]int{
+			ReasonReferenceNotFound: -2,
+			ReasonSecretNotFound:    -2,
+			ReasonWaitingForOwner:   -2,
+			ReasonReconciling:       -1,
```
theunrepentantgeek, line 229: Maybe a comment to explain why negative numbers are used. Is it so that zero takes priority over these?

theunrepentantgeek: And maybe stash the map in a global so we don't create it every time you call `shouldOverwrite()`?

theunrepentantgeek (changes requested):

v2/internal/controllers/to_azure_secrets_test.go
matthchr: Fixed

v2/pkg/genruntime/conditions/conditions.go
matthchr: Did both of these

v2/pkg/genruntime/conditions/conditions.go
```diff
@@ -207,17 +207,55 @@ func (c Condition) String() string {
...
+// set if the new condition is different from the existing condition of the same type.
+// See Condition.IsEquivalent and Condition.ShouldOverwrite for more details.
 func SetCondition(o Conditioner, new Condition) {
+	setCondition(o, new, func(new Condition, old Condition) bool { return new.ShouldOverwrite(old) })
+}
+
+// Reasons other than those explicitly called out here have the default priority of 0.
+// These are given a negative priority so that they "lose" to the default of 0 and are overwritten.
+// Take care to not modify this structure (Golang doesn't support a readonly map). We could use
+// /home/matthchr/work/github/azure-service-operator/v2/tools/generator/internal/readonly/readonly_map.go but given
```
theunrepentantgeek, line 219:
```suggestion
// /v2/tools/generator/internal/readonly/readonly_map.go but given
```

theunrepentantgeek (approved):

v2/pkg/genruntime/conditions/conditions.go
matthchr: Doh, good catch

PR #2429, opened 2022-08-03
Improve EnvTest performance by not launching so many

matthchr:
* Stop envtest (kube-apiserver, etcd) when they are no longer needed.
* Prevent more than `numCores` extra envtests from being run at once. Total running should be 1 + `numCores`.

### DIFF
--- a/v2/internal/testcommon/kube_per_test_context.go
+++ b/v2/internal/testcommon/kube_per_test_context.go
@@ -43,7 +43,6 @@ type KubePerTestContext struct {
 	*KubeGlobalContext
 	KubeBaseTestContext
 
-	Ctx        context.Context
 	kubeClient client.Client
 	G          gomega.Gomega
 	Verify     *Verify
@@ -118,10 +117,21 @@ func (ctx KubeGlobalContext) ForTest(t *testing.T) *KubePerTestContext {
 	// Test configs never want SyncPeriod set as it introduces jitter
 	cfg.SyncPeriod = nil
 
-	return ctx.ForTestWithConfig(t, cfg)
+	return ctx.forTestWithConfig(t, cfg, bypassesParallelLimits)
 }
 
+type testConfigParallelismLimit string
+
+const (
+	countsAgainstParallelLimits = testConfigParallelismLimit("countsAgainstParallelLimits")
+	bypassesParallelLimits      = testConfigParallelismLimit("bypassesParallelLimits")
+)
+
 func (ctx KubeGlobalContext) ForTestWithConfig(t *testing.T, cfg config.Values) *KubePerTestContext {
+	return ctx.forTestWithConfig(t, cfg, countsAgainstParallelLimits)
+}
+
+func (ctx KubeGlobalContext) forTestWithConfig(t *testing.T, cfg config.Values, parallelismRestriction testConfigParallelismLimit) *KubePerTestContext {
 	/*
 		Note: if you update this method you might also need to update TestContext.Subtest.
 	*/
@@ -130,6 +140,7 @@ func (ctx KubeGlobalContext) ForTestWithConfig(t *testing.T, cfg config.Values) 
 	if err != nil {
 		t.Fatal(err)
 	}
+	perTestContext.CountsTowardsParallelLimits = parallelismRestriction == countsAgainstParallelLimits
 
 	baseCtx, err := ctx.createBaseTestContext(perTestContext, cfg)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


v2/internal/testcommon/kube_per_test_context.go
```diff
@@ -118,10 +117,14 @@ func (ctx KubeGlobalContext) ForTest(t *testing.T) *KubePerTestContext {
 	// Test configs never want SyncPeriod set as it introduces jitter
 	cfg.SyncPeriod = nil
 
-	return ctx.ForTestWithConfig(t, cfg)
+	return ctx.forTestWithConfig(t, cfg, false)
```
theunrepentantgeek, line 120: Bool parameters are difficult to read - suggest changing this so `ForTestWithConfig()` defaults to _included in limit_ but with a method you can call to exclude it.

So this code would become

``` go
return ctx.ForTestWithConfig(t, config).exludeFromConcurrentLimit()
```

v2/internal/testcommon/kube_per_test_context.go
```diff
@@ -118,10 +117,14 @@ func (ctx KubeGlobalContext) ForTest(t *testing.T) *KubePerTestContext {
...
 }
 
 func (ctx KubeGlobalContext) ForTestWithConfig(t *testing.T, cfg config.Values) *KubePerTestContext {
+	return ctx.forTestWithConfig(t, cfg, true)
+}
+
+func (ctx KubeGlobalContext) forTestWithConfig(t *testing.T, cfg config.Values, countsTowardsLimit bool) *KubePerTestContext {
```
theunrepentantgeek, line 127: See above comment. The value `countsTowardsLimit` is intelligible here in the implementation, but at callsites it's mysterious and cryptic - which is cool in Film Noir, but less so in code.

v2/internal/testcommon/kube_per_test_context.go
```diff
@@ -157,7 +159,6 @@ func (ctx KubeGlobalContext) ForTestWithConfig(t *testing.T, cfg config.Values)
 		Verify:              verify,
 		Match:               match,
 		scheme:              scheme,
-		Ctx:                 context,
```
theunrepentantgeek, line 160: Why remove context here?

theunrepentantgeek (changes requested):
Mostly looks good. One API tweak, please.

v2/internal/testcommon/kube_per_test_context.go
matthchr: I pulled it up into the `PerTestContext` rather than the `KubePerTestContext` so that I could use it in the semaphore acquire call.

v2/internal/testcommon/kube_per_test_context.go
matthchr: The issue with doing it this way is that the call to `ctx.createBaseTestContext` in `forTestWithConfig` is the thing that actually creates and launches the envtest. That call is made in `forTestWithConfig` quite early on in the process. Once envtest is launched, none of the tracking stuff can be done. Doing it the way you suggested is difficult because `.excludeFromConcurrentLimit()` gets called after the whole EnvTest launch sequence has already happened.

That's part of the reason why I moved `forTestWithConfig` to be unexported. Callers either use `ForTest`, which implies they're using the default config which is always running, _or_ they use `ForTestWithConfig` which implies they are not using the default config and they get the concurrency limiting and garbage collection I've added. 

So from the callers perspective there is no confusing `bool` parameter, it's hidden. 
I could:
1. Change this to an enum so it reads clearer.
2. Improve the name of the boolean (doesn't really help at the call-site but reads clearer in the internal `forTestWithConfig`).
3. Do a larger refactor to come up with a configuration struct and pass that?

I'm definitely happy to do either 1 or 2... I could probably be convinced to do 3, although I'm not sure it really makes sense as this function only has 3 arguments. 

Thoughts?

theunrepentantgeek: Ergh. I didn't realise it launched quite so quickly.

Change it to clarify the parameter name with a comment and I'll be happy:

```suggestion
	return ctx.forTestWithConfig(t, cfg, /* countsTowardsLimit: */ false)
```

theunrepentantgeek (approved):

super-harsh (approved):

v2/internal/testcommon/kube_per_test_context.go
matthchr: Moved to a simple enum instead for a smidge more clarity.

PR #2086, opened 2022-02-11
Add PrivateDNSZone resource

vaspahomov:
Signed-off-by: vaspahomov <vas2142553@gmail.com>

**What this PR does / why we need it**:

I want to add PrivateDNSZone resource and their PrivateDNSZone/VirtualNetworkLink dependency. 
This can help us to create [PostgreSQL connected with private network](https://github.com/Azure/azure-service-operator/blob/main/v2/api/dbforpostgresql/v1alpha1api20210601/flexible_server_types_gen.go#L2299) without az cli or others.

**Special notes for your reviewer**:

I've found that `'*api201*'` versions of network are excluded. I think we need explicitly add PrivateDnsZones and VirtualNetworkLinks (but I can not find right resource name) resources.

After this manipulations I'm receiving: 
```
E0211 14:19:54.561995   27841 gen_kustomize.go:111] Error during code generation:
failed during pipeline stage 23/57: Verify there are no ErroredType's containing errors: "github.com/Azure/azure-service-operator/v2/api/network/v1alpha1api20180901/PrivateDnsZone" has property "Status" with errors: "missing status information for github.com/Azure/azure-service-operator/v2/api/network/v1alpha1api20180901/PrivateDnsZones"
```

I've tried to add this in status section, but it did not help:

```
    {
      # give the privations types their own namespace
      basePath: 'privatedns/resource-management',
      namespace: 'Microsoft.Network'
    },
```

Can you tell me what I'm doing wrong?

**How does this PR make you feel**:
gif

**If applicable**:
- [ ] this PR contains documentation
- [ ] this PR contains tests

PR #2430, opened 2022-08-03
Add support for PrivateDnsZone

theunrepentantgeek:

Adds support for the resource `PrivateDnsZone`.

This resource is a bit odd, as it has no properties of it's own, but exists more as something for other resources to reference. It seems to only exist in the `2018-09-01` version, so the imports needed to be carefully scoped to avoid importing other resources that we can't handle.

Will allow us to close PR #2086 by providing the requested resource, ahead of Swagger-only support.

gif

**If applicable**:
- [x] this PR contains tests


v2/internal/testcommon/kube_per_test_context.go
```diff
@@ -602,21 +602,46 @@ func (tc *KubePerTestContext) AsExtensionOwner(obj client.Object) *genruntime.Ar
...
 
 	tc.T.Logf("Exported resource to %s", filepath)
 }
 
+func (tc *KubePerTestContext) cleanSample(resource any) {
+
+	if kr, ok := resource.(genruntime.KubernetesResource); ok {
+		// Remove Status
+		emptyStatus := kr.NewEmptyStatus()
+		_ = kr.SetStatus(emptyStatus) // Ignore errors
```
super-harsh, line 627: do we need empty status here or nil status here? As seen in the sample `status: {}` is present.

v2/config/samples/network/v1beta/v1beta20180901_privatednszone.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: network.azure.com/v1beta20180601
+kind: PrivateDnsZone
+metadata:
+  creationTimestamp: null
```
super-harsh, line 4: Do we require this and empty status below?

super-harsh (changes requested):
Resource addition LGTM, left a few comments for sample. Also, you might need to run `TEST_FILTER=Test_Samples_CreationAndDeletion task controller:test-integration-envtest` to make sure sample is tested and recording is added for the new sample.

On the above, its my bad I did not add step for recording samples test in documentation. Have created an issue #2433 to track it.

v2/internal/testcommon/kube_per_test_context.go
theunrepentantgeek: We actually want no status at all; will modify.

v2/config/samples/network/v1beta/v1beta20180901_privatednszone.yaml
theunrepentantgeek: Not required; will remove.

v2/config/samples/network/v1beta/v1beta20180901_privatednszone.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: network.azure.com/v1beta20180601
```
super-harsh, line 1:
```suggestion
apiVersion: network.azure.com/v1beta20180901
```

I just noticed failed on your PR.. ​runtime.notRegisteredErr​error indicates about either either of gvk in the sample is not correct. This should fix the error.

theunrepentantgeek: Thanks. I've just merged the same fix into my local branch.

v2/internal/controllers/crd_networking_privatednszone_test.go
```diff
@@ -0,0 +1,39 @@
...
+			Owner:    testcommon.AsOwner(rg),
+		},
+	}
+
+	tc.CreateResourceAndWait(zone)
+
+	tc.Expect(zone.Status.MaxNumberOfRecordSets).ToNot(BeNil())
+	tc.Expect(zone.Status.MaxNumberOfVirtualNetworkLinks).ToNot(BeNil())
+
+	tc.DeleteResourceAndWait(zone)
```
matthchr, line 38: Minor: Could add assertion after this that delete really happened in Azure. Look for example tests that use the `HEAD` API to do that. For example the `crd_cache_redis_test.go` does that

v2/internal/testcommon/kube_per_test_context.go
```diff
@@ -602,21 +602,46 @@ func (tc *KubePerTestContext) AsExtensionOwner(obj client.Object) *genruntime.Ar
...
+	}
+
+	if oa, ok := resource.(metav1.ObjectMetaAccessor); ok {
+		// Remove runtime objectmeta information
+		om := oa.GetObjectMeta()
+		om.SetAnnotations(nil)
+		om.SetFinalizers(nil)
+		om.SetManagedFields(nil)
+		om.SetLabels(nil)
+		om.SetOwnerReferences(nil)
```
matthchr, line 637: Consider setting namespace to `default` as well.

v2/internal/controllers/crd_networking_privatednszone_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+	tc := globalTestContext.ForTest(t)
+
+	rg := tc.CreateTestResourceGroupAndWait()
+
+	name := tc.Namer.GenerateName("pdz") + ".com"
+	zone := &network.PrivateDnsZone{
+		ObjectMeta: tc.MakeObjectMetaWithName(name),
+		Spec: network.PrivateDnsZones_Spec{
+			Location: to.StringPtr("global"),
```
matthchr, line 28: This is an interesting "location" that I haven't seen before. Do you know what it means?

matthchr (approved):

v2/internal/controllers/crd_networking_privatednszone_test.go
theunrepentantgeek: Checking against Azure requires an ID or something to query for, and there's nothing on the object. Other tests use an Id property that doesn't exist here.

v2/internal/controllers/crd_networking_privatednszone_test.go
theunrepentantgeek: DNS settings aren't (cannot be!) scoped to a region, so while all ARM resources need a location, `PrivateDNSZone` doesn't use it - and seems to require `global` as the value.

v2/internal/controllers/crd_networking_privatednszone_test.go
matthchr: We also compute the Azure ID and store it in an annotation: `serviceoperator.azure.com/resource-id`, which is accessible as `genruntime.ResourceIDAnnotation` or readable via `genruntime.GetResourceID(obj)`

theunrepentantgeek: Added.

v2/config/samples/network/v1beta/v1beta20180901_privatednszone.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: network.azure.com/v1beta20180901
+kind: PrivateDnsZone
+metadata:
+  name: aso-sample-pdz.com
+  namespace: aso-test-networking-privatednszone-crud
```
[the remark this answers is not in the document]
theunrepentantgeek: Done.

v2/internal/testcommon/kube_per_test_context.go
theunrepentantgeek: Done.

super-harsh (approved):

PR #2437, opened 2022-08-09
Fix lint issues

theunrepentantgeek:

The linter in our base container image has been updated, and it detected a couple of things we need to address.

gif

### DIFF
--- a/v2/internal/controllers/crd_cache_redis_test.go
+++ b/v2/internal/controllers/crd_cache_redis_test.go
@@ -47,10 +47,13 @@ func Test_Cache_Redis_CRUD(t *testing.T) {
 	// same time seems to wreak havoc (or at least make things take a
 	// lot longer), so test firewall rules synchronously first -
 	// they're quick when not clobbering links.
-	tc.T.Run("Redis firewall rule CRUD", func(t *testing.T) {
-		t.Parallel()
-		Redis_FirewallRule_CRUD(tc.Subtest(t), redis1)
-	})
+	tc.RunSubtests(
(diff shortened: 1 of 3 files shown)


super-harsh (approved):

v2/internal/controllers/crd_cache_redis_test.go
```diff
@@ -47,6 +48,7 @@ func Test_Cache_Redis_CRUD(t *testing.T) {
 	// lot longer), so test firewall rules synchronously first -
 	// they're quick when not clobbering links.
 	tc.T.Run("Redis firewall rule CRUD", func(t *testing.T) {
```
matthchr, line 50: You could instead switch this to use `tc.RunSubtests` which I think the linter doesn't notice. We really don't want to run this subtest in parallel and I am not sure that doing it this way will actually do what we think it's supposed to. If you look at the comment in `RunParallelSubtests` it says:
```
	// this looks super weird but is correct.
	// parallel subtests do not run until their parent test completes,
	// and then the parent test does not finish until all its subtests finish.
	// so "subtests" will run and complete, then all the subtests will run
	// in parallel, and then "subtests" will finish. ¯\_(ツ)_/¯
	// See: https://blog.golang.org/subtests#TOC_7.2.
```

Or alternatively we can just call `RunParallelSubtests` here and it'll do a parallel of 1 (which is sequential).

theunrepentantgeek: Done, using `tc.RunSubtests()`

PR #2442, opened 2022-08-10
Support new extension points for ARM resources

matthchr:
- Hook for deletion
- Hook for after-successful-creation
- Hook for claim

### DIFF
--- a/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
+++ b/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
@@ -329,6 +329,12 @@ func (r *azureDeploymentReconcilerInstance) handleCreatePollerSuccess(ctx contex
 		return ctrl.Result{}, err
 	}
 
+	onSuccess := extensions.CreateSuccessfulCreationHandler(r.Extension, r.Log)
+	err = onSuccess(r.Obj)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
 	ClearPollerResumeToken(r.Obj)
 	return ctrl.Result{}, nil
 }
--- a/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
+++ b/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
@@ -199,36 +199,9 @@ func (r *azureDeploymentReconcilerInstance) StartDeleteOfResource(ctx context.Co
 	r.Log.V(Status).Info(msg)
 	r.Recorder.Event(r.Obj, v1.EventTypeNormal, string(DeleteActionBeginDelete), msg)
 
-	// If we have no resourceID to begin with, the Azure resource was never created
-	resourceID := genruntime.GetResourceIDOrDefault(r.Obj)
-	if resourceID == "" {
-		return ctrl.Result{}, nil
-	}
-
-	// Check that this objects owner still exists
-	// This is an optimization to avoid excess requests to Azure.
-	_, err := r.ResourceResolver.ResolveResourceHierarchy(ctx, r.Obj)
-	if err != nil {
-		var typedErr *resolver.ReferenceNotFound
-		if errors.As(err, &typedErr) {
-			return ctrl.Result{}, nil
-		}
-		return ctrl.Result{}, err
-	}
-
-	apiVersion, verr := r.GetAPIVersion()
-	if verr != nil {
-		return ctrl.Result{}, errors.Wrapf(verr, "error getting api version for resource %s while starting deletion of resource", r.Obj.GetName())
-	}
-
-	// retryAfter = ARM can tell us how long to wait for a DELETE
-	retryAfter, err := r.ARMClient.DeleteByID(ctx, resourceID, apiVersion)
-	if err != nil {
-		return ctrl.Result{}, errors.Wrapf(err, "deleting resource %q", resourceID)
-	}
-
-	// Normally don't need to set both of these fields but because retryAfter can be 0 we do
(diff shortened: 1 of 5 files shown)


v2/pkg/genruntime/extensions/claimer.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+
+	. "github.com/Azure/azure-service-operator/v2/internal/logging"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+)
+
+// Claimer can be implemented to customize how the reconciler claims a resource
```
theunrepentantgeek, line 17: Do you need to doc what Claiming is about?

v2/pkg/genruntime/extensions/successful_resource_modifier.go
```diff
@@ -0,0 +1,46 @@
...
+
+	. "github.com/Azure/azure-service-operator/v2/internal/logging"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+)
+
+// SuccessfulResourceModifier can be implemented to customize the resource upon successful creation
+type SuccessfulResourceModifier interface {
+	// Success modifies the resource based on a successful creation
+	Success(obj genruntime.ARMMetaObject) error
+}
```
theunrepentantgeek, line 20: What's the scenario where this extension point is needed?

theunrepentantgeek (approved):

v2/pkg/genruntime/extensions/claimer.go
matthchr: Its part of the `Reconciler` interface we define and is described in that interface already:
```
	// Claim performs resource specific claim actions. This must be idempotent.
	// A standard finalizer is added to all resources, Claim
	// should deal with any resource specific claiming actions (such as setting a resource ID annotation, etc).
	// If Claim returns an error then reconciliation will be retried according to the returned Ready condition until
	// no error is returned. Once Claim succeeds CreateOrUpdate is called.
	Claim(
		ctx context.Context,
		log logr.Logger,
		eventRecorder record.EventRecorder,
		obj MetaObject) error
```

Do you think we need more than that?

v2/internal/reconcilers/arm/deleter.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"github.com/pkg/errors"
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	"github.com/Azure/azure-service-operator/v2/internal/genericarmclient"
+	"github.com/Azure/azure-service-operator/v2/internal/resolver"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+)
+
+func deleteResource(
```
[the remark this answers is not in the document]
matthchr: Ive moved this back into `azure_generic_arm_reconciler` (which is where it is used, so that should be more clear), and added the following comment:
```
// deleteResource deletes a resource in ARM. This function is used as the default deletion handler and can
// have its behavior modified by resources implementing the genruntime.Deleter extension
```

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -356,6 +329,12 @@ func (r *azureDeploymentReconcilerInstance) handleCreatePollerSuccess(ctx contex
 		return ctrl.Result{}, err
 	}
 
+	onSuccess := extensions.CreateSuccessfulResourceModifier(r.Extension, r.Log)
```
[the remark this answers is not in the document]
matthchr: Yeah I don't like the name either. The hook is: "When a resource has been created successfully". We could call it something more action-oriented like `OnSuccessfulResourceCreation`. The issue is that's not really `CreateX` which is the pattern we have for other extension handlers, where X is the name of the extension.

`CreateResourceCreationHandler`?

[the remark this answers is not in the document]
matthchr: I've moved this to `SuccessfulCreationHandler` + `CreateSuccessfulCreationHandler`

v2/pkg/genruntime/extensions/successful_resource_modifier.go
matthchr: Because `subscriptionAliases` don't have the right url to assign child resources (like RoleAssignment) to, I've got in another branch the addition of a concept called `ChildResourceIDOverride`, which is an annotation that can be added to an object. If that annotation is added to the object, when child resources are attached to that object, instead of using the normal ID/Type of the object they just use the URL in the annotation. This hook is so that I can manually add that annotation to the `subscriptionAlias` type, which then causes it to be able to correctly have child resources created off of the subscription (rather than off of the alias).

Basically it's a way to combine two separate resources into 1. It's somewhat hacky so we shouldn't really use it regularly, but it saves us from having to implement our own reconciler for subscription, it can just use the normal reconciler.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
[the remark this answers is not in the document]
theunrepentantgeek: That works. :+1:

v2/pkg/genruntime/extensions/claimer.go
theunrepentantgeek: No, that'll do. I missed that.

v2/pkg/genruntime/extensions/successful_resource_modifier.go
theunrepentantgeek: Good use of an extension point.

PR #2443, opened 2022-08-11
Add flag to skip tests from running in live mode

super-harsh:


matthchr (approved):

v2/internal/controllers/samples_test.go
```diff
@@ -40,6 +40,11 @@ var randomNameExclusions = []string{
 
 func Test_Samples_CreationAndDeletion(t *testing.T) {
 	t.Parallel()
+
+	if *isLive {
```
theunrepentantgeek, line 44: Don't you need a **nil** check before dereferencing `isLive` for safety?

theunrepentantgeek (approved):

v2/internal/controllers/samples_test.go
super-harsh: default for `isLive` would always be `false`. So it never ends up being nil

PR #2445, opened 2022-08-11
Record tests for national cloud better

matthchr:
Also add support for billing invoice ID test variable

### DIFF
--- a/v2/internal/config/vars_test.go
+++ b/v2/internal/config/vars_test.go
@@ -7,6 +7,8 @@ import (
 	"reflect"
 	"testing"
 
+	// Importing this for side effects is required as it initializes cloud.AzurePublic
+	_ "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
 	. "github.com/onsi/gomega"
 
--- a/v2/internal/testcommon/test_context.go
+++ b/v2/internal/testcommon/test_context.go
@@ -45,18 +45,19 @@ type TestContext struct {
 
 type PerTestContext struct {
 	TestContext
-	T                   *testing.T
-	logger              logr.Logger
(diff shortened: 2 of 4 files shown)


v2/internal/config/vars_test.go
```diff
@@ -7,6 +7,8 @@ import (
 	"reflect"
 	"testing"
 
+	// TODO: Importing this for side effects is required as it initializes cloud.AzurePublic
```
theunrepentantgeek, line 10: Don't think this needs `TODO`.
```suggestion
	// Importing this for side effects is required as it initializes cloud.AzurePublic
```

v2/internal/testcommon/test_context.go
```diff
@@ -93,28 +96,31 @@ func (tc TestContext) ForTest(t *testing.T, cfg config.Values) (PerTestContext,
 	logger := NewTestLogger(t)
 
 	cassetteName := "recordings/" + t.Name()
-	creds, azureIDs, recorder, err := createRecorder(cassetteName, tc.RecordReplay)
+	details, err := createRecorder(cassetteName, cfg, tc.RecordReplay)
 	if err != nil {
 		return PerTestContext{}, errors.Wrapf(err, "creating recorder")
 	}
+	// Use the recorder-specific CFG, which will URLs and AADAuthorityHost (among other things) to default
```
theunrepentantgeek, line 103: Missing word?

```suggestion
	// Use the recorder-specific CFG, which will force URLs and AADAuthorityHost (among other things) to default
```

theunrepentantgeek (approved):

v2/internal/testcommon/test_context.go
matthchr: Thanks, fixed

v2/internal/config/vars_test.go
matthchr: Good point

super-harsh (approved):

PR #2446, opened 2022-08-11
Add support for Subscription creation in ASO

matthchr:
This resource is a bit wonky/tricky and as such required a few hooks to customize its reconciliation behavior from what the default is.

The underlying API is that creating a subscription requires creating an Alias. That Alias is not the subscription itself, just a mechanism to produce a new one. Unfortunately the Alias cannot have RoleAssignments or other things put on it once it's created, that must be done on the subscription itself.

In order to manage this resource in ASO we combine the two concepts into a single one: `subscription.Alias`. Deleting the Alias causes the subscription to be cancelled. A subscription cannot be cancelled if it has resources (delete will fail with an error). A subscription that is cancelled will be deleted 90 days later (or can be manually deleted after 3 days).

**Blocked until**
 - [x] #2445 merges

gif

**If applicable**:
- [x] this PR contains tests


v2/api/subscription/customizations/alias_extensions.go
```diff
@@ -0,0 +1,88 @@
...
+
+	// Using armClient.ClientOptions() here ensures we share the same HTTP connection, so this is not opening a new
+	// connection each time through
+	subscriptionClient, err := armsubscription.NewClient(armClient.Creds(), armClient.ClientOptions())
+	if err != nil {
+		return ctrl.Result{}, errors.Wrapf(err, "failed to create new workspaceClient")
+	}
+
+	// Don't need to do anything with the response here so just ignore it
+	_, err = subscriptionClient.Cancel(ctx, subscriptionID, nil)
```
theunrepentantgeek, line 49: Might the cancellation be a long-running-operation that needs monitoring?

v2/internal/genericarmclient/arm_ids.go
```diff
@@ -66,5 +66,14 @@ func MakeResourceGroupScopeARMID(subscription string, resourceGroup string, prov
...
-	return fmt.Sprintf("/subscriptions/%s/resourceGroups/%s", subscription, resourceGroup)
+	return fmt.Sprintf("%s/resourceGroups/%s", MakeSubscriptionID(subscription), resourceGroup)
+}
+
+// MakeSubscriptionID makes an ARM ID representing a subscription. This has the format:
+// /subscriptions/00000000-0000-0000-0000-000000000000
+// This is "special" because there is no provider at all
+func MakeSubscriptionID(subscription string) string {
+	// This is a "special" format as it doesn't actually list a provider in it anywhere. I was expecting that it would
+	// have a Microsoft.Resources provider but it doesn't
```
theunrepentantgeek, line 77: Putting the second sentence on the next line makes it more visible.
```suggestion
	// This is a "special" format as it doesn't actually list a provider in it anywhere. 
	// I was expecting that it would have a Microsoft.Resources provider but it doesn't
```

v2/pkg/genruntime/base_types.go
```diff
@@ -30,6 +30,13 @@ const (
 // TODO: to serviceoperator-internal.azure.com to signify they are internal?
 const (
 	ResourceIDAnnotation = "serviceoperator.azure.com/resource-id"
+
+	// ChildResourceIDOverrideAnnotation is an annotation that can be used to force child resources
+	// to be owned by a different resource ID than it would normally. This is primarily used for
+	// resources like SubscriptionAlias + Subscription, where the create API doesn't use the same
+	// ResourceID as needed by child resources of the subscription.
+	// When present, this takes precedent over
```
theunrepentantgeek, line 38: Incomplete sentence?

"This takes precedence over ... _what?_"

theunrepentantgeek (commented):
I think I need to read through this again to see how it works, but my first pass looks good, just a trio of comments.

v2/internal/config/vars.go
```diff
@@ -96,9 +100,9 @@ func (v Values) Cloud() cloud.Configuration {
 
 	// Special handling if we've got all the defaults just return the official public cloud
 	// configuration
-	hasDefaultAzureAuthorityHost := v.AzureAuthorityHost == "" || v.AzureAuthorityHost == cloud.AzurePublic.ActiveDirectoryAuthorityHost
-	hasDefaultResourceManagerEndpoint := v.ResourceManagerEndpoint == "" || v.ResourceManagerEndpoint == cloud.AzurePublic.Services[cloud.ResourceManager].Endpoint
-	hasDefaultResourceManagerAudience := v.ResourceManagerAudience == "" || v.ResourceManagerAudience == cloud.AzurePublic.Services[cloud.ResourceManager].Audience
+	hasDefaultAzureAuthorityHost := v.AzureAuthorityHost == "" || v.AzureAuthorityHost == DefaultAADAuthorityHost
```
super-harsh, line 103: You have similar changes in your other [PR](https://github.com/Azure/azure-service-operator/pull/2445) which might be conflicting

v2/internal/controllers/samples_test.go
```diff
@@ -27,6 +27,7 @@ const samplesPath = "../../config/samples"
 var skipTests = []string{
 	// TODO: Will re-record test .. as resource was having some issues in creation.
 	"cache",
+	"subscription", // Can't be run in our live env currently
```
super-harsh, line 30: Maybe you want to delete this? We already have a skip flag for live

v2/internal/resolver/resource_hierarchy.go
```diff
@@ -122,23 +130,76 @@ func (h ResourceHierarchy) FullyQualifiedARMID(subscriptionID string) (string, e
 		// Join them together
 		interleaved := genruntime.InterleaveStrSlice(resourceTypes, remainingNames)
 		return genericarmclient.MakeResourceGroupScopeARMID(subscriptionID, rgName, provider, interleaved...)
+	case ResourceHierarchyRootTenant:
+		// The only resource we actually care about for figuring out resource types is the
+		// most derived resource
+		res := h[len(h)-1]
```
super-harsh, line 136: could this run into indexOutOfBounds if `len==0`?

super-harsh: ok, makes sense .. it panics in `rootKind` method if `len==0`

v2/internal/testcommon/creds.go
```diff
@@ -18,17 +18,20 @@ import (
...
+var cachedIds AzureIDs
+
+const TestBillingIDVar = "TEST_BILLING_ID"
 
 type AzureIDs struct {
-	subscriptionID string
-	tenantID       string
+	subscriptionID   string
+	tenantID         string
+	billingInvoiceID string
```
super-harsh, line 28: Same, These changes and some below also exist in your other PR

v2/api/subscription/customizations/alias_extensions.go
matthchr: It's not, this finishes synchronously. Which is good too because otherwise we'd have to hack this up further.

v2/internal/controllers/samples_test.go
matthchr: We can't easily record it, is the issue actually. I need to fix the comment to explain better. Since we can't record it there is no recording so it wont' work.

v2/internal/testcommon/creds.go
matthchr: Yeah I need to merge the other one and rebase this, will do.

matthchr: Fixed this and pushed

v2/internal/resolver/resource_hierarchy.go
matthchr: It could, yes. There are other places where we explicitly panic on `len==0`, but basically a `ResourceHierarchy` is expected to be at least length = 1. Moreover, these are created by functions we control and they always have at least the resource being deployed in them, so there's no way in practice to end up with 1 that's length = 0.

Good question though.

v2/internal/controllers/samples_test.go
matthchr: Ive updated the comment here

v2/internal/config/vars.go
matthchr: Fixed

v2/internal/genericarmclient/arm_ids.go
matthchr: Fixed

v2/api/subscription/customizations/alias_extensions.go
matthchr: I've added a note about this in the code. Also worth clarifying that when I said it finishes synchronously I mean that the SDK also doesn't do any polling here. There is no azure async operation to poll and this operation completes quickly as the result of a single POST response.

v2/pkg/genruntime/base_types.go
matthchr: Fixed

theunrepentantgeek (approved):

PR #2451, opened 2022-08-16
[Automated] Add API Docs

Porges:
Add new API Docs for latest release


theunrepentantgeek (changes requested):
This looks wrong - I don't think we've actually removed `UserAssignedIdentitiesfrom` the `compute` package.

matthchr:
@theunrepentantgeek - I think we did. It was removed in this PR: #2409. Specifically [here](https://github.com/Azure/azure-service-operator/pull/2409/files#diff-5794927accddb9621b12855f68facb91046f58a24fbb12fbcd0961585748eb8dR245)

Although in retrospect we probably shouldn't have removed it from the `_Status` types.

theunrepentantgeek (approved):

PR #2461, opened 2022-08-22
Add workflow to scan controller image (trivy)

### LINKED ISSUE #2426: Run Trivy on our container images
See https://github.com/Azure/azure-workload-identity/blob/main/.pipelines/templates/scan-images.yaml#L16 for an example

super-harsh:
Closes #2426 

Added workflow for scanning controller image using trivy

Decided to not include the check in other workflows, as we don't want the CI to be blocked due to these. We can track the vulnerabilities through the separate workflow and fix as we go.


matthchr (approved):

.github/workflows/scan-controller-image.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+  push:
+    # run when changes pushed to go.mod on main branch
+    branches:
+      - main
+    paths:
+      - ./v2/go.mod
+
+jobs:
+  deploy-site:
```
theunrepentantgeek, line 19: Is this the right name for this job?

theunrepentantgeek (approved):

.github/workflows/scan-controller-image.yaml
super-harsh: Good catch! Thanks. I copied some of this from other workflow and might've hit undo somewhere.

super-harsh: Making it to be called `scan-image`

PR #2462, opened 2022-08-23
Refactor delete to use async operation tracking

matthchr:
Refactors deletion code to use the async operation ID to track the delete operation. This results in better error messages if something goes wrong.

**If applicable**:
- [x] this PR contains tests


v2/internal/controllers/crd_servicebus_namespace_basic_test.go
```diff
@@ -53,10 +53,7 @@ func Test_ServiceBus_Namespace_Basic_CRUD(t *testing.T) {
 	tc.DeleteResourceAndWait(namespace)
 
 	// Ensure that the resource was really deleted in Azure
-	exists, retryAfter, err := tc.AzureClient.HeadByID(tc.Ctx, armId, string(servicebus.APIVersion_Value))
-	tc.Expect(err).ToNot(HaveOccurred())
-	tc.Expect(retryAfter).To(BeZero())
-	tc.Expect(exists).To(BeFalse())
+	tc.ExpectResourceIsDeletedInAzure(armId, string(servicebus.APIVersion_Value))
```
matthchr, line 56: It used to be that we could reliably say that when the ASO resource deletion completed, the resource in Azure was 100% gone. It seems like for most services that is still the case, but some services you can still GET the resource for a very short time even after the async operation ID has reported that the delete succeeded. Possibly this is due to some caching layer for those services and possibly it's just their API behaving a bit weird.

While we _could_ continue to maintain that invariant by explicitly doing a GET and waiting for the resource to have disappeared in the operator before removing it, I didn't think it made sense to do so since it costs it's an extra request that we otherwise wouldn't need.

AFAIK, all of the SDKs do what we're doing here (poll the operation until its done). They don't check that GET 404's after that, so we don't either.

That means that in the tests, we needed to change some "assert this is true _now_" to "assert this is true eventually". This is one such example.

v2/internal/genericarmclient/poller.go
```diff
@@ -62,3 +68,44 @@ func (l *PollerResponse) Resume(ctx context.Context, client *GenericClient, toke
 	l.RawResponse = resp
 	return nil
 }
+
+//
+//// CreateOrUpdatePollerResponse is the response from issuing a PUT to Azure. It contains a poller (for polling the long-running
```
matthchr, line 73: Delete this

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -211,33 +211,35 @@ func (r *azureDeploymentReconcilerInstance) MonitorDelete(ctx context.Context) (
 	r.Log.V(Verbose).Info(msg)
 	r.Recorder.Event(r.Obj, v1.EventTypeNormal, string(DeleteActionMonitorDelete), msg)
 
-	resourceID, hasResourceID := genruntime.GetResourceID(r.Obj)
+	// TODO: Could tear this part out
```
matthchr, line 214: Deal with this TODO

theunrepentantgeek (approved):

v2/internal/testcommon/test_context.go
```diff
@@ -306,14 +306,20 @@ func createRecorder(cassetteName string, cfg config.Values, recordReplay bool) (
...
+		for key, values := range i.Response.Headers {
 			for i := range values {
 				values[i] = hide(values[i], azureIDs.subscriptionID, uuid.Nil.String())
 				values[i] = hide(values[i], azureIDs.tenantID, uuid.Nil.String())
 				if azureIDs.billingInvoiceID != "" {
 					values[i] = hide(values[i], azureIDs.billingInvoiceID, DummyBillingId)
 				}
 			}
+			// Hide the base request URL in the AzureOperation and Location headers
+			if key == "Azure-Asyncoperation" || key == "Location" {
```
super-harsh, line 318: Do we have any constants for these response headers somewhere? Can these response header values change in future?

super-harsh (approved):

v2/internal/testcommon/test_context.go
super-harsh: Tryna use them below? 

https://github.com/Azure/azure-sdk-for-go/blob/sdk/armcore/v0.8.1/sdk/armcore/internal/pollers/pollers.go#L22

matthchr: Those ones are internal so we cant use them, but Ill define our own and use them

PR #2465, opened 2022-08-24
Add web/Sites and web/Serverfarm

### LINKED ISSUE #2039: Feature: Support Azure Functions
Support creation of Azure Functions in ASOv2.

Functions is actually a `Microsoft.Web/sites` of type `functionapp` (see [ARM template](https://docs.microsoft.com/en-us/azure/azure-functions/functions-create-first-function-resource-manager?tabs=visual-studio-code%2Cazure-cli)), which requires the following additional resources as well:

- Microsoft.Storage/storageAccounts
- Microsoft.Web/serverfarms
- Microsoft.Web/sites
- Microsoft.Insights/components

ASO v2 already supports **Microsoft.Insights/components** and **Microsoft.Storage/storageAccounts**, but we need the support for **Microsoft.Web/serverfarms** and **Microsoft.Web/sites**.

super-harsh:
Closes #2039 

This PR adds Azure Functions(web/Sites) and App Service(web/Serverfarm)

**If applicable**:
- [x] this PR contains tests


v2/internal/controllers/crd_web_test.go
```diff
@@ -0,0 +1,91 @@
...
+func Test_Web_ServerFarm_CRUD(t *testing.T) {
+	t.Parallel()
+
+	tc := globalTestContext.ForTest(t)
+
+	rg := tc.CreateTestResourceGroupAndWait()
+
+	serverfarm := newServerFarm(tc, rg)
+
+	tc.CreateResourceAndWait(serverfarm)
```
theunrepentantgeek, line 27: Are any of the properties of server farm updatable after creation? It might be a good idea to test that updates work as well as creation.

v2/internal/controllers/crd_web_test.go
```diff
@@ -0,0 +1,91 @@
...
+			Enabled:             to.BoolPtr(true),
+			Owner:               testcommon.AsOwner(rg),
+			Location:            tc.AzureRegion,
+			ServerFarmReference: tc.MakeReferenceFromResource(serverfarm),
+		},
+	}
+	tc.ExportAsSample(serverfarm)
+
+	tc.CreateResourceAndWait(serverfarm)
+	tc.CreateResourceAndWait(site)
```
theunrepentantgeek, line 63: Similar comment to what I said above. It may be worth doing an update to the site after it's been created.

v2/tools/generator/internal/functions/index_registration_function.go
```diff
@@ -200,7 +200,8 @@ func (f *IndexRegistrationFunction) handleArray(result *dst.Ident, locals *astmo
 func (f *IndexRegistrationFunction) handleMap(result *dst.Ident, locals *astmodel.KnownLocalsSet, ident dst.Expr, remainingChain []*astmodel.PropertyDefinition) []dst.Stmt {
 	p := remainingChain[0]
 
-	key := locals.CreateLocal("key")
+	// key in the loop would always be un-used.
+	key := "_"
```
theunrepentantgeek, line 204: Is this the first time we've used handle map? It seems weird to need to make this change if we've already used it, and wouldn't this change also impact across additional generated files? I don't see any in this PR.

theunrepentantgeek (approved):

v2/tools/generator/internal/functions/index_registration_function.go
super-harsh: Yeah, that was weird for me.. I did check that a few times and did not find any usages.

theunrepentantgeek: I'd expect our linters to catch anything the compiler missed, so no worries there. Seems weird is all.

v2/internal/controllers/crd_web_test.go
super-harsh: Done!

v2/internal/controllers/crd_web_test.go
super-harsh: Done!

v2/internal/controllers/crd_web_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	tc := globalTestContext.ForTest(t)
+
+	rg := tc.CreateTestResourceGroupAndWait()
+
+	serverfarm := newServerFarm(tc, rg)
+
+	site := &v1beta20220301.Site{
+		ObjectMeta: tc.MakeObjectMeta("function"),
+		Spec: v1beta20220301.Sites_Spec{
```
matthchr, line 57: So, according to [the documentation](https://docs.microsoft.com/en-us/azure/azure-functions/functions-create-first-function-resource-manager?tabs=azure-cli), if we want to use this API to make a Function, we need to allow users to set a value in `siteConfig.AppSettings` that looks like `"[format('DefaultEndpointsProtocol=https;AccountName={0};EndpointSuffix={1};AccountKey={2}', variables('storageAccountName'), environment().suffixes.storage, listKeys(resourceId('Microsoft.Storage/storageAccounts', variables('storageAccountName')), '2021-08-01').keys[0].value)]`

The issue is that we don't have a way to do that securely, and we also don't have a way to let them dynamically build that string (since AFAIK we don't support exporting such a field on the storage side).

We could add the ability to export such a field on the storage side but that still leaves us with an inability to specify a connection string here in a secure reference-based way.

Maybe we need to think about somehow being able to tag the `KeyValuePair` type or whatever it's called in `AppSettings` to allow either a raw string _or_ a reference? That's not support we have today though.

super-harsh: Holding this off until a user actually asks for it and have a user scenario. For now have added a TODO in here to track.

v2/internal/controllers/crd_web_serverfarm_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	tc.DeleteResourcesAndWait(serverfarm)
+
+	exists, _, err := tc.AzureClient.HeadByID(
+		tc.Ctx,
+		armId,
+		string(v1beta20220301.APIVersion_Value))
+	tc.Expect(err).ToNot(gomega.HaveOccurred())
+	tc.Expect(exists).To(gomega.BeFalse())
+
```
matthchr, line 43: super minor: Unneeded newline here

matthchr (approved):

PR #2466, opened 2022-08-25
Skip generating tests using gopter for large structs

theunrepentantgeek:

Property tests using [gopter](https://github.com/leanovate/gopter) have a limitation where they panic if a struct his more than 50 fields. This is due to [a limitation of the GO runtime](https://github.com/golang/go/issues/54669). 

This PR specifically excludes these large objects from our testing, giving us a workaround for the problem.

A log line is generated for each. Exclusion so that we know that it's happening.

gif

### DIFF
--- a/v2/tools/generator/internal/codegen/pipeline/json_serialization_test_cases.go
+++ b/v2/tools/generator/internal/codegen/pipeline/json_serialization_test_cases.go
@@ -8,6 +8,7 @@ package pipeline
 import (
 	"context"
 	"github.com/pkg/errors"
+	"k8s.io/klog/v2"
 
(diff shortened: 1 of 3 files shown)


super-harsh (approved):
Thanks for this. Should we consider putting the code in a common place and reuse?

theunrepentantgeek:
The most important part of each change is the comment explaining _why we need it_, and I think we'd want to retain that at each location. 

While there is a lot of similarity between each check, each of them is somewhat unique because of the context it's in, so I'm not sure how much we could refactor out into a common method., I suspect we'd end up with a similar amount of code.

PR #2472, opened 2022-08-26
Add support for AppConfiguration/ConfigurationStore

### LINKED ISSUE #1390: Feature: Integration with Azure App Configuration
As a k8s user, I would like to be able to define and retrieve resources from Azure App Configuration

super-harsh:
Closes #1390 

This PR adds support for AppConfiguration/ConfigurationStore

**If applicable**:
- [x] this PR contains tests


v2/api/appconfiguration/v1beta20220501/configuration_store_types_gen.go
```diff
@@ -0,0 +1,2787 @@
...
+// Generated from: https://schema.management.azure.com/schemas/2022-05-01/Microsoft.AppConfiguration.json#/definitions/ResourceIdentity
+type ResourceIdentity struct {
+	// Type: The type of managed identity used. The type 'SystemAssigned, UserAssigned' includes both an implicitly created
+	// identity and a set of user-assigned identities. The type 'None' will remove any identities.
+	Type *ResourceIdentity_Type `json:"type,omitempty"`
+
+	// UserAssignedIdentities: The list of user-assigned identities associated with the resource. The user-assigned identity
+	// dictionary keys will be ARM resource ids in the form:
+	// '/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{identityName}'.
+	UserAssignedIdentities map[string]v1.JSON `json:"userAssignedIdentities,omitempty"`
```
matthchr, line 1762: Prune this in the azure-arm.yaml like we do for other `UserAssignedIdentities`. Search `UserAssignedIdentities` in `azure-arm.yaml` to see what I am talking about.

v2/internal/controllers/crd_appconfiguration_configstore_test.go
```diff
@@ -0,0 +1,64 @@
...
+				Name: to.StringPtr("standard"),
+			},
+			PublicNetworkAccess: &publicNetworkAccessDisabled,
+		},
+		Status: v1beta20220501.ConfigurationStore_STATUS{},
+	}
+
+	tc.CreateResourceAndWait(cs)
+
+	tc.ExportAsSample(cs)
```
matthchr, line 48: I don't think this is supposed to be checked in, or am I misunderstanding @theunrepentantgeek?

v2/internal/controllers/crd_appconfiguration_configstore_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/Azure/azure-service-operator/v2/api/appconfiguration/v1beta20220501"
+	"github.com/Azure/azure-service-operator/v2/internal/testcommon"
+	"github.com/Azure/go-autorest/autorest/to"
+	"github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// If recording this test, might need to manually purge the old App config: az appconfig purge --name asotest-confstore-fsrajl
+
+func Test_AppConfiguration_ConfigurationStore_CRUD(t *testing.T) {
```
matthchr, line 20: This resource has keys too: https://docs.microsoft.com/en-us/rest/api/appconfiguration/stable/configuration-stores/list-keys?tabs=HTTP

Wondering if we should support keys here like we do for storage accounts, etc?

v2/internal/controllers/crd_appconfiguration_configstore_test.go
super-harsh: That should've been removed. My bad.

v2/api/appconfiguration/customizations/configuration_store_extensions.go
```diff
@@ -0,0 +1,149 @@
...
+		pager = confClient.NewListKeysPager(id.ResourceGroupName, obj.AzureName(), nil)
+		resp, err = pager.NextPage(ctx)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to retreive response")
+		}
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed listing keys")
+		}
+
+		keys = secretsByName(resp.Value)
```
theunrepentantgeek, line 79: You need a loop here to iterate through all the pages - there's no guarantee that everything you want will be available in the first page, nor even that the page will contain anything. AFAIK, it is entirely within contract to return empty pages or single row pages.

v2/api/appconfiguration/customizations/configuration_store_extensions.go
```diff
@@ -0,0 +1,149 @@
...
+		secrets.SecondaryKey != nil ||
+		secrets.SecondaryKeyID != nil ||
+		secrets.PrimaryReadOnlyConnectionString != nil ||
+		secrets.PrimaryReadOnlyKeyID != nil ||
+		secrets.PrimaryReadOnlyKey != nil ||
+		secrets.SecondaryReadOnlyConnectionString != nil ||
+		secrets.SecondaryReadOnlyKey != nil ||
+		secrets.SecondaryReadOnlyKeyID != nil {
+		return true
+	}
```
theunrepentantgeek, line 110: Needs `go fmt`? Inconsistent use of spaces vs tabs.

v2/api/appconfiguration/customizations/configuration_store_extensions.go
```diff
@@ -0,0 +1,149 @@
...
+
+	return false
+}
+
+func secretsByName(keys []*armappconfiguration.APIKey) map[string]armappconfiguration.APIKey {
```
theunrepentantgeek, line 115: Given the need to iterate through the pager, noted above, I suggest this method should accept a map that gets populated, rather than creating a new one each time.

theunrepentantgeek (changes requested):

v2/api/appconfiguration/customizations/configuration_store_extensions.go
super-harsh: It's already go fmted. I'll make it neater and push

v2/api/appconfiguration/customizations/configuration_store_extensions.go
super-harsh: Done!

v2/api/appconfiguration/customizations/configuration_store_extensions.go
super-harsh: Done!

v2/api/appconfiguration/customizations/configuration_store_extensions.go
```diff
@@ -0,0 +1,149 @@
...
+
+func secretsSpecified(obj *storage.ConfigurationStore) bool {
+	if obj.Spec.OperatorSpec == nil || obj.Spec.OperatorSpec.Secrets == nil {
+		return false
+	}
+
+	secrets := obj.Spec.OperatorSpec.Secrets
+
+	if secrets.PrimaryConnectionString != nil ||
+		secrets.PrimaryKeyID != nil ||
```
matthchr: What is `PrimaryKeyID`/`SecondaryKeyID`?

matthchr: Oh doh, I see from their API...

v2/api/appconfiguration/customizations/configuration_store_extensions.go
```diff
@@ -94,35 +97,31 @@ func secretsSpecified(obj *storage.ConfigurationStore) bool {
...
 		return true
 	}
 
 	return false
 }
 
-func secretsByName(keys []*armappconfiguration.APIKey) map[string]armappconfiguration.APIKey {
-	result := make(map[string]armappconfiguration.APIKey)
-
+func secretsByName(keys []*armappconfiguration.APIKey, result map[string]armappconfiguration.APIKey) {
```
theunrepentantgeek, line 118: Now wondering about the name of this method, given its now a repeated-action, not a transformation. Maybe `addSecretsToMap()`?

theunrepentantgeek (approved):

v2/api/appconfiguration/customizations/configuration_store_extensions.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func secretsSpecified(obj *storage.ConfigurationStore) bool {
+	if obj.Spec.OperatorSpec == nil || obj.Spec.OperatorSpec.Secrets == nil {
+		return false
+	}
+
+	secrets := obj.Spec.OperatorSpec.Secrets
+
+	if secrets.PrimaryKeyID != nil ||
```
matthchr, line 100: I actually think you can remove `PrimaryKeyID` and all of the other `ID` fields? From what I can tell, it's just a readonly ID that identifies if the key has changed. I don't think their clients actually use/need it, it's like the `LastModifiedTime`. There in the response but doesn't need to be exposed to the pods?

v2/api/appconfiguration/customizations/configuration_store_extensions.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func secretsToWrite(obj *storage.ConfigurationStore, keys map[string]armappconfiguration.APIKey) ([]*v1.Secret, error) {
+	operatorSpecSecrets := obj.Spec.OperatorSpec.Secrets
+	if operatorSpecSecrets == nil {
+		return nil, errors.Errorf("unexpected nil operatorspec")
+	}
+
+	collector := secrets.NewSecretCollector(obj.Namespace)
+	collector.AddSecretValue(operatorSpecSecrets.PrimaryConnectionString, *keys["Primary"].ConnectionString)
```
matthchr, line 134: Should we protect against if `keys` doesn't have these values?

Maybe add some if-guards and return error rather than here where we would panic? Could read them into vars too to avoid repeating the same `*keys["Primary"]` over and over too.

v2/api/appconfiguration/customizations/configuration_store_extensions.go
super-harsh: I had a same thought on IDs, Then I ended up creating a resource in Azure and checked on portal. While reading the Access Keys, there it shows ID to be `credential`. That's the only reason for adding Ids
image

super-harsh: Its the same for Read-Write keys btw.

v2/internal/controllers/crd_appconfiguration_configstore_test.go
theunrepentantgeek: No misunderstanding; it's a dev time helper method.

v2/api/appconfiguration/customizations/configuration_store_extensions.go
matthchr: I think thats their UI weirdness. The example API response shows :
```
    {
      "id": "439AD01B4BE67DB1",
      "name": "Primary",
      "value": "000000000000000000000000000000000000000000000000000000",
      "connectionString": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
      "lastModified": "2018-04-24T16:30:54+00:00",
      "readOnly": false
    },
```

The `ID` field isnt long enough to be what they are showing as the `credential`.

super-harsh: I think the example API is not correct or too old? Check the response below.

List keys using `az rest` on resource:

image

On portal: 

image

v2/api/appconfiguration/customizations/configuration_store_extensions.go
super-harsh: Done!

v2/api/appconfiguration/customizations/configuration_store_extensions.go
matthchr: Hmm, the fact that it's in the connection string suggests it may actually be needed, although I can't find any examples of theirs showing its use (everybody uses the connectionString).

Alright, happy to keep it then.

matthchr (approved):

v2/api/appconfiguration/customizations/configuration_store_extensions.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	secondary, ok := keys["Secondary"]
+	if ok {
+		collector.AddSecretValue(operatorSpecSecrets.SecondaryConnectionString, *secondary.ConnectionString)
+		collector.AddSecretValue(operatorSpecSecrets.SecondaryKeyID, *secondary.ID)
+		collector.AddSecretValue(operatorSpecSecrets.SecondaryKey, *secondary.Value)
+	}
+
+	SecondaryReadOnly, ok := keys["Secondary Read Only"]
```
matthchr, line 155: minor: `secondaryReadOnly`

matthchr (approved):
LGTM, had only one minor style comment

PR #2475, opened 2022-08-29
Limit concurrency of CI builds

theunrepentantgeek:

Limits the concurrency of CI builds because the build machines are small and we suspect we are overloading them.

I'm hoping the CI build for this PR will run with the workflow modification included. If it doesn't, we'll need to merge this, and then run some tests.

gif

### DIFF
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -96,7 +96,7 @@ jobs:
           container_id=${{steps.devcontainer.outputs.id}}
 
           set +e # don't exit instantly on failure, we need to produce Markdown summary
-          docker exec "$container_id" task ci
+          docker exec "$container_id" task ci --concurrency 1 # limit concurrency because build machines are small
           EXIT_CODE=$?
           set -e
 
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-validation.yml
```diff
@@ -114,8 +114,8 @@ jobs:
       - name: Build docker image & build configuration YAML
         run: |
           container_id=${{steps.devcontainer.outputs.id}}
-          docker exec "$container_id" task controller:docker-build-and-save
-          docker exec "$container_id" task controller:run-kustomize-for-envtest
+          docker exec "$container_id" task controller:docker-build-and-save --concurrency 1 # limit concurrency because build machines are small
```
super-harsh, line 117: Wouldn't this hit timeout/take too much time while running the tests in parallel?

theunrepentantgeek: If our build machines were highly spec'd, limiting concurrency to just one task at a time would leave processors idle and memory unused. But we know from experience that the build machines are small, just two vCPU's, and so my hypothesis is that running too many tasks at the same time results in them fighting for resources and everything gets slowed down.

super-harsh (approved):

theunrepentantgeek:
Based on some rudimentary testing, I think this makes some improvement.

image

PR #2476, opened 2022-08-30
Add links to reference docs from supported resources page

### LINKED ISSUE #2360: Add direct links to resource documentation from summary page

On the [Supported Resources](https://azure.github.io/azure-service-operator/reference/) page of our documentation, we currently link to a sample for each resource, but not to the documentation.

**Describe the improvement**

One approach would be to link from the name of the resource (in the first column) to the documentation page:

image

Another would be to add a new column, explicitly called "Documentation". 

Maybe we do both?

**Additional context**

Closely related to #2361

theunrepentantgeek:

Makes each resource name into a link that goes to the API reference docs for that specific version of the resource.

Closes #2360 

URL is template driven to allow easy reuse for other targets.

gif


super-harsh (approved):
LGTM! Not sure about a scenario where we have generated a new resource and still don't have the docs generated from the tool ( which we run on new release ). Then the link will stay broken.

PR #2483, opened 2022-09-02
Add pre release upgrade tests

### LINKED ISSUE #2235: Automate upgrade testing
For the Beta-0 release, @matthchr manually tested that resources deployed with Alpha-5 would correctly operate with the new release. We should automate these tests so that we can ensure we don't accidentally break backwards compatibility in any future release.

super-harsh:
Closes #2235 

This PR adds task, tests and workflow for pre release upgrade testing.

Decided to create a seperate workflow for this, as we can extend this to add more pre release tests in the future. 

Will update the `Creating a new release` doc in the follow up PR to include the link to new workflow for manual triggering before release. 

**If applicable**:
- [x] this PR contains tests


v2/test/pre-release/pre_release_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	tc := globalTestContext.ForTest(t)
+	tc.Namespace = preReleaseNamespace
+
+	// TODO: Will have to change the version here when we go from beta to stable
+	vnetAfterUpgrade := newVnet(tc, tc.Namer.GenerateName("vn"))
+	tc.CreateResourceAndWait(vnetAfterUpgrade)
+	tc.DeleteResourcesAndWait(vnetAfterUpgrade)
+
+	// Patch for existing resource. This resource already will exist in kind as will be created without cleanup in test 'Test_Pre_Release_ResourceCanBeCreated_BeforeUpgrade'.
```
theunrepentantgeek, line 68: How do you control the sequence of execution?

I thought `go test` was deliberately non-deterministic in the order it executes tests.

super-harsh: Yep, I decided to use these tests instead of `kubectl` so that they can fail/pass when something goes wrong. We're controlling them through the task [here](https://github.com/Azure/azure-service-operator/pull/2483/files#diff-cd2d359855d0301ce190f1ec3b4c572ea690c83747f6df61c9340720e3d2425eR264)

.github/workflows/pre-release-tests.yaml
```diff
@@ -0,0 +1,55 @@
+name: Pre Release tests
+on:
+  workflow_dispatch:
```
matthchr, line 3: Is this intentionally the only thing that triggers this workflow? Can we automate them somehow... maybe once a week on the weekend?

Taskfile.yml
```diff
@@ -240,6 +240,39 @@ tasks:
...
+      --create-namespace \
+      --namespace={{.ASO_NAMESPACE}} \
+      --set azureSubscriptionID=$AZURE_SUBSCRIPTION_ID \
+      --set azureTenantID=$AZURE_TENANT_ID \
+      --set azureClientID=$AZURE_CLIENT_ID \
+      --set azureClientSecret=$AZURE_CLIENT_SECRET"
+    - "kubectl create namespace pre-release"
+    - "kubectl wait --for=condition=ready --timeout=2m pod -n {{.ASO_NAMESPACE}} -l control-plane=controller-manager"
+    - "go test -timeout 15m -v -run Test_Pre_Release_ResourceCanBeCreated_BeforeUpgrade ./test/pre-release"
+    - "helm upgrade --devel aso2 ./charts/azure-service-operator \
```
matthchr, line 265: Can you not just call `controller:install-helm` here, rather than duplicating it? From what I can tell, this is the same content as that command?

Even if you can't, presumably you should still use the `{{.LOCAL_REGISTRY_CONTROLLER_DOCKER_IMAGE}}` here rather than hardcode `localhost:5000/azureserviceoperator:latest`.

v2/test/pre-release/pre_release_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	rg := tc.NewTestResourceGroup()
+	rg.Name = resourceGroupName
+
+	vnet := newVnet(tc, vnetBeforeUpgradeName)
+	// We want to use 'CreateResourceUntracked' here as we don't want the resource cleanup for this test. We will be checking the
+	// backward compatibility for these resources after we upgrade the controller.
+	tc.CreateResourceAndWaitWithoutCleanup(rg)
+	tc.CreateResourceAndWaitWithoutCleanup(vnet)
+
```
matthchr, line 38: minor: unneeded newline

v2/test/pre-release/pre_release_test.go
```diff
@@ -0,0 +1,78 @@
...
+	network "github.com/Azure/azure-service-operator/v2/api/network/v1beta20201101"
+	"github.com/Azure/azure-service-operator/v2/internal/testcommon"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+	"github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+const (
+	resourceGroupName     = "aso-sample-rg"
```
matthchr, line 19: minor: I think you should use a name like `aso-upgrade-test-rg` or something. Not a name that we already use in samples. Less risk of us colliding with ourselves.

v2/test/pre-release/pre_release_test.go
```diff
@@ -0,0 +1,78 @@
...
+	t.Parallel()
+
+	tc := globalTestContext.ForTest(t)
+	tc.Namespace = preReleaseNamespace
+
+	rg := tc.NewTestResourceGroup()
+	rg.Name = resourceGroupName
+
+	vnet := newVnet(tc, vnetBeforeUpgradeName)
+	// We want to use 'CreateResourceUntracked' here as we don't want the resource cleanup for this test. We will be checking the
```
matthchr, line 34: minor: comment doesn't match method name

v2/test/pre-release/pre_release_test.go
```diff
@@ -0,0 +1,78 @@
...
+	tc.Namespace = preReleaseNamespace
+
+	// TODO: Will have to change the version here when we go from beta to stable
+	vnetAfterUpgrade := newVnet(tc, tc.Namer.GenerateName("vn"))
+	tc.CreateResourceAndWait(vnetAfterUpgrade)
+	tc.DeleteResourcesAndWait(vnetAfterUpgrade)
+
+	// Patch for existing resource. This resource already will exist in kind as will be created without cleanup in test 'Test_Pre_Release_ResourceCanBeCreated_BeforeUpgrade'.
+	vnetBeforeUpgrade := newVnet(tc, vnetBeforeUpgradeName)
+	defer tc.DeleteResourceAndWait(vnetBeforeUpgrade)
```
matthchr, line 70: I think you should put this `vnetBeforeUpgrade` creation and the `defer tc.DeleteResourceAndWait()` at the top of this method, in case this method fails before you get to this step we'll still delete things. Also should probably do the `defer` on the `rg` rather than the vnet, otherwise the vnet might get leaked.

v2/test/pre-release/pre_release_test.go
matthchr: Actually, you should also use the `TEST_LIVE_RESOURCE_PREFIX` on this resource, which is `asolivetest`, that way if we do for some reason end up leaking this resource, the ` cleanup-old-live-azure-resources` step in the Taskfile which is already run periodically will clean it up.

v2/test/pre-release/pre_release_test.go
matthchr: Sorry I meant: "otherwise the rg might get leaked" Obviously if you delete the vnet it won't get leaked =)

v2/test/pre-release/pre_release_test.go
```diff
@@ -0,0 +1,82 @@
...
+	old := vnetBeforeUpgrade.DeepCopy()
+	vnetBeforeUpgrade.Spec.Tags = map[string]string{"newTag": "afterUpgradeTag"}
+	tc.Patch(old, vnetBeforeUpgrade)
+	tc.Eventually(vnetBeforeUpgrade).Should(tc.Match.BeProvisioned(1))
+	tc.GetResource(types.NamespacedName{Name: vnetBeforeUpgrade.Name, Namespace: vnetBeforeUpgrade.Namespace}, vnetBeforeUpgrade)
+	tc.Expect(vnetBeforeUpgrade.Status.Tags).To(gomega.HaveKey("newTag"))
+}
+
+func getLiveTestResourceName(name string) string {
+	return fmt.Sprintf("%s-%s", testcommon.LiveResourcePrefix, name)
```
matthchr, line 81: So technically, you could do this by passing `testcommon.LiveResourcePrefix` to `nameConfig` on line 39 of the `suite_test.go`, and then creating a custom namer in the test via:
`newNamer := tc.Namer.WithNumRandomChars(0)`.

Then you can do `newNamer.GenerateName(resourceGroupName)`. to generate a name that is the `resourceGroupName` + just the prefix.

The only advantage I see to doing it that way is that you don't need this local function to do that and it mirrors the way the live tests do it so that if there's any change in theory this place would also get it.

Pretty minor though.

matthchr (approved):

PR #2487, opened 2022-09-07
Add support for CosmosDB Role Assignments

matthchr:
**If applicable**:
- [x] this PR contains tests


theunrepentantgeek (approved):

v2/api/documentdb/customizations/sql_role_assignment_extension_types.go
```diff
@@ -0,0 +1,54 @@
...
+	return details, nil
+}
+
+// isRetryableConflict checks the passed error to see if it is a retryable conflict, returning true if it is.
+func isRetryableRoleAssignmentBadRequest(err *genericarmclient.CloudError) bool {
+	if err == nil {
+		return false
+	}
+
+	if err.Code() != "BadRequest" {
```
super-harsh, line 49: Would err.Code() == "400" ? or text?

super-harsh (approved):

v2/internal/testcommon/samples_tester.go
```diff
@@ -48,6 +48,9 @@ var exclusions = []string{
 
 	// Excluding dbformysql/user as is not an ARM resource
 	"user",
+
+	// Excluding sqlroleassignment because it needs to reference a principalId
+	"sqlroleassignment",
```
super-harsh, line 53: can this be done the same way we do `roleassignment`?

v2/api/documentdb/customizations/sql_role_assignment_extension_types.go
matthchr: `err.Code()` is a string always. It's part of the contract from Azure w.r.t. what the error shape is.

v2/internal/testcommon/samples_tester.go
matthchr: It could but since that requires hardcoding of a principalID or editing of the sample before re-recording I decided not to do it and instead just bypass this test for now. I'm got a proposal out that will make this easier because we can then link via the UMI name. Once that's implemented we can run this sample.

PR #2489, opened 2022-09-07
Design document for reading status properties of other resources

matthchr:
**If applicable**:
- [ ] this PR contains tests


theunrepentantgeek (approved):

docs/hugo/content/design/ADR-2022-09-Reading-Status-Properties-Of-Resources.md
```diff
@@ -0,0 +1,261 @@
...
+  owner:
+    name: samplevnet
+    group: network.azure.com
+    kind: VirtualNetwork
+  principalId:
+    ref:
+      name: sample-uai
+      group: manangedidentity.azure.com
+      kind: UserAssignedIdentity
+      path: status.principalId
```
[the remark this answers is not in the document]
matthchr: Version isn't required for these because there's only 1 logical entity even if there are N versions of it. We're linking to the logical entity here.

theunrepentantgeek, line 167: Worth observing (perhaps in a separate `Implementation` section below `Pros` & `Cons`) that evaluation of an arbitrary path has some challenges - we could code it ourselves, and handle all the edge cases, or take a shortcut 
(perhaps by serialization to JSON and using a query library) at the risk of not having control over some cases.

theunrepentantgeek: Also make a note that every ASO resource would have to register for events related to every other ASO resource, resulting in a significant load on the Kubernetes Event system, even for workloads that don't use this feature.

A good principle to reference here is "Pay for Play" - if someone isn't using a particular ASO feature, they shouldn't have to pay very much (if anything) for it's unused presence.

theunrepentantgeek: Because there would be literally no restriction on which property a user might reference from another resource's status, it would be possible for one resource to consume a volatile property from another (say, CPU load from a VM).

If a user did this, every reconcile would result in an update to the resource, which might trigger a busy wait loop and a nasty failure mode.

docs/hugo/content/design/ADR-2022-09-Reading-Status-Properties-Of-Resources.md
```diff
@@ -0,0 +1,261 @@
...
+4. Uncertain general reconcile behavior. This reference constitutes a dependency on the referenced resource. Does this mean that ideally, the resource holding the reference will be `Watch`-ing the referenced resource and immediately react to changes in its dependencies? Implementing that dynamic watch behavior given the current structure of controller-runtime would be very difficult.
+
+### Option 3: Support reading/writing to a ConfigMap or Secret, and indirect through that
+
+The idea here is that `UserManagedIdentity` would allow exporting a Secret or ConfigMap value, and then the `RoleAssignment` resource would support (optionally) sourcing from that store instead of having the value hardcoded.
+The main difference between this and what we have now is for references we can't replace the entire reference with a `SecretReference`, we need a new type like `OptionalSecretReference`:
+
+We would only opt a select few resource properties into this `OptionalSecretReference` structure. We could (via serialization magic) support new properties adopting this shape in a nonbreaking way, in case there are properties we didn't think
```
theunrepentantgeek, line 188: **Task** does this kind of serialization magic in a number of places - one example is around [variables](https://taskfile.dev/usage/#variables) where they can be directly defined:

```
    vars:
      VAR: Hello!
```
or dynamically defined:
```
    vars:
      GIT_COMMIT:
        sh: git log -n 1 --format=%h
```

We can reference the [Task implementation](https://github.com/go-task/task/blob/master/taskfile/var.go) for an approach.

docs/hugo/content/design/ADR-2022-09-Reading-Status-Properties-Of-Resources.md
```diff
@@ -0,0 +1,261 @@
...
+spec:
+  location: Germany West Central
+  owner:
+    name: samplevnet
+    group: network.azure.com
+    kind: VirtualNetwork
+  principalId:
+    ref:
+      name: uai-secret
+      key: principalid
```
theunrepentantgeek, line 228: I think we could simplify this and make it more self documenting:
``` yaml
principalId:
  fromConfigmap: uai-secret
  key: principalid
```

docs/hugo/content/design/ADR-2022-09-Reading-Status-Properties-Of-Resources.md
```diff
@@ -0,0 +1,261 @@
...
+  roleDefinitionReference:
+    armId: /subscriptions/00000000-0000-0000-0000-000000000000/providers/Microsoft.Authorization/roleDefinitions/b24988ac-6180-42a0-ab88-20f7382dd24c
+```
+
+Pros:
+1. Generic
+2. Would solve another related problem as well (see [2349](https://github.com/Azure/azure-service-operator/issues/2349) and [2350](https://github.com/Azure/azure-service-operator/issues/2350), that of how a customer pod can access some of these variables.
+3. Solves the priviledge escalation risk of option 2
```
theunrepentantgeek, line 236: Break this out as a separate `Security` section, and cover that it resolves the privilege escalation risk by requiring the source resource (the identity) to opt-in to sharing the property.

docs/hugo/content/design/ADR-2022-09-Reading-Status-Properties-Of-Resources.md
```diff
@@ -0,0 +1,261 @@
...
+
+Cons:
+1. More complicated to implement. Operator would need to make sure to watch the ConfigMap or Secret for changes and trigger reconciliation based on that.
+
```
theunrepentantgeek, line 240: Add in an `Implementation` section and contrast with Option 2, pointing out that the resources will only need to subscribe to configMap events, keeping the load on the Kubernetes events subsystem fairly low.

theunrepentantgeek (approved):
I've made comments reflecting all my notes; many of these overlap our discussion in the review meeting.

super-harsh (approved):
LGTM. Would poll in for option 3 too.

docs/hugo/content/design/ADR-2022-09-Reading-Status-Properties-Of-Resources.md
```diff
@@ -0,0 +1,261 @@
...
+    // Name of the identity in Kuberentes
+    Name string `json:"name,omitempty"`
+
+    // PrincipalId of the identity in AAD. This is mutually exclusive with Name.
+    PrincipalId string `json:"principalId,omitempty"`
+}
+```
+
+Usage would be something like:
+```yaml
```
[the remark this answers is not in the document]
matthchr: Did this

docs/hugo/content/design/ADR-2022-09-Reading-Status-Properties-Of-Resources.md
```diff
@@ -0,0 +1,261 @@
...
+    ref:
+      name: uai-secret
+      key: principalid
+  roleDefinitionReference:
+    armId: /subscriptions/00000000-0000-0000-0000-000000000000/providers/Microsoft.Authorization/roleDefinitions/b24988ac-6180-42a0-ab88-20f7382dd24c
+```
+
+Pros:
+1. Generic
+2. Would solve another related problem as well (see [2349](https://github.com/Azure/azure-service-operator/issues/2349) and [2350](https://github.com/Azure/azure-service-operator/issues/2350), that of how a customer pod can access some of these variables.
```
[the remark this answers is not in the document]
matthchr: Yeah this is the general sentiment/agreement.

docs/hugo/content/design/ADR-2022-09-Reading-Status-Properties-Of-Resources.md
matthchr: > Worth observing (perhaps in a separate Implementation section below Pros & Cons) that evaluation of an arbitrary path has some challenges - we could code it ourselves, and handle all the edge cases, or take a shortcut
(perhaps by serialization to JSON and using a query library) at the risk of not having control over some cases.

I've just put this into the existing con, as I think it's a con and makes more sense there. I've added `Implementation` as a section in a few other places where there are "notes" on implementation that aren't strictly a pro/con.

docs/hugo/content/design/ADR-2022-09-Reading-Status-Properties-Of-Resources.md
matthchr: Changed this to:
```
    configMap: uai-config  # Reference the config map created above
    key: principalid
```

I think the `from` prefix is relatively superfluous because it exists on a property that is clearly being supplied by the user. For `SecretReference` we don't say `fromName` or something like that we just say `name` and I think it makes sense to follow that pattern here as well since this is probably going to be called `OptionalConfigReference`.

Thoughts?

docs/hugo/content/design/ADR-2022-09-Reading-Status-Properties-Of-Resources.md
matthchr: Done

docs/hugo/content/design/ADR-2022-09-Reading-Status-Properties-Of-Resources.md
matthchr: I've just listed this in `Pros`, as it feels like it fits there better than in the Implementation section. I did add an implementation section here talking about custom serialization/deserialization though.

docs/hugo/content/design/ADR-2022-09-Reading-Status-Properties-Of-Resources.md
theunrepentantgeek: Happy with that. :+1:

docs/hugo/content/design/ADR-2022-09-Reading-Status-Properties-Of-Resources.md
```diff
@@ -0,0 +1,301 @@
+---
+title: '2022-09: Reading Status properties from other ASO resources'
+---
+
+## Context
+
+It isn't currently possible to create a `UserManagedIdentity` and use it in the same YAML, because 
+usage of it requires knowing its `PrincipalId` which is generated by Azure only when the identity is created.
+This is discussed in a variety of issues. See [2435](https://github.com/Azure/azure-service-operator/issues/2435) 
+and [2350](https://github.com/Azure/azure-service-operator/issues/2350) for example.
```
theunrepentantgeek, line 10:
```suggestion
and [2350](https://github.com/Azure/azure-service-operator/issues/2350) for two examples.
```

docs/hugo/content/design/ADR-2022-09-Reading-Status-Properties-Of-Resources.md
```diff
@@ -0,0 +1,301 @@
...
+This is also somewhat related to [2474](https://github.com/Azure/azure-service-operator/issues/2474) as well.
```
theunrepentantgeek, line 11:
```suggestion
This is also somewhat related to [2474](https://github.com/Azure/azure-service-operator/issues/2474).
```

docs/hugo/content/design/ADR-2022-09-Reading-Status-Properties-Of-Resources.md
```diff
@@ -0,0 +1,301 @@
...
+infrastructure as code and deploy things simply with no manual steps. 
+
+It's worth noting that the general problem of reading a value from another resource to use in _this_ 
+resource comes up in more than just the AAD context.
+
+
+### Options
+
+In all of the proposed solutions, an annotation would need to be added into `azure-arm.yaml` which indicates that the 
+property in question is special. Something like:
```
theunrepentantgeek, line 50:
```suggestion
In all of the proposed solutions, an annotation would need to be added into `azure-arm.yaml` to indicate the 
property in question is special:
```

docs/hugo/content/design/ADR-2022-09-Reading-Status-Properties-Of-Resources.md
```diff
@@ -0,0 +1,301 @@
...
+```yaml
+RoleAssignmentProperties:
+  PrincipalId:
+    $aadReference: true
+```
+
+or (for the more generic proposals)
```
theunrepentantgeek, line 57:
```suggestion
or, for the more generic proposals:
```

docs/hugo/content/design/ADR-2022-09-Reading-Status-Properties-Of-Resources.md
```diff
@@ -0,0 +1,301 @@
...
+2. Simple to use
+3. Can give a good error message if the identity in question doesn't exist
+
+Cons:
+
+1. Not generic. What if some resources require `ClientId` instead of `PrincipalId`? We would need a 
+   new type for that.
+
+
+### Option 2: Support a generic resource for reading from other ASO resources Status
```
theunrepentantgeek, line 119:
```suggestion
### Option 2: Support a generic definition for reading from other ASO resource's Status
```

docs/hugo/content/design/ADR-2022-09-Reading-Status-Properties-Of-Resources.md
```diff
@@ -0,0 +1,301 @@
...
+   we could use an existing JMESPath library but doing so would require serializing objects to JSON regularly,
+   which could get expensive from a performance perspective.
+4. Uncertain general reconcile behavior. This reference constitutes a dependency on the referenced 
+   resource. Does this mean that ideally, the resource holding the reference will be `Watch`-ing the
+   referenced resource and immediately react to changes in its dependencies? Implementing that 
+   dynamic watch behavior given the current structure of controller-runtime would be very difficult. In the worst
+   case it would mean every ASO resource having a watch on every other resource.
+
+Security:
+1. Risk for privilege escalation, where a user in the namespace that doesn't have permission to read resources of type
```
theunrepentantgeek, line 203:
```suggestion
Security:
1. Risk for privilege escalation, where a user in the namespace that doesn't have permission to read resources of type 

```

docs/hugo/content/design/ADR-2022-09-Reading-Status-Properties-Of-Resources.md
```diff
@@ -0,0 +1,301 @@
...
+1. Risk for privilege escalation, where a user in the namespace that doesn't have permission to read resources of type 
+   `X` instead asks ASO to read that resource. Azure then reflects that value back to them in the error message.
+
+
+### Option 3: Support reading/writing to a ConfigMap or Secret, and indirect through that
+
+<!-- TODO: Expand on serialization trickery -->
+<!-- TODO: Implementation notes. For example on the second, how does eventing work? -->
+
+The idea here is that `UserManagedIdentity` would allow exporting ConfigMap value, and then the
```
theunrepentantgeek, line 212:
```suggestion
The idea here is that `UserManagedIdentity` would allow exporting `PrincipalId` to a ConfigMap value, and then the 
```

docs/hugo/content/design/ADR-2022-09-Reading-Status-Properties-Of-Resources.md
```diff
@@ -0,0 +1,301 @@
...
+   `X` instead asks ASO to read that resource. Azure then reflects that value back to them in the error message.
+
+
+### Option 3: Support reading/writing to a ConfigMap or Secret, and indirect through that
+
+<!-- TODO: Expand on serialization trickery -->
+<!-- TODO: Implementation notes. For example on the second, how does eventing work? -->
+
+The idea here is that `UserManagedIdentity` would allow exporting ConfigMap value, and then the 
+`RoleAssignment` resource would support (optionally) sourcing from that store instead of having the value hardcoded.
```
theunrepentantgeek, line 213:
```suggestion
`RoleAssignment` resource would support (optionally) reading from that store instead of having a hardcoded value.
```

docs/hugo/content/design/ADR-2022-09-Reading-Status-Properties-Of-Resources.md
```diff
@@ -0,0 +1,301 @@
...
+  roleDefinitionReference:
+    armId: /subscriptions/00000000-0000-0000-0000-000000000000/providers/Microsoft.Authorization/roleDefinitions/b24988ac-6180-42a0-ab88-20f7382dd24c
+```
+
+Pros:
+1. Generic
+2. Would solve related problems where customers want access to these variables in their pods as well. See: 
+   1. [2349](https://github.com/Azure/azure-service-operator/issues/2349)
+   2. [2350](https://github.com/Azure/azure-service-operator/issues/2350)
+   3. [2317](https://github.com/Azure/azure-service-operator/issues/2317).
```
theunrepentantgeek, line 269:
```suggestion
   1. [#2349](https://github.com/Azure/azure-service-operator/issues/2349) - Access serviceBusEndpoint in ServiceBus namespace
   2. [#2350](https://github.com/Azure/azure-service-operator/issues/2350) - Export configuration from status
   3. [#2317](https://github.com/Azure/azure-service-operator/issues/2317) - Export clientId from UserAssignedIdentity
```

docs/hugo/content/design/ADR-2022-09-Reading-Status-Properties-Of-Resources.md
```diff
@@ -0,0 +1,301 @@
...
+3. Solves the privilege escalation risk of option 2
+4. Resources only need to watch ConfigMap, as opposed to "all other ASO resources" as in option 2.
+
+Cons:
+1. More complicated to implement. Operator would need to make sure to watch the ConfigMap for changes and 
+   trigger reconciliation based on that.
```
theunrepentantgeek, line 275: More complicated than Option 1, but less so than Option 2.

docs/hugo/content/design/ADR-2022-09-Reading-Status-Properties-Of-Resources.md
```diff
@@ -0,0 +1,301 @@
...
+metadata:
+  name: 6a2d44f5-57d8-4916-9f46-ff7c9c1b338f
+spec:
+  location: Germany West Central
+  owner:
+    name: samplevnet
+    group: network.azure.com
+    kind: VirtualNetwork
+  principalId:
+    configMap: uai-config  # Reference the config map created above
```
theunrepentantgeek, line 258:
```suggestion
    configMap: uai-config  # Reference the config map value exported above
```

docs/hugo/content/design/ADR-2022-09-Reading-Status-Properties-Of-Resources.md
```diff
@@ -0,0 +1,301 @@
...
+which does this.
+
+## Decision
+
+We will go with Option 3, using `ConfigMaps` and `OptionalConfigReference`. This is the most flexible option and
+solves the most user requests while also limiting downsides.
+
+## Status
+
+Proposed.
```
theunrepentantgeek, line 289:
```suggestion
Agreed. Implementation pending.
```

😁

theunrepentantgeek (approved):

PR #2494, opened 2022-09-12
Fix fork check so it can be made required

matthchr:


super-harsh (approved):

theunrepentantgeek (approved):
Missing EOL at EOF, otherwise looks good.

PR #2499, opened 2022-09-15
Design document for supporting multiple credentials under global operator

super-harsh:
**If applicable**:
- [ ] this PR contains tests


docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
```diff
@@ -0,0 +1,153 @@
...
+1. Flexible, as users have the control for which resource they want to use which credential
+
+**Cons:**
+1. Will have to specify annotation on each resource explicitly
+
+### Option 3: Secret lookup + configuration using annotations
+
+Here, we would use a fixed pattern for namespaces secrets + annotations for per-resource-group and per-resource secrets. Operator would look up for the annotation(`serviceoperator.azure.com/credential-from`) on resource created, if annotation found, then use credentials from secret in annotation. If not, then look for the namespaced secret and use the namespaced credential.If none provided, operator shall continue with the global credential.
+
+NOTE: If namespaced secret or global secret is used, operator would inject `serviceoperator.azure.com/credential-from` annotation with the secret name value.
```
matthchr, line 72: Does it make sense to inject this for the global secret? It's supposed to point to a secret in _this_ namespace, right? But the credential for the global secret is not in this namespace probably.  I think we shouldn't inject this for the global secret but should still inject the `serviceoperator.azure.com/subscription` annotation I mention above.

For the namespace scoped secret, we could inject this annotation. can you call out the win of doing so here though? I'm assuming it's so that the user understands which secret we've actually used with this resource (rather than it being some implicit thing where they need to go "oh does it not have an annotation, then let me check the namespace secret). Which I think I like, but if you have other reasons call them out here too.

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
```diff
@@ -0,0 +1,153 @@
...
+
+**Annotation for per-resource-group and per-resource secret**
+```
+annotations: 
+  serviceoperator.azure.com/credential-from: any-secret
+```
+
+**Pros:**
+1. Users don't have to specify the configuration for each resource explicitly if using a namespaced secret
+2. Users get the flexibility of over-riding the credential they want to use as per resource and as per resource group
```
matthchr, line 87: minor:
```suggestion
2. Users get the flexibility of overriding the credential they want to use as per resource and as per resource group
```

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
```diff
@@ -0,0 +1,153 @@
...
+
+**Cons:**
+1. Could be complicated/confusing for users to understand the usage and credentials being used
```
matthchr, line 90: You could call out here the mitigation mentioned above of automatically adding the `credential-from` annotation when they use at least the namespace scope secret mode, so that it's more clear to them which secret is being used.

I personally think that mitigation makes this problem pretty minimal. Curious what @theunrepentantgeek thinks.

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Option 4: Secret lookup + configuration using global secret
+
+This option, relates to the option 3. However, the change is how we configure to use the per-resource-group and per-resource secret. We can have a unique slice property(CREDENTIAL_FROM) in the global secret where admin users can specify the secrets they want to load on the startup.
+Then only these loaded secrets would be used further for per-resource and per-resource-group. This option would restrict the other non-admin users(who don't have access to the global secret) to add their credential. 
+
+**Pros:**
+1. Users don't have to specify the configuration for each resource explicitly if using a namespaced secret
+2. Security conscious, not anyone in the cluster can add to use their credentials.
```
matthchr, line 99: I think this is also kinda a con, because it means that to onboard new credentials/users, the administrators of the ASO namespace must be contacted.

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
```diff
@@ -0,0 +1,153 @@
...
+**Cons:**
+1. Could be complicated/confusing for users to understand the usage and credentials being used
+2. Not very flexible, as pod will have to startup again if a new secret needs to be loaded
+
+## Handling for the failure cases and logging
+
+- Failure on loading a secret would result in failure on resource creation
+- Failure on authenticating the credentials would result in failure on resource creation
+- Logging should be clear about which credential is being used for a resource
+- Logging should be clear about the credential load/authentication failure
```
matthchr, line 110: minor: Mention errors can be reported through the standard `Condition` mechanism

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Milestones
+
+### Support for namespaced secrets
```
matthchr, line 114: minor: You could possibly collapse this section into just an ordered list (or a table?) rather than headings, since there's not a lot under them right now.

```
## Milestones

1. Namespaced secrets: 
2. per-ResourceGroup secrets
3. per-Resource secrets: Will not be implemented unless there is strong user demand. We believe that per-ResourceGroup secrets should be sufficiently granular.
```

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
```diff
@@ -0,0 +1,153 @@
...
+
+**Timelines:** TBD
+
+### Support for per-resource secrets
+
+We aim to support per-resource secrets under this milestone and configuration work required.
+
+**Timelines:** TBD
+
+### Support for credential cache
```
matthchr, line 132: Isn't there usually a Decision section where you propose a particular solution?

I think I prefer option 3 personally.

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
```diff
@@ -0,0 +1,153 @@
...
+
+TBD
+
+## Open questions
+1. Do we need to support both per-resource and per-resource-group credentials?
```
matthchr, line 137: IMO no, let's document per-resource in our design like you have. So we have considered how to implement it if we had to... but then let's say that we're not actually going to implement it unless there is strong user demand.

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
```diff
@@ -0,0 +1,153 @@
...
+2. Do we want a serviceoperator.azure.com/subscription-id annotation that we write? We can use this to protect against users accidentally updating credential to move resources between subscriptions.
```
matthchr, line 138: IMO yes (see my earlier comments). I think this is an important protection. It helps even the current global credential case. User can always get around it if they want by manually clearing that annotation.

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
```diff
@@ -0,0 +1,153 @@
...
+3. Spell out what fields are in aso-credential. clientid/subscriptionid/tenantId/etc?
+4. How are we going to make this work for Managed Identity?
+5. Replace the above-mentioned(in options) annotations with spec property?
```
matthchr, line 141: Honestly the more I look at the annotation the more I like it. Although possibly we should consider the name of the annotation and change it to something like `serviceoperator.azure.com/credential-name`. Then if we end up needing to support cross-namespace secrets (I dont think we will, and there are lots of security issues, etc) we can just use `serviceoperator.azure.com/credential-namespace`.

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Support for credential cache
+
+TBD
+
+## Open questions
+1. Do we need to support both per-resource and per-resource-group credentials?
+2. Do we want a serviceoperator.azure.com/subscription-id annotation that we write? We can use this to protect against users accidentally updating credential to move resources between subscriptions.
+3. Spell out what fields are in aso-credential. clientid/subscriptionid/tenantId/etc?
+4. How are we going to make this work for Managed Identity?
```
matthchr, line 140: I think we should have a meeting with @aramase about this maybe. He might have some good suggestions or raise up some gotchas we haven't thought of.

My initial thinking is something like the following:
1. User creates ManagedIdentity
2. User registers the managed identity Federated identity credential. When they do this, they need to specify the serviceaccount and namespace from which requests will originate. For ASO I think that's always going to be `system:serviceaccount:azureserviceoperator-system:azureserviceoperator-default`... which is a bit awkward because they don't necessarily even have permission to that namespace. We can document it though so it's maybe not the end of the world.
3. User creates a secret in their namespace with their MI details. They may also need to set a flag in that secret like `IsManagedIdentity` or something so that ASO understands this is a managedIdentity and not a SP. Or else we can just use the lack of `ClientSecret` field for that.
4. ASO reads that secret, gets the clientId of the identity it's supposed to use, and uses it somehow. We may need another operator that proxies that identity onto the ASO ServiceAccount, or maybe there's a trick we can use to avoid needing to do that... this is where talking with @aramase would help.

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
```diff
@@ -0,0 +1,153 @@
+---
+title: '2022-09: Support For Multiple Credentials Under Global Operator'
+---
+
+## Context
+
+The operator needs to support creating Azure resources using multiple credentials, managed by different groups, within one cluster by a single global instance of operator.
```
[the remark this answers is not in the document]
theunrepentantgeek: Suggest inverting the structure of the current intro to start with the problem being solved, rather than the change we're making to ASO.

Maybe 

> Some users of Azure Service Operator want to have a single instance that manages Azure resources across multiple subscriptions. This is useful, for example, if the organisation has a large cluster used by multiple teams, each team having their own Azure subscription. 
> To support this, ASO will introduce support for multiple credentials, each associated with a specific scope:
> * *Namespaced Secrets* will be used for all Azure Resources within a specific cluster namespace
> * *ResourceGroup Secrets* will be used for all Azure Resources within a specific Resource Group
> * *Resource Secrets* will be used for specific resources

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
```diff
@@ -0,0 +1,153 @@
...
+
+![hierarchy](images/adr-2022-09-multiple-credential-operator.png) 
+
+## Options
+
+In all the proposed solutions, we would need to cache the credentials to be used for resources when we reconcile and a mechanism to identify the credentials used for a resource. 
+
+### Option 1: Secrets lookup
+
+In secrets lookup option, we would need a fixed secret pattern like below. Which then will be looked up for each time a new resource is created. Operator would look up for the secret pattern for resource group first, then namespace. If secret exists, cache the credentials for the resource to be used for the next time or use the default global.
```
theunrepentantgeek, line 36: Grammar tip - generally want to avoid using "which" as a linking word like this.

```suggestion
In secrets lookup option, fixed naming patterns would be used to look up a secret each time a new resource is created. Operator would first use the naming pattern for the resource group, then for the cluster namespace. If a secret is found with the expected name, the credentials for the resource would be cached for use the next time. If no secret was found matching any patter, the default global secret would be used.
```

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
```diff
@@ -0,0 +1,153 @@
...
+
+NOTE: Annotation/condition will be added on the resource while creation which will have information about the secret name used for creation/reconcile.
+
+**Fixed secret name pattern:**
+```
+Namespaced secret -> aso-credential (in the namespace)
+Per-resource-group secret -> aso-credential-{rg-name}
+Per-resource secret -> aso-credential-{resource-name}
+```
```
theunrepentantgeek, line 45: Suggest using a table
```suggestion
| Secret Type        | Pattern                        | Notes                            |
| ------------------ | ------------------------------ | -------------------------------- |
| Namespaced         | aso-credential                 | Created in the Cluster namespace |
| Per-resource-group | aso-credential-{rg-name}       |                                  |
| Per-resource       | aso-credential-{resource-name} |                                  | 
```

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
```diff
@@ -0,0 +1,153 @@
...
+annotations: 
+  serviceoperator.azure.com/credential-from: any-secret
+```
+
+**Pros:**
+1. Flexible, as users have the control for which resource they want to use which credential
+
+**Cons:**
+1. Will have to specify annotation on each resource explicitly
```
theunrepentantgeek, line 66:
```suggestion
1. Will have to specify annotation on every resource explicitly, involving lots of repetition
```

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
```diff
@@ -0,0 +1,153 @@
...
+```
+Namespaced secret -> aso-credential (in the namespace)
+```
+
+**Annotation for per-resource-group and per-resource secret**
+```
+annotations: 
+  serviceoperator.azure.com/credential-from: any-secret
+```
+
```
theunrepentantgeek, line 84: Another case where a table would be clearer.

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
```diff
@@ -0,0 +1,153 @@
...
+**Pros:**
+1. Users don't have to specify the configuration for each resource explicitly if using a namespaced secret
+2. Users get the flexibility of over-riding the credential they want to use as per resource and as per resource group
+
+**Cons:**
+1. Could be complicated/confusing for users to understand the usage and credentials being used
+
+### Option 4: Secret lookup + configuration using global secret
+
+This option, relates to the option 3. However, the change is how we configure to use the per-resource-group and per-resource secret. We can have a unique slice property(CREDENTIAL_FROM) in the global secret where admin users can specify the secrets they want to load on the startup.
```
theunrepentantgeek, line 94: If this is a variant of Option 3, call it Option 3b.

What's a `slice property`? Is this a Go thing? You talk later about secrets (plural) so maybe this is a list property? or a multi-valued property?

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
theunrepentantgeek: Good idea.

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
```diff
@@ -0,0 +1,153 @@
...
+
+TBC
+
+## Experience Report
+
+TBC
+
+## References
+
+TBC
```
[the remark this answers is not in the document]
theunrepentantgeek: Do we need to use a persistent cache like that? In conversations so far, I've assumed that any caching would be in-memory, to avoid hitting the api-server any more often than needed.

If the ASO pod restarts, loading all the secrets afresh (not from the cache) is a feature, as it gives users a simple (if draconian) way to ensure that up to date secrets are being used.

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
super-harsh: Would help users to know which secret is being used for a resource, and for the operator in a manner that we don't have to do the secret lookup each time for namespaced secret and know from where to use the credential on a reconcile.

theunrepentantgeek: This brings up an interesting scenario - if what happens if a resource is created using a lower priority secret (e.g. a namespaced secret), but then the user creates a higher priority one (e.g. a resource group secret), should the operator continue to use the original one, or should it switch?

I can see migration scenarios where users have things set up one way and then create new secrets, expecting the operator to switch over. Does having annotations on the resource prevent this from working? How would the user do the migration?

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
```diff
@@ -0,0 +1,131 @@
...
+- Errors can be reported through the standard `Condition` mechanism
+
+## Milestones
+
+1. Namespaced secrets: TBD
+2. per-ResourceGroup secrets: TBD
+3. per-Resource secrets: Will not be implemented unless there is strong user demand. We believe that per-ResourceGroup secrets should be sufficiently granular.
+
+## Open questions
+1. Do we need to support both per-resource and per-resource-group credentials?
```
matthchr, line 115: I think you have actually answered this question now - we've designed for it so we _could_ support it, but not planned to implement unless user asks. To me, that means you can remove this from the list of open questions.

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
matthchr: I think it would make sense to switch the secret used at that point @theunrepentantgeek. As you say, there are scenarios where they do this on purpose.

I do see some possible awkwardness with the annotations though, it would be good exploring these options more in depth in the design to see if we can come up with solutions for them.

1. Moving from global to per-namespace, there was no annotation to begin with so no issues there.
2. Moving from per-namespace to per-resourceGroup: There would be an automatically added annotation pointing to the per-namespace secret on all of the existing resources. If the user then changes the annotation on the ResourceGroup to point to a different secret, that does not change the an annotation on any of the other resources in that resourceGroup, so we have an awkward situation where RG points at secret A, but all other resources point at secret B. Who wins?

We could possibly solve that by, next time the child resources are reconciled, fixing the annotation to point to the same secret as the resourceGroup, but that seems awkward because now things aren't so simple as "the secret used will always be what is set in `serviceoperator.azure.com/credential-from`". Another problem with reusing the `serviceoperator.azure.com/credential-from` for us setting it and the user also having the option to set it, is that we need to watch that annotation to trigger reconciles, but if we can also set it ourselves in a reconcile it's possible that a reconcile runs, changes that value and then triggers another reconcile based off of that. So I've tried to keep annotations in a spec/status style, where it's either something we write and the user _never_ writes, or it's something the user writes and we _never_ write. We can even write those as long as it's in a webhook, (won't trigger extra reconciles), but I don't see a way to deal with this case really cleanly off the top of my head.

Is it clearer to _not_ annotate any resources other than ResourceGroups with the `serviceoperator.azure.com/credential-from` annotation automatically? So basically in our initial implementation that annotation would never be on anything other than `ResourceGroup` (and possibly other top-level resources we designate)?

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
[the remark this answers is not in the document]
super-harsh: Using @theunrepentantgeek's suggestion here. FYI @matthchr

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
super-harsh: I think we SHOULD not suggest shifting credentials? Do we want to allow users to change the credential setting once the resource is created?

Should suggest re-creating the resource with new credentials rather. If user tries to change credential on already present resource, we should set condition to false.

theunrepentantgeek: > I think we SHOULD not suggest shifting credentials?

Sorry, I don't understand this phrasing.
Are you saying we shouldn't comment on shifting credentials (our docs would be silent on the subject) or that we should document that shifting credentials is unsupported?

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
theunrepentantgeek: Yup. The usual sections are

- Context
- Decision
- Status
- Consequences
- Experience Report
- References

For a design like this, most of the content goes in _Context_, with some detail in _Decision_, and _Status_ as pending.

We then loop back after implementation to fill in the later sections.

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
super-harsh: > This brings up an interesting scenario - if what happens if a resource is created using a lower priority secret (e.g. a namespaced secret), but then the user creates a higher priority one (e.g. a resource group secret), should the operator continue to use the original one, or should it switch?

As on the reply to this, Operator would use whatever is present on annotation. The idea of adding annotation is that, we're not storing any state of which credential is being used for which resource. If annotation says `credential-from: secretA` then operator on reconcile, would simple fetch the credential for secretA(if exist) or run into failure if does not exist. The idea of adding annotation for namespaced secret is to eliminate the need of repetitive lookup and storing state.

super-harsh: > > I think we SHOULD not suggest shifting credentials?
> 
> Sorry, I don't understand this phrasing. Are you saying we shouldn't comment on shifting credentials (our docs would be silent on the subject) or that we should document that shifting credentials is unsupported?

Sorry for being unclear there, we should document that it's not suggested, as it can lead to failures on reconcile. As the new credentials might not have permissions to update the resource.

matthchr: > Sorry for being unclear there, we should document that it's not suggested, as it can lead to failures on reconcile. As the new credentials might not have permissions to update the resource.

I think we should consider the "I want to change credentials used to reconcile this object" case. This is something that you can change pretty easily if you're say, running your own code or using something like the `az cli`. It seems to me that we should support it. If users attempt to change to a credential that doesn't have permissions to do what they wanna do, then that's their issue. It seems weird to block or not recommend that because they might make a mistake though. 

Any issue they could run into with changing credentials they could also run into on first-use (the credential they try to use is wrong). So we're going to need to make sure we give a good error in those cases either way.

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
matthchr: Sounds like there's a path forward for this. I will write up the steps and you can incorporate it into this document @super-harsh.

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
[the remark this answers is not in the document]
matthchr: Going to resolve this, as it looks like you've taken this feedback.

docs/hugo/content/design/ADR-2022-09-Multiple-Credential-Operator.md
matthchr: Ok so to answer this question, here's what we need to do:

1. ASO must be installed with workload identity. This is because we need access to the Service Account Token, which will only be injected and used if they configure that for the global credential.
2. Users create a `UserManagedIdentity` which they want to use as the identity in a particular namespace. They assign permissions to this identity. This can be done either using ASO itself, or 
3. Users register a `FederatedIdentityCredential` for that `ManagedIdentity`. The `FederatedIdentityCredential` must be registered for the ASO system service account (`"subject": "system:serviceaccount:azureserviceoperator-system:azureserviceoperator-default"`). Note that the users don't need access to our namespace to do this, that subject is basically just a magic string they need to use.
4. User would then create the namespace-scoped `aso-credential` Secret, containing the `clientId` of the User Managed Identity. ASO will read and use that `clientId` when authenticating with Azure.

theunrepentantgeek (approved):

matthchr (approved):

PR #2500, opened 2022-09-16
Refactor secrets handling to prepare for configmap support

matthchr:
Refactors some code which can be shared between configmap and secrets.

**If applicable**:
- [ ] this PR contains tests


.devcontainer/install-dependencies.sh
```diff
@@ -160,14 +160,12 @@ go-install gen-crd-api-reference-docs github.com/ahmetb/gen-crd-api-reference-do
 go-install setup-envtest sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
 
 # Install golangci-lint
-if [ "$DEVCONTAINER" != true ]; then
```
theunrepentantgeek, line 163: Why remove the `DEVCONTAINER` check? I thought we could/should skip installation in the devcontainer.

theunrepentantgeek (approved):

.devcontainer/install-dependencies.sh
matthchr: As far as I can tell, the devcontainer already has golangci-lint installed. _But_ it's a version we don't control as it's part of the base image, which means that we don't control its version and it's hard to match locally if you're not running the container.

For example you could pretty easily get into a situation where the devcontainer was on version A, but when you run install-dependencies.sh you get version B, and you get different linter errors because of that. It seemed to make more sense to just be consistent between the two, that way no matter what env you're using (devcontainer vs normal machine that just ran `install-dependencies`) you see the same lint errors.

Also gives us more control over the linter version so I think we'll stop seeing things randomly break because the linter catches new stuff after an update. We will need to remember to update the version used from time to time, but that's probably better as it's less disruptive to us (we opt-in to updates as opposed to getting them even if we don't want)

PR #2502, opened 2022-09-16
Use Swagger as first source of truth for ARM references

### LINKED ISSUE #2473: Use arm-id annotation in Swagger instead of $armReference in azure-arm.yaml
See [formatToPattern](https://github.com/Azure/azure-service-operator/blob/d5c322fde392e27b39ab540df7a293120f195f14/v2/tools/generator/internal/jsonast/jsonast.go#L332) for where we are ignoring this today.

matthchr:
Closes #2473

gif

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/v2/tools/generator/internal/codegen/pipeline/add_cross_resource_references.go
+++ b/v2/tools/generator/internal/codegen/pipeline/add_cross_resource_references.go
@@ -29,12 +29,34 @@ func AddCrossResourceReferences(configuration *config.Configuration, idFactory a
 		AddCrossResourceReferencesStageID,
 		"Replace cross-resource references with genruntime.ResourceReference",
 		func(ctx context.Context, definitions astmodel.TypeDefinitionSet) (astmodel.TypeDefinitionSet, error) {
-			result := make(astmodel.TypeDefinitionSet)
+			typesWithARMIDs := make(astmodel.TypeDefinitionSet)
 
 			var crossResourceReferenceErrs []error
 
 			isCrossResourceReference := func(typeName astmodel.TypeName, prop *astmodel.PropertyDefinition) bool {
+				// First check if we know that this property is an ARMID already
 				isReference, err := configuration.ARMReference(typeName, prop.PropertyName())
+				if primitive, ok := astmodel.AsPrimitiveType(prop.PropertyType()); ok {
+					if primitive == astmodel.ARMIDType {
+						if err == nil {
+							if !isReference {
+								// We've overridden the ARM spec details
+								return false
+							} else {
+								// Swagger has marked this field as a reference, and we also have it marked in our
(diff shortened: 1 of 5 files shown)


v2/tools/generator/internal/codegen/pipeline/add_cross_resource_references.go
```diff
@@ -29,12 +29,31 @@ func AddCrossResourceReferences(configuration *config.Configuration, idFactory a
...
+						}
+						// Mark configuration as unconsumed here to force an error since
+						// Swagger is overriding it so we don't need the annotation anymore
+						err = configuration.MarkARMReferenceUnconsumed(typeName, prop.PropertyName())
+						if err != nil && !config.IsNotConfiguredError(err) {
+							crossResourceReferenceErrs = append(
+								crossResourceReferenceErrs,
+								errors.Wrap(err, "failed to reset ARM reference flag"))
+						}
+						return true
```
theunrepentantgeek, line 53: Instead of marking the configuration as unconsumed, maybe it would be better to return errors from the stage indicating that the configuration of `$armReference: true` is no longer required.

After all, the value _has_ been consumed - it's just been found to be redundant.

v2/tools/generator/internal/codegen/pipeline/add_cross_resource_references.go
```diff
@@ -211,3 +235,27 @@ func makeResourceReferenceProperty(idFactory astmodel.IdentifierFactory, existin
 
 	return newProp
 }
+
+func stripRemainingARMIDPrimitiveTypes(types astmodel.TypeDefinitionSet) (astmodel.TypeDefinitionSet, error) {
+	// Any astmodel.ARMReference's which have escaped need to be turned into
+	// strings
+	armIDStrippingVisitor := astmodel.TypeVisitorBuilder{
+		VisitPrimitive: func(_ *astmodel.TypeVisitor, it *astmodel.PrimitiveType, ctx interface{}) (astmodel.Type, error) {
+			if it == astmodel.ARMIDType {
```
theunrepentantgeek, line 244: This is probably safe, but other type comparisons use the `TypeEquals()` method, so this should as well.

theunrepentantgeek (approved):

v2/tools/generator/internal/codegen/pipeline/add_cross_resource_references.go
matthchr: Yeah that's also less code because we don't need to plumb the Unconsumed function through the config. Changed.

v2/tools/generator/internal/codegen/pipeline/add_cross_resource_references.go
matthchr: Fixed

PR #2510, opened 2022-09-20
Use KubernetesExporter interface rather than AzureSecrets extension

matthchr:
KubernetesExporter is more generic so scales better to ConfigMaps and possibly other Kubernetes resource types in the future

gif

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -327,7 +327,10 @@ tasks:
     deps: [controller:run-kustomize-for-envtest]
     cmds:
     # -race fails at the moment in controller-runtime
+    # We run these in sequence right now because it doesn't gain us much speed to go in parallel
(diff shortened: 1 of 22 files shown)


theunrepentantgeek (approved):
Overall looks really good - nicely taking things to be more generic.

Should `.../genruntime/test/recordings/Test_ApplyObjAndEnsureOwner.yaml` be empty though?

matthchr:
> Should .../genruntime/test/recordings/Test_ApplyObjAndEnsureOwner.yaml be empty though?

So, it uses the envtest infrastructure in order to ensure that the Apply codepath works, but because that codepath produces no calls to ARM, it records no HTTP content. We have a few other tests that also do this. I could've put in a mode into testcommon that indicated I just want envtest but without the recording infrastructure, but that would make it more complicated than it was so it felt better to live with the artifact of an empty recording file.

super-harsh (approved):

PR #2512, opened 2022-09-21
Generate reports showing resource structure to allow change detection

theunrepentantgeek:

As we work towards migration to Swagger as a single source of truth (see #2323), we face the issue of encountering unexpected breaking changes.

This PR adds a pipeline stage to create a structure report for each package, showing the shape of each resource in that package. Nested types are inlined just as they would be in the CRDs, and our custom type names are largely elided.

I considered checking in the raw CRDs, but the changes to documentation make those less suitable for detecting structure changes.

gif


v2/api/appconfiguration/customizations/structure.txt
```diff
@@ -0,0 +1,2 @@
+github.com/Azure/azure-service-operator/v2/api/appconfiguration/customizations
```
super-harsh, line 1: Would we need structure for `customizations` folder?

v2/tools/generator/internal/reporting/type_catalog_report.go
```diff
@@ -0,0 +1,296 @@
...
+		if r := recover(); r != nil {
+			os.Remove(filePath)
+			panic(r)
+		}
+	}()
+
+	err = tcr.WriteTo(file)
+	if err != nil {
+		// cleanup in case of errors
+		file.Close()
```
super-harsh, line 53: Do we need this block here? We'll handle this in defer function already?

v2/api/appconfiguration/customizations/structure.txt
theunrepentantgeek: A report is generated for each output package, including the `customizations` folder, though that one is trivially small because the object we generate is, well, trivially small. I'm not sure we need it, but it'd be more code to suppress it and I don't see any harm.

v2/tools/generator/internal/reporting/type_catalog_report.go
theunrepentantgeek: I think we do - the defer block will only remove the file if we're in mid-panic, whereas we always want to remove it if there was a problem.

super-harsh (approved):

v2/tools/generator/internal/reporting/type_catalog_report.go
super-harsh: Correct me if I'm wrong - If we close here, and again close in defer block(before exitting), it'll panic anyway(as we're closing a closed file) and never return the error?

theunrepentantgeek: The second call to `Close()` returns an error (which is ignored anyway); [it doesn't panic](https://pkg.go.dev/os#File.Close).

FWIW, I copied the structure from `go_source_file.go` (in our `astmodel` package), so I hope it's not a broken technique!

v2/internal/controllers/controller_resources_gen.go
```diff
@@ -227,27 +227,15 @@ func getKnownStorageTypes() []*registration.StorageType {
...
-			{
-				Key:  ".spec.initContainers.environmentVariables.secureValue",
-				Func: indexContainerinstanceContainerGroupInitContainersSecureValue,
-			},
 			{
 				Key:  ".spec.imageRegistryCredentials.password",
 				Func: indexContainerinstanceContainerGroupPassword,
 			},
-			{
-				Key:  ".spec.diagnostics.logAnalytics.workspaceKey",
```
matthchr, line 243: This feels like something that should be a secret. Did it get deleted in a more recent API or something?

v2/api/containerinstance/v1beta20211001/container_group_types_gen.go
```diff
@@ -7783,6 +7548,190 @@ const (
...
+	// Metadata: Metadata for log analytics.
+	Metadata map[string]string `json:"metadata,omitempty"`
+
+	// +kubebuilder:validation:Required
+	// WorkspaceId: The workspace id for log analytics
+	WorkspaceId *string `json:"workspaceId,omitempty"`
+
+	// +kubebuilder:validation:Required
+	// WorkspaceKey: The workspace key for log analytics
+	WorkspaceKey *string `json:"workspaceKey,omitempty"`
```
matthchr, line 7565: I think this should be a secret

v2/api/containerinstance/v1beta20211001/container_group_types_gen.go
```diff
@@ -7149,10 +6782,130 @@ func (view *ContainerProperties_InstanceView_STATUS) AssignProperties_To_Contain
...
 }
 
+// Generated from: https://schema.management.azure.com/schemas/2021-10-01/Microsoft.ContainerInstance.json#/definitions/EnvironmentVariable
+type EnvironmentVariable struct {
+	// +kubebuilder:validation:Required
+	// Name: The name of the environment variable.
+	Name *string `json:"name,omitempty"`
+
+	// SecureValue: The value of the secure environment variable.
+	SecureValue *string `json:"secureValue,omitempty"`
```
matthchr, line 6792: I think this should be a secret ref, like it was in the `ContainerGroup_Properties_Containers_Properties_EnvironmentVariables_Spec` type.

v2/internal/controllers/controller_resources_gen.go
matthchr: Maybe separate out the structure.txt stuff from the submodule update, and we can do the submodule update in a separate PR?

v2/tools/generator/internal/reporting/structure_report.go
```diff
@@ -16,32 +17,32 @@ const (
...
+func (dr *StructureReport) SaveTo(writer io.Writer) error {
 	var indents []string
 
 	return dr.writeBlock(writer, indents, "", "")
 }
 
-// writeTo writes a block of lines from this debugReport to a writer
-func (dr *debugReport) writeBlock(
+// writeTo writes a block of lines from this StructureReport to a writer
+func (dr *StructureReport) writeBlock(
```
matthchr, line 45: minor: `sr` would make more sense than `dr` as the receiver name

matthchr (changes requested):
structure.txt stuff looks good, but there are some issues with the submodule update + subsequent changes to generated code.

v2/api/containerinstance/v1beta20211001/container_group_types_gen.go
matthchr: Hmm, it's flagged as `x-ms-secret` in the Swagger too... see [here](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/containerinstance/resource-manager/Microsoft.ContainerInstance/stable/2021-10-01/containerInstance.json#L1560)

```
        "workspaceKey": {
          "description": "The workspace key for log analytics",
          "type": "string",
          "x-ms-secret": true
        },
```

v2/api/containerinstance/v1beta20211001/container_group_types_gen.go
matthchr: This is _also_ flagged as `x-ms-secret` in the Swagger, [see](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/containerinstance/resource-manager/Microsoft.ContainerInstance/stable/2021-10-01/containerInstance.json#L1737)

So for some reason neither this nor the above workspace key are getting correctly set after the submodule update. Possibly because we've just updated the Azure REST API Specs (Swagger) submodule here but not the Azure Resource Manager Schema too? I think we might need to update them both (in sync). These types seem to have been renamed because of a change in the Swagger, but if we don't also update Resource Manager Schema then it doesn't have the new names so the type-merging doesn't work? (Just thinking out loud here but it seems plausible to me)

matthchr (approved):

v2/tools/generator/internal/reporting/structure_report.go
theunrepentantgeek: Fixed.

v2/internal/controllers/controller_resources_gen.go
theunrepentantgeek: Fixed submodule reference and this change went away.

v2/api/containerinstance/v1beta20211001/container_group_types_gen.go
theunrepentantgeek: Fixed submodule reference and this change went away.

PR #2516, opened 2022-09-22
Register RP if not registered in RP

### LINKED ISSUE #2513: RPs are not registered for first use
A description of how things are today.

For the new subscription, RPs are not registered by default when used via REST for the first time. As a result, provisioning of the resources will fail until it is manually registered and resources are stuck/time out in reconcile error.

```
"error"="The subscription is not registered to use namespace 'Microsoft.DocumentDB'
```

**Describe the improvement**

Check on RP registration HTTP 409 "MissingSubscriptionRegistration" error and retry the operation with exponential back off

**Additional context**

```
{
    "code": "MissingSubscriptionRegistration",
    "message": "The subscription is not registered to use namespace 'Microsoft.DocumentDB'. See https://aka.ms/rps-not-found for how to register subscriptions.",
    "details": [
      {
        "code": "MissingSubscriptionRegistration",
        "target": "Microsoft.DocumentDB",
        "message": "The subscription is not registered to use namespace 'Microsoft.DocumentDB'. See https://aka.ms/rps-not-found for how to register subscriptions."
      }
    ]
  }
}
```

matthchr:
This closes #2513.

**If applicable**:
- [x] this PR contains tests


v2/internal/genericarmclient/policy_register_rp.go
```diff
@@ -0,0 +1,221 @@
...
+	azpolicy "github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+
+	"github.com/Azure/azure-service-operator/v2/internal/version"
+)
+
+// NewRPRegistrationPolicy creates a policy object configured using the specified options.
+// The policy controls whether an unregistered resource provider should automatically be
+// registered. See https://aka.ms/rps-not-found for more information.
+func NewRPRegistrationPolicy(cred azcore.TokenCredential, o *azpolicy.ClientOptions) (azpolicy.Policy, error) {
```
super-harsh, line 34: rename `o *azpolicy.ClientOptions` to `options *azpolicy.ClientOptions` ?

v2/internal/genericarmclient/policy_register_rp.go
```diff
@@ -0,0 +1,221 @@
...
+	armruntime "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
+	azpolicy "github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+
+	"github.com/Azure/azure-service-operator/v2/internal/version"
+)
+
+// NewRPRegistrationPolicy creates a policy object configured using the specified options.
+// The policy controls whether an unregistered resource provider should automatically be
```
super-harsh, line 32: Do we want to automatically register the RP? Thinking if it would be a right thing to do without the user's permission?

v2/internal/genericarmclient/policy_register_rp.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+	if _, err = rpOps.Register(req.Raw().Context(), rp); err != nil {
+		return resp, err
+	}
+	return resp, fmt.Errorf("registering Resource Provider %s with subscription. Try again later", rp)
+}
+
+func getSubscription(path string) (string, error) {
+	parts := strings.Split(path, "/")
+	for i, v := range parts {
```
super-harsh, line 101: Can use regex matcher here?

v2/internal/genericarmclient/policy_register_rp.go
matthchr: So it's pretty common to automatically register subscriptions on the users behalf. The Go SDK (and AFAIK all of the other SDKs) actually do it by default. We turn it off on our usage of the Go SDK because it also by default waits 5m for the registration to take place, which blocks the running reconcilers for 5m which is bad.

There's not really a lot of risk/harm in registering an RP AFAIK. We also have a user requesting it in #2513, and not registering sorta defeats some of the IAC purpose of ASO because you have to perform manual actions to initially set up your sub. In a perfect world ASO would take care of that for you.

v2/internal/genericarmclient/policy_register_rp.go
matthchr: Could, but since I copied this code from the Go SDK I think it makes sense to (as much as possible) keep it similar so that we can easily apply changes if we want.

I copied their code rather than using their code because theirs forces retries and a wait on you (or else you get an ugly timeout error), and I wanted our users to see a better error than that.

super-harsh (approved):

v2/internal/genericarmclient/policy_register_rp.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+type rpRegistrationPolicy struct {
+	endpoint string
+	pipeline runtime.Pipeline
+	options  *azpolicy.ClientOptions
+}
+
+func (r *rpRegistrationPolicy) Do(req *azpolicy.Request) (*http.Response, error) {
+	const unregisteredRPCode = "MissingSubscriptionRegistration"
```
theunrepentantgeek, line 55: Suggest adding a comment to indicate what this constant is for
```suggestion
        // The unique identifier returned by ARM when the required resource provider has not been registered
	const unregisteredRPCode = "MissingSubscriptionRegistration"
```

v2/internal/genericarmclient/policy_register_rp.go
theunrepentantgeek: I'm not sure it would be easy to match with a Go regex anyway - isn't the limitation on prior-matches one of the issues?

theunrepentantgeek (approved):

v2/internal/genericarmclient/policy_register_rp.go
matthchr: Same comment about keeping code mostly the same as its original source (for easier keeping up to date).

PR #2519, opened 2022-09-27
Change MySQL livetest location to Eastus

super-harsh:

Looked like capacity issues to me, however the error was `Subscription 'sub' is not allowed to provision in West US`


matthchr:
> Subscription 'sub' is not allowed to provision in West US

Where did you see that error? We should make sure it's more clear in the test logs. All I think I saw was InternalServerError

matthchr (approved):

Issue #2435, opened 2022-08-05
Feature: Reference Managed Identities in Role Assignments


thriqon:
Similarly to Public IP Addresses in Load Balancers, it should be possible to reference a Managed Identity by name in a Role Assignment.

Consider the following:

```yaml
---
apiVersion: managedidentity.azure.com/v1beta20181130
kind: UserAssignedIdentity
metadata:
  name: sample-uai
spec:
  location: Germany West Central
  owner:
    name: dev-sample-rg
---
apiVersion: authorization.azure.com/v1beta20200801preview
kind: RoleAssignment
metadata:
  name: 6a2d44f5-57d8-4916-9f46-ff7c9c1b338f
spec:
  location: Germany West Central
  owner:
    name: samplevnet
    group: network.azure.com
    kind: VirtualNetwork
  principalId: <NOT YET KNOWN>
  roleDefinitionReference:
    armId: /subscriptions/00000000-0000-0000-0000-000000000000/providers/Microsoft.Authorization/roleDefinitions/b24988ac-6180-42a0-ab88-20f7382dd24c
```

This cannot be applied directly to the cluster, as the `principalId` of the Managed Identity is yet unknown. After creation, it has to be copied manually (!) and applied afterwards.

There is prior art, as a Load Balancer accepts a Kubernetes reference of the form

```yaml
reference:
  name: ...
  kind: ...
  group: ...
```

The `principalId` is part of the `status` subresource.

matthchr:
We totally agree. It's possible we could use something like #2350 for this, or alternative we could do like you suggest and have special handling like we do for ARM IDs.

super-harsh:
Related to #2056

Issue #2317, opened 2022-05-25
Feature: Export clientId from UserAssignedIdentity


Sbou:
Hello,

To be able to use the Pod identity feature in a AKS clsuter, it could be great to have the **resource Id** and the **client Id** of the generated managed identity in a ConfigMap or Secret.

Thanks

theunrepentantgeek:
This may be possible by adding `$azureGeneratedSecrets`. We'll add this to our backlog - given it's a simple change, we'd also accept a PR.

matthchr:
We were just talking about this in #2489 and #2435. This is something that we want to do.

PR #2520, opened 2022-09-27
Add ConfigMapReference and friends

matthchr:
Part of ongoing work related to #2317 and #2435.

Also removes SecretsRetriever which isn't used anymore

**If applicable**:
- [x] this PR contains tests


v2/pkg/genruntime/configmaps.go
```diff
@@ -0,0 +1,161 @@
...
+func (c ConfigMapReference) Copy() ConfigMapReference {
+	return c
+}
+
+func (c ConfigMapReference) String() string {
+	return fmt.Sprintf("Name: %q, Key: %q", c.Name, c.Key)
+}
+
+// ToNamespacedRef creates a NamespacedSecretReference from this SecretReference in the given namespace
+func (c ConfigMapReference) ToNamespacedRef(namespace string) NamespacedConfigMapReference {
```
theunrepentantgeek, line 42: Minor: I think we've been using the prefix `As` on these sorts of conversion methods elsewhere in ASO.

v2/pkg/genruntime/configmaps.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// NamespacedConfigMapReference is a ConfigMapReference with namespace information included
+type NamespacedConfigMapReference struct {
+	ConfigMapReference
+	Namespace string
+}
+
+func (s NamespacedConfigMapReference) String() string {
+	return fmt.Sprintf("Namespace: %q, %s", s.Namespace, s.ConfigMapReference.String())
```
theunrepentantgeek, line 56: Do you need the call to `String()` here? I thought `fmt.Sprintf()` would automagically use the `fmt.Stringer` interface if it's present.

v2/pkg/genruntime/configmaps.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+type OptionalConfigMapReference struct {
```
theunrepentantgeek, line 59: Type needs a comment.

v2/pkg/genruntime/configmaps.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func (ref *OptionalConfigMapReference) UnmarshalJSON(data []byte) error {
+	var value string
+	strErr := json.Unmarshal(data, &value)
+	if strErr == nil {
+		// Format was string, set and return
+		ref.Value = &value
+		return nil
+	}
```
theunrepentantgeek, line 72: Nice!

v2/pkg/genruntime/configmaps.go
```diff
@@ -0,0 +1,161 @@
...
+	Name string `json:"name"`
+
+	// Key is the key in the ConfigMap being referenced
+	// +kubebuilder:validation:Required
+	Key string `json:"key"`
+
+	// This is a type separate from ConfigMapReference as in the future we may want to support things like
+	// customizable annotations or labels, instructions to not delete the ConfigMap when the resource is
+	// deleted, etc. None of those things make sense for ConfigMapReference so using the exact same type isn't
+	// advisable.
```
theunrepentantgeek, line 123: Also nice. #goodtoknow

v2/pkg/genruntime/configmaps.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func makeKeyPairFromConfigMap(dest *ConfigMapDestination) keyPair {
+	return keyPair{
+		name: dest.Name,
+		key:  dest.Key,
+	}
+}
+
+// ValidateConfigMapDestinations checks that no destination is writing to the same configmap/key, as that could cause
```
theunrepentantgeek, line 142:
```suggestion
// ValidateConfigMapDestinations checks that no two destinations are writing to the same configmap/key, as that could cause
```

v2/pkg/genruntime/configmaps/configmaps.go
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package configmaps
```
theunrepentantgeek, line 6: Why have a separate nested package like this, especially one that essentially contains just a single helper function?

v2/pkg/genruntime/configmaps_test.go
```diff
@@ -0,0 +1,141 @@
...
+			g.Expect(err).ToNot(HaveOccurred())
+			g.Expect(string(result)).To(Equal(tt.expected))
+		})
+	}
+}
+
+func Test_DeserializeOptionalConfigMapReference(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
```
theunrepentantgeek, line 71: Elsewhere we've called this `cases` and used `c` as the local variable.

theunrepentantgeek (approved):
A few small changes, but overall looks very good.

gif

v2/pkg/genruntime/configmaps/configmaps.go
matthchr: It's going to get more stuff in future PRs. For example `configmaps.Collector` (similar to `secrets.Collector` we already have).

Since `genruntime` is already pretty big, I'm trying to (where it makes sense) make subpackages. Some stuff still has to go at the top level so it's not like genruntime is going to stop growing, but trying to combat bloat some by putting things like `configmaps.Collector`, `configmaps.SliceToClientObjectSlice`, `secrets.Collector`, etc into their own sub packages. It also feels more idiomatic in Go than having `genruntime.ConfigMapCollector` and `genruntime.SecretCollector`.

v2/pkg/genruntime/configmaps.go
matthchr: Fixed

v2/pkg/genruntime/configmaps.go
matthchr: Fixed

v2/pkg/genruntime/configmaps.go
matthchr: Fixed (and in secrets too)

v2/pkg/genruntime/configmaps_test.go
matthchr: Fixed

v2/pkg/genruntime/configmaps.go
matthchr: `ResourceReference`, and `SecretReference` are already using the `To` prefix here, which was the pattern I was following for this type.

I went ahead and changed them all though as part of this PR to the `AsX` pattern.

v2/pkg/genruntime/configmaps/configmaps.go
theunrepentantgeek: Fair call. :+1:

super-harsh (approved):

PR #2523, opened 2022-09-29
Add configmap Collector and generic Merger

matthchr:
**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/api/dbformariadb/customizations/server_extensions.go
+++ b/v2/api/dbformariadb/customizations/server_extensions.go
@@ -74,8 +74,8 @@ func secretsToWrite(obj *mariadb.Server) ([]*v1.Secret, error) {
 		return nil, errors.Errorf("unexpected nil OperatorSpec")
 	}
 
-	collector := secrets.NewSecretCollector(obj.Namespace)
-	collector.AddSecretValue(operatorSpecSecrets.FullyQualifiedDomainName, to.String(obj.Status.FullyQualifiedDomainName))
+	collector := secrets.NewCollector(obj.Namespace)
+	collector.AddValue(operatorSpecSecrets.FullyQualifiedDomainName, to.String(obj.Status.FullyQualifiedDomainName))
 
-	return collector.Secrets(), nil
+	return collector.Values()
 }
--- a/v2/pkg/genruntime/configmaps/collector.go
+++ b/v2/pkg/genruntime/configmaps/collector.go
@@ -1,0 +1,120 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package configmaps
+
+import (
+	"sort"
+
+	"github.com/pkg/errors"
+	"golang.org/x/exp/maps"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+)
+
+// TODO: This is similar to secrets.Collector. If this is updated, that should likely be as well
+
+// Collector collects configmap values and their associated genruntime.ConfigMapDestination's
+// and produces a merged set of v1.ConfigMap's that can be written.
+type Collector struct {
+	configs   map[string]*v1.ConfigMap
+	namespace string
+	errors    []error
+}
+
+// NewCollector creates a new Collector for collecting multiple config map writes together
+func NewCollector(namespace string) *Collector {
+	return &Collector{
+		configs:   make(map[string]*v1.ConfigMap),
+		namespace: namespace,
+	}
+}
+
+func (c *Collector) get(dest *genruntime.ConfigMapDestination) *v1.ConfigMap {
+	existing, ok := c.configs[dest.Name]
+	if !ok {
+		existing = &v1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      dest.Name,
+				Namespace: c.namespace,
+			},
+			Data:       make(map[string]string),
+			BinaryData: make(map[string][]byte),
+		}
+		c.configs[dest.Name] = existing
+	}
+	return existing
+}
+
+func (c *Collector) errIfKeyExists(val *v1.ConfigMap, key string) error {
+	if _, ok := val.Data[key]; ok {
+		return errors.Errorf("key collision, entry exists for key %s in Data", key)
+	}
+
+	if _, ok := val.BinaryData[key]; ok {
+		return errors.Errorf("key collision, entry exists for key %s in BinaryData", key)
+	}
+
+	return nil
+}
+
+// AddValue adds the dest and ConfigMapDestination pair to the collector. If another value has already
+// been added going to the same config map (but with a different key) the new key is merged into the
+// existing map.
+func (c *Collector) AddValue(dest *genruntime.ConfigMapDestination, value string) {
+	if dest == nil || value == "" {
+		return
+	}
+
+	existing := c.get(dest)
+	err := c.errIfKeyExists(existing, dest.Key)
+	if err != nil {
+		c.errors = append(c.errors, err)
+		return
+	}
+
+	existing.Data[dest.Key] = value
+}
+
+// AddBinaryValue adds the dest and ConfigMapDestination pair to the collector. If another value has already
+// been added going to the same config map (but with a different key) the new key is merged into the
+// existing map.
+func (c *Collector) AddBinaryValue(dest *genruntime.ConfigMapDestination, value []byte) {
+	if dest == nil || value == nil {
+		return
+	}
+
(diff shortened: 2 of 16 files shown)


v2/api/dbformariadb/customizations/server_extensions.go
```diff
@@ -74,8 +74,8 @@ func secretsToWrite(obj *mariadb.Server) ([]*v1.Secret, error) {
 		return nil, errors.Errorf("unexpected nil OperatorSpec")
 	}
 
-	collector := secrets.NewSecretCollector(obj.Namespace)
-	collector.AddSecretValue(operatorSpecSecrets.FullyQualifiedDomainName, to.String(obj.Status.FullyQualifiedDomainName))
+	collector := secrets.NewCollector(obj.Namespace)
```
theunrepentantgeek, line 77: I was confused for a moment, but now I get it - saying `secrets.NewCollector()` is definitely nicer than having to stutter when saying `secretes.NewSecreteCollector()`.

v2/pkg/genruntime/configmaps/collector.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	existing.Data[dest.Key] = value
+}
+
+// Values returns the set of Values that have been collected.
+func (c *Collector) Values() ([]*v1.ConfigMap, error) {
+	err := kerrors.NewAggregate(c.errors)
+	if err != nil {
+		return nil, err
+	}
```
theunrepentantgeek, line 70: I'm still on the fence with this idiom - would prefer if the check was more like this

```suggestion
	if len(c.errors) > 0 {
		return nil, kerrors.NewAggregate(c.errors)
	}
```

v2/pkg/genruntime/merger/merger.go
```diff
@@ -0,0 +1,90 @@
...
+
+	result := append(other, secrets.SliceToClientObjectSlice(mergedSecrets)...)
+	result = append(result, configmaps.SliceToClientObjectSlice(mergedConfigMaps)...)
+	return result, nil
+}
+
+func mergeSecrets(namespace string, s []*v1.Secret) ([]*v1.Secret, error) {
+	collector := secrets.NewCollector(namespace)
+	for _, secret := range s {
+		// TODO: We're not merging Data here, which is possibly problematic?
```
theunrepentantgeek, line 63: Should you log a warning if this happens, so we can at least detect the problem?

v2/pkg/genruntime/merger/merger.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+	}
+
+	return collector.Values()
+}
+
+func mergeConfigMaps(namespace string, c []*v1.ConfigMap) ([]*v1.ConfigMap, error) {
+	collector := configmaps.NewCollector(namespace)
+	for _, configMap := range c {
+		// TODO: We're not merging BinaryData here, which is possibly problematic?
```
theunrepentantgeek, line 79: Again, should we log something so the error is detectable?

theunrepentantgeek (approved):

v2/pkg/genruntime/configmaps/collector.go
```diff
@@ -0,0 +1,83 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+)
+
+// TODO: This is similar to secrets.Collector. If this is updated, that should likely be as well
+
+// Collector collects secret values and their associated genruntime.ConfigMapDestination's
+// and produces a merged set of v1.Secret's that can be written.
```
super-harsh, line 23:
```suggestion
// Collector collects configmap values and their associated genruntime.ConfigMapDestination's
// and produces a merged set of v1.ConfigMap's that can be written.
```

v2/pkg/genruntime/secrets/collector.go
```diff
@@ -8,33 +8,38 @@ package secrets
...
 	"golang.org/x/exp/maps"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
 )
 
-// SecretCollector collects secret values and their associated genruntime.SecretDestination's
+// TODO: This is similar to configmaps.Collector. If this is updated, that should likely be as well
```
super-harsh, line 20: maybe we should consider extract `Collector` interface and extend it here and for configmaps?

super-harsh: Not sure if something like type `Collector[T v1.Secret|v1.ConfigMap] struct {}` would work? Kinda new to generics, but would be clean if we can use generics here.

v2/api/dbformariadb/customizations/server_extensions.go
matthchr: Yeah I think this is a somewhat standard Go idiom, where you use the package name to help keep things short. @babbageclunk also did it and drew a double-take from me with `config.Values` (which is just `Values` when unqualified). This shortening drew inspiration from that.

v2/pkg/genruntime/secrets/collector.go
matthchr: I thought about using generics here, but the whole `StringData`/`BinaryData` and how secret writes to `StringData` show up in data as base64 encoded, etc means the two have different application functions, so the type ends up needing to be something like:

```
type Collector[T metav1.Object] struct {
	things  map[string]*T
	namespace string
	applierFunc func(t T) error
}
```

or something like that.

We could do it, and it would save a small bit of code mostly the `Values` implementation, but the actual meat of `AddValue` couldn't be shared, so I decided against it at least for now. If we get a 3rd impl then maybe it makes sense to generalize, but with 2 I'm not sure that it does.

v2/pkg/genruntime/configmaps/collector.go
matthchr: Ah I feel sorta the opposite. This idiom matches the `err := func(); err != nil` idiom that Go has with almost all functions. The `len` checking one on the other hand doesn't immediately look like every other error handling block/return. I took a look through https://github.com/kubernetes-sigs and see all 3 usages of the package:

`return errors.NewAggregate(errs)` was probably the most common.
There was a good number of `if len(errs) > 0 { return errors.NewAggregate(errs)`, and also a good number of `err := errors.NewAggregate(errs); err != nil` (this idiom).

If folks feel strongly we should change this, I'm fine doing so, but I do think that the `err := <x>` idiom is the standard Go err handling idiom and this matches that and so scans easier. We also use this pattern elsewhere in the generator and generated code already (see for example [here](https://github.com/Azure/azure-service-operator/blob/main/v2/tools/generator/internal/codegen/pipeline/status_from_swagger.go#L176)). There are ~20 hits of the `if len(errs) > 0` idiom in ASO today and 18 of this one, so we're pretty evenly split in our overall usage on it right now as well.

v2/pkg/genruntime/merger/merger.go
matthchr: I didn't want to pull a logger parameter into this function, so instead I just added support for merging `Data` and `BinaryData` (including tests) for both secrets and configmaps.

We don't need it right now but who knows when we will.

v2/pkg/genruntime/merger/merger.go
matthchr: See above, added support for this and removed TODO

v2/pkg/genruntime/secrets/collector.go
matthchr: We'd also need to define an interface other than `metav1.Object` because there's no shared way to get/set values from `SecretData`/`BinaryData` somehow (not in the v1 package though so can't add methods to v1.ConfigMap/v1.Secret to do this easily...).

v2/pkg/genruntime/configmaps/collector.go
theunrepentantgeek: Evenly split! I would have _sworn_ we were mostly (or completely) using the `if len(errs) > 0` approach. 

My problem with it is that it relies on _hidden magic_ (behaviour that's not explicit nor even implied by the method name).

That said, if you prefer this approach, leave it the way it is.

v2/pkg/genruntime/configmaps/collector.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+		c.configs[dest.Name] = existing
+	}
+	return existing
+}
+
+func (c *Collector) errIfKeyExists(val *v1.ConfigMap, key string) error {
+	if _, ok := val.Data[key]; ok {
+		return errors.Errorf("key collision, entry exists for key %s in Data", key)
+	}
```
theunrepentantgeek, line 57: Missing blank line #nit

theunrepentantgeek (approved):

super-harsh (approved):

PR #2530, opened 2022-10-05
Add ability to export operatorSpec.ConfigMaps

matthchr:
Part of ongoing work related to https://github.com/Azure/azure-service-operator/issues/2317 and https://github.com/Azure/azure-service-operator/issues/2435.

**If applicable**:
- [x] this PR contains tests


v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -475,24 +477,45 @@ func (r *azureDeploymentReconcilerInstance) updateStatus(ctx context.Context) er
...
+		var additionalResources []client.Object
+		additionalResources, err = exporter.ExportKubernetesResources(ctx, r.Obj, r.ARMClient, r.Log)
+		if err != nil {
+			return err
+		}
+
+		resources = append(resources, additionalResources...)
+	}
+
+	// TODO: We do a bit of duplicate work here because each handler also does merging of it's own, but then we
```
matthchr, line 499: Possibly just turn this into a note rather than a TODO

v2/tools/generator/internal/codegen/pipeline/add_operator_spec.go
```diff
@@ -82,26 +112,172 @@ func createOperatorSpecIfNeeded(
...
+
+	return result, configMapDestinationDetails, nil
 }
 
 type operatorSpecBuilder struct {
 	idFactory     astmodel.IdentifierFactory
 	configuration *config.Configuration
 	resource      astmodel.TypeDefinition
 	definitions   astmodel.TypeDefinitionSet
+
```
matthchr, line 263: minor: extra newline here that is unneeded

v2/tools/generator/internal/codegen/pipeline/exported_configmaps.go
```diff
@@ -0,0 +1,54 @@
...
+ * Licensed under the MIT license.
+ */
+
+package pipeline
+
+import "github.com/Azure/azure-service-operator/v2/tools/generator/internal/astmodel"
+
+// ExportedProperties is a map of string property name to a chain of properties. For allowing the export of a property "foo" that writes
+// the status.a.b.c property to a config map would be "foo" -> []{status, a, b, c}
+type ExportedProperties = map[string][]*astmodel.PropertyDefinition
```
matthchr, line 12: This type name doesn't actually match the file name. I think I need to rename file.

v2/tools/generator/internal/functions/kubernetes_exporter_function.go
```diff
@@ -0,0 +1,221 @@
...
+	"fmt"
+	"go/token"
+
+	"github.com/dave/dst"
+
+	"github.com/Azure/azure-service-operator/v2/tools/generator/internal/astbuilder"
+	"github.com/Azure/azure-service-operator/v2/tools/generator/internal/astmodel"
+)
+
+type KubernetesExporterBuilder struct {
```
matthchr, line 18: minor: Missing some doc comments on these types.

v2/tools/generator/internal/config/object_model_configuration.go
```diff
@@ -191,6 +191,33 @@ func (omc *ObjectModelConfiguration) VerifyAzureGeneratedSecretsConsumed() error
...
+	err := visitor.Visit(omc)
+	if err != nil {
+		return nil, err
+	}
+
+	return result, nil
+}
+
+// VerifyAzureGeneratedConfigsConsumed returns an error if Azure generated configs were not used, nil otherwise.
+func (omc *ObjectModelConfiguration) VerifyAzureGeneratedConfigsConsumed() error {
```
matthchr, line 213: Confirm this is actually called/used

matthchr: It is

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -475,24 +477,44 @@ func (r *azureDeploymentReconcilerInstance) updateStatus(ctx context.Context) er
 // saveAssociatedKubernetesResources retrieves Kubernetes resources to create and saves them to Kubernetes.
 // If there are no resources to save this method is a no-op.
 func (r *azureDeploymentReconcilerInstance) saveAssociatedKubernetesResources(ctx context.Context) error {
+	// Check if this resource has a handcrafted extension for exporting
 	retriever := extensions.CreateKubernetesExporter(ctx, r.Extension, r.ARMClient, r.Log)
 	resources, err := retriever(r.Obj)
-	//_, err := retriever(r.Obj)
 	if err != nil {
 		return err
```
theunrepentantgeek, line 484: Do we need to wrap this error to indicate it was generated by extension code?

v2/tools/generator/internal/codegen/pipeline/add_kubernetes_exporter.go
```diff
@@ -0,0 +1,46 @@
...
+
+const AddKubernetesExporterStageID = "addKubernetesExporter"
+
+func AddKubernetesExporter(idFactory astmodel.IdentifierFactory) *Stage {
+	return NewStage(
+		AddKubernetesExporterStageID,
+		"Adds the KubernetesExporter interface to resources that need it",
+		func(ctx context.Context, state *State) (*State, error) {
+			defs := state.Definitions()
+			updateDefs := make(astmodel.TypeDefinitionSet)
```
theunrepentantgeek, line 23: #nitpick
```suggestion
			updatedDefs := make(astmodel.TypeDefinitionSet)
```

v2/tools/generator/internal/codegen/pipeline/add_kubernetes_exporter.go
```diff
@@ -0,0 +1,46 @@
...
+
+			mappings := state.GeneratedConfigMaps()
+
+			for _, def := range astmodel.FindResourceDefinitions(defs) {
+				resourceType := def.Type().(*astmodel.ResourceType)
```
theunrepentantgeek, line 28: While I think this works now, you should probably use `astmodel.AsResourceType()` in case it's wrapped with something else by a later change to the pipeline.

v2/tools/generator/internal/codegen/pipeline/add_operator_spec.go
```diff
@@ -82,26 +112,171 @@ func createOperatorSpecIfNeeded(
...
+	result.walker = walker
+
+	return result
+}
+
+func (w *configMapTypeWalker) includeSpecStatus(this *astmodel.TypeVisitor, it *astmodel.ResourceType, ctx interface{}) (astmodel.Type, error) {
+	typedCtx := ctx.(configMapContext)
+
+	// Fabricate a fake PropertyDetails here
+	prop := astmodel.NewPropertyDefinition("Status", "status", it.StatusType()) // TODO: This isn't actually guaranteed right now?
```
theunrepentantgeek, line 180: Do you need to take care of this `TODO` before merging?

theunrepentantgeek: Actually, call `it.Property("Status")` and you'll get the property you need. `ResourceType` implements `PropertyContainer`.

v2/tools/generator/internal/codegen/pipeline/exported_properties.go
```diff
@@ -0,0 +1,54 @@
...
+ * Licensed under the MIT license.
+ */
+
+package pipeline
+
+import "github.com/Azure/azure-service-operator/v2/tools/generator/internal/astmodel"
+
+// ExportedProperties is a map of string property name to a chain of properties. For allowing the export of a property "foo" that writes
+// the status.a.b.c property to a config map would be "foo" -> []{status, a, b, c}
+type ExportedProperties = map[string][]*astmodel.PropertyDefinition
```
theunrepentantgeek, line 12: Can you use (or generalize) PropertyChain here?

v2/tools/generator/internal/codegen/pipeline/state.go
```diff
@@ -16,10 +16,11 @@ import (
 
 // State is an immutable instance that captures the information being passed along the pipeline
 type State struct {
-	definitions     astmodel.TypeDefinitionSet // set of type definitions generated so far
-	conversionGraph *storage.ConversionGraph   // graph of transitions between packages in our conversion graph
-	stagesSeen      set.Set[string]            // set of ids of the stages already run
-	stagesExpected  map[string]set.Set[string] // set of ids of expected stages, each with a set of ids for the stages expecting them
+	definitions        astmodel.TypeDefinitionSet // set of type definitions generated so far
+	conversionGraph    *storage.ConversionGraph   // graph of transitions between packages in our conversion graph
+	exportedConfigMaps *ExportedTypeNameProperties
```
theunrepentantgeek, line 21: Why do we need this here, when we didn't add secrets here earlier?

v2/tools/generator/internal/functions/kubernetes_exporter_function.go
```diff
@@ -0,0 +1,237 @@
...
+
+		for i := len(propertyPath) - 1; i >= 0; i -= 1 {
+			propType := propertyPath[i].PropertyType()
+			if _, ok := astmodel.AsMapType(propType); ok {
+				panic(
+					fmt.Sprintf(
+						"Exporting Map elements as configmaps is not supported currently. Property %s has type %s",
+						propertyNames[i],
+						propType.String()))
+			}
```
theunrepentantgeek, line 123: #nit missing blank line

v2/tools/generator/internal/functions/kubernetes_exporter_function.go
```diff
@@ -0,0 +1,237 @@
...
+				receiverIdent,
+				collectorIdent,
+				propertyPath,
+				propertyNames,
+				operatorSpecPropertyName)
+		}
+
+		stmts = append(
+			stmts,
+			astbuilder.SimpleIf(condition, ifBlock))
```
theunrepentantgeek, line 172: Check out `astmodel.Statements()` - not only will it combine statements together, it will also flatten any `[]Stmt` you pass in.

Give your individual statements meaningful names, then call `Statements()` at the end of the sequence to stitch them together.

theunrepentantgeek (changes requested):
Some questions I'd like to discuss before approving. Looks good though.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
matthchr: Done, and for below.

v2/tools/generator/internal/codegen/pipeline/add_kubernetes_exporter.go
matthchr: This was a typo, I meant to call it `updatedDefs` -- good catch.

v2/tools/generator/internal/codegen/pipeline/add_kubernetes_exporter.go
matthchr: Done

v2/tools/generator/internal/codegen/pipeline/add_operator_spec.go
matthchr: That's part of the problem, but the other part of the problem is that there's no guarantee elsewhere in the code that the path actually is part of status. So probably I need to rewrite this to append spec to the spec path and status to the status path.

Basically the issue here is I am making the assumption that we're going to take the `Status` branch, but that's not actually required (although in practice it is because we won't export _spec_ fields it'd be weird)

v2/tools/generator/internal/codegen/pipeline/state.go
matthchr: The reason that secrets doesn't need this is that we're not actually codegenerating the implementation of gathering the secrets. Whereas for configmaps we are. So for secrets the pattern is:
1. Create the `OperatorSpec.Secrets` object.
2. The actual mapping between the `OperatorSpec.Secrets.<prop>` properties and their data source is defined by hand in an extension, so the code generator doesn't need to know what it is.

For configMaps the pattern is:
1. Create the `OperatorSpec.ConfigMaps` object.
2. Code generate an implementation of `KubernetesExporter` which automatically maps the user specified values in `OperatorSpec.ConfigMaps` onto a `v1.ConfigMap` and saves the corresponding property.

To accomplish item 2 for ConfigMaps, we need to remember which property corresponds to which `OperatorSpec.ConfigMaps` entry. So for example `OperatorSpec.ConfigMaps.PrincipalId` maps to the `.Status.PrincipalId` field. We need this information twice, once to generate the `OperatorSpec.ConfigMaps` type, and once to generate the implementation of the `KubernetesExporter` interface. Since I already had it, I stashed it into `Stage` and pulled it out later.

I suppose an alternative implementation would be to re-call the `getConfigMapProperties(defs, configuration, resource)` function on each resource definition and rebuild the mapping in the following stage. We have the information in the type graph + config to do that. It's a bit "risky" though because it does remove the direct linkage between the set of properties used in `OperatorSpec.ConfigMaps` and the set of properties that are autogenerated. So if for some reason stages were moved such that we found a different list for some reason the second time round, that would be bad (I'd have to panic/error in that case, which can't happen now because we're using exactly the same data source). In practice right now that's not a risk though because the two stages one after the other.

Thoughts?

v2/tools/generator/internal/functions/kubernetes_exporter_function.go
matthchr: Done. I still needed the list for the loop though, but renamed it to `collectStmts` (as it's the set of statements that involve the nil-guards + `collector.Add` call).

v2/tools/generator/internal/codegen/pipeline/exported_properties.go
matthchr: Yes, this is a `propertyChain`. It's already basically generalized enough I believe, but will fix up some comments and use it.

v2/tools/generator/internal/codegen/pipeline/state.go
theunrepentantgeek: This makes sense. Stashing the information on the `State` of the pipeline makes sense for this - it's a similar argument to why the conversion graph is there.

Thanks for clarifying!

v2/tools/generator/internal/codegen/pipeline/add_operator_spec.go
matthchr: Fix this.

v2/tools/generator/internal/codegen/pipeline/exported_properties.go
matthchr: Actually, this is not quite true. `propertyChain` is structured in such a way as to allow us to remove certain items from the path it generates. That is needed for the context where we are generating minimal index method names (we want as few properties in the path as we can fit while still making the method unique).

In this instance, the properties will actually be used in code (`a.b.c.d`) so there's no need to remove properties from the chain.

We could still use `propertyChain` here (it doesn't hurt), but it's defined in `pipeline` and we eventually pass this map to something in `astmodel/functions`. We can't do that with `propertyChain` so we'd end up having to call `chain.properties()` on it and then rebuild the map and pass it, which generates a good bit of unneeded boilerplate. Since the actual usage of the map is in the `astmodel/functions` helper, it doesn't get us anything to use `propertyChain` unless we can pass it everywere.

A solution to this would be to promote `propertyChain` to a first class citizen in `astmodel`. I'm 50-50 as to if it's worth doing that for this case because we're not getting a lot of advantage out of using it here, and if we do feel doing that is worth it, I am not sure we should do it in this PR as it'll be a pretty big refactor of `propertyChain` usage and this PR is already pretty large.

Thoughts?

theunrepentantgeek: Sounds like it should stay the way it is. We _might_ want to refactor it as you suggest, but certainly not in this PR.

theunrepentantgeek (approved):
Looks good. :shipit:

PR #2532, opened 2022-10-06
Helm Chart: add AAD Pod Identity authorization

mehighlow:

- Updates Helm Chart: automates AAD Pod Identity authorization
- Updates the documentation

**If applicable**:
- [ ] this PR contains tests


matthchr:
Currently the entire `templates` directory is generated via `Kustomize` and then modified with the following script: https://github.com/Azure/azure-service-operator/blob/main/scripts/generate-helm-manifest.sh

Because this functionality is Helm only, it probably doesn't make sense to include these in the Kustomize targets (since then they would also be included in the non-Helm release, which we don't want). cc @super-harsh, I'm wondering if we want to support this, and if we do, can we make the script aware of the handcrafted Helm files and not delete them during the CI run? Right now we delete the entire `/templates` directory in CI.

v2/charts/azure-service-operator/values.yaml
```diff
@@ -13,6 +13,15 @@ azureClientID: ""
 # This is required when using Service Principal authentication.
 azureClientSecret: ""
 
+# [Deprecated] Managed Identity (aad-pod-identity) Authentication
+# azureManagedIdentityId is the client ID (00000000-0000-0000-0000-00000000000) of the Azure Managed Identity used to authenticate with Azure if using aad-pod-identity(deprecated) authorization.
+# azureManagedIdentityResourceId is the resource ID (/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{managedIdentityName})
+# of the Azure Managed Identity used to authenticate with Azure if using aad-pod-identity authorization.
+aadPodIdentity:
+  enable: false
+  azureManagedIdentityClientId: ""
```
matthchr, line 22: I think you can just use the existing azureClientID variable. We don't need a separate one specific to aadPodIdentity.

mehighlow:
> Currently the entire `templates` directory is generated via `Kustomize` and then modified with the following script: https://github.com/Azure/azure-service-operator/blob/main/scripts/generate-helm-manifest.sh
> 
> Because this functionality is Helm only, it probably doesn't make sense to include these in the Kustomize targets (since then they would also be included in the non-Helm release, which we don't want). cc @super-harsh, I'm wondering if we want to support this, and if we do, can we make the script aware of the handcrafted Helm files and not delete them during the CI run? Right now we delete the entire `/templates` directory in CI.
- with the new workload identity released a week ago, add pod identity, indeed, will be deprecated. however, it would take 2-3 months to create all IaC and automated scripts to run this solution in production( I mean the IaC scripts, terraform, and other tools engineers rely on running repeatable production infrastructure)

- it is still getting updates and patches( the last one is 25 days ago) and will be getting patches. I bet it is still widely being used in production and would require a significant amount of time for those who run it in production to migrate to the workload identity (another 2-3 months of planning and implementation).

matthchr:
As pointed out in Teams, this actually works just fine out of the box. I misread our script.

docs/hugo/content/introduction/authentication.md
```diff
@@ -230,3 +232,18 @@ stringData:
...
+helm repo add aso2 https://raw.githubusercontent.com/Azure/azure-service-operator/main/v2/charts
+helm repo update
+
+helm upgrade --install --devel aso2 aso2/azure-service-operator \
+     --create-namespace \
+     --namespace=azureserviceoperator-system \
+     --set azureSubscriptionID=$AZURE_SUBSCRIPTION_ID \
+     --set aadPodIdentity.enable=true \
+     --set aadPodIdentity.azureManagedIdentityResourceId=${IDENTITY_RESOURCE_ID} \
+     --set aadPodIdentity.azureManagedIdentityClientId=${IDENTITY_CLIENT_ID}
```
matthchr, line 248:
```suggestion
     --set azureClientID=${IDENTITY_CLIENT_ID}
```

matthchr (approved):
Approved with 1 documentation suggestion

PR #2544, opened 2022-10-12
Update resource registration to support ConfigMaps

matthchr:
- Correctly indexes ConfigMaps in the resource registration function if needed.
- Use Indexer interface in controller_resources_gen index functions.

  The indexer interface makes it easier to abstract across various shapes that can be indexed (ConfigMap, Secret, etc). This helps us avoid generating code for each distinct shape, instead we can target the interface.

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/v2/api/cache/v1beta20201201/redis_linked_server_types_gen.go
+++ b/v2/api/cache/v1beta20201201/redis_linked_server_types_gen.go
@@ -431,9 +431,7 @@ func (server *Redis_LinkedServer_Spec) PopulateFromARM(owner genruntime.Arbitrar
 	}
 
 	// Set property ‘Owner’:
-	server.Owner = &genruntime.KnownResourceReference{
-		Name: owner.Name,
-	}
+	server.Owner = &genruntime.KnownResourceReference{Name: owner.Name}
 
 	// Set property ‘ServerRole’:
 	// copying flattened property:
--- a/v2/pkg/genruntime/indexer.go
+++ b/v2/pkg/genruntime/indexer.go
@@ -1,0 +1,13 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package genruntime
+
+type Indexer interface {
+	// Index returns the index of the Indexer. The index can be passed to a registration.Index to
+	// build an index for the controller-runtime client. If Index returns nil, there is nothing to index.
+	// See controller-runtime mgr.GetFieldIndexer().IndexField() for more details.
+	Index() []string
+}
--- a/v2/tools/generator/internal/astbuilder/composite_literals.go
+++ b/v2/tools/generator/internal/astbuilder/composite_literals.go
@@ -13,6 +13,7 @@ import (
 type CompositeLiteralBuilder struct {
 	structType dst.Expr
 	elts       []dst.Expr
+	newLines   bool
 }
 
 // NewCompositeLiteralBuilder creates a new instance for initialization of the specified struct
@@ -20,7 +21,14 @@ type CompositeLiteralBuilder struct {
 func NewCompositeLiteralBuilder(structType dst.Expr) *CompositeLiteralBuilder {
 	return &CompositeLiteralBuilder{
 		structType: structType,
+		newLines:   true,
 	}
+}
+
+// WithoutNewLines returns the CompositeLiteralBuilder without NewLines enabled
+func (b *CompositeLiteralBuilder) WithoutNewLines() *CompositeLiteralBuilder {
+	b.newLines = false
+	return b
 }
 
 // AddField adds initialization of another field
@@ -30,9 +38,10 @@ func (b *CompositeLiteralBuilder) AddField(name string, value dst.Expr) *Composi
 		Key:   dst.NewIdent(name),
 		Value: dst.Clone(value).(dst.Expr),
 	}
-
-	expr.Decs.Before = dst.NewLine
-	expr.Decs.After = dst.NewLine
+	if b.newLines {
+		expr.Decs.Before = dst.NewLine
+		expr.Decs.After = dst.NewLine
+	}
 
 	b.elts = append(b.elts, expr)
 	return b
--- a/v2/tools/generator/internal/astbuilder/composite_literals.go
+++ b/v2/tools/generator/internal/astbuilder/composite_literals.go
@@ -38,17 +38,26 @@ func (b *CompositeLiteralBuilder) AddField(name string, value dst.Expr) *Composi
 		Key:   dst.NewIdent(name),
 		Value: dst.Clone(value).(dst.Expr),
 	}
-	if b.newLines {
-		expr.Decs.Before = dst.NewLine
-		expr.Decs.After = dst.NewLine
-	}
+	expr.Decs.Before = dst.NewLine
+	expr.Decs.After = dst.NewLine
 
 	b.elts = append(b.elts, expr)
 	return b
 }
 
 // Build constructs the actual dst.CompositeLit that's required
-func (b CompositeLiteralBuilder) Build() *dst.CompositeLit {
+func (b *CompositeLiteralBuilder) Build() *dst.CompositeLit {
+	// If we only have a single element, remove the NewLine directives to force it onto a single line.
+	// While there are places where this doesn't look amazing, on balance it is cleaner than having them
+	// on multiple lines. This is especially true when the CompositeLit is used in a deeply nested context.
+	if len(b.elts) == 1 {
+		for _, elt := range b.elts {
+			kvExpr := elt.(*dst.KeyValueExpr) // This is safe because it's the only type we put into elts
+			kvExpr.Decs.Before = dst.None
+			kvExpr.Decs.After = dst.None
+		}
+	}
+
 	return &dst.CompositeLit{
 		Type: b.structType,
 		Elts: b.elts,
--- a/v2/tools/generator/internal/codegen/pipeline/export_controller_type_registrations.go
+++ b/v2/tools/generator/internal/codegen/pipeline/export_controller_type_registrations.go
@@ -48,6 +48,7 @@ func ExportControllerResourceRegistrations(idFactory astmodel.IdentifierFactory,
 						if err != nil {
 							return nil, errors.Wrapf(err, "failed to catalog %s secret property chains", def.Name())
 						}
+
 						configMapChains, err := catalogConfigMapPropertyChains(def, definitions)
 						if err != nil {
 							return nil, errors.Wrapf(err, "failed to catalog %s configmap property chains", def.Name())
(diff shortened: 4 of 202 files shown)


matthchr:
Only 2 PRs left after this until ConfigMap changes are all done.

v2/pkg/genruntime/indexer.go
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package genruntime
+
+type Indexer interface {
```
super-harsh, line 8: Should we consider having `Index` from `v2/pkg/genruntime/registration/registration.go` and this at the same place?

v2/tools/generator/internal/codegen/pipeline/resource_registration_file.go
```diff
@@ -503,3 +470,103 @@ func (r *ResourceRegistrationFile) getImportedPackages() map[astmodel.PackageRef
...
+	eventHandler := astbuilder.CallFunc(
+		"watchSecretsFactory",
+		astbuilder.SliceLiteral(dst.NewIdent("string"), secretKeyParams...),
+		astbuilder.AddrOf(astbuilder.NewCompositeLiteralBuilder(listTypeName).Build()))
+	newWatchBuilder.AddField("MakeEventHandler", eventHandler)
+
+	return newWatchBuilder.Build()
+}
+
+func (r *ResourceRegistrationFile) makeConfigMapWatchesExpr(typeName astmodel.TypeName, codeGenerationContext *astmodel.CodeGenerationContext) dst.Expr {
```
super-harsh, line 530: Are we using this method anywhere as of now?

v2/tools/generator/internal/codegen/pipeline/resource_registration_file.go
```diff
@@ -503,3 +470,103 @@ func (r *ResourceRegistrationFile) getImportedPackages() map[astmodel.PackageRef
...
+	}
+
+	if len(configMapKeys) == 0 {
+		return nil
+	}
+
+	newWatchBuilder := astbuilder.NewCompositeLiteralBuilder(astmodel.WatchRegistrationType.AsType(codeGenerationContext))
+	sort.Strings(configMapKeys)
+
+	// Not using astbuilder here because we want this to go onto a single line
```
super-harsh, line 543: maybe extract this into a method with a parameter `Type` so that can be used by secretExpr as well?

v2/pkg/genruntime/indexer.go
matthchr: Possibly. `Indexer` is a bit more generic than `Index`, as `Index` is really used specifically for our registration, whereas in theory `Indexer` could be used generically to define any index, so I do think it's probably ok that one of them is in `genruntime` while one is in `genruntime/registration`, reflecting its slightly more scoped nature.

In practice right now they're both basically used in the same places, but that isn't necessarily going to be the case forever.

v2/tools/generator/internal/codegen/pipeline/resource_registration_file.go
matthchr: I updated the `astbuilder` helper so that we could use it here. It's not tons shorter but it it somewhat shorter.

v2/tools/generator/internal/codegen/pipeline/resource_registration_file.go
matthchr: This is used now

v2/tools/generator/internal/codegen/pipeline/resource_registration_file.go
matthchr: Or did you mean this whole method? I can look into that although there's more to it than just a single type parameter.

v2/tools/generator/internal/astbuilder/composite_literals.go
```diff
@@ -13,26 +13,35 @@ import (
...
+		newLines:   true,
 	}
 }
 
+// WithoutNewLines returns the CompositeLiteralBuilder without NewLines enabled
+func (b *CompositeLiteralBuilder) WithoutNewLines() *CompositeLiteralBuilder {
+	b.newLines = false
+	return b
+}
+
```
theunrepentantgeek, line 33: Do we need this to be explicit, or could we make this automatic?
(I'm not a fan of pushing more cognitive load onto the API unless we have to.)

In `Build()` we could check the length of `elts` and force newlines on the members only if there's more than one.
We'd need to also change the field from `elts []dst.Expr` to `elts []*dst.KeyValueExpr`, but I think that's safe.

We already have some smarts like this elsewhere in `astbuilder`, when generating function calls IIRC.

v2/tools/generator/internal/codegen/pipeline/export_controller_type_registrations.go
```diff
@@ -43,14 +44,21 @@ func ExportControllerResourceRegistrations(idFactory astmodel.IdentifierFactory,
 					if resource.IsStorageVersion() {
 						storageVersionResources = append(storageVersionResources, def.Name())
 
-						chains, err := catalogSecretPropertyChains(def, definitions)
+						secretChains, err := catalogSecretPropertyChains(def, definitions)
 						if err != nil {
-							return nil, errors.Wrapf(err, "failed to catalog %s property chains", def.Name())
+							return nil, errors.Wrapf(err, "failed to catalog %s secret property chains", def.Name())
+						}
```
theunrepentantgeek, line 50: Missing blank line #nit

v2/pkg/genruntime/indexer.go
```diff
@@ -0,0 +1,13 @@
...
+	// Index returns the index of the Indexer. The index can be passed to a registration.Index to
+	// build an index for the controller-runtime client. If Index returns nil, there is nothing to index.
+	// See controller-runtime mgr.GetFieldIndexer().IndexField() for more details.
+	Index() []string
```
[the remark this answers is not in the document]
matthchr: Because `controller-runtime mgr.GetFieldIndexer().IndexField()` operators on an `IndexerFunc` whose signature is `func(Object) []string`.

The main reason I kept that support for our `Indexer` is twofold:
1. Just to match.
2. If we ever have a `Ref` type that actually refers to a collection of resources, it fits into the existing interface. With this said it would have to be the type itself that referred to a collection, and I'm not sure if we'll ever actually do that in practice.

Happy to restrict this to `string` if you think that's better/clearer.

v2/tools/generator/internal/astbuilder/composite_literals.go
matthchr: I took a look into doing this. It's totally doable, but the problem is that I am not sure we universally want all CompositeLiterals to be on a single line if they only have one property.

In my mind, it's more about the context they're being used in. If they're being assigned to a variable directly and aren't already deeply nested, then this is ok (or preferable, even):
```
       account.Owner = &genruntime.KnownResourceReference{
               Name: owner.Name,
       }
```

whereas, if they're deeply nested, then I think we prefer this:
```
Watches: []registration.Watch{
			{
				Src:              &source.Kind{Type: &v1.ConfigMap{}},
				MakeEventHandler: watchConfigMapsFactory([]string{".spec.principalIdConfigRef"}, &authorization_v20200801ps.RoleAssignmentList{}),
			},
		},
```

vs

```
Watches: []registration.Watch{
			{
				Src: &source.Kind{
					Type: &v1.ConfigMap{}
				},
				MakeEventHandler: watchConfigMapsFactory([]string{".spec.principalIdConfigRef"}, &authorization_v20200801ps.RoleAssignmentList{}),
			},
		},
```

Since context matters it feels like we should allow the user to choose. Thoughts?

matthchr: Hmm... thinking further on this, I don't hate 
```
       account.Owner = &genruntime.KnownResourceReference{Name: owner.Name}
```

so if folks are ok with that everywhere I can do this (single-line if 1 element) and we can expand later when we run into a case where we really dislike it?

matthchr: Check the "Increase CompositeLiteralBuilder smarts" commit to see the effect of this change. Let me know if you dislike it, otherwise I will keep it.

v2/tools/generator/internal/astbuilder/composite_literals.go
```diff
@@ -38,17 +38,23 @@ func (b *CompositeLiteralBuilder) AddField(name string, value dst.Expr) *Composi
...
 
 	b.elts = append(b.elts, expr)
 	return b
 }
 
 // Build constructs the actual dst.CompositeLit that's required
-func (b CompositeLiteralBuilder) Build() *dst.CompositeLit {
+func (b *CompositeLiteralBuilder) Build() *dst.CompositeLit {
+	if len(b.elts) == 1 {
+		for _, elt := range b.elts {
```
theunrepentantgeek, line 51: Add a comment explaining what we're doing here

v2/tools/generator/internal/astbuilder/composite_literals.go
theunrepentantgeek: I'm happy with that change - just made some comment requesting a comment. #recursion

theunrepentantgeek (approved):

v2/api/cache/v1beta20201201/redis_linked_server_types_gen.go
```diff
@@ -431,9 +431,7 @@ func (server *Redis_LinkedServer_Spec) PopulateFromARM(owner genruntime.Arbitrar
 	}
 
 	// Set property ‘Owner’:
-	server.Owner = &genruntime.KnownResourceReference{
-		Name: owner.Name,
-	}
+	server.Owner = &genruntime.KnownResourceReference{Name: owner.Name}
```
super-harsh, line 434: This inline does not look right? it was fine earlier I guess.

v2/tools/generator/internal/codegen/pipeline/resource_registration_file.go
matthchr: Reduced duplication between the `secrets` and `configmap` paths here by refactoring to use a `makeSimpleWatchesExpr`.

We might need other variants in the future, but for now the two are similar enough that they can share the same code path.

v2/tools/generator/internal/astbuilder/composite_literals.go
matthchr: Resolving this. Verdict is to take the change that @theunrepentantgeek suggested.

v2/api/cache/v1beta20201201/redis_linked_server_types_gen.go
matthchr: As discussed in Teams, will take this change (the new one-line way) as the consistency is nice and it's overall better when considering the use of all of our Composite Literals.

v2/pkg/genruntime/indexer.go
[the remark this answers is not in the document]
theunrepentantgeek: Happy to leave them returning `[]string` - I didn't infer this reasoning by reading the code, which is why I queried.

super-harsh (approved):

PR #2549, opened 2022-10-17
Change mysql live tests location to Australia East

super-harsh:
RP keeps restricting the US zones for the org subs, so changing it to Australia East.


v2/test/mysql_test.go
```diff
@@ -417,7 +417,7 @@ func newSecret(tc *testcommon.KubePerTestContext, key string, password string) *
 func newMySQLServer(tc *testcommon.KubePerTestContext, rg *resources.ResourceGroup, adminUsername string, adminKey string, adminSecretName string) *mysql.FlexibleServer {
 	// Force this test to run in a region that is not capacity constrained.
 	// location := tc.AzureRegion TODO: Uncomment this line when West US 2 is no longer constrained
-	location := to.StringPtr("East US")
+	location := to.StringPtr("Australia East")
```
theunrepentantgeek, line 420: I'm surprised using the display name for the region works - would have expected to need `australiaeast` or `eastus`.

Suggest changing to use the region code as that's more standard.

theunrepentantgeek (approved):

PR #2559, opened 2022-10-20
Add support for namespaced secrets

super-harsh:

Closes:  #1173 

This PR adds support to run operator in multitenant mode using one operator and multiple credentials using namespaced secrets. PR includes the following:
1. Support for namespaced secret using ARMClientFactory
2. Block users to change subscriptionID once resource is created in a subscription
3. Push events for credential used on resource


v2/internal/reconcilers/arm/azure_generic_arm_reconciler.go
```diff
@@ -91,28 +93,47 @@ func (r *AzureDeploymentReconciler) asARMObj(obj genruntime.MetaObject) (genrunt
 	return typedObj, nil
 }
 
-func (r *AzureDeploymentReconciler) makeInstance(log logr.Logger, eventRecorder record.EventRecorder, obj genruntime.MetaObject) (*azureDeploymentReconcilerInstance, error) {
+func (r *AzureDeploymentReconciler) makeInstance(log logr.Logger, eventRecorder record.EventRecorder, obj genruntime.MetaObject, ctx context.Context) (*azureDeploymentReconcilerInstance, error) {
```
matthchr, line 96: minor: `ctx` as first parameter.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler.go
```diff
@@ -91,28 +93,47 @@ func (r *AzureDeploymentReconciler) asARMObj(obj genruntime.MetaObject) (genrunt
...
 	log = log.WithValues("azureName", typedObj.AzureName())
 
+	armClient, err := r.ARMClientFactory(typedObj, ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	eventRecorder.Eventf(typedObj, v1.EventTypeNormal, "CredentialFrom", "Using credential from %q", armClient.CredentialFrom().String())
+
+	resourceID, ok := genruntime.GetResourceID(typedObj)
```
matthchr, line 111: I think this check should be in `azure_generic_arm_reconciler_instance.go`, in two places:
1. `func BeginCreateOrUpdateResource`
2. `func deleteResource`

and the error that we return should be a `conditions.ReadyConditionImpactingError`, with severity `conditions.Error` and a new reason `SubscriptionMismatch`

matthchr: The reason I think that is:
1. `makeInstance` isn't supposed to have a lot of buisness logic in it
2. You have access to the `ReadyCondition` stuff there which will make it easier to produce an error which will lead to a user visible condition which more easily conveys this problem to them.

matthchr: Probably since you're putting this into two places you'll wanna put this check into a helper function and just call that in both spots.

matthchr (commented):
Great start! Have left a few comments and let me know what you think.

v2/internal/genericarmclient/generic_client.go
```diff
@@ -39,17 +40,18 @@ type GenericClient struct {
...
 	metrics        metrics.ARMClientMetrics
+	credentialFrom types.NamespacedName
 }
 
 // TODO: Need to do retryAfter detection in each call?
 
 // NewGenericClient creates a new instance of GenericClient
-func NewGenericClient(cloudCfg cloud.Configuration, creds azcore.TokenCredential, subscriptionID string, metrics metrics.ARMClientMetrics) (*GenericClient, error) {
-	return NewGenericClientFromHTTPClient(cloudCfg, creds, nil, subscriptionID, metrics)
+func NewGenericClient(cloudCfg cloud.Configuration, creds azcore.TokenCredential, subscriptionID string, metrics metrics.ARMClientMetrics, credentialFrom types.NamespacedName) (*GenericClient, error) {
```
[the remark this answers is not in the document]
super-harsh: Yep, had the same thought about it. Its just I wanted to have the events getting pushed for the credentials used for latest reconcile. Will see if I can pass event recorded in ARMClientfactory and push events there

v2/internal/genericarmclient/arm_ids.go
```diff
@@ -73,3 +73,13 @@ func MakeResourceGroupID(subscription string, resourceGroup string) string {
 func MakeSubscriptionID(subscription string) string {
 	return fmt.Sprintf("/subscriptions/%s", subscription)
 }
+
+func GetSubscription(path string) (string, error) {
```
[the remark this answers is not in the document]
super-harsh: ah! This is just moved from [here ](https://github.com/Azure/azure-service-operator/pull/2559/files/92bb540d5443e2a3039e953e1f634a7884a34294#diff-0614b9ea6824ce28ae2dc1bc07900f2929014ee58a6b4cb2da29f5a52f6d3521L100). Just wanted for it to be at the right place and be exported to be re-used.

v2/internal/reconcilers/arm/arm_client_factory.go
```diff
@@ -6,8 +6,143 @@
...
+}
+
+func (c ARMClients) newClientFromSecret(secret *v1.Secret, nsName types.NamespacedName, cloud cloud.Configuration) (*genericarmclient.GenericClient, error) {
+	subscriptionID, ok := secret.Data[config.SubscriptionIDVar]
+	if !ok {
+		return nil, resolver.NewSecretNotFoundError(nsName, errors.Errorf("Credential Secret %q does not contain key %q", nsName.String(), config.SubscriptionIDVar))
+	}
+	tenantID, ok := secret.Data[config.TenantIDVar]
+	if !ok {
+		return nil, resolver.NewSecretNotFoundError(nsName, errors.Errorf("Credential Secret %q does not contain key %q", nsName.String(), config.TenantIDVar))
```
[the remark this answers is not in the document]
super-harsh: I used it because its already there. Good idea to move it to a generic place and use it.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler.go
super-harsh: done!

v2/internal/genericarmclient/arm_ids.go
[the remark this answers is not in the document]
matthchr: Yeah, I mean honestly you could probably not even bother with the `GetSubscription` function and instead just call `arm.ParseResourceID` right there, so that you have access to other stuff like `ResourceGroup` too if you need it (which as we discussed, you might actually).

v2/internal/reconcilers/arm/arm_client.go
```diff
@@ -0,0 +1,166 @@
...
+
+func (c armClient) GenericClient() *genericarmclient.GenericClient {
+	return c.genericClient
+}
+
+func (c armClient) CredentialFrom() types.NamespacedName {
+	return c.credentialFrom
+}
+
+type armClients struct {
```
matthchr, line 46: minor: Consider calling this `armClientsCache` (or maybe `armClientCache`, not sure if the `s` is better or not).

v2/internal/reconcilers/arm/arm_client.go
```diff
@@ -0,0 +1,166 @@
...
+func newARMClient(client *genericarmclient.GenericClient, secretData map[string][]byte, credentialFrom types.NamespacedName) *armClient {
+	return &armClient{
+		genericClient:  client,
+		secretData:     secretData,
+		credentialFrom: credentialFrom,
+	}
+
+}
+
+func (c armClients) Register(client *armClient) {
```
matthchr, line 75: Should use a ptr receiver here and all down the file, to match the `*armClients` returned from `NewARMClients` (I know it doesn't _really_ matter but it's a small clarity thing that helps IMO)

v2/internal/reconcilers/arm/arm_client_factory.go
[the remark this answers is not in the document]
matthchr: Yeah, as discussed maybe you can move it to `genruntime/core/errors.go`

v2/internal/reconcilers/arm/arm_client.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func (c armClients) Lookup(key string) (*armClient, bool) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	client, ok := c.clients[key]
+	return client, ok
+}
+
+func (c armClients) GetClient(ctx context.Context, obj genruntime.ARMMetaObject, kubeClient kubeclient.Client, cloud cloud.Configuration) (*armClient, error) {
```
matthchr, line 88: This method also can instead just take a namespace as a parameter rather than the whole `genruntime.ARMMetaObject`.

matthchr, line 88: I'm wondering if it would make sense to pass `kubeClient` to `armClients` at construction, so that you don't have to pass it every time here?

This could then be: ` GetClient(ctx context.Context, cloud cloud.Configuration, namespace string) (*armClient, error)`

v2/internal/reconcilers/arm/arm_client_factory.go
```diff
@@ -6,8 +6,11 @@
...
 import (
+	"context"
+
 	"github.com/Azure/azure-service-operator/v2/internal/genericarmclient"
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+	"k8s.io/client-go/tools/record"
 )
 
-type ARMClientFactory func(genruntime.ARMMetaObject) *genericarmclient.GenericClient
+type ARMClientFactory func(context.Context, genruntime.ARMMetaObject, record.EventRecorder) (*genericarmclient.GenericClient, error)
```
matthchr, line 16: I actually think it's OK for you to turn this into a struct like you had before (at least, I think you had that before).

Then you could pass the `record.EventRecorder` to that struct and do the event emission in the bit of code that constructs a new client (in `GetClient`).

Thoughts?

matthchr: Basically I am not sure we need both `armClients` _and_ `armClientFactory`. They seem two things that are tackling the same or very similar problem?

Am I missing the need for both?

v2/internal/reconcilers/arm/arm_client.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	credential, err := azidentity.NewClientSecretCredential(string(tenantID), string(clientID), string(clientSecret), nil)
+	if err != nil {
+		return nil, "", errors.Wrap(err, errors.Errorf("Invalid Client Secret Credential for %q encountered", nsName.String()).Error())
+	}
+	return credential, string(subscriptionID), nil
+}
+
+func getSecret(ctx context.Context, kubeClient kubeclient.Client, namespace string, secretName string) (*v1.Secret, error) {
```
matthchr, line 147: Yeah and you could tack this on as a private method of `armClients` and if you took `kubeClient` as a parameter of `armClients` you wouldn't need to pass it here either.

v2/internal/reconcilers/arm/arm_client.go
```diff
@@ -0,0 +1,166 @@
...
+		nsName := types.NamespacedName{Namespace: secret.Namespace, Name: secret.Name}
+		client, ok := c.Lookup(nsName.String())
+		// if client does not already exist, or secret data is changed, we need to get new token.
+		if !ok || !matchSecretData(client.secretData, secret.Data) {
+			credential, subscriptionID, err := c.newCredentialFromSecret(secret, nsName)
+			if err != nil {
+				return nil, err
+			}
+
+			newClient, err := genericarmclient.NewGenericClient(cloud, credential, subscriptionID, metrics.NewARMClientMetrics())
```
matthchr, line 105: Still need to pass `httpClient` here. (Resolving my other comment as it was out of date)

v2/internal/reconcilers/arm/azure_generic_arm_reconciler.go
matthchr: Is this actually fixed, it doesn't seem to be?

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -608,6 +630,11 @@ func deleteResource(
 		return ctrl.Result{}, nil
 	}
 
+	err := checkSubscription(resourceID, armClient.SubscriptionID())
```
matthchr, line 633: Nice, this works out well.

v2/internal/testcommon/test_context.go
```diff
@@ -20,18 +20,17 @@ import (
 	"testing"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	resources "github.com/Azure/azure-service-operator/v2/api/resources/v1beta20200601"
```
matthchr, line 23: Don't reorder this files imports. If you're using GoLand consider setting your imports options like this:

image

That matches the options we use in ASO and AKS.

v2/pkg/genruntime/conditions/ready_condition_builder.go
```diff
@@ -6,6 +6,9 @@ Licensed under the MIT license.
 package conditions
 
 const (
+	// Auth reasons
+	ReasonSubscriptionMismatch = "SubscriptionMismatch"
```
matthchr, line 10: minor: IMO this fits OK into "other reasons", it doesn't need its own section.

matthchr (commented):
Some more small comments, thanks for the quick turnaround!

v2/pkg/genruntime/conditions/ready_condition_builder.go
super-harsh: we'll have to add more reasons like `CredentialSecretNotFound` into it when we do per-rg secrets, so thought lets have a separate place for these.

v2/internal/reconcilers/arm/arm_client.go
super-harsh: Have passed in the whole object here to keep the access to annotations in mind for per-rg credentials. This method is anyway gonna change a lot when we add more logic for other credential type as well.

v2/internal/reconcilers/arm/arm_client_factory.go
super-harsh: `armClientFactory` is just a factory method which gets passed down to the reconciler so that each time we reconcile a resource, we can have values put into the factory method, which then has access to the to the `armClientCache`.

super-harsh: I used eventRecorder as a parameter so that we don't have to call `mgr.getRecorder` and have another instance of recorder, rather we can just pass it [here](https://github.com/Azure/azure-service-operator/pull/2559/files#diff-48b587518825e853a2effad5b6a94c94327f2a645f701d9f71c18d954c0643ecR102) and use it.

v2/internal/reconcilers/arm/arm_client_factory.go
```diff
@@ -6,8 +6,143 @@
...
+}
+
+func (c ARMClients) Lookup(key string) (ARMClient, bool) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	client, ok := c.clients[key]
+	return client, ok
+}
+
+func (c ARMClients) GetClient(obj genruntime.ARMMetaObject, kubeClient kubeclient.Client, ctx context.Context, cloud cloud.Configuration) (*genericarmclient.GenericClient, error) {
```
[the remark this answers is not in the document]
super-harsh: Have moved the traction for CredentialFrom to armClient

v2/internal/reconcilers/arm/arm_client.go
super-harsh: Done!

v2/internal/reconcilers/arm/arm_client_factory.go
super-harsh: As we're using `armClientFactory` in `kubeTestContext` as well with a completely different logic, I think for now, as we discussed, its better to keep both and tackle this in a separate PR

v2/internal/reconcilers/arm/arm_client.go
super-harsh: done!

v2/internal/reconcilers/arm/arm_client.go
super-harsh: done!

v2/internal/testcommon/test_context.go
super-harsh: Thanks for this, Have updated.

v2/internal/config/vars.go
```diff
@@ -14,8 +14,11 @@ import (
 )
 
 const (
+	// #nosec
+	AzureClientSecret          = "AZURE_CLIENT_SECRET"
 	SubscriptionIDVar          = "AZURE_SUBSCRIPTION_ID"
 	TenantIDVar                = "AZURE_TENANT_ID"
+	AzureClientID              = "AZURE_CLIENT_ID"
```
matthchr, line 21: minor: put `var` suffix on the end of these, to indicate they're the value of the variable name, not the actual secret or client id themselves? (Follows pattern of other names here)

v2/internal/genericarmclient/arm_ids.go
```diff
@@ -73,3 +74,11 @@ func MakeResourceGroupID(subscription string, resourceGroup string) string {
...
```
matthchr, line 78: minor: Method comment

v2/pkg/genruntime/conditions/ready_condition_builder.go
matthchr: Ah, good point. Resolving this as I agree with you above.

v2/internal/reconcilers/arm/arm_client.go
```diff
@@ -0,0 +1,34 @@
...
+
+package arm
+
+import (
+	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/Azure/azure-service-operator/v2/internal/genericarmclient"
+)
+
+type armClient struct {
```
matthchr, line 14: minor: A comment on this type might help explain what exactly it is: It holds the client as well as the secret used to create it and the credential from which that secret was retrieved (or something like that?)

v2/internal/reconcilers/arm/arm_client_cache.go
```diff
@@ -0,0 +1,160 @@
...
+func NewARMClientCache(
+	client *genericarmclient.GenericClient,
+	podNamespace string,
+	kubeClient kubeclient.Client,
+	configuration cloud.Configuration,
+	httpClient *http.Client) *armClientCache {
+
+	globalClient := &armClient{
+		genericClient:  client,
+		credentialFrom: types.NamespacedName{Name: "aso-controller-settings", Namespace: podNamespace},
```
matthchr, line 54: minor: should `aso-controller-settings` be a const?

v2/internal/reconcilers/arm/arm_client_cache.go
```diff
@@ -0,0 +1,160 @@
...
+		lock:         sync.Mutex{},
+		clients:      make(map[string]*armClient),
+		globalClient: globalClient,
+		kubeClient:   kubeClient,
+		cloudConfig:  configuration,
+		httpClient:   httpClient,
+	}
+}
+
+func (c *armClientCache) Register(client *armClient) {
```
matthchr, line 67: Should this be `register` instead, since I don't think it's used outside of this object?

v2/internal/reconcilers/arm/arm_client_cache_test.go
```diff
@@ -0,0 +1,226 @@
...
+const (
+	fakeID = "00000000-0000-0000-0000-000000000000"
+)
+
+func NewFakeKubeClient(s *runtime.Scheme) kubeclient.Client {
+	fakeClient := fake.NewClientBuilder().WithScheme(s).Build()
+	return kubeclient.NewClient(fakeClient)
+}
+
+func NewTestArmClientCache(client kubeclient.Client) (*armClientCache, error) {
```
matthchr, line 39: very minor: `Arm` should be `ARM` I think

v2/internal/reconcilers/arm/arm_client_cache_test.go
```diff
@@ -0,0 +1,226 @@
...
+		return nil, err
+	}
+
+	return &testResources{
+		armClientCache: cache,
+		kubeClient:     client,
+	}, nil
+}
+
+func Test_ARMClientCache_ReturnsNamespacedSecretClient(t *testing.T) {
```
matthchr, line 87: minor: Would calling this `Test_ARMClientCache_ReturnsNamespaceScopedClient` be slightly clearer? IIRC the client it's returning isn't a `SecretClient`, which is sorta what this test makes me think at first glance.

v2/internal/reconcilers/arm/arm_client_cache_test.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	g.Expect(len(res.armClientCache.clients)).To(BeEquivalentTo(1))
+	g.Expect(credentialFrom).To(BeEquivalentTo(credentialNamespacedName.String()))
+	g.Expect(client.SubscriptionID()).To(BeEquivalentTo(fakeID))
+	g.Expect(client).To(Not(BeEquivalentTo(res.armClientCache.globalClient.GenericClient())))
+
+}
+
+func Test_ARMClientCache_ReturnsNamespacedSecretClient_SecretChanged(t *testing.T) {
```
matthchr, line 122: Same as above: `Test_ARMClientCache_ReturnsNamespaceScopedClient_SecretChanged`

v2/internal/reconcilers/arm/azure_generic_arm_reconciler.go
```diff
@@ -91,28 +92,35 @@ func (r *AzureDeploymentReconciler) asARMObj(obj genruntime.MetaObject) (genrunt
...
 	}
 	// Augment Log with ARM specific stuff
 	log = log.WithValues("azureName", typedObj.AzureName())
 
+	armClient, credentialFrom, err := r.ARMClientFactory(ctx, typedObj)
+	if err != nil {
+		return nil, err
+	}
+
+	eventRecorder.Eventf(obj, v1.EventTypeNormal, "credentialFrom", "Using credential from %q", credentialFrom)
```
matthchr, line 108:
```suggestion
	eventRecorder.Eventf(obj, v1.EventTypeNormal, "CredentialFrom", "Using credential from %q", credentialFrom)
```

I think our one other event is uppercase currently and I took a quick look at Kubernetes documentation and it showed them using Uppercase events as well.

matthchr (approved):
This came out great!

Left a few minor comments around test naming, variable casing, but nothing major.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler.go
super-harsh: I saw this one, k8s automagically makes it upper case. But yeah, worth making it upper case here

v2/internal/reconcilers/arm/arm_client_cache.go
```diff
@@ -0,0 +1,162 @@
...
+	err := c.kubeClient.Get(
+		ctx,
+		types.NamespacedName{Namespace: namespace, Name: secretName},
+		secret)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, nil
+		} else {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
theunrepentantgeek: Should this error be wrapped with context?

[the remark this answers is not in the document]
theunrepentantgeek: (I won't tag other cases.)

v2/internal/reconcilers/arm/arm_client_cache_test.go
```diff
@@ -0,0 +1,226 @@
...
+	client, credentialFrom, err := res.armClientCache.GetClient(ctx, rg)
+	if err != nil {
+		return
+	}
+
+	g.Expect(len(res.armClientCache.clients)).To(BeEquivalentTo(1))
+	g.Expect(credentialFrom).To(BeEquivalentTo(credentialNamespacedName.String()))
+	g.Expect(client.SubscriptionID()).To(BeEquivalentTo(fakeID))
+	g.Expect(client).To(Not(BeEquivalentTo(res.armClientCache.globalClient.GenericClient())))
+
```
theunrepentantgeek, line 119: Extra line.

PR #2560, opened 2022-10-20
Remove unneeded delete optimization

matthchr:
The optimization is faulty because there is no guarantee that the client caches are up to date. It is possible for delete of a child resource to happen when the parent resource isn't in the cache. This is especially probably immediately after pod restart or if the parent has the reconcile-action: skip annotation and has already been deleted.

Most of the rest of the PR is re-recording tests and samples.

**If applicable**:
- [x] this PR contains tests


v2/internal/genericarmclient/generic_client.go
```diff
@@ -267,6 +267,12 @@ func (client *GenericClient) BeginDeleteByID(ctx context.Context, resourceID str
 		return nil, err
 	}
 
+	// TODO: Poller doesn't understand how to deal with a 404 response, but we want to treat 404 as success, so we need
+	// TODO: to return a poller that is just done right away
+	//if runtime.HasStatusCode(resp, http.StatusNotFound) {
```
matthchr, line 272: Remove this

v2/samples/authorization/v1beta/v1beta20200801preview_roleassignment.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: authorization.azure.com/v1beta20200801preview
 kind: RoleAssignment
 metadata:
-  name: b5ce4677-d53d-4532-aa18-ac32aecb1117 # Should be UUID
+  name: fee8b6b1-fe6e-481d-8330-0e950e4e6b86 # Should be UUID
```
matthchr, line 4: Once my other PR merges I can update these samples to use a dynamically provisioned identity and avoid hardcoding principalId

theunrepentantgeek (approved):
LGTM

v2/samples/documentdb/mongodb/v1beta/v1beta20210515_mongodbdatabase.yaml
```diff
@@ -1,14 +1,14 @@
 apiVersion: documentdb.azure.com/v1beta20210515
 kind: MongodbDatabase
 metadata:
-  name: sample-db-mongodb
+  name: sample-db-mongodb-beta
```
super-harsh, line 4: Why do we need this?

super-harsh (approved):
Just a minor comment to know why we need `-beta` post fix in db samples. Rest LGTM.

v2/samples/documentdb/mongodb/v1beta/v1beta20210515_mongodbdatabase.yaml
matthchr: We have name randomization turned off for documentdb (not entirely sure why). So the two samples alpha/beta having the same names means they conflict when you try to re-record them both at the same time.

Changing the name means that you can re-record both samples at the same time which makes things faster.

PR #2576, opened 2022-11-01
Add support for per resource credential

### LINKED ISSUE #1173: As a cluster admin, I can configure separate set of credentials for deployment from different AKS namespaces

super-harsh:
Closes #1173 

**Special notes for your reviewer:**
Follow up PR would include usage docs and updated design doc

### DIFF
--- a/v2/internal/reconcilers/arm/arm_client_cache.go
+++ b/v2/internal/reconcilers/arm/arm_client_cache.go
@@ -93,7 +93,7 @@ func (c *armClientCache) GetClient(ctx context.Context, obj genruntime.ARMMetaOb
 		return client.GenericClient(), client.CredentialFrom(), nil
 	}
 	// Namespaced secret
-	client, err = c.checkNamespacedSecret(ctx, obj, err)
+	client, err = c.checkNamespacedSecret(ctx, obj)
 	if err != nil {
 		return nil, "", err
 	} else if client != nil {
@@ -104,7 +104,7 @@ func (c *armClientCache) GetClient(ctx context.Context, obj genruntime.ARMMetaOb
 	return c.globalClient.GenericClient(), c.globalClient.CredentialFrom(), nil
 }
 
-func (c *armClientCache) checkNamespacedSecret(ctx context.Context, obj genruntime.ARMMetaObject, err error) (*armClient, error) {
+func (c *armClientCache) checkNamespacedSecret(ctx context.Context, obj genruntime.ARMMetaObject) (*armClient, error) {
 	secret, err := c.getSecret(ctx, obj.GetNamespace(), namespacedSecretName)
 	if err != nil {
 		return nil, err
--- a/v2/internal/reconcilers/arm/arm_client_cache.go
+++ b/v2/internal/reconcilers/arm/arm_client_cache.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"net/http"
 	"reflect"
+	"strings"
 	"sync"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
@@ -31,7 +32,9 @@ const (
 	// #nosec
 	globalCredentialSecretName = "aso-controller-settings"
 	// #nosec
-	namespacedSecretName = "aso-credential"
+	namespacedSecretName        = "aso-credential"
+	perResourceSecretAnnotation = "serviceoperator.azure.com/credential-from"
+	namespacedNameSeparator     = "/"
 )
 
 // armClientCache is a cache for armClients to hold multiple credential clients and global credential client.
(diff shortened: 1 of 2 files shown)


v2/internal/reconcilers/arm/arm_client_cache_test.go
```diff
@@ -199,19 +302,19 @@ func newSecret(name string, namespace string) *v1.Secret {
 	}
 }
 
-func newResourceGroup(name string, namespace string) *resources.ResourceGroup {
+func newResourceGroup(namespace string) *resources.ResourceGroup {
```
theunrepentantgeek, line 305: Why hard code the name of the resource group?

theunrepentantgeek (approved):

v2/internal/reconcilers/arm/arm_client_cache_test.go
super-harsh: Its to avoid linter complaining about the same name being passed every time

v2/internal/reconcilers/arm/arm_client_cache.go
```diff
@@ -82,37 +85,96 @@ func (c *armClientCache) lookup(key string) (*armClient, bool) {
...
 	}
+	return nil, nil
+}
 
-	// If not found, default would be global
-	return c.globalClient.GenericClient(), c.globalClient.CredentialFrom(), nil
+func (c *armClientCache) checkPerResourceSecret(ctx context.Context, obj genruntime.ARMMetaObject) (*armClient, error) {
+	credentialFrom, ok := obj.GetAnnotations()[perResourceSecretAnnotation]
+	// annotation exists, use per Resource Group secret credential
+	if ok {
```
matthchr, line 126: IMO invert this check, it'll reduce the nesting in this method:

```
credentialFrom, ok := obj.GetAnnotations()[perResourceSecretAnnotation]
if !ok {
    return nil, nil
}

// Stuff here
```

v2/internal/reconcilers/arm/arm_client_cache.go
```diff
@@ -82,37 +85,96 @@ func (c *armClientCache) lookup(key string) (*armClient, bool) {
...
 		}
+		return armClient, nil
 	}
+	return nil, nil
+}
 
-	// If not found, default would be global
-	return c.globalClient.GenericClient(), c.globalClient.CredentialFrom(), nil
+func (c *armClientCache) checkPerResourceSecret(ctx context.Context, obj genruntime.ARMMetaObject) (*armClient, error) {
+	credentialFrom, ok := obj.GetAnnotations()[perResourceSecretAnnotation]
```
matthchr, line 124: minor: Maybe call this `credentialFromAnnotation` or something? Since in theory in the future it might be used for more than just perResourceSecret (if we end up doing perRg for example)

v2/internal/reconcilers/arm/arm_client_cache.go
```diff
@@ -9,6 +9,7 @@ import (
 	"context"
 	"net/http"
 	"reflect"
+	"strings"
 	"sync"
```
matthchr, line 14: What other stuff is remaining for this?

1. WorkloadIdentity/ManagedIdentity support
2. E2E tests?

Are you planning on doing those in separate PRs? It would be good to get E2E tests ASAP IMO.

v2/internal/reconcilers/arm/arm_client_cache.go
```diff
@@ -82,37 +85,96 @@ func (c *armClientCache) lookup(key string) (*armClient, bool) {
...
+		return armClient, nil
 	}
+	return nil, nil
+}
 
-	// If not found, default would be global
-	return c.globalClient.GenericClient(), c.globalClient.CredentialFrom(), nil
+func (c *armClientCache) checkPerResourceSecret(ctx context.Context, obj genruntime.ARMMetaObject) (*armClient, error) {
+	credentialFrom, ok := obj.GetAnnotations()[perResourceSecretAnnotation]
+	// annotation exists, use per Resource Group secret credential
```
matthchr, line 125: minor:
```suggestion
	// annotation exists, use specificed secret
```

As this doesn't have to be a resourceGroup?

v2/internal/reconcilers/arm/arm_client_cache.go
```diff
@@ -82,37 +85,96 @@ func (c *armClientCache) lookup(key string) (*armClient, bool) {
...
+	if ok {
+		secretNamespacedName := getSecretNameFromAnnotation(credentialFrom, obj.GetNamespace())
+
+		secret, err := c.getSecret(ctx, secretNamespacedName.Namespace, secretNamespacedName.Name)
+		if err != nil {
+			return nil, err
+		}
+		// secret not found
+		if secret == nil {
+			return nil, core.NewSecretNotFoundError(secretNamespacedName, errors.Errorf("Credential Secret %q not found for '%s/%s'", secretNamespacedName.String(), obj.GetNamespace(), obj.GetName()))
```
matthchr, line 135: very minor: errors aren't supposed to start with a capital letter (Go is weird).

```suggestion
			return nil, core.NewSecretNotFoundError(secretNamespacedName, errors.Errorf("credential Secret %q not found for '%s/%s'", secretNamespacedName.String(), obj.GetNamespace(), obj.GetName()))
```

matthchr: Actually wondering if it would be cleaner to change `getSecret` to return the `NotFound` error, rather than swallowing it and returning nil.

That way you can do this:
```
		secret, err := c.getSecret(ctx, secretNamespacedName.Namespace, secretNamespacedName.Name)
		if err != nil {
			if apierrors.IsNotFound(err) {
				return nil, core.NewSecretNotFoundError(secretNamespacedName, err)
			}
			return nil, err
		}
```

The other caller (namespacedName) could do:
```
	secret, err := c.getSecret(ctx, obj.GetNamespace(), namespacedSecretName)
	if err != nil {
		if apierrors.IsNotFound(err) {
			return nil, nil // Not finding this secret is allowed, allow caller to proceed to higher scope secret
		}
		return nil, err
	}
```

Which I think reduces some extra nesting and the fact that you have to reconstruct the secret error in the per-resource secret case.

v2/internal/reconcilers/arm/arm_client_cache.go
```diff
@@ -82,37 +85,96 @@ func (c *armClientCache) lookup(key string) (*armClient, bool) {
...
 	// Namespaced secret
-	secret, err := c.getSecret(ctx, obj.GetNamespace(), namespacedSecretName)
+	client, err = c.checkNamespacedSecret(ctx, obj)
 	if err != nil {
 		return nil, "", err
+	} else if client != nil {
+		return client.GenericClient(), client.CredentialFrom(), nil
+	}
+
+	// If not found, default would be global
```
matthchr, line 103: minor, phrasing:
```suggestion
	// If not found, default is the global client
```

v2/internal/reconcilers/arm/arm_client_cache.go
```diff
@@ -82,37 +85,96 @@ func (c *armClientCache) lookup(key string) (*armClient, bool) {
 
 // GetClient finds and returns a client and credential to be used for a given resource
 func (c *armClientCache) GetClient(ctx context.Context, obj genruntime.ARMMetaObject) (*genericarmclient.GenericClient, string, error) {
+
+	client, err := c.checkPerResourceSecret(ctx, obj)
```
matthchr, line 89: minor, naming: Would these `checkX` methods be more clearly named `getClientFromX` or maybe just `getXClient`?

So `getPerResourceClient`, `getNamespacedClient`?

Obviously very minor, ignore if you don't think that's clearer.

v2/internal/reconcilers/arm/arm_client_cache.go
```diff
@@ -82,37 +85,96 @@ func (c *armClientCache) lookup(key string) (*armClient, bool) {
...
+	if strings.Contains(credentialFrom, namespacedNameSeparator) {
+		slice := strings.Split(credentialFrom, namespacedNameSeparator)
+		return types.NamespacedName{Namespace: slice[0], Name: slice[1]}
+	} else {
+		// If namespace is not specified, look into the resource namespace
+		return types.NamespacedName{Namespace: resourceNamespace, Name: credentialFrom}
+	}
+}
+
+func (c *armClientCache) getARMClientFromSecret(secret *v1.Secret) (*armClient, error) {
```
matthchr, line 157: Nice, I like this method.

v2/internal/reconcilers/arm/arm_client_cache.go
```diff
@@ -82,37 +85,96 @@ func (c *armClientCache) lookup(key string) (*armClient, bool) {
...
+	nsName := types.NamespacedName{Namespace: secret.Namespace, Name: secret.Name}
+	client, ok := c.lookup(nsName.String())
+	// if client does not already exist, or secret data is changed, we need to get new token.
+	if !ok || !matchSecretData(client.secretData, secret.Data) {
```
matthchr, line 161: minor: You could also remove nesting here if you wanted via by swapping the order of these:

```
	client, ok := c.lookup(nsName.String())
	if ok && matchSecretData(client.secretData, secret.Data) {
		return client, nil
	}

	// Other stuff you had in the if

```

v2/internal/reconcilers/arm/arm_client_cache.go
super-harsh: and Secret watcher. 

Yep, for e2e tests we still need to discuss how to use this in `testContext`.

v2/internal/reconcilers/arm/arm_client_cache.go
```diff
@@ -82,65 +85,130 @@ func (c *armClientCache) lookup(key string) (*armClient, bool) {
...
+		return nil, nil
+	}
+
+	// annotation exists, use specified secret
+	secretNamespacedName := getSecretNameFromAnnotation(credentialFrom, obj.GetNamespace())
+
+	secret, err := c.getSecret(ctx, secretNamespacedName.Namespace, secretNamespacedName.Name)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, core.NewSecretNotFoundError(secretNamespacedName, errors.Errorf("credential Secret %q not found for '%s/%s'", secretNamespacedName.String(), obj.GetNamespace(), obj.GetName()))
```
matthchr, line 140: Don't need to do the `errors.Errorf("credential Secret %q not found for '%s/%s'", secretNamespacedName.String(), obj.GetNamespace(), obj.GetName())` anymore, as I think that's all already included in the err, so just do:

`core.NewSecretNotFoundError(secretNamespacedName, err)`. Or if you have to add some extra details, wrap the error so its context isn't lost w/ `errors.Wrapf`.

matthchr (approved):
Nice! Had 1 minor comment

v2/internal/reconcilers/arm/arm_client_cache.go
super-harsh: Thanks for mentioning. Have improved the other errors as well.

PR #2582, opened 2022-11-03
More customizable retry

### LINKED ISSUE #2556: Detect when parent resource has been created faster
We currently have a backoff of up to a minute for when resources are waiting for their owner to be created. This can cause some awkward pauses if a large graph of resources is being created, especially if one of the resources in that graph is slow to create such that all of the resources it owns have reached max backoff by the time it finishes.

Consider the graph
ResourceA: 10m to provision
  ResourceB: 1s to provision
    ResourceC: 1s to provision

In a perfect world, this would take ~10m 2s to provision. In reality today it takes:
10 + ~30s (average wait after 1m backoff) + ~30s (average wait after 1m backoff) = 11m to provision.

**Describe the improvement**
Child resources should either watch their parent resource for notifications when their specific parent completes provisioning, or we should have a more intelligent backoff that understands that the check "is my parent done" is cheap, and allows a faster retry for those resources. This could be accomplished by writing a custom wrapper around `controllers.NewRateLimiter` that reads our ready condition and chooses a backoff based on that.

I tend to favor the more intelligent RateLimiter, as doing this with eventing would drive quite a  bit of excess load and it's only used in the initial provisioning case after which point all of the events are useless. Plus the events we get out of the box with Kubebuilder are for changes, not "done".

### LINKED ISSUE #2575: 1m maximum backoff is too aggressive
Today we only back off up to 1m before retrying a failing request to Azure.

**Describe the improvement**
We should back off longer than 1m, but less than 1m for certain errors such as waiting for owner or waiting to resolve references. See #2556 which is related to this.

In an ideal world we could differentiate between expected errors and unexpected errors, and have expected errors take the fast-retry path (especially if they don't trigger requests to Azure) while unexpected errors take the slower path (especially if they do trigger requests to Azure).

matthchr:
Retry less aggressively for most Azure returned errors. Retry more aggressively for Kube-local errors such as WaitingForOwner.

Closes #2575.
Closes #2556

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/cmd/controller/main.go
+++ b/v2/cmd/controller/main.go
@@ -8,6 +8,7 @@ package main
 import (
 	"context"
 	"flag"
+	"math/rand"
 	"os"
 	"time"
 
@@ -29,7 +30,9 @@ import (
 	. "github.com/Azure/azure-service-operator/v2/internal/logging"
 	asometrics "github.com/Azure/azure-service-operator/v2/internal/metrics"
 	armreconciler "github.com/Azure/azure-service-operator/v2/internal/reconcilers/arm"
+	"github.com/Azure/azure-service-operator/v2/internal/util/interval"
 	"github.com/Azure/azure-service-operator/v2/internal/util/kubeclient"
+	"github.com/Azure/azure-service-operator/v2/internal/util/lockedrand"
 	"github.com/Azure/azure-service-operator/v2/internal/version"
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/conditions"
@@ -170,7 +173,18 @@ func makeControllerOptions(log logr.Logger, cfg config.Values) controllers.Optio
 				// TODO: do we need GVK here too?
 				return log.WithValues("namespace", req.Namespace, "name", req.Name)
 			},
+			// These rate limits are used for happy-path backoffs (for example polling async operation IDs for PUT/DELETE)
 			RateLimiter: controllers.NewRateLimiter(1*time.Second, 1*time.Minute),
 		},
+		RequeueIntervalCalculator: interval.NewCalculator(
+			// These rate limits are primarily for ReadyConditionImpactingError's
+			interval.CalculatorParameters{
+				//nolint:gosec // do not want cryptographic randomness here
+				Rand:              rand.New(lockedrand.NewSource(time.Now().UnixNano())),
+				ErrorBaseDelay:    1 * time.Second,
+				ErrorMaxFastDelay: 30 * time.Second,
+				ErrorMaxSlowDelay: 3 * time.Minute,
+				SyncPeriod:        cfg.SyncPeriod,
+			}),
 	}
 }
--- a/v2/internal/util/interval/interval_calculator.go
+++ b/v2/internal/util/interval/interval_calculator.go
@@ -1,0 +1,170 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package interval
+
+import (
+	"math"
+	"math/rand"
+	"sync"
+	"time"
+
+	"github.com/pkg/errors"
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	"github.com/Azure/azure-service-operator/v2/internal/util/kubeclient"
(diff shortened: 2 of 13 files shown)


v2/cmd/controller/main.go
```diff
@@ -170,7 +173,18 @@ func makeControllerOptions(log logr.Logger, cfg config.Values) controllers.Optio
...
 			},
+			// These rate limits are used for happy-path backoffs (for example polling async operation IDs for PUT/DELETE)
 			RateLimiter: controllers.NewRateLimiter(1*time.Second, 1*time.Minute),
 		},
+		RequeueIntervalCalculator: interval.NewCalculator(
+			// These rate limits are primarily for ReadyConditionImpactingError's
+			interval.CalculatorParameters{
+				//nolint:gosec // do not want cryptographic randomness here
+				Rand:         rand.New(lockedrand.NewSource(time.Now().UnixNano())),
+				BaseDelay:    1 * time.Second,
```
matthchr, line 184: Possibly these should be made configurable? At least the MaxFast/MaxSlow and maybe the 1m (which is sorta "MaxHappyPath"? Thoughts?

v2/cmd/controller/main.go
```diff
@@ -170,7 +173,18 @@ func makeControllerOptions(log logr.Logger, cfg config.Values) controllers.Optio
...
+				MaxFastDelay: 30 * time.Second,
+				MaxSlowDelay: 3 * time.Minute,
```
matthchr, line 186: Is 3m slow enough? We _could_ make this 5m but that's pretty darn long

v2/cmd/controller/main.go
theunrepentantgeek: There's merit in making them configurable, but not enough to block this PR. I'd suggest logging an issue to do that as a followup.

v2/cmd/controller/main.go
theunrepentantgeek: The Slow delay is used for Azure stuff, right? Some Azure things - like provisioning and deletion - can take a super long time.

That said, 3m = 20 requests per hour, and 5m = 12 requests per hour, so the difference isn't that great in the general scheme of things.

v2/internal/util/interval/interval_calculator.go
```diff
@@ -0,0 +1,170 @@
...
+	// potential drift from the state in Azure. Note that we cannot use mgr.Options.SyncPeriod for this because we filter
+	// our events by predicate.GenerationChangedPredicate and the generation will not have changed.
+	if i.syncPeriod != nil {
+		result.RequeueAfter = randextensions.Jitter(i.rand, *i.syncPeriod, 0.1)
+	}
+
+	return result
+}
+
+func (i *calculator) calculateExponentialDelay(base time.Duration, exp int, max time.Duration) time.Duration {
```
theunrepentantgeek, line 157: Shouldn't this have some jitter too?

theunrepentantgeek (approved):

v2/cmd/controller/main.go
matthchr: Will improve the comments here to clearly indicate. Really `MaxFast` and `MaxSlow` are for errors only. So when we're getting an error from Azure, we'll back off to one request every 3m on retries. That'd be the case for example for 5xx's or 4xx level errors that we're retrying on.

The happy path always takes the upper ratelimiter time (I hate that there are 2 but there have to be to make this work unfortunately). So when things are working correctly and we're polling an async operation, we'll poll every 1m. AFAIK, polling async operations even by the SDKs is done at a pretty snappy pace, so backing off to 3m there seems unnecessary. The compute API for example specifically says something along the lines of: "if you want to poll fast, poll the async operation ID, don't issue GETs")

matthchr: Changed these names to `ErrorMaxFastDelay` and `ErrorMaxSlowDelay` to clarify where they are actually used.

matthchr: Will leave this at 3 minutes for now.

v2/internal/util/interval/interval_calculator.go
matthchr: Maybe, but the existing interface which we're replacing (rate_limiter) didn't, so we're in no different a situation jitter-wise now than we were before.

v2/cmd/controller/main.go
matthchr: I filed #2584

PR #2596, opened 2022-11-18
High level design for Resource Import tool

theunrepentantgeek:

Captures current thinking about resource importing and how that would address the needs of ASO v1 users wanting to migrate to using ASO v2.

gif


docs/hugo/content/design/ADR-2022-11-Resource-Import.md
```diff
@@ -0,0 +1,78 @@
...
+The most likely form of this tool would be as a command-line utility for users to run locally. Idiomatically for the Kubernetes ecosystem, we'd call this `asoctl`.
+
+* Pro: Straightforward migration for ASO v1 customers (though, not zero touch).
+* Pro: Would benefit all customers wanting to migrate existing Azure resources, not just those using ASO v1.
+* Pro: Could also be used to snapshot existing resources (e.g. to capture a hand configured resource as YAML for reuse).
+* Con: We'd need to extend the code generator to provide the required support (but much of what we need already exists).
+
+## Decision
+
+Create a new command line tool, `asoctl` to house multiple functions, starting with resource import. The tool will follow the usual convention of supporting a variety of verbs, each activating a different mode.
```
matthchr, line 36: I do still think we should include the "do it with a CRD" option and pros/cons for why we decided not to do it.

We can't just blindly apply the YAML we import as there are things we need the user to provide us (secrets being the main one we can't do without). There are many resources which don't have secrets though (or have secrets but which can be used w/ AAD authentication modes where they don't have secrets) and for those resources a direct import would work.

Amazon's offering does it this way: [ACK Adoption](https://aws-controllers-k8s.github.io/community/docs/user-docs/adopted-resource/)

Google's offering does it with a cmdline tool like you've proposed here: [ConfigConnector Export](https://cloud.google.com/config-connector/docs/how-to/import-export/export)

There are pros and cons to both so I'm not saying we should use the CRD approach but I think we should document it and say why we didn't.

The arguments I see are:
Pro: You don't need to use an `asoctl` tool, which means you can import/export using just `kubectl`.
Con: Awkward to model a run-once action as a CRD as there is no reconcile to be done after the first pass. There are examples of such resources in core k8s API such as `Job` and a pattern we could follow
Con: Would need to automatically mark resources as `skip` reconcile for safety, but then users need to go remove that which takes more work.

There may be other pros/cons, but given at least one of our competitors does it that way it seems like we should describe why we don't.

docs/hugo/content/design/ADR-2022-11-Resource-Import.md
```diff
@@ -0,0 +1,78 @@
...
+For example, to export the YAML for an existing Virtual Network, the user would run:
+
+``` bash
+$ asoctl export resource http://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/rg1/providers/Microsoft.Network/virtualNetworks/vnet1
+```
+
+As an extension, we could also support `export resourcegroup` to generate a single YAML file containing all the resources in a resource group. This would be useful for users who want to snapshot an existing resource group as YAML for reuse.
+
+``` bash
+$ asoctl export resourcegroup http://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/rg1
```
matthchr, line 49: Rather than limit this to `resourcegroup` it feels like we should just call it `export-all` or something, and it works by exporting the resource in question and any subresources (as per Azure's ownership rules which we know).

That way if you wanted to for example export a Kubernetes cluster and all of its AgentPools, you just `export-all` the cluster. Or a MySQL Server, its firewall, and all of the firewall rules, you can just `export-all` the `MySQLServer`

docs/hugo/content/design/images/adr-2022-11-import-flow.dot
```diff
@@ -0,0 +1,22 @@
...
+    rsrc_yaml [label="«yaml»\nResource"]
+
+    edge [arrowhead=vee, arrowtail=vee, dir=forward]
+
+    arm -- rsrc_arm  [label="GET", arrowhead=vee]
+    rsrc_arm -- rsrc_status [label="Conversion", arrowhead=vee]
+    rsrc_spec -- rsrc_yaml [label="Export", arrowhead=vee]
+    
+    edge [style=dashed, color=gray]
+    rsrc_status -- rsrc_spec [label="Update", arrowhead=vee]
```
matthchr, line 21: minor: Is this really an update? It also feels like a conversion to me?

docs/hugo/content/design/ADR-2022-11-Resource-Import.md
super-harsh: also maybe `export subscription` or something else to export all resources under a subscription?

docs/hugo/content/design/ADR-2022-11-Resource-Import.md
```diff
@@ -0,0 +1,78 @@
...
+```
+### Data Flow
+
+The process of generating the YAML for a given resource can be simplified to the following flow:
+
+![Data Flow](./images/adr-2022-11-import-flow.png)
+
+1. The user specified URL is used to ***GET*** the resource from Azure in its native ARM format.
+2. We applly a ***conversion*** this to a Status object for the relevant ASO custom resource.
```
super-harsh, line 58:
```suggestion
2. We apply a ***conversion*** this to a Status object for the relevant ASO custom resource.
```

docs/hugo/content/design/ADR-2022-11-Resource-Import.md
theunrepentantgeek: Done.

docs/hugo/content/design/ADR-2022-11-Resource-Import.md
theunrepentantgeek: My initial instinct is that `export subscription` would be too wide a net, though it would be an easy extension if someone asked for it.

theunrepentantgeek: Actually, I don't think we need a separate verb - we can make the decision based on what the URL points to.

docs/hugo/content/design/images/adr-2022-11-import-flow.dot
theunrepentantgeek: I've gone with _Populate_, let me know what you think.

super-harsh (approved):

docs/hugo/content/design/ADR-2022-11-Resource-Import.md
matthchr: > Given ResourceGroup doesn't have a collection within, I thought it needed special treatment.

Not 100% sure I follow this? This is true for most resources which have subresources, there's no way to know if those subresources exist unless you:
1. Know about their types explicitly, and know the specific URL used to list them and call that.
2. You call a generic List Resources API with the resource as the prefix.

We know the full hierarchy of the types we support (that's the ownership stuff), _and_ we also have access to the generic List Resources API via the Go SDK. So either approach should allow us to treat ResourceGroups the same as other resources, I think?

theunrepentantgeek: You may be right - I'll update the design to reflect this.

PR #2608, opened 2022-11-22
Prune embedded resources in Swagger

### LINKED ISSUE #2605: Bug: Unexplained missing properties
**Describe the bug**

Comparing resource structures between main and the Swagger branch in #2323, identified a few properties that are missing without full explanation.

For example, on `main` we have:

```
github.com/Azure/azure-service-operator/v2/api/compute/v1beta20201201
├── VirtualMachine: Resource
│   ├── Spec: Object (26 properties)
│   │   ├── ...
│   └── Status: Object (32 properties)
│       ├── Resources: Object (15 properties)
│       │   ├── AutoUpgradeMinorVersion: *bool
│       │   ├── EnableAutomaticUpgrade: *bool
│       │   ├── ForceUpdateTag: *string
│       │   ├── Id: *string
│       │   ├── InstanceView: Object (5 properties)
│       │   │   ├── Name: *string
│       │   │   ├── Statuses: Object (5 properties)
│       │   │   │   ├── Code: *string
│       │   │   │   ├── DisplayStatus: *string
│       │   │   │   ├── Level: Enum (3 values)
│       │   │   │   │   ├── "Error"
│       │   │   │   │   ├── "Info"
│       │   │   │   │   └── "Warning"
│       │   │   │   ├── Message: *string
│       │   │   │   └── Time: *string
│       │   │   ├── Substatuses: Object (5 properties)
│       │   │   │   ├── Code: *string
│       │   │   │   ├── DisplayStatus: *string
│       │   │   │   ├── Level: Enum (3 values)
│       │   │   │   │   ├── "Error"
│       │   │   │   │   ├── "Info"
│       │   │   │   │   └── "Warning"
│       │   │   │   ├── Message: *string
│       │   │   │   └── Time: *string
│       │   │   ├── Type: *string
│       │   │   └── TypeHandlerVersion: *string
│       │   ├── Location: *string
│       │   ├── Name: *string
│       │   ├── PropertiesType: *string
│       │   ├── ProtectedSettings: map[string]v1.JSON
│       │   ├── ProvisioningState: *string
│       │   ├── Publisher: *string
│       │   ├── Settings: map[string]v1.JSON
│       │   ├── Tags: map[string]string
│       │   ├── Type: *string
│       │   └── TypeHandlerVersion: *string
```

The property `VirtualMachine.Status.Resources` is completely missing from #2323

Similarly, we also have 

```
github.com/Azure/azure-service-operator/v2/api/compute/v1beta20201201
├── VirtualMachineScaleSet: Resource
│   ├── Spec: Object (22 properties)
│   │   ├── ...
│   │   ├── VirtualMachineProfile: Object (11 properties)
│   │   │   ├── ...
│   │   │   ├── ExtensionProfile: Object (2 properties)
│   │   │   │   ├── Extensions: Object (5 properties)
│   │   │   │   │   ├── Name: *string
│   │   │   │   │   ├── Publisher: Validated<*string> (1 rule)
│   │   │   │   │   │   └── Rule 0: {MinLength %!s(int64=1)}
│   │   │   │   │   ├── Settings: map[string]v1.JSON
│   │   │   │   │   ├── Type: Validated<*string> (1 rule)
│   │   │   │   │   │   └── Rule 0: {MinLength %!s(int64=1)}
│   │   │   │   │   └── TypeHandlerVersion: Validated<*string> (1 rule)
│   │   │   │   │       └── Rule 0: {MinLength %!s(int64=1)}
│   │   │   │   └── ExtensionsTimeBudget: *string
│   │   │   ├── LicenseType: *string
```

where `VirtualMachineScaleSet.Spec.VirtualMachineProfile.ExtensionProfile.Extensions` is missing from #2323.

matthchr:
* Detect and prune resources embedded inside other resources.
* This was previously done with a different approach based on the JSON schemas, but needed to be rewritten to work with Swagger as a single source of truth.

Resolves #2605


matthchr:
Notes for reviewers:

**Check out the various structure.txt diffs with this in mind:**
* Batch
  * Removed PrivateEndpointConnections in status - This is correct as PrivateEndpointConnections are real resources and should be examined there.
* Compute 2020
  * Brought ExtensionProfile back, which we have in main
* Compute 2022
  * Brought ExtensionProfile back, which we have in main
* ContainerRegistry
  * PrivateEndpointConnections from Status was removed, but I think that's correct (see above)
* EventHub
  * Also lost PrivateEndpointConnections, but again I think correct
* MachineLearningServices
  * Also lost PrivateEndpointConnections, but again I think correct
* ServiceBus
  * Also lost PrivateEndpointConnections
* Signalrservice
  * Also lost PrivateEndpointConnections

The biggest offender by far was networking, which I did a detailed diff through structure.txt compared with main 3-4 times to make sure things look good. There are specific notes in `azure-arm.yaml` for various properties and why they are/aren't pruned. I would recommend not bothering to look at the diff of networking in this PR as there was _so much_ stuff that shouldn't have been there. This PR gets it way closer to correct (I think it's 100% correct but we can do another pass to confirm).

**When reviewing remover.go, it might be helpful to look at it as all new code, rather than a diff. While the basic structure is the same as before all of the guts have been rewritten which makes the diff hard to read. You can probably discount the old code and just comment on the new code.**

PR #2612, opened 2022-11-28
Add workload identity support for single operator multitenancy

super-harsh:
**If applicable**:


v2/internal/reconcilers/arm/arm_client_cache.go
```diff
@@ -242,3 +259,38 @@ func (c *ARMClientCache) getSecret(ctx context.Context, namespace string, secret
...
+	w.cred = cred
+	return w, nil
+}
+
+func (w *workloadIdentityCredential) GetToken(ctx context.Context, opts policy.TokenRequestOptions) (azcore.AccessToken, error) {
+	return w.cred.GetToken(ctx, opts)
+}
+
+func (w *workloadIdentityCredential) getAssertion(context.Context) (string, error) {
+	if now := time.Now(); w.lastRead.Add(3600 * time.Second).Before(now) {
```
matthchr, line 287: 3600 is too long I think? The way this works (AFAIK) is that it re-reads the token periodically. We need to pick a time which is significantly faster than the underlying token expiry? The example written by @chlowell had 5m here but possibly that was just a dummy value.

Can you confirm with him what we should be setting this value to? I can't actually tell if/when the `AssertionCallback` is called from a quick read of their code. I see it registered but not used in the `GetToken` call. Apologies if you've already figured that out and 3600 is in fact correct.

Either way, once we have an answer maybe leave a comment here (or by the `3600` const I mentioned above) that explains why it's 3600 (or whatever the other value is).

v2/internal/reconcilers/arm/arm_client_cache.go
```diff
@@ -191,35 +194,49 @@ func (c *ARMClientCache) getARMClientFromSecret(secret *v1.Secret) (*armClient,
...
 		errs = append(errs, err)
 	}
 	clientSecret, ok := secret.Data[config.AzureClientSecretVar]
-	// TODO: We check this for now until we support Workload Identity. When we support workload identity for multitenancy, !ok would mean to check for workload identity.
 	if !ok {
-		err := core.NewSecretNotFoundError(nsName, errors.Errorf("credential Secret %q does not contain key %q", nsName.String(), config.AzureClientSecretVar))
-		errs = append(errs, err)
+		// Here we check for workload identity if client secret is not provided, AZURE_FEDERATED_TOKEN_FILE env var gets populated if workload identity sidecar is injected to the pod,
+		// so we can check if its populated, which means workload identity is enabled.
+		if tokenFile := os.Getenv(config.AzureFederatedTokenFileVar); tokenFile != "" {
```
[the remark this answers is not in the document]
matthchr: As discussed on Teams I think this makes sense. If we include a token like so in our container (possibly with a different path so it doesn't collide with the WI one if that's enabled):
```
    - name: azure-identity-token
      projected:
        defaultMode: 420
        sources:
        - serviceAccountToken:
            audience: api://AzureADTokenExchange
            expirationSeconds: 3600
            path: azure-identity-token
```

Then we don't actually need to user to install workload identity for WI to work. We also don't need them to have installed ASO using WI, we can instead just default to trying this if the `clientSecret` isn't set (same way you're doing now). If we do this, you don't need to add the note I mentioned above about: "Possibly worth a bold warning about how multitenant secrets won't work unless the main operator is also deployed using workload identity?"

matthchr (approved):
Approving this as I will be on vacation starting next week, but please take a look at comments before merging.

v2/internal/reconcilers/arm/arm_client_cache.go
chlowell: The 5 minute assertion cache was the feature team's recommendation when we first designed this last year. @aramase, is that still the best approach?

docs/hugo/content/introduction/multitenant-deployment.md
```diff
@@ -37,6 +37,30 @@ To deploy the operator in single-operator multi-tenant mode:
...
 
+### Deployment using Workload Identity
+
+To deploy the operator in single-operator multi-tenant mode:
+
+1. Follow the normal ASO [workload identity installation](https://azure.github.io/azure-service-operator/introduction/authentication/#azure-workload-identity)
+2. To use namespace scoped credential, create a credential secret named `aso-credential` in the desired namespace. Using this, all the resources in that namespace will use namespace scoped credential.
+3. To use per-resource credential, create a credential secret and add an annotation to the resource like `serviceoperator.azure.com/credential-from: <SECRET_NAMESPACE>/<SECRET_NAME>`
+
+**Note:** Each credential(namespaced and per-resource credential) created further, should have established trust between your OIDC issuer URL and the backing Service Principal or Managed Identity. See [how to configure trust](https://azure.github.io/azure-service-operator/introduction/authentication/#configure-trust) for more details.
```
theunrepentantgeek, line 48: Suggested wording change.

```suggestion
**Note:** Each credential (both namespaced and per-resource) you create must have a trust relationship between your OIDC issuer URL and the backing Service Principal or Managed Identity. See [how to configure trust](https://azure.github.io/azure-service-operator/introduction/authentication/#configure-trust) for more details.
```

v2/internal/reconcilers/arm/arm_client_cache.go
```diff
@@ -191,35 +194,49 @@ func (c *ARMClientCache) getARMClientFromSecret(secret *v1.Secret) (*armClient,
 
 func (c *ARMClientCache) newCredentialFromSecret(secret *v1.Secret, nsName types.NamespacedName) (azcore.TokenCredential, string, error) {
 	var errs []error
+	var err error
+	var credential azcore.TokenCredential
+
 	subscriptionID, ok := secret.Data[config.SubscriptionIDVar]
 	if !ok {
-		err := core.NewSecretNotFoundError(nsName, errors.Errorf("credential Secret %q does not contain key %q", nsName.String(), config.SubscriptionIDVar))
+		err = core.NewSecretNotFoundError(nsName, errors.Errorf("credential Secret %q does not contain key %q", nsName.String(), config.SubscriptionIDVar))
```
theunrepentantgeek, line 202: Why switch to a shared `err` variable - these all look to be isolated cases where having independently decared single-write variables would make sense.

v2/internal/reconcilers/arm/arm_client_cache.go
```diff
@@ -242,3 +259,38 @@ func (c *ARMClientCache) getSecret(ctx context.Context, namespace string, secret
 func matchSecretData(old, new map[string][]byte) bool {
 	return reflect.DeepEqual(old, new)
 }
+
+// TODO: Not sure if its a right thing to put this here or create a separate package for identity/auth and put workloadIdentityCredential there?
+// The below bit is from https://github.com/Azure/azure-sdk-for-go/issues/15615#issuecomment-1211012677
```
theunrepentantgeek, line 264: I'd lean towards putting it into a separate package as you suggested.

theunrepentantgeek (approved):
Some documentation suggestions. Looks good. Once we've addressed the comments from @matthchr, should be good to merge.

v2/internal/reconcilers/arm/arm_client_cache.go
super-harsh: Yep, I'll keep this one as it is for now. And since I'm gonna work more on workload identity stuff which Matt and I had a conversation about(Matt has mentioned above). Basically, the global credential code in `main.go` would be using this bit as well so will move it in next PR.

v2/internal/reconcilers/arm/arm_client_cache.go
super-harsh: Linter complained about `shadow declaration`

v2/internal/reconcilers/arm/arm_client_cache.go
aramase: The default token expiry configured with Azure Workload Identity is `1h` and the max allowed value is `24h`. The `kubelet` process renews the token at 80% expiry (so every 40 mins). With those numbers, every 30 min would work fine.

v2/internal/reconcilers/arm/arm_client_cache.go
```diff
@@ -242,3 +257,38 @@ func (c *ARMClientCache) getSecret(ctx context.Context, namespace string, secret
...
+	w.cred = cred
+	return w, nil
+}
+
+func (w *workloadIdentityCredential) GetToken(ctx context.Context, opts policy.TokenRequestOptions) (azcore.AccessToken, error) {
+	return w.cred.GetToken(ctx, opts)
+}
+
+func (w *workloadIdentityCredential) getAssertion(context.Context) (string, error) {
+	if now := time.Now(); w.lastRead.Add(5 * time.Minute).Before(now) {
```
chlowell, line 285: Note that there's a data race here, so you may want to add synchronization, if only to avoid setting off the race detector.

theunrepentantgeek (approved):

PR #2623, opened 2022-12-01
Address Lint issues

theunrepentantgeek:

An update to the linting tools has made them overtly hostile to much of our documentation.

* Comments must now have exactly one space between the `//` and any content, so indenting paragraphs or lists is now impossible
* Any line containing a colon (`:`) is now presumed to be a directive and is flagged if it doesn't conform

gif

**If applicable**:


v2/tools/generator/internal/codegen/storage/property_converter.go
```diff
@@ -79,6 +79,7 @@ func (p *PropertyConverter) useBaseTypeForEnumerations(
 	return tv.Visit(et.BaseType(), ctx)
 }
 
+//nolint:gofmt
```
matthchr, line 82: Curious why not change this one too? To a single space + `*`?

v2/tools/generator/internal/astbuilder/maps.go
```diff
@@ -12,10 +12,13 @@ import (
 )
 
 // MakeMap returns the call expression for making a map
-//
-// 	make(map[<key>]<value>)
-//
 func MakeMap(key dst.Expr, value dst.Expr) *dst.CallExpr {
+	/*
```
matthchr, line 16: The gofmt documentation suggests that it supports code blocks, as long as they're formatted "correctly": See https://tip.golang.org/doc/comment - section about "code blocks" (can't link to it for some reason).

For example check out: https://go.dev/play/p/zPOvetWAPA1 - if you click format on that, it adds a space between the top level method text and the code block, but leaves it alone other than that. So I think all of these places where we've moved the code block into the method can be done in that style and kept in the comment (which IMO looks better?)

I think you can get `go fmt` to do that automatically by just running the latest version of `go fmt` on the original set of files, at least for these `astbuilder` files where the issue was code block formatting.

v2/tools/generator/internal/conversions/property_conversions.go
```diff
@@ -101,60 +103,62 @@ func init() {
 //
 // The method works by considering the conversion requested by sourceEndpoint & destinationEndpoint,
 // with recursive calls breaking the conversion down into multiple steps that are then combined.
-//
```
matthchr, line 104: It might be possible to make this comment work w/ the code-block style formatting as well, see https://go.dev/play/p/AGebhZXwX7-, which passes `gofmt` so I think should pass the linter too?

v2/tools/generator/internal/conversions/property_conversions.go
```diff
@@ -192,31 +196,34 @@ func NameOfPropertyAssignmentFunction(
 }
 
 // writeToBagItem will generate a conversion where the destination is in our property bag
```
matthchr, line 198: Same here, see https://go.dev/play/p/_q-RST9zBFD, which passes `go fmt`

matthchr: I think this applies to most of the stuff below as well.

theunrepentantgeek:
Superseeded by #2629

PR #2629, opened 2022-12-01
Fix Linting Issues (redux)

theunrepentantgeek:

An update to the linting tools has made them overtly hostile to much of our documentation, due to the `gofmt` linter always using the latest version of the rules _despite_ the version of Go being used for the project.

Upgrading to `Go 1.19` means that `go fmt` is now able to automatically fix these problems.

gif

**If applicable**:


v2/tools/generator/internal/functions/resource_conversion_function.go
```diff
@@ -20,31 +20,30 @@ import (
...
+//	func (<receiver> <receiverType>) Convert<From|To>(hub conversion.Hub) error {
+//		source, ok := hub.(*<hub.Type>)
+//		if !ok {
+//			return fmt.Errorf("expected <hub.Type> but received %T instead", hub)
+//		}
+//		return <receiver>.AssignProperties<From|To><Type>(source)
+//	}
 //
-// Indirect conversion, multiple steps via an intermediate instance
+// # Indirect conversion, multiple steps via an intermediate instance
```
matthchr, line 31: I think you can remove the added `#` here and make it not have that. I had that show up in some go playground tests but fussing with the comment a bit got it to go away.

Feels like a gofmt bug to me...

matthchr (approved):
One small comment, LGTM otherwise

v2/tools/generator/internal/functions/resource_conversion_function.go
theunrepentantgeek: Fixed.

PR #2638, opened 2022-12-03
Fix ARM ID case miss

### LINKED ISSUE #2501: Incorrectly named cross resource references
We've mistakenly missed a few properties which should have been marked as ARM ID references. We instead left them as strings. The fix for this is easy, but it would be breaking... We need to decide what we can do here to evolve in a positive direction, either by breaking prior to going GA or fixing this in a newer version of the service.

Can look through the repo for `TODO[breaking]` to find some examples of this.

matthchr:
This is nonbreaking and will put us in a good position going forward. I will deal with the transformations required as part of the new ManagedCluster version.

This fixes #2501


theunrepentantgeek:
After offline discussion, we concluded that we do want this - but not as a breaking change.

Scheduling for `beta.5` so that we can do the appropriate design to make it so.

Issue #2621, opened 2022-11-30
Improve support for new api versions in samples testing


super-harsh:
**Describe the current behavior**

Currently, we have a structure where we only have `v1alpha1api` and `v1beta` folders for samples which is not very helpful when we're adding new API versions for an existing resource. 

**Describe the improvement**

As we discussed, we need to have separate folders for new API versions for an existing resource. As a result of this, we can will have a structure like below: 

- v1alpha1api
- v1beta20210601
- v1beta20220301

**Additional context**
As a part of this, we'll have to make sure all the documentation for samples have correct links to the samples and make sure sample checker script checks the sample in right folders.

PR #2642, opened 2022-12-05
Add support for Dbforpostgresql/FlexibleServer/20220120preview

### LINKED ISSUE #2579: aso v2.0.0 doesn't support 'Some zone' High availability mode for postgresql flexible server
we are testing postgresql flexible server in azure China north3 region using aso v2.0.0-beta.2 version, this new region currently only supports 'Same zone' HA, and 'Zone redundant' feature is not ready in azure china. I want to enable 'Same zone' HA setting via aso, currently this setting can be updated in azure portal, but it seems aso v2.0.0-beta.2 and v2.0.0-beta.3 still don't support 'Same zone' HA for postgresql flexible server. from file apiextensions.k8s.io_v1_customresourcedefinition_flexibleservers.dbforpostgresql.azure.com.yaml I cannot find keyword 'SameZone', but from apiextensions.k8s.io_v1_customresourcedefinition_flexibleservers.dbformysql.azure.com.yaml file I can find highAvailability.mode has enum value SameZone, see screenshot attached.
aso1
aso2
aso3

super-harsh:
Closes #2579 & #2621


v2/azure-arm.yaml
```diff
@@ -1016,6 +1021,29 @@ objectModelConfiguration:
       FlexibleServers_FirewallRule:
         $exportAs: FlexibleServersFirewallRule
         $supportedFrom: v2.0.0-alpha.2
+    2022-01-20-preview:
+      FlexibleServer:
+        $export: true
+        $supportedFrom: v2.0.0-beta.4
+        $azureGeneratedSecrets:
+          - FullyQualifiedDomainName
```
theunrepentantgeek, line 1029: Is this really a secret? Shouldn't this be exported as a config map instead?

super-harsh: I matched it in from `2021-06-01`. Although, it should be a configmap. I'm thinking, we have other resources(dbformariadb, etc) as well for which we're doing the same. So maybe all at once, I can change them to be exported to a configmap?

theunrepentantgeek: As discussed, offline, we have a number of older resources that publish non-sensitive information as secrets because that was the only mechanism we had at the time. We don't want to modify those, because it would be a needless breaking change.

New resources (including new versions of existing resources) should do things properly:
* Sensitive information (e.g. api keys, Azure generated passwords, etc) should be made available to secrets.
* All other information (e.g. domain names etc) should be made available as config maps.

theunrepentantgeek (approved):

PR #2649, opened 2022-12-07
Design for a new extension point to allow customization of when a resource reconciles

theunrepentantgeek:

There are some Azure resources where long running operations preclude any PUT from succeeding, but currently ASO tries anyway. When ASO encounters the error, it uselessly retries over and over again when it can't possibly succeed. This causes excess load (up to and including ARM throttle triggering).

By creating a new extension point, we can allow customization of behaviour for specific resources as they are identified.

gif

**If applicable**:


docs/hugo/content/design/ADR-2022-12-Reconciliation-Extensions.md
```diff
@@ -0,0 +1,65 @@
...
+
+We'll define a new extension point called `ReconcileActionSelector` in a style similar to our already existing extension points.
+
+The extension point will receive the following parameters:
+
+* The current resource, with a status freshly updated from Azure.
+* A client allowing for ARM operations.
+* A logger to allow for tracing of what the extension did.
+* A context to allow cancellation of long running operations.
+* A function that will return the next action to take.
```
super-harsh, line 38: maybe, do you think an `Event` should also join the list?

super-harsh (approved):
LGTM! Just left a minor suggestion.

docs/hugo/content/design/ADR-2022-12-Reconciliation-Extensions.md
theunrepentantgeek: What would the event signal?

super-harsh: A k8s events.. so when a user `k describe` a resource, they can see whether ASO issued a `GET` or a `PUT`.

theunrepentantgeek: I don't think this is the place for that. 

This is the parameter list for the extension - and for it to work, it has to have the latest information, so at that point we'll always have just done a GET.

PR #2666, opened 2022-12-21
Setup asoctl infrastructure

### LINKED ISSUE #2613: Create shell for `asoctl`
We have need for a command line tool that will have multiple functions - see  1238, #1519, 2119 #2172, #2596

Since there's a bit of infrastructure to set up, it makes sense to break out this foundational work as a separate task so that the overhead needn't be borne by the first feature off the start line.

Source code for `asoctl` should live in a folder of the same name under `v2/tools`.

Startup should mirror the way `generator` has a `main.go` that in turn delegates to `root.go`, and allow for extension the same way that `generator` can be extended with extra commands.

We need new targets to `Taskfile.yaml` to allow for building and CI integration. The list of required targets will parallel those for `generator`

* `generator:quick-checks`
* `generator:ci`
* `generator:unit-tests`
* `generator:unit-tests-cover`
* `generator:lint`
* `generator:lint-full`
* `generator:build`

super-harsh:
Closes #2613 

This PR includes infrastructure setup for asoctl


Taskfile.yml
```diff
@@ -100,10 +103,61 @@ tasks:
...
+    desc: Run {{.ASOCTL_APP}} unit tests and output coverage.
+    dir: '{{.ASOCTL_ROOT}}'
+    cmds:
+      - defer: # want to run even on failure
+          task: produce-markdown-summary
+          vars:
+            INPUT_FILE: '{{.TEST_OUT}}/asoctl-unit-tests.json'
+            OUTPUT_FILE: '{{.TEST_OUT}}/asoctl-unit-tests.md'
+      - go test ./... -tags=noexit -race -covermode atomic -coverprofile=asoctl-coverage.out -json -coverpkg=./... -run '{{default ".*" .TEST_FILTER}}' > '{{.TEST_OUT}}/asoctl-unit-tests.json'
+
```
matthchr, line 139: minor: extra newline here

Taskfile.yml
```diff
@@ -100,10 +103,61 @@ tasks:
...
 
+  ############### Asoctl targets ###############
+
+  asoctl:quick-checks:
+    cmds:
+      - task: asoctl:unit-tests
+      - task: basic-checks
+      - task: asoctl:lint
+
+  asoctl:ci:
```
matthchr, line 117: Should call this from the `ci` target, so that it runs as part of CI.

Taskfile.yml
```diff
@@ -100,10 +103,61 @@ tasks:
...
+      - go test ./... -tags=noexit -race -covermode atomic -coverprofile=asoctl-coverage.out -json -coverpkg=./... -run '{{default ".*" .TEST_FILTER}}' > '{{.TEST_OUT}}/asoctl-unit-tests.json'
+
+
+  asoctl:lint:
+    desc: Run {{.ASOCTL_APP}} fast lint checks.
+    dir: '{{.ASOCTL_ROOT}}'
+    cmds:
+      - golangci-lint run --verbose
+
+  asoctl:lint-full:
```
matthchr, line 147: I know you followed the pattern here, but I'm thinking we actually should just remove all the `lint-full`'s and make `lint` use `--fast=false` everywhere. Otherwise CI and quick-checks might run different validations which feels awkward (passes locally, fails in CI).

I can fix that once this merges.

matthchr (approved):
LGTM. Left a few minor build comments. 

Didn't review the code _that_ carefully as it's obviously just scaffolding and will get added to significantly.

Issue #2172, opened 2022-03-18
Support deleting deprecated CRD versions


matthchr:
Kubernetes doesn't have great support for removing unsupported CRD versions using `kubectl`. 

A vanilla `kubectl apply -f` of a CRD that no longer contains a version that it once contained results in:
>The CustomResourceDefinition "resourcegroups.resources.azure.com" is invalid: status.storedVersions[0]: Invalid value: "v1alpha1api20200601": must appear in spec.versions

See https://github.com/kubernetes/website/issues/18248 which discusses this. It's not clear based on that issue if `k edit` is _supposed_ to work or not. That issue refers to https://github.com/elastic/cloud-on-k8s/issues/2196 which has a golang script for doing this. I did try to use `k edit` and it still doesn't seem to work.

There are some operators that do this though, for example see https://github.com/cert-manager/cert-manager/releases/tag/v1.7.0, which deleted the old alpha CRDs and provides a tool to help clean up `storedVersions`

matthchr:
See https://github.com/kubernetes-sigs/kube-storage-version-migrator as well

PR #2678, opened 2023-01-12
Add support for deleting deprecated CRD versions

### LINKED ISSUE #2677: Update asoctl project structure
**Describe the improvement**

Proposed structure: 

```
├── cmd
│   ├── main.go
│   └── root.go
├── internal
│   ├── client
│   │   └── client.go
│   └── crd
│       └── cleaner.go
└── pkg
    ├── crd
    │   ├── crd.go
    │   └── crd_clean.go
    └── export
        ├── export.go
        └── export_resource.go
├── go.mod
├── go.sum
```

super-harsh:
Closes #2677 & #2172 

This PR adds asoctl command `crd clean` and updated asoctl dir structure. 

The approach adopted in `crd clean` is to find and remove `v1alpha1api` storedVersion from the installed CRDs and update the CRD with the new version set.


theunrepentantgeek (commented):
A great starting point; have made some suggestions.

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,81 @@
...
+
+	if list.Items == nil || len(list.Items) == 0 {
+		fmt.Println("found 0 results, make sure you have ASO CRDs installed")
+	}
+
+	var updated bool
+	for _, item := range list.Items {
+		crdName := item.Name
+		crdRegexp := regexp.MustCompile(".*\\.azure\\.com")
+		deprecatedVersionRegexp := regexp.MustCompile("v1alpha1api.{8}.*")
```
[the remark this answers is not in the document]
super-harsh: @matthchr and I had a conversation on this already. We don't want the end user to specify what they want to delete/deprecate(which can break their cluster). The tool we're providing is just to remove `storedVersion` for the deprecated CRDs which we specify in the tool. 

For sure we want a workaround to solve the problem dynamically where in the future we'll be deprecating `v1beta` which'll be a few versions after we go GA(I assume).

v2/tools/asoctl/cmd/root.go
```diff
@@ -32,7 +35,8 @@ func newRootCommand() (*cobra.Command, error) {
 	rootCmd.Flags().SortFlags = false
 
 	cmdFuncs := []func() (*cobra.Command, error){
-		NewExportCommand,
+		export.NewExportCommand,
```
matthchr, line 38: minor: Given these are in packages we probably could just call it `NewCommand`? If we've got a package per top-level command anyway?

Then we get:
```
export.NewCommand(),
crd.NewCommand()
```

which maybe reads 25% prettier?

v2/tools/asoctl/internal/client/client.go
```diff
@@ -0,0 +1,22 @@
...
+
+package client
+
+import (
+	"github.com/pkg/errors"
+	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+func NewClient() (*v1.ApiextensionsV1Client, error) {
```
matthchr, line 14: I'm not totally sold that we actually need this helper, since:
1. It seems a pretty thin wrapper for the already existing `v1.NewForConfig(cfg)`
2. If we had multiple different types of clients we needed to construct, it probably would make sense to just call `config.GetConfigOrDie` once and then pass it to each of the `v1.NewForConfig` equivalents. Otherwise we're doing excess file reads. If we made that change, we're now _exactly_ wrapping `v1.NewForConfig` which isn't super useful?

So IMO just remove this and do it in the `crd_clean.go` file directly.

v2/tools/asoctl/pkg/export/export_resource.go
```diff
@@ -3,31 +3,13 @@
...
-	exportResourceCommand, err := newExportResourceCommand()
-	if err != nil {
-		return nil, err
-	}
-	cmd.AddCommand(exportResourceCommand)
-
-	return cmd, nil
-}
-
 func newExportResourceCommand() (*cobra.Command, error) { //nolint:unparam // TODO: Remove this comment when the tool is actually functional
```
matthchr, line 13: Same as above

v2/tools/asoctl/pkg/crd/crd_clean.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"context"
+
+	"github.com/spf13/cobra"
+
+	"github.com/Azure/azure-service-operator/v2/tools/asoctl/internal/client"
+	"github.com/Azure/azure-service-operator/v2/tools/asoctl/internal/crd"
+)
+
+func newCRDCleanCommand() (*cobra.Command, error) { //nolint:unparam
```
matthchr, line 17: Actually, should we just remove the `error` return here? I'm not seeing the case where the _creation of the cmd_ will fail. Execution yes of course (that's `RunE` below), but constructing the `cmd` object seems pretty unlikely to hit an error? Then we could drop the `unapram` ignoring here.

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,78 @@
...
+	"fmt"
+	"regexp"
+
+	apiextensionsV1 "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func CleanDeprecatedCRDVersions(ctx context.Context, cl apiextensionsV1.CustomResourceDefinitionInterface) error {
+	var crdRegexp = regexp.MustCompile(".*\\.azure\\.com")
+	var deprecatedVersionRegexp = regexp.MustCompile("v1alpha1api.{8}.*")
```
matthchr, line 19:
```suggestion
	var deprecatedVersionRegexp = regexp.MustCompile("v1alpha1api\d{8}(preview)?(storage)?")
```

Is a more precise match, I think?

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,78 @@
...
+	"context"
+	"fmt"
+	"regexp"
+
+	apiextensionsV1 "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func CleanDeprecatedCRDVersions(ctx context.Context, cl apiextensionsV1.CustomResourceDefinitionInterface) error {
+	var crdRegexp = regexp.MustCompile(".*\\.azure\\.com")
```
matthchr, line 18:
```suggestion
	var crdRegexp = regexp.MustCompile(`.*\.azure\.com`)
```
I think this works?

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+package crd
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+
+	apiextensionsV1 "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
```
matthchr, line 13: minor: standard pattern is that package names are all lower - so this should maybe just be `apiextensions`

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,78 @@
...
+func CleanDeprecatedCRDVersions(ctx context.Context, cl apiextensionsV1.CustomResourceDefinitionInterface) error {
+	var crdRegexp = regexp.MustCompile(".*\\.azure\\.com")
+	var deprecatedVersionRegexp = regexp.MustCompile("v1alpha1api.{8}.*")
+
+	list, err := cl.List(ctx, v1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	if list.Items == nil || len(list.Items) == 0 {
```
matthchr, line 26: minor: Possibly `nil` deref here, `list == nil`? IMO change to

```suggestion
	if list == nil || len(list.Items) == 0 {
```

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,78 @@
...
+
+	var updated bool
+	for _, item := range list.Items {
+		crdName := item.Name
+
+		if !crdRegexp.MatchString(crdName) {
+			continue
+		}
+
+		crd, err := cl.Get(ctx, crdName, v1.GetOptions{})
```
matthchr, line 38: Why do you need to issue a `Get` here? We already did a `List` above? Isn't `crd` == `item`?

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	if !updated {
+		fmt.Println("updated 0 CRDs, no deprecated versions found")
+	}
+
+	return nil
+}
+
+func getNewVersionSet(oldStoredVersions []string, versionRegexp *regexp.Regexp) (bool, []string) {
```
matthchr, line 64: minor: would calling this `removeStoredVersions` or `removeMatchingStoredVersions` be clearer?

matthchr, line 64: minor: I'm not sure you actually need the `found` boolean. Instead you could just do:

```
		found, newStoredVersions := removeMatchingStoredVersions(crd.Status.StoredVersions, deprecatedVersionRegexp)

		if len(newStoredVersions) != len(crd.Status.StoredVersions) {
```

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,78 @@
...
+		found, newStoredVersions := getNewVersionSet(crd.Status.StoredVersions, deprecatedVersionRegexp)
+
+		if found {
+			crd.Status.StoredVersions = newStoredVersions
+			crd, err = cl.UpdateStatus(ctx, crd, v1.UpdateOptions{})
+			if err != nil {
+				return err
+			}
+
+			updated = true
```
matthchr, line 52: minor: Maybe make this a counter instead, and then at the end log `updated N CRDs`? That way callers have a summary of what was changed in addition to the log-lines of each change.

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return err
+	}
+
+	if list.Items == nil || len(list.Items) == 0 {
+		fmt.Println("found 0 results, make sure you have ASO CRDs installed")
+	}
+
+	var updated bool
+	for _, item := range list.Items {
```
matthchr, line 31: Given that this is somewhat risky and destructive, we should probably have a `--dry-run` option, and document that users run that first to log what would be deleted (before deleting it)

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,78 @@
...
+		crd, err := cl.Get(ctx, crdName, v1.GetOptions{})
+		if err != nil {
+			return err
+		}
+
+		found, newStoredVersions := getNewVersionSet(crd.Status.StoredVersions, deprecatedVersionRegexp)
+
+		if found {
+			crd.Status.StoredVersions = newStoredVersions
+			crd, err = cl.UpdateStatus(ctx, crd, v1.UpdateOptions{})
```
matthchr, line 47: How do we know that executing this is safe? IMO we need to consider the following scenarios:
1. `asoctl crd clean` is run against a cluster that is running `alpha.6`. There are no `beta` storage versions, so no alpha storage versions should be removed. This could be protected against by a check above saying `if len(newStoredVersions) == 0 then error`.
2. Do we know that it's safe to remove the alpha stored version without forcing the version stored in etcd to be the beta version? I know that's what the cert-manager tool does and we're not doing it here, I believe because of some testing you did that suggested it wasn't needed - but it would be good to document what that testing was here and explain how we know it's OK to do that in a comment.

v2/tools/asoctl/internal/crd/cleaner_test.go
```diff
@@ -0,0 +1,83 @@
...
+	crd, err := fakeClientSet.CustomResourceDefinitions().Get(context.TODO(), crdName, metav1.GetOptions{})
+	asserter.Expect(err).To(BeNil())
+
+	asserter.Expect(crd.Status.StoredVersions).ToNot(BeNil())
+	asserter.Expect(crd.Status.StoredVersions).ToNot(HaveLen(2))
+	asserter.Expect(crd.Status.StoredVersions).ToNot(ContainElement(alphaVersion))
+	asserter.Expect(crd.Status.StoredVersions).To(ContainElement(betaVersion))
+}
+
+func Test_CleanDeprecatedCRDVersions_DoesNothing_IfAlphaVersionDoesNotExist(t *testing.T) {
```
matthchr, line 48: As mentioned above, we need another `DoesNothing` if `BetaVersionDoesNotExist`

v2/tools/asoctl/internal/crd/cleaner.go
theunrepentantgeek: Suggest making a dry-run the default, and requiring a `--update-cluster` option (or similar) to actually make the changes. That way it's safe by default.

v2/tools/asoctl/internal/crd/cleaner.go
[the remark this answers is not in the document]
matthchr: The logic that got me to "we just shouldn't accept users parameters here (at least initially)" is:

1. This is a dangerous operation because we're removing stored versions which is possibly destructive, so we're going to need a lot of validation on what the user passes in to make sure they don't remove _all_ of their storage versions (or really any storage version that is "in use")
2. The operator pod expects a certain set of versions to exist based on its version (`alpha.6` expects all of alpha versions, `beta.0` expects all of the beta versions but doesn't look for the alpha versions (it only cares about stored version), and `stable` will expect the stable versions and presumably not care about beta or alpha). If the expected versions are missing, the pod will crash/not function.

Achieving those two validations means that the tool is going to need to introspect the cluster to determine exactly what versions can be safely removed. At that point, it seemed pointless to give the user any real control because we'd just be making them guess at what we allowed, and each currently existing version of ASO has exactly one right "allowed set" of removals. We could just take no input and remove exactly what is supposed to be removed.

I do agree with @theunrepentantgeek's sentiment that we should structure the code such that it can be expanded in the future to possibly support the `--remove` syntax without needing a huge internal refactor, but I think exposing such an option now when the operator pod expects certain versions of certain CRDs is a bad idea because it would be very easy to remove something you shouldn't and put the pod into crashloop backoff.

OTOH: `--remove` can already be done via `kubectl` assuming that the user wants to remove the entire CRD - the unique issue that this tool is solving is removing just one version of a CRD rather than the whole CRD. I don't think that necessarily means we don't need to to worry about structuring things to be expandable (we always should just do that by default), but I am not sure that the `--remove` is going to be needed for whole CRDs because that already exists in `kubectl`. It's specifically versions within a CRD that needs this tool.

v2/tools/asoctl/internal/crd/cleaner.go
super-harsh: good catch!

v2/tools/asoctl/internal/crd/cleaner.go
super-harsh: do we want `len(newStoredVersions) == 0 then error` or `len(newStoredVersions) == 0 then ignore`. Later one seems fine, as there could be other CRDs which we want to update in the bulk

v2/tools/asoctl/pkg/crd/crd.go
```diff
@@ -0,0 +1,23 @@
...
+ * Licensed under the MIT license.
+ */
+
+package crd
+
+import "github.com/spf13/cobra"
+
+// NewCommand creates a new cobra to clean deprecated crd versions
+func NewCommand() (*cobra.Command, error) {
+
```
matthchr, line 12: minor: You have an extra newline here but not for some other functions. I don't mind the newline at the start but I think we should be consistent. Most of your functions don't have a newline here so maybe remove it?

v2/tools/asoctl/pkg/export/export.go
```diff
@@ -0,0 +1,23 @@
...
+ * Licensed under the MIT license.
+ */
+
+package export
+
+import "github.com/spf13/cobra"
+
+// NewCommand creates a new cobra Command when invoked from the command line
+func NewCommand() (*cobra.Command, error) {
+
```
matthchr, line 12: minor: You have an extra newline here but not for some other functions. I don't mind the newline at the start but I think we should be consistent. Most of your functions don't have a newline here so maybe remove it?

matthchr (commented):
This looks pretty much there to me - the only gap I think is the `--dry-run` comment [here](https://github.com/Azure/azure-service-operator/pull/2678/files#r1072638974) and the new comment I put on etcd version. I left a few other comments but they're all really minor.

v2/tools/asoctl/go.mod
```diff
@@ -2,17 +2,91 @@ module github.com/Azure/azure-service-operator/v2/tools/asoctl
 
 go 1.18
 
-// Needed to reference shared version numbering:
 replace github.com/Azure/azure-service-operator/v2 => ../../
```
matthchr, line 5: I think we can eventually do away with this in favor of a publicly exported `CreateScheme` method, but fine with this for now

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,187 @@
...
+package crd
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"time"
+
+	apiextensionsapi "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apiextensions "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
```
matthchr, line 15: minor: instead of `apiextensionsapi` and `apiextensions`, consider `apiextensions` (for `pkg/apis/apiextensions/v1`) and `apiextensionsclient` (for `pkg/client/clientset/cientset/typed...`)

matthchr: I suggest this mostly because `apiextensionsapi` reads kinda weird to me

v2/tools/asoctl/pkg/crd/crd_clean.go
```diff
@@ -0,0 +1,54 @@
...
+			if err != nil {
+				return errors.Wrap(err, "unable to create kubernetes client")
+			}
+
+			cl, err := client.New(cfg, client.Options{Scheme: controllers.CreateScheme()})
+			if err != nil {
+				return err
+			}
+
+			dryrun, err := cmd.Flags().GetBool(dryRun)
```
matthchr, line 39:
```suggestion
			dryRun, err := cmd.Flags().GetBool(dryRun)
```

matthchr: Oh wait -- the name `dryRun` is already taken. Maybe need to have a different variable name as having two names that just differ by case is awkward.

Maybe call the top one something like `dryRunFlagName` or `dryRunFlag`?
Then this can just be `dryRun`

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,187 @@
...
+		client:              client,
+		migrationBackoff:    migrationBackoff,
+		dryRun:              dryRun,
+	}
+}
+
+func (c *Cleaner) Run(ctx context.Context) error {
+	list, err := c.apiExtensionsClient.List(ctx, v1.ListOptions{})
+	if err != nil {
+		return err
```
matthchr, line 49: minor: Consider adding some extra error text here like "failed to list CRDs" or something

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,187 @@
...
+
+	return nil
+}
+
+func (c *Cleaner) updateStorageVersions(
+	ctx context.Context,
+	crd apiextensionsapi.CustomResourceDefinition,
+	newStoredVersions []string) error {
+
+	if c.dryRun {
```
matthchr, line 102: minor: On all of these functions, instead of this `if/else` pattern, maybe consider:

```
// Log
if c.dryRun {
    // return
}


```

I think that's a bit cleaner?

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,187 @@
...
+			if err != nil {
+				return err
+			}
+
+			err = c.updateStorageVersions(ctx, crd, newStoredVersions)
+			if err != nil {
+				return err
+			}
+
+			c.updated++
```
matthchr, line 83: minor: Why is this a variable on the struct rather than just a function local variable here? Seems like it could just be function local?

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	return nil
+}
+
+func (c *Cleaner) migrateObjects(ctx context.Context, objectsToMigrate *unstructured.UnstructuredList) error {
+	for _, obj := range objectsToMigrate.Items {
+		obj := obj
+		if c.dryRun {
+			fmt.Printf("resource '%s' to migrate for kind '%s'", obj.GetName(), obj.GroupVersionKind().Kind)
```
matthchr, line 120: if + continue here, then can drop the else for slightly clearer?

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,187 @@
...
+	deprecatedVersionRegexp := regexp.MustCompile(`v1alpha1api\d{8}(preview)?(storage)?`)
+	for _, crd := range list.Items {
+		crd := crd
+
+		if !crdRegexp.MatchString(crd.Name) {
+			continue
+		}
+
+		newStoredVersions := removeMatchingStoredVersions(crd.Status.StoredVersions, deprecatedVersionRegexp)
+		objectsToMigrate, err := c.getObjectsForMigration(ctx, crd, deprecatedVersionRegexp)
```
matthchr, line 66: minor: Move this inside the if, since it's not used til then?

v2/tools/asoctl/internal/crd/cleaner.go
matthchr, line 120: minor: Both of these log messages might be quite verbose if there are a lot of resources. Do we need a `Logger` interface with verbosity defined, so that users can do `--v 2` or whatever in the cmdline to get more verbose logs, but by default they don't get these because they're kinda spammy?

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+func (c *Cleaner) migrateObjects(ctx context.Context, objectsToMigrate *unstructured.UnstructuredList) error {
+	for _, obj := range objectsToMigrate.Items {
+		obj := obj
+		if c.dryRun {
+			fmt.Printf("resource '%s' to migrate for kind '%s'", obj.GetName(), obj.GroupVersionKind().Kind)
+		} else {
+			err := retry.OnError(c.migrationBackoff, isValidError, func() error { return c.client.Update(ctx, &obj) })
+			if isValidError(err) {
```
matthchr, line 123: minor: The name of this `isValidError` function confused me a bit. Consider changing the name to something like `isErrorFatal`?

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,187 @@
...
+			}
+			fmt.Printf("migrated '%s' for %s\n", obj.GetName(), obj.GroupVersionKind().Kind)
+		}
+	}
+
+	return nil
+}
+
+func isValidError(err error) bool {
+	if err != nil {
```
matthchr, line 134: same comment as elsewhere that you can invert this and do:

```
if err == nil {
    return false
}

// original contents of if err != nil here
```

and then it looks a bit cleaner

matthchr (commented):
Didn't finish but giving you what I have so far.

v2/tools/asoctl/pkg/crd/crd_clean.go
```diff
@@ -0,0 +1,55 @@
...
+			if err != nil {
+				return errors.Wrap(err, "unable to create kubernetes client")
+			}
+
+			cl, err := client.New(cfg, client.Options{Scheme: controllers.CreateScheme()})
+			if err != nil {
+				return errors.Wrap(err, "unable to create kubernetes client")
+			}
+
+			dryRun, err := cmd.Flags().GetBool(dryRunFlagName)
```
matthchr, line 40: Set up `klog` here too, call `klog.InitFlags(cmd.Flags())`, that way users can configure the log level via the cmdline

v2/tools/asoctl/pkg/crd/crd_clean.go
```diff
@@ -0,0 +1,55 @@
...
+			}
+
+			return crd.NewCleaner(
+				apiExtClient.CustomResourceDefinitions(),
+				cl,
+				dryRun).Run(context.TODO())
+		},
+	}
+
+	cmd.PersistentFlags().Bool(dryRunFlagName, false, "")
```
matthchr, line 52: May want to document the logging verbosity flag here:
`(&logging.verbosity, "v", "number for the log level verbosity")` is I believe what it should be?

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,194 @@
...
+			}
+
+			err = c.updateStorageVersions(ctx, crd, newStoredVersions)
+			if err != nil {
+				return err
+			}
+
+			updated++
+		} else {
+			klog.Infof("nothing to update for '%s'\n", crd.Name)
```
matthchr, line 89: minor, style: Update your log calls to use `klog.V(0)` for the log messages people should always see, 1 or 2 for more verbose messages. Or if you're using the ASO defined levels I think `Status` == 0

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,194 @@
...
+		crd := crd
+
+		if !crdRegexp.MatchString(crd.Name) {
+			continue
+		}
+
+		newStoredVersions := removeMatchingStoredVersions(crd.Status.StoredVersions, deprecatedVersionRegexp)
+
+		if len(newStoredVersions) > 0 && len(newStoredVersions) != len(crd.Status.StoredVersions) {
+			klog.Infof("starting cleanup for '%s'", crd.Name)
```
matthchr, line 71: minor: can use %q instead of '%s'

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,194 @@
...
+
+	return nil
+}
+
+func (c *Cleaner) migrateObjects(ctx context.Context, objectsToMigrate *unstructured.UnstructuredList) error {
+	for _, obj := range objectsToMigrate.Items {
+		obj := obj
+		if c.dryRun {
+			// TODO: verbose
+			klog.V(logging.Verbose).Infof("resource '%s' to migrate for kind '%s'", obj.GetName(), obj.GroupVersionKind().Kind)
```
matthchr, line 126: minor: Same comment on '%s' vs %q

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,194 @@
...
+	klog.Infof("updated '%s' CRD status storedVersions to : %s\n", crd.Name, updatedCrd.Status.StoredVersions)
+
+	return nil
+}
+
+func (c *Cleaner) migrateObjects(ctx context.Context, objectsToMigrate *unstructured.UnstructuredList) error {
+	for _, obj := range objectsToMigrate.Items {
+		obj := obj
+		if c.dryRun {
+			// TODO: verbose
```
matthchr, line 125: Looks like you've maybe resolved this TODO already? Can delete it probably.

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+func isErrorFatal(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	if apierrors.IsGone(err) { // If resource no longer exists, we don't want to retry
+		return false
+	} else if apierrors.IsConflict(err) { // If resource is already in the state of update, we don't want to retry either
```
matthchr, line 147: minor: Possibly expand on why this isn't a problem.

The scenario is:
We try to Update an object, but we get conflict. Conflict means somebody else updated this object between when we listed it and now. BUT because all updates will trigger a conversion to the latest storage version, their update is as good as our update and we know this resource is safe.

Maybe write that up in the comment here somehow just so people understand why this is OK - because at first thought "my update was ignored because somebody else did an update" _seems_ problematic. It's just that we don't actually care about our update (we're not changing anything on the object even)

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+
+		err := retry.OnError(c.migrationBackoff, isErrorFatal, func() error { return c.client.Update(ctx, &obj) })
+		if isErrorFatal(err) {
+			return err
+		}
+		klog.V(logging.Verbose).Infof("migrated '%s' for %s\n", obj.GetName(), obj.GroupVersionKind().Kind)
+	}
+
+	return nil
```
matthchr, line 137: minor: Same as you log how many CRDs we updated at the top level, you might want to log how many CRs we updated here? That could be logged at a higher verbosity.

v2/tools/asoctl/internal/crd/cleaner_test.go
```diff
@@ -0,0 +1,311 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	fake2 "sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+// TODO: Currently we need to create clientsets for each test as they run in parallel and we run into `resource already exists` error.
+// TODO: We may require a testing suite re-use the clientsets efficiently.
+func getClientSets() (apiextensions.ApiextensionsV1Interface, client.WithWatch, *Cleaner) {
```
matthchr, line 29: minor: consider calling this `makeClientSets` to match slightly closer to the languages terminology - we're not really "get"-ing something from somewhere in here we're constructing stuff, so either make or new make more sense IMO.

matthchr, line 29: super duper mega minor (feel free to ignore this if you want): My personal style here is that 3+ argument return functions are a bit wonky. Sometimes they make the most sense but sometimes it's worth defining a little structure that holds the 3 returns.

The reason I like that is:
1. Caller has a field named `fakeAPIExtClient`, rather than needing to know that the 1st argument is that
2. It makes it more clear when checking for errors if the last argument is/isn't an error.

but obviously this is personal preference so no worries if you wanna leave this like it is.

If you had 4 fields and an err (for 5 fields), I'd turn that into a struct with 4 fields and then return `(struct, err)`. In this case because you have no `err` I'd do a struct w/ 3 fields.

v2/tools/asoctl/internal/crd/cleaner_test.go
```diff
@@ -0,0 +1,311 @@
...
+	asserter := NewGomegaWithT(t)
+
+	fakeApiExtClient, fakeClient, cleaner := getClientSets()
+
+	alphaVersion := "v1alpha1api20200601"
+	betaVersion := "v1beta20200601"
+
+	// create CRD
+	definition := newCRDWithStoredVersions(alphaVersion, betaVersion)
+	definition.Spec.Versions = []v1.CustomResourceDefinitionVersion{
```
matthchr, line 74: I think you have this test and the next test backwards? AFAIK, `Storage` will only be set for 1 CRD version. We have to make sure that it's _not_ set for the version we're removing, because that's the version that is saved into etcd. If alpha is storage, and we do the migration, then everything in etcd will be alpha. Then we delete alpha and boom, that's bad.

We want the reverse: we're removing alpha, make sure alpha is _not_ storage. Then do migrate and then we're good.

v2/tools/asoctl/internal/crd/cleaner_test.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+
+	ns := newNamespace("test-ns")
+
+	_, err := fakeApiExtClient.CustomResourceDefinitions().Create(context.TODO(), definition, metav1.CreateOptions{})
+	asserter.Expect(err).To(BeNil())
+
+	// create Namespace
+	err = fakeClient.Create(context.TODO(), ns)
+	asserter.Expect(err).To(BeNil())
```
matthchr, line 91: minor: we usually just call this `asserter` `g`. Applies to all the tests.

v2/tools/asoctl/internal/crd/cleaner_test.go
```diff
@@ -0,0 +1,311 @@
...
+	err = cleaner.Run(context.TODO())
+	asserter.Expect(err).To(BeNil())
+
+	crd, err := fakeApiExtClient.CustomResourceDefinitions().Get(context.TODO(), definition.Name, metav1.GetOptions{})
+	asserter.Expect(err).To(BeNil())
+
+	var updatedRG resources.ResourceGroup
+	err = fakeClient.Get(context.TODO(), types.NamespacedName{Name: rg.Name, Namespace: rg.Namespace}, &updatedRG)
+	asserter.Expect(err).To(BeNil())
+	asserter.Expect(updatedRG.ResourceVersion).ToNot(BeEquivalentTo(rg.ResourceVersion))
```
matthchr, line 107: Just confirming: This is making sure that the migrate actually happened?

v2/tools/asoctl/internal/crd/cleaner_test.go
```diff
@@ -0,0 +1,311 @@
...
+
+	asserter.Expect(crd.Status.StoredVersions).ToNot(BeNil())
+	asserter.Expect(crd.Status.StoredVersions).ToNot(BeEquivalentTo(definition.Status.StoredVersions))
```
matthchr, line 110: minor: I think this 1 assertion is technically not needed. Just by asserting that it doesn't contain alpha you sorta imply that it isn't equal to `definition.Status.StoredVersions`

v2/tools/asoctl/internal/crd/cleaner_test.go
```diff
@@ -0,0 +1,311 @@
...
+
+	crd, err := fakeApiExtClient.CustomResourceDefinitions().Get(context.TODO(), definition.Name, metav1.GetOptions{})
+	asserter.Expect(err).To(BeNil())
+
+	asserter.Expect(crd.Status.StoredVersions).ToNot(BeNil())
+	asserter.Expect(crd.Status.StoredVersions).To(BeEquivalentTo(definition.Status.StoredVersions))
+	asserter.Expect(crd.Status.StoredVersions).To(ContainElement(betaVersion))
+}
+
+func Test_CleanDeprecatedCRDVersions_DoesNothing_BetaVersionDoesNotExist(t *testing.T) {
```
matthchr, line 193: minor: Match test name above: 
```suggestion
func Test_CleanDeprecatedCRDVersions_DoesNothing_IfBetaVersionDoesNotExist(t *testing.T) {
```

v2/tools/asoctl/internal/crd/cleaner_test.go
```diff
@@ -0,0 +1,311 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "resourcegroups.resources.azure.com",
+		},
+		Spec: v1.CustomResourceDefinitionSpec{
+			Group: "resources.azure.com",
+			Names: v1.CustomResourceDefinitionNames{
+				ListKind: "ResourceGroup",
+			},
+		},
+
```
matthchr, line 305: minor: extra space

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,194 @@
...
+		if err := c.client.List(ctx, list); err != nil {
+			return nil, err
+		}
+	}
+
+	return list, nil
+}
+
+// requireMigrate checks if resources under CRD need migration. Requirement for migration would depend on if the deprecated version is the storage version.
+func requireMigrate(crd apiextensions.CustomResourceDefinition, versionRegexp *regexp.Regexp) (bool, string) {
```
matthchr, line 173: See my comment on your tests, I think you have this check backwards.

I also think it might make more sense to do this check in `removeMatchingStoredVersions` and have that function return an error if it attempts to remove a version that matches the regex but is also marked as the storage version.

So basically the flow would be:

1. List CRDs
2. For each crd...
    1. Remove the version in storedVersions that matches the pattern, as long as it's not marked the storageVersion
    2. ... Do everything you're already doing.

Then I think you could delete this method and its usage?

matthchr (commented):
Looks really good - there's one gotcha which is that I think you have your check for StorageVersion backwards.

I left a few other minor comments but it LGTM other than the one StorageVersion thing.

v2/tools/asoctl/internal/crd/cleaner.go
super-harsh: isn't `klog.Infof` a default `klog.V(0)` ?

v2/tools/asoctl/internal/crd/cleaner_test.go
super-harsh: Yep, this is checking if resource is updated or not, when resourceVersion is not equal here, we have a confirmation, that the resource did change.

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,194 @@
...
+	"github.com/pkg/errors"
+	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
+	"k8s.io/klog/v2"
```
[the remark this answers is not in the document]
super-harsh: it is sorted already? what do you think is not right?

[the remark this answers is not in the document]
super-harsh: It is sorted already by golang

v2/tools/asoctl/pkg/crd/crd_clean.go
super-harsh: its already set up in main.go

v2/tools/asoctl/pkg/crd/crd_clean.go
super-harsh: Don't need to add that.. the above is already a part of `klog.InitFlags(cmd.Flags())`

v2/tools/asoctl/internal/crd/cleaner.go
super-harsh: This method would still be required to fetch a version we want to update right? Basically to get version for GVK of resources we want to list and migrate? which would be alpha version only I assume?

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,193 @@
...
+	for _, crd := range list.Items {
+		crd := crd
+
+		if !crdRegexp.MatchString(crd.Name) {
+			continue
+		}
+
+		newStoredVersions, matchedStoredVersion := removeMatchingStoredVersions(crd.Status.StoredVersions, deprecatedVersionRegexp)
+
+		// In the check below, where CRD only has an only matched version, then the newStoredVersions == 0
```
matthchr, line 71: minor: The phrasing here is a bit weird -- maybe improve?

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,193 @@
...
+	deprecatedVersionRegexp := regexp.MustCompile(`v1alpha1api\d{8}(preview)?(storage)?`)
+	for _, crd := range list.Items {
+		crd := crd
+
+		if !crdRegexp.MatchString(crd.Name) {
+			continue
+		}
+
+		newStoredVersions, matchedStoredVersion := removeMatchingStoredVersions(crd.Status.StoredVersions, deprecatedVersionRegexp)
+
```
matthchr, line 70: I thought you were going to split these two cases:

`len(newStoredVersions) == 0` becomes an error because that shouldn't ever happen and if it does we should just abort the whole run right now given the user is probably trying to run this tool against a cluster w/ the wrong version of ASO installed.

`len(newStoredVersions) != len(crd.Status.StoredVersions)` means there was nothing to remove (already removed from this CRD), so your `klog.Infof("nothing to update")` is correct

matthchr (approved):

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,193 @@
...
+type Cleaner struct {
+	apiExtensionsClient apiextensionsclient.CustomResourceDefinitionInterface
+	client              client.Client
+	migrationBackoff    wait.Backoff
+	dryRun              bool
+}
+
+func NewCleaner(apiExtensionsClient apiextensionsclient.CustomResourceDefinitionInterface, client client.Client, dryRun bool) *Cleaner {
+	migrationBackoff := wait.Backoff{ // TODO: Still need to see if we want exponential backoff or linear is fine
```
theunrepentantgeek, line 36: Theoretically linear is useful, but it's just as easy to use Exponential and that has none of the untoward side effects.

theunrepentantgeek: Best practice is to include some jitter as well.

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,193 @@
...
+
+		newStoredVersions, matchedStoredVersion := removeMatchingStoredVersions(crd.Status.StoredVersions, deprecatedVersionRegexp)
+
+		// In the check below, where CRD only has an only matched version, then the newStoredVersions == 0
+		if len(newStoredVersions) > 0 && len(newStoredVersions) != len(crd.Status.StoredVersions) {
+			klog.Infof("starting cleanup for %q", crd.Name)
+			objectsToMigrate, err := c.getObjectsForMigration(ctx, crd, matchedStoredVersion)
+			if err != nil {
+				return err
+			}
```
theunrepentantgeek, line 77: How does this short-circuit if there's nothing to do?

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,193 @@
...
+
+			err = c.updateStorageVersions(ctx, crd, newStoredVersions)
+			if err != nil {
+				return err
+			}
+
+			updated++
+		} else {
+			klog.Infof("nothing to update for %q\n", crd.Name)
```
theunrepentantgeek, line 91: Error messages should start with lowerCase, but log messages should use TitleCase.

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,193 @@
...
+		klog.Infof("would update storedVersions for %q CRD to: %s\n", crd.Name, newStoredVersions)
+		return nil
+	}
+
+	crd.Status.StoredVersions = newStoredVersions
+	updatedCrd, err := c.apiExtensionsClient.UpdateStatus(ctx, &crd, v1.UpdateOptions{})
+	if err != nil {
+		return err
+	}
+	klog.Infof("updated %q CRD status storedVersions to : %s\n", crd.Name, updatedCrd.Status.StoredVersions)
```
theunrepentantgeek, line 118:
```suggestion
	klog.Infof("Updated %q CRD status storedVersions to : %s\n", crd.Name, updatedCrd.Status.StoredVersions)
```

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,193 @@
...
+
+	return nil
+}
+
+func (c *Cleaner) migrateObjects(ctx context.Context, objectsToMigrate *unstructured.UnstructuredList) error {
+	for _, obj := range objectsToMigrate.Items {
+		obj := obj
+		if c.dryRun {
+			klog.V(logging.Verbose).Infof("resource %q to migrate for kind %q", obj.GetName(), obj.GroupVersionKind().Kind)
```
theunrepentantgeek, line 127: Phrasing here is a bit clunky - here's a suggestion, feel free to take a different approach:

```suggestion
			klog.V(logging.Verbose).Infof("Would migrate resource %q of kind %q", obj.GetName(), obj.GroupVersionKind().Kind)
```

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -0,0 +1,193 @@
...
+			continue
+		}
+
+		err := retry.OnError(c.migrationBackoff, isErrorFatal, func() error { return c.client.Update(ctx, &obj) })
+		if isErrorFatal(err) {
+			return err
+		}
+
+		klog.V(logging.Verbose).Infof("migrated %q for %s\n", obj.GetName(), obj.GroupVersionKind().Kind)
```
theunrepentantgeek, line 136:
```suggestion
		klog.V(logging.Verbose).Infof("Migrated %q of kind %s", obj.GetName(), obj.GroupVersionKind().Kind)
```

v2/tools/asoctl/internal/crd/cleaner.go
super-harsh: `retry.OnError` uses exponential backoff under the hood.
https://github.com/kubernetes/client-go/blob/master/util/retry/util.go#L50

PR #2683, opened 2023-01-18
Create PreReconciliation Extension Point

### LINKED ISSUE #2644: Feature: Extension point to control reconciliation behaviour
For some ARM resources there are times when it is ineffective to do a a PUT to update the status of the resource.

A prominent example of this is an AKS Cluster with status Updating - the AKS RP will categorically refuse any PUT request while the cluster is processing a prior request.

The current behaviour of ASO is to retry relatively quickly when those errors occur, which in turn can cause ARM Throttling on the subscription because of the high number of PUT requests. 

By creating an extension point for the reconciliation loop, we would be able to selectively defer reconciliation attempts while the cluster was in an updating state.

theunrepentantgeek:

Creates an extension point allowing resources to customize ASO to avoid sending PUT requests that can't possibly succeed.

Implements #2649
Closes #2644

Follow-up PRs will implement this extension point for selected resources.

gif

### DIFF
--- a/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
+++ b/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
@@ -277,12 +277,14 @@ func (r *azureDeploymentReconcilerInstance) BeginCreateOrUpdateResource(
 
 	// If the check says we're postponing reconcile, we're done for now as there's nothing to do.
 	if check.PostponeReconciliation() {
+		r.Log.V(Status).Info("Extension recommended postponing reconciliation")
 		return ctrl.Result{}, nil
 	}
 
 	// If the check says we're blocking reconcile, we return ReadyConditionImpactingError here to update the Ready
 	// condition is updated so the user can see why we're not reconciling right now, and to trigger a retry in a bit.
 	if check.BlockReconciliation() {
+		r.Log.V(Status).Info("Extension recommended blocking reconciliation", "message", check.Message())
 		return ctrl.Result{}, check.CreateConditionError()
 	}
 
--- a/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
+++ b/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
@@ -275,10 +275,14 @@ func (r *azureDeploymentReconcilerInstance) BeginCreateOrUpdateResource(
 		return ctrl.Result{}, impactingError
 	}
 
-	// If the check says we don't need to reconcile, we're done
-	// We use a ReadyConditionImpactingError here to ensure the Ready condition is updated so the user can see why
-	// we're not reconciling right now. We'll try again later.
-	if !check.ShouldReconcile() {
+	// If the check says we're postponing reconcile, we're done for now as there's nothing to do.
+	if check.PostponeReconciliation() {
+		return ctrl.Result{}, nil
+	}
+
+	// If the check says we're blocking reconcile, we return ReadyConditionImpactingError here to update the Ready
+	// condition is updated so the user can see why we're not reconciling right now, and to trigger a retry in a bit.
+	if check.BlockReconciliation() {
 		return ctrl.Result{}, check.CreateConditionError()
 	}
 
--- a/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
+++ b/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
@@ -9,8 +9,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
-	"net/http"
+
 	"time"
 
 	"github.com/Azure/azure-service-operator/v2/internal/util/kubeclient"
@@ -197,7 +196,8 @@ func (r *azureDeploymentReconcilerInstance) DetermineCreateOrUpdateAction(
 	if refreshRequired {
 		r.Log.V(Verbose).Info("Refreshing Status of resource")
 		err := r.updateStatus(ctx)
-		if err != nil {
+		if err != nil && !genericarmclient.IsNotFoundError(err) {
+			// We have an error and it's not because the resource doesn't exist yet
 			return CreateOrUpdateActionNoAction, NoAction, errors.Wrapf(err, "error refreshing status of resource for pre-reconciliation check")
 		}
 	}
@@ -462,10 +462,7 @@ func (r *azureDeploymentReconcilerInstance) getStatus(ctx context.Context, id st
 	// Get the resource
 	retryAfter, err := r.ARMClient.GetByID(ctx, id, apiVersion, armStatus)
 	if err != nil {
-		// If our error is anything other than a 404, we should return it
-		if !IsNotFoundError(err) {
-			return nil, retryAfter, errors.Wrapf(err, "getting resource with ID: %q", id)
-		}
+		return nil, retryAfter, errors.Wrapf(err, "getting resource with ID: %q", id)
 	}
 
 	if r.Log.V(Debug).Enabled() {
@@ -755,13 +752,4 @@ func deleteResource(
 
 	// Normally don't need to set both of these fields but because retryAfter can be 0 we do
 	return ctrl.Result{Requeue: true, RequeueAfter: retryAfter}, nil
-}
-
-func IsNotFoundError(err error) bool {
-	var responseError azcore.ResponseError
-	if errors.As(err, &responseError) {
-		return responseError.RawResponse.StatusCode == http.StatusNotFound
-	}
-
-	return false
 }
--- a/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
+++ b/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
@@ -9,8 +9,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/Azure/azure-service-operator/v2/internal/util/kubeclient"
 	"time"
+
+	"github.com/Azure/azure-service-operator/v2/internal/util/kubeclient"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/go-logr/logr"
@@ -193,7 +194,10 @@ func (r *azureDeploymentReconcilerInstance) DetermineCreateOrUpdateAction(
 	checker, refreshRequired := extensions.CreatePreReconciliationChecker(r.Extension, r.alwaysReconcile)
 	if refreshRequired {
 		r.Log.V(Verbose).Info("Refreshing Status of resource")
-		r.updateStatus(ctx)
+		err := r.updateStatus(ctx)
+		if err != nil {
+			return CreateOrUpdateActionNoAction, NoAction, errors.Wrapf(err, "error refreshing status of resource for pre-reconciliation check")
+		}
 	}
 
 	check, err := checker(ctx, r.Obj, r.KubeClient, r.ARMClient, r.Log)
--- a/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
+++ b/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
@@ -9,10 +9,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-
-	"time"
-
 	"github.com/Azure/azure-service-operator/v2/internal/util/kubeclient"
+	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/go-logr/logr"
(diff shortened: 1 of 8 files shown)


v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -187,9 +192,40 @@ func (r *azureDeploymentReconcilerInstance) DetermineCreateOrUpdateAction() (Cre
...
+		}
+	}
+
+	check, err := checker(ctx, r.Obj, r.KubeClient, r.ARMClient, r.Log)
+	if err != nil {
+		return CreateOrUpdateActionNoAction, NoAction, errors.Wrapf(err, "error during pre-reconciliation check")
+	}
+
+	if !check.ShouldReconcile() {
+		//!! Is this correct? Will this correctly display the reason why the resource is not being reconciled to the user?
```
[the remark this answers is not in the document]
matthchr: For example this is probably a `SeverityWarning`, as it's not fatal but it does mean that something might be up...

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -175,7 +178,9 @@ func (r *azureDeploymentReconcilerInstance) DetermineDeleteAction() (DeleteActio
 	return DeleteActionBeginDelete, r.StartDeleteOfResource, nil
 }
 
-func (r *azureDeploymentReconcilerInstance) DetermineCreateOrUpdateAction() (CreateOrUpdateAction, CreateOrUpdateActionFunc, error) {
+func (r *azureDeploymentReconcilerInstance) DetermineCreateOrUpdateAction(
+	ctx context.Context,
```
matthchr, line 182: The original point of this method was to help drive the state-machine we have between create/update or monitor. the reason it didn't take `ctx` was because that decision was supposed to be made based on local state.

Violating that isn't really an issue though, but I am wondering if you considered putting the new code into `BeginCreateOrUpdateResource` instead? I think it would have the same effect there as it does here but would leave this method purely state-machine-management. Possibly we could rename the `BeginCreateorUpdateResource` if you felt its name didn't match the new behavior, although I think it probably still does - it's just that the first part of "CreateOrUpdate" is "Diff" or "Check if we should Create" or whatever.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -9,8 +9,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+
 	"time"
 
+	"github.com/Azure/azure-service-operator/v2/internal/util/kubeclient"
```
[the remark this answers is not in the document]
theunrepentantgeek: Moved.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -9,8 +9,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
theunrepentantgeek: I looked at putting the code into `BeginCreateOrUpdateResource` but figured the intent was to only call that once we'd committed that either create or update was required. 😁 

I'll rejig things to put the code in that method.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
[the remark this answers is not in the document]
theunrepentantgeek: Updated. Thanks!

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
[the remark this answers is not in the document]
matthchr: Was this actually moved? I still see it up above as opposed to down in the `"github.com/Azure/azure-service-operator/v2` section. Or did you move the `time` bit? I meant `"github.com/Azure/azure-service-operator/v2/internal/util/kubeclient"` should be at the bottom of the file.

There's a setting in Goland that does this automatically for you. Let me know if you want and I can go dig it up and share what I have configured.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -175,7 +176,9 @@ func (r *azureDeploymentReconcilerInstance) DetermineDeleteAction() (DeleteActio
 	return DeleteActionBeginDelete, r.StartDeleteOfResource, nil
 }
 
-func (r *azureDeploymentReconcilerInstance) DetermineCreateOrUpdateAction() (CreateOrUpdateAction, CreateOrUpdateActionFunc, error) {
+func (r *azureDeploymentReconcilerInstance) DetermineCreateOrUpdateAction(
+	ctx context.Context,
```
matthchr, line 180: minor: After rejig, you can now remove this `ctx` as this method doesn't make any external calls so has no need for `ctx` (param is unused).

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -248,10 +251,59 @@ func (r *azureDeploymentReconcilerInstance) MonitorDelete(ctx context.Context) (
...
+		// Something went wrong running the check. We assume it's not fatal, and we'll try again later
+		// Make sure we return a ReadyConditionImpactingError so that the Ready condition is updated for the user
+		// Ideally any implementation of the checker should return a ReadyConditionImpactingError, but we can't
+		// guarantee that, so we wrap as required
+		impactingError, ok := conditions.AsReadyConditionImpactingError(err)
+		if !ok {
+			impactingError = conditions.NewReadyConditionImpactingError(
+				err,
+				conditions.ConditionSeverityWarning,
+				conditions.ReasonAzureResourceNotReady)
```
matthchr, line 292: is `ReasonAzureResourceNotReady` the right `Reason` here? Given we've already issued the GET on the Azure resource above, the Azure resource must exist here? What is "not ready" about it? The actual error details will be included so I think users will probably figure it out - I'm just not following exactly what this `Reason` means.

matthchr: Ooh I see, in the context of specifically checking something like "Is the resource in a good state for me to send an update" this makes sense.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -248,10 +251,59 @@ func (r *azureDeploymentReconcilerInstance) MonitorDelete(ctx context.Context) (
...
+		}
+
+		return ctrl.Result{}, impactingError
+	}
+
+	// If the check says we don't need to reconcile, we're done
+	// We use a ReadyConditionImpactingError here to ensure the Ready condition is updated so the user can see why
+	// we're not reconciling right now. We'll try again later.
+	if !check.ShouldReconcile() {
+		return ctrl.Result{},
```
matthchr, line 302: Aren't there two cases?
1. The resource doesn't need to reconcile because it _can't yet_ (for example, `ProvisioningState` is updating currently so no point in sending a request?). This is definitely a `Warning` as you've set up here.
2. The resource doesn't need to reconcile because it's already in a good state (i.e. we wrote custom diffing logic in this handler and everything is OK.

How do we represent that second case here? That feels like it should be `Condition` == `Ready` (i.e. no `ReadyConditionImpactingError`, we're just done).

matthchr: This maybe gets at the contract w/ this extension, is `err` only for exceptional cases, or can the extension use `err` to say: "The resource is in an -ing state, so we can't reconcile now"? In which case, the upper `if` covers case 1 in my question above and this lower case could cover case 2, either by assuming you're in that case or checking something on `check`

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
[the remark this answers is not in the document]
theunrepentantgeek: I haven't pushed my branch yet. 😁

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
theunrepentantgeek: This is why the result has a `ShouldReconcile()` method, so we can extend the type in the future.

theunrepentantgeek: We now have three factory methods for the result struct, which I think gives us the semantics we want.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -248,15 +248,43 @@ func (r *azureDeploymentReconcilerInstance) MonitorDelete(ctx context.Context) (
...
+			impactingError = conditions.NewReadyConditionImpactingError(
+				err,
+				conditions.ConditionSeverityWarning,
+				conditions.ReasonFailed)
+		}
+
+		return ctrl.Result{}, impactingError
+	}
+
+	// If the check says we don't need to reconcile, we're done
```
matthchr, line 278: minor: Clarify this, as I think you mean "we're done for now?" -- since if you were done we wouldn't be returning a `ConditionImpactingError`

matthchr: Ah, see my comment below in `prereconciliation_checker` - I think there's an issue w/ how you're handling the `Postpone` case here. `CreateConditionError` should return `nil` in the `Postpone` case.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -284,20 +312,54 @@ func (r *azureDeploymentReconcilerInstance) BeginCreateOrUpdateResource(ctx cont
...
+	// Having a checker requires our resource to have an up-to-date status
+	r.Log.V(Verbose).Info("Refreshing Status of resource")
+	statusErr := r.updateStatus(ctx)
+	if statusErr != nil && !genericarmclient.IsNotFoundError(statusErr) {
+		// We have an error, and it's not because the resource doesn't exist yet
+		return extensions.PreReconcileCheckResult{}, statusErr
+	}
+
+	// We also need to have our owner, it too with an up-to-date status
+	owner, ownerErr := r.ResourceResolver.ResolveOwner(ctx, r.Obj)
```
matthchr, line 347: minor: Wondering if it's better to leave this to the `checker` to choose to do if it wants? Rather than passing owner you could pass `ResourceResolver` - then the checker could decide to call it, or not if it didn't need it.

That would reduce unneeded owner lookups -- although I do think they're cached so we're probably ok either way.

v2/pkg/genruntime/extensions/prereconciliation_checker.go
```diff
@@ -0,0 +1,167 @@
...
+// PreReconciliationChecker is implemented by resources that want to do extra checks before proceeding with
+// a full ARM reconcile.
+type PreReconciliationChecker interface {
+	// PreReconcileCheck does a pre-reconcile check to see if the resource is in a state that can be reconciled.
+	// ARM resources should implement this to avoid reconciliation attempts that cannot possibly succeed.
+	// Returns ProceedWithReconcile if the reconciliation should go ahead.
+	// Returns BlockReconcile and a human-readable reason if the reconciliation should be skipped.
+	// ctx is the current operation context.
+	// obj is the resource about to be reconciled. The resource's State will be freshly updated.
+	// kubeClient allows access to the cluster for any required queries.
```
matthchr, line 28: minor: missing owner documentation here -- although see above question about if it'd be better to just pass `ResourceResolver` instead

v2/pkg/genruntime/extensions/prereconciliation_checker.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+}
+
+// PostponeReconcile indicates reconciliation of a resource is not currently required by returning a
+// PreReconcileCheckResult with action `Postpone`.
+// reason is an explanatory reason to show to the user via an info condition on the resource.
+func PostponeReconcile(reason string) PreReconcileCheckResult {
+	return PreReconcileCheckResult{
+		action:   preReconcileCheckResultTypePostpone,
+		severity: conditions.ConditionSeverityInfo,
```
matthchr, line 85: Ah, for `Postpone`, if we just want to allow the standard "eventually we will re-check" interval, then we shouldn't return an error here.

The success case (== no reconcile needed at this time, try again after the standard `azure_sync_period` is `ctrl.Result{}, nil`). If you return an err, you'll end up getting reconciled again after only a small interval (3m is our max backoff I think), issuing another GET to Azure.
The block case should be `ctrl.Result{}, err` (which you already have).

theunrepentantgeek: So how do we show the reason for the postponement on the condition for the resource? I thought returning an error was required here.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
theunrepentantgeek: I've modified the check result to properly do this.

v2/pkg/genruntime/extensions/prereconciliation_checker.go
theunrepentantgeek: As discussed offline, I agree that "success is success", so we don't need to show a reason for postponement.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -248,15 +248,47 @@ func (r *azureDeploymentReconcilerInstance) MonitorDelete(ctx context.Context) (
...
+				conditions.ConditionSeverityWarning,
+				conditions.ReasonFailed)
+		}
+
+		return ctrl.Result{}, impactingError
+	}
+
+	// If the check says we're postponing reconcile, we're done for now as there's nothing to do.
+	if check.PostponeReconciliation() {
+		return ctrl.Result{}, nil
```
matthchr, line 280: minor: Did you want a log message here?

v2/tools/generator/internal/testcases/json_serialization_test_case.go
```diff
@@ -315,6 +315,7 @@ func (o *JSONSerializationTestCase) createTestMethod(codegenContext *astmodel.Co
 		astbuilder.CallQualifiedFunc("err", "Error"))
 
 	// match := cmp.Equal(subject, actual, cmpopts.EquateEmpty())
+	// We include cmpopts.EquateEmpty() to allow empty slices and maps to match nil values
```
matthchr, line 318: Is this change supposed to be in this PR?

v2/tools/generator/internal/testcases/property_assignment_test_case.go
```diff
@@ -279,11 +280,15 @@ func (p *PropertyAssignmentTestCase) createTestMethod(
 		dst.NewIdent(errId),
 		astbuilder.CallQualifiedFunc("err", "Error"))
 
+	// match := cmp.Equal(subject, actual, cmpopts.EquateEmpty())
+	// We include cmpopts.EquateEmpty() to allow empty slices and maps to match nil values
```
matthchr, line 284: Same question: Is this change right? Looks like change from a different PR you already merged?

matthchr (approved):
LGTM, but check the "is this supposed to be in this PR comments" -- possibly need to rebase to clear those diffs out?

super-harsh (approved):

v2/tools/generator/internal/testcases/json_serialization_test_case.go
theunrepentantgeek: Ahh, that's a mystery solved. I wondered what had happened - I had a commit that didn't show up when I pushed, so I redid it. These should null out when I merge `main`.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
theunrepentantgeek: Yup. Added.

PR #2685, opened 2023-01-19
Add HTTP transport with settings to support multiple connections

### LINKED ISSUE #2672: Use multiple HTTP connections to ARM

Currently we use a single HTTP connection to talk to ARM. This results in all our PUT requests being processed by a single ARM endpoint, and our users are running into throttling issues due to the limit of 1200 PUT requests per sub per hour.

**Describe the improvement**

If we change ASO to use multiple HTTP connections, our activity will be distributed across multiple ARM endpoints, mitigating the throttling issue we're seeing.

**Additional context**

@matthchr has confirmed throttling is 1200 per ARM endpoint. Options seem to be:

1. Use multiple connections/connection pool (configurable in Go HTTP, see [the MaxIdleConnsPerHost property](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/c33f413c0ef58a7c7af4bdaed87fd1a6e4714f99/pkg/azureclients/armclient/azure_armclient.go#L80).

2. Set ForceAttemptHTTP2 to false

If ForceAttemptHTTP2 is true (the default), connections using HTTP2 will be multiplexed over a single TCP connection, which means still hitting a single ARM FE. See https://github.com/golang/go/issues/39302 for details about disabling HTTP2 (though we may not need to disable it, just not force it?).

**More reading**

https://github.com/OJ/gobuster/issues/127
https://github.com/Azure/azure-sdk-for-go/issues/19192

super-harsh:
Closes #2672 

Did some manual testing around it to make 2500 successful calls using a single controller and multiple controllers. Before the change, we were running into `SubscriptionRequestsThrottled` error.

As confirmed by @matthchr, we can use [go-armbalancer](https://github.com/Azure/go-armbalancer) project to tackle this problem better by mid feb. 

gif

**If applicable**:
- [ ] this PR contains tests


v2/internal/genericarmclient/generic_client.go
```diff
@@ -84,6 +84,16 @@ func NewGenericClientFromHTTPClient(cloudCfg cloud.Configuration, creds azcore.T
 	// the value IN the interface IS nil).
 	if httpClient != nil {
 		opts.Transport = httpClient
+	} else {
+		// If httpClient is not provided, we use a HTTPClient with default Transport + settings
+		// to establish multiple TCP ARMClient connections to avoid throttling.
+		// TODO: Use https://github.com/Azure/go-armbalancer here once its prod ready.
```
matthchr, line 90: IMO we should construct a default HTTPClient of our own similar to this: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/c33f413c0ef58a7c7af4bdaed87fd1a6e4714f99/pkg/azureclients/armclient/azure_armclient.go#L70

Otherwise we're:
1. Modifying the default HTTPClient, which seems bad?
2. We aren't' actually using the default settings that the SDK uses, we should use basically their default settings but with some modifications. Possibly we could clone their transport (if it's accessible, not sure if it is).

v2/internal/genericarmclient/generic_client.go
```diff
@@ -84,6 +85,27 @@ func NewGenericClientFromHTTPClient(cloudCfg cloud.Configuration, creds azcore.T
...
+		httpTransport := &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
+			DialContext: (&net.Dialer{
+				Timeout:   30 * time.Second, // default transport value
+				KeepAlive: 30 * time.Second, // default transport value
+			}).DialContext,
+			ForceAttemptHTTP2:     false,            // default is true; since HTTP/2 multiplexes a single TCP connection. we'd want to use HTTP/1, which would use multiple TCP connections.
+			MaxIdleConns:          100,              // default transport value
+			MaxIdleConnsPerHost:   10,               // default is 2, so we want to increase the number to use establish more connections.
```
matthchr, line 100: This is missing TLS min version, which the SDK has I think?

See https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/runtime/transport_default_http_client.go#L18

```
	defaultTransport := &http.Transport{
		Proxy: http.ProxyFromEnvironment,
		DialContext: (&net.Dialer{
			Timeout:   30 * time.Second,
			KeepAlive: 30 * time.Second,
		}).DialContext,
		ForceAttemptHTTP2:     true,
		MaxIdleConns:          100,
		IdleConnTimeout:       90 * time.Second,
		TLSHandshakeTimeout:   10 * time.Second,
		ExpectContinueTimeout: 1 * time.Second,
		TLSClientConfig: &tls.Config{
			MinVersion: tls.VersionTLS12,
		},
	}
```

v2/internal/genericarmclient/generic_client.go
```diff
@@ -84,6 +86,30 @@ func NewGenericClientFromHTTPClient(cloudCfg cloud.Configuration, creds azcore.T
 	// the value IN the interface IS nil).
 	if httpClient != nil {
 		opts.Transport = httpClient
+	} else {
+		// If httpClient is not provided, we use a HTTPClient with default Transport + settings
+		// to establish multiple TCP ARMClient connections to avoid throttling.
+		// TODO: Use https://github.com/Azure/go-armbalancer here once its prod ready.
+		httpTransport := &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
```
matthchr, line 94: Ahh one more thing -- we should have a single global one of these, not 1 per call to this function (that's how the SDK does it as well, see the `transport_default_http_client` link I gave below.

super-harsh: So rather than having a `httpClient` per `GenericClient`, we want to share a default `httpClient` across all `GenericClients`?

matthchr: Share the transport, the HTTPClient can be constructed 1 for each.

matthchr (approved):

Issue #2643, opened 2022-12-05
Bug: using azure service operator deploy PostgreSQL flexible servers encounter error in azure "Number of write requests for subscription xxx exceeded the limit of '1200' for time interval '01:00:00'. Please try again after xx seconds"


Gan-Jun:
**Version of Azure Service Operator**
aso/azure_service_operator:v2.0.0-beta.2, kubebuilder/kube-rbac-proxy:v0.13.0

**Describe the bug**
  We are planning to migrate our project platform from azure ChinaNorth2 region to ChinaNorth3 region so that we can switch from postgresql singleserver to postgresql flexibleserver. there are 40~50 postgresql flexibleservers to be created in ChinaNorth3 region by using azure_service_operator:v2.0.0-beta.2.
  While testing postgresql flexibleservers deployment in lower stage environment, we see there are many "**Write FirewallRules, Write Configurations, Write Databases**" requests in activity log in azure portal, those too frequent write request caused error: **Number of write requests for subscription 'xxx' exceeded the limit of '1200' for time interval '01:00:00'. Please try again after '303' seconds**. 
  Even for successfully created postgresql flexibleserver, still azure service operator sends "**Write FirewallRules, Write Configurations, Write Databases**" requests to azure.  and in values.yaml for azure service operator deployment, I am using default setting of **azureSyncPeriod: "", azureOperatorMode: ""**, but not sure why it still sends these write requests frequently.

**To Reproduce**
screenshot of settings and log:
1. values.yaml file for azure_service_operator:v2.0.0-beta.2 deployment using helm3 in azure kubernates:
![valuesyaml](https://user-images.githubusercontent.com/48749989/205610784-fc3b16eb-a038-4a55-80f1-fcf11af15c00.png)
2. use concourse pipeline job to deploy azure_service_operator and override some default settings in values.yaml file:
![concourse-job](https://user-images.githubusercontent.com/48749989/205610828-6f435136-5d74-403b-8f16-aefabb12304e.png)
3. template file for postgresql flexibleserver:
![psql-template](https://user-images.githubusercontent.com/48749989/205610863-b6ffdc88-52e2-49cf-a111-2b1f2b2c8649.png)
4. too many "**Write FirewallRules, Write Configurations, Write Databases**" requests in activity log in azure portal:
![write requests log](https://user-images.githubusercontent.com/48749989/205610938-75b46666-3a79-4c11-b66e-416df48675a9.png)

**Expected behavior**
[...]

PR #2688, opened 2023-01-20
Add Prereconcile check to PostgreSQL Flexible Server

theunrepentantgeek:

Implements the pre-reconciliation check to skip reconciliation of flexible servers that are currently updating.

**Prerequisites**

- [x] PR #2683 

Check out the changes in the recordings to see the change in behaviour.

gif

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/api/dbforpostgresql/customizations/flexible_server_extensions.go
+++ b/v2/api/dbforpostgresql/customizations/flexible_server_extensions.go
@@ -7,6 +7,11 @@ package customizations
 
 import (
 	"context"
+	"fmt"
+	"github.com/Azure/azure-service-operator/v2/internal/set"
+	"github.com/Azure/azure-service-operator/v2/internal/util/kubeclient"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/extensions"
+	"strings"
 
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/go-logr/logr"
@@ -78,4 +83,43 @@ func secretsToWrite(obj *postgresql.FlexibleServer) ([]*v1.Secret, error) {
 	collector.AddValue(operatorSpecSecrets.FullyQualifiedDomainName, to.String(obj.Status.FullyQualifiedDomainName))
 
 	return collector.Values()
+}
+
+var _ extensions.PreReconciliationChecker = &FlexibleServerExtension{}
+
+var blockingFlexibleServerStates = set.Make(
+	"starting",
+	"stopping",
+	"updating",
+)
+
+func (ext *FlexibleServerExtension) PreReconcileCheck(
+	_ context.Context,
+	obj genruntime.MetaObject,
+	_ kubeclient.Client,
+	_ *genericarmclient.GenericClient,
+	_ logr.Logger,
+	_ extensions.PreReconcileCheckFunc,
+) (extensions.PreReconcileCheckResult, error) {
+	// This has to be the current hub storage version. It will need to be updated
+	// if the hub storage version changes.
+	flexibleServer, ok := obj.(*postgresql.FlexibleServer)
+	if !ok {
+		return extensions.SkipReconcile("Expected Flexible Server"),
+			errors.Errorf("cannot run on unknown resource type %T, expected *postgresql.FlexibleServer", obj)
+	}
+
+	// Type assert that we are the hub type. This will fail to compile if
+	// the hub type has been changed but this extension has not
+	var _ conversion.Hub = flexibleServer
+
+	if state := flexibleServer.Status.State; state != nil {
+		if blockingFlexibleServerStates.Contains(strings.ToLower(*state)) {
+			return extensions.SkipReconcile(
+					fmt.Sprintf("Flexible Server is in provisioning state %q", *state)),
+				nil
+		}
+	}
+
+	return extensions.ProceedWithReconcile(), nil
 }
--- a/v2/api/dbforpostgresql/customizations/flexible_server_extensions.go
+++ b/v2/api/dbforpostgresql/customizations/flexible_server_extensions.go
@@ -88,9 +88,9 @@ func secretsToWrite(obj *postgresql.FlexibleServer) ([]*v1.Secret, error) {
 
 var _ extensions.PreReconciliationChecker = &FlexibleServerExtension{}
 
-// If a flexible server does not have a provisioningState in this set, it will reject any attempt to PUT a new state
+// If the provisioningState of a flexible server is not in this set, it will reject any attempt to PUT the resource
 // out of hand; so there's no point in even trying. This is true even if the PUT we're doing will have no effect on
-// the state of the cluster.
+// the state of the server.
 // These are all listed lowercase, so we can do a case-insensitive match.
 var nonBlockingFlexibleServerStates = set.Make(
 	"succeeded",
@@ -119,21 +119,17 @@ func (ext *FlexibleServerExtension) PreReconcileCheck(
 	// the hub type has been changed but this extension has not been updated
 	var _ conversion.Hub = server
 
-	if flexibleServerStateBlocksReconciliation(server) {
+	state := server.Status.State
+	if state != nil && flexibleServerStateBlocksReconciliation(*state) {
 		return extensions.BlockReconcile(
 			fmt.Sprintf(
 				"Flexible Server is in provisioning state %q",
-				*server.Status.State)), nil
+				state)), nil
 	}
 
 	return extensions.ProceedWithReconcile(), nil
 }
 
-func flexibleServerStateBlocksReconciliation(server *postgresql.FlexibleServer) bool {
-	state := server.Status.State
-	if state == nil {
-		return false
-	}
-
-	return !nonBlockingFlexibleServerStates.Contains(strings.ToLower(*state))
+func flexibleServerStateBlocksReconciliation(state string) bool {
+	return !nonBlockingFlexibleServerStates.Contains(strings.ToLower(state))
 }
--- a/v2/api/dbforpostgresql/customizations/flexible_server_extensions.go
+++ b/v2/api/dbforpostgresql/customizations/flexible_server_extensions.go
@@ -88,14 +88,14 @@ func secretsToWrite(obj *postgresql.FlexibleServer) ([]*v1.Secret, error) {
 
 var _ extensions.PreReconciliationChecker = &FlexibleServerExtension{}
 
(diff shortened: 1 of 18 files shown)


v2/api/dbforpostgresql/customizations/flexible_server_extensions.go
```diff
@@ -79,3 +84,42 @@ func secretsToWrite(obj *postgresql.FlexibleServer) ([]*v1.Secret, error) {
...
+
+var _ extensions.PreReconciliationChecker = &FlexibleServerExtension{}
+
+var blockingFlexibleServerStates = set.Make(
+	"starting",
+	"stopping",
+	"updating",
+)
+
+func (ext *FlexibleServerExtension) PreReconcileCheck(
```
matthchr, line 96: What's the reasoning behind having this check for PostgreSQL FlexibleServer specifically? Is it just showcasing the extension? Did we have a bug where somebody had an issue with the behavior we had previously (is there a corresponding issue?)

Generally speaking the only issue we've had that I'm aware of with blindly PUT-ing and backing off if the resource is in an `-ing` state is with AKS AgentPool, when:
1. Our backoff was to 1m, rather than 3m like now (not a ton of difference I know, but it's something).
2. The AKS AgentPool was stuck in the `-ing` state and would never leave, causing the 1m requests to continue forever. (RP bug).
3. PUT-ing to AKS AgentPool triggers AKS RP to make expensive calls to the Compute RP, which triggered compute throttling. This is also an RP bug.

I like the existence of this extension as a place where we could do GET + diff manually (if we had to, before we have it automated), or if we find a resource that specifically needs the optimization, but I'm not sure what the specific need for it here is?

If there is a specific need - can we leave a comment explaining what it is?

matthchr: On the other hand, if there is no specific need, why bother implementing this extension now? Just leave the hook for if we need it in the future?

theunrepentantgeek: We've had a bug reported for PostgreSQL Flexible server, though I don't recall the exact details and can't find them in my email. But, that's why it's listed in [ADR-2022-12-Reconciliation-Extensions.md](https://github.com/Azure/azure-service-operator/blob/main/docs/hugo/content/design/ADR-2022-12-Reconciliation-Extensions.md) as an affected resource.

matthchr: It feels like we should have some details left in the code as breadcrumbs for "why we bothered to do this for this resource" - for example in this case is the problem due to a bug or bad behavior on the server side? Have we reported it to the PostgreSQL team if so? 

Do you recall how this issue was reported? Through email or otherwise? I can also try to check mine and see if I find anything, if I know more about what I'm looking for.

v2/api/dbforpostgresql/customizations/flexible_server_extensions.go
```diff
@@ -79,3 +84,42 @@ func secretsToWrite(obj *postgresql.FlexibleServer) ([]*v1.Secret, error) {
...
+	}
+
+	// Type assert that we are the hub type. This will fail to compile if
+	// the hub type has been changed but this extension has not
+	var _ conversion.Hub = flexibleServer
+
+	if state := flexibleServer.Status.State; state != nil {
+		if blockingFlexibleServerStates.Contains(strings.ToLower(*state)) {
+			return extensions.SkipReconcile(
+					fmt.Sprintf("Flexible Server is in provisioning state %q", *state)),
```
super-harsh, line 119: Maybe adding some resource information like `obj.Name` would be helpful here?

v2/api/dbforpostgresql/customizations/flexible_server_extensions.go
super-harsh: I guess the issue is here #2643

v2/api/dbforpostgresql/customizations/flexible_server_extensions.go
```diff
@@ -79,3 +85,55 @@ func secretsToWrite(obj *postgresql.FlexibleServer) ([]*v1.Secret, error) {
...
 	return collector.Values()
 }
+
+var _ extensions.PreReconciliationChecker = &FlexibleServerExtension{}
+
+// If a flexible has a provisioningState in this set, it will reject any attempt to PUT a new state out of hand;
+// so there's no point in even trying. This is true even if the PUT we're doing will have no effect on the state of the
+// cluster.
+// These are all listed lowercase, so we can do a case-insensitive match.
+var blockingFlexibleServerStates = set.Make(
```
matthchr, line 95: TBH it might be better to define this as a set of allowed states, rather than a set of disallowed states. ARM defines only [3 terminal states](https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/async-api-reference.md#provisioningstate-property)

Those states are:
- Succeeded
- Failed
- Canceled 

Anything other than those 3 states is a transitionary state and will likely cause request reject (that's how AKS works at least, 99% sure most other RPs with "only 1 request active at a time" are the same)

matthchr: This also makes us more hardened against possible futures where they add new "-ing" states in the future (which they are allowed to do)

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -248,10 +248,38 @@ func (r *azureDeploymentReconcilerInstance) MonitorDelete(ctx context.Context) (
...
+) (ctrl.Result, error) {
 	if r.Obj.AzureName() == "" {
 		err := errors.New("AzureName was not set. A webhook should default this to .metadata.name if it was omitted. Is the ASO webhook service running?")
-		return ctrl.Result{}, conditions.NewReadyConditionImpactingError(err, conditions.ConditionSeverityError, conditions.ReasonFailed)
+		return ctrl.Result{},
+			conditions.NewReadyConditionImpactingError(
+				err, conditions.ConditionSeverityError, conditions.ReasonFailed)
+	}
+
+	check, err := r.preReconciliationCheck(ctx)
```
matthchr, line 261: Looks like this maybe needs a rebase after the merge of your other PR?

matthchr (commented):
Basically LGTM, but waiting to sign-off til after the rebase as PR is a bit hard to fully grok right now as it shows the other PR that has already been merged as well.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
theunrepentantgeek: Yup. Done.

v2/api/dbforpostgresql/customizations/flexible_server_extensions.go
theunrepentantgeek: Fair call. Done.

v2/api/dbforpostgresql/customizations/flexible_server_extensions.go
```diff
@@ -79,3 +85,55 @@ func secretsToWrite(obj *postgresql.FlexibleServer) ([]*v1.Secret, error) {
...
 	return collector.Values()
 }
+
+var _ extensions.PreReconciliationChecker = &FlexibleServerExtension{}
+
+// If a flexible server does not have a provisioningState in this set, it will reject any attempt to PUT a new state
+// out of hand; so there's no point in even trying. This is true even if the PUT we're doing will have no effect on
+// the state of the cluster.
+// These are all listed lowercase, so we can do a case-insensitive match.
+var nonBlockingFlexibleServerStates = set.Make(
```
matthchr, line 95: minor: fix varname too? This should be `blockingFlexibleServerStates`, right?

v2/api/dbforpostgresql/customizations/flexible_server_extensions.go
```diff
@@ -79,3 +85,55 @@ func secretsToWrite(obj *postgresql.FlexibleServer) ([]*v1.Secret, error) {
...
+
+	// Type assert that we are the hub type. This will fail to compile if
+	// the hub type has been changed but this extension has not been updated
+	var _ conversion.Hub = server
+
+	if flexibleServerStateBlocksReconciliation(server) {
+		return extensions.BlockReconcile(
+			fmt.Sprintf(
+				"Flexible Server is in provisioning state %q",
+				*server.Status.State)), nil
```
matthchr, line 126: Do we 100% know that `server.Status.State` is not nil? Otherwise we could panic here.

v2/api/dbforpostgresql/customizations/flexible_servers_database_extensions.go
```diff
@@ -0,0 +1,44 @@
...
+	_ logr.Logger,
+	_ extensions.PreReconcileCheckFunc,
+) (extensions.PreReconcileCheckResult, error) {
+	// Check to see if our owning server is ready for the database to be reconciled
+	if server, ok := owner.(*postgresql.FlexibleServer); ok {
+		if flexibleServerStateBlocksReconciliation(server) {
+			return extensions.BlockReconcile(
+				fmt.Sprintf(
+					"Owning FlexibleServer is in provisioning state %q",
+					*server.Status.State)), nil
```
matthchr, line 39: Do we 100% know that `server.Status.State` is not nil? Otherwise we could panic here.

matthchr (approved):

v2/api/dbforpostgresql/customizations/flexible_server_extensions.go
theunrepentantgeek: No. I changed the name to `nonBlockingFlexibleServerStates` because I changed it to the list of non-blocking states - the states that don't block us from trying to reconcile.

v2/api/dbforpostgresql/customizations/flexible_server_extensions.go
theunrepentantgeek: Theoretically, `flexibleServerStateBlocksReconciliation()` ensures that `State` is not nil, but to hedge against future issues I refactored anyway.

v2/api/dbforpostgresql/customizations/flexible_servers_database_extensions.go
theunrepentantgeek: Here too.

v2/api/dbforpostgresql/customizations/flexible_server_extensions.go
matthchr: Clearly I was tired at the end of the day

PR #2690, opened 2023-01-20
Fix bug where some resources could be missing an owner

matthchr:
This could happen if there was a name collision as the ARM URL was lost.

**If applicable**:
- [ ] this PR contains tests


theunrepentantgeek (approved):
How on earth did you find that? 😮

matthchr:
> How on earth did you find that? 😮

Painstaking debugging when a particular resource didn't have an owner. I was like "WHAT IS HAPPENING?!?" but finally tracked it down to here.

PR #2692, opened 2023-01-20
Fix bug detecting some collections of ARM IDs

matthchr:
**If applicable**:
- [x] this PR contains tests


hack/crossplane/apis/sql/v1beta20201101preview/structure.txt
```diff
@@ -147,7 +147,7 @@ github.com/Azure/azure-service-operator/hack/crossplane/apis/sql/v1beta20201101p
     │       ├── ResourceGroupNameRef: *v1alpha1.Reference
     │       ├── ResourceGroupNameSelector: *v1alpha1.Selector
     │       ├── RestorableDroppedDatabaseId: *string
-    │       ├── RestorePointInTime: Validated<*string> (0 rules)
+    │       ├── RestorePointInTime: *string
```
matthchr, line 150: Note that these changes (from `Validated<*string> (0 rules)` to `*string`) don't have any structural impact to the CRDs and so are nonbreaking.

v2/tools/generator/internal/codegen/pipeline/testdata/TestAddCrossResourceReferences_HandlesMap/person-v20200101.golden
```diff
@@ -0,0 +1,43 @@
...
+type PersonList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []Person `json:"items"`
+}
+
+type Person_Spec struct {
+	// FullName: As would be used to address mail
+	FullName   string                                  `json:"fullName,omitempty"`
+	References map[string]genruntime.ResourceReference `armReference:"Ids" json:"references,omitempty"`
```
matthchr, line 30: This is the important line here -- we detect the collection and swap the type to `genruntime.ResourceReference`

v2/tools/generator/internal/codegen/pipeline/testdata/TestAddCrossResourceReferences_HandlesArray/person-v20200101.golden
```diff
@@ -0,0 +1,43 @@
...
+type PersonList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []Person `json:"items"`
+}
+
+type Person_Spec struct {
+	// FullName: As would be used to address mail
+	FullName   string                         `json:"fullName,omitempty"`
+	References []genruntime.ResourceReference `armReference:"Ids" json:"references,omitempty"`
```
matthchr, line 30: This is the important line here. We detect the collection and swap the type to `genruntime.ResourceReference`

v2/tools/generator/internal/codegen/pipeline/add_cross_resource_references_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"testing"
+
+	. "github.com/onsi/gomega"
+
+	"github.com/Azure/azure-service-operator/v2/tools/generator/internal/astmodel"
+	"github.com/Azure/azure-service-operator/v2/tools/generator/internal/config"
+	"github.com/Azure/azure-service-operator/v2/tools/generator/internal/test"
+)
+
+var IDsArrayProperty = astmodel.NewPropertyDefinition("Ids", "ids", astmodel.NewArrayType(astmodel.ARMIDType))
```
theunrepentantgeek, line 18: Suggest adding a comment to this property, so the golden file points out the important thing:

```suggestion
var IDsArrayProperty = astmodel.NewPropertyDefinition("Ids", "ids", astmodel.NewArrayType(astmodel.ARMIDType)).
    WithComment("This is the important line here. The final type should be genruntime.ResourceReference")
```

v2/tools/generator/internal/astmodel/meta_type.go
```diff
@@ -161,3 +161,21 @@ func ExtractTypeName(aType Type) (TypeName, bool) {
...
+	if primitiveType, ok := AsPrimitiveType(aType); ok {
+		return primitiveType, ok
+	}
+
+	if arrayType, ok := AsArrayType(aType); ok {
+		return ExtractPrimitiveType(arrayType.Element())
+	}
+
+	if mapType, ok := AsMapType(aType); ok {
+		return ExtractPrimitiveType(mapType.ValueType())
```
theunrepentantgeek, line 177: Looking only at the value type of the map might be surprising to future consumers, so I'm wondering whether this method belongs here.

Either move it closer to the (only) usage, so it's clear it's specialised, OR add a comment to the code indicating why only the map value needs to be checked.

theunrepentantgeek (approved):
A couple of minor comments, but otherwise looks good.

v2/tools/generator/internal/codegen/pipeline/add_cross_resource_references_test.go
matthchr: Done

v2/tools/generator/internal/astmodel/meta_type.go
matthchr: This is a fair point. The reason I put this here is because it's structurally very similar to `ExtractTypeName` - which also has the "only checks map values" aspect. But... I also wrote that function and put it here, and it also only has a single usage (in the embedded resource handling).

I've moved the new function closer to its usage and removed it from here. Possibly the same should happen for `ExtractTypeName`, but I haven't moved it for now. Let me know if you think it should be moved too - although in its case the only checking map values has more validity as TypeNames are never the generated maps key.

PR #2695, opened 2023-01-24
Samples tests should only auto-fill secrets it must

matthchr:
Some secrets are already provided by refs and those must be accounted for when autogenerating secrets.

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/internal/controllers/samples_test.go
+++ b/v2/internal/controllers/samples_test.go
@@ -17,6 +17,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/Azure/azure-service-operator/v2/internal/reflecthelpers"
+	"github.com/Azure/azure-service-operator/v2/internal/set"
 	"github.com/Azure/azure-service-operator/v2/internal/testcommon"
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
 )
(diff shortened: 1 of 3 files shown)


v2/internal/controllers/samples_test.go
```diff
@@ -106,23 +110,58 @@ func runGroupTest(tc *testcommon.KubePerTestContext, groupVersionPath string) {
...
+	allReferences := set.Make[genruntime.SecretReference]()
+
+	// Join samples and refs
+	resources := make(map[string]genruntime.ARMMetaObject, len(samples)+len(refs))
+	for k, v := range samples {
+		resources[k] = v
+	}
+	for k, v := range refs {
+		resources[k] = v
+	}
```
theunrepentantgeek, line 137: Is there any possibility that `samples` and `refs` will have keys in common? 
If there _is_, the current handling is that `refs` wins by overwrite - is that correct?

Even if the current behaviour is correct, we need a comment explaining that.

theunrepentantgeek (approved):
Looks good. One comment, which maybe can be resolved by adding some comments to the code. Nice use of generics.

v2/internal/controllers/samples_test.go
matthchr: There's not really any possibility that they collide because the key is filepath. With that said I added:
```
		// It shouldn't be possible for these to collide as the key is a file path and samples and refs should have distinct
		// file paths given they're in different folders. In case somehow this does happen, we panic
		if _, ok := resources[k]; ok {
			panic(fmt.Sprintf("unexpected resource sample and ref collision at %s", k))
		}
```

matthchr: Doh, they _can_ collide. Fixed this to prevent that possibility.

super-harsh (approved):

PR #2697, opened 2023-01-24
Ensure we prune empty types

### LINKED ISSUE #1474: Fix handling of "empty" types
There are a number of these that are due to conversion from Swagger where all of the fields of a type are read-only:

For example:
1. https://schema.management.azure.com/schemas/2019-03-01/Microsoft.Compute.Galleries.json 
    - GalleryProperties.Identifier
2. https://schema.management.azure.com/schemas/2019-07-01/Microsoft.Compute.json 
    - VirtualMachineIdentity.UserAssignedIdentities
Both of the above are types that are entirely readonly

This causes at least a few errors when controller-gen runs that come out looking like this:
```
/home/matthchr/work/github/k8s-infra/hack/generator/apis/microsoft.compute.galleries/v20180601/galleries_types.go:75:15: unsupported AST kind *ast.InterfaceType
```

matthchr:
Fixes #1474

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/v2/tools/generator/internal/codegen/golden_files_test.go
+++ b/v2/tools/generator/internal/codegen/golden_files_test.go
@@ -160,9 +160,6 @@ func NewTestCodeGenerator(
 			pipeline.CheckForAnyTypeStageID,
 			pipeline.CreateResourceExtensionsStageID,
 			pipeline.CreateTypesForBackwardCompatibilityID,
-			// TODO: Once the stage is enabled in the pipeline, we may need to remove it here for testing
-			// pipeline.InjectHubFunctionStageID,
-			// pipeline.ImplementConvertibleInterfaceStageId,
 			pipeline.ReportOnTypesAndVersionsStageID,
 			pipeline.ReportResourceVersionsStageID,
 			pipeline.ReportResourceStructureStageId)
--- a/v2/azure-arm.yaml
+++ b/v2/azure-arm.yaml
@@ -136,63 +136,55 @@ typeTransformers:
     remove: true
     because: AccessKeys is only set on response to PUT/CREATE, but we fill out Status via GET so this field is always empty. It also contains secrets we wouldn't want to expose in status anyway.
 
-  # Deal with readonly properties that were not properly pruned in the JSON schema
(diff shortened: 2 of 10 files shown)


v2/tools/generator/internal/codegen/golden_files_test.go
```diff
@@ -160,12 +160,10 @@ func NewTestCodeGenerator(
...
 			pipeline.CreateResourceExtensionsStageID,
 			pipeline.CreateTypesForBackwardCompatibilityID,
-			// TODO: Once the stage is enabled in the pipeline, we may need to remove it here for testing
-			// pipeline.InjectHubFunctionStageID,
-			// pipeline.ImplementConvertibleInterfaceStageId,
 			pipeline.ReportOnTypesAndVersionsStageID,
 			pipeline.ReportResourceVersionsStageID,
-			pipeline.ReportResourceStructureStageId)
+			pipeline.ReportResourceStructureStageId,
+			pipeline.RemoveEmptyObjectsStageId)
```
theunrepentantgeek, line 166: We shouldn't remove stages from the test pipeline unless absolutely necessary - we want those tests to be as close as possible to the real thing. Does including this cause any failures?

theunrepentantgeek (approved):

v2/tools/generator/internal/codegen/golden_files_test.go
matthchr: It did cause failures, but I didn't look that closely into why - I just added it to the ignore. Let me check exactly what caused the failure first and make 100% sure it makes sense.

matthchr: Ah, the issue is that there is a golden test for the `additionalProperties` behavior:

> Here we check that if 'additionalProperties' is explicitly set to 'false' (and there are no other fields) the generated struct will be empty

That expects to generate an empty object. I've updated the test so that it has 2 properties, 1 which is the `additionalProperties: false` object, and the other is a simple string, and re-recorded the test so that we don't have to remove this stage.

PR #2698, opened 2023-01-25
Add Azure SQL resources, tests, and samples

### LINKED ISSUE #2318: Feature: Add support for Azure SQL
In addition to MySQL/PostgreSQL/MariaDB which we already support, we should add support for AzureSQL

matthchr:
Closes #2318

This cannot be merged until the corresponding Swagger changes are also merged: See https://github.com/Azure/azure-rest-api-specs/pull/22292

**If applicable**:
- [x] this PR contains tests


v2/api/sql/v1beta20211101/server_types_gen.go
```diff
@@ -0,0 +1,2817 @@
...
+	Administrators *ServerExternalAdministrator_STATUS `json:"administrators,omitempty"`
+
+	// Conditions: The observed state of the resource
+	Conditions []conditions.Condition `json:"conditions,omitempty"`
+
+	// FederatedClientId: The Client id used for cross tenant CMK scenario
+	FederatedClientId *string `json:"federatedClientId,omitempty"`
+
+	// FullyQualifiedDomainName: The fully qualified domain name of the server.
+	FullyQualifiedDomainName *string `json:"fullyQualifiedDomainName,omitempty"`
```
super-harsh, line 993: Do we want to export FQDN for this into a configmap?

matthchr: Good catch! Yes I think that makes sense.

Other DBs export this as a secret but that was because we didn't have the auto-configmap export. We can add it to them as well if we want, but it probably makes sense here to use that rather than into a secret.

matthchr: This is fixed

super-harsh (approved):

Issue #1440, opened 2021-05-14
Feature: Generate interfaces for augmentation/interception of conversions


theunrepentantgeek:
Add generation of the interfaces to allow the conversion process to be customized with handwritten code and integrate use of those interfaces into the generated conversion routines.

Prerequisites: Azure/k8s-infra#354

matthchr:
This is something we should probably do before we need it badly.

matthchr:
No change from above message, still need to do it.

PR #2702, opened 2023-01-26
ADR for allowing hooks to alter conversions

theunrepentantgeek:

Design for #1440 

Writing this up so we can estimate/size the work and decide whether to tackle now or later.

@matthchr has a situation where altering the generated conversion is needed.

gif

**If applicable**:


docs/hugo/content/design/ADR-2023-01-Conversion-Augmentation.md
```diff
@@ -0,0 +1,135 @@
...
+    err := src.AssignFrom(source)
+    if err != nil {
+        return errors.Wrap(
+            err, 
+            "calling custom AssignFrom() for conversion from v20210101s.BatchAccount")
+    }
+}
+```
+
+**Pro**: Minimizes the amount of code we need to generate because we only need an extra clause in the existing conversion methods.
```
matthchr, line 49: Would it be in conversion methods for every type? Or just for top level resource type?

I assume the proposal here is just for the top level resource type because as I understand it, that's the type which implements `conversion.Convertible`

docs/hugo/content/design/ADR-2023-01-Conversion-Augmentation.md
```diff
@@ -0,0 +1,135 @@
...
+
+**Con**: Requires the user to correctly provide the right type parameter when implementing the interface to augment the property conversions. For example, in this above situation, only an implementation of `genruntime.Assignable[*v20210101s.BatchAccount]` will be invoked; if the user instead implemented `genruntime.Assignable[**v20210101.BatchAccount]` the code would compile, but be silently ignored.
+
+**Con**: Go prohibits doing a type cast on a concrete type, so we need to stash `source` in an intermediate variable first.
```
matthchr, line 53: FWIW this isn't _that_ much of a con - we do this elsewhere already.

docs/hugo/content/design/ADR-2023-01-Conversion-Augmentation.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Code-gen interface implementation
+
+Instead of using a generic interface, we can define a new interface for each version of each resource. This interface will be implemented by the user and will be called by the code generator.
+
+For example, in the `v1beta20210101` package, we'll generate an interface for the conversion hook:
+
+``` go
+type assignableBatchAccount interface {
```
matthchr, line 62: This is actually kinda nice.

docs/hugo/content/design/ADR-2023-01-Conversion-Augmentation.md
```diff
@@ -0,0 +1,135 @@
...
+if err != nil {
+    return errors.Wrap(
+        err, 
+        "calling custom AssignFrom() for conversionfrom v20210101s.BatchAccount")
+}
+```
+
+**Pro**: The configuration is very explicit and easy to understand.
+
+**Con**: We need to add a new configuration option and the code to process it.
```
matthchr, line 113: Where would the implementation of this function live? It's authored by the user so the generated code won't compile on its own? That's sorta a downside as AFAIK up until now an invariant of the code generator has been that it produces code that compiles.

docs/hugo/content/design/ADR-2023-01-Conversion-Augmentation.md
theunrepentantgeek: I've updated the ADR to explicitly call this out, and included details from our earlier discussion indicating that implementing it for all resources and objects makes things easier.

docs/hugo/content/design/ADR-2023-01-Conversion-Augmentation.md
theunrepentantgeek: Agreed. I've called this out explicitly. The implementation would be written by the end user, and until that was done nothing would work.

matthchr (approved):

PR #2705, opened 2023-01-30
Move controller-resources from `internal` to  `pkg` package

super-harsh:

This PR helps us to expose controller-resources methods for `asoctl` tool to create scheme and to get resource extensions by moving them out of `internal` package


theunrepentantgeek (changes requested):
Having two packages called `controllers` (one in `internal` and one in `pkg`) is creating a rather large trip hazard for future developers. I thought the proposed package was `pkg/scheme` - which wouldn't conflict.

You also need to change the appropriate functions so that the generated methods are public

v2/internal/reconcilers/generic/register.go
```diff
@@ -0,0 +1,155 @@
...
+type Options struct {
+	controller.Options
+
+	// options specific to our controller
+	RequeueIntervalCalculator interval.Calculator
+	Config                    config.Values
+	LoggerFactory             func(obj metav1.Object) logr.Logger
+}
+
+func RegisterWebhooks(mgr ctrl.Manager, objs []client.Object) error {
```
matthchr, line 43: I like this new file - helps clean up the `generic_reconciler.go` mess a bit.

v2/internal/reconcilers/generic/generic_reconciler.go
```diff
@@ -3,7 +3,7 @@ Copyright (c) Microsoft Corporation.
 Licensed under the MIT license.
 */
 
-package controllers
+package generic
```
matthchr, line 6: I thought about it more and I think I like the `internal/reconcilers/generic` path, over `internal/controllers`.

That does mean:
`internal/reconcilers/generic` runs inner reconcilers of types `internal/reconcilers/arm` and `internal/reconcilers/mysql` (and in the future more), BUT I don't think that's all that confusing and it puts all the reconciliation stuff next to one another which makes a lot of sense.

v2/pkg/controllers/controller_resources.go
```diff
@@ -47,7 +48,7 @@ func GetKnownStorageTypes(
 	armClientFactory arm.ARMClientFactory,
```
matthchr, line 48: I don't think we should be exporting this `GetKnownStorageTypes` function -- it's really for internal use only.

What if instead of moving all of the contents of controller_resources and/or controller_resources_gen, we just moved the whole `internal/controllers` folder into `internal/reconcilers/generic`, including the tests _and_ these generated files, but we handcrafted a single `scheme.go` file in `v2/api` which contained specific API exports, like:

```
func Scheme() *runtime.Scheme {
    return generic.CreateScheme()
}
```

This way we're only exporting a select few curated methods, _and_ they're in the `api` package which corresponds to where people will likely expect them (since all of the actual schemes themselves are also in the API package). Alternatively we could export them as an alias:
```
var Scheme = generic.CreateScheme
```

The only weirdness I see with doing this is that we have a bit of a weird package import hierarchy:
v2/api imports v2/internal/reconcilers/generic
v2/internal/reconcilers/generic imports v2/api/* (all of the individual API storage types)

You could make the argument that the `v2/internal/reconcilers/generic package` should not have a dependency on `v2/api/*`, because it's the _generic_ controller. We could accomplish that by splitting `controller_resources.go`, `controller_resources_gen.go`, and all of the tests off into an `internal/resources` or something, which would then result in:

v2/api imports v2/internal/resources
v2/internal/resources imports v2/api/* 
v2/internal/reconcilers/generic imports neither of the above (it's all wired up in main and passed to the generic reconciler via interfaces)

IMO that's the direction we should be heading.

theunrepentantgeek: We do still want to be able to create different forms of ASO for different scenarios, right? 
We've been pushing back on crossplane for a while, and there's been discussion around having variants that only support selected resources too.

To support those, we need the import direction to be that `v2/api` imports the generic controller, not the other way around.

It's always felt weird that the `controller_resources.gen.go` file wasn't under `v2/api`.

matthchr: > It's always felt weird that the controller_resources.gen.go file wasn't under v2/api.

Parts of it could go under API (like the scheme stuff) but parts IMO do not. Especially because v2/api is public so anything exported from it is public, but many of the "Make a collection of resources for us to operate a controller on" such as the `GetKnownStorageTypes` function are internal not public IMO.

> To support those, we need the import direction to be that v2/api imports the generic controller, not the other way around.

This is basically what I suggested in my second suggestion right?
v2/api imports v2/internal/resources
v2/internal/resources imports v2/api/*
v2/internal/reconcilers/generic imports neither of the above (it's all wired up in main and passed to the generic reconciler via interfaces)

just that instead of `controller_resources_gen` being put into `v2/api` directly, it's put into an internal package and we manually pick which methods from it we want to export as part of `v2/api`.

super-harsh:
Closing this as the refactor is no longer required.

PR #2707, opened 2023-01-30
Treat empty maps and slices as equal to nil in property tests

### LINKED ISSUE #2706: Bug: ASO storage types don't correctly roundtrip empty slices
The autogenerated tests want to assert that two storage versions roundtrip to eachother:

- The two versions in question, `20221101storage` and `20221102previewstorage` for new AKS Managed Cluster. 
- The 20221102previewstorage version has a new property NetworkProfile which contains a few slices. When roundtripping between `20221102previewstorage` and `20221101storage`, the `NetworkProfile` field is stored in the `propertyBag` as expected, BUT there's an issue when dealing with empty slices w/ `omitempty` as they are omitted, meaning that on one object we have an empty slice and on the other object we end up with a nil slice. This causes the autogenerated roundtrip tests to fail.

The actual issue is that the generated code is just serializing the entire new `NetworkProfile` structure into the `propertyBag` using the default JSOn serialization logic (`omitempty` on our slices):
```
 // NetworkProfile    
if pool.NetworkProfile != nil {
       propertyBag.Add("NetworkProfile", *pool.NetworkProfile)
} else {
       propertyBag.Remove("NetworkProfile")
}
```

this results in `pool.NetworkProfile.myCollection` being omitted from serialization. Later on we assert that an empty slice (original) == a nil slice (round-tripped), which fails.

theunrepentantgeek:

When a complex property (one with an object type) is serialized into a property bag, empty slices and maps are omitted. When the value is rehydrated from the property bag, those empty slices and maps are returned as nil values.

This is causing our tests to fail.

Since nothing else in the ecosystem cares about the difference, it seems fair to fix our tests by ignoring the difference between empty and nil. 

Closes #2706 

gif

### DIFF
--- a/v2/tools/generator/internal/testcases/property_assignment_test_case.go
+++ b/v2/tools/generator/internal/testcases/property_assignment_test_case.go
@@ -229,6 +229,7 @@ func (p *PropertyAssignmentTestCase) createTestMethod(
 	)
 
 	cmpPackage := codegenContext.MustGetImportedPackageName(astmodel.CmpReference)
+	cmpoptsPackage := codegenContext.MustGetImportedPackageName(astmodel.CmpOptsReference)
 	prettyPackage := codegenContext.MustGetImportedPackageName(astmodel.PrettyReference)
 	diffPackage := codegenContext.MustGetImportedPackageName(astmodel.DiffReference)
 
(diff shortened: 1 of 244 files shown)


v2/tools/generator/internal/testcases/property_assignment_test_case.go
```diff
@@ -279,11 +280,14 @@ func (p *PropertyAssignmentTestCase) createTestMethod(
 		dst.NewIdent(errId),
 		astbuilder.CallQualifiedFunc("err", "Error"))
 
+	// match := cmp.Equal(subject, actual, cmpopts.EquateEmpty())
+	equateEmpty := astbuilder.CallQualifiedFunc(cmpoptsPackage, "EquateEmpty")
```
matthchr, line 284: minor: Maybe leave a comment referring to the open issue, or just explain the problem w.r.t. empty slices, so that viewers understand why this bit is key.

matthchr (approved):

PR #2711, opened 2023-02-03
Add InterfaceType

matthchr:
**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/tools/generator/internal/astmodel/interface_type.go
+++ b/v2/tools/generator/internal/astmodel/interface_type.go
@@ -1,0 +1,209 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package astmodel
+
+import (
+	"go/token"
+	"sort"
+	"strings"
+
+	"github.com/dave/dst"
+
+	"github.com/Azure/azure-service-operator/v2/tools/generator/internal/astbuilder"
+	"github.com/Azure/azure-service-operator/v2/tools/generator/internal/readonly"
+)
+
+var EmptyInterfaceType = NewInterfaceType()
+
+// InterfaceType represents an interface
+type InterfaceType struct {
+	functions readonly.Map[string, Function]
+}
+
+var _ Type = &InterfaceType{}
+
+// NewInterfaceType creates a new InterfaceType
+func NewInterfaceType(functions ...Function) *InterfaceType {
+	var f readonly.Map[string, Function]
+	if len(functions) > 0 {
+		inner := make(map[string]Function, len(functions))
+		for _, v := range functions {
+			inner[v.Name()] = v
+		}
+		f = readonly.CreateMap(inner)
+	} else {
+		f = readonly.EmptyMap[string, Function]()
+	}
+
+	return &InterfaceType{
+		functions: f,
+	}
+}
+
+// WithFunction creates a new InterfaceType with a function (method) attached to it
+func (i *InterfaceType) WithFunction(function Function) *InterfaceType {
+	// Create a copy of objectType to preserve immutability
+	result := i.copy()
+	result.functions = result.functions.With(function.Name(), function)
+
+	return result
+}
+
+// Functions returns all the function definitions
+// A sorted slice is returned to preserve immutability and provide determinism
+func (i *InterfaceType) Functions() []Function {
+	functions := i.functions.Values()
+
+	sort.Slice(functions, func(i int, j int) bool {
+		return functions[i].Name() < functions[j].Name()
+	})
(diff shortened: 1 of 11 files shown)


v2/tools/generator/internal/astmodel/interface_type.go
```diff
@@ -0,0 +1,185 @@
...
+
+// InterfaceType represents an interface
+type InterfaceType struct {
+	functions readonly.Map[string, Function]
+}
+
+var _ Type = &InterfaceType{}
+
+// NewInterfaceType creates a new InterfaceType
+func NewInterfaceType() *InterfaceType {
```
theunrepentantgeek, line 29: Suggest having a varidic parameter allowing functions to be added straight away:

```suggestion
func NewInterfaceType(functions ...Function) *InterfaceType {
```

v2/tools/generator/internal/astmodel/interface_type.go
```diff
@@ -0,0 +1,185 @@
...
+func (i *InterfaceType) References() TypeNameSet {
+	result := NewTypeNameSet()
+	i.functions.ForEach(
+		func(name string, method Function) {
+			result.AddAll(method.References())
+		})
+	return result
+}
+
+func (i *InterfaceType) AsType(codeGenerationContext *CodeGenerationContext) dst.Expr {
```
theunrepentantgeek, line 54: Missing doc-comment

theunrepentantgeek (changes requested):
Looks really good - simpler than I expected.

Need to add into `type_catalog_report.go` as well, so that interfaces are included in the reports. Once that's in, I'll approve.

v2/tools/generator/internal/astmodel/interface_type.go
matthchr: Fixed

v2/tools/generator/internal/astmodel/type_visitor_test.go
```diff
@@ -146,6 +146,8 @@ func TestIdentityVisitorReturnsEqualResult(t *testing.T) {
 
 	resource := NewResourceType(person, individual)
 
+	iface := NewInterfaceType().WithFunction(transform)
```
theunrepentantgeek, line 149: Now that `NewInterfaceType()` takes functions, you might simplify this:

```suggestion
	iface := NewInterfaceType(transform)
```

v2/tools/generator/internal/reporting/testdata/Test_TypeCatalogReport_GivenInterface_ShowsExpectedDetails.golden
```diff
@@ -0,0 +1,2 @@
+github.com/Azure/azure-service-operator/testing/person/v20200101
+└── Person: Interface
```
theunrepentantgeek, line 2: Does your test case include a function? Maybe it should include a couple so we can see the sorting works?

theunrepentantgeek: I just found the test - it doesn't. Suggest adding a couple.

theunrepentantgeek (approved):
A couple of minor suggestions, but nothing significant. Looks good.

v2/tools/generator/internal/reporting/testdata/Test_TypeCatalogReport_GivenInterface_ShowsExpectedDetails.golden
matthchr: It doesn't but the default is also to not show functions, right? I used the same "should I show functions" option as ObjectType which I think we default to false (don't show functions)?

I can write a test that tests the function showing but wasn't sure if it made sense given that's not what the default behavior is and so isn't what people are going to see in the generated files.

theunrepentantgeek: The TypeCatalog is used in a couple of places, one of which shows functions, one of which doesn't.

Using the same behaviour as ObjectType is good - but since it's a separate test, wouldn't it be easy to show functions from `Test_TypeCatalogReport_GivenInterface_ShowsExpectedDetails`? We already have `test.NewFakeFunction()` which can be used.

theunrepentantgeek: You're right that we're missing this test for `ObjectType` - I'll add one after this PR merges.

PR #2714, opened 2023-02-07
Add helper method to export controllers.CreateScheme

super-harsh:

This PR includes `v2/api/scheme.go` to expose `internal/controllers.CreateScheme()` method externally

### DIFF
--- a/v2/api/scheme.go
+++ b/v2/api/scheme.go
@@ -11,7 +11,7 @@ import (
 	"github.com/Azure/azure-service-operator/v2/internal/controllers"
 )
 
-// CreateScheme is a helper method to export 'internal/controllers/controller_resources.go/CreateScheme()' method outside internal package.
+// CreateScheme is a helper method to export 'internal/controllers/controller_resources.go/CreateScheme()' method outside internal package for external use.
 func CreateScheme() *runtime.Scheme {
 	return controllers.CreateScheme()
 }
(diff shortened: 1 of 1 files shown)


v2/api/scheme.go
```diff
@@ -0,0 +1,17 @@
...
+
+package api
+
+import (
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/Azure/azure-service-operator/v2/internal/controllers"
+)
+
+// CreateScheme is a helper method to export 'internal/controllers/controller_resources.go/CreateScheme()' method outside internal package for external use.
```
matthchr, line 14: minor: This documentation should be public facing, so just say what the method does ("creates a runtime.Scheme containing all ASO resource definitions"). if you want to include a comment about the fact that it's exporting the internal function (and/or _why_ it does that), IMO put that bit of the comment inside the method so that it doesn't show up in the public facing documentation.

matthchr (approved):
LGTM. Left a small documentation suggestion.

v2/api/scheme.go
```diff
@@ -0,0 +1,17 @@
...
+func CreateScheme() *runtime.Scheme {
```
theunrepentantgeek, line 15: Given the way Go imports are named (by default), usage of this becomes `api.CreateScheme()` - this reads very nicely. :+1:

theunrepentantgeek (approved):

v2/api/scheme.go
super-harsh: Done!

PR #2718, opened 2023-02-08
Fix build

matthchr:


v2/tools/generator/internal/config/configuration.go
```diff
@@ -59,7 +59,7 @@ type Configuration struct {
 	// Additional information about our object model
 	ObjectModelConfiguration *ObjectModelConfiguration `yaml:"objectModelConfiguration"`
 
-	GoModulePath string // TODO: Since this isn't yaml annotated it can't be set, right?
+	goModulePath string // TODO: Since this isn't yaml annotated it can't be set, right?
```
theunrepentantgeek, line 62: I think we can remove the `// TODO:`

theunrepentantgeek (approved):

v2/tools/generator/internal/config/configuration.go
matthchr: Doh, good catch

PR #2719, opened 2023-02-08
Fix bug where an update to the resource being reconciled could cause reconciliation to fail to requeue quickly like it should.

matthchr:
* Fix a bug where a PUT or PATCH to the resource being reconciled which changed only fields which we were not watching (ignored annotations for example) could cause reconciliation to fail to requeue quickly like it should.

  The root cause of this issue is that we were treating Conflict as an ignorable error, expecting that we could avoid requeing on it because we would get an updated event shortly. This only holds true when the update changes a property of the object we actually care about. If it changes a property we aren't watching we didn't requeue but also didn't get an updated event and so were "stuck" (at least until the periodic reconciliation interval was hit).
* Fix bug were we weren't triggering reconciliation when the credential-from annotation changed.

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/internal/reconcilers/annotations.go
+++ b/v2/internal/reconcilers/annotations.go
@@ -6,12 +6,16 @@ Licensed under the MIT license.
 package reconcilers
 
 import (
+	"reflect"
+
 	"github.com/go-logr/logr"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	"github.com/Azure/azure-service-operator/v2/internal/util/annotations"
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
 )
+
+const PerResourceSecretAnnotation = "serviceoperator.azure.com/credential-from"
 
 // ARMReconcilerAnnotationChangedPredicate creates a predicate that emits events when annotations
 // interesting to the generic ARM reconciler are changed
@@ -20,6 +24,16 @@ func ARMReconcilerAnnotationChangedPredicate(log logr.Logger) predicate.Predicat
 		log,
 		map[string]annotations.HasAnnotationChanged{
 			ReconcilePolicyAnnotation: HasReconcilePolicyAnnotationChanged,
+		})
+}
+
+// ARMPerResourceSecretAnnotationChangedPredicate creates a predicate that emits events when annotations
+// interesting to the generic ARM reconciler are changed
+func ARMPerResourceSecretAnnotationChangedPredicate(log logr.Logger) predicate.Predicate {
+	return annotations.MakeSelectAnnotationChangedPredicate(
+		log,
+		map[string]annotations.HasAnnotationChanged{
+			PerResourceSecretAnnotation: HasAnnotationChanged,
 		})
 }
 
@@ -36,4 +50,9 @@ func GetReconcilePolicy(obj genruntime.MetaObject, log logr.Logger) ReconcilePol
 	}
 
 	return policy
+}
+
+// HasAnnotationChanged returns true if the annotation has changed
+func HasAnnotationChanged(old *string, new *string) bool {
+	return !reflect.DeepEqual(old, new)
 }
--- a/v2/internal/reconcilers/arm/arm_client_cache.go
+++ b/v2/internal/reconcilers/arm/arm_client_cache.go
@@ -25,6 +25,7 @@ import (
 	"github.com/Azure/azure-service-operator/v2/internal/config"
 	"github.com/Azure/azure-service-operator/v2/internal/genericarmclient"
 	"github.com/Azure/azure-service-operator/v2/internal/metrics"
+	"github.com/Azure/azure-service-operator/v2/internal/reconcilers"
 	"github.com/Azure/azure-service-operator/v2/internal/util/kubeclient"
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/core"
@@ -34,9 +35,8 @@ const (
 	// #nosec
 	globalCredentialSecretName = "aso-controller-settings"
 	// #nosec
-	NamespacedSecretName        = "aso-credential"
-	PerResourceSecretAnnotation = "serviceoperator.azure.com/credential-from"
-	namespacedNameSeparator     = "/"
+	NamespacedSecretName    = "aso-credential"
+	namespacedNameSeparator = "/"
 )
 
 // ARMClientCache is a cache for armClients to hold multiple credential clients and global credential client.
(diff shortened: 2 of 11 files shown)


v2/internal/reconcilers/annotations.go
```diff
@@ -6,13 +6,17 @@ Licensed under the MIT license.
 package reconcilers
 
 import (
+	"reflect"
+
 	"github.com/go-logr/logr"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
```
matthchr, line 13: Note to self... we might be able to write a test that forces this to happen... maybe should do that in this PR.

v2/internal/reconcilers/generic/generic_reconciler.go
```diff
@@ -120,14 +122,16 @@ func (gr *GenericReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 	if err != nil {
 		// NotFound is a superfluous error as per https://github.com/kubernetes-sigs/controller-runtime/issues/377
 		// The correct handling is just to ignore it and we will get an event shortly with the updated version to patch
-		// We must also ignore conflict here because updating a resource that
-		// doesn't exist returns conflict unfortunately: https://github.com/kubernetes/kubernetes/issues/89985. This is OK
-		// to ignore because a conflict means either the resource has been deleted (in which case there's nothing to do) or
-		// it has been updated, in which case there's going to be a new event triggered for it and we can count this
-		// round of reconciliation as a success and wait for the next event.
-		return ctrl.Result{}, kubeclient.IgnoreNotFoundAndConflict(err)
+		// We do NOT ignroe conflict here because it's hard to tell if it's coming from an attempt to update a non-existing resource
```
theunrepentantgeek, line 125:
```suggestion
		// We do NOT ignore conflict here because it's hard to tell if it's coming from an attempt to update a non-existing resource
```

v2/internal/reconcilers/generic/generic_reconciler.go
```diff
@@ -120,14 +122,16 @@ func (gr *GenericReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+		// (see https://github.com/kubernetes/kubernetes/issues/89985), or if it's from an attempt to update a resource which
+		// was updated by a user. If we ignore the user-update case, we MAY get another event since they changed the resource,
+		// but since we don't trigger updates on all changes (some annotations are ignored) we also may not get a fresh event
```
theunrepentantgeek, line 128: "may not" sounds like the system won't let it happen.

```suggestion
		// but since we don't trigger updates on all changes (some annotations are ignored) we also MIGHT NOT get a fresh event
```

v2/internal/util/interval/interval_calculator.go
```diff
@@ -107,12 +107,13 @@ func (i *calculator) failureResult(req ctrl.Request, err error) (ctrl.Result, er
 	if !ok {
 		// NotFound is a superfluous error as per https://github.com/kubernetes-sigs/controller-runtime/issues/377
 		// The correct handling is just to ignore it and we will get an event shortly with the updated version to patch
-		// We must also ignore conflict here because updating a resource that
-		// doesn't exist returns conflict unfortunately: https://github.com/kubernetes/kubernetes/issues/89985. This is OK
-		// to ignore because a conflict means either the resource has been deleted (in which case there's nothing to do) or
-		// it has been updated, in which case there's going to be a new event triggered for it and we can count this
-		// round of reconciliation as a success and wait for the next event.
-		return ctrl.Result{}, kubeclient.IgnoreNotFoundAndConflict(err)
+		// We do NOT ignroe conflict here because it's hard to tell if it's coming from an attempt to update a non-existing resource
```
theunrepentantgeek, line 110:
```suggestion
		// We do NOT ignore conflict here because it's hard to tell if it's coming from an attempt to update a non-existing resource
```

v2/internal/util/interval/interval_calculator.go
```diff
@@ -107,12 +107,13 @@ func (i *calculator) failureResult(req ctrl.Request, err error) (ctrl.Result, er
...
+		// (see https://github.com/kubernetes/kubernetes/issues/89985), or if it's from an attempt to update a resource which
+		// was updated by a user. If we ignore the user-update case, we MAY get another event since they changed the resource,
+		// but since we don't trigger updates on all changes (some annotations are ignored) we also may not get a fresh event
```
theunrepentantgeek, line 113: "may not" sounds like the system won't let it happen.

```suggestion
		// but since we don't trigger updates on all changes (some annotations are ignored) we also MIGHT NOT get a fresh event
```

theunrepentantgeek (approved):

v2/internal/reconcilers/arm/arm_client_cache.go
```diff
@@ -115,7 +116,7 @@ func (c *ARMClientCache) GetClient(ctx context.Context, obj genruntime.ARMMetaOb
 }
 
 func (c *ARMClientCache) getPerResourceCredential(ctx context.Context, obj genruntime.ARMMetaObject) (*armClient, error) {
-	return c.getCredentialFromAnnotation(ctx, obj, PerResourceSecretAnnotation)
+	return c.getCredentialFromAnnotation(ctx, obj, reconcilers.PerResourceSecretAnnotation)
```
super-harsh, line 119: Could the same be for namespacedCredential, if we're not watching the namespaced secret?

v2/internal/reconcilers/arm/arm_client_cache.go
```diff
@@ -34,9 +35,9 @@ const (
 	// #nosec
 	globalCredentialSecretName = "aso-controller-settings"
 	// #nosec
-	NamespacedSecretName        = "aso-credential"
-	PerResourceSecretAnnotation = "serviceoperator.azure.com/credential-from"
-	namespacedNameSeparator     = "/"
+	NamespacedSecretName = "aso-credential"
+
```
super-harsh, line 39: maybe we don't need a space here?

v2/internal/reconcilers/arm/arm_client_cache.go
matthchr: There were 2 separate issues actually:
1. When an update was made to an object that changed a property of that object which we were not watching (for example an annotation we are ignoring) it was possible if the timing was right that the ongoing reconcile of that object was interrupted but a new one wasn't queued. This would cause us to get "stuck" at least for a while.
2. We were ignoring changes to this annotation.

1 was the "bigger" bug, as it can happen with any annotation, including user annotations that we'll never care about.
2 was just an easy additional fix, because we don't have to be watching the corresponding secret to detect changes, we're already watching our own object we just need to stop ignoring changes to this annotation.

Watching the namespace scoped secret is harder because it requires setting up a watch, etc. We can do it but it's not going to cause a bug if we don't.

v2/internal/reconcilers/arm/arm_client_cache.go
matthchr: Good eye

v2/internal/reconcilers/annotations.go
matthchr: I thought about this some more and actually, rather than trying to force this in and end-to-end test, which is fragile and sensitive to timing, really what we need is just another `interval_calculator` unit test which ensures that we have the right backoff when we hit conflict.

I've added that unit test.

PR #2722, opened 2023-02-13
Fix naming issues

theunrepentantgeek:

Fixes some naming where the singular form returned from `flect` wasn't quite right.

Notable improvements include:
* `cooky` -> `cookie`
* `cpuCfsQuotum` -> `cpuCfsQuota`
* `continuou` -> `continuous`
* `knownA` -> `knownAs`

I also submitted a [bug-report/possible-fix](https://github.com/gobuffalo/flect/pull/65) for `flect`. 

gif

**If applicable**:
- [x] this PR contains tests


v2/api/containerservice/v1alpha1api20210501/managed_cluster_types_gen.go
```diff
@@ -9917,18 +9917,18 @@ func (profile *ManagedClusterLoadBalancerProfile) AssignProperties_From_ManagedC
 
 	// EffectiveOutboundIPs
 	if source.EffectiveOutboundIPs != nil {
-		effectiveOutboundIPList := make([]ResourceReference, len(source.EffectiveOutboundIPs))
-		for effectiveOutboundIPIndex, effectiveOutboundIPItem := range source.EffectiveOutboundIPs {
+		effectiveOutboundIpList := make([]ResourceReference, len(source.EffectiveOutboundIPs))
```
matthchr, line 9920: minor: the previous casing was actually slightly more correct (for Go).

It's not a big deal though if this is hard to fix.

v2/tools/generator/go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/devigned/tab v0.1.1
 	github.com/go-openapi/jsonpointer v0.19.5
 	github.com/go-openapi/spec v0.20.4
-	github.com/gobuffalo/flect v0.2.3
+	github.com/gobuffalo/flect v1.0.0
```
matthchr, line 18: Ooh that's a big version bump!

v2/tools/generator/internal/astmodel/known_locals_set.go
```diff
@@ -28,15 +28,44 @@ func NewKnownLocalsSet(idFactory IdentifierFactory) *KnownLocalsSet {
 	}
 }
 
+// flectOverrides is an ordered list of overrides for flect pluralization
+// Earlier overrides take precedence over later ones
+var flectOverrides = []struct {
```
matthchr, line 33: There is a similar construct in `type_name.go`, see the `Singularize` function there.

Should we consider combining these into a flect-library wrapper? So instead of this we would do:

`ourflect.Singularize()` or `ourflect.Pluralize()` in both `type_name.go` and here, and we could have a shared set of overrides for all.

v2/tools/generator/internal/astmodel/known_locals_set.go
```diff
@@ -28,15 +28,44 @@ func NewKnownLocalsSet(idFactory IdentifierFactory) *KnownLocalsSet {
...
+	{"Ip", "Ips"},
+	{"knownAs", "knownAs"},
+	{"ssh", "ssh"}, // workaround flect bug - see https://github.com/gobuffalo/flect/pull/65
+}
+
 // CreateSingularLocal creates a new unique Go local variable for a single value with one of the specified suffixes.
 func (locals *KnownLocalsSet) CreateSingularLocal(nameHint string, suffixes ...string) string {
 	hint := flect.Singularize(nameHint)
+	lowerHint := strings.ToLower(nameHint)
+	for _, o := range flectOverrides {
```
matthchr, line 48: minor: could this be turned into a function so that its implementation could be shared?

I'm thinking something like:
```
func (locals *KnownLocalsSet) applyFlectOverrides(flectValue string, originalValue string) string {
	...
}
```

Although maybe better to do what my other comment suggested of a flect-wrapper so we can use it in `type_name` too

v2/tools/generator/go.mod
sio4: Yeah, the module was already quite stable but we just kept it as v0, but I wanted to make it v1 since we are planning the next version, v2. Also, there was a heavy change in supporting irregular plurals, but the major version number doesn't mean a big change though.

v2/tools/generator/internal/astmodel/known_locals_set.go
sio4, line 33: Just for your information, Flect supports user data. If you add your own set of "single":"plural" data (`inflections.json`) in the app's working directory, the JSON file will be parsed and override the Flect defaults. It could be useful.

For example:

```json
{
  "FIPS": "FIPS",
  "Id": "Ids",
  "Ip": "Ips",
  "as": "as"
}
```

theunrepentantgeek: That's a good idea, but not one I want to pursue in this PR. I'll make a note to do this later.

v2/tools/generator/internal/astmodel/known_locals_set.go
theunrepentantgeek: Agreed. But a job for a different PR.

matthchr (approved):
Approved but I do think in the long term a single flect wrapper would be better than the 2-ish we have now

v2/tools/generator/internal/astmodel/known_locals_set.go
matthchr: Maybe leave a TODO here?

PR #2733, opened 2023-02-19
Add PrivateLinkService and PrivateEndpoint support

### LINKED ISSUE #1313: Support for Private Link
Azure services such as Azure Database for PostgreSQL, MySQL, SQL support private link.
Using private links enables AKS to connect to these database services using private ip addresses.
It appears that ASO doesn't provide a way to create services with private endpoint.

Can ASO create Azure services with private endpoint?

This will help to Partners who are building their IP solutions for deployments across multiple clouds. In Azure context, they can use private link support to avoid use of public ip addresses and improve security posture.

super-harsh:
Closes #1313 

With the support for PrivateLinkService and PrivateEndpoint, we don't need to export any field as DNS and Ip address to reach the endpoint/service is either provided by the user, `ServiceConnection`, or by the `vnet`.
 
**If applicable**:
- [x] this PR contains tests


v2/api/network/v1alpha1api20201101storage/network_interface_types_gen_test.go
```diff
@@ -5,6 +5,7 @@ package v1alpha1api20201101storage
 
 import (
 	"encoding/json"
+	v20180901s "github.com/Azure/azure-service-operator/v2/api/network/v1beta20180901storage"
```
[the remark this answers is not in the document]
matthchr: Does this change need to be rebased onto `main` where the per-resource storage version stuff is, at which point this change will go away, maybe?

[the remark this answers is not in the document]
matthchr: Hmm, no it seems to have that change.

v2/azure-arm.yaml
```diff
@@ -1164,6 +1164,9 @@ objectModelConfiguration:
       PrivateDnsZone:
         $export: true
         $supportedFrom: v2.0.0-beta.2
+      PrivateDnsZones_VirtualNetworkLink:
```
matthchr, line 1167: Consider exporting `2020-06-01` `PrivateDnsZones_VirtualNetworkLink` instead? I'm not sure what the difference between the two API versions is but you don't have to match the parents API version for the child resource. Should probably use the latest version if we can.

matthchr, line 1167: Do we also need to support DNS RecordSets? This is the URL in their Swagger:
`"/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/privateDnsZones/{privateZoneName}/{recordType}/{relativeRecordSetName}":`

It's a bit funky because actually it's defining like 8 different resourceTypes there (`recordType` is an enum with a bunch of options including "A", etc).

I don't know what our generator will do with that, but do we need these resources too for users to create actual DNS entires?

v2/azure-arm.yaml
```diff
@@ -1274,6 +1277,35 @@ objectModelConfiguration:
       VirtualNetworks_VirtualNetworkPeering:
         $exportAs: VirtualNetworksVirtualNetworkPeering
         $supportedFrom: v2.0.0-alpha.1
+    2022-07-01:
+      PrivateEndpoint:
+        $export: true
+        $supportedFrom: v2.0.0
+      PrivateEndpointConnection_STATUS:
+        $isResource: true
```
matthchr, line 1285: Curious about this - making a note to myself to understand this workflow

v2/internal/controllers/crd_networking_privatednszone_test.go
```diff
@@ -48,3 +53,48 @@ func Test_Networking_PrivateDnsZone_CRUD(t *testing.T) {
 	tc.Expect(retryAfter).To(BeZero())
 	tc.Expect(exists).To(BeFalse())
 }
+
+func VirtualNetworkLinks_CRUD(tc *testcommon.KubePerTestContext, zone *network.PrivateDnsZone, rg *resources.ResourceGroup) {
```
matthchr, line 57: minor: Maybe `PrivateDNSZone_VirtualNetworkLinks_CRUD`?

v2/internal/controllers/crd_networking_privatednszone_test.go
```diff
@@ -48,3 +53,48 @@ func Test_Networking_PrivateDnsZone_CRUD(t *testing.T) {
...
+
+	old := links.DeepCopy()
+	key := "foo"
+	links.Spec.Tags = map[string]string{key: "bar"}
+
+	tc.PatchResourceAndWait(old, links)
+	tc.Expect(links.Status.Tags).To(HaveKey(key))
+
+	tc.DeleteResource(links)
+
```
matthchr, line 75: minor: extra newline

v2/internal/controllers/crd_networking_privatednszone_test.go
```diff
@@ -48,3 +53,48 @@ func Test_Networking_PrivateDnsZone_CRUD(t *testing.T) {
...
 	tc.Expect(exists).To(BeFalse())
 }
+
+func VirtualNetworkLinks_CRUD(tc *testcommon.KubePerTestContext, zone *network.PrivateDnsZone, rg *resources.ResourceGroup) {
+	vnet := newVMVirtualNetwork(tc, testcommon.AsOwner(rg))
+	tc.CreateResourceAndWait(vnet)
+
+	links := newVirtualNetworkLink(tc, zone, vnet)
+
+	tc.CreateResourcesAndWait(links)
```
matthchr, line 63: Can we make the link and the vnet at the same time? So:

```
vnet := newVMVirtualNetwork(tc, testcommon.AsOwner(rg))
links := newVirtualNetworkLink(tc, zone, vnet)
tc.CreateResourcesAndWait(vnet, links)
```

Should be able to, right?

v2/internal/controllers/crd_networking_privatednszone_test.go
```diff
@@ -48,3 +53,48 @@ func Test_Networking_PrivateDnsZone_CRUD(t *testing.T) {
 	tc.Expect(retryAfter).To(BeZero())
 	tc.Expect(exists).To(BeFalse())
 }
+
+func VirtualNetworkLinks_CRUD(tc *testcommon.KubePerTestContext, zone *network.PrivateDnsZone, rg *resources.ResourceGroup) {
+	vnet := newVMVirtualNetwork(tc, testcommon.AsOwner(rg))
+	tc.CreateResourceAndWait(vnet)
+
+	links := newVirtualNetworkLink(tc, zone, vnet)
```
matthchr, line 61: minor: It's just 1 link, right?
```suggestion
	link := newVirtualNetworkLink(tc, zone, vnet)
```

v2/internal/controllers/crd_networking_privatednszone_test.go
```diff
@@ -24,19 +26,22 @@ func Test_Networking_PrivateDnsZone_CRUD(t *testing.T) {
...
+	zone := newPrivateDNSZone(tc, name, rg)
 
 	tc.CreateResourceAndWait(zone)
 
 	tc.Expect(zone.Status.MaxNumberOfRecordSets).ToNot(BeNil())
 	tc.Expect(zone.Status.MaxNumberOfVirtualNetworkLinks).ToNot(BeNil())
 
+	tc.RunParallelSubtests(
+		testcommon.Subtest{
+			Name: "Test_WorkspaceCompute_CRUD",
```
matthchr, line 38: minor: Wrong test name.

v2/internal/controllers/crd_networking_privateendpoint_test.go
```diff
@@ -0,0 +1,117 @@
...
+	sa.Spec.Kind = &saKind
+
+	vnet := newVMVirtualNetwork(tc, testcommon.AsOwner(rg))
+	subnet := newVMSubnet(tc, testcommon.AsOwner(vnet))
+
+	tc.CreateResourcesAndWait(sa, vnet, subnet)
+
+	tc.Expect(sa.Status.Id).ToNot(BeNil())
+
+	endpoint := newPrivateEndpoint(tc, rg, sa, subnet)
```
matthchr, line 39: Minor: Same question as in other test. Can we make this in parallel too? That's how it would be done w/ `kubectl apply`

So:
```
tc.CreateResourcesAndWait(sa, vnet, subnet, endpoint)
```

v2/internal/controllers/crd_networking_privateendpoint_test.go
```diff
@@ -0,0 +1,117 @@
...
+	)
+
+	tc.DeleteResourceAndWait(endpoint)
+
+	// Ensure delete
+	exists, retryAfter, err := tc.AzureClient.HeadByID(tc.Ctx, armId, string(network.APIVersion_Value))
+	tc.Expect(err).ToNot(HaveOccurred())
+	tc.Expect(retryAfter).To(BeZero())
+	tc.Expect(exists).To(BeFalse())
+
```
matthchr, line 69: minor: extra newline

v2/internal/controllers/crd_networking_privateendpoint_test.go
```diff
@@ -0,0 +1,117 @@
...
+	endpoint.Spec.Tags = map[string]string{key: "bar"}
+
+	tc.PatchResourceAndWait(old, endpoint)
+	tc.Expect(endpoint.Status.Tags).To(HaveKey(key))
+
+	tc.RunParallelSubtests(
+		testcommon.Subtest{
+			Name: "Test_DNSZoneGroup_CRUD",
+			Test: func(tc *testcommon.KubePerTestContext) {
+				DNSZoneGroup_CRUD(tc, vnet, endpoint, rg)
```
matthchr, line 57: minor: Consider calling this `PrivateEndpoint_DNSZoneGroup_CRUD`?

v2/internal/controllers/crd_networking_privateendpoint_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	// Ensure delete
+	exists, retryAfter, err := tc.AzureClient.HeadByID(tc.Ctx, armId, string(network.APIVersion_Value))
+	tc.Expect(err).ToNot(HaveOccurred())
+	tc.Expect(retryAfter).To(BeZero())
+	tc.Expect(exists).To(BeFalse())
+
+}
+
+func DNSZoneGroup_CRUD(tc *testcommon.KubePerTestContext, vnet *v1beta20201101.VirtualNetwork, endpoint *network.PrivateEndpoint, rg *resources.ResourceGroup) {
```
matthchr, line 72: Ok so this test is basically testing the scenario where the user has a VNET + PrivateDNSZone already configured on that VNET, and they want the PrivateEndpoint to create an entry in that existing PrivateDNSZone?

This seems pretty OK from a user-experience for me.

What about the other scenarios? User has NO private DNS zone and just wants to associate the Private Endpoint with a VNET directly. Should we have another test for that scenario? That scenario will make use of the `customDnsConfigs` property of the `PrivateEndpoint`, right?

v2/internal/controllers/crd_networking_privateendpoint_test.go
```diff
@@ -0,0 +1,117 @@
...
+	. "github.com/onsi/gomega"
+
+	"github.com/Azure/azure-service-operator/v2/api/network/v1beta20201101"
+	network "github.com/Azure/azure-service-operator/v2/api/network/v1beta20220701"
+	resources "github.com/Azure/azure-service-operator/v2/api/resources/v1beta20200601"
+	storage "github.com/Azure/azure-service-operator/v2/api/storage/v1beta20210401"
+	"github.com/Azure/azure-service-operator/v2/internal/testcommon"
+)
+
+func Test_Networking_PrivateEndpoint_CRUD(t *testing.T) {
```
matthchr, line 21: Did you reference an ARM template/bicep example to produce this test? If so can you link it in a comment up here at the top?

v2/internal/controllers/crd_networking_privateendpoint_test.go
```diff
@@ -0,0 +1,117 @@
...
+	endpoint := &network.PrivateEndpoint{
+		ObjectMeta: tc.MakeObjectMeta("endpoint"),
+		Spec: network.PrivateEndpoint_Spec{
+			Location: tc.AzureRegion,
+			Owner:    testcommon.AsOwner(rg),
+			PrivateLinkServiceConnections: []network.PrivateLinkServiceConnection{
+				{
+					Name:                        to.StringPtr("testEndpoint"),
+					PrivateLinkServiceReference: tc.MakeReferenceFromResource(sa),
+					GroupIds:                    []string{"blob"}, // TODO: This is a bit weird that user has to figure out the group ID(s).
```
matthchr, line 108: is this the same thing they need to do in an ARM template?

v2/internal/controllers/crd_networking_privatelinkservice_test.go
```diff
@@ -0,0 +1,149 @@
...
+
+	rg := tc.CreateTestResourceGroupAndWait()
+
+	vnet := newVMVirtualNetwork(tc, testcommon.AsOwner(rg))
+
+	privateLinkServiceNetworkPolicies := network20201101.SubnetPropertiesFormat_PrivateLinkServiceNetworkPolicies_Disabled
+	subnet := newVMSubnet(tc, testcommon.AsOwner(vnet))
+	subnet.Spec.PrivateLinkServiceNetworkPolicies = &privateLinkServiceNetworkPolicies
+
+	lb, frontendIPConfigurationARMID := newLbForPLS(tc, rg, subnet)
```
matthchr, line 33:
```suggestion
	lb, frontendIPConfigurationARMID := newLoadBalancerForPLS(tc, rg, subnet)
```

I think one acronym per method is probably ideal =)

v2/internal/controllers/crd_networking_privatelinkservice_test.go
```diff
@@ -0,0 +1,149 @@
...
+			Location: tc.AzureRegion,
+			Owner:    testcommon.AsOwner(rg),
+			Visibility: &network20220701.ResourceSet{
+				Subscriptions: []string{tc.AzureSubscription},
+			},
+			OperatorSpec: &network20220701.PrivateLinkServiceOperatorSpec{
+				ConfigMaps: &network20220701.PrivateLinkServiceOperatorConfigMaps{
+					Alias: &genruntime.ConfigMapDestination{
+						Name: plsConfigMap,
+						Key:  "alias",
```
matthchr, line 72: What exactly is alias and why do users want to consume it?

v2/azure-arm.yaml
matthchr: Do you have anywhere in tests where you show how this is used? How does this work? it seems weird for this to be a resource, and it also feels weird for this to be its name.

I assume this has something to do with showing the status of the PrivateEndpointConnection (This is a GET-only resource in ARM, basically?)

Can you walk me through how the user uses this? I don't actually see an example of it in the samples?

matthchr (commented):
Bunch of comments but this overall looks pretty good to me. Some of my comments boil down to "I am not sure I understand this workflow" - which is Azure's issue... but I wanna make sure it's clear for users of ASO.

v2/internal/controllers/crd_networking_privatelinkservice_test.go
super-harsh: Alias is a globally unique name for your service. It helps you mask the customer data for your service and at the same time creates an easy-to-share name for your service. When you create a Private Link service, Azure generates an alias for your service that you can share with your customers. Your customers can use this alias to request a connection to your service.

The alias is composed of three parts: Prefix.GUID.Suffix

Prefix is the service name. You can pick your own prefix. After "Alias" is created, you can't change it, so select your prefix appropriately.

GUID will be provided by platform. This GUID makes the name globally unique.

Suffix is appended by Azure: region.azure.privatelinkservice

Complete alias: Prefix. {GUID}.region.azure.privatelinkservice

https://learn.microsoft.com/en-us/azure/private-link/private-link-service-overview#alias

v2/internal/controllers/crd_networking_privatednszone_test.go
theunrepentantgeek: If we can't, we have a problem, because users will definitely want to put them into the same YAML file and have ASO work out the sequencing.

v2/internal/controllers/crd_networking_privateendpoint_test.go
theunrepentantgeek: If it is the same as the ARM template story, we may want to have the GroupIds configurable from config maps so that we can automate the workflow.

v2/internal/controllers/crd_networking_privatelinkservice_test.go
theunrepentantgeek: This sounds like something that should be called out in our docs. Consider adding a doc fragment to `docs/v2/azure/supported-resources` so that this included.

v2/api/network/v1alpha1api20201101storage/network_interface_types_gen_test.go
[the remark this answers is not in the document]
theunrepentantgeek: This does seem weird. Some investigation is in order to verify whether these changes are correct. I'll do that.

theunrepentantgeek (changes requested):
Blocking merge until we finish investigation of the version changes in the generated tests.

v2/internal/controllers/crd_networking_loadbalancer_test.go
```diff
@@ -116,3 +107,19 @@ func Test_Networking_LoadBalancer_CRUD(t *testing.T) {
 	g.Expect(retryAfter).To(BeZero())
 	g.Expect(exists).To(BeFalse())
 }
+
+// TODO: This is still really awkward
+func getFrontendIPConfigurationARMID(tc *testcommon.KubePerTestContext, rg *resources.ResourceGroup, lbName string, lbFrontendName string) (string, error) {
```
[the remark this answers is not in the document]
matthchr: I kinda think that he should just return the error there rather than panicing -- since it's already asserted against w/ `g.expect` above and that feels like it'll generate a clearer error.

v2/internal/controllers/crd_networking_privateendpoint_test.go
super-harsh: There is an ARM bicep example identical to this, however, they use `SQLServer` and we're using `StorageAccount` in this test. Still do we add the link?

v2/api/network/customizations/private_endpoint_extensions.go
```diff
@@ -0,0 +1,48 @@
...
+	_ kubeclient.Client,
+	_ *genericarmclient.GenericClient,
+	_ logr.Logger,
+	_ extensions.PreReconcileCheckFunc,
+) (extensions.PreReconcileCheckResult, error) {
+
+	if endpoint, ok := obj.(*network.PrivateEndpoint); ok && endpoint.Status.PrivateLinkServiceConnections != nil {
+
+		for _, connection := range endpoint.Status.PrivateLinkServiceConnections {
+			if *connection.PrivateLinkServiceConnectionState.Status != "Approved" {
```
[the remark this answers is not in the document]
super-harsh: Not really, I looked in the specs as well. The comment on `connection.PrivateLinkServiceConnectionState.Status` (below) states what would be the response. 

> // Status: Indicates whether the connection has been Approved/Rejected/Removed by the owner of the service.

eskaufel:
How is the progress on this? We need this as well

v2/internal/controllers/crd_networking_privateendpoint_test.go
super-harsh: the problem is, GroupIDs are not findable from a resource, it can be funky to get groupIDs for each type of resource into a configMap. For example, for Batch/BatchAccounts, it can be either `batchAccount` or `nodeManagement`, for CosmosDB/DatabaseAccounts it is `Sql`, for ServiceBus/namespaces it'll be `namespace` etc.

v2/internal/controllers/crd_networking_privateendpoint_test.go
matthchr: Yeah I think still worth adding the link. Just say you replaced SQLServer with StorageAccount

v2/internal/controllers/crd_networking_privateendpoint_test.go
matthchr: I am not sure I follow the configMap angle either. I just wanted to confirm that we're at parity with ARM templates or BICEP here, which I think we are. The user needs to (somehow) know what the GroupID is for their resource. I think if that's the same experience users get in the ARM template/BICEP/Terraform world then that's OK we're at parity with how it is in the rest of Azure.

v2/azure-arm.yaml
```diff
@@ -1238,6 +1238,32 @@ objectModelConfiguration:
...
         $supportedFrom: v2.0.0-beta.2
         $importable: false
+    2020-06-01:
+      PrivateDnsZones_VirtualNetworkLink:
+        $exportAs: PrivateDnsZonesVirtualNetworkLink
+        $supportedFrom: v2.0.0
+        $importable: false
+      PrivateDnsZones_A:
+        $exportAs: PrivateDnsZonesRecordsetTypeA
+        $supportedFrom: v2.0.0
```
matthchr, line 1248: Do we need to support `FromConfig` for some of the A/AAAA record inputs, such as the IP addr?

matthchr (commented):
There were a bunch of user scenarios we were tracking with this work, here's my take on what we wanted and what you've got:

- [x] Private Link Service CRUD support
- [x] Private endpoint CRUD support

The Private Endpoint connection scenarios:
- [ ] PE connects to VNET with no PrivateDNSZone. Supported via ??? (how does this case work? Azure Magic?)
- [x] PE connects to VNET with PrivateDNSZone: Supported via `PrivateEndpointsPrivateDnsZoneGroup`. do we also need to support via raw IP (IP exported from PE, imported into ARecord, via configmap?)
- [ ] PE connects to VNET using on-prem DNS. Requires DNS forwarder stuff we don't have set up here.

Is that right?

v2/api/network/customizations/private_endpoint_extensions.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/go-logr/logr"
+
+	network "github.com/Azure/azure-service-operator/v2/api/network/v1beta20220701storage"
+	"github.com/Azure/azure-service-operator/v2/internal/genericarmclient"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+)
+
+var _ extensions.PreReconciliationChecker = &PrivateEndpointExtension{}
+
+func (ext *PrivateEndpointExtension) PreReconcileCheck(
```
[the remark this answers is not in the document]
super-harsh: Don't we reconcile immediately after creation?

v2/samples/network/v1beta20220701/v1beta20220701_privateendpointsprivatednszonegroup.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: network.azure.com/v1beta20220701
+kind: PrivateEndpointsPrivateDnsZoneGroup
```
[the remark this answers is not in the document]
matthchr: > Previously, the DNS records for the private endpoint were created via scripting (retrieving certain information about the private endpoint and then adding it on the DNS zone). With the DNS zone group, there is no need to write any additional CLI/PowerShell lines for every DNS zone. Also, when you delete the private endpoint, all the DNS records within the DNS zone group will be deleted as well.

Seems to imply auto-deletion too, and it'd be weird to have auto-deletion but not auto-creation? I'm pretty sure it creates an entry too.

v2/api/network/customizations/private_endpoint_extensions.go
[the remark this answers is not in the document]
matthchr: > Don't we reconcile immediately after creation?

No. By default each reconcile is a PUT. We only reconcile when there's a `spec` change, or after `azureSyncPeriod` (default 1h) passes.

v2/azure-arm.yaml
super-harsh: > do we also need to support via raw IP (IP exported from PE, imported into ARecord, via configmap?

A record gets automatigically created in privateDNSZone

super-harsh: Hence, we don't require such

v2/api/network/customizations/private_endpoint_extensions.go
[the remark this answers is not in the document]
super-harsh: Considering we don't require this, resolving other improvement comments for `v2/api/network/customizations/private_endpoint_extensions.go`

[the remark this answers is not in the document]
super-harsh: Filed an issue to track above #2799

matthchr (approved):
LGTM. We should probably merge this before my huge PR that adds the v1api version and removes v1alpha.

That way I need to rebase and fix it as opposed to you need to.

v2/internal/controllers/crd_networking_privateendpoint_test.go
theunrepentantgeek: I don't see the comment, yet.

theunrepentantgeek (approved):
Looks good.

v2/internal/controllers/crd_networking_privateendpoint_test.go
super-harsh: Added!

PR #2744, opened 2023-02-23
Move asoctl:ci target to run sequentially

super-harsh:

Moving asoctl:ci target to run sequentially. As we're experiencing timeouts randomly this could be because of a race in the `generator:ci` and `asoctl:ci` trying to modify the files parallelly.


Taskfile.yml
```diff
@@ -67,9 +67,9 @@ tasks:
     deps: 
     - generator:ci
     - controller:ci
-    - asoctl:ci
     - crossplane:ci
     cmds:
+    - task: asoctl:ci
```
matthchr, line 72: Im not convinced that this is the issue. If you look at an example stuck run: 
https://github.com/Azure/azure-service-operator/actions/runs/4249026647/jobs/7388790440

The final logs are:
```
[generator:unit-tests-cover] go: downloading golang.org/x/text v0.7.0
[generator:unit-tests-cover] go: downloading github.com/josharian/intern v1.0.0
[generator:unit-tests-cover] go: downloading golang.org/x/sys v0.5.0
task: [controller:generate-genruntime-deepcopy] find ./pkg/genruntime -type f -name "zz_generated.*" -delete
task: [controller:generate-genruntime-deepcopy] cd pkg/genruntime && controller-gen object:headerFile=/workspace/v2/boilerplate.go.txt paths=./...
task: [controller:generate-genruntime-deepcopy] cd pkg/genruntime && gofmt -l -s -w .
[controller:generate-genruntime-deepcopy] conditions/zz_generated.deepcopy.go
[controller:generate-genruntime-deepcopy] zz_generated.deepcopy.go
Error: The operation was canceled.
```

To me that looks like either `controller:generate-genruntime-deepcopy` or `generator:unit-tests-cover` are getting stuck, not `asoctl`, which can't even be building at the time those steps run as it has a dependency on `controller:generate-crds` which hasn't happened yet.

I think we should add some logs to the end of the  `controller:generate-genruntime-deepcopy` and `generator:unit-tests-cover` steps to see if they are actually finishing. Then when things get stuck again we can determine if the issue is either of those commands.

We may also want to stick a log at the beginning of `produce-markdown-summary` as that task is deferred til after the `generator:unit-tests-cover` finishes -- so it should run immediately after, but obviously isn't.

matthchr:
For reference here's a successful run to compare that stuck run to: https://github.com/Azure/azure-service-operator/actions/runs/4249442170/jobs/7389599661

matthchr:
Actually it looks like this might be related to:
https://github.com/keycloak/keycloak-benchmark/issues/113
https://github.com/go-task/task/issues/715

It sounds like it's possible to deadlock go-task if we use `run: when_changed` or `run: once` -- some of the issues say when `-C` is 1, but others seem to suggest any usage of `-C` can cause it. We use `-C 2`.

theunrepentantgeek:
This has been discussed over on the Task Discord - best hypothesis was that something was trying to acquire the lock to select the next task to run before the lock was released for the previous one.

matthchr:
If that's the case, rather than force asoctl to run sequentially maybe we should just drop our usage of  `-C`?

theunrepentantgeek:
We added `-C` flag to limit the concurrency of tasks because we were overloading the build machines (they only have two vCPUs and not that much memory) and things were running really _really_ slowly. 

Have a look at #2475 (Limit concurrency of CI builds). 

And we're already using `-C 2` - see #2485 (Lift concurrency to 2 simultaneous tasks).

PR #2748, opened 2023-02-24
Code generate ResourceGroup

matthchr:
Removes the handcrafted ResourceGroup in favor of a code generated one. The two are the same in structure _except_ for `Status.ProvisioningState`, which is now `Status.Properties.ProvisioningState` (matching ARM).

This change is breaking for:
1. Consumers of the status `provisioningState`
2. Consumers of the API types directly (which we currently don't give a go mod guarantee for anyway)

**If applicable**:
- [x] this PR contains tests


matthchr:
Note that I had to edit the recordings but only because the generated code has the same properties in a slightly different order which impacts JSON serialization order. In practice this isn't an issue.

I specifically tested old RG to new RG in kind manually by installing the old shape, then installing the new CRDs and operator and using the new shape -- exisitng RGs get reconciled just fine. This is as expected because the JSON shape in etcd hasn't changed even though the storage type has.

theunrepentantgeek (approved):

PR #2753, opened 2023-02-27
Lift concurrency to 3

theunrepentantgeek:

**Task** can experience deadlocks while waiting on executing tasks to complete. 

We were using `--concurrency 2` but this only worked because we were only ever waiting on an outstanding task once. With the recent introduction of the build steps for `asoctl` we can now end up with two concurrent waits on an outstanding tasks at the same time. 

Lifting the concurrent limit should avoid the deadlock until the fix has been merged.

Fix for **Task** is here in [PR#1025](https://github.com/go-task/task/pull/1025).

gif

### DIFF
--- a/.github/workflows/pr-validation-fork.yml
+++ b/.github/workflows/pr-validation-fork.yml
@@ -94,7 +94,8 @@ jobs:
       - name: Run integration tests
         run: |
           container_id=${{steps.devcontainer.outputs.id}}
-          docker exec -e AZURE_TENANT_ID -e AZURE_CLIENT_ID -e AZURE_CLIENT_SECRET -e AZURE_SUBSCRIPTION_ID -e AZURE_CLIENT_SECRET_MULTITENANT -e AZURE_CLIENT_ID_MULTITENANT "$container_id" task controller:ci-integration-tests --concurrency 2 # limit concurrency because build machines are small
+          docker exec -e AZURE_TENANT_ID -e AZURE_CLIENT_ID -e AZURE_CLIENT_SECRET -e AZURE_SUBSCRIPTION_ID -e AZURE_CLIENT_SECRET_MULTITENANT -e AZURE_CLIENT_ID_MULTITENANT "$container_id" task controller:ci-integration-tests --concurrency 3 # limit concurrency because build machines are small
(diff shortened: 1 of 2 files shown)


super-harsh:
or we can let the concurrency be == 2 since and change the order of how we run task deps for `ci` target. Which would result in below, or do you think that'll still be an issue? 
 
 ``` 
   ci:
    desc: Run all CI checks.
    deps: 
    - asoctl:ci
    - generator:ci
    - controller:ci
    - crossplane:ci
    cmds:
    - task: verify-no-changes

.github/workflows/pr-validation-fork.yml
```diff
@@ -94,7 +94,8 @@ jobs:
       - name: Run integration tests
         run: |
           container_id=${{steps.devcontainer.outputs.id}}
-          docker exec -e AZURE_TENANT_ID -e AZURE_CLIENT_ID -e AZURE_CLIENT_SECRET -e AZURE_SUBSCRIPTION_ID -e AZURE_CLIENT_SECRET_MULTITENANT -e AZURE_CLIENT_ID_MULTITENANT "$container_id" task controller:ci-integration-tests --concurrency 2 # limit concurrency because build machines are small
+          docker exec -e AZURE_TENANT_ID -e AZURE_CLIENT_ID -e AZURE_CLIENT_SECRET -e AZURE_SUBSCRIPTION_ID -e AZURE_CLIENT_SECRET_MULTITENANT -e AZURE_CLIENT_ID_MULTITENANT "$container_id" task controller:ci-integration-tests --concurrency 3 # limit concurrency because build machines are small
+          # Change concurrency to --concurrency 1 once https://github.com/go-task/task/pull/1025 is merged
```
matthchr, line 98:
```suggestion
          # Change concurrency to --concurrency 2 once https://github.com/go-task/task/pull/1025 is merged
```

Unless we wanted 1 before but had deadlock issues?

I thought we picked 2 because the build machines have 2 cores, in which case a default of 2 seems OK to me?

matthchr: Same comment applies below as well.

matthchr (approved):

PR #2758, opened 2023-03-01
Make default credential optional

### LINKED ISSUE #2725: Make global secret entirely optional
Global secret with `AZURE_SUBSCRIPTION_ID` set is required.

**Describe the improvement**
Now that we have `credential-from`, this should be entirely optional for those who want to run their operator without a global credential.

**Additional context**
We need a test to make sure that the resulting authentication error if the secret is not specified (but is attempted to be used) is good.

super-harsh:
Closes #2725 

I did consider and tried to have envtest for this, however, changes are in `main` or `ARMClientCache` for which envtest makes no sense as we create test environment distinct from `main` so any test would not test any changes in main and will have to replicate changes in envtest environment. We can have an e2e suite test(not sure if we still want to have it) for this by using a task to create operator without `subscriptionID` and run the suite test on it(like how we do for `multi-tenant` and `pre-release` tests. 

For now, I have included an `ARMClientCache` unit test.

**If applicable**:
- [x] this PR contains tests


v2/cmd/controller/main.go
```diff
@@ -169,6 +159,33 @@ func main() {
 	}
 }
 
+func getGetDefaultCredential(cfg config.Values, setupLog logr.Logger) (azcore.TokenCredential, error) {
+
+	// If subscriptionID is not supplied, then set default credential to not be used/nil
+	if cfg.SubscriptionID == "" {
+		setupLog.Info("No default credential set found, continuing without default credential")
```
theunrepentantgeek, line 166: Are we calling this a _global_ credential or a _default_ one? 

Our users will be reading both the documentation and the logs - we should be consistent, else we run the risk of being confusing.

Remember also that some associate `default` with failure - when you _default_ on a loan, you've failed to pay it back.

Also, is this one credential, or many? A "credential set" implies one or more different credentials, but that doesn't match the code.

Maybe this?

```suggestion
		setupLog.Info("No global credential configured, continuing without.")
```

v2/internal/reconcilers/arm/arm_client_cache.go
```diff
@@ -110,6 +110,9 @@ func (c *ARMClientCache) GetClient(ctx context.Context, obj genruntime.ARMMetaOb
 		return client.GenericClient(), client.CredentialFrom(), nil
 	}
 
+	if c.globalClient.GenericClient() == nil {
+		return nil, "", errors.New("Default global credential is not configured. Use either namespaced or per-resource credential")
```
theunrepentantgeek, line 114:
```suggestion
		return nil, "", errors.New("Global credential not configured, you must use either namespaced or per-resource credentials")
```

theunrepentantgeek (approved):
Pleasantly surprised by how brief this change is. Some minor changes to wording required, but otherwise looks good. :shipit:

docs/hugo/content/introduction/authentication.md
```diff
@@ -14,7 +14,8 @@ Azure Service Operator supports three different styles of authentication today.
 ### Prerequisites
 1. An existing Azure Service Principal.
 
-To use Service Principal authentication, specify an `aso-controller-settings` secret with the `AZURE_CLIENT_ID` and `AZURE_CLIENT_SECRET` keys set.
+To use Service Principal authentication, specify an `aso-controller-settings` secret with the `AZURE_CLIENT_ID` and `AZURE_CLIENT_SECRET` keys set. This Service Principal is used as a default global credential and is optional. However, it is only suggested to exclude above in [single-operator-multitenancy](https://azure.github.io/azure-service-operator/introduction/multitenant-deployment/#single-operator-multitenancy-default) mode
```
matthchr, line 17: Consider rephrasing the second sentence here to: "This Service Principal is used as a default global credential and is optional. If you do not specify a global credential, you must follow the instructions in [single-operator-multitenancy](https://azure.github.io/azure-service-operator/introduction/multitenant-deployment/#single-operator-multitenancy-default) to specify per-namespace or per-resource credentials"

v2/cmd/controller/main.go
```diff
@@ -169,6 +159,33 @@ func main() {
...
+		return nil, nil
+	}
+
+	var credential azcore.TokenCredential
+	var err error
+	if cfg.UseWorkloadIdentityAuth {
+		credential, err = identity.NewWorkloadIdentityCredential(cfg.TenantID, cfg.ClientID)
+		if err != nil {
+			setupLog.Error(err, "unable to get workload identity credential")
+			os.Exit(1)
```
matthchr, line 176: IMO do `errors.Wrapf` here, drop the `os.Exit(1)` and return the error. Then where this is called, do:
```
if err != nil { 
    log the error
    os.Exit(1)
}
```

Currently where this method is called you don't do anything with the error, which I think should trip our linters too...

v2/cmd/controller/main.go
```diff
@@ -90,27 +90,17 @@ func main() {
...
-	var credential azcore.TokenCredential
-	if cfg.UseWorkloadIdentityAuth {
-		credential, err = identity.NewWorkloadIdentityCredential(cfg.TenantID, cfg.ClientID)
-		if err != nil {
-			setupLog.Error(err, "unable to get workload identity credential")
-			os.Exit(1)
-		}
-	} else {
-		credential, err = azidentity.NewDefaultAzureCredential(nil)
+	credential, err := getGetDefaultCredential(cfg, setupLog)
```
matthchr, line 93: Need to check `err` here?

v2/internal/reconcilers/arm/arm_client_cache.go
```diff
@@ -110,6 +110,9 @@ func (c *ARMClientCache) GetClient(ctx context.Context, obj genruntime.ARMMetaOb
 		return client.GenericClient(), client.CredentialFrom(), nil
 	}
 
+	if c.globalClient.GenericClient() == nil {
+		return nil, "", errors.New("Global credential not configured, you must use either namespaced or per-resource credentials")
```
matthchr, line 114: nice

matthchr (changes requested):
I think you also need to update azure-service-operator/v2/config/manager/manager_image_patch.yaml to make the `SUBSCRIPTION_ID` optional.

Same for `AZURE_TENANT_ID` and `AZURE_CLIENT_ID` which also aren't marked optional right now.

matthchr (approved):

nojnhuh:
> I think you also need to update azure-service-operator/v2/config/manager/manager_image_patch.yaml to make the `SUBSCRIPTION_ID` optional.
> 
> Same for `AZURE_TENANT_ID` and `AZURE_CLIENT_ID` which also aren't marked optional right now.

Sorry to dig up ancient history, but was this change ever made? I found that I still had to do this to avoid having to create the aso-controller-settings Secret, otherwise I see an Event prevent ASO from starting:
```
89s         Warning   Failed                  pod/azureserviceoperator-controller-manager-84446f4877-z8jw7    Error: secret "aso-controller-settings" not found
```

https://github.com/Azure/azure-service-operator/blob/239eea73cdc26007005e35d05fe2442ec0e3214a/v2/config/manager/manager_image_patch.yaml#L27-L37

matthchr:
There was a reason we didnt' do this but I need to dig into why. It may be that it doesn't matter anymore... (IIRC it had to do with pod boot ordering and Helm)

PR #2766, opened 2023-03-08
Add support for AZURE_CLIENT_CERTIFICATE_PATH

### LINKED ISSUE #2761: Allow authentication using Service Principle with Certificate
ASO currently only supports using a client secret for Service Principal-backed authentication.

**Describe the improvement**
Allow Service Principals to use a certificate for authentication like is described here: https://learn.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication-service-principal?tabs=azure-cli

**Additional context**

super-harsh:
Closes #2761 

Changes are to add `AZURE_CLIENT_CERTIFICATE_PATH` as an env var which would be handled by `azidentity.DefaultAzureCredential` to fetch credentials. Also, to read the certificate as an env var from a secret we have helm settings that can be used. However, for the yaml installation, users would have to add a secret mount manually.


scripts/generate-helm-manifest.sh
```diff
@@ -46,6 +46,9 @@ sed -i "s@$KUBE_RBAC_PROXY.*@{{.Values.image.kubeRBACProxy}}@g" "$GEN_FILES_DIR"
 sed -i "s@$LOCAL_REGISTRY_CONTROLLER_DOCKER_IMAGE@{{.Values.image.repository}}@g" "$GEN_FILES_DIR"/*_deployment_* # Replace hardcoded ASO image
 # Perl multiline replacements - using this because it's tricky to do these sorts of multiline replacements with sed
 perl -0777 -i -pe 's/(template:\n.*metadata:\n.*annotations:\n(\s*))/$1\{\{- if .Values.podAnnotations \}\}\n$2\{\{ toYaml .Values.podAnnotations \}\}\n$2\{\{- end \}\}\n$2/igs' "$GEN_FILES_DIR"/*_deployment_* # Add pod annotations
+perl -0777 -i -pe 's/(volumeMounts:\n(\s*))/$1\{\{- if .Values.azureClientCert.path \}\}\n$2\- mountPath: .Values.azureClientCert.path\n$2\  name: azure-client-ca\n$2\  readOnly: true\n$2\{\{- end \}\}\n$2/igs' "$GEN_FILES_DIR"/*_deployment_* # Add volumeMount for azureClientCert secret
+perl -0777 -i -pe 's/(volumes:\n(\s*))/$1\{\{- if .Values.azureClientCert.path \}\}\n$2\- name: azure-client-ca\n$2\  secret:\n$2\    defaultMode: 420\n$2\    secretName: \{\{ .Values.azureClientCert.secretName \}\}\n$2\{\{- end \}\}\n$2/igs' "$GEN_FILES_DIR"/*_deployment_* # Add volume for azureClientCert secret
```
theunrepentantgeek, line 50: The amount of perl in our scripts is starting to make me concerned. Perl is famously a write-only language, even for those fluent in it. 

I was once fluent in perl (some 30 years ago), so I'm sure I _could_ understand what these lines are doing, but it'd likely take me some heavy consultation with manuals to work it out.

Why are we using sed/perl to make these changes instead of a tool that understands YAML? 
This seems like a brittle approach that's likely to founder on whitespace issues and other trivia.

For JSON, there's `jq`, is there a good equivilent for YAML?

super-harsh: Good point! I tried to use `yq` initially, but the problem is with adding helm statements. So, if we add a helm `if` statement, it'll look something like below. Which then, runs into an error for yq processing as it's no more a valid YAML. 

```
      {{- if .Values.azureClientCert.path }}
      - name: azure-client-ca
        secret:
          defaultMode: 420
          secretName: {{ .Values.azureClientCert.secretName }}
      {{- end }}

Comment (matthchr) [reply scripts/generate-helm-manifest.sh:50]: Yeah we can't use `yq` because our files actually aren't valid YAML (because of the Helm template substitutions). I also ran into this in my "manage CRDs" PR.

Honestly, I think the ecosystem for producing Helm charts is just not great, especially from Kustomize which is how we do it. I'm not sure that we're really using Kustomize "right" either, given the amount of scripting we have wrapped around it (I begrudgingly added more...)

https://jfrog.com/blog/power-up-helm-charts-using-kustomize-to-manage-kubernetes-deployments/ talks about some ways to use Kustomize to natively do these things.

I've filed #2778 but I think to solve it we need a holistic look at how we do things.

File: v2/charts/azure-service-operator/values.yaml [hunk 2]
```diff
@@ -13,6 +13,12 @@ azureClientID: ""
 # This is required when using Service Principal authentication.
 azureClientSecret: ""
 
+azureClientCert:
```
matthchr, line 16: I was expecting to see something more like how [CAPZ supports it]( https://capz.sigs.k8s.io/topics/multitenancy.html#service-principal-with-certificate)

Rather than us just taking the PATH here, which will only work if the user somehow modifies our deployment to inject a certificate into it, I think we need to support it in a way that can also be multitenant (like how CAPZ does).

I think that means:
1. The user would give us an `AZURE_CLIENT_CERTIFICATE` and `AZURE_CLIENT_CERTIFICATE_PASSWORD` entries in the operator secret.
2. We would use these entries to construct the client, not the `AZURE_CLIENT_CERTIFICATE_PATH` environment 
variable.

matthchr (changes requested):

super-harsh:
Will be updating/adding documentation around the usage of this once we agree on how we want to use/implement this.

super-harsh:
Closing this and will be creating a new one for supporting `AZURE_CLIENT_CERTIFICATE` and `AZURE_CLIENT_CERTIFICATE_PASSWORD`.

PR #2769, opened 2023-03-10
Make the operator pod manage CRDs

matthchr:
Adds a new command line parameter to the operator pod, allowing configuration of what CRDs to install. Initially only supports "*" (or empty).

"*" means install all the CRDs.
unspecified means don't install any CRDs

We originally were using a CRD to configure this, but there were a number of downsides:
1. It required extra work to `kubectl apply` the operator because the CRD creation races with the instance of CRD. This significantly complicated initial install (1 step to 3)
2. It required extra work to `helm install`. The extra work could be hidden inside `helm` for the most part, but it did require `helm --wait` - if you didn't use `--wait` installation would error with a somewhat obscure error.

I considered moving to a ConfigMap, but given our desired goal of a clear failure mode, it actually made the most sense to move to a pod cmdline parameter. Changing the parameters in the deployment will automatically trigger a cycle of the underlying pods. Using a ConfigMap but trying to put the status onto the pod was awkward as you have to launch the `ctrl.Manager` first in order to start the reconciler, at which point the pod will start responding to /healthz as healthy.

Another advantage of the cmdline approach is that it's promising the least. If we have a CRD or ConfigMap which we're watching and reconciling they might be surprised that any update triggers a pod restart. Making it a cmdline parameter makes it very clear that the operator pod will have to restart to apply the new changes. If we want to evolve to promise more in the future based on user requests, we can but we're not committing to it up front.

Also adds a /readyz probe to ensure that the pod is ready. In this case ready _includes_ CRD reconciliation, so the pod will not be marked as ready until it has listed current CRDs and possibly added new ones.

Helm chart testing:
- Default configuration - Passes
- `installCRDs` = false - Confirmed we only install + use the crd-reader role. Pod starts successfully but watches 0 resources (because 0 resources installed). For users to get this to work, they can manually install whatever CRDs they want BEFORE chart and operator will dynamically discover them.
- Multitenant standard - Confirmed it works, updated Taskfile test to use Helm rather than raw YAML as this actually didn't work before.
- Multitenant CRDs off - Confirmed we use crd-reader role
- Upgrade from previous chart

Future possible improvements:
- Support more than just `*`.

**If applicable**:
- [x] this PR contains tests


v2/internal/crdmanagement/crds.go
```diff
@@ -0,0 +1,238 @@
...
+	for _, crd := range list.Items {
+		m.logger.V(0).Info("Found a CRD", "CRD", crd.Name)
+	}
+
+	return list.Items, nil
+}
+
+func (m *Manager) LoadOperatorCRDs(path string) ([]apiextensions.CustomResourceDefinition, error) {
+	// Expectation is that every file in this folder is a CRD
+	entries, err := os.ReadDir(path)
```
theunrepentantgeek, line 73: Should you ~dig~ recurse into nested folders so that we've got an option later on to organize the CRDs in some way, or is that a case of YAGNI?

v2/internal/crdmanagement/crds.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	return needUpdate
+}
+
+// fixCRDNamespace fixes up namespace references in the CRD to match the provided namespace.
+// This could in theory be done with a string replace across the JSON representation of the CRD, but that's risky given
+// we don't know what else might have the "azureserviceoperator-system" string in it. Instead, we hardcode specific places
+// we know need to be fixed up. This is more brittle in the face of namespace additions but has the advantage of guaranteeing
+// that we can't break our own CRDs with a string replace gone awry.
```
theunrepentantgeek, line 165: :+1:

theunrepentantgeek (approved):

theunrepentantgeek (commented):
Changing to a comment so I get to review your upcoming changes.

Taskfile.yml
```diff
@@ -720,27 +724,47 @@ tasks:
...
       - "kubectl create namespace t1-items"
       - "kubectl create namespace t1-more"
-      # Wait for all the operator pods to be ready.
-      - "kubectl wait --for=condition=ready --timeout=2m pod -n {{.ASO_NAMESPACE}} -l control-plane=controller-manager"
+      # Wait for the other operator to be ready
       - "kubectl wait --for=condition=ready --timeout=2m pod -n tenant1-system -l control-plane=controller-manager"
-      # Wait for the webhook client config caBundles to be set on the
-      # ASO CRDs.
-      - "{{.SCRIPTS_ROOT}}/wait-for-ca-bundles.sh"
+
```
super-harsh, line 741: do we need a `- task: controller:wait-for-operator-ready` here?

v2/charts/azure-service-operator/Chart.yaml
```diff
@@ -6,6 +6,3 @@ home: https://github.com/Azure/azure-service-operator/v2
 sources:
   - https://github.com/Azure/azure-service-operator/v2
 dependencies:
```
super-harsh, line 8: Maybe should remove this as well?

v2/cmd/controller/main.go
```diff
@@ -197,3 +303,96 @@ func makeControllerOptions(log logr.Logger, cfg config.Values) generic.Options {
 			}),
 	}
 }
+
+func getCRDsToSkip(ctx context.Context, logger logr.Logger, k8sConfig *rest.Config, cfg config.Values) (map[string]apiextensions.CustomResourceDefinition, error) {
```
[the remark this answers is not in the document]
matthchr: Changed this to `getReadyCRDs`. I support both now in the `crds` library as well (just invert the comparators and call a single impl). I think this solves your issue and is a bit cleaner.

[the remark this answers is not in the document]
matthchr: Actually upon further investigation, there is a downside to doing it this way:

1. controller_gen.go methods return the set of all resources we understand (GetKnownTypes). This includes `InstalledResourceDefinitions` (because we understand it).
2. `crdManager.LoadOperatorCRDs` returns the set of all CRDs the operator itself is managing. This includes all CRDs _except_ the `InstalledResourceDefinitions` CRD. That CRD is special because the operator doesn't manage it.
3. `crdManager.ListOperatorCRDs` returns the set of all CRDs that the operator finds on APIServer. In the happy path flow this is first _just_ `InstalledResourceDefinitions`, and then after the `InstalledResourceDefinitions` resource is reconciled it's all CRDs.

We effectively are defining the following sets of things:

```
crdsToReconcile = crdManager.LoadOperatorCRDs ∩ crdManager.ListOperatorCRDs
crdsToInstall = crdManager.LoadOperatorCRDs \ crdManager.ListOperatorCRDs
clusterScopeCRDs = controllers.GetClusterScopeStorageTypes
crdsToRunWebhooksOn = ???
```

```
∩ = set intersection
∪  = set union
\ = set difference
```

In order to correctly generate the set of types to run webhooks on we want:
`crdsToReconcile ∪ clusterScopeCRDs` - but the types here don't actually line up, so the easier way to produce this set is actually `controllers.GetKnownTypes \ crdsToInstall`

`crdsToInstall` was the `skip` I had before.

I've kept the other changes but moved this back to a `skip` (set difference) approach because of the above.

v2/charts/azure-service-operator/values.yaml
```diff
@@ -80,7 +80,6 @@ metrics:
   address: 0.0.0.0:{{ .Values.metrics.port }}
   port: 8080
 
-# installCRDs determines if the CRDs subchart should be installed.
 installCRDs: true
```
super-harsh, line 83: Maybe still need a comment here to explain this?

v2/config/default/kustomization.yaml
```diff
@@ -6,12 +6,15 @@ namespace: azureserviceoperator-system
 # "wordpress" becomes "alices-wordpress".
 # Note that it should also match with the prefix (text before '-') of the namespace
 # field above.
-namePrefix: azureserviceoperator-
+#namePrefix: azureserviceoperator-
```
super-harsh, line 9: Remove this if unused?

super-harsh (approved):

v2/internal/config/vars.go
```diff
@@ -111,7 +111,8 @@ var _ fmt.Stringer = Values{}
 // Returns the configuration as a string
 func (v Values) String() string {
 	return fmt.Sprintf(
-		"SubscriptionID:%s/TenantID:%s/ClientID:%s/PodNamespace:%s/OperatorMode:%s/TargetNamespaces:%s/SyncPeriod:%s/ResourceManagerEndpoint:%s/ResourceManagerAudience:%s/AzureAuthorityHost:%s/UseWorkloadIdentityAuth:%t",
+		"SubscriptionID:%s/TenantID:%s/ClientID:%s/PodNamespace:%s/OperatorMode:%s/TargetNamespaces:%s/SyncPeriod:%s"+
+			"/ResourceManagerEndpoint:%s/ResourceManagerAudience:%s/AzureAuthorityHost:%s/UseWorkloadIdentityAuth:%t",
```
theunrepentantgeek, line 115: Would this be better, for readability, if it created a buffer and wrote each name/value pair into that buffer individually?

v2/internal/crdmanagement/crds.go
```diff
@@ -0,0 +1,340 @@
...
+		crdsToApply[name] = crd
+	}
+
+	if len(crdsToApply) > 0 {
+		for _, crd := range crdsToApply {
+			crd := crd
+
+			_, isDifferentVersion := goalCRDsWithDifferentVersion[crd.Name]
+			_, isDifferentSpec := goalCRDsWithDifferentSpec[crd.Name]
+			m.logger.V(Verbose).Info("Found CRD in need of update", "CRD", crd.Name, "SpecDifferent", isDifferentSpec, "VersionDifferent", isDifferentVersion)
```
theunrepentantgeek, line 187: For example, based on the above comment

```suggestion
			m.logger.V(Verbose).Info("Found CRD %d/%d in need of update", "CRD", 
			        index, count, crd.Name, "SpecDifferent", isDifferentSpec, "VersionDifferent", isDifferentVersion)
```

v2/internal/crdmanagement/crds.go
```diff
@@ -0,0 +1,340 @@
...
+				return errors.Wrapf(err, "failed to apply CRD %s", crd.Name)
+			}
+
+			m.logger.V(Debug).Info("Successfully applied CRD", "name", crd.Name, "result", result)
+		}
+
+		// If we make it to here, we have successfully updated all the CRDs we needed to. We need to kill the pod and let it restart so
+		// that the new shape CRDs can be reconciled.
+		m.logger.V(Status).Info("Restarting operator pod after updating CRDs", "count", len(crdsToApply))
+		os.Exit(0) // TODO: Should this be nonzero?
```
theunrepentantgeek, line 211: I think zero is appropriate - this isn't a failure condition, it's normal operation.

v2/internal/crdmanagement/crds_test.go
```diff
@@ -0,0 +1,308 @@
...
+		Strategy: apiextensions.WebhookConverter,
+		Webhook: &apiextensions.WebhookConversion{
+			ClientConfig: &apiextensions.WebhookClientConfig{
+				Service: &apiextensions.ServiceReference{
+					Name:      "azureserviceoperator-webhook-service",
+					Namespace: "azureserviceoperator-system",
+					Path:      to.StringPtr("/convert"),
+					Port:      to.Int32Ptr(443),
+				},
+				CABundle: []byte{17, 14, 12, 21, 33, 61, 25, 99, 111},
```
theunrepentantgeek, line 148: Magic numbers?

v2/internal/crdmanagement/crds_test.go
```diff
@@ -0,0 +1,308 @@
...
```
theunrepentantgeek, line 227: Maybe factor these magic numbers out into an informatively named function?

theunrepentantgeek (approved):

v2/internal/crdmanagement/crds_test.go
matthchr: Hah, yes -- let me leave a comment here. This is just "some random byte data to simulate a CRD bundle".

matthchr: Ah, your function idea below is even better. Will do that.

Taskfile.yml
```diff
@@ -264,14 +266,14 @@ tasks:
     dir: "{{.CONTROLLER_ROOT}}"
     # controller:run-kustomize-for-envtest isn't strictly required here, but running it here so that
     # integration tests can be run manually (via IDE) after running "task"
-    deps: [controller:generate-crds, controller:run-kustomize-for-envtest]
+    deps: [controller:generate-crds, controller:run-kustomize-for-envtest, controller:bundle-crds] # bundle-crds is here because some tests assert on generated CRD shape
```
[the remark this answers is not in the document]
matthchr: Done

Taskfile.yml
```diff
@@ -264,14 +266,14 @@ tasks:
...
     cmds:
     - go test -short -tags=noexit -timeout 10m -run '{{default ".*" .TEST_FILTER}}' ./...
 
   controller:test-cover:
     desc: Run {{.CONTROLLER_APP}} unit tests and output coverage.
     dir: "{{.CONTROLLER_ROOT}}"
-    deps: [controller:generate-crds]
+    deps: [controller:generate-crds, controller:bundle-crds] # bundle-crds is here because some tests assert on generated CRD shape
```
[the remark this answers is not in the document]
matthchr: Done

Taskfile.yml
```diff
@@ -301,10 +303,13 @@ tasks:
       --set azureClientID=$AZURE_CLIENT_ID \
       --set azureClientSecret=$AZURE_CLIENT_SECRET"
     - "kubectl create namespace pre-release"
-    - "kubectl wait --for=condition=ready --timeout=2m pod -n {{.ASO_NAMESPACE}} -l control-plane=controller-manager"
+    # TODO: Can remove the -c parameter once we GA
```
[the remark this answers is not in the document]
matthchr: Done, also explained more why we're using it here and thus why it can be removed when we GA.

Specifically:
```
   # TODO: Can remove the -c parameter once we GA. -c means skip waiting for CRDs to reach established state.
   # TODO: this needs to be skipped because the CRD label for filtering doesn't exist in the old beta charts.
```

Taskfile.yml
```diff
@@ -319,18 +324,19 @@ tasks:
...
-    - "cmd/main.go"
+    - "cmd/controller/main.go"
     cmds:
     - CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build {{.VERSION_FLAGS}} -o ./bin/{{.CONTROLLER_APP}} ./cmd/controller/
 
   controller:docker-build:
     desc: Builds the {{.CONTROLLER_APP}} Docker image.
     dir: "{{.CONTROLLER_ROOT}}"
-    deps: [controller:build]
+    deps: [controller:build, controller:bundle-crds]
```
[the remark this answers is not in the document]
matthchr: Done

v2/internal/crdmanagement/crds.go
matthchr: Removed the TODO

v2/internal/crdmanagement/crds.go
```diff
@@ -0,0 +1,340 @@
...
+	// The same CRD may be in both sets, but we don't want to apply twice, so combine the sets
+	crdsToApply := make(map[string]apiextensions.CustomResourceDefinition)
+	for name, crd := range goalCRDsWithDifferentVersion {
+		crdsToApply[name] = crd
+	}
+	for name, crd := range goalCRDsWithDifferentSpec {
+		crdsToApply[name] = crd
+	}
+
+	if len(crdsToApply) > 0 {
```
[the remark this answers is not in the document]
matthchr: Added `m.logger.V(Status).Info("Will apply CRDs", "count", len(crdsToApply))`

v2/internal/crdmanagement/crds.go
matthchr: Did it slightly differently than this because the logger wants us to put variables into the key/value parameters at the end. I did it like this:
```
			m.logger.V(Verbose).Info(
				"Found CRD in need of update",
				"Progress", fmt.Sprintf("%d/%d", i, len(crdsToApply)),
				"CRD", crd.Name,
				"SpecDifferent", isDifferentSpec,
				"VersionDifferent", isDifferentVersion)
```

v2/internal/config/vars.go
matthchr: It's probably slower, but yeah it's a bit clearer. I've changed.

PR #2772, opened 2023-03-12
Code gardening around astbuilder

theunrepentantgeek:

Refactors some code that should have been using functions from the `astbuilder` package.

Introduced the new func `PointerTo()` as it makes for semantically better reading when the parameter is a type.

gif


super-harsh (approved):

v2/tools/generator/internal/functions/kubernetes_admissions_defaulter.go
```diff
@@ -109,18 +109,14 @@ func (d *DefaulterBuilder) localDefault(k *ResourceFunction, codeGenerationConte
 	for _, def := range d.defaults {
 		defaults = append(
 			defaults,
-			&dst.ExprStmt{
-				X: astbuilder.CallQualifiedFunc(receiverIdent, def.Name()),
-			})
+			astbuilder.InvokeQualifiedFunc(receiverIdent, def.Name()))
```
matthchr, line 112: Not necessarily in this PR but I always have a hard time differentiating `Invoke<x>` vs `Call<x>` - one returns a `Stmt` and one returns an `Expr`. I wonder if, given that's the difference, we ought to just put the name of the return type in the method name?

`CallQualifiedFuncExpr` vs `CallQualifiedFuncStmt` (or maybe just `CallQualifiedFunc`?) might be a bit easier to find the different return-type variants.

Do you have a trick you use to remember which one gives a stmt and which one gives an expr when looking? I think my struggle is that they are "the same" but their first characters don't match so I often find myself finding one or the other, and then being like: "No this isn't the one I want I want the other one" and so I look for other functions with the `Invoke` prefix, not realizing that it's the `Call` prefix that has what I want.

matthchr (approved):

v2/tools/generator/internal/functions/kubernetes_admissions_defaulter.go
theunrepentantgeek: The _trick_ I use is probably not transferrable.

To me `Call` is functional, implying the return value is consumed; `Invoke` implies it's a command where there is no return value.

But that association comes out of my own experience, and it's clearly not a definition that's working for you. 

Happy to include a renaming in this PR if we can agree on better names.

PR #2773, opened 2023-03-13
Adds configuration for which resources are importable

theunrepentantgeek:

Adds configuration allowing us to control which resources can be imported by `asoctl import azure-resource`.

Initially, we'll use this to suppress all resources so that the initial commits of the feature are reasonably sized and can be effectively reviewed.

Once we're happy with the new code being generated, we'll remove all (or almost all) the suppressions, leaving the `$importable` flag available for any problematic resources we discover in the future.

gif


v2/azure-arm.yaml
```diff
@@ -837,12 +864,14 @@ objectModelConfiguration:
         $azureGeneratedSecrets:
           - AdminCredentials
           - UserCredentials
+        $importable: false
```
matthchr, line 867: minor: Might be helpful to always put this after the `$supportedFrom`? My eyes at least have a harder time parsing these changes when it's after a (possibly long-ish) list of other things.

matthchr (approved):

v2/azure-arm.yaml
theunrepentantgeek: Good suggestion - but I'll leave things the way they are, given I'll shortly be removing all of these.

PR #2775, opened 2023-03-13
Restructure asoctl command line argument handling

theunrepentantgeek:

As discussed offline, we don't need the separate commands of `asoctl` divided into separate packages, so I'm merging them into the `cmd` folder.

Also renames `export` to `import` and updates command descriptions.

gif


v2/tools/asoctl/cmd/crd.go
```diff
@@ -3,12 +3,12 @@
...
 
-package crd
+package main
 
 import "github.com/spf13/cobra"
 
-// NewCommand creates a new cobra to clean deprecated crd versions
-func NewCommand() (*cobra.Command, error) {
+// newCRDCommand creates a new cobra command for working with CRDs
+func newCRDCommand() (*cobra.Command, error) {
```
matthchr, line 11: As I alluded to in Teams, I feel like this is _less_ idiomatic Go than we had before. AFAIK, Go has no issue with small packages, and having them means we can hide impl details with internal methods/functions rather than having all commands see all other commands implementation internals.

I'm not sold that there was any issue with the structure we had. What was the problem with it?

For example the change below:
```
		- export.NewCommand,
		- crd.NewCommand,
		+ newCRDCommand,
		+ newImportCommand,
```

where these are used feels like a downgrade to me.

I'm not super religious about this I just feel like we're going to end up undoing this change if/when we add 2-3 more commands and the `cmd` folder becomes bigger.

theunrepentantgeek: We've previously agreed that all the heavy lifting should go into `internal` packages, so the only stuff that ends up in `cmd` are the launching points. 

Put another way, we should never end up with all commands seeing the internal implementation details of each other because those details should never be here to begin with.

theunrepentantgeek: We should end up with a very small number of files here - one for each top level verb, one for each second level command, and maybe one for each third level if we end up needing them.

matthchr: > but having a separate package for each trio of files, all of them under 50 lines, seems more like obfuscation than anything else.

I'm not sure I understand how it's obfuscation? F12 still takes you there, right? And the import at the top of the file clearly shows where that package is being imported from?

The `cmd` folders that I usually see are pretty much just a `main.go`. Here are the first 3 tools I thought of and all have only main.go (or in 1 case main.go and version.go) at the top level:
- [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder/tree/master/cmd)
- [clusterctl](https://github.com/kubernetes-sigs/cluster-api/tree/main/cmd/clusterctl), which has only main.go at the top level but has a substructure under that (which we could do).
- [certctl](https://github.com/cert-manager/cert-manager/tree/master/cmd/ctl), which also has main.go at the top level but has a substructure.

Looking into certctl's substructure, you can see there are plenty of single-file packages: <100 line packages, for example: https://github.com/cert-manager/cert-manager/tree/master/cmd/ctl/pkg/inspect

Unlike `clusterctl` and `certctl` we have `tools/asoctl/cmd` as opposed to `cmd/asoctl` and so our structure is a bit different. If we had `cmd/asoctl/main.go` then I think we'd want other asoctl packages to be `cmd/asoctl/<packages>` (in that directory).

Is the obfuscation issue just that they're in a different directory than `main`?

theunrepentantgeek: Heh. `clusterctl` does things essentially the way that I want - the only difference is that they've moved `main.go` up to the root, instead of leaving it under `cmd` - which I like, because then you can `go build` in the root folder and it works.

`certctl` takes a different approach, and seems to comingle the commands with the implementation, which (in my mind) raises testability concerns.

And `kubebuilder` takes a wholly different approach with a plugin model.

matthchr (approved):
as discussed, fine with this especially if main.go moves, although I think we should consider moving `asoctl` up to `v2/cmd/asoctl` as it's an externally exported exe we ship.

PR #2788, opened 2023-03-17
Expand path enums

### LINKED ISSUE #2787: Bug: Fail to import resources with enums in operation path
**Describe the bug**

We currently only support importing resources with enumerations in their PUT operation path if those enumerations have exactly one value.

This means we're failing to import DNS record types, as that operation path has multiple values.

**To Reproduce**

Put a breakpoint in `swagger_type_extractor.go`, on the following code (starting at line #546) and look at what happens when `isEnum` is true, and there is more than one option available.

``` go
		enum, isEnum := paramType.(*astmodel.EnumType)
		if isEnum && len(enum.Options()) == 1 {
			theOption := strings.Trim(enum.Options()[0].Value, "\"")
			toReplace := fmt.Sprintf("{%s}", resolved.Name)
			result = strings.ReplaceAll(result, toReplace, theOption)
		}
```

**Additional context**

Discovered while investigating a problem encounted by @super-harsh.

In `dns.json` on L#30 we find following operation path:

```
/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/
dnsZones/{zoneName}/{recordType}/{relativeRecordSetName}
```

The `recordType` parameter is an enumeration with these values (L#323):

``` json
          {
            "name": "recordType",
            "in": "path",
            "required": true,
            "type": "string",
            "description": "The type of DNS record in this record set. Record sets of type SOA cannot be deleted (they are deleted when the DNS zone is deleted).",
            "enum": [
              "A",
              "AAAA",
              "CAA",
              "CNAME",
              "MX",
              "NS",
              "PTR",
              "SOA",
              "SRV",
              "TXT"
            ],
            "x-ms-enum": {
              "name": "RecordType",
              "modelAsString": false
            }
```

Because there are too many values, we're incorrectly rejecting this as a resource.

Full path to `dns.json`: `v2\specs\azure-rest-api-specs\specification\dns\resource-manager\Microsoft.Network\stable\2018-05-01\dns.json`

**Possible Fix**
Speculating on a possible change.

* Modify `SwaggerTypeExtractor.expandEnumsInPath()` to return `[]string`, returning all the possible expansions of the path. 
  * Would need to handle the combinatorial results of having multiple enum parameters).
  * Possibly would also need to return the parameter values used, for construction of the resource name
* Modify `SwaggerTypeExtractor.ExtractResourceTypes()` to generate a unique resource type for each path returned by `expandEnumsInPath()`.
  * The bulk of this method, from after the call to `expandEnumsInPath()` on L#133 until the end of the loop on L#205 would likely factored out into a method that handles a single operation path, perhaps `extractResourceType()`.

matthchr:
Closes #2787


theunrepentantgeek (approved):
Looks good. I was expecting `inferNameFromURLPath()` to need updates too, but I guess it works as is.

super-harsh (approved):

PR #2790, opened 2023-03-19
Document issues with downgrades

### LINKED ISSUE #1446: How do upgrades of the Service Operator work?
A bunch of issues to consider here, including:

* How do we recover from a failed upgrade? Do we try the upgrade again or can we revert to an earlier verison of the operator?
* How does the Operator Lifecycle Manager play into this? Can it prevent upgrades that won't work?
* When do the web-hooks for upgrades trigger? 
* Is an upgrade triggered all at once or JIT? If JIT, how do we roll back if it fails?

theunrepentantgeek:

Documents how to downgrade ASO and the likely issues that users might run into if they need to use this remedy of last resort.

Closes #1446 (Everything else mentioned in #1446 has already been discussed and/or documented)

gif

**If applicable**:


super-harsh (approved):

docs/hugo/content/introduction/upgrading.md
```diff
@@ -52,3 +52,27 @@ helm upgrade --devel --version v2.0.0-beta.3 aso2 aso2/azure-service-operator \
...
+
+Each ASO resource is stored within the cluster using a canonical storage (or hub) version, regardless of the actual version used for interaction with the cluster or with Azure. These stored versions are based on the latest Stable Azure API version of that resource. (See [Resource Versioning](https://azure.github.io/azure-service-operator/design/versioning/) for all the background on this design decision.)
+
+When a new version of ASO adds support for a later version of an already supported resource (say, adding support for the latest release of Azure Redis, v2022-06-01, in addition to the existing support for v2021-12-01), then the new version of ASO will use the newer API version of the resource as the storage version. In this situation, downgrading ASO to the prior version will orphan the resource, as the older version of ASO is unaware of the newer storage version and is unable to interact with it. This means that any resources created with the newer version of ASO will be unable to be managed by the older version of ASO. They will also be inaccessible via `kubectl` or any other tool.
```
matthchr, line 64: minor: Should refer to CRD versions here, which would be something like `v1api20220601`, rather than ARM API versions, since you're talking about CRDs

matthchr, line 64: It's not just any version created with the newer version. It's any version created or updated after the new ASO version ( with the updated storage version) was installed. There's sorta a matrix or list here:

1. Was the ASO operator pod ever running successfully on the new version? If yes, resources may have been reconciled, which means they would be at the new storage version. Downgrade is unsafe.
2. Did a client tool _successfully_ issue a `kubectl apply` or raw HTTP PUT on any CR? If yes, resources may have been updated to the new storage version. Downgrade is unsafe.*

* Note that, if the operator pod webhooks are down `kubectl apply` of any resource would fail as the webhooks are needed to convert to the storage version. In this case, downgrade is safe.

docs/hugo/content/introduction/upgrading.md
```diff
@@ -52,3 +52,27 @@ helm upgrade --devel --version v2.0.0-beta.3 aso2 aso2/azure-service-operator \
...
+
+These orphaned resources can be recovered by reapplying the resource (using `kubectl apply`) and overwriting the resource with a new version. Alternatively, if you are going to be re-upgrading to ASO soon, you can leave those resources in place. The older version of ASO will ignore them, and any newer version should pick them up and start reconciliation automatically.
```
matthchr, line 66:
```suggestion
These orphaned resources can be recovered by reapplying the resource (using `kubectl apply`) and overwriting the resource with a new version after the downgrade. Alternatively, if you are going to be re-upgrading to ASO soon, you can leave those resources in place. The older version of ASO will ignore them, and any newer version should pick them up and start reconciliation automatically.
```

docs/hugo/content/introduction/upgrading.md
```diff
@@ -25,7 +25,7 @@ For example `v2.0.0-alpha.6` to `v2.0.0-beta.1` should be accomplished with two
 
 We recommend that you upgrade ASO using the same tool you installed it with. 
 
-**Note:** The instructions below all assume you're upgrading from the previous version to the latest version. Please follow 
+**Note:** The instructions below all assume you're upgrading from the immediatey prior version to the latest version. Please follow
```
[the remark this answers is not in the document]
theunrepentantgeek: Done.

docs/hugo/content/introduction/upgrading.md
```diff
@@ -52,3 +52,27 @@ helm upgrade --devel --version v2.0.0-beta.3 aso2 aso2/azure-service-operator \
 ## Supported Versions
 
 We provide best effort support via GitHub for ASO, focusing on the latest version and the previous version. If you're running an older version, we may ask you to upgrade to the latest release before we can help you.
+
+## Downgrading
+
+If you encounter an issue after an upgrade of ASO that can't be addressed by any other means, you may want to downgrade to the prior version. This isn't an officially supported scenario, but we'll try to provide some guidance here.
+
+Summary: If the new version of ASO didn't change the stored version of your resource, then you should be able to downgrade without issue. If the new version of ASO did change the stored version of your resource, then you will need to manually fix things up.
```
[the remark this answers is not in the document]
theunrepentantgeek: Users should be able to work it out from the release notes or the list of supported resources.

docs/hugo/content/introduction/upgrading.md
theunrepentantgeek: Done.

theunrepentantgeek: The actions required by a user are the same regardless of how the resource was updated. Also, there's no easy visibility (and any answer might change between running a `kubectl` command and when the downgrade occurs), so I think adding this level of detail would be counterproductive.

docs/hugo/content/introduction/upgrading.md
```diff
@@ -52,3 +52,27 @@ helm upgrade --devel --version v2.0.0-beta.3 aso2 aso2/azure-service-operator \
...
+When a new version of ASO does not change the stored version of a resource (say, adding support for an old version of Azure Redis, v2020-12-01, in addition fo the existing support for v2021-12-01) then the new version of ASO will be using the same API version of the resource as the prior release. In this situation, downgrading ASO to the prior version should work without issue.
+
+Given that it's unusual for a release of ASO to upgrade all resources, it's likely that you'll find both of the above scenarios in your cluster, depending on the variety of ASO supported resource types in play.
+
+### Preparation for downgrading
+
+For each ASO managed resource in your cluster
+
+* Check the ASO release notes to see if the ASO upgrade added a new version for that resource type.
+* Download the YAML definition for the resource using `kubectl get` so you have a basis for recovery if needed.
```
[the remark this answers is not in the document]
theunrepentantgeek: Updated the wording.

docs/hugo/content/introduction/upgrading.md
matthchr: My concern is specifically around this sentence:

> This means that any resources **created** with the newer version of ASO will be unable to be managed by the older version of ASO.

(emphasis mine). That makes it sound like you're safe if you don't create new resources.

Should we rephrase to:
> This means that any resources with updated storage versions in the newer version of ASO will be unable to be managed by the older version of ASO.

and just not say "created"?

matthchr (approved):
LGTM.

Have you tested this btw?

It might not hurt to just make 100% sure we're right about what the behavior here is. I'm pretty sure we are, but it should be pretty easy to manually test on a kind cluster with a few resources. Might also be able to then show exactly what error the user would see, which might help.

docs/hugo/content/introduction/upgrading.md
theunrepentantgeek: I've reworded the sentence to this (emphasis added):

> This means that any resources created with, _or modified by,_ the newer version of ASO will be unable to be managed by the older version of ASO.

PR #2794, opened 2023-03-21
Remove dependency on xcobra

theunrepentantgeek:

Discovered while debugging `asoctl`, if we use the `xcobra` library then any errors we return from command handlers are silently lost. Given that `cobra` now has support for running commands with Context, simplest fix seems to be to remove it.

Not sure about factoring out the interrupt-context signal handler into a common library, open to suggestions.

gif


super-harsh (approved):

v2/tools/generator/root.go
```diff
@@ -17,7 +21,17 @@ func Execute() {
 		klog.Fatalf("fatal error: commands failed to build! %s\n", err)
 	}
 
-	if err := cmd.Execute(); err != nil {
+	ctx, cancel := context.WithCancel(context.Background())
```
matthchr, line 24: Any reason not to put this bit into a shared function?

`makeSignalCtx` or something?

matthchr (approved):

v2/tools/generator/root.go
theunrepentantgeek: Good idea. I just need to work out where that shared function should live.

theunrepentantgeek: Done.

PR #2796, opened 2023-03-21
Improve resource path canonicalization

matthchr:
Fix issue where infer name would include parameters which were the resource "name" such as "current", when really we only want to include parameters which define resource type.

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/tools/generator/internal/jsonast/swagger_type_extractor.go
+++ b/v2/tools/generator/internal/jsonast/swagger_type_extractor.go
@@ -129,7 +129,7 @@ func (extractor *SwaggerTypeExtractor) ExtractResourceTypes(ctx context.Context,
 		fullPutParameters := append(op.Parameters, op.Put.Parameters...)
 
 		nameParameterType := extractor.getNameParameterType(ctx, rawOperationPath, scanner, fullPutParameters)
-		operationPaths := extractor.expandEnumsInPath(ctx, rawOperationPath, scanner, fullPutParameters)
+		operationPaths := extractor.expandAndCanonicalizePath(ctx, rawOperationPath, scanner, fullPutParameters)
 
 		for _, operationPath := range operationPaths {
 			err := extractor.extractOneResourceType(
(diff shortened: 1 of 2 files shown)


v2/tools/generator/internal/jsonast/swagger_type_extractor.go
```diff
@@ -552,9 +578,43 @@ func (extractor *SwaggerTypeExtractor) expandEnumsInPath(
...
+			continue
+		}
+
+		if urlPart[0] != '{' {
+			// Not a parameter, so by definition doesn't need to be expanded
+			continue
+		}
+
+		paramName := strings.Trim(urlPart, "{}")
+		param, ok := paramMap[paramName]
```
theunrepentantgeek, line 611: Does Swagger require things to match by case? Or should this be a case-insensitive lookup?

v2/tools/generator/internal/jsonast/swagger_type_extractor.go
```diff
@@ -569,19 +629,40 @@ func (extractor *SwaggerTypeExtractor) expandEnumsInPath(
...
+			// We only expand enums, skip everything else
+			continue
+		}
+
+		if len(enum.Options()) == 0 {
+			// This shouldn't happen but guard against it anyway
+			continue
+		}
+
+		// resource types come on the even counts: Microsoft.Group/resourceType/{resourceName}/{childResourceTypeParameter}/{childName}
```
theunrepentantgeek, line 642: :thinking: ... "even counts" assumes you treat `Microsoft.Group` as item 0, but not if you start counting from 1.

Not sure how to clarify though ...

v2/tools/generator/internal/jsonast/swagger_type_extractor.go
```diff
@@ -569,19 +629,40 @@ func (extractor *SwaggerTypeExtractor) expandEnumsInPath(
...
+		isResourceTypeParameter := i%2 == 0
+		isDefaultEnum := false
+		if len(enum.Options()) == 1 {
+			opt := enum.Options()[0]
+			theOption := strings.Trim(opt.Value, "\"")
```
theunrepentantgeek, line 647: Are double quotes (`"`) always used or are other forms sometimes used? I'm thinking primarily of single quotes (`'`) but also of more exotic kinds.

theunrepentantgeek (approved):

v2/tools/generator/internal/jsonast/swagger_type_extractor.go
matthchr: When in doubt, give a better example:

```
		// resource types come on the even indices after the group.
		// Example: Microsoft.Group/resourceType/{resourceName}/{childResourceTypeParameter}/{childName},
		// has "resourceType" at index=0, "{childResourceTypeParameter}" at index=2
```

v2/tools/generator/internal/jsonast/swagger_type_extractor.go
matthchr: For strings, yes, `"` is always used. That's an artifact of how we store enums in the code generator, not how they are specified in Swagger.

They are stored that way because they become string literals (which have "'s around them).

There are some enum types (ints, etc) where the value isn't a string literal and so doesn't have quotes, but those values won't be equal to "default" (the only enum we expand) so this comparison is safe.

PR #2807, opened 2023-03-24
Refactor generic client to remove SubscriptionID and make metrics optional

matthchr:
**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/api/cache/customizations/redis_extensions.go
+++ b/v2/api/cache/customizations/redis_extensions.go
@@ -57,7 +57,7 @@ func (ext *RedisExtension) ExportKubernetesResources(
 	var accessKeys armredis.AccessKeys
 	// Only bother calling ListKeys if there are secrets to retrieve
 	if hasSecrets {
-		subscription := armClient.SubscriptionID()
+		subscription := id.SubscriptionID
 		// Using armClient.ClientOptions() here ensures we share the same HTTP connection, so this is not opening a new
 		// connection each time through
 		var redisClient *armredis.Client
--- a/v2/internal/metrics/arm_client_metrics.go
+++ b/v2/internal/metrics/arm_client_metrics.go
@@ -6,20 +6,11 @@
 package metrics
 
 import (
-	"net/http"
 	"strconv"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
-)
-
-type HTTPMethod string
-
-const (
-	HttpPut    HTTPMethod = http.MethodPut
-	HttpDelete HTTPMethod = http.MethodDelete
-	HttpGet    HTTPMethod = http.MethodGet
 )
 
 type ARMClientMetrics struct {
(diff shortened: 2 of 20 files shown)


v2/internal/reconcilers/arm/arm_client.go
```diff
@@ -33,3 +40,19 @@ func (c *armClient) GenericClient() *genericarmclient.GenericClient {
 func (c *armClient) CredentialFrom() string {
 	return c.credentialFrom.String()
 }
+
+// ARMClientDetails is the type returned by the ARMClientCache containing the actual client and some details about the destination
+// of the resource
+type ARMClientDetails struct {
```
matthchr, line 46: I am not in love with this type name.

It's not really _just_ `ClientDetails` it feels more like `DestinationDetails` or something? It has a client obviously, and also the target subscription, as well as the location that the credentials being used came from.

`AuthenticationDetails`?

It's a bit of a grab-bag but at the same time feels like an actual construct because all of the stuff is about:
1. Where we got the secret we're using.
2. What we read out of that secret (including a client constructed from the secret)

v2/api/cache/customizations/redis_extensions.go
```diff
@@ -57,7 +57,7 @@ func (ext *RedisExtension) ExportKubernetesResources(
 	var accessKeys armredis.AccessKeys
 	// Only bother calling ListKeys if there are secrets to retrieve
 	if hasSecrets {
-		subscription := armClient.SubscriptionID()
+		subscription := id.SubscriptionID
```
theunrepentantgeek, line 60: Nice that `SubscriptionID` was already available from another source.

v2/internal/metrics/arm_client_metrics.go
```diff
@@ -60,16 +51,16 @@ func (a *ARMClientMetrics) RegisterMetrics() {
 }
 
 // RecordAzureSuccessRequestsTotal records the total successful number requests to ARM by increasing the counter.
-func (a *ARMClientMetrics) RecordAzureSuccessRequestsTotal(resourceName string, statusCode int, method HTTPMethod) {
+func (a *ARMClientMetrics) RecordAzureSuccessRequestsTotal(resourceName string, statusCode int, method string) {
```
theunrepentantgeek, line 54: Why go for the less well typed parameter?

v2/internal/reconcilers/arm/arm_client.go
theunrepentantgeek: Maybe `ARMConnectionDetails` or maybe just `ARMConnection` ?

theunrepentantgeek (approved):
Looks good to me. More involved than I expected, but looks like it came out cleanly.

v2/internal/metrics/arm_client_metrics.go
matthchr: My reasoning here is that it's what the actual `http.Request` type has for `Method`, and so when writing the `metrics_policy` if we had an enum, I would need to write an:
```
if method == http.PUT {
    typedMethod = HTTPPut
}
...
```

or a switch of the same.

That felt more awkward than just doing `method := raw.Method`. Not that we couldn't do the above but if the Go stdlib authors don't care to do it, it felt awkward for us to bolt it on above as it made the interface between us and HTTP more awkward.

v2/internal/reconcilers/arm/arm_client.go
matthchr: Yeah I like that more.

Since the module is ARM I actually dropped the prefix too and just called it `arm.Connection`

v2/internal/metrics/arm_client_metrics.go
theunrepentantgeek: Fair enough.

v2/internal/reconcilers/arm/arm_client.go
theunrepentantgeek: Nice.

super-harsh (approved):

PR #2812, opened 2023-03-28
Design document for Post Reconciliation Extension

super-harsh:

This PR adds a design document for Post Reconciliation Extension to facilitate the reconciler to run checks after a successful reconcile on a resource

**If applicable**:
- [ ] this PR contains tests


docs/hugo/content/ADR-2023-03-Post-Reconciliation-Extension.md
```diff
@@ -0,0 +1,46 @@
+---
+title: '2023-03: Post Reconciliation Extensions'
+---
+
+## Context
+
+The current behaviour of Azure Service Operator (ASO) is to set the `Ready` status on a resource by default after a successful creation of a resource. 
+
+For some resources, this is not ideal as we need to check the status of a created resource post reconciliation and set a condition accordingly.
```
theunrepentantgeek, line 9: Remember the extension point is generic - while our current use is for `PrivateEndpoint`, we're not designing it to handle that case only. The design and the description need to allow for many uses - the defining characteristic of this extension point is _when_ it gets applied, not _which_ resources it's applied to.

```suggestion
For some resources, this is not sufficient as there may be additional steps required after initial creation of the resource. We need the ability to check the status of a created resource post reconciliation and set a condition accordingly. 
```

docs/hugo/content/ADR-2023-03-Post-Reconciliation-Extension.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Context
+
+The current behaviour of Azure Service Operator (ASO) is to set the `Ready` status on a resource by default after a successful creation of a resource. 
+
+For some resources, this is not ideal as we need to check the status of a created resource post reconciliation and set a condition accordingly. 
+
+An example of this:
+
+- `PrivateEndpoints` with `ConnectionState` of `Approved`, `Rejected` or `Pending`.
```
theunrepentantgeek, line 13: If there's one example, put it in  sentence. Only use a bullet point if there's more than one.

docs/hugo/content/ADR-2023-03-Post-Reconciliation-Extension.md
```diff
@@ -0,0 +1,46 @@
...
+
+We need a way to allow the operator to handle the above case in a way where we can write an extension point to be called post reconciliation. Which allows the operator to set a custom condition on the resource based on the status we've received from ARM.
```
theunrepentantgeek, line 15:
```suggestion
We need a way to extend the operator to handle the above case, an extension point to be called post reconciliation allowing handwritten code to set a custom condition on the resource based on the status we've received from ARM.
```

docs/hugo/content/ADR-2023-03-Post-Reconciliation-Extension.md
```diff
@@ -0,0 +1,46 @@
...
+- A Kubernetes client allowing for Cluster operations.
+- An ARM client allowing for ARM operations.
+- A logger to allow for tracing of what the extension did.
+- A context to allow cancellation of long-running operations.
+
+The above would return one of two below options:
+
+- `Success` if the post reconcile check is met.
+- `Failure` with `NewReadyConditionImpactingError` if the post reconcile check is not met.
+
```
theunrepentantgeek, line 33: What does the return value mean?

If an implementation of this extension returns `Success`, what happens? How is that return value structured?
If it returns `Failure`, what does that mean for the reconciler? 
How does the `NewReadyConditionImpactingError` get returned? 
Is it possible for other errors to be returned? How?

theunrepentantgeek (changes requested):

docs/hugo/content/ADR-2023-03-Post-Reconciliation-Extension.md
super-harsh: Done!

docs/hugo/content/ADR-2023-03-Post-Reconciliation-Extension.md
super-harsh: Have elaborated more on this.

docs/hugo/content/design/ADR-2023-03-Post-Reconciliation-Extension.md
```diff
@@ -0,0 +1,45 @@
...
+title: '2023-03: Post Reconciliation Extensions'
+---
+
+## Context
+
+The current behaviour of Azure Service Operator (ASO) is to set the `Ready` status on a resource by default after a successful creation of a resource. 
+
+For some resources, this is not sufficient as there may be additional steps required after initial creation of the resource. We need the ability to check the status of a created resource post reconciliation and set a condition accordingly. 
+
+An example of this is `PrivateEndpoints` with `ConnectionState` of `Approved`, `Rejected` or `Pending`.
```
theunrepentantgeek, line 11:
```suggestion
An example of this is `PrivateEndpoints` with `ConnectionState` of `Approved`, `Rejected` or `Pending`. We don't want a `PrivateEndpoint` to go to condition `Ready` until the connection is approved.
```

theunrepentantgeek (approved):
This looks good. In an [earlier ADR](https://azure.github.io/azure-service-operator/design/adr-2022-01-reconciler-extensions/) we included the interface definition of the extension point, maybe do the same here?

docs/hugo/content/design/ADR-2023-03-Post-Reconciliation-Extension.md
```diff
@@ -30,6 +30,95 @@ The return will be one of three possibilities:
...
+We create the required extension interface in the package genruntime/extensions:
+
+``` go
+// postreconciliation_checker.go
+
+package extensions
+
+type PostReconciliationChecker interface {
+	// PostReconcileCheck does a post-reconcile check to see if the resource is in a state to set 'Ready' condition.
+	// ARM resources should implement this to avoid reconciliation attempts that cannot possibly succeed.
```
theunrepentantgeek, line 46: I think this comment is stale.

docs/hugo/content/design/ADR-2023-03-Post-Reconciliation-Extension.md
```diff
@@ -30,6 +30,95 @@ The return will be one of three possibilities:
...
+	if endpoint, ok := obj.(*network.PrivateEndpoint); ok && endpoint.Status.PrivateLinkServiceConnections != nil {
+
+		for _, connection := range endpoint.Status.PrivateLinkServiceConnections {
+			if *connection.PrivateLinkServiceConnectionState.Status != "Approved" {
+				// Returns 'conditions.NewReadyConditionImpactingError' error
+				return extensions.PostReconcileCheckResultFailure(
+					fmt.Sprintf(
+						"Private connection '%s' to the endpoint requires approval %q",
+						*connection.Id,
+						*connection.PrivateLinkServiceConnectionState)), nil
```
theunrepentantgeek, line 112: Not an issue for this spec, but should this aggregate all errors - in case there are multiple connections requiring approval? Might help our users avoid playing config-what-a-mole.

theunrepentantgeek (approved):
Looks good.

docs/hugo/content/design/ADR-2023-03-Post-Reconciliation-Extension.md
super-harsh: I'll just let this be as is here, as this is an example and is more clear this way as an example. However, makes sense what you mentioned above.. Will modify the actual code in the other PR.

PR #2815, opened 2023-03-28
Remove deprecated alpha CRD versions and add v1

### LINKED ISSUE #2720: Feature: Create GA resource versions
We need to create the GA resource versions _and_ delete the alpha resource versions as part of the v2.0.0 release.

matthchr:
Closes #2720

Removes the v1alpha1 CRDs versions.
Adds a new v1api CRD version, including samples for those new versions. These versions are exact clones of the v1beta versions but have the new v1api stable version prefix.
Update tests to use the v1api stable versions

Don't try to review this PR all at once, go commit by commit and skip the "Regenerate types" commit. The rest of the commits should be reasonably reviewable.

**If applicable**:
- [x] this PR contains tests


v2/tools/generator/internal/codegen/pipeline/detect_skipping_properties.go
```diff
@@ -253,11 +258,131 @@ func (detector *skippingPropertyDetector) lookupNext(ref astmodel.PropertyRefere
...
+		return false
+	}
+
+	packageRefs := set.Make[astmodel.PackageReference]()
+	for name := range right {
+		packageRefs.Add(name.PackageReference)
+	}
+
+	if len(packageRefs) != 1 {
+		panic("oops")
```
super-harsh: We need a better panic error here?

v2/tools/generator/internal/codegen/pipeline/detect_skipping_properties_test.go
```diff
@@ -151,6 +151,40 @@ func Test_DetectSkippingProperties_WhenPropertyTypesIdentical_ReturnsNoError(t *
...
+		NewState().WithDefinitions(defs),
+		CreateStorageTypes(),            // First create the storage types
+		CreateConversionGraph(cfg, "v"), // Then, create the conversion graph showing relationships
+	)
+	g.Expect(err).To(Succeed())
+
+	_, err = RunTestPipeline(
+		initialState,
+		DetectSkippingProperties(), // and then we get to run the stage we're testing
+	)
```
theunrepentantgeek: I don't think you need to use `RunTestPipeline()` twice - it's Ok to assume the initial stages run properly.

v2/tools/generator/internal/codegen/pipeline/report_resource_versions.go
```diff
@@ -410,6 +410,12 @@ func (report *ResourceVersionsReport) generateSupportedFrom(typeName astmodel.Ty
 		return "v2.0.0-beta.0", nil
 	}
 
+	// Special case for resources that existed prior to GA
+	// the `v1` versions of those resources are only available from "v2.0.0"
```
theunrepentantgeek: Suggest 

```
// the `v1api` versions of those resources are only available from "v2.0.0"
```

v2/tools/generator/internal/astmodel/package_reference.go
```diff
@@ -35,6 +35,15 @@ type PackageReference interface {
 	GroupVersion() (string, string)
 }
 
+type LocalLikePackageReference interface {
```
theunrepentantgeek: Not sure why this interface is needed.

v2/tools/generator/internal/codegen/code_generator.go
```diff
@@ -154,7 +154,7 @@ func createAllPipelineStages(idFactory astmodel.IdentifierFactory, configuration
 		pipeline.AddSecrets(configuration).UsedFor(pipeline.ARMTarget),
 		pipeline.AddConfigMaps(configuration).UsedFor(pipeline.ARMTarget),
 
-		pipeline.CreateTypesForBackwardCompatibility("v2.0.0-alpha", configuration.ObjectModelConfiguration).UsedFor(pipeline.ARMTarget),
+		pipeline.CreateTypesForBackwardCompatibility("v2.0.0-", configuration.ObjectModelConfiguration).UsedFor(pipeline.ARMTarget),
```
theunrepentantgeek: Shouldn't this be `v2.0.0-beta` - I think it's the prefix for the older versions.

scripts/check_samples.py
```diff
@@ -42,7 +42,7 @@ def get_expected_samples(crd_path):
         group = group.split(' ')[1]
         group = strip(group, ".azure.com")
         version = version.split(' ')[1]
-        versionPrefix = re.match('v1(api|alpha1|beta)([0-9][a-z]+)?', version).group() + "*"
+        versionPrefix = re.match('v1(api|alpha1|beta)?([0-9][a-z]+)?', version).group() + "*"
```
theunrepentantgeek, line 45: Why do we need `api|alpha1|beta` to be optional?

theunrepentantgeek (approved):
A few minor questions.

scripts/check_samples.py
matthchr: because the handcrafted MySQL User resource (and other handcrafted resources in the future that don't correspond to an azure API verison) will be called `v1` only, not `v1api`.

v2/tools/generator/internal/codegen/pipeline/detect_skipping_properties_test.go
matthchr: I think I just copied this from the test you wrote above =P

Will take a look and see if I can combine

matthchr: Fixed

v2/tools/generator/internal/codegen/pipeline/report_resource_versions.go
matthchr: Fixed

v2/tools/generator/internal/astmodel/package_reference.go
matthchr: I added the following comment:
```
// LocalLikePackageReference describes a package reference that points to a local package (either a storage package
// or a standard one). It can be used to abstract across the exact package type (storage vs local)
```

It's used like this:
```
	leftLPR, isLeftLocalRef := left.PackageReference.(astmodel.LocalLikePackageReference)
	rightLPR, isRightLocalRef := right.PackageReference.(astmodel.LocalLikePackageReference)
```

In `detect_skipping_properties.go`.

Basically there's code that needs to treat `PackageReference` the same as long as it's one of the two local flavors (Storage | Local).

Is there a better way to do this?

v2/tools/generator/internal/codegen/code_generator.go
matthchr: No because it needs to match anything that had a beta version, and we had a beta version for anything that had an alpha version, and anything that was added in beta.

So we need to match these two things:
`supportedFrom: v2.0.0-alpha.5` (added _before_ beta, so has a beta version)
`supportedFrom: v2.0.0-beta.0` (added _during_ beta, so has a beta version)

The only thing we don't want backcompat types for is `v2.0.0` (or `v2.1.0`, etc)

I've added a comment here to that effect for clarity.

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -68,17 +68,19 @@ func (c *Cleaner) Run(ctx context.Context) error {
...
+			return errors.New(fmt.Sprintf("it doesn't look like your version of ASO is one that supports deprecating version %q. Have you upgraded ASO yet?", deprecatedVersion))
 		}
 
+		activeVersion := newStoredVersions[0]
+
 		// If we found an updated slice, which implies, we have found a version to deprecate. Then only continue with the cleaning process
 		if len(newStoredVersions) != len(crd.Status.StoredVersions) {
 			klog.Infof("Starting cleanup for %q", crd.Name)
-			objectsToMigrate, err := c.getObjectsForMigration(ctx, crd, matchedStoredVersion)
+			objectsToMigrate, err := c.getObjectsForMigration(ctx, crd, activeVersion)
```
super-harsh: Why would we get objects for migration with activeVersion?

super-harsh: We're supposed to get the objects for migration, which are under deprecatedVersion, and then issue a PUT onto them?

matthchr: So the issue is that we want `asoctl` to successfully be able to remove deprecated API versions even if the user is running a newer `asoctl` and they're on an older cluster. 

For example, when when built with `v2.0.0`, or even `v2.1.0`.

The code we had before though, was trying to use the deprecated api version to perform the `c.client.List`. The issue is, that deprecated version isn't in the `scheme` of `v2.0.0` (because it's been removed/deprecated), so that list will fail.

TL;DR: When migrating CRs, we need to use an _active_ version to migrate them, not a deprecated version.

v2/tools/generator/internal/astmodel/package_reference.go
matthchr: I looked into `astmodel.IsExternalPackageReference`, but using that in this situation is awkward because:

1. I need access to `LocalPathPrefix` in addition to `Group`, so I would still need to do 2 casts (one for `LocalPackageReference` and one for `StoragePackageReference`)
2. Performing this check by looking for the "set of things they aren't" is more confusing than a positive assertion. That is: `IsLocalLikePackageReference` should be preferred over `!ExternalPackageReference()` - who knows if/when we add another type and that `!IsExternalPackageReference()` is now a bug.

Additionally, our exisitng `PackageReference` interface shape has always been a peeve of mine, specifically these methods:
```
	// IsPreview returns true if this package reference has a suffix indicating it's a preview
	// release, false otherwise
	IsPreview() bool
	// TryGroupVersion returns the group and version of this reference.
	// Returns true if the reference has a group and version, false otherwise.
	TryGroupVersion() (string, string, bool)
	// GroupVersion returns the group and version of this reference, triggering a panic if they aren't available
	GroupVersion() (string, string)
```

As all of them only make sense for `LocalLikePackageReferences` but we have them on the generic interface even when they don't apply to all that interfaces implementers.

Would you be open to me leaving this as is for now (it definitely makes the checks I need to do easier), and I come back in another PR and resolve some of the above issues as well so that _more_ callers use this interface rather than it just being the one?

PR #2817, opened 2023-03-28
Remove autorest dependency as it is deprecated

### LINKED ISSUE #2814: Remove usage of autorest/to
autorest is deprecated, we should replace this with the package from the track2 SDK or roll our own.

matthchr:
Closes #2814

Replaced with our own generic version as the total required code is ~20 lines now. There is technically a new Azure SDK version but it doesn't support `to.Value` which we use relatively heavily. (was `to.String` and friends)

**If applicable**:
- [ ] this PR contains tests


super-harsh (approved):

v2/internal/util/to/to.go
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package to
+
+// Value extracts a value from a pointer. If the pointer is nil, the default value is returned.
+func Value[T any](v *T) T {
```
theunrepentantgeek, line 7: Nice use of generics.

theunrepentantgeek (approved):

PR #2823, opened 2023-03-30
Add Extension point to control post reconciliation behaviour

### LINKED ISSUE #2799: Extension point to control post reconciliation behaviour
For some ARM resources, we need the ability to set a condition on the resource based on the resource `status` we receive from ARM.

An example of this is PrivateEndpoints - Where we need to check the `ConnectionState` of a `PrivateEndpointConnection` from the status received by ARM to set a condition based on if it's approved by the network admin.

super-harsh:
Closes #2799 

This PR adds a `PostReconciliationChecker` extension to run post reconcile checks on a resource. Also includes `PostReconciliationChecker` extension implementation for `PrivateEndpoint` to set a condition on the resource based on if the connection is Approved. 

**If applicable**:
- [x] this PR contains tests


v2/api/network/customizations/private_endpoints_extensions.go
```diff
@@ -0,0 +1,57 @@
...
+	// Type assert that we are the hub type. This will fail to compile if
+	// the hub type has been changed but this extension has not
+	var _ conversion.Hub = endpoint
+
+	// We want to check `ManualPrivateLinkServiceConnections` as these are the ones which are not auto-approved.
+	if connections := endpoint.Status.ManualPrivateLinkServiceConnections; connections != nil {
+		for _, connection := range connections {
+			if *connection.PrivateLinkServiceConnectionState.Status != "Approved" {
+				// Returns 'conditions.NewReadyConditionImpactingError' error
+				return extensions.PostReconcileCheckResultFailure(
```
theunrepentantgeek, line 47: Would it make sense to list all the connections that need approval in one go?

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -448,6 +468,31 @@ func (r *azureDeploymentReconcilerInstance) handleCreatePollerSuccess(ctx contex
...
+		// No extension found, nothing to do
+		return extensions.PostReconcileCheckResultSuccess(), nil
+	}
+
+	// We also need to have our owner, it too with an up-to-date status
+	owner, ownerErr := r.ResourceResolver.ResolveOwner(ctx, r.Obj)
+	if ownerErr != nil {
+		// We can't obtain the owner, so we can't run the extension
+		return extensions.PostReconcileCheckResult{}, ownerErr
+	}
```
theunrepentantgeek, line 484: What about resources that don't have an owner that's an ASO resource?

v2/pkg/genruntime/extensions/postreconciliation_checker.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/Azure/azure-service-operator/v2/internal/util/kubeclient"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/conditions"
+)
+
+// PostReconciliationChecker is implemented by resources that want to do extra status checks after
+// a full ARM reconcile.
+type PostReconciliationChecker interface {
+	// PostReconcileCheck does a post-reconcile check to see if the resource is in a state to set 'Ready' condition.
+	// ARM resources should implement this to avoid reconciliation attempts that cannot possibly succeed.
```
theunrepentantgeek, line 25:
```suggestion
	// ARM resources should implement this if they need to defer the Ready condition until later
```

v2/pkg/genruntime/extensions/postreconciliation_checker.go
```diff
@@ -0,0 +1,160 @@
...
+func PostReconcileCheckResultFailure(reason string) PostReconcileCheckResult {
+	return PostReconcileCheckResult{
+		action:   postReconcileCheckResultTypeFailure,
+		severity: conditions.ConditionSeverityWarning,
+		reason:   conditions.ReasonPostReconcileFailure,
+		message:  reason,
+	}
+}
+
+func (r PostReconcileCheckResult) ReconciliationFailure() bool {
```
theunrepentantgeek, line 80: Suggest renaming these to use the past tense, since that matches the semantics.

```suggestion
func (r PostReconcileCheckResult) ReconciliationFailed() bool {
```

v2/pkg/genruntime/extensions/postreconciliation_checker.go
```diff
@@ -0,0 +1,160 @@
...
+	return r.action == postReconcileCheckResultTypeFailure
+}
+
+func (r PostReconcileCheckResult) ReconciliationSuccess() bool {
```
theunrepentantgeek, line 84:
```suggestion
func (r PostReconcileCheckResult) ReconciliationSucceeded() bool {
```

v2/pkg/genruntime/extensions/postreconciliation_checker.go
```diff
@@ -0,0 +1,160 @@
...
+
+		log.V(Status).Info("Extension post-reconcile check succeeded", "Result", result)
+
+		return result, nil
+	}, true
+}
+
+// alwaysSucceed is a PostReconciliationChecker that always indicates a reconciliation is successful.
+// We have this here, so we can set up a chain, even if it's only one link long.
+// When we start doing proper comparisons between Spec and Status, we'll have an actual chain of checkers.
```
theunrepentantgeek, line 150: Will this extension be involved in comparisons between `Spec` & `Status`?

theunrepentantgeek (approved):
Looking good.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
super-harsh: Is there any resource that does not have an owner except `ResourcrGroups` ?

v2/pkg/genruntime/extensions/postreconciliation_checker.go
super-harsh: good catch! I left this one behind. Removing it.

v2/api/network/customizations/private_endpoints_extensions.go
super-harsh: Updated!

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
theunrepentantgeek: `ResolveOwner()` returns a `genruntime.ARMMetaObject`, so it can only return a resource that exists in the cluster - and there's no requirement that people create their `ResourceGroup` in the cluster.

For example, in ARM a `BatchAccount` is created within a `ResourceGroup`, but in ASO you could just create the `BatchAccount`, putting it inside a ResourceGroup that already exists.

super-harsh: Do we allow this scenario where the owner does not exist in ASO? I think it fails while we try to `Claim` the resource

Issue #2822, opened 2023-03-29
Feature: Add an option to source workspaceResourceReference to Component (insights)


mehighlow:
The traditional Application Insights will be phased out by February 2024 and will be replaced by the Workspace-based version. In this new version, the `Component` allows the workspace to be sourced via `workspaceResourceReference.armId`. It would be advantageous to expose armId as a K8s configMap (https://github.com/Azure/azure-service-operator/issues/2821) and to provide the capability to reference it.

matthchr:
You can already do this automatically. You just use the `workspaceResourceReference.Group/Kind/Name`.

Just point it to the OperationalInsights workspace kind + name, for example:

```
workspaceResourceReference:
  group: operationalinsights.azure.com
  kind: Workspace
  name: <your workspace name>
```

We maybe need to document that capability better though...

mehighlow:
This won't work for cross-subscription installation, will it? `armId` defines workspace more precisely. 
Another issue might be if appinsights gets created successfully, but the workspace fails. If there is no workspace resource reference, appinsights could through a clearer error message, imho.

matthchr:
> This won't work for cross-subscription installation, will it?

It will. The subid used by the workspace is attached to the workspace. When you reference it in Kubernetes, we have that context as well as the resource name, etc and use it all to actually build the ARM ID (under the hood).

There's no need to export the ARM ID to a configmap and allow import here. the Group + Kind + Name reference style should be preferred for linking two ASO resources together and accomplishes the same goal. `workspaceResourceReference.armID` should really only be used if you have a resource created outside of ASO that is not managed by ASO and you wanna point to that.

Issue #2821, opened 2023-03-29
Feature: expose Workspace's (operationalinsights) armId


mehighlow:
Feature: expose Workspace's (operationalinsights) armId

Azure Service Operator version: 2
apiVersion: operationalinsights.azure.com/v1beta20210601

Please consider adding operatorSpec to the Workspace

```
  operatorSpec:
    configMaps:
      armId:
        name: workspace-properties-cm
        key: armId
```

matthchr:
Can you expand on the workflow you need it for?

mehighlow:
for workspace-based appinsights the `workspaceResourceReference.armId` needs to be set. 
with https://github.com/Azure/azure-service-operator/issues/2822 implemented it gets easier to blend one into another.

matthchr:
Closing this based on the conversation in the other issue - if all you need this for is referencing another ASO resource, we already have support for that (as outlined in #2822)

PR #2827, opened 2023-03-30
Improve/insights samples

mehighlow:
Refers #2821 and #2822

This PR improves insights sample

None


matthchr:
We also have tests for each sample (to ensure that the samples are correct). The tests are done using HTTP recordings and the results are checked in.

These samples have two tests associated with them: 
https://github.com/Azure/azure-service-operator/blob/main/v2/internal/controllers/recordings/Test_Samples_CreationAndDeletion/Test_Insights_v1api_CreationAndDeletion.yaml
and
https://github.com/Azure/azure-service-operator/blob/main/v2/internal/controllers/recordings/Test_Samples_CreationAndDeletion/Test_Insights_v1beta_CreationAndDeletion.yaml

In order to re-record those tests, you need to follow the instructions [here](https://azure.github.io/azure-service-operator/contributing/#running-integration-tests). Mainly you need to set up the `AZURE_SUBSCRIPTION_ID`, `AZURE_TENANT_ID`, etc.

You don't need to run every integration test, if you just run the one sample test: `Test_Samples_CreationAndDeletion`, that will re-record the those files. The actual workflow to do this is:
1. Set `AZURE_SUBSCRIPTION_ID`, etc vars
2. Delete the two existing recording files (linked above).
3. Run `TEST_FILTER=Test_Samples_CreationAndDeletion task controller:test-integration-envtest`

This should produce 2 new files (named the same as before but with different content) containing the HTTP recordings of successfully deploying that component + workspace

mehighlow:
@matthchr, tests associated with samples have been added. PTAL.

matthchr (approved):

PR #2836, opened 2023-04-02
Publish raw YAML with just CRDs each release

### LINKED ISSUE #2818: Publish raw YAML with just CRDs each release
As part of making the operator pod manage CRDs (#2769), we removed the CRDs from the published YAML. This means there's no longer any place to get them.

**Describe the improvement**
We should publish a standalone CRD-only YAML for folks who want to manage the CRDs themselves, or who want the raw CRDs for other uses.

super-harsh:
Closes #2818 

This is a one-liner addition to enabling the creation of raw YAML which is aided by #2830. The raw CRDs file gets generated into `v2/bin` as part of `controller:run-kustomize` with this change, then it is picked up by `make-release-artifacts` to copy into `v2/out/release` for release artifacts.


matthchr (approved):
Huh, that was easy.

Nice!

PR #2841, opened 2023-04-05
Add support for importing extension resources

### LINKED ISSUE #2803: Feature: Import should support extension resources
Extension resources (such as `roleAssignment`) should be importable via `asoctl`.

This is a follow-up to #2728

theunrepentantgeek:

Extends `asoctl import azure-resource` to handle extension resources too.

* Adds `owner` parameter to `Import()`.
* Populates `Status` before calling any implementation of `extensions.Importer` to give more information to the implementation.
* Implements `extensions.Importer` for RoleAssignment, allowing us to omit role assignments that are inherited.
* Adds concurrency to the import to improve performance.

Closes #2803

**Prerequisites**

- [ ] #2837 

gif


v2/api/authorization/customizations/role_assignment_extensions.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/extensions"
+)
+
+var _ extensions.Importer = &RoleAssignmentExtension{}
+
+func (extension *RoleAssignmentExtension) Import(
+	ctx context.Context,
+	rsrc genruntime.ImportableResource,
+	owner *genruntime.ResourceReference,
```
matthchr, line 23: An alternative to passing `owner` here would be, if there's a way for each resource to get its own ARM ID, you could use the ParseARMID function of the Azure SDK to decompose the ARM ID and then determine its owner from that.

I don't know if you have such a way to get the resources ARM ID already, but if you do it would save you from needing to modify your interface (which might be cleaner?)

Not critical though if there is no way for the resource to get its own ARM ID already.

v2/tools/asoctl/internal/importing/find_child_resources.go
```diff
@@ -45,6 +45,12 @@ func createTypeToSubtypesMap() map[string]set.Set[string] {
 
 		rsrc, ok := obj.(genruntime.KubernetesResource)
 		if !ok {
+			// Skip non-resources
```
matthchr, line 48: Curious when this happens? I would've thought that every GVK pointed to a resource?

v2/tools/asoctl/internal/importing/import_skipped_error.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// ImportSkippedError is an error that indicates that a resource cannot be imported for a reason we know about.
+// This allows us to continue the import even if some expected errors occur.
+type ImportSkippedError struct {
```
matthchr, line 16: Are these errors all by definition "known and thus ignorable"? Or do we need a `Severity` or something like that attached to this too?

matthchr: Put another way, is there any such thing as a "Known and _not_ ignorable" error?

v2/tools/asoctl/internal/importing/importable_arm_resource.go
```diff
@@ -75,24 +90,21 @@ func (i *importableARMResource) Resource() genruntime.MetaObject {
...
 	if err != nil {
 		return nil, err
 	}
 
 	var result []ImportableResource
-	subTypes := FindSubTypesForType(id.ResourceType.String())
-	for _, subType := range subTypes {
+
+	childTypes := FindChildResourcesForResourceType(i.armID.ResourceType.String())
+	childTypes = append(childTypes, FindExtensionTypes()...)
```
matthchr, line 102: minor: Might be worth leaving a comment here saying something like: "Extension types are children of every other type" or something like that?

v2/tools/asoctl/internal/importing/importable_arm_resource.go
```diff
@@ -167,30 +187,83 @@ func (i *importableARMResource) loader() extensions.ImporterFunc {
 		}
 
 		// Get the current status of the object from ARM
-		status, err := i.getStatus(ctx, i.armID, importable)
+		status, err := i.getStatus(ctx, i.armID.String(), importable)
 		if err != nil {
+			var responseError *azcore.ResponseError
```
matthchr, line 192: minor: You could probably stick this bit into some sort of `classifyError` function if you wanted (and/or if it needed to be reused elsewhere)? Might help the code read a bit cleaner too?

v2/tools/asoctl/internal/importing/importable_arm_resource.go
```diff
@@ -143,22 +155,30 @@ func (i *importableARMResource) createImportFunction(
...
+		loader = func(
+			ctx context.Context,
+			resource genruntime.ImportableResource,
+			owner *genruntime.ResourceReference,
+		) (extensions.ImportResult, error) {
+			return ex.Import(ctx, resource, owner, next)
 		}
 	}
 
+	// Lastly, we need to wrap importFn to remove Status
```
matthchr, line 171: Explain why status needs to be cleared here, rather than (or in addition to) that it must be cleared.

I assume the reason is because if status is not cleared, it will show up in the exported resource YAML which is both confusing and not-helpful?

v2/tools/asoctl/internal/importing/importable_arm_resource.go
```diff
@@ -220,13 +293,39 @@ func (i *importableARMResource) importChildResources(
 		containerURI,
 		imp.GetAPIVersion())
 	if err != nil {
+		var responseError *azcore.ResponseError
+		if errors.As(err, &responseError) {
```
matthchr, line 297: _Almost_ the same error processing logic as above? Would these two places likely evolve in tandem? Can we share the logic of error classification such that each spot can use the one function (but possibly omit its own logs, or something?)

Or maybe we really want a log on the forbidden case above too?

v2/tools/asoctl/internal/importing/resource_importer.go
```diff
@@ -36,91 +38,157 @@ func NewResourceImporter(
...
+		return errors.Wrapf(err, "failed to create importer for %q", armID)
+	}
+
 	ri.Add(importer)
+	return nil
+}
+
+// addCore is the internal implementation of Add that assumes the lock is already held
+// This allows us to call it from other methods that already have the lock
+func (ri *ResourceImporter) addCore(importer ImportableResource) {
```
matthchr, line 68: is `Core` being used here to mean: "The actual logic"?

I've seen `addImpl` or `addLocked` too - `core` was a bit confusing to me for about 5s as I was thinking this somehow had something to do with threading (mentioned in your PR description) and the core you were referring to was somehow a compute core.

I'm not necessarily opposed to `Core` as a suffix here it's just I haven't seen it used this way before. I do _vaguely_ think I saw in the actual Go source code usage of the `xLocked` approach, but I can't recall where so might be hallucinating it

v2/tools/asoctl/internal/importing/resource_importer.go
```diff
@@ -36,91 +38,157 @@ func NewResourceImporter(
...
+	processed := 0
+
+	// Why the nested loop?
+	// We can potentially trigger all the pending resources listed in ri.queue to be imported in parallel
+	// and end up with nothing on the queue because they're still running.
+	// The outer loop gives us a last look at the queue and allows us to keep going if there are new items
+	// requiring import.
+	for len(ri.queue) > 0 {
+		var eg errgroup.Group
+		eg.SetLimit(8)
```
matthchr, line 96: Why 8?

matthchr (approved):
Had some small comments/questions. Approving in the interest of time

v2/api/authorization/customizations/role_assignment_extensions.go
theunrepentantgeek: I considered doing that - but realized it wouldn't work for non-ARM resources (such as database users).

v2/tools/asoctl/internal/importing/import_skipped_error.go
theunrepentantgeek: Any not-ignorable error can just be returned and will do the right thing. We only need a special case for errors that are non-fatal.

v2/tools/asoctl/internal/importing/importable_arm_resource.go
theunrepentantgeek: Correct. I'll update the code.

v2/tools/asoctl/internal/importing/resource_importer.go
theunrepentantgeek: Semi random number chosen so we get a reasonable degree of parallelism without trying to do too much. I'll add a comment.

PR #2844, opened 2023-04-05
bump initialDelaySeconds for readiness and liveness probe

super-harsh:

The operator pod was running into timeouts on my arm64 mac machine due to readiness probe timeout. This PR bumps the `initialDelaySeconds` for readiness and liveness probes so that operator gets more time to load and apply CRDs.


matthchr (approved):
Good catch

PR #2846, opened 2023-04-06
Fix bugs found by bug bash

matthchr:
- Use non-storage API version in CRD Cleaner.
- Update Taskfile Helm target to use same Helm installation name as public docs.
- Avoid v1alpha1api versions in Controller.
- Delete kind cluster after upgrade test.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -330,7 +330,7 @@ tasks:
     - "cmctl check api --wait=2m"
     - "helm repo add asov2 https://raw.githubusercontent.com/Azure/azure-service-operator/main/v2/charts"
     - "helm repo update"
-    - "helm upgrade --install --devel asov2 asov2/azure-service-operator \
+    - "helm upgrade --install --devel aso2 asov2/azure-service-operator \
       --create-namespace \
       --namespace={{.ASO_NAMESPACE}} \
       --set azureSubscriptionID=$AZURE_SUBSCRIPTION_ID \
@@ -348,6 +348,7 @@ tasks:
     - task: controller:wait-for-operator-ready
 #   TODO: Will have to change versions in 'Test_Pre_Release_ResourceCanBeCreated_AfterUpgrade' test once we move from beta to stable
     - "go test -timeout 15m -v -run Test_Pre_Release_ResourceCanBeCreated_AfterUpgrade ./test/pre-release"
+    - task controller:kind-delete
 
   controller:build:
     desc: Generate the {{.CONTROLLER_APP}} binary.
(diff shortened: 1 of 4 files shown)


v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -199,3 +200,9 @@ func removeMatchingStoredVersions(oldVersions []string, versionRegexp *regexp.Re
 
 	return newStoredVersions, matchedStoredVersion
 }
+
+// getVersionFromStoredVersion returns the public (non-storage) API version for a given version
+func getVersionFromStoredVersion(version string) string {
```
matthchr, line 205: So technically this fix is enough and we don't need the fix in `group_version_kind_aware`, but I put both just to be on the safe side.

Without this change we see the following error:
> E0406 00:01:43.131605       1 generic_reconciler.go:348] controllers/VirtualNetworksSubnetController "msg"="Encountered error impacting Ready condition" "error"="Reason: Failed, Severity: Warning, RetryClassification: RetrySlow, Cause: unable to get spec from network.azure.com/v1api20201101storage, Kind=VirtualNetworksSubnet: getting versioned spec: unable to create new network.azure.com/v1alpha1api20201101, Kind=VirtualNetworksSubnet: no kind \"VirtualNetworksSubnet\" is registered for version \"network.azure.com/v1alpha1api20201101\" in scheme \"pkg/runtime/scheme.go:100\"" "name"="samplesubnet" "namespace"="default"

This is because the `asoctl crd clean` command was using the storage version to do the update not the API version, which wasn't correctly changing the latest used API version.

Taskfile.yml
```diff
@@ -330,7 +330,7 @@ tasks:
     - "cmctl check api --wait=2m"
     - "helm repo add asov2 https://raw.githubusercontent.com/Azure/azure-service-operator/main/v2/charts"
     - "helm repo update"
-    - "helm upgrade --install --devel asov2 asov2/azure-service-operator \
+    - "helm upgrade --install --devel aso2 asov2/azure-service-operator \
```
matthchr, line 333: Changed to `aso2` to match public docs

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -75,32 +76,32 @@ func (c *Cleaner) Run(ctx context.Context) error {
...
-		} else {
+		// If the slice was not updated, there is no version to deprecate.
+		if len(newStoredVersions) == len(crd.Status.StoredVersions) {
 			klog.Infof("Nothing to update for %q\n", crd.Name)
+			continue
+		}
+
+		// Make sure to use a version that hasn't been deprecated for migration. Deprecated versions will not be in our
+		// scheme, and so we cannot List/PUT with them. Instead, use the next available version.
+		activeVersion := getVersionFromStoredVersion(newStoredVersions[0])
```
theunrepentantgeek, line 87: Not sure that this is the right one to use. 

If `newStoredVersions` is in non-deterministic order, this runs the risk of using a version that is missing some properties, and thus experiencing data-loss.

If `newStoredVersions` is in oldest-to-newest order, then selecting the oldest one is almost guaranteed to cause data loss because properties are unknown. 

Or maybe the result will be that `asoctl` will fail?

Either way, not good. Need to proactively select the latest non-preview version.

theunrepentantgeek (approved):

v2/tools/asoctl/internal/crd/cleaner.go
theunrepentantgeek: To clarify the problem.

Say we currently have support for three different ARM API versions:
* `v2020-01-01`
* `v2021-01-01` - introduces property `Foo`
* `v2022-01-01` - introduces property `Bar`

We'll therefore have the following ASO CRD versions:
* `v1alpha1api20200101`
* `v1alpha1api20210101`
* `v1alpha1api20220101`
* `v1beta20200101`
* `v1beta20210101`
* `v1beta20220101`

Now suppose we have a `v1alpha1api20220101` resource which is using both the `Foo` and `Bar` properties.

When doing the upgrade, `newStoredVersions[0]` will be `v1beta20200101` - this version doesn't have `Foo` nor `Bar`, so it's not possible to save the amended resource as that version; I think the save will then fail because the api-server doesn't know what to do with the extra properties.

Issue #2829, opened 2023-03-30
Refactor ToARMConverter + FromARMConverter


matthchr:
The shape of these interfaces evolved somewhat organically. We generate both conversion directions for all types, but in reality it would make more sense to generate only ToARM for Spec types, and only FromARM for Status types, as those are the directions that things flow.

For example the fact that `FromARMConverter.PopulateFromARM` takes an owner is weird because it's only ever used for spec types and we never actually call that method on spec types.

matthchr:
This is still something we'd like to do.

theunrepentantgeek:
Still something we thingk would be nice to do.

matthchr:
Still interested in this (low priority).

matthchr:
No change from above.

matthchr:
No change from above. Very low priority but would be nice to do.

PR #2850, opened 2023-04-07
Support UserAssignedIdentities on all relevant resources

### LINKED ISSUE #2528: Feature: Specify user managed identity on objects cleanly
The `identity.userAssignedIdentities` field which we have been pruning from objects thinking it is readOnly is _not_ readonly. It is expected that users provide, as a map key, the ARM IDs of the userAssignedIdentities they would like to use on their object.

See [this](https://learn.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/qs-configure-template-windows-vm#assign-a-user-assigned-managed-identity-to-an-azure-vm) for example, or the type [in the AKS Swagger](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/containerservice/resource-manager/Microsoft.ContainerService/stable/2022-07-01/managedClusters.json#L4080) as another.

We should bring things property back, and change the code generator to detect and fix the keys to be `genruntime.ResourceReference`'s instead of full ARM IDs.

matthchr:
Resolves #2528.

- Add DoNotPrune protection for map value of UserAssignedIdentities field. Previously this field was being pruned because it was a map[string]EmptyObject.
- Ensure that all UserAssignedIdentities are map[ARMID]Object, as that is the expected shape.
- Add tests for UserAssignedIdentities to ensure that they work properly.

**If applicable**:
- [x] this PR contains tests


v2/api/appconfiguration/v1api20220501/structure.txt
```diff
@@ -17,12 +17,14 @@ github.com/Azure/azure-service-operator/v2/api/appconfiguration/v1api20220501
 │   │   │   └── KeyVaultProperties: *Object (2 properties)
 │   │   │       ├── IdentityClientId: *string
 │   │   │       └── KeyIdentifier: *string
-│   │   ├── Identity: *Object (1 property)
-│   │   │   └── Type: *Enum (4 values)
-│   │   │       ├── "None"
-│   │   │       ├── "SystemAssigned"
-│   │   │       ├── "SystemAssigned, UserAssigned"
-│   │   │       └── "UserAssigned"
+│   │   ├── Identity: *Object (2 properties)
```
matthchr, line 20: I've gone through and confirmed that there are no breaking changes (only addition of a new property with this shape) for all types, but good to get a second set of eyes on it to confirm.

The properties should all look like this too:
```
│   │   │   └── UserAssignedIdentities: Object (1 property)[]
│   │   │       └── Reference: genruntime.ResourceReference
```

v2/tools/generator/internal/armconversion/convert_to_arm_function_builder.go
```diff
@@ -455,6 +492,89 @@ func (builder *convertToARMBuilder) propertiesWithSameNameHandler(
 	), true
 }
 
+// convertReferenceProperty handles conversion of reference properties.
+// This function generates code that looks like this:
+//
+// TODO: Show structure of generated code in comment
```
matthchr, line 498: Need to fix TODO

v2/tools/generator/internal/armconversion/convert_to_arm_function_builder.go
```diff
@@ -455,6 +492,89 @@ func (builder *convertToARMBuilder) propertiesWithSameNameHandler(
 	), true
 }
 
+// convertReferenceProperty handles conversion of reference properties.
```
matthchr, line 495: Wrong name

v2/tools/generator/internal/codegen/pipeline/create_arm_types.go
```diff
@@ -245,6 +264,32 @@ func (c *armTypeCreator) createARMNameProperty(prop *astmodel.PropertyDefinition
...
 }
 
+func (c *armTypeCreator) createUserAssignedIdentitiesProperty(prop *astmodel.PropertyDefinition, _ bool) (*astmodel.PropertyDefinition, error) {
+	typeName, ok := astmodel.IsUserAssignedIdentityProperty(prop)
+	if !ok {
+		return nil, nil
+	}
+
+	newTypeName := astmodel.CreateARMTypeName(typeName)
+	// TODO: Does this need to be based on their Swagger?
```
matthchr, line 274: Clarify this TODO comment (for future readers). Won't be implementing this now

v2/tools/generator/internal/codegen/pipeline/handle_user_assigned_identities.go
```diff
@@ -0,0 +1,120 @@
...
+	it = it.WithProperty(prop)
+
+	return astmodel.IdentityVisitOfObjectType(this, it, ctx)
+}
+
+func newUserAssignedIdentityDefinition(pr astmodel.PackageReference) astmodel.TypeDefinition {
+	name := astmodel.MakeTypeName(pr, astmodel.UserAssignedIdentitiesTypeName)
+
+	prop := astmodel.NewPropertyDefinition("Reference", "reference", astmodel.ResourceReferenceType)
+	prop = prop.WithTag(astmodel.ARMReferenceTag, "Reference") // TODO: This is not great
```
matthchr, line 115: Remove/clarify TODOs

v2/tools/generator/internal/codegen/pipeline/inject_spec_initialization_functions.go
```diff
@@ -313,6 +313,11 @@ func (s *specInitializationScanner) visitArrayType(
 ) (astmodel.Type, error) {
 	status, ok := astmodel.AsArrayType(statusAny.(astmodel.Type))
 	if !ok {
+		// TODO: need to improve this check significantly
```
matthchr, line 316: Fix this TODO

v2/tools/generator/internal/codegen/pipeline/json_serialization_test_cases.go
```diff
@@ -7,9 +7,11 @@ package pipeline
...
 import (
 	"context"
+
+	"github.com/pkg/errors"
+
 	"github.com/Azure/azure-service-operator/v2/tools/generator/internal/astmodel"
 	"github.com/Azure/azure-service-operator/v2/tools/generator/internal/testcases"
-	"github.com/pkg/errors"
 
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
```
matthchr, line 16: minor: I think this should move too

v2/tools/generator/internal/conversions/property_conversions.go
```diff
@@ -1171,6 +1172,103 @@ func assignMapFromMap(
...
+		return nil, nil
+	}
+
+	if destinationEndpoint.Name() != astmodel.UserAssignedIdentitiesProperty {
+		return nil, nil
+	}
+
+	destinationElement := destinationArray.Element()
+
+	// TODO: Need an outer nil-check
```
matthchr, line 1217: Remove TODO, as it's done

v2/tools/generator/internal/conversions/property_conversions.go
```diff
@@ -1171,6 +1172,103 @@ func assignMapFromMap(
...
+		declaration := astbuilder.ShortDeclaration(
+			tempId,
+			astbuilder.MakeAppendSlice(destinationArray.AsType(generationContext), astbuilder.CallFunc("len", reader)))
+
+		loopLocals := knownLocals.Clone()
+		keyID := loopLocals.CreateLocal(sourceEndpoint.Name(), "Key")
+
+		intermediateDestination := loopLocals.CreateLocal(destinationEndpoint.Name(), "Ref")
+
+		conversion, err := CreateTypeConversion(
```
matthchr, line 1231: @theunrepentantgeek - other callers of CreateTypeConversion do it above, rather than in this function, but it wasnt' totally clear to me how to pull that off and use `knownLocals` to pick an intermediate destination name.

Let me know if there's a different way this should be being done.

theunrepentantgeek: Definitely should be doing all the creation of conversions in the outer method - the returned func should never panic, it should be doing code generation that's known to succeed.

matthchr: Do you have any suggestions on how exactly I should go about doing that in this case, where the `intermediateDestination` of the conversion is to a local variable whose name is determined by `loopLocals.CreateLocal`?

I don't have access to the `knownLocals` outside of the context of this function.

Maybe I'm confused and the "Name" aspect of the conversion endpoint doesn't matter as much as I thought it did because the function returned (the actual conversion function) takes in the name to assign to? Looking at `assignMapFromMap` that must be the case...

I'll move this up

matthchr: Moved

v2/api/appconfiguration/v1api20220501/configuration_store_spec_arm_types_gen.go
```diff
@@ -136,6 +137,9 @@ const (
 	SystemData_LastModifiedByType_User            = SystemData_LastModifiedByType("User")
 )
 
+type UserAssignedIdentityDetails_ARM struct {
```
theunrepentantgeek, line 140: Shouldn't this have a doc comment?

v2/tools/generator/internal/conversions/property_conversions.go
```diff
@@ -1171,6 +1172,105 @@ func assignMapFromMap(
...
+		return nil, nil
+	}
+
+	// Require destination to be an array type
+	destinationArray, destinationIsArray := astmodel.AsArrayType(destinationEndpoint.Type())
+	if !destinationIsArray {
+		return nil, nil
+	}
+
+	if sourceEndpoint.Name() != astmodel.UserAssignedIdentitiesProperty {
```
theunrepentantgeek, line 1207:
```suggestion
        // Require the source endpoint to have the expected property name
	if sourceEndpoint.Name() != astmodel.UserAssignedIdentitiesProperty {
```

v2/tools/generator/internal/conversions/property_conversions.go
```diff
@@ -1171,6 +1172,105 @@ func assignMapFromMap(
...
+		return nil, nil
+	}
+
+	if destinationEndpoint.Name() != astmodel.UserAssignedIdentitiesProperty {
```
theunrepentantgeek, line 1211:
```suggestion
        // Require the destination endpoint to have the expected property name
	if destinationEndpoint.Name() != astmodel.UserAssignedIdentitiesProperty {
```

v2/tools/generator/internal/conversions/property_conversions.go
```diff
@@ -1171,6 +1172,105 @@ func assignMapFromMap(
...
+		return nil, nil
+	}
+
+	// Require source to be a map type
+	_, sourceIsMap := astmodel.AsMapType(sourceEndpoint.Type())
+	if !sourceIsMap {
+		return nil, nil
+	}
+
+	// Require destination to be an array type
```
theunrepentantgeek, line 1201: Should you check the array time has the expected element type?

v2/tools/generator/internal/conversions/property_conversions.go
```diff
@@ -1171,6 +1172,105 @@ func assignMapFromMap(
...
+	destinationEndpoint *TypedConversionEndpoint,
+	conversionContext *PropertyConversionContext) (PropertyConversion, error) {
+
+	// Require both source and destination to not be bag items
+	if sourceEndpoint.IsBagItem() || destinationEndpoint.IsBagItem() {
+		return nil, nil
+	}
+
+	// Require source to be a map type
+	_, sourceIsMap := astmodel.AsMapType(sourceEndpoint.Type())
```
theunrepentantgeek, line 1196: Should you check the map type has the expected key and value types?

v2/tools/generator/internal/conversions/property_conversions.go
```diff
@@ -1171,6 +1172,105 @@ func assignMapFromMap(
...
+		conversionContext)
+	if err != nil {
+		return nil, errors.Wrapf(
+			err,
+			"finding UserAssignedIdentities conversion from %s to %s",
+			astmodel.DebugDescription(astmodel.StringType),
+			astmodel.DebugDescription(astmodel.ResourceReferenceType))
+	}
+
+	return func(reader dst.Expr, writer func(dst.Expr) []dst.Stmt, knownLocals *astmodel.KnownLocalsSet, generationContext *astmodel.CodeGenerationContext) []dst.Stmt {
```
theunrepentantgeek, line 1229: I'd suggest annotating individual statements with comments showing what code will be generated.

v2/tools/generator/internal/conversions/property_conversions.go
```diff
@@ -1171,6 +1172,105 @@ func assignMapFromMap(
...
+// a map whose key is the ARM ID of the userAssignedIdentity
+// <arr> := make([]<arrType>, 0, len(source.UserAssignedIdentities))
+//
+//	for key, _ := range source.UserAssignedIdentities {
+//	   ref := genruntime.CreateResourceReferenceFromARMID(key)
+//	   <arr> = append(<arr>, UserAssignedIdentityDetails{Reference: ref})
+//	}
+//
+// <writer> = <arr>
+func assignUserAssignedIdentityMapFromArray(
```
theunrepentantgeek, line 1185: Maybe explain why we don't need the other direction.

theunrepentantgeek (commented):
Partial review.

v2/tools/generator/internal/codegen/pipeline/create_arm_types.go
```diff
@@ -389,3 +440,22 @@ func (c *armTypeCreator) convertObjectPropertiesForARM(t *astmodel.ObjectType, i
...
 }
+
+func skipUserAssignedIdentity(def astmodel.TypeDefinition) bool {
+	return def.Name().Name() == astmodel.UserAssignedIdentitiesTypeName
+}
+
+func requiresARMType(def astmodel.TypeDefinition) bool {
+	skipFuncs := []func(it astmodel.TypeDefinition) bool{
+		skipUserAssignedIdentity,
+	}
```
theunrepentantgeek, line 451: Maybe move this to a global so it's constructed once?

v2/tools/generator/internal/astmodel/std_properties.go
```diff
@@ -18,10 +18,34 @@ const (
...
 
 // IsKubernetesResourceProperty returns true if the supplied property name is one of the properties required by the
 // KubernetesResource interface.
 func IsKubernetesResourceProperty(name PropertyName) bool {
 	return name == AzureNameProperty || name == OwnerProperty
 }
+
+func IsUserAssignedIdentityProperty(prop *PropertyDefinition) (TypeName, bool) {
+	if !prop.HasName(UserAssignedIdentitiesProperty) {
+		return TypeName{}, false
```
theunrepentantgeek, line 33:
```suggestion
		return astmodel.EmptyTypeName, false
```

And other places too.

v2/tools/generator/internal/astbuilder/slices.go
```diff
@@ -13,7 +13,7 @@ import (
 
 // MakeSlice returns the call expression for making a slice
```
theunrepentantgeek, line 14:
```suggestion
// MakeSlice returns the call expression for making a slice with a specified length
```

v2/tools/generator/internal/astbuilder/slices.go
```diff
@@ -24,6 +24,24 @@ func MakeSlice(listType dst.Expr, len dst.Expr) *dst.CallExpr {
 	}
 }
 
+// MakeAppendSlice returns the call expression for making a slice with a specified length that can be
+// appended to using append(...)
+//
+// make([]<value>, 0, <len>)
+func MakeAppendSlice(listType dst.Expr, len dst.Expr) *dst.CallExpr {
```
theunrepentantgeek, line 31: Not overly happy with this name.

What about this?

```suggestion
func MakeSliceWithCapacity(listType dst.Expr, capacity dst.Expr) *dst.CallExpr {
```

v2/tools/generator/internal/astbuilder/slices.go
```diff
@@ -24,6 +24,24 @@ func MakeSlice(listType dst.Expr, len dst.Expr) *dst.CallExpr {
...
+		Fun: dst.NewIdent("make"),
+		Args: []dst.Expr{
+			listType,
+			&dst.BasicLit{
+				Value: "0",
+				Kind:  token.INT,
+			},
+			len,
+		},
+	}
```
theunrepentantgeek, line 42: Working too hard.

```suggestion
	return CallFunc(
		"make",
		listType,
		IntLiteral(0),
		len)
	}
```

theunrepentantgeek: You might want to improve `MakeSlice()` too.

v2/tools/generator/internal/armconversion/convert_from_arm_function_builder.go
```diff
@@ -177,6 +178,18 @@ func (builder *convertFromARMBuilder) namePropertyHandler(
...
+func (builder *convertFromARMBuilder) userAssignedIdentitiesPropertyHandler(
+	toProp *astmodel.PropertyDefinition,
+	_ *astmodel.ObjectType,
+) ([]dst.Stmt, bool) {
+	if _, ok := astmodel.IsUserAssignedIdentityProperty(toProp); !ok {
+		return nil, false
+	}
+
+	// TODO: For now we are not assigning these, as we don't know how to rebuild the reference
+	return nil, true
```
theunrepentantgeek, line 190: Still true?

theunrepentantgeek (approved):

v2/tools/generator/internal/conversions/property_conversions.go
matthchr: Done

v2/tools/generator/internal/conversions/property_conversions.go
matthchr: Have added a check that the destination is a `TypeName`

v2/tools/generator/internal/astbuilder/slices.go
matthchr: Done

v2/tools/generator/internal/astmodel/std_properties.go
matthchr: Done

v2/api/appconfiguration/v1api20220501/configuration_store_spec_arm_types_gen.go
matthchr: Fixed (and on the non-ARM type too)

v2/tools/generator/internal/conversions/property_conversions.go
matthchr: Added these

v2/tools/generator/internal/codegen/pipeline/create_arm_types.go
matthchr: Done

v2/tools/generator/internal/astbuilder/slices.go
matthchr: I don't love the name either. With that said, both `MakeSlice` functions are making a slice with a `Capacity`, because the parameter we are calling `len` here is actually always capacity, see: https://go.dev/tour/moretypes/11.

I've changed the `len` argument here to `capacity` on both functions to correct that issue.

That leaves the question of what the name of this function should be... technically we're making a zero length slice with a capacity... I've changed this function name to `MakeSliceZeroLength`. It still has the doc comment that explains "MakeSliceZeroLength returns the call expression for making a slice with a specified length that can be appended to using append(...)"

I don't love this name either but I guess it's at least technically correct. Thoughts?

v2/tools/generator/internal/astbuilder/slices.go
matthchr: Fixed both

v2/tools/generator/internal/armconversion/convert_from_arm_function_builder.go
matthchr: Yes, this is still true. Related to #2829, this other direction is actually not used.

v2/tools/generator/internal/astbuilder/slices.go
theunrepentantgeek: How about `MakeEmptySlice()`?

super-harsh (approved):

PR #2855, opened 2023-04-10
Cater to duplicate properties with configmap export

matthchr:
* Document how to use the $generatedConfigs property.
* Refactor config map export to work on types that may have multiple paths to the property in question. For example
  .Status.PrimaryEndpoints.Blob is a different config map export than .Status.SecondaryEndpoints.Blob, even though the two 
  Blob properties are on the same type.
* Export Storage endpoint details to configmap. Previously these were exported as secrets, but that doesn't chain as well into other ASO types

**If applicable**:
- [x] this PR contains tests


v2/azure-arm.yaml
```diff
@@ -542,6 +542,22 @@ status:
...
+#     to the property to export. Collections (or really any fancy jsonpath expressions) are not currently supported.
+#     Only valid for resources.
+#     Example:
+#         BlobEndpoint: .Status.PrimaryEndpoints.Blob
+#
+# $manualAzureGeneratedConfigs: <array of strings>
+#     An array of names identifying the properties which will be exported to config maps manually. Prefer $azureGeneratedConfigs
+#     if the property is on Status or otherwise accessible on the resource itself. $manualAzureGeneratedConfigs can be used
+#     to have more control over exactly what value goes into the exported config map, at the cost of needing to manually implement
+#     the config map export The key of the map is the name of the property to export, the value of the map is a json-path like expression
```
super-harsh, line 556:
```suggestion
#     the config map export. The key of the map is the name of the property to export, and the value of the map is a json-path like expression
```

v2/azure-arm.yaml
```diff
@@ -542,6 +542,22 @@ status:
...
+#     to the property to export. Collections (or really any fancy jsonpath expressions) are not currently supported.
+#     Example:
```
super-harsh, line 558: Do we need a better example here? This one looks identical to `$azureGeneratedConfigs` example which does not really explain the difference in both.

v2/azure-arm.yaml
matthchr: Fixed this

v2/azure-arm.yaml
matthchr: Fixed, they actually weren't supposed to have the same example.

v2/tools/generator/internal/config/object_model_configuration.go
```diff
@@ -244,6 +244,33 @@ func (omc *ObjectModelConfiguration) VerifyAzureGeneratedConfigsConsumed() error
 	return visitor.Visit(omc)
 }
 
+// ManualAzureGeneratedConfigs looks up a type to determine if it has any Azure generated configs
+func (omc *ObjectModelConfiguration) ManualAzureGeneratedConfigs(name astmodel.TypeName) ([]string, error) {
```
theunrepentantgeek, line 248: I'm confused by this name. How can something be _manual_ and _generated_ at the same time?

v2/azure-arm.yaml
matthchr: Fixed, I think

v2/azure-arm.yaml
```diff
@@ -542,6 +542,23 @@ status:
 #     implemented as an extension on the resource type in question.
 #     Only valid for resource types.
 #
+# $azureGeneratedConfigs: <map of string -> string>
+#     The key of the map is the name of the property to export, the value of the map is a json-path like expression
+#     to the property to export. Collections (or really any fancy jsonpath expressions) are not currently supported.
+#     Only valid for resources.
+#     Example:
+#         $azureGeneratedConfigs:
+#           BlobEndpoint: .Status.PrimaryEndpoints.Blob
```
theunrepentantgeek, line 551: Is the leading period necessary? It'll be really easy to leave out, and I guess it wouldn't work.

theunrepentantgeek (approved):
A couple of minor questions only.

v2/tools/generator/internal/config/object_model_configuration.go
matthchr: So basically what I am trying to evoke here is `$azureGeneratedSecrets` equivalent, which is that the code-generator will generate the property on `spec.OperatorSpec.ConfigMaps.<YourPropertyNameHere>`, but nothing will ever populate it unless you write a manual extension.

This is exactly how `$azureGeneratedSecrets` works too.

Maybe what we should do is call this `$manualConfigs` and drop the `AzureGenerated`?

matthchr: Did the above. Also changed `$azureGeneratedConfigs` to just `$generatedConfigs`

v2/azure-arm.yaml
matthchr: So my intent here was to mimic JSONPath. I missed that that required a leading `$`. I've changed these to be `$.Status.PrimaryEndpoints.Blob` as you're right, the leading `.` wasn't correct.

You'll get an error if you get the right structure but the wrong names, the error looks like:
```
F0411 15:19:19.345605 4144497 root.go:26] failed during pipeline stage 45/73 [addOperatorSpec]: Adds the property 'OperatorSpec' to all Spec types that require it: finding properties allowed to export as config maps: $azureGeneratedConfigs property "FullyQualifiedDomainName" not found at path "$.Status.FullyQualifiedDomainNamez"
```

I also just added an error for if you forget the `$.`

```
F0411 15:24:23.442393 4144829 root.go:26] configuration file loaded from "azure-arm.yaml" is not valid YAML: decoding yaml for "dbforpostgresql": decoding yaml for "2022-01-20-preview": decoding yaml for "FlexibleServer": $generatedConfigs entry "Status.FullyQualifiedDomainName" must begin with $.
```

PR #2859, opened 2023-04-11
Add pre-upgrade helm hook to verify helm/resource-policy on CRDs

### LINKED ISSUE #2831: Helm hook to block upgrade from beta.0-4 -> v2.0.0
The only safe Helm upgrade is from beta.5 -> v2.0.0. We can and should document this clearly, but ideally we could block the upgrade with a Helm [`pre-upgrade` hook](https://helm.sh/docs/topics/charts_hooks/#writing-a-hook)

These hooks are usually `batch/v1` `Jobs` which launch a container and then can perform some checks.

In order to do this we would need to:
- [ ] Create a `Job` YAML specifically for Helm, with the `pre-upgrade` hook.
- [ ] Create a ServiceAccount, Role, and RoleBinding giving the container access to the Kubernetes CRD API.
- [ ] Launch a container as part of the job above (possibly the same container as the operator pod uses, but with different arguments, for ease of publishing?). This container would use the CRDs API to list CRDs, find the ASO CRDs and ensure that they have the Helm `keep` policy.
- [ ] Ensure that the ServiceAccount, Role, RoleBinding, and Job are all deleted if the hook succeeds.

super-harsh:
Closes #2831 

We needed a pre-upgrade check for GA version to make sure the users are upgrading their clusters from `beta.5` instead of any earlier versions. This PR adds a helm job with a pre-upgrade helm hook to run the ASOv2 binary in `PreUpgradeCheck` mode to verify the existence of `helm/resource-policy: keep` annotation on CRDs.


v2/charts/azure-service-operator/templates/pre-release-hook/batch_v1_job_pre-upgrade-checker.yaml
```diff
@@ -0,0 +1,27 @@
...
+        image: {{.Values.image.repository}}
+        imagePullPolicy: Always
+        name: pre-upgrade-checker
+        resources:
+          limits:
+            cpu: 500m
+            memory: 512Mi
+          requests:
+            cpu: 200m
+            memory: 256Mi
```
matthchr, line 27: minor: should end in a newline

v2/cmd/controller/app/flags.go
```diff
@@ -53,5 +57,10 @@ func ParseFlags(args []string) (Flags, error) {
 		healthAddr:           healthAddr,
 		enableLeaderElection: enableLeaderElection,
 		crdPatterns:          crdPatterns,
+		preUpgradeCheck:      preUpgradeCheck,
 	}, nil
 }
+
+func (f *Flags) PreUpgradeCheck() bool {
```
matthchr, line 64: Might be clearer to instead just make all of the flags exported fields?

v2/cmd/controller/app/setup.go
```diff
@@ -44,6 +50,40 @@ import (
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/conditions"
 )
 
+func SetupPreUpgradeCheck(ctx context.Context) error {
+	cfg := clientconfig.GetConfigOrDie()
```
matthchr, line 54: IMO use the `GetConfig` flavor since we're returning an error here. `OrDie` is really only appropriate in a top level caller that doesn't return an error.

Since you _are_ returning an error here you can just return the error from `GetConfig` too.

v2/cmd/controller/app/setup.go
```diff
@@ -44,6 +50,40 @@ import (
...
+	apiExtClient, err := apiextensionsclient.NewForConfig(cfg)
+	if err != nil {
+		return errors.Wrap(err, "unable to create kubernetes client")
+	}
+
+	// Had to list CRDs this way and not with crdManager, since we did not have "serviceoperator.azure.com/version" labels in earlier versions.
```
matthchr, line 60: Makes sense.

v2/cmd/controller/app/setup.go
```diff
@@ -44,6 +50,40 @@ import (
...
+			continue
+		}
+
+		if policy, ok := crd.Annotations["helm.sh/resource-policy"]; !ok || policy != "keep" {
+			err = errors.New(fmt.Sprintf("looks like CRD '%s' does not contain helm keep policy. Make sure the upgrade is from beta.5", crd.Name))
+			errs = append(errs, err)
+		}
+	}
+
+	if err = kerrors.NewAggregate(errs); err != nil {
```
matthchr, line 80: You can condense this bit down to just:
```
return kerrors.NewAggregate(errs)
```

v2/cmd/controller/main.go
```diff
@@ -26,11 +26,18 @@ func main() {
...
-	mgr := app.SetupControllerManager(ctx, setupLog, flgs)
-
-	setupLog.Info("starting manager")
-	if err = mgr.Start(ctx); err != nil {
-		setupLog.Error(err, "problem running manager")
-		os.Exit(1)
+	if flgs.PreUpgradeCheck() {
+		err = app.SetupPreUpgradeCheck(ctx)
+		if err != nil {
+			setupLog.Error(err, "failed to run pre-upgrade checks")
```
matthchr, line 32: minor:
```suggestion
			setupLog.Error(err, "pre-upgrade check failed")
```

matthchr (approved):
Looks pretty good to me!

Have you tested it and made sure it works (try upgrade from beta.4 -> v2.0.0?)

v2/cmd/controller/app/setup.go
```diff
@@ -44,6 +50,40 @@ import (
...
+	crdRegexp := regexp.MustCompile(`.*\.azure\.com`)
+	var errs []error
+	for _, crd := range list.Items {
+		crd := crd
+		if !crdRegexp.MatchString(crd.Name) {
+			continue
+		}
+
+		if policy, ok := crd.Annotations["helm.sh/resource-policy"]; !ok || policy != "keep" {
+			err = errors.New(fmt.Sprintf("looks like CRD '%s' does not contain helm keep policy. Make sure the upgrade is from beta.5", crd.Name))
```
theunrepentantgeek, line 79: Suggested wording change. #nitpick

```suggestion
			err = errors.New(fmt.Sprintf("CRD '%s' does not have annotation for helm keep policy. Make sure the upgrade is from beta.5", crd.Name))
```

PR #2862, opened 2023-04-12
Improve asoctl user experience

### LINKED ISSUE #2854: Improve logging output of `asoctl`

Currently `asoctl` is using the `klog` logger, which is better suited for services, not least because of the hard coded prefix used for every log entry:

```
 I0406 08:02:36.555009   23800 resource_importer.go:130]
```
Given that `asoctl` is a command line tool, I believe we should pay a little more attention to the user experience and make it more polished for end users.

**Describe the improvement**

For plain logging, I suggest we use [`zerolog`](https://github.com/rs/zerolog) in console mode. 
We should also investigate using [`mpb`](https://github.com/vbauerster/mpb) to show overall progress.

theunrepentantgeek:

Improves the user experience working with `asoctl`

* Swaps out `klog` (which is optimized for microservice use) with `zerolog`, configured as a console logger.
* Provides progress indications when importing azure resources

image

Closes #2854 

Also fixes the bug where we got different help text for different command line errors.

gif


v2/tools/asoctl/cmd/crd_clean.go
```diff
@@ -13,18 +13,25 @@ import (
...
 
 	cmd := &cobra.Command{
 		Use:   "clean",
 		Short: "Clean deprecated CRD versions from cluster",
+		Args:  cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			cfg := config.GetConfigOrDie()
+			cfg, err := config.GetConfig()
+			if err != nil {
+				return errors.Wrap(err, "unable to get kubernetes config")
```
matthchr, line 32:
```suggestion
				return errors.Wrap(err, "unable to get Kubernetes config")
```

matthchr: I think it's OK to cap in the err just not the first letter (why not who knows?)

v2/tools/asoctl/cmd/import_azure_resource.go
```diff
@@ -11,13 +11,11 @@ import (
...
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
-	"github.com/pkg/errors"
-	"github.com/spf13/cobra"
-
+	"github.com/Azure/azure-service-operator/v2/api"
 	"github.com/Azure/azure-service-operator/v2/internal/genericarmclient"
 	"github.com/Azure/azure-service-operator/v2/internal/version"
-
-	"github.com/Azure/azure-service-operator/v2/api"
+	"github.com/pkg/errors"
```
matthchr, line 17: minor: I think the old import ordering was more correct, as the ASO stuff should be at the bottom and the GH stuff should be in the middle.

v2/tools/asoctl/cmd/log.go
```diff
@@ -0,0 +1,45 @@
...
+	// as parameters where needed.
+	log      logr.Logger
+	progress *mpb.Progress
+
+	verbose bool
+)
+
+func init() {
+	// Create Progressbar for console output
+	progress = mpb.New()
```
matthchr, line 27: Is there a compelling reason to do this in an `init` function vs some sort of `makeLogger() (logr.Logger, *mpb.Progress)` called in `main`? You're already passing them around everywhere pretty much rather than using them as globals so it seems like it'd be pretty easy to drop the `init` + globals entirely?

Is the issue the `root.go` usage?

v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -78,7 +83,8 @@ func (c *Cleaner) Run(ctx context.Context) error {
 
 		// If the slice was not updated, there is no version to deprecate.
 		if len(newStoredVersions) == len(crd.Status.StoredVersions) {
-			klog.Infof("Nothing to update for %q\n", crd.Name)
+			c.log.Info("Nothing to update",
```
matthchr, line 86: minor: kinda weird linebreaks here.

Consider:
```
c.log.Info(
    "Nothing to update",
    "crd-name", crd.Name)
```

or just putting everything on one line since it's pretty short.

matthchr: Applies below too

v2/tools/asoctl/internal/importing/resource_importer.go
```diff
@@ -159,6 +159,51 @@ func (ri *ResourceImporter) Import(ctx context.Context) (*ResourceImportResult,
...
+		if errors.As(err, &skipped) {
+			ri.log.V(1).Info(
+				"Skipped",
+				"kind", gk,
+				"name", rsrc.Name(),
+				"because", skipped.Because)
+			return nil
+		}
+
+		ri.log.Error(err,
```
matthchr, line 192: Do we need to log the error here and also return an error?

Will this error be logged above (resulting in a duplicate error log?)

matthchr (approved):
Some minor formatting comments primarily

v2/tools/asoctl/cmd/log.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/rs/zerolog"
+	"github.com/vbauerster/mpb/v8"
+)
+
+var (
+	// Global logger and progress bar
+	// While these are global here, they can't be referenced from internal packages and should be passed through
+	// as parameters where needed.
+	log      logr.Logger
+	progress *mpb.Progress
```
matthchr, line 20: How does the progress bar work with no `--output` flag? Will it break piping?

v2/tools/asoctl/cmd/log.go
theunrepentantgeek: Want to guarantee they're initialized before use. 

Have deliberately put them into `cmd` so they can't be referenced from `internal` without creating a circular reference, forcing them to be passed in, but this makes it possible for the variables to be directly referenced, bypassing any lazy-initialization.

Can't easily pass them through `cobra` from `root.go`, so this seemed a good compromise.

v2/tools/asoctl/cmd/crd_clean.go
theunrepentantgeek: Fixed in #2867

v2/tools/asoctl/internal/crd/cleaner.go
theunrepentantgeek: Deliberately put each name/value pair on a separate line to make them easier to visually parse.

theunrepentantgeek:
Having any logging at all will break piping, so it's not the progress bar that breaks it. 

Maybe we need a `--silent` flag to suppress logging?

v2/tools/asoctl/cmd/log.go
theunrepentantgeek: Having any logging at all will break piping, so it's not the progress bar that breaks it.

Maybe we need a --silent flag to suppress logging?

matthchr: This is why many Linux tools emit their logs to `stderr` rather than `stdout`, I think.

Structured output goes to `stdout`, log output goes to `stderr` always, and then piping works by default.

If piping doesn't work, then `-o` being an optional flag seems somewhat pointless, because there's not a lot of value in running the tool and emitting to `stdout` given you can't actually pipe that to a file (or to `kubectl` directly)?

v2/tools/asoctl/internal/crd/cleaner.go
matthchr: That's fine, but you should put a newline after the `Info(`, otherwise everything ends up over-tabbed, right?

v2/tools/asoctl/internal/importing/resource_importer.go
theunrepentantgeek: Want to log it here so the presence of errors can be seen during the run, and in the right context.

Also need to return an error to ensure `asoctl` doesn't exit with a success code. 

Maybe need some fine tuning on what's in each message?

v2/tools/asoctl/internal/crd/cleaner.go
theunrepentantgeek: Agreed. Updated.

PR #2865, opened 2023-04-12
Documentation for Import Azure Resource

theunrepentantgeek:

Some basic documentation on `asoctl import azure-resource` and how it works.

gif

**If applicable**:


docs/hugo/content/tools/asoctl.md
```diff
@@ -33,4 +30,75 @@ TBC
...
-TBC
+When you have an existing Azure resource that needs to be managed by ASO, you can use the `import` command to generate a YAML file that can be used to create a new ASO resource. This is useful when:
+
+* You want to migrate an existing Azure resource from ASvO v1 to ASO v2.
+* You have an existing Azure resource and want ASO to maintain it in a particular state.
+* You have a hand-configured Azure resource and you want to replicate it for development, staging, or production use.
+
+``` bash
+$ asoctl import azure-resource --help
+imports an ARM resource as a CR
```
matthchr, line 41: I think we should clarify in this help text (in the tool and here) that it imports the resource and all child resources?

It sounds like a single resource here but then it's actually not.

matthchr (approved):

PR #2866, opened 2023-04-12
Improve/relocate asoctl

### LINKED ISSUE #2782: Relocate asoctl to v2/cmd

Currently, `asoctl` lives under `v2/tools` alongside `generator`.

**Describe the improvement**

We should consider moving it to `v2/cmd` alongside `controller` as it's an executable we're shipping for user consumption. This contrasts with `generator` which is an internal tool used to generate code for the operator.

**Additional context**

For reference, there is a [standard go project layout](https://github.com/golang-standards/project-layout) which should be considered,.

theunrepentantgeek:

Relocates `asoctl` from `v2/tools` into `v2/cmd`.

User facing command line utilities should be in the `cmd` folder; `tools` is for internal tooling.

Closes #2782 

**Prerequisites**

- [x] #2862 

gif


v2/cmd/asoctl/go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/Azure/azure-service-operator/v2/tools/asoctl
```
super-harsh, line 1: Do we need to be consistent on where we have the go.mod for the `controller` as well?

super-harsh:
I'd suggest moving the `cmd` and `internal` package in `v2/cmd/asoctl` to `v2/cmd/asoctl/app/`. Just like how we have for `v2/cmd/controller`

theunrepentantgeek:
I disagree. If we need consistency, the `controller/app` directory could be removed - but I don't see the need.

v2/cmd/asoctl/go.mod
theunrepentantgeek: The `go.mod` and `go.sum` files have to be at the root of the project, so we don't have any option on where to put them.

matthchr:
https://github.com/golang-standards/project-layout/tree/master/cmd#readme suggests:

> **Don't put a lot of code in the application directory.** If you think the code can be imported and used in other projects, then it should live in the /pkg directory. If the code is not reusable or if you don't want others to reuse it, put that code in the /internal directory. You'll be surprised what others will do, so be explicit about your intentions!
> It's common to have a small main function that imports and invokes the code from the /internal and /pkg directories and nothing else.

(emphasis mine)

Moving from `v2/tools/asoctl` to `v2/cmd/asoctl` is definitely good and we should do it.

From the project-layout links, I observe that some tools definitely don't obey the "Don't put a lot of code in the application directory" rule. Although I guess it's in the eye of the beholder what constitutes "a lot of code".

I don't see any of those linked projects that have a `v2/cmd/<tool>/cmd` structure, nor do any have a `v2/cmd/<tool>/internal` structure. I do see that [cluster API](https://github.com/kubernetes-sigs/cluster-api/tree/main/cmd/clusterctl) does this but that's the only project I've found that does.

My interpretation of the directive: "If the code is not reusable or if you don't want others to reuse it, put that code in the /internal directory" is that it is directing us to put internal code in `v2/internal`, _not_ `v2/cmd/asoctl/internal`.

This is I think a somewhat "religious" discussion as unfortunately Go's project structure guidelines seem pretty loosely followed as a whole and so different projects do different things. Still, I think my preference is to follow the written guideline and do something like this:

```
v2/
    cmd/
        asoctl/
            main.go
            cmd (or app, I don't really care although I think we should probably be consistent)/
                clean-crds.go
                clean.go
                import.go
                import-azure-resource.go
                log.go
                root.go
    internal/
        asoctl/
            crd/
                cleaner.go
                cleaner_test.go
            importing/
                (all the importing files)
```

Or we could drop the `internal/asoctl` folder and promote its contents to `internal/crd` and `internal/importing` - it depends on how specific to asoctl we feel those things are.

theunrepentantgeek:
I feel that packages specific to `asoctl` - stuff that _cannot_ and _will not_ used elsewhere should be under `asoctl/internal`. Everything reusable should be under `/internal` and `/pkg`.

matthchr:
> I feel that packages specific to asoctl - stuff that cannot and will not used elsewhere should be under asoctl/internal. Everything reusable should be under /internal and /pkg.

In practice what does that end up looking like though? Is there a compelling argument for that over `v2/internal/asoctl/<stuff>` for stuff that is "for asoctl"?

I'm not necessarily saying that I _love_ everything existing under the one `internal` directory, but that seems to me to be what the recommendation from go-standards is and also what a large number of linked projects do.

It's not a hill I'm ready to die on though, I just don't see many projects with `internal` as a subfolder under `cmd`. I really only see CAPI doing that.

matthchr (approved):
Approved, though we should revisit the `cmd/internal/...` structure at some point in the future.

PR #2871, opened 2023-04-13
Document policy and history of breaking changes

### LINKED ISSUE #2192: Design: How do we manage ASO upgrades with breaking changes?

Thus far we've managed to ensure each release of ASO v2 is backward compatible with resources used with the previous releases. This applies even for the prefix change from `v1alpha1api` to `v1beta`.

We need to have a plan, and hopefully a design, for what we are going to do when we need to release a breaking change of ASO. 

How do we make a breaking change to the structure of our generated resources without causing havoc for our existing users?

**Describe the improvement**

Ideally, we'd be able to make this change to the actual structure of our custom resources (CRs) without breaking any existing users, including those who are deploying ASO "v.latest" automatically using a gitOps workflow.

One suggested approach is to serialize the old CR to ARM API format (which won't have changed) and then deserialize to the new object, effectively tunnelling through the change via ARM.

A prerequisite for this approach would be to always retain the existing CR object structure and _add_ the new structure by using a new prefix (following the same approach as we used in PR #2163 when moving to beta).

theunrepentantgeek:

We need to be more explicit with our approach to breaking changes and how we're avoiding issues.

Closes #2192

gif

**If applicable**:


docs/hugo/content/introduction/breaking-changes/_index.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Released Breaking Changes
+
+### v2.0.0
+
+Breaking changes are:
+
+* Upgrades from releases prior to v2.0.0-beta.5 are disallowed
+* Alpha CRD versions have been removed
+
```
matthchr, line 15: Missing mention on the `ResourceGroup` status change?

docs/hugo/content/introduction/breaking-changes/_index.md
```diff
@@ -0,0 +1,65 @@
...
+* Objects that were incorrectly generated as nested properties have been directly inlined
+* ARM Id fields are now correctly identified as references, allowing linking to a resource in Kubernetes instead of only in Azure.
+* Status properties that cannot be set by the end user on a Spec that were included in the Spec in error
+* Sub-resources that were incorrectly inlined into the parent resource have been moved to a separate resource.
+* Properties that previously included on Spec but actually had no function have been removed.
+* Validation rules have been tightened, or added to properties that previously had no validation rules.
+* 
+For detailed information, including an exhaustive list of all affected resource properties, see [v2.0.0 Breaking Changes](./breaking-changes-v2.0.0-beta.4).
+
+## Upcoming Breaking Changes
```
matthchr, line 35: I feel like this should be at the top, rather than after the released breaking changes?

Or alternatively, it could be in a section under the version it is announced with, as otherwise the "upcoming" list of breaking changes may change over time in a way that doesn't preserve history very well?

Something like this:

```
## v2.0.0

### Upcoming Breaking Changes
...

### Breaking changes
...
```

maybe?

matthchr: Actually I take it back, I think I like it how you have it here, but maybe we should call out the versions it is upcoming in more clearly (and move this to the top I think?)

```
## Upcoming Breaking Changes

### 2.3.0 or later
Beta CRD versions (any version with `v1beta` prefix) will be deprecated no sooner than **v2.3.0**. 
We recommend you start using `v1api` prefixed versions now. 

You can easily swap from a `v1beta` version to a `v1api` version by just replacing `v1beta` with `v1api` in your CRD YAML, the resource shapes are the same.
```

That way we can also have a `2.4.0 or later` section in the future?

docs/hugo/content/introduction/breaking-changes/_index.md
```diff
@@ -0,0 +1,65 @@
...
+You can easily swap from a `v1beta` version to a `v1api` version by just replacing `v1beta` with `v1api` in your CRD YAML, the resource shapes are the same.
+
+## Policies
+
+Prior to the v2.0.0 GA release, our policy was *avoid breaking changes*. 
+We're now upgrading that to *no avoidable breaking changes*.
+
+If we do encounter potential breaking changes, we have a number of potential mitigations.
+
+### Conversion augmentation
```
matthchr, line 49: This feels very developer-facing rather than user-facing, as users may not understand how we do intra-version conversions (or that we even do).

I wonder if it would be better to outline the commitments we're giving first, before talkin about mitigations. I think those commitments are:
1. We will not introduce a breaking change in an existing version of a resource such as `v1api20230101`
2. We _may_ introduce breaking changes between versions of resources, such as between `v1api20230101` and `v1api20230201`. These two versions correspond to two different Azure API versions, and if the upstream Azure service introduces a breaking change in the newer of these two API versions we may pass that breaking change along. When this happens we will do our best to hide these breaking changes from our users (see mitigations below).

This more explicitly spells out what exactly "no avoidable breaking changes" means.

docs/hugo/content/introduction/breaking-changes/_index.md
theunrepentantgeek: Added.

docs/hugo/content/introduction/breaking-changes/_index.md
theunrepentantgeek: As suggested, moved to the top with a heading for the version.

docs/hugo/content/introduction/breaking-changes/_index.md
theunrepentantgeek: Good idea. Adopted and expanded.

docs/hugo/content/introduction/breaking-changes/_index.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Mitigation techniques
+
+**Automatic conversion**: By manually augmenting the intra-version conversions produced by our code-generator, we may be able to automatically upgrade resources from the older version to the new. Migration would be zero-touch for our users.
+
+We've done this before when a new version of a resource API introduced a breaking change, see [managed_cluster_conversion_overrides.go](https://github.com/Azure/azure-service-operator/blob/main/v2/api/containerservice/v1api20210501storage/managed_cluster_conversion_overrides.go).
+
+**Custom Tooling**: For our **v2.0.0** release, we provided a specialized tool (`asoctl`) specifically designed to smooth the way for upgrading users. In addition to the existing two modes, we may add further functions to cater for other scenarios.
+
+For users upgrading from ASO v1, `asoctl import azure-resource` provides a way to scaffold an ASO v2 resource based on an existing Azure resource.
```
super-harsh, line 67: Maybe add links here to the docs we've got for `asoctl import azure-resource` and the below `asoctl clean crds`

theunrepentantgeek: Done.

docs/hugo/content/introduction/breaking-changes/breaking-changes-v2.0.0.md
```diff
@@ -0,0 +1,25 @@
...
+
+We changed how we manage CRDs in this release (see [PR#2769](https://github.com/Azure/azure-service-operator/pull/2769)), and as a result if using Helm you _must_ upgrade from v2.0.0-beta.5 to v2.0.0.
+
+You cannot upgrade from v2.0.0-beta.4 or earlier directly to v2.0.0. This is enforced with a Helm upgrade hook.
+
+This restriction is just for upgrades to the v2.0.0 version, although [we always recommend](https://azure.github.io/azure-service-operator/introduction/upgrading/#recommended-upgrade-pattern) upgrading one version at a time.
+
+## Alpha CRD versions have been removed
+
+**You cannot successfully upgrade to v2.0.0 until you have followed our [migration guide](TODO)**.
```
matthchr, line 18: I will update this TODO link when I send a PR for the v2.0.0 migration guide

matthchr (approved):

PR #2880, opened 2023-04-14
Improve asoctl clean crds logging

super-harsh:

Found a bug where having no ASO CRDs installed, would not return a correct error. This PR adds logic to count the seen ASO CRDs and return the error if not found.


v2/tools/asoctl/internal/crd/cleaner.go
```diff
@@ -114,6 +111,10 @@ func (c *Cleaner) Run(ctx context.Context) error {
 		updated++
 	}
 
+	if asoCrdsSeen <= 0 {
+		return errors.New("found 0 results, make sure you have ASO CRDs installed")
```
theunrepentantgeek, line 115:
```suggestion
		return errors.New("found no Azure Service Operator CRDs, make sure you have ASO installed.")
```

theunrepentantgeek (approved):

Issue #2687, opened 2023-01-19
Plan for supporting lots of versions in a single CRD


matthchr:
Some services release API versions very frequently. AKS for example ships a new API version nearly every month. Each version often supports a few new features which might be interesting to a small subset of customers.

We don't currently automatically generate a new CRD version for each API version. Regardless, as time goes on the number of CRD versions we support for `ManagedCluster` will increase without bound. At some point, we will reach the limit on CRD size and be forced to deprecate some versions (even if the service itself has not deprecated that particular API version yet).

**We should define how this deprecation will happen and which APIs will be chosen to remove**

For example, we could say that certain API versions every year are LTS (meaning we'll keep them for 3+ years):
20200101 == LTS (supported for 3 years)
20200301 == non-LTS but supported (will remove after a year?)
20200501 == non-LTS
20200801 == non-LTS
20210101 == LTS or something?

theunrepentantgeek:
Still something we need to consider, especially as the number of CRDs in ASO climbs.

matthchr:
We also need to define a timeline for how long we announce a deprecation of an API version before we remove it. I've spoken with some customers who want at least a year, which at least for GA APIs seems pretty reasonable to me.

dtzar:
I think it's fair for preview API versions to only support something like N-1.

matthchr:
I just ran into this when adding the newer version of AKS ManagedCluster API. It looks like at least for that API, having 6 variants of the API (including preview/GA) along with all of the documentation for each field pushes the CRD file to ~2.5M, which ends up crossing the 1.5M max size limit when we submit the CRD.

I don't fully follow why the filesize differs from the actual limit (I assume something to do with encodings).

In terms of what we should do in the near-term at least, I think we need to announce the deprecation of some of the API versions of ManagedCluster to reduce the size of this CRD to a more manageable level.

At least for this resource, we should commit to 2 preview versions and 3 GA versions? We can fit more if we strip documentation, which I don't love.

matthchr:
We can also (eventually) fit more when AKS supports #2791, as we can strip the AgentPool from inside MC...

dtzar:
For AKS - I propose only the latest preview version and N-1 GA versions.

PR #2881, opened 2023-04-14
Document handling of Azure Deprecations

### LINKED ISSUE #2070: Deprecate CRDs when Azure offerings deprecations are announced
Today, there's no change to the support for an Azure Service Operator CRD when the underlying Azure service/offering is being deprecated. How should we communicate these deprecations and what timelines should we follow? How will we keep our catalog up to date?

theunrepentantgeek:

Document how we can handle Azure resource deprecation.

Closes #2070 

gif

**If applicable**:


docs/hugo/content/design/ADR-2023-04-Deprecation.md
```diff
@@ -0,0 +1,48 @@
...
+title: '2023-04: Resource Deprecation'
+---
+
+## Context
+
+As succinctly raised by `@sakthi-vetrivel` in (#2070)[https://github.com/Azure/azure-service-operator/issues/2070]:
+
+> Today, there's no change to the support for an Azure Service Operator CRD when the underlying Azure service/offering is being deprecated. How should we communicate these deprecations and what timelines should we follow? How will we keep our catalog up to date?
+
+Starting in Kubernetes v1.19, it's possible to flag individual versions of a custom resource as (deprecated)[https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#version-deprecation]. This results in the follow behaviour:
```
matthchr, line 11: I believe there is also a CRD max-size, so we may end up with an issue where there are so many supported versions of a resource that we can't fit them all.

#2687 is tracking this.

Unsure if you want to talk about that here at all or not, but in theory we can use a similar approach where we realize we're getting close to the limit and start deprecating older versions.

matthchr (approved):

docs/hugo/content/design/ADR-2023-04-Deprecation.md
theunrepentantgeek: You're absolutely correct, but this ADR is specifically about deprecation of Azure resources, so I don't think that information belongs here.

super-harsh (approved):

PR #2887, opened 2023-04-17
Fix Live Validation tests workflow

super-harsh:

Live validation tests are constantly failing on `waiting for CRD cabundles` task. This PR is to bump up the timeout from 1m to 2m. 

Workflow run from current branch: [run](https://github.com/Azure/azure-service-operator/actions/runs/4717534379)


matthchr (approved):

v2/internal/controllers/crd_keyvault_test.go
```diff
@@ -130,6 +128,9 @@ func newVault(tc *testcommon.KubePerTestContext, rg *resources.ResourceGroup) *k
 					Name:   &skuName,
 				},
 				TenantId: to.Ptr(tc.AzureTenant),
+				// EnableSoftDelete is true as default. We need to set this false to purge delete the KeyVault so there's no while re-running tests.
```
theunrepentantgeek, line 131:
```suggestion
				// EnableSoftDelete is true as default. We need to set this false to purge delete the KeyVault so there's no conflict while re-running tests.
```

theunrepentantgeek (approved):
Just one wording change.

PR #2905, opened 2023-04-18
Improve UX for asoctl import azure-resource

### LINKED ISSUE #2868: Add --quiet flag to asoctl to silence non-error logging

Some users may wish to pipe the output of asoctl into another tool, and the logging might get in the way.

**Describe the improvement**

Add a `--quiet` flag to suppress non-error messages.

### LINKED ISSUE #2900: Bug: --verbose doesn't work for asoctl
**Describe the bug**

Using `--verbose` has no effect on the output.

**Expected behavior**

More information (`DBG` level) should be included if the commandline includes `--verbose`.

### LINKED ISSUE #2904: Feature: Summarize results of asoctl import azure-resource
When `asoctl import azure-resource` completes, log a summary of what has been imported.

theunrepentantgeek:

Add --quiet flag to suppress most logging 
Add summary of what was imported (or not) #2904
Fixes for --verbose not working.  #2900
Fix for partial progress bars remaining on screen

Closes #2868 
Closes #2900
Closes #2904 

**Prerequisites**

- [x] #2866 

gif

### DIFF
--- a/v2/cmd/asoctl/cmd/log.go
+++ b/v2/cmd/asoctl/cmd/log.go
@@ -38,8 +38,6 @@ func CreateLogger() logr.Logger {
 
 	// Use standard interface for logging
 	zerologr.VerbosityFieldName = "" // Don't include verbosity in output
-	zerologr.SetMaxV(0)              // Default to quiet
-
 	log := zerologr.New(&zl)
 	return log
 }
@@ -63,7 +61,6 @@ func CreateLoggerAndProgressBar() (logr.Logger, *mpb.Progress) {
 
 	// Use standard interface for logging
 	zerologr.VerbosityFieldName = "" // Don't include verbosity in output
-	zerologr.SetMaxV(0)              // Default to quiet
 
 	log := zerologr.New(&zl)
 	return log, progress
--- a/v2/cmd/asoctl/cmd/log.go
+++ b/v2/cmd/asoctl/cmd/log.go
@@ -16,7 +16,7 @@ import (
 
 var (
 	verbose bool // True if the user specifies --verbose or -v
-	quiet   bool // True if the user specifies --quiet or -s
+	quiet   bool // True if the user specifies --quiet or -q
 )
 
 // CreateLogger creates a logger  for console output.
--- a/v2/cmd/asoctl/cmd/log.go
+++ b/v2/cmd/asoctl/cmd/log.go
@@ -15,16 +15,13 @@ import (
 )
 
 var (
-	// Global logger and progress bar
-	// While these are global here, they can't be referenced from internal packages and should be passed through
-	// as parameters where needed.
-	verbose bool
+	verbose bool // True if the user specifies --verbose or -v
+	quiet   bool // True if the user specifies --quiet or -s
 )
 
 // CreateLogger creates a logger  for console output.
 // Use this when your command wants to show only log messages
 func CreateLogger() logr.Logger {
-
 	// Configure console writer for ZeroLog
 	output := zerolog.ConsoleWriter{
 		Out:        os.Stderr,      // Write to StdErr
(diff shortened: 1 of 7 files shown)


v2/cmd/asoctl/cmd/log.go
```diff
@@ -15,16 +15,13 @@ import (
 )
 
 var (
-	// Global logger and progress bar
-	// While these are global here, they can't be referenced from internal packages and should be passed through
-	// as parameters where needed.
-	verbose bool
+	verbose bool // True if the user specifies --verbose or -v
+	quiet   bool // True if the user specifies --quiet or -s
```
matthchr, line 19: `--quiet` == `-s` seems weird?

Why not `-q`? Or is `-s` supposed to be short for `silent` - in which case why not `--silent` to match?

v2/cmd/asoctl/cmd/root.go
```diff
@@ -56,8 +60,14 @@ func newRootCommand() (*cobra.Command, error) {
 	}
 
 	rootCmd.PersistentPreRun = func(cmd *cobra.Command, args []string) {
+		// Configure logging; --verbose overrides --quiet
```
matthchr, line 63: Could instead error if both are set?

v2/cmd/asoctl/internal/importing/importable_arm_resource.go
```diff
@@ -326,6 +326,14 @@ func (*importableARMResource) classifyError(err error) (string, bool) {
 			// If we're not allowed to look, we can still import the rest
 			return "access forbidden", true
 		}
+
+		if responseError.StatusCode == http.StatusBadRequest &&
+			strings.Contains(responseError.Error(), "RequestUrlInvalid") {
+			// We constructed an invalid URL
+			// (Seems that some extension resources aren't permitted in all locations)
```
matthchr, line 333: What exactly does "not permitted in all locations" mean? Are you referring to location here as an Azure region? Or do you mean "URL location"?

Maybe an example of when this might happen?

Or alternatively, is there a way we can know that a URL is/isn't bad and just not construct it in the first place? Rather than inferring from response that we must have generated something bad but it's OK? For example if this method is used to classify errors for both standard and extension resources, presumably this `RequestUrlInvalid` is not expected for a standard resource, but we'd ignore it anyway?

v2/cmd/asoctl/internal/importing/resource_importer_report.go
```diff
@@ -0,0 +1,134 @@
...
+	"sync"
+
+	"github.com/go-logr/logr"
+	"golang.org/x/exp/maps"
+	"golang.org/x/exp/slices"
+)
+
+// resourceImportReport is used to generate a report of the resources that were imported (or not).
+type resourceImportReport struct {
+	content map[resourceImportReportKey]int
```
matthchr, line 18: Leave a comment explaining what exactly the int is/represents?

matthchr (approved):
Had a few comments but looks good overall

v2/cmd/asoctl/cmd/log.go
theunrepentantgeek: The short form is `-q`; have fixed the comment.

v2/cmd/asoctl/cmd/root.go
theunrepentantgeek: Now logging an error, but don't want to abort the run for a trivial matter.

v2/cmd/asoctl/internal/importing/importable_arm_resource.go
theunrepentantgeek: Got a couple of errors where the constructed URL for listing extension resources underneath another resource returned `RequestUrlInvalid`. I've managed to tweak the logic to mostly avoid doing this (by avoiding nesting of extension resources) but it still came up in a couple cases.

v2/cmd/asoctl/internal/importing/resource_importer_report.go
theunrepentantgeek: Updated.

PR #2908, opened 2023-04-19
Allow import of entire Resource Groups

### LINKED ISSUE #2878: Feature: Allow import of entire resource group
Currently `asoctl import azure-resource` will import a ResourceGroup but not any of the resources within the group.

All supported resources within the resource group should be found and imported.

theunrepentantgeek:

Allows `asoctl import azure-resource` to import all the supported resources in a resource group.

Also improves the uniq'ing of resource import so we don't try to process the same resource multiple times (useful to avoid reprocessing the same `RoleAssignment` dozens of times).

Closes #2878 

**Prerequisites**

- [x] #2866 

gif


v2/cmd/asoctl/internal/importing/find_resources_by_scope.go
```diff
@@ -0,0 +1,81 @@
...
+
+// ensureResourceTypesByScope ensures that the resourceTypesByScope map is populated
+func ensureResourceTypesByScope() {
+	populateResourceTypesByScope.Do(func() {
+		resourceTypesByScope = createResourceTypesByScope()
+	})
+}
+
+func createResourceTypesByScope() map[genruntime.ResourceScope]set.Set[string] {
+	result := make(map[genruntime.ResourceScope]set.Set[string], 4)
```
matthchr, line 54: minor: You don't need to specify 4 here, the linter is fine without it. I know 4 is technically correct (as there are 4 possible scopes), but in the absence of a way to actually pin the initial size of this collection to the number of supported scopes (there's no `len(enum)` unfortunately) it feels like it would be better to just not specify an initial size at all?

v2/cmd/asoctl/internal/importing/find_resources_by_scope.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"sync"
+
+	"github.com/Azure/azure-service-operator/v2/api"
+	"github.com/Azure/azure-service-operator/v2/internal/set"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+)
+
+var (
+	resourceTypesByScope         map[genruntime.ResourceScope]set.Set[string]
```
matthchr, line 17: minor, leave a comment explaining the set of `strings` is of `typeName`'s?

v2/cmd/asoctl/internal/importing/resource_importer.go
```diff
@@ -74,18 +77,14 @@ func (ri *ResourceImporter) AddARMID(armID string) error {
 // This allows us to call it from other methods that already have the lock
 func (ri *ResourceImporter) addImpl(importer ImportableResource) {
 	// If we've already handled this resource, skip it
```
matthchr, line 79: Why didn't the previous implementation work?

If something is in the `pending` set or the `completed` set, we're already processing it (or processed it) and it doesn't need to be acted on again. Otherwise, it does?

Why do we also need `unique`?

theunrepentantgeek: Because `completed` only lists resources we _successfully_ completed; If we have an error trying to handle the resource (say, `access denied`) then it didn't show up in `completed` and we retried (and refailed) every time we encountered it.

v2/cmd/asoctl/internal/importing/find_resources_by_scope.go
theunrepentantgeek: I added this line _because the linter complained_.
It feels very nondeterministic. 
I'll remove the size and wait for it to complain.

v2/cmd/asoctl/internal/importing/find_resources_by_scope.go
theunrepentantgeek: Comment added.

v2/cmd/asoctl/internal/importing/resource_importer.go
matthchr: Should `completed` be called `completedSuccessfully` or `successful` then? Obviously minor

v2/cmd/asoctl/internal/importing/find_resources_by_scope.go
matthchr: I believe the rule is that if the loop has no returns or breaks, it wants you to prealloc it. If there _are_ returns or breaks such that it's not obvious what the actual resultant count will be, you don't need to prealloc.

See: https://github.com/alexkohler/prealloc#flags

matthchr (approved):

v2/cmd/asoctl/internal/importing/resource_importer.go
theunrepentantgeek: Renamed the field to `imported` (past tense).

PR #2909, opened 2023-04-19
Create workflow to validate documentation links in PRs

theunrepentantgeek:

Adds a workflow to validate documentation links on PRs.

gif

### DIFF
--- a/.github/workflows/pr-validation-docs.yml
+++ b/.github/workflows/pr-validation-docs.yml
@@ -62,7 +62,7 @@ jobs:
           docker start "$container_id"
           echo "id=$container_id" >> $GITHUB_OUTPUT
 
-      - name: Build & validate docs site
+      - name: Validate docs site
         run: |
           container_id=${{steps.devcontainer.outputs.id}}
           docker exec "$container_id" task check-docs-site-links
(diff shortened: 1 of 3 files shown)


super-harsh (approved):
LGTM! You might wanna remove this check from [deploy-site workflow file](https://github.com/Azure/azure-service-operator/blob/main/.github/workflows/deploy-site.yml#L57)

theunrepentantgeek:
Given the check is cheap (takes a couple of minutes), I think leaving it in the deploy site workflow is advisable, at least until we gain confidence in the new workflow working.

theunrepentantgeek:
Ready for merge once @matthchr has confirmed the trigger for forks is correct.

.github/workflows/pr-validation-docs.yml
```diff
@@ -0,0 +1,73 @@
...
+  workflow_dispatch:
+    # no content, allows manual triggering
+
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "docs/**"
+
+  repository_dispatch:
```
matthchr, line 12: I actually don't think we want/need `ok-to-test` to do anything with this. It can be triggered the same way the `test-generator` job is in `pr-validation`, which is to say, like:

```
  pull_request: 
    branches:
      - main

  merge_group:
    branches:
      - main
```

(with the added docs restriction).

It doesn't need ok-to-test because it doesn't require subscription access.

matthchr (approved):

PR #2916, opened 2023-04-21
Add support for Search/SearchService

### LINKED ISSUE #2843: Feature: Add Azure Cognitive Search
Feature: Azure Cognitive Search

Did not find a CRD for Azure Cognitive Search. Please, consider adding it.
[ARM tempalte](https://learn.microsoft.com/en-us/azure/templates/microsoft.search/searchservices?pivots=deployment-language-arm-template)

super-harsh:
Closes #2843 

**If applicable**:
- [x] this PR contains tests


v2/api/network/v1api20220401preview/structure.txt
```diff
@@ -0,0 +1,213 @@
+github.com/Azure/azure-service-operator/v2/api/network/v1api20220401preview
```
matthchr, line 1: Confused where these network files are coming from? Do you need to rebase or otherwise update the PR? I don't see any reason from `azure-arm.yaml` why they would be here.

matthchr (changes requested):
They have a more recent GA API: https://github.com/Azure/azure-rest-api-specs/tree/main/specification/search/resource-manager/Microsoft.Search/stable/2022-09-01

We should use that instead, right?

v2/api/network/v1api20220401preview/structure.txt
super-harsh: Good catch! I've removed these files and updated the version to `2022-09-01`. Thanks!

theunrepentantgeek (approved):

matthchr (approved):

v2/api/search/v1api20220901/structure.txt
```diff
@@ -0,0 +1,222 @@
...
+│   │   │   ├── "default"
+│   │   │   └── "highDensity"
+│   │   ├── Identity: *Object (1 property)
+│   │   │   └── Type: *Enum (2 values)
+│   │   │       ├── "None"
+│   │   │       └── "SystemAssigned"
+│   │   ├── Location: *string
+│   │   ├── NetworkRuleSet: *Object (1 property)
+│   │   │   └── IpRules: Object (1 property)[]
+│   │   │       └── Value: *string
```
matthchr, line 31: Wait is this supposed to be an ARM ID? We may need to annotate it as such in our YAML

v2/api/search/v1api20220901/structure.txt
```diff
@@ -0,0 +1,222 @@
...
+│   └── "2022-09-01"
+├── SearchService: Resource
+│   ├── Owner: github.com/Azure/azure-service-operator/v2/api/resources/v1apiv20191001.ResourceGroup
+│   ├── Spec: Object (14 properties)
+│   │   ├── AuthOptions: *Object (2 properties)
+│   │   │   ├── AadOrApiKey: *Object (1 property)
+│   │   │   │   └── AadAuthFailureMode: *Enum (2 values)
+│   │   │   │       ├── "http401WithBearerChallenge"
+│   │   │   │       └── "http403"
+│   │   │   └── ApiKeyOnly: map[string]v1.JSON
```
matthchr, line 13: Hmm, what is this `APiKeyOnly` field? Its type looks weird.

v2/api/search/v1api20220901/structure.txt
super-harsh: Not really, have a look at the [docs](https://learn.microsoft.com/en-us/azure/templates/microsoft.search/2021-04-01-preview/searchservices?pivots=deployment-language-bicep#iprule)

v2/api/search/v1api20220901/structure.txt
super-harsh: yes it does, even I'm a bit unsure about the type of this. Official documentation mentions about using `any()`. Terraform completely omits this field though 🤔

super-harsh: https://learn.microsoft.com/en-us/azure/templates/microsoft.search/2021-04-01-preview/searchservices?pivots=deployment-language-bicep#dataplaneauthoptions

super-harsh: Makes more sense to make it `map[string]string` though?

v2/api/search/v1api20220901/structure.txt
matthchr: Looks good.

v2/api/search/v1api20220901/structure.txt
theunrepentantgeek: I'm not sure this is related to #2924, though it's coincidentally similar.

The _semantics_ of `akiKeyOnly` are that it's a Bool, with the default being `false`. 

I think we need to use a type transformation in our config to force this on the resource, and then do something to ensure we can generate the right ARM payload separately.

matthchr: We could consider suppressing generation of the `authOptions` section here for now, until we have a good way to model it so as to avoid breaking change?

Idea courtesy of @theunrepentantgeek

PR #2918, opened 2023-04-24
Add support for network/dnszone and network/dnszone/recordSets

### LINKED ISSUE #2901: Feature: add support for public DNS
As `privateDNS` is already supported, i would like to have `publicDNS` supported as well.

https://azure.github.io/azure-service-operator/reference/#network

super-harsh:
Closes #2901 

Add support for `network/dnszone` and `network/dnszone/recordSets`

**If applicable**:
- [x] this PR contains tests


v2/api/network/v1api20180501/dns_zone_types_gen.go
```diff
@@ -0,0 +1,1361 @@
...
+
+const APIVersion_Value = APIVersion("2018-05-01")
+
+type DnsZone_Spec struct {
+	// AzureName: The name of the resource in Azure. This is often the same as the name of the resource in Kubernetes but it
+	// doesn't have to be.
+	AzureName string `json:"azureName,omitempty"`
+
+	// Etag: The etag of the zone.
+	Etag *string `json:"etag,omitempty"`
```
matthchr, line 346: I am pretty sure that this field should be readonkly? Maybe fix it in our config, and/or in Swagger upstream?

v2/api/network/v1api20180501/dns_zones_aaaa_record_types_gen.go
```diff
@@ -0,0 +1,4064 @@
...
+	AzureName string `json:"azureName,omitempty"`
+
+	// CNAMERecord: The CNAME record in the  record set.
+	CNAMERecord *CnameRecord `json:"CNAMERecord,omitempty"`
+
+	// CaaRecords: The list of CAA records in the record set.
+	CaaRecords []CaaRecord `json:"caaRecords,omitempty"`
+
+	// Etag: The etag of the record set.
+	Etag *string `json:"etag,omitempty"`
```
matthchr, line 353: Same here and presumably for all of the other types as well?

v2/api/network/v1beta20201101storage/load_balancer_types_gen.go
```diff
@@ -1358,8 +1359,13 @@ func (embedded *BackendAddressPool_STATUS_LoadBalancer_SubResourceEmbedded) Assi
 			if err != nil {
 				return errors.Wrap(err, "calling AssignProperties_From_SubResource_STATUS() to populate field SubResource_STATUSStash from LoadBalancingRules")
 			}
+			var subResourceSTATUSStashLocal v1api20180501s.SubResource_STATUS
+			err = subResourceSTATUSStashLocal.AssignProperties_From_SubResource_STATUS(&subResourceSTATUSStash)
+			if err != nil {
```
matthchr, line 1364: @theunrepentantgeek just confirming this is right. I think it is given the addition of the older API version but just confirming.

v2/internal/controllers/crd_networking_dnszone_test.go
```diff
@@ -0,0 +1,88 @@
...
+			Test: func(tc *testcommon.KubePerTestContext) {
+				DNSZones_CNAME_Record_Test(tc, zone)
+			},
+		},
+	)
+
+	tc.DeleteResourceAndWait(zone)
+
+	//Ensure that the resource was really deleted in Azure
+	armId, hasID := genruntime.GetResourceID(zone)
```
matthchr, line 46: Weird to do this here -- move up above the delete?

v2/samples/network/v1api20180501/v1api20180501_dnszonesarecord.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: network.azure.com/v1api20180501
+kind: DnsZonesARecord
+metadata:
+  name: aso-sample-record-a
+  namespace: default
+spec:
+  ARecords:
```
matthchr, line 7: Case looks wrong here?

matthchr: Ugh, i'ts wrong in their API. Ok yeah we should match that.

v2/api/network/v1api20180501/dns_zone_types_gen.go
super-harsh: Did not seem like a read-only field to me - https://learn.microsoft.com/en-us/azure/dns/dns-zones-records#etags

matthchr: Are you sure? They say this:
"At the level of the Azure DNS REST API, Etags are specified using HTTP headers. Their behavior is given in the following table:"

So they are basically saying what you do is, you do a GET, you take the value of the Etag returned by the server, and you put it as the HTTP header `if-none-match`/`if-match`, etc.

AFAIK the value in the body is always only read from the server. The instruction to the server comes in the form of an HTTP header `if-none-match`/`if-match`. Because Etags are only ever generated by the server, it doesn't make any sense to allow the clients to set them - what would they set them to? They don't understand how to make new ones.

At least, that's how all the ETag's I've ever used work.

matthchr: Maybe try setting it to something and see what the server does? They might just ignore it in which case that doesn't really tell us anything, but they might reject.

matthchr: I'm 99% sure that this should be readonly though

v2/api/network/v1api20180501/dns_zones_aaaa_record_types_gen.go
matthchr: I don't think this was fixed?

v2/azure-arm.yaml
```diff
@@ -213,6 +213,60 @@ typeTransformers:
     property: Type
     remove: true
     because: it is supposed to be marked ReadOnly in Swagger
+  - group: network
+    version: "*"
+    name: DnsZone_Spec
+    property: Etag
+    remove: true
+    because: it is supposed to be marked ReadOnly in Swagger
```
theunrepentantgeek, line 221: The values here are multi-valued, so you can combine these together:

```suggestion
  - group: network
    version: "*"
    name: DnsZone_Spec; DnsZones_AAAA_Spec; DnsZones_CAA_Spec; etc ...
    property: Etag
    remove: true
    because: it is supposed to be marked ReadOnly in Swagger
```

See the interface [`StringMatcher`](https://github.com/Azure/azure-service-operator/blob/main/v2/tools/generator/internal/config/string_matcher.go) and it's implementations for details.

v2/api/network/v1beta20201101storage/load_balancer_types_gen.go
theunrepentantgeek: Yes it is correct. All `beta` versions come before all `api` versions, so the sequence is this:

... -> `v1beta20201101storage` -> `v1api20180501storage` -> `v1api20200601storage` -> `v1api20201101storage`

The extra code is to handle the conversion required by introducing `v1api20180501storage` into the midst of the conversion graph.

theunrepentantgeek (approved):
Some simplification of `azure-arm.yaml` possible, otherwise looks good.

v2/azure-arm.yaml
super-harsh: oohh.. This is something I did not realise. Thanks, I'll update.

PR #2915, opened 2023-04-20
Refactor authentication and multitenancy documentation some

matthchr:
Closes #2903

**If applicable**:
- [x] this PR contains documentation
- [ ] this PR contains tests

PR #2919, opened 2023-04-24
ARMClientCache should only check the namespace the resource is in

matthchr:
No cross namespace secret references should be allowed as there is a risk for escalation of privilege.

This is a breaking change, but unfortunately a required one from a security perspective.

**If applicable**:
- [x] this PR contains tests


matthchr:
I've also updated my documentation PR to talk about security best practices: #2915

v2/internal/reconcilers/arm/arm_client_cache.go
```diff
@@ -166,13 +175,7 @@ func (c *ARMClientCache) getCredentialFromAnnotation(ctx context.Context, obj ge
 }
 
 func getSecretNameFromAnnotation(credentialFrom string, resourceNamespace string) types.NamespacedName {
```
super-harsh, line 177: Now we can totally get rid of this method and just use `types.NamespacedName{Namespace: resourceNamespace, Name: credentialFrom}` instead?

matthchr: Fixed

super-harsh (approved):

Issue #2914, opened 2023-04-20
Bug: Storage Account resource does not detect drift in networkAcls if the original is empty


gareth-baker:
**Version of Azure Service Operator**
v2.0.0-beta.5

**Describe the bug**
I am trying to create StorageAccount resources that maintain state for the networkAcls on those accounts. If a resource is originally created with `virtualNetworkRules` or `ipRules` configured with at least one entry, drift will be managed properly and any new entries created in the Azure portal will be deleted during reconciliation.

However, if I create a resource without any `virtualNetworkRules` or `ipRules` configured, reconciliation does not seem to be deleting rules created in the portal during reconciliation.

**To Reproduce**

- Create a StorageAccount resource like the following:

```
apiVersion: storage.azure.com/v1beta20210401
kind: StorageAccount
metadata:
  name: iprulesnorules
  namespace: default
spec:
  azureName: iprulesnorules
  kind: StorageV2
  largeFileSharesState: Enabled
  location: canadacentral
  minimumTlsVersion: TLS1_2
  networkAcls:
    bypass: Logging
    defaultAction: Deny
  owner:
    name: gareth-worker
  sku:
    name: Standard_LRS
  supportsHttpsTrafficOnly: true
```
- Add a new IpRule or VirtualNetworkRule through the Azure portal on the Storage Account after it is deployed
- Wait for next reconciliation loop

**Expected behavior**
Additional rules created in the portal should be deleted.

**Additional context**
I'm not sure if this is intentional but wanted to see if there is a particular reason why it is working if there are rules configured on the resource.

matthchr:
No, this is not intentional. It's a combination of a couple things coming together in a way that is bad.

Here's what's causing this:

1. Azure Storage accounts don't quite have PUT == full replace semantics. They have a PATCH-like behavior on PUT which effectively boils down to: They will replace sub-objects/sub-collections if the collection is in the JSON payload, but they will leave untouched properties which were omitted.
2. When you remove `virtualNetworkRules` or `ipRules`, what actually happens on the backend is those properties are `nil` and don't get serialized. This falls afoul of item 1 above because since those properties were not included in the JSON payload storage handles them with PATCH semantics and doesn't clear/remove the contents of the collection.

[...]

PR #2924, opened 2023-04-27
Design for handling PATCH semantics for collections

theunrepentantgeek:

Design to address the issue identified in #2914 

gif

**If applicable**:


docs/hugo/content/design/ADR-2023-04-Patch-Collections.md
```diff
@@ -0,0 +1,115 @@
...
+To illustrate, assume you have a Person object with a Name and a list of Addresses.
+
+* If you PUT a Person object with a new Name and _no addresses_, the new name will be persisted, but the addresses will remain _unchanged_.
+* If you PUT a Person object with a new Name and an _empty list_ of addresses, the new name will be persisted, and the addresses collection will be cleared.
+
+This is a problem for ASO because we use `omitempty` on slice/map properties, which means that if you remove the last item from a slice/map, the property will be entirely omitted from the JSON payload.
+
+In principle, an ASO custom resource (CR) should be the goal-state for the resource and should be a complete definition of the desired state. Whether the collection is empty or missing entirely should not matter, we should interpret that as "clear the collection".
+
+### Option 1: Optional Properties
```
matthchr, line 20: This is actually a problem for more than just slices/maps btw, it can impact even `*string` fields that are `omitempty`. The AKS API is a classic example of this problem.

docs/hugo/content/design/ADR-2023-04-Patch-Collections.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+* PRO: Behaves exactly like we want from a serialization perspective.
+* CON: Horrible to use in code because you must dereference the property to range over it (or really do much of anything with it).
+* CON: Breaking change for any consumers of our object model.
+* CON: Modifies our Spec and Status object definitions for an issue that is specific to ARM
+
+### Option 2: Remove `omitempty` 
+
+Remove omitempty from slice/map properties across all generated types.
```
matthchr, line 43: As above, for impacted RPs we possibly ought to remove `omitempty` from... everything? That has its own set of problems but we should probably talk about collections _and_ non-collections here because the problem can impact them both.

docs/hugo/content/design/ADR-2023-04-Patch-Collections.md
```diff
@@ -0,0 +1,115 @@
...
+``` go
+MyProperty []Type 'json:"myProperty"'. 
+```
+
+This means that the slice or map will always be serialized, even if empty.
+
+* PRO: Behaves exactly like we want from a serialization perspective.
+* PRO: Not a breaking change for consumers of our object model.
+* CON: Empty collections will always be present when an ASO CR is read from the cluster, even if the user did not specify them at all.
+* CON: If we extend this to all properties, all properties will be present when an ASO CR is read from the cluster, even if the user did not specify them at all.
```
matthchr, line 54: Ah, it's called out here but could be more explicit

docs/hugo/content/design/ADR-2023-04-Patch-Collections.md
```diff
@@ -0,0 +1,115 @@
...
+* CON: Modifies our Spec and Status object definitions for an issue that is specific to ARM
+* QUERY: It is possible that some Azure services treat `myProperty: null` as different than omitting the property entirely, although we do not actually know of any cases where that is true.
+
+### Option 3: Custom JSON Serialization
+
+Use a custom JSON serialization library that supports an `omitnil` equivalent, such as [jettison](https://github.com/wI2L/jettison).
```
matthchr, line 60: Does this solve non-collection properties issue? I'm not sure that it does. The only way to clear those properties may be to explicitly pass json `null` to the server for the field. `omitnil` would prevent that. `omitnil` saves us for collections because we can still pass the empty collection.

docs/hugo/content/design/ADR-2023-04-Patch-Collections.md
```diff
@@ -0,0 +1,115 @@
...
+* PRO: Not a breaking change for consumers of our object model.
+* CON: Breaks CRD YAML deserialization for the ASO controller (as the [sigs.k8s.io/yaml]( https://github.com/kubernetes-sigs/yaml) library we use works with standard JSON serialization attributes)
+* CON: Also breaks YAML serialization for `asoctl` (for the same reason)
+* CON: Modifies our Spec and Status object definitions for an issue that is specific to ARM
+
+### Option 4: Remove `omitempty` for selected ARM types only
+
+As for Option 2, above, we apply the change only selected ARM Spec types, leaving the core ASO Spec and Status types unmodified.
+
+Selection occurs through our existing `ObjectModelConfiguration` allowing us to explicitly target collections (and other properties) that we know are problematic.
```
matthchr, line 72: Do we do this at the property level or at the type level (or maybe at the resource level?)

I kinda think we shouldn't do it at the property level because it's _too_ fine-grained? AFAIK, when RPs have this problem they have it for the whole resource (including complex sub-properties, etc). Though that may be up to RPs individual implementations...

Basically wondering if something liek: "You can specify that a whole resource including its whole type hierarchy shouldn't use `omitempty`" and when we run into this for say, `storageAccount`, we just blat that onto the whole resource and call it good rather than try and be overly surgical?

docs/hugo/content/design/ADR-2023-04-Patch-Collections.md
```diff
@@ -0,0 +1,115 @@
...
+
+* PRO: Behaves exactly like we want from a serialization perspective.
+* PRO: Limits the scope of the change to only ARM types. These are not directly used by consumers of our object model.
+* PRO: Limits the scope of the change to only properties that we know are problematic, reducing any potential for unintended consequences.
+* PRO: No change to the YAML serialization for ASO Spec and Status types, so no visible change to our users.
+* CON: We can only configure properties we know have this problem. We may miss some, requiring a new release to fix affected users.
+
+## Decision
+
+Adopt Option 4, as it is a minimal change that addresses the issue without impacting on current users.
```
matthchr, line 82: Generally agree this is the best medium term option. I think some debate on at exactly what level we do the specification (i.e. resource, type, or property)

docs/hugo/content/design/ADR-2023-04-Patch-Collections.md
theunrepentantgeek: I'll update the description of the problem.

docs/hugo/content/design/ADR-2023-04-Patch-Collections.md
theunrepentantgeek: I've added Option 2a to be more explicit.

docs/hugo/content/design/ADR-2023-04-Patch-Collections.md
theunrepentantgeek: I've added that as an outstanding question.

theunrepentantgeek (commented):
Updated to address Matt's feedback.

docs/hugo/content/design/ADR-2023-04-Patch-Collections.md
matthchr: Or maybe we want the ability to do it at either the type or the property level? See my comment on #2916 about this

matthchr (approved):

PR #2926, opened 2023-04-28
Re-record Workspace_BigDataPool test

matthchr:
Prior to re-recording this test failed roughly 1 in 5 runs on my local.

After re-recording, it ran for 50 runs without a failure. I'm not sure exactly what timing issue is being hit here, and definitely this isn't fixing the underlying problem, but it will help our test pass rate so worth doing.

**If applicable**:
- [x] this PR contains tests


theunrepentantgeek (approved):
Finding the underlying issue would be good, but improving our pass rates in the meantime is worthwhile.

PR #2946, opened 2023-05-05
Don't retry on OperationNotAllowed for all RPs

matthchr:
While this operation is retryable for Compute, it is not retryable for all RPs.

**If applicable**:
- [x] this PR contains tests


super-harsh (approved):

v2/api/compute/customizations/vm_extensions.go
```diff
@@ -0,0 +1,37 @@
...
+	log logr.Logger,
+	next extensions.ErrorClassifierFunc) (core.CloudErrorDetails, error) {
+	details, err := next(cloudError)
+	if err != nil {
+		return core.CloudErrorDetails{}, err
+	}
+
+	// It's weird to retry on OperationNotAllowed as it certainly sounds like a fatal error, but
+	// it primarily happens for quota errors on VM/VMSS, which we do want to retry on as the quota may free
+	// up at some point in the future.
```
theunrepentantgeek, line 31: Are there other resources with quota where this might be needed? 

I don't think the requisite things in Batch are resources, but maybe other providers?

theunrepentantgeek (approved):

v2/api/compute/customizations/vm_extensions.go
matthchr: The problem is that the setting of this code is going to be specific to individual RPs. Not every RP will use `OperationNotAllowed` for "no quota". Compute does, so making this check specific to them for now makes sense. In the future if we have specific codes from other RPs that are quota that are reported to us (or that we find) we can expand to have custom error classifiers for those RPs that check for their specific `code`, but it's not universal across all RPs so I think we need to go RP by RP.

PR #2947, opened 2023-05-06
Split ARMClientCache into CredentialProvider and Cache

matthchr:
* A separate CredentialProvider encapsulates logic for looking up the specific credential.
* ARMClientCache now only concerns itself with caching logic for clients. Actual credential lookup is deferred to the   CredentialProvider.

Prior to this change, the assumption was that every credential was always used to produce a GenericARMClient. This enables scenarios where the user for the credential is something other than an ARM client. One example of such a usage is to extract an AAD token to authenticate with the MySQL Dataplane.

**If applicable**:
- [x] this PR contains tests


v2/internal/identity/credential_provider.go
```diff
@@ -0,0 +1,260 @@
...
+		kubeClient:       kubeClient,
+		globalCredential: globalCredential,
+	}
+}
+
+// GetCredential returns the credential used to manage the obj. It performs credential lookup according to the following
+// hierarchy:
+// 1. Per-Resource credential specified as an annotation (serviceoperator.azure.com/credential-from) on the obj
+// 2. Per-Namespace credential provided at the namespace scope ("aso-credential" secret in the obj's namespace)
+// 3. Global credential for the operator.
```
theunrepentantgeek, line 80: Minor tweaking of wording for clarity

```suggestion
// GetCredential returns the credential used to manage the obj. It performs credential lookup according to the 
// following precedence:
// 1. Per-Resource credential specified as an annotation (`serviceoperator.azure.com/credential-from`) directly on the resource
// 2. Per-Namespace credential provided at the namespace scope (a secret called `aso-credential` in the resource's namespace)
// 3. Global credential for the operator.
```

I'm assuming annotations can't be inherited, so they have to be directly on the resource itself - is that right?

v2/internal/identity/credential_provider.go
```diff
@@ -0,0 +1,260 @@
...
+//
+// If no matching credential can be found, an error is returned.
+func (c *credentialProvider) GetCredential(ctx context.Context, obj genruntime.MetaObject) (*Credential, error) {
+	cred, err := c.getCredentialFromAnnotation(ctx, obj, reconcilers.PerResourceSecretAnnotation)
```
theunrepentantgeek, line 84:
```suggestion
        // Resource annotation
	cred, err := c.getCredentialFromAnnotation(ctx, obj, reconcilers.PerResourceSecretAnnotation)
```

v2/internal/identity/credential_provider.go
```diff
@@ -0,0 +1,260 @@
...
+	// Namespaced secret
+	cred, err = c.getCredentialFromNamespaceSecret(ctx, obj.GetNamespace())
+	if err != nil {
+		return nil, err
+	}
+	if cred != nil {
+		return cred, nil
+	}
+
+	if c.globalCredential == nil {
```
theunrepentantgeek, line 101:
```suggestion
        // Global credential
	if c.globalCredential == nil {
```

v2/internal/identity/credential_provider.go
```diff
@@ -0,0 +1,260 @@
...
+		return nil, errors.New("global credential not configured, you must use either namespaced or per-resource credentials")
+	}
+	// If not found, default is the global credential
```
theunrepentantgeek, line 104: Add blank line prior

v2/internal/identity/credential_provider.go
```diff
@@ -0,0 +1,260 @@
...
+	return c.globalCredential, nil
+}
+
+func (c *credentialProvider) getCredentialFromNamespaceSecret(ctx context.Context, namespace string) (*Credential, error) {
```
theunrepentantgeek, line 108: Suggest adding a doc comment explaining what this method does

v2/internal/identity/credential_provider.go
```diff
@@ -0,0 +1,260 @@
...
+	secretNamespacedName := types.NamespacedName{Namespace: namespace, Name: NamespacedSecretName}
+	cred, err := c.getCredentialFromSecret(ctx, secretNamespacedName)
+	if err != nil {
+		var secretNotFound *core.SecretNotFound
+		if errors.As(err, &secretNotFound) {
+			return nil, nil // Not finding this secret is allowed, allow caller to proceed to higher scope secret
+		}
+		if apierrors.IsNotFound(err) {
+
+		}
```
theunrepentantgeek, line 118: Empty body for this if is weird.

v2/internal/identity/credential_provider.go
```diff
@@ -0,0 +1,260 @@
...
+		return nil, err
+	}
+
+	return cred, nil
+}
+
+func (c *credentialProvider) getCredentialFromAnnotation(ctx context.Context, obj genruntime.MetaObject, annotation string) (*Credential, error) {
```
theunrepentantgeek, line 125: Add doc comment

v2/internal/identity/credential_provider.go
```diff
@@ -0,0 +1,260 @@
...
+	credentialFrom, ok := obj.GetAnnotations()[annotation]
+	if !ok {
+		return nil, nil
+	}
+
+	// Disallow credentials with `/` in their credentialFrom
+	if strings.Contains(credentialFrom, "/") {
```
theunrepentantgeek, line 132: Are there any potential encodings we need to be aware of here?

E.g. we're blocking `/` but not `%2F`

v2/internal/identity/credential_provider.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+func (c *credentialProvider) newCredentialFromSecret(secret *v1.Secret) (*Credential, error) {
+	var errs []error
+
+	nsName := types.NamespacedName{Namespace: secret.GetNamespace(), Name: secret.GetName()}
+
+	subscriptionID, ok := secret.Data[config.SubscriptionIDVar]
+	if !ok {
+		err := core.NewSecretNotFoundError(nsName, errors.Errorf("credential Secret %q does not contain key %q", nsName.String(), config.SubscriptionIDVar))
```
theunrepentantgeek, line 169: I thought you could omit `.String()` when formatting, because the `fmt.Stringer` interface would be automatically invoked by `%q`.

(Applies to other formatting strings too).

v2/internal/identity/credential_provider.go
```diff
@@ -0,0 +1,260 @@
...
+		errs = append(errs, err)
+	}
+
+	tenantID, ok := secret.Data[config.TenantIDVar]
+	if !ok {
+		err := core.NewSecretNotFoundError(nsName, errors.Errorf("credential Secret %q does not contain key %q", nsName.String(), config.TenantIDVar))
```
theunrepentantgeek, line 175: Suggest formatting for readability

```suggestion
		err := core.NewSecretNotFoundError(
		    nsName, 
		    errors.Errorf(
		        "credential Secret %q does not contain key %q", 
    		        nsName.String(), 
    		        config.TenantIDVar))
```

v2/internal/reconcilers/arm/arm_client_cache_test.go
```diff
@@ -30,33 +30,37 @@ import (
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/core"
 )
 
-const (
-	fakeID = "00000000-0000-0000-0000-000000000000"
-)
+const testPodNamespace = "azureserviceoperator-system-test"
+const testSubscriptionID = "00000011-1111-0011-1100-110000000000"
```
theunrepentantgeek, line 34: Why this particular constant? It seems very specific.

v2/internal/identity/credential_provider.go
```diff
@@ -0,0 +1,260 @@
...
+	globalCredentialSecretName = "aso-controller-settings"
+	// #nosec
+	NamespacedSecretName = "aso-credential"
+)
+
+// Credential describes a credential used to connect to Azure
+type Credential struct {
+	TokenCredential azcore.TokenCredential
+	CredentialFrom  types.NamespacedName
+	SubscriptionID  string
```
theunrepentantgeek, line 39: Should these be publicly visible (and thus mutable), or should you create some getter methods to safely expose them?

v2/internal/reconcilers/arm/arm_client.go
```diff
@@ -51,7 +47,7 @@ type Connection struct {
 func newConnection(client *armClient) *Connection {
```
theunrepentantgeek, line 47: Instead of this being a bare function that accepts a client, consider making it a method on `armClient`.

```suggestion
func (client *armClient) newConnection() *Connection {
```

theunrepentantgeek (approved):

v2/internal/identity/credential_provider.go
super-harsh: I think it should be removed, as we're checking the `secretNotFound` above

v2/internal/reconcilers/arm/arm_client.go
super-harsh, line 47: This method is called inside `armreconciler.ARMClientFactory` and returns a `Connection` instead of ARMClient. Maybe we need to fix some naming around here for `Connection`.

v2/internal/identity/credential_provider.go
matthchr: Updated the docs here according to your suggestions.

v2/internal/identity/credential_provider.go
matthchr: Done

v2/internal/identity/credential_provider.go
matthchr: Done

v2/internal/identity/credential_provider.go
matthchr: Done

v2/internal/identity/credential_provider.go
matthchr: @super-harsh is correct here, I had meant to delete this. The actual handling for NotFound is done in the `errors.As` above.

matthchr: Fixed

v2/internal/identity/credential_provider.go
matthchr: Done

v2/internal/identity/credential_provider.go
matthchr: No, in fact this `if` isn't really required to be "safe" - it's just for a more clear user error. If the user passed us a secret with name: `my/secret`, we would lookup a secret with: `types.NamespacedName{Namespace: obj.Namespace, Name: "my/secret"}`. That lookup will fail because no such secret exists.

v2/internal/identity/credential_provider.go
matthchr: Fixed

v2/internal/identity/credential_provider.go
matthchr: Done

v2/internal/reconcilers/arm/arm_client_cache_test.go
matthchr: Technically any hardcoded GUID is _quite_ specific. There's not real reason for this specific GUID. I just needed a GUID that wasn't all zeroes. I've left a comment `// Arbitrary GUID that isn't all 0s` here to help.

v2/internal/identity/credential_provider.go
matthchr: Moved to methods

v2/internal/reconcilers/arm/arm_client.go
matthchr: Done

matthchr: I see what you mean.

I've changed a few things here:

1. `Connection` is now an interface which `armClient` implements, so there's not two separate types.
2. The method is `GetConnection` now rather than `GetClient`, and it returns a `Connection`. This seemed clearer than `GetClient` as there is already such a thing as a client (i.e. `GenericARMClient`) and this was "bigger" than that as it also has identity/credential details on it.

PR #2958, opened 2023-05-10
Assert the PerResourceSecret annotation is not set

matthchr:
It should not be set in the global secret case. This test just ensures that.

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/test/single_operator_multitenant_test.go
+++ b/v2/test/single_operator_multitenant_test.go
@@ -9,7 +9,6 @@ import (
 	"testing"
 
 	"github.com/pkg/errors"
-	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


super-harsh (approved):

v2/test/single_operator_multitenant_test.go
```diff
@@ -127,6 +127,7 @@ func Test_Multitenant_SingleOperator_PerResourceCredential(t *testing.T) {
 
 	objKey := client.ObjectKeyFromObject(acct)
 	tc.GetResource(objKey, acct)
+	tc.Expect(acct.Annotations).ToNot(HaveKey(reconcilers.PerResourceSecretAnnotation))
```
super-harsh, line 130: oh wait! on line 116 above, the annotation has `nsName.String()` as a secret name which should get rejected as `.String()` method would add a separator `/`?

matthchr: Ahh good catch, will fix.

PR #2960, opened 2023-05-11
Add support for FileService, TableService, FileShare and Table

### LINKED ISSUE #2824: Support configuring Microsoft.Storage/storageAccounts/fileservices and tableServices
While playing with StorageAccounts in the `apiextensions.k8s.io_v1_customresourcedefinition_storageaccounts.storage.azure.com.yaml` contract we reaslized that there is no way to specify a fileShare soft delete policy on a storage account forcing us to go directly and call an AZ API endpoint to change that value.
This is the missing corresponding AZ property
```
 {
            "name": "[concat(parameters('storageAccountName'), '/default')]",
            "type": "Microsoft.Storage/storageAccounts/fileservices",
            "apiVersion": "2022-05-01",
            "properties": {
                "shareDeleteRetentionPolicy": {
                    "enabled": "[parameters('isShareSoftDeleteEnabled')]"
                }
            },
            "dependsOn": [
                "[concat('Microsoft.Storage/storageAccounts/', parameters('storageAccountName'))]",
                "[concat(concat('Microsoft.Storage/storageAccounts/', parameters('storageAccountName')), '/blobServices/default')]"
            ]
        }
```

Is it possible to somehow extend the schema of a givenCRD so when the operator creates the resource it takes the new property into account?
While searching through the schemas in `apiextensions.k8s.io_v1_customresourcedefinition_storageaccounts.*` namespace almost every one of the them defines `additionalProperties` all over the place? How can one use those properties? Any code examples?

super-harsh:
Closes #2824 

This PR adds Fileservice, TableService, FileShare, and Table resources and some lift-shift around storageAccount samples. 

**If applicable**:
- [x] this PR contains tests


v2/api/storage/v1api20220901storage/structure.txt
```diff
@@ -0,0 +1,187 @@
...
+│   │   │   │   └── PropertyBag: genruntime.PropertyBag
+│   │   │   └── PropertyBag: genruntime.PropertyBag
+│   │   ├── OriginalVersion: string
+│   │   ├── Owner: *genruntime.KnownResourceReference
+│   │   ├── PropertyBag: genruntime.PropertyBag
+│   │   ├── ProtocolSettings: *Object (2 properties)
+│   │   │   ├── PropertyBag: genruntime.PropertyBag
+│   │   │   └── Smb: *Object (6 properties)
+│   │   │       ├── AuthenticationMethods: *string
+│   │   │       ├── ChannelEncryption: *string
```
super-harsh, line 24: Not sure if this one should be a secretRef?

v2/api/storage/v1api20220901storage/structure.txt
```diff
@@ -0,0 +1,187 @@
...
+│   │   │       ├── KerberosTicketEncryption: *string
```
super-harsh, line 25: Not sure if this one should be a secretRef?

v2/api/storage/v1api20220901storage/structure.txt
matthchr: Not sure either -- look up an example where it's used?

v2/azure-arm.yaml
```diff
@@ -1588,6 +1588,25 @@ objectModelConfiguration:
 #        $exportAs: ServersDatabasesWorkloadGroupsWorkloadClassifier
 #        $supportedFrom: v2.0.0
   storage:
+    2022-09-01:
+      StorageAccounts_TableService:
+        $exportAs: StorageAccountsTableService
```
matthchr, line 1593: Normally I am all in favor of using the latest API version, especially when adding a new resource or a totally new RP. In this case though we already have lots of this RPs peer items but at a lower API version and now we're going to have a mismatch in consistency.

2021-04-01 StorageAccounts_BlobService, but 2022-09-01 `StorageAccounts_FileService`.

I wonder if we should either add support for BlobService/FileService at the old version rather than this new verison, OR add support for this new version for the existing types so that one version has all of the types.

That would have the nice benefit of updating storage to support a newer API which does seem possibly desirable?

super-harsh: In that case, I personally would prefer to add an older matching resource version for the new resources. That said, users running the older version of parent/peer resources who want to implement the new resource, would then still have to install the newer version which would not be consistent and would be the same scenario as we are in right now or they'll have to migrate the resources to be on the same API version to be consistent.

matthchr: > In that case, I personally would prefer to add an older matching resource version for the new resources. That said, users running the older version of parent/peer resources who want to implement the new resource, would then still have to install the new resource with a newer version which would not be consistent and would be the same scenario as we are in right now or they'll have to migrate the resources to be on the same API version to be consistent.

Not sure I follow this.

If you added for `2022-09-01` all of the existing StorageAccount stuff too, then we would have exactly the same experience for people who are using the old `StorageAccount` and want `TableService`, with the added benefit that we also support the new StorageAccount with its new features?

PR #2963, opened 2023-05-11
Allow saving imported resources to separate YAML files

### LINKED ISSUE #2870: Add option to export each resource to a separate file

Currently `asoctl import azure-resource` will create a single output YAML file containing all the resources. 

**Describe the improvement**

Some users may want each resource to end up in a separate file.

theunrepentantgeek:

Writing all imported resources to a single YAML file doesn't work for some use cases; writing to separate files has been requested by some users so they have better visibility.

Closes #2870 

gif


v2/cmd/asoctl/cmd/import_azure_resource.go
```diff
@@ -75,26 +83,62 @@ func importAzureResource(ctx context.Context, armIDs []string, outputPath *strin
...
 		}
-	} else {
+	} else if folder, ok := options.writeToFolder(); ok {
 		log.Info(
-			"Writing to file",
-			"path", *outputPath)
-		err := result.SaveToFile(*outputPath)
+			"Writing to individual files in folder",
+			"folder", folder)
+		err := result.SaveToFolder(folder)
```
matthchr, line 111: minor: pedant in me but one of these save methods is `SaveAllToFile` and one is `SaveToFolder`. Are we not saving `All` to folder?

Feels like they should mirror one another and either both have `All` or neither have `All`.

v2/cmd/asoctl/internal/importing/resource_import_result.go
```diff
@@ -22,7 +24,71 @@ type ResourceImportResult struct {
...
+
+func (r *ResourceImportResult) SaveToFolder(folder string) error {
+	// We name the files after the resource type and name
+	// We allocate resources to files using a map, just in case we have a naming collision
+	// (If that happens, all the similarly named resources will be in the same file, which is not ideal,
+	// but better than dropping one or more)
+	fileMap := make(map[string][]genruntime.MetaObject, len(r.resources))
+	for _, resource := range r.resources {
+		resourceName := resource.GetName()
+		typeName := resource.GetObjectKind().GroupVersionKind().Kind
```
matthchr, line 48: You could include the group in here too, in a `group-kind-name`, which would guarantee uniqueness and avoid the awkward possibility of collisions entirely?

matthchr: Presumably if you did this, the map could just be `map[string]genruntime.MetaObject` I guess?

matthchr (approved):
Approved with some comments

v2/cmd/asoctl/cmd/import_azure_resource.go
theunrepentantgeek: I was trying to reflect the differing semantics in the names - one creates a single file, the other creates _n_ files. I'll rename the methods.

v2/cmd/asoctl/internal/importing/resource_import_result.go
theunrepentantgeek: I think we'd still need a `map[string][]genruntime.MetaObject` because we can (in theory) follow ARM URL references into different subscriptions. I'd rather gather all the resources (and force the user to resolve the conflicts in the yaml files) than non-deterministically drop one of them.

v2/cmd/asoctl/cmd/import_azure_resource.go
theunrepentantgeek: Renamed to `SaveToSingleFile()` and `SaveToIndividualFilesInFolder()` which are a bit unwieldy, but clear.

PR #2972, opened 2023-05-12
Add support for MySQL FlexibleServer AAD Users

### LINKED ISSUE #1401: Bug: Cannot log into MySQL as MySQLAADUser representing a group
**Describe the bug**
MySQLAADUser representing an AAD Group cannot be logged in as on the MySQL server.

**To Reproduce**
1. Create a `MySQLAADUser`  for a group. Specify the group name as the `Username` and the group object ID as the `AADID`.
2. Observe that the user reconciles successfully.
3. Attempt to log in to the user in MySQL, following [the instructions](https://docs.microsoft.com/en-us/azure/mysql/howto-configure-sign-in-azure-ad-authentication#creating-azure-ad-groups-in-azure-database-for-mysql). Specifically: "When logging in, members of the group will use their personal access tokens, but sign with the group name specified as the username."

**Expected behavior**
I am able to log in as the user.

### LINKED ISSUE #1402: Clarify expectations with MySQLAADUser and long user/group names
We currently do not support creating aliases for users or groups on the `MySQLAADUser` CRD. According to [the documentation](https://docs.microsoft.com/en-us/azure/mysql/howto-configure-sign-in-azure-ad-authentication#creating-azure-ad-users-in-azure-database-for-mysql)
> For user names that exceed 32 characters, it is recommended you use an alias instead, to be used when connecting

Even though it says "recommended" we've had customer reports of errors being raised when attempting to register names longer than 32 characters.

**Describe the improvement**
We need to either confirm that >32 character users and groups work, or we need to add support for aliasing so that they can be aliased to a <=32 character MySQL username.

**Additional context**
We should probably add a test for this.

matthchr:
Includes support for creating AAD users.

In order to manage AAD users, the identity the operator is running as (either the global identity or the identity specified in the users credential-from annotation) must be the AAD Administrator of the MySQL server. This requires that ASO is using a managed identity of some kind. 
We recommend workload identity.

Note that the added test is skipped right now as it cannot run in our CI subscription, but I have run it in a subscription that does work and it passes. We will work towards getting it to run in the CI sub.

Closes #1401 
Closes #1402

**If applicable**:
- [x] this PR contains tests


v2/internal/controllers/crd_dbformysql_flexibleserver_test.go
```diff
@@ -205,7 +205,7 @@ func MySQLFlexibleServer_AADAdmin_CRUD(tc *testcommon.KubePerTestContext, rg *re
 		Spec: mysqlpreview.FlexibleServers_Administrator_Spec{
 			Owner:             testcommon.AsOwner(server),
 			AdministratorType: &aadAdmin,
-			Login:             &mi.Name,
+			Login:             &mi.Name, // TODO: Should we be able to source this from a configmap?
```
matthchr, line 208: Remove this TODO, I think for now it doesn't make sense to source from configMap. We can revisit if user asks

v2/api/dbformysql/v1/user_types.go
```diff
@@ -150,6 +152,104 @@ func (user *User) deleteValidations() []func() error {
...
+	}
+
+	// If we don't have a finalizer yet, it's OK to change things
+	hasFinalizer := controllerutil.ContainsFinalizer(oldUser, genruntime.ReconcilerFinalizer)
+	if !hasFinalizer {
+		return nil
+	}
+
+	if oldUser.Spec.AzureName != user.Spec.AzureName {
+		errs = append(errs, errors.Errorf("updating 'AzureName' is not allowed for '%s : %s", oldObj.GetObjectKind().GroupVersionKind(), oldUser.GetName()))
```
theunrepentantgeek, line 183: Would it be clearer to separate creation of the error from the append? 
And maybe to use one parameter per line to increase visibility of the parameters used?

```suggestion
		err := errors.Errorf(
			"updating 'AzureName' is not allowed for '%s : %s", 
			oldObj.GetObjectKind().GroupVersionKind(), 
			oldUser.GetName())
		errs = append(errs, err)
```

v2/api/dbformysql/v1/user_types.go
```diff
@@ -150,6 +152,104 @@ func (user *User) deleteValidations() []func() error {
...
+	// Ensure that owner has not been changed
+	oldOwner := oldUser.Owner()
+	newOwner := user.Owner()
+
+	bothHaveOwner := oldOwner != nil && newOwner != nil
+	ownerAdded := oldOwner == nil && newOwner != nil
+	ownerRemoved := oldOwner != nil && newOwner == nil
+
+	if (bothHaveOwner && oldOwner.Name != newOwner.Name) || ownerAdded {
+		errs = append(errs, errors.Errorf("updating 'Owner.Name' is not allowed for '%s : %s", oldObj.GetObjectKind().GroupVersionKind(), oldUser.GetName()))
```
theunrepentantgeek, line 195: For illustration, suggesting the same change here:

```suggestion
		err := errors.Errorf(
			"updating 'Owner.Name' is not allowed for '%s : %s", 
			oldObj.GetObjectKind().GroupVersionKind(), 
			oldUser.GetName())
		errs = append(errs, err)
```

v2/api/dbformysql/v1/user_types.go
```diff
@@ -166,14 +266,20 @@ type UserList struct {
...
-	//AzureName: The name of the resource in Azure. This is often the same as the name of the resource in Kubernetes but it
-	//doesn't have to be.
+	// AzureName: The name of the resource in Azure. This is often the same as the name of the resource in Kubernetes but it
+	// doesn't have to be.
+	// If not specified, the default is the name of the Kubernetes object.
+	// When creating a local user, this will be the name of the user created.
+	// When creating an AAD user, this must have a specific format depending on the type of AAD user being created.
+	// For managed identity: "my-managed-identity-name"
+	// For standard AAD user: "myuser@mydomain.onmicrosoft.com"
+	// For AAD group: "my-group"
```
theunrepentantgeek, line 276: How does this syntax discriminate between managed-identity and AAD-group?

v2/api/dbformysql/v1/user_types.go
```diff
@@ -166,14 +266,20 @@ type UserList struct {
...
 	AzureName string `json:"azureName,omitempty"`
 
 	// +kubebuilder:validation:Required
-	//Owner: The owner of the resource. The owner controls where the resource goes when it is deployed. The owner also
-	//controls the resources lifecycle. When the owner is deleted the resource will also be deleted. Owner is expected to be a
-	//reference to a dbformysql.azure.com/FlexibleServer resource
+	// Owner: The owner of the resource. The owner controls where the resource goes when it is deployed. The owner also
+	// controls the resources lifecycle. When the owner is deleted the resource will also be deleted. Owner is expected to be a
+	// reference to a dbformysql.azure.com/FlexibleServer resource
```
theunrepentantgeek, line 282: Has `controller-gen` fixed the bug where leading spaces on comments cause issues?

v2/internal/reconcilers/mysql/aad_user.go
```diff
@@ -0,0 +1,149 @@
...
+	db, err := u.connectToDB(ctx)
+	if err != nil {
+		return err
+	}
+	defer db.Close()
+
+	// TODO: There's still probably some ways that this user can be deleted but that we don't detect (and
+	// TODO: so might cause an error triggering the resource to get stuck).
+	// TODO: Cases where the server is in the process of being
+	// TODO: deleted (or all system tables have been wiped?) might also exist...
```
theunrepentantgeek, line 74: Does this need to be addressed before merging?

v2/internal/reconcilers/mysql/connector.go
```diff
@@ -0,0 +1,48 @@
...
+	// Get the owner - at this point it must exist
+	owner, err := resourceResolver.ResolveOwner(ctx, user)
+	if err != nil {
+		return "", err
+	}
+
+	flexibleServer, ok := owner.(*dbformysql.FlexibleServer)
+	if !ok {
+		return "", errors.Errorf("owner was not type FlexibleServer, instead: %T", owner)
+	}
```
theunrepentantgeek, line 36: Missing blank line. #pedant

v2/internal/reconcilers/mysql/local_user.go
```diff
@@ -0,0 +1,157 @@
...
+		return errors.Wrap(err, "failed to look up .spec.localUser.Password")
+	}
+
+	// Create or update the user. Note that this updates password if it has changed
+	username := u.user.Spec.AzureName
+	err = mysqlutil.CreateOrUpdateUser(ctx, db, username, u.user.Spec.Hostname, password)
+	if err != nil {
+		return errors.Wrap(err, "failed to create user")
+	}
+	// Ensure that the privileges are set
```
theunrepentantgeek, line 59: Missing blank line after **if**. #pedant

v2/internal/util/mysql/mysql.go
```diff
@@ -88,9 +73,30 @@ func CreateOrUpdateUser(ctx context.Context, db *sql.DB, username string, hostna
 	return nil
 }
 
+// See https://learn.microsoft.com/en-us/azure/mysql/flexible-server/how-to-azure-ad for details on how to create AAD
+// users.
+// username can be either the actual AAD user name (for real AAD users)
+// TODO: ??? for managed identities
```
theunrepentantgeek, line 79: Does this need to be resolved before merge?

v2/internal/util/mysql/mysql.go
```diff
@@ -88,9 +73,30 @@ func CreateOrUpdateUser(ctx context.Context, db *sql.DB, username string, hostna
...
+// or the group name for groups
+func CreateOrUpdateAADUser(ctx context.Context, db *sql.DB, username string, alias string) error {
+	var err error
+	if alias != "" {
+		statement := "CREATE AADUSER IF NOT EXISTS ? AS ?"
+		_, err = db.ExecContext(ctx, statement, username, alias)
```
theunrepentantgeek, line 85: What's being ignored here?

v2/internal/util/mysql/mysql.go
```diff
@@ -111,3 +117,37 @@ func DropUser(ctx context.Context, db *sql.DB, username string) error {
 	_, err := db.ExecContext(ctx, "DROP USER IF EXISTS ?", username)
 	return err
 }
+
+func newConfig(serverAddress string, database string, port int, user string, password string) *mysql.Config {
```
theunrepentantgeek, line 121: Maybe this needs a more specific name to avoid conflating this `config` with other `config` elsewhere?

```suggestion
func newMySQLConfig(
        serverAddress string, 
        database string, 
        port int, 
        user string, 
        password string) *mysql.Config {
```

v2/samples/dbformysql/v1api/v1_user.yaml
```diff
@@ -21,6 +21,8 @@ spec:
       - SELECT
   localUser:
     serverAdminUsername: admin
+    # serverAdminPassword is optional. If NOT specified, the operator uses its identity to log into the server.
+    # The operator can only successfully log into the server if its identity is the Administrator of the server.
```
theunrepentantgeek, line 25: Can the Administrator of the server be a group?

v2/samples/dbformysql/v1api/v1_user_aad.yaml
```diff
@@ -0,0 +1,31 @@
...
+  # REPLICATION SLAVE, REPLICATION CLIENT, CREATE USER
+  privileges:
+    - PROCESS
+    - CREATE USER
+  databasePrivileges:
+    mysqldatabase-sample:
+      # Privileges include the following:
+      # SELECT, INSERT, UPDATE, DELETE, CREATE, DROP, REFERENCES, INDEX,
+      # ALTER, CREATE TEMPORARY TABLES, LOCK TABLES, EXECUTE, CREATE VIEW,
+      # SHOW VIEW, CREATE ROUTINE, ALTER ROUTINE, EVENT, TRIGGER
```
theunrepentantgeek, line 20: Saying "include" implies there are other privileges that might be used. Is this canonical or just representative?

v2/test/mysql_test.go
```diff
@@ -277,7 +277,49 @@ func MySQL_User_CRUD(tc *testcommon.KubePerTestContext, server *mysql.FlexibleSe
...
+	// Confirm that we cannot change the user type from local to AAD
+	user = originalUser.DeepCopy()
+	old = user.DeepCopy()
+	user.Spec.LocalUser = nil
+	user.Spec.AADUser = &mysqlv1.AADUserSpec{
+		ServerAdminUsername: "someadminuser",
+	}
+	err = tc.PatchAndExpectError(old, user)
+	tc.Expect(err).To(HaveOccurred())
+	tc.Expect(err.Error()).To(ContainSubstring("cannot change from local user to AAD user"))
```
theunrepentantgeek, line 306: Do we need the reverse test (cannot change from AAD user to local user) too?

theunrepentantgeek (approved):

v2/api/dbformysql/v1/user_types.go
```diff
@@ -150,6 +152,104 @@ func (user *User) deleteValidations() []func() error {
...
+		},
+		user.validateUserTypeNotChanged,
+		user.validateWriteOnePropertiesNotChanged,
+		user.validateUserAADAliasNotChanged,
+	}
+}
+
+// validateWriteOnePropertiesNotChanged function validates the update on WriteOnce properties.
+// TODO: Note this should be kept in sync with admissions.ValidateWriteOnceProperties
+func (user *User) validateWriteOnePropertiesNotChanged(oldObj runtime.Object) error {
```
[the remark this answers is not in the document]
matthchr: Fixed

v2/api/dbformysql/v1/user_types.go
matthchr: Done

v2/api/dbformysql/v1/user_types.go
matthchr: Done

v2/api/dbformysql/v1/user_types.go
matthchr: As far as I can tell, it doesn't. 

I just confirmed that it is possible to create a UMI and a group with the same name. I've reached out to our MySQL contacts to ask this question to them, because [the public documentation](https://learn.microsoft.com/en-us/azure/mysql/flexible-server/how-to-azure-ad#create-azure-ad-groups-in-azure-database-for-mysql) doesn't really cover it.

v2/api/dbformysql/v1/user_types.go
matthchr: All of our existing code-generated resources have spaces after the comment:
```
	// MinimumTlsVersion: Optional: requires clients to use a specified TLS version (or higher) to connect (e,g, '1.0', '1.1',
	// '1.2')
	MinimumTlsVersion *RedisCreateProperties_MinimumTlsVersion `json:"minimumTlsVersion,omitempty"`
```

That's a snippet from `redis_types_gen.go`.

The actual documentation in the CRD.yaml for this property is:
```
              owner:
                description: 'Owner: The owner of the resource. The owner controls
                  where the resource goes when it is deployed. The owner also controls
                  the resources lifecycle. When the owner is deleted the resource
                  will also be deleted. Owner is expected to be a reference to a dbformysql.azure.com/FlexibleServer
                  resource'
```

Which looks correct to me?

So I think the answer must be yes?

v2/internal/reconcilers/mysql/aad_user.go
matthchr: No, we had this comment even on the existing MySQL local user support. This is just saying that, we're not confident that the deletion logic here is perfect. Unfortunately actually _finding_ the edge cases where things get stuck is very tricky. It passes in all of our tests, but could get stuck for some time during something like a SQL failover.

I think we just need to keep our eyes open for user reports that suggest an issue here and fix them as they arise.

v2/internal/reconcilers/mysql/connector.go
matthchr: Fixed

v2/internal/util/mysql/mysql.go
matthchr: I don't think so? It's privately scoped to the `mysql` package so it won't cause any issues for other packages, and AFAIK Go's recommendation is that you not prefix the package name on the type names in the package (at least, Goland gives a warning squiggly if you do that).

v2/test/mysql_test.go
matthchr: Not a bad idea, although it's tricky because right now we can't actually successfully create an AAD user in our CI subscription. I've added this to the AAD user test but it's skipped right now like all of the other bits requiring the ability to create an AAD user in our CI sub.

v2/samples/dbformysql/v1api/v1_user_aad.yaml
matthchr: It's representative. The full set is here I think: https://dev.mysql.com/doc/refman/5.7/en/privileges-provided.html

It depends on the version of MySQL, and may also be expandable through MySQL addons/plugins (I'm not enough of an expert in the space to know).  We don't do validation at the operator level on these, we let MySQL perform validation. Being exhaustive here is difficult, but these are the common ones.

v2/samples/dbformysql/v1api/v1_user.yaml
matthchr: Yes! I've added some clarifying documentation about how exactly to log in when the admin is a member of the admin group.

Specifically:
```
    # If the administrator is a group, the ServerAdminUsername should be the group name, not the actual username of the
    # identity to log in with. For example if the administrator group is "admin-group" and identity "my-identity" is
    # a member of that group, the ServerAdminUsername should be "admin-group"
```

That same documentation now exists in the CRD types too (Local and AAD flavors)

Issue #2920, opened 2023-04-25
kube-apiserver persistently high memory usage with large number of CRDs


matthchr:
kube-apiserver uses a surprisingly large amount of memory when CRDs are added. 

ASO v2.0.0 installs ~125 CRDs. Testing quickly on a local `kind` cluster (version 1.26.3) shows that resident memory grows from:
304mb (empty cluster) -> 380mb (cert-manager installed) -> **2.5g** (aso just installed) -> **2.0g** (steady state).

This means that each of our 125 CRDs causes ~13mb of memory usage in kube-apiserver.

## Customer impact
It's very easy to cross managed cluster `kube-apiserver` memory limits. If this happens, kube-apiserver may get OOMKilled by the managed cluster provider (such as AKS). This has negative downstream effects on pretty much everything in the cluster, including triggering pod restarts due to expired watches, failed kube api requests, etc.

## Quick instructions for profiling `kind` apiserver
1. Start `kind`.
2. Install `aso`
3. `kubectl proxy --port=8080 &`
4. `go tool pprof -png "http://localhost:8080/debug/pprof/heap" > out.png` - The actual trailing path can be anything: `/debug/pprof/profile`, `debug/pprof/heap`, etc
5. Or alternatively you can do `curl localhost:8080/debug/pprof/heap > out.pprof` and then analyze it with `go tool pprof out.pprof`

## Experiments
- [ ] Is memory usage reduced if CRDs have no documentation included?
- [ ] Is memory usage reduced if `v1beta` versions are removed?
- [ ] Examine in detail w/ pprof where `kube-apiserver` memory usage is coming from

## Prior art
- https://blog.upbound.io/scaling-kubernetes-to-thousands-of-crds/
- [Summary of crossplane CRD scaling issues](https://github.com/crossplane/crossplane/issues/2895)
- [Great Crossplane summary of kube-apiserver memory usage](https://github.com/crossplane/crossplane/blob/master/design/one-pager-crd-scaling.md#api-server-resource-consumption)

Some key snippets from Crossplane's summary:

[...]

Issue #4, opened 2019-07-23
EventHub Service Operator


frodopwns:
@todo: link to operator spec, add description

PR #2973, opened 2023-05-12
Create ADR for installing only selected CRDs

matthchr:
Related to #2920.

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
+++ b/docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
@@ -1,0 +1,270 @@
+---
+title: '2023-05: Installing only selected CRDs'
+---
+
+## Context
+
+Outstanding issues:
+1. [#2920: kube-apiserver persistently high memory usage with large number of CRDs](https://github.com/Azure/azure-service-operator/issues/2920).
+2. [#1433: Allow users to control which CRDs are installed](https://github.com/Azure/azure-service-operator/issues/1433)
+
+CRDs consume kube-apiserver resources even if no instances of the CRDs are ever created. For more details, see 
+[#2920](https://github.com/Azure/azure-service-operator/issues/2920).
+
+
+We have had a longstanding item ([#1433](https://github.com/Azure/azure-service-operator/issues/1433)) to allow users 
+to be more selective about what CRDs to install. As the number of CRDs installed by ASO increases, 
+this becomes more critical in order to mitigate the kube-apiserver memory usage problem.
+
+### Requirements
+
+Before getting into different options for solving this problem we should go over the requirements.
+
+1. Reduction in default number of CRDs installed.
+    * The goal is that the default ASO configuration should fit into a minimal memory environment
+      (such as a small `kind` cluster or a `Free` tier AKS cluster).
+2. Seamless upgrade experience for existing users.
+    * If the user has installed a previous version of ASO and upgrade to a new version, whatever 
+      resources they had installed should continue to function. The previous default was "everything".
+3. Seamless upgrade experience going forward.
+    * If a user selects a subset of CRDs when initially installing ASO, they should get that exact 
+      subset automatically during upgrade unless they say otherwise.
+    * No "oops I forgot to select any CRDs, so none were installed"
+    * No "oops I forgot to filter CRDs, so they were all installed"
+    * Exactly the CRDs the user originally specified should be upgraded, and only those CRDs.
+4. No CRDs uninstalled ever.
+    * CRD uninstallation is dangerous. If users want to remove CRDs they can do it manually. 
+    * TODO: A reduction in the list of CRDs to install should result in only the new list CRDs of CRDs getting new versions?
+      Or maybe it needs to actually upgrade those versions too even though they weren't specified, to maintain a consistent
+      storage version?
+
+### Question 1: What to install by default?
+
+As mentioned above, a key goal is a reduction in the default number of CRDs installed. That means we somehow need to 
+choose which CRDs we will install and which CRDs we will not.
+
+#### Answer 1a: Handpicked default
+
+We handpick a set of resources we feel are "critical" and install just those by default
+
+Pros:
+* Has at least a chance of working for some user-scenarios out of the box.
+* Ensures that resources which are extremely likely to be needed (such as Microsoft.Resources/resourceGroup) are included 
+  by default.
+
+Cons:
+* Arbitrary.
+* May result in higher kube-apiserver memory consumption than is strictly needed for users as it is likely to include some
+  resources which they may not use.
+
+#### Answer 1b: Nothing by default
+
+We install nothing by default. The operator may treat "no CRDs" as an error case and exit or it may
+treat it as a standard case and run but do nothing.
+
+Pros:
+* Users choose exactly what they need.
+
+Cons:
+* The operator is nonfunctional out of the box, which is a poor new-user experience.
+
+#### Answer 1c: Mandatory configuration
+
+Selection of CRDs to install is required and the operator pod will not launch without it.
+
+Can also use [Helm required values](https://stackoverflow.com/questions/53866196/how-best-to-say-a-value-is-required-in-a-helm-chart)
+to ensure that the value is required at the Helm chart level, although making it required may run afoul of requirement #3.
+
+Pros:
+* Users choose exactly what they need.
+
+Cons:
+* The operator is nonfunctional out of the box, but at least it's clearly nonfunctional and has a good error why.
+
+### Question 2: How do users specify what CRDs to install?
+
+#### Answer 2a: Using the existing command line parameter
+
+Continue using the existing `--crd-pattern` command line parameter on the operator pod
+
+Pros:
+* Mostly exists today
+
+Cons:
+* Command-line isn't preserved between upgrades, which may fail requirement #3 (seamless upgrade experience).
+
+#### Answer 2b: Use a ConfigMap
+
+A `ConfigMap` would persist between upgrades and so in theory could be used to remember what was installed previously
+without the user needing to re-specify it.
+
+One challenge is that in the Helm context if the `ConfigMap` is created by Helm, Helm will overwrite it in
+subsequent upgrades. This could be dealt with by using a
+[pre-install hook](https://stackoverflow.com/questions/55503893/helm-patch-default-service-account). We would create and
+patch the `ConfigMap` in the pre-install hook.
+
+One possible ConfigMap structure:
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: managed-resources
+  namespace: azureserviceoperator-system
+data:
+  "all": ""  # magic key meaning "*" which isn't supported in keys
+  microsoft.resources: "*"
+  microsoft.storage: "StorageAccount,StorageAccountsBlobService"
+```
+
+The following `patch` commands could be run either directly by the user or by our pre-install hook:
+
+* Addition of new resources: `kubectl patch configmap managed-resources -n azureserviceoperator-system --patch '{"data":{"microsoft.storage":"*"}}'`
+* Removal of old resources: `kubectl patch configmap managed-resources -n azureserviceoperator-system --type=json -p='[{"op": "remove", "path": "/data/microsoft.resources"}]'`
+
+If nothing is specified as part of the upgrade, the set of supported resources will not change.
+
+If the `ConfigMap` doesn't exist when we go to **upgrade**, we can assume we're coming from an older ASO version that didn't
+support the `ConfigMap` and default to creating one that signals to install all resources.
+
+Pros:
+* Meets all the requirements.
+* Works for both Helm and raw-YAML situations. Helm as discussed above and raw YAML by having users.
+
+Cons:
+* Complicated to implement.
+* Not the behavior that Helm users would necessarily expect for Helm values.
+* Upgrade scenario from versions prior to this feature will still have all CRDs and thus still have the memory usage problem.
+
+Open questions:
+* Do we even need to support a Helm value for configuring this? Technically speaking the only reason we would need to is
+  to enable ASO as an Extension or Addon, as there everything must be done through Helm. For users of Helm itself we likely
+  can just tell them to run a `kubectl` command afterwards to create the `ConfigMap`. That does make the Helm install
+  experience worse though.
+* We may be able to create the `ConfigMap` as empty given Helms use of
+  [3 way strategic merge](https://helm.sh/docs/faq/changes_since_helm2/#improved-upgrade-strategy-3-way-strategic-merge-patches)
+  and then patch it afterwards.
+* Is it OK that _providers_ would be patchable, but you would need to include all the resources when patching something
+  like `microsoft.storage: "StorageAccount,StorageAccountsBlobService"`? We could just disallow individual resource selection
+  for now, although that might eventually become a problem for certain large providers (`Microsoft.Network` comes to mind).
+
+See also Helm [post rendering](https://helm.sh/docs/topics/advanced/#post-rendering), which we could also consider.
+pre-install hook (or post-install) seems easier though.
+
+#### Answer 2c: Using a CRD
+
+We investigated doing this originally for [Helm chart size limitations](../ADR-2023-02-Helm-Chart-Size-Limitations)
+
+Pros:
+* More structured than a `ConfigMap`.
+
+Cons:
+* Chicken and egg problem of using the CRD that's part of the operator to configure the operator behavior.
+  Helm has a very hard time creating a CRD _and_ an instance of the CRD in the same deployment.
+* May need to implement collection strategic merge behavior that already exists for `ConfigMap`.
+
+### Question 3: How do we ensure consistent configuration as ASO is upgraded
+
+#### Answer 3a: Require the user to be consistent when upgrading
+
+The user would need re-specify what resources they want every time they upgrade
+
+Pros:
+* Easy for us
+
+Cons:
+* Very easy to make a mistake and accidentally leave a resource off they want.
+
+#### Answer 3b: Use a configuration technique that persists across upgrades
+
+Pros:
+* Satisfies the requirement.
+
+Cons:
+* Would augment/complicate the approach we already have
+* More complex for us to implement, especially with Helm. Helm owns any resource it creates, 
+  and deletes it if not specified on the next upgrade. See [Using a ConfigMap](#answer-2b-use-a-configmap) for some 
+  workarounds.
+
+**Variant: Store the configuration independently** 
+
+For example user could specify configuration in a non-durable way (Helm variables that pass to cmdline) and then we could
+merge this configuration into a "store" such as a `ConfigMap`. This differs from the above pattern only in that the user
+would not interact with this `ConfigMap` directly, it would be managed by ASO.
+
+It has similar pros/cons as above.
+
+#### Answer 3c: Query existing CRDs and always support those
+
+Combined with some form of non-persistent storage such as [command line](#answer-2a-using-the-existing-command-line-parameter)
+we would query all existing CRDs and install/upgrade them, while adding whatever new CRDs were specified in 
+the non-persistent storage.
+
+Pros:
+* Automatically supports seamless upgrades.
+* Installed resources are always upgrade to the latest version, even if user has forgotten about them.
+* Could be built onto existing cmdline configuration.
+* Hard for users to get wrong.
+* Single source of truth.
+
+Cons:
+* ???
+
+## Decision
+
+### Answer to [question #1: What to install by default?](#question-1-what-to-install-by-default)
+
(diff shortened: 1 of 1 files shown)


docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
```diff
@@ -0,0 +1,174 @@
...
+      Or maybe it needs to actually upgrade those versions too even though they weren't specified, to maintain a consistent
+      storage version?
+
+### Question 1: How to identify the subset of resources to install by default?
+
+As mentioned above, a key goal is a reduction in the default number of CRDs installed. That means we somehow need to 
+choose which CRDs we will install and which CRDs we will not.
+
+#### Answer 1a: We handpick a set of resources we feel are "critical" and install just those by default
+
```
theunrepentantgeek, line 47:
```suggestion
#### Answer 1a: Handpicked Default

We handpick a set of resources we feel are "critical" and install just those by default

```

docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
```diff
@@ -0,0 +1,174 @@
...
+* Ensures that resources which are extremely likely to be needed (such as Microsoft.Resources/resourceGroup) are included 
+  by default.
+
+Cons:
+* Arbitrary.
+* May result in higher kube-apiserver memory consumption than is strictly needed for users as it is likely to include some
+  resources which they may not use.
+
+#### Answer 1b: We install nothing by default
+
```
theunrepentantgeek, line 59:
```suggestion
#### Answer 1b: Nothing by default

We install nothing by default

```

docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
```diff
@@ -0,0 +1,174 @@
...
+    * No "oops I forgot to select any CRDs, so none were installed"
+    * No "oops I forgot to filter CRDs, so they were all installed"
+    * Exactly the CRDs the user originally specified should be upgraded, and only those CRDs.
+4. No CRDs uninstalled ever.
+    * CRD uninstallation is dangerous. If users want to remove CRDs they can do it manually. 
+    * TODO: A reduction in the list of CRDs to install should result in only the new list CRDs of CRDs getting new versions?
+      Or maybe it needs to actually upgrade those versions too even though they weren't specified, to maintain a consistent
+      storage version?
+
+### Question 1: How to identify the subset of resources to install by default?
```
theunrepentantgeek, line 41:
```suggestion
### Question 1: What to install by default?
```

docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
```diff
@@ -0,0 +1,174 @@
...
+
+#### Answer 1b: We install nothing by default
+
+Pros:
+* Users choose exactly what they need.
+
+Cons:
+* The operator is nonfunctional out of the box, which is a poor new-user experience.
+
+#### Answer 1c: Selection of CRDs to install is required and the operator pod will not launch without it
```
theunrepentantgeek, line 66:
```suggestion
#### Answer 1c: Mandatory configuration

The selection of CRDs to install is required and the operator pod will not launch without it
```

docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Question 2: How do users specify what CRDs to install?
+
+#### Answer 2a: using our existing --crd-pattern command line parameter on the operator pod
+
+Pros:
+* Exists today
+
+Cons:
+* Command-line isn't preserved between upgrades, which fails requirement #3 (seamless upgrade experience).
```
theunrepentantgeek, line 85:
```suggestion
* Command-line isn't preserved between upgrades, which may fail requirement #3 (seamless upgrade experience).
```

docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
```diff
@@ -0,0 +1,174 @@
...
+
+A `ConfigMap` would persist between upgrades and so in theory could be used to remember what was installed previously
+without the user needing to re-specify it.
+
+One challenge is that in the Helm context if the `ConfigMap` is created by Helm, Helm will overwrite it in 
+subsequent upgrades. This could be dealt with by using a 
+[pre-install hook](https://stackoverflow.com/questions/55503893/helm-patch-default-service-account). We would create and
+patch the `ConfigMap` in the pre-install hook.
+
+ConfigMap structure:
```
theunrepentantgeek, line 97:
```suggestion
One possible configMap structure:
```

docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
```diff
@@ -0,0 +1,174 @@
...
+
+The following `patch` commands could be run either directly by the user or by our pre-install hook:
+
+* Addition of new resources: `kubectl patch configmap managed-resources -n azureserviceoperator-system --patch '{"data":{"microsoft.storage":"*"}}'`
+* Removal of old resources: `kubectl patch configmap managed-resources -n azureserviceoperator-system --type=json -p='[{"op": "remove", "path": "/data/microsoft.resources"}]'`
+
+If nothing is specified as part of the upgrade, the set of supported resources will not change.
+
+If the `ConfigMap` doesn't exist when we go to upgrade, we can assume we're coming from an older ASO version that didn't
+support the `ConfigMap` and default to creating one that signals to install all resources.
```
theunrepentantgeek, line 118: This would put us back into the high memory use scenario.

docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
```diff
@@ -0,0 +1,174 @@
...
+
+Pros:
+* Users choose exactly what they need.
+
+Cons:
+* The operator is nonfunctional out of the box, but at least it's clearly nonfunctional and has a good error why.
+
+### Question 2: How do users specify what CRDs to install?
+
+#### Answer 2a: using our existing --crd-pattern command line parameter on the operator pod
```
theunrepentantgeek, line 79:
```suggestion
#### Answer 2a: Use the existing command line parameter

Continue using the existing `--crd-pattern` command line parameter on the operator pod.
```

docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
```diff
@@ -0,0 +1,174 @@
...
+
+Pros:
+* Exists today
+
+Cons:
+* Command-line isn't preserved between upgrades, which fails requirement #3 (seamless upgrade experience).
+
+#### Answer 2b: Using a ConfigMap
```
theunrepentantgeek, line 87:
```suggestion
#### Answer 2b: Use a ConfigMap
```

docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
```diff
@@ -0,0 +1,174 @@
...
+
+See also Helm [post rendering](https://helm.sh/docs/topics/advanced/#post-rendering), which we could also consider.
+pre-install hook (or post-install) seems easier though.
+
+#### Answer 2c: Using a CRD
+
+We investigated doing this originally for [Helm chart size limitations](../ADR-2023-02-Helm-Chart-Size-Limitations)
+
+Pros:
+* Slightly more structured than a `ConfigMap`.
```
theunrepentantgeek, line 148:
```suggestion
* Significantly more structured than a `ConfigMap`.
```

docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
```diff
@@ -0,0 +1,174 @@
...
+
+Cons:
+* Chicken and egg problem of using the CRD that's part of the operator to configure the operator behavior. 
+  Helm has a very hard time creating a CRD _and_ an instance of the CRD in the same deployment.
+* May need to implement collection strategic merge behavior that already exists for `ConfigMap`.
+
+## Decision
+
+For question #1: [Answer 1c: Selection of CRDs to install is required and the operator pod will not launch without it](#answer-1c--selection-of-crds-to-install-is-required-and-the-operator-pod-will-not-launch-without-it)
+For question #2: [Answer 2b: Using a ConfigMap](#answer-2b--using-a-configmap)
```
theunrepentantgeek, line 158: With the separation of question 2 into "how to configure" and "how to ensure consistency" (as mentioned elsewhere), I think we take a different approach:

* Mandatory configuration
* Retain Command line parameter
* Always support CRDs that have already been installed

theunrepentantgeek (changes requested):
A good write up of the problem, but I disagree with your conclusion.

I think your answers for Question 2 are actually trying to answer two different questions:
* How do users configure which CRDs to install
* How do we ensure consistent configuration as ASO is upgraded?

For the later question, there are at least four answers:

1. Require the user to be consistent when installing any new version 
   * Pro: Easy for us
   * Con: Complex for the user, easy for them to make a mistake
2. Use a configuration technique that persists across upgrades
   * Pro: Nothing special for us to do
   * Con: Chick and egg problem for installation (incl race conditions)
   * Con: More complex for us to implement
   * Con: Discarding the approach we have already implemented
3. Store our configuration in a config map and merge it after upgrade
   * Similar to the way Helm stashes a copy of installed CRDs for later use
   * Con: Might be surprising to users
   * Con: Fragile - things go awry if user deletes the config map item
5. Query for existing CRDs and always support those
   * Pro: Seemless upgrades (Requirements #2, #3)
   * Pro: consistent with no-CRD Deletion (Requirement #4)
   * Pro: Installed resources never stop working
   * Pro: We can use existing commandline configuration
   * Pro: hard for users to get wrong
   * Pro: automatically works when users manually deploy CRDs

docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
matthchr: Yes but it'd be a breaking change IMO to do anything other than this, right?

docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
```diff
@@ -0,0 +1,174 @@
...
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: managed-resources
+  namespace: azureserviceoperator-system
+data:
+  "all": ""  # magic key meaning "*" which isn't supported in keys
+  microsoft.resources: "*"
+  microsoft.storage: "StorageAccount,StorageAccountsBlobService"
```
super-harsh, line 107: We might need handling somewhere to check if parent CRD is installed or not. As for in this example, what if a new user, does something like `microsoft.storage: "StorageAccountsBlobService"` and assume it'll include StorageAccount or if its not required.

docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
theunrepentantgeek: Yup. I just figure it's worth calling out as a Con for this approach.

docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
matthchr: Done

docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
matthchr: Note I'm specifically here referring to just upgrade scenarios. New installs user would need to choose still. I'll call that out as well.

matthchr: Added the con

docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
matthchr: Dropped the qualifier and said "More structured than a `ConfigMap`"

docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
matthchr: I'll add an open questions section and put this there. My current thinking on this is that we avoid this complexity and strongly recommend users install entire providers such as `compute/*`.

There are other links to consider too:
- Owners (i.e. resourceGroup owns a lot of things even though it's not in their group)
- Cross resource references: i.e. compute references networking. This is pretty involved to work out automatically and there's also no guarantee that just because a resource _can_ link to a Subnet or a LoadBalancer that the customer actually has that use-case.

matthchr:
> A good write up of the problem, but I disagree with your conclusion.
> 
> I think your answers for Question 2 are actually trying to answer two different questions:
> 
> * How do users configure which CRDs to install
> * How do we ensure consistent configuration as ASO is upgraded?
> 
> For the later question, there are at least four answers:
> 
> 1. Require the user to be consistent when installing any new version
>    
>    * Pro: Easy for us
>    * Con: Complex for the user, easy for them to make a mistake
> 2. Use a configuration technique that persists across upgrades
>    
>    * Pro: Nothing special for us to do
>    * Con: Chick and egg problem for installation (incl race conditions)
>    * Con: More complex for us to implement
>    * Con: Discarding the approach we have already implemented
> 3. Store our configuration in a config map and merge it after upgrade
>    
>    * Similar to the way Helm stashes a copy of installed CRDs for later use
>    * Con: Might be surprising to users
>    * Con: Fragile - things go awry if user deletes the config map item
> 4. Query for existing CRDs and always support those
>    
>    * Pro: Seemless upgrades (Requirements [initial pr template draft #2](https://github.com/Azure/azure-service-operator/pull/2), #3)
>    * Pro: consistent with no-CRD Deletion (Requirement [EventHub Service Operator #4](https://github.com/Azure/azure-service-operator/issues/4))
>    * Pro: Installed resources never stop working
>    * Pro: We can use existing commandline configuration
>    * Pro: hard for users to get wrong
>    * Pro: automatically works when users manually deploy CRDs

I'm not sure that I follow what the difference is between 2 and 3. Isn't 2 just a generic version of 3?

On 4: Agree this seems like a great option. I've updated the document to include it, although I did remove this pro:
```
   * Pro: consistent with no-CRD Deletion
```

As every option has this pro (we never issue CRD delete no matter what so it feels weird to call it out explicitly here when we haven't anywhere else).

docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
matthchr: Not sure exactly what "Retain command line parameter" means - does it just mean "use the option we have now" or are you talking about some cmdline parameter that is retained between deployments somehow automatically?

I agree with your argument towards "Always support CRDs that have already been installed", but I think that actually forces us away from "Mandatory configuration", at least in the sense of Helm required parameters or anything like that.

I think what we want is:
- Operator exits with a clear reason (visible on Pod status) if no CRDs are configured.
- Cmdline parameter to configure which _new_ CRDs to install
- All existing CRDs are automatically upgraded to the latest version, even if not specified in cmdline parameter.

If we have some form of mandatory configuration _and_ the "existing CRDs" behavior it's not clear what the user should put into the mandatory configuration if they want to just use the resources they were already using.

theunrepentantgeek:
> I'm not sure that I follow what the difference is between 2 and 3. Isn't 2 just a generic version of 3?

Option 2 is suggesting that we specifically use a form of configuration that persists across upgrades.

We could use a config map entry or a custom resource, but if the user overwrites it (either deliberately or accidentally) we would have the failure mode of orphaning existing resources.

Option 3 is suggesting that we make a copy of the configuration (regardless of the form) and stash it away somewhere so that we can reference it later, comparing it with what we're given later.

We can then compare/contrast/merge user provided configuration with our "memory" of what happened last time to ensure things run smoothly - or that we give better diagnostics.

docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
```diff
@@ -0,0 +1,248 @@
...
+
+Pros:
+* Satisfies the requirement.
+
+Cons:
+* Would augment/complicate the approach we already have
+* More complex for us to implement, especially with Helm. Helm owns any resource it creates, 
+  and deletes it if not specified on the next upgrade. See [Using a ConfigMap](#answer-2b--use-a-configmap) for some 
+  workarounds.
+
```
matthchr, line 187: Add variant: "Store the configuration independently..."

docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
```diff
@@ -0,0 +1,248 @@
...
+### Answer to [question #2: How do users specify what CRDs to install?](#question-2--how-do-users-specify-what-crds-to-install)
+
+[Answer 2a: Using the existing command line parameter](#answer-2a--using-the-existing-command-line-parameter)
+
+This parameter will come to mean: "New CRDs to install to the cluster".
+
+### Answer to [question #3: How do we ensure consistent configuration as ASO is upgraded?](#question-3--how-do-we-ensure-consistent-configuration-as-aso-is-upgraded)
+
+[Answer 3c: Query existing CRDs and always support those](#answer-3c--query-existing-crds-and-always-support-those)
```
matthchr, line 222: Note: We need to deal with the users-manage-their-own CRDs case, explain what we do for that here.

matthchr: Confirm that in the "we dont manage CRDs case, we still have GET CRDs permissions"

docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
```diff
@@ -0,0 +1,248 @@
...
+Combined with some form of non-persistent storage such as [command line](#answer-2a--using-the-existing-command-line-parameter)
+we would query all existing CRDs and install/upgrade them, while adding whatever new CRDs were specified in 
+the non-persistent storage.
+
+Pros:
+* Automatically supports seamless upgrades.
+* Installed resources are always upgrade to the latest version, even if user has forgotten about them.
+* Could be built onto existing cmdline configuration.
+* Hard for users to get wrong.
+
```
matthchr, line 199: + Single source of truth

docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
```diff
@@ -0,0 +1,248 @@
...
+If a user imports `microsoft.storage/storageAccountsBlobService` should we automatically determine that they 
+also want `microsoft.storage/storageAccounts` since that's the `owner` type?
+
+Conclusion: No, users must list all resources they want explicitly. We strongly recommend using full provider imports such as 
+`microsoft.networking/*` rather than tightly scoping to specific resources.
+
+## Status
+
+Proposed.
```
matthchr, line 236: We need to document:
1. What the CRD uninstall process is (especially for users from v2.0.0 where they have everything).
2. Explain how to actually select new CRDs
3. Explain how CRDs are managed/installed.

matthchr, line 236: Open question, even without a predefined set of "default" resources, do we want a "default" set in the ASO install command on the readme so that people know what to do?

Consensus seems to be: yes.

docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
matthchr: Updated documentation to cover this case. Fundamentally, it should work exactly as planned. See the document which outlines the basic algorithm

docs/hugo/content/design/ADR-2023-05-Installing-Only-Selected-CRDs.md
matthchr: Have made notes of all this. Documentation will be included in the implementation PR.

matthchr: Made a note of this. Docs will be in the implementation PR

theunrepentantgeek (approved):

PR #2980, opened 2023-05-15
Improve workflows

theunrepentantgeek:

Actions requiring Node v12 will stop working soon; we need to migrate to newer versions of these actions.

Some recurring workflows have been silently failing for some time; I've added badges to our README to make these visible.

image

Update _Validate Pull Request (fork)_ to use newer version of `actions/github-script`
* No workflow dispatch option for triggering, untested

Update _Validate Pull Request_ to user newer version of `actions/upload-artifact`
* No workflow dispatch option for triggering, untested

Update _Render Visualizations of ASO Components_ to use newer version of `peterjgrainger/action-create-branch`
* Tested by manual trigger

Update _Publish helm charts_ to user newer version of `peterjgrainger/action-create-branch`
* Has a workflow dispatch trigger, but don't want to actually publish a new helm chart. Untested

Update _Add 'needs-triage' label to new issues_ to replace action `andymckay/labeler` which is now archived.
* No workflow dispatch option for triggering, untested

We'll need to carefully review these, then test after merging.

gif


.github/workflows/add-needs-triage-label.yml
```diff
@@ -1,18 +1,23 @@
-name: Add 'needs-triage' label to new issues
+name: Add 'needs-triage' label to new issuesZ
```
super-harsh, line 1:
```suggestion
name: Add 'needs-triage' label to new issues
```

super-harsh (approved):

matthchr:
> Some recurring workflows have been silently failing for some time; I've added badges to our README to make these visible.

I'm not sold that we should put the badges of our failing workflows on our front page. It feels like we should fix them and then put a badge.

We had a badge for "Validate with Live Azure resources" that was failing for ages before I removed it, so I'm not sure that having the badge really means we'll have time to dig in. I wonder if that workflow is fundamentally flawed. It takes a long time to run and is only getting longer as we add more resources. It feels destined to be flaky due to real Azure flakes or quota. AFAIK its purpose is "what if the actual RP breaks their contract, would we notice" - with pure recording tests the answer is no.

It's not clear to me if:
1. That's really something we're all that concerned about happening in practice (is it worth the $ cost in terms of eng hours or Azure billing $$)
2. It's our job to catch such a regression in the first place.

The tester in me wants to say: "Hey having your tests run periodically to ensure that they can pass in live mode is valuable because if they're so flaky they can't pass then how can you re-record them?" but given quota realities, etc, if we want to pursue that avenue it still feels like we need to choose a curated list of the most interesting/import resources rather than everything.

v2/README.md
```diff
@@ -3,6 +3,14 @@ title: Azure Service Operator v2
 type: docs
 description: "Manage your Azure resources from within your Kubernetes cluster."
 ---
+
+[![Validate with live Azure resources](https://github.com/Azure/azure-service-operator/actions/workflows/live-validation.yml/badge.svg)](https://github.com/Azure/azure-service-operator/actions/workflows/live-validation.yml) [![Build Devcontainer image](https://github.com/Azure/azure-service-operator/actions/workflows/build-devcontainer-image.yml/badge.svg)](https://github.com/Azure/azure-service-operator/actions/workflows/build-devcontainer-image.yml)
+
+[![Publish docs site](https://github.com/Azure/azure-service-operator/actions/workflows/deploy-site.yml/badge.svg)](https://github.com/Azure/azure-service-operator/actions/workflows/deploy-site.yml)
```
matthchr, line 9: IMO the README.md is for users, not for us. I don't think we should put these here. They don't add clarity for our users regardless of their state (green or red).

If we want a way of notifying us all of failed actions we could do something [like this](https://www.ravsam.in/blog/send-email-notification-when-github-action-fails/) with a dummy gmail account... I don't love that but I can't see any way to actually get proactive notifications from GitHub itself. Feels like we should put in a feature ask for that (or more likely find the existing one... there is one [here](https://github.com/orgs/community/discussions/43415))

The only one that maybe makes sense to put here is `Validate with live Azure resources`, although see my other comment about hesitancy there - we used to have it here and I removed it because it's always failing. IMO we either need to make it pass and then add it or just remove it altogether.

Rest maybe could go into CONTRIBUTING.md instead if not using some email solution, as they're all for contributors?

theunrepentantgeek: Badges removed from README.

PR #2982, opened 2023-05-15
Add logo to website

theunrepentantgeek:

Now we've selected a logo, this PR integrates it into our documentation website, as both favicon (shows in the site tab) and as logo.

image

Visibility of Clampy Op Bot was poor against the default blue colour used by our Hugo theme, Docsy. Changing it to an Azure Blue seemed appropriate.

gif


super-harsh (approved):

super-harsh:
LGTM! Minor suggestion: Should we have the logo to be a bit bigger than the `Azure Service Operator` title?

theunrepentantgeek:
> Should the space between their "clamps" be transparent too? 
> Otherwise it looks like they are holding something (empty space?) but it's not clear what?

That's the image as supplied - it deliberately has non-transparent squares in the grips of the clamps; I think the intent is to ensure the clamps remain visible.

theunrepentantgeek:
> LGTM! Minor suggestion: Should we have the logo to be a bit bigger than the `Azure Service Operator` title?

It's an SVG and scales according to how the docsy theme specifies. No doubt we can make it better, but it's not obvious to me how to do that. It's not as simple as using a larger PNG.

PR #2984, opened 2023-05-16
Add support for LoadBalancersInboundNatRule

### LINKED ISSUE #2894: Feature: Support LoadBalancer InboundNATRules
The actual resource is: microsoft.network/loadbalancers/inboundnatrules

https://learn.microsoft.com/en-us/azure/templates/microsoft.network/loadbalancers/inboundnatrules

super-harsh:
Closes #2894 

This PR adds support for LoadBalancersInboundNatRule to be created and managed separately. 

Added load_balancer_extension.go to handle the edge case where InboundNatRule is present on LoadBalancer and independently.

**If applicable**:
- [x] this PR contains tests


v2/api/network/customizations/load_balancer_extension.go
```diff
@@ -0,0 +1,171 @@
...
+		return errors.Wrapf(err, "failed to unmarshal inboundNatRule JSON")
+	}
+
+	// Safety check that these are actually the same:
+	// We can't use reflect.DeepEqual because the types are not the same.
+	embeddedInboundNatRuleJSON, err := json.Marshal(embeddedInboundNatRule.Interface())
+	if err != nil {
+		return errors.Wrap(err, "unable to check that embedded inboundNatRule is the same as inboundNatRule")
+	}
+	if string(embeddedInboundNatRuleJSON) != string(inboundNatRuleJSON) {
```
theunrepentantgeek, line 166: Does JSON serialization have a guarantee of ordering? I suspect this may be fragile.

v2/api/network/customizations/load_balancer_extension.go
```diff
@@ -0,0 +1,171 @@
...
+	typedObj, ok := obj.(*network.LoadBalancer)
+	if !ok {
+		return nil, errors.Errorf("cannot run on unknown resource type %T, expected *network.LoadBalancer", obj)
+	}
+
+	// Type assert that we are the hub type. This will fail to compile if
+	// the hub type has been changed but this extension has not been updated to match
+	var _ conversion.Hub = typedObj
+
+	inboundNatRuleGVK := getInboundNatRuleGVK(obj)
```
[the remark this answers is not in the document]
matthchr: The problem is that we are injecting one type into another. Imagine for example:

LoadBalancer OriginalGVK `20230101`
LoadBalancerNATRule OriginalGVK `20230601`

The "contract" (if you can call it that) is that `20230101` LoadBalancer can embed `20230101` NAT rules, but it can't embed the `20230601` version because that's newer. We need to run conversion.

This has a requirement is poorly documented here and should probably be called out (here and in the vmss and RouteTable extenions too): In each API version, there must be a version of the subresource _and_ the parent resource (no skipping one or the other). We should probably put a comment in azure-arm.yaml to that effect as well next to these resources.

I believe if the GVK doesn't exist, the `List` call will fail saying no such GVK, so at runtime we're "protected" here somewhat.

**The issue is** you can't know statically what GVK you want, so you can't do what you suggest and get the GVK statically. We need to get the GVK dynamically based on GVK that we're going to use to generate the ARM payload.

v2/api/network/v1api20201101/load_balancer_types_gen.go
```diff
@@ -8530,25 +8530,6 @@ func (embedded *Subnet_STATUS_LoadBalancer_SubResourceEmbedded) AssignProperties
...
-
-// The transport protocol for the endpoint.
-type TransportProtocol_STATUS string
-
-const (
-	TransportProtocol_STATUS_All = TransportProtocol_STATUS("All")
-	TransportProtocol_STATUS_Tcp = TransportProtocol_STATUS("Tcp")
-	TransportProtocol_STATUS_Udp = TransportProtocol_STATUS("Udp")
-)
-
```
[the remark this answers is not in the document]
super-harsh: Its added in [here](https://github.com/Azure/azure-service-operator/pull/2984/files#diff-f54d917944323904808a0736d1e9a4ffb5814a0ff1ab7b147d8730b8647c11f6R1408)

v2/api/network/customizations/load_balancer_extension.go
super-harsh: This is the same logic we've been using for other resource extensions(vnet, routetable) as well and noticed no serialization issues. Can you think of a better way to do this?

v2/api/network/customizations/load_balancer_extension.go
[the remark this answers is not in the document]
matthchr: Reiterating this bit:

> This has a requirement that is poorly documented here and should probably be called out (here and in the vmss and RouteTable extenions too): In each API version, there must be a version of the subresource and the parent resource (no skipping one or the other). We should probably put a comment in azure-arm.yaml to that effect as well next to these resources.

My takeaway from @theunrepentantgeek's comment, which we could try to find a way to do, is to ensure that there really _is_ such a version for the parent type and the child type. right now we just make the assumption. Obviously that happens to work in practice but we're not very hardened against issues with that assumption here. Some combination of:
1. Comment in `azure-arm.yaml` saying that there must always be a parent and child in the same API version (to call it out for others when adding new versions)
2. Dynamic checks via scheme lookup here to ensure that such a type actually exists.

Seems the best way out to me?

I at least would be happy to see this done in a separate PR where we tackled all 3-4 of the resources that have this requirement together, rather than trying to do them all here, but also curious what @theunrepentantgeek thinks.

v2/api/network/customizations/load_balancer_extension.go
matthchr: According to this [stackoverflow post](https://stackoverflow.com/questions/44755089/does-serialized-content-strictly-follow-the-order-in-definition-use-encoding-jso), which references the Go JSON serialization code, struct order will be consistent.

That's an implementation detail though and is technically subject to change.

Map ordering is not guaranteed, so if the structs had maps we could be at risk here. Since we can't know for sure that the structs don't have maps, I think in general I agree with Bevan that this is not ideal.

_Possibly_ what we could do here is:
1. Serialize these types to JSON
2. Deserialize them into a generic `map[string]interface{}` (map of maps)
3. use `reflect.DeepEqual` on those.

While the actual JSON representations of the types might differ because of ordering, I _believe_ that `reflect.DeepEqual` is not ordering sensitive for maps and so using that would be safe.

As above it might make sense to merge this as-is (given it's a bug we have across all the impls right now) and fix it across all of them together for consistency.

matthchr: if we do decide to fix this but in a different PR, let's file a bug for it so we don't forget. Make sure to mention RouteTable and VNETs there too.

v2/internal/controllers/controller_resources.go
```diff
@@ -132,9 +132,17 @@ func getGeneratedStorageTypes(
 
 	// Verify we're using the hub version of VirtualNetworksSubnet in the loop below
 	var _ ctrlconversion.Hub = &networkstorage.VirtualNetworksSubnet{}
+	var _ ctrlconversion.Hub = &networkstorage.LoadBalancersInboundNatRule{}
+	var _ ctrlconversion.Hub = &networkstorage.RouteTablesRoute{}
```
matthchr, line 136: minor: Why is this line (136) needed here?

super-harsh: Hub version verification was missing for `RouteTablesRoute` so just added that here

v2/api/network/customizations/load_balancer_extension.go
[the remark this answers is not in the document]
theunrepentantgeek: Ergh. I get it now. 

My key problem here is that any failure happens at runtime of the controller, while the fix requires changes to the configuration of the generator - these are _very_ far apart in the code base, and the errors will happen at very different times. It's also more than likely the person experiencing the failure won't be the person who needs to change the configuration.

I'd suggest the following attack:
* In this PR, add comments in `azure-arm.yaml` to all the afflicted resources (not just the new resources in this PR).
* Log an issue to track the problem
* Add config + check to the pipeline to enforce the relationship, somehow, so that we can't possibly leave things out
* Revisit this code to see if we can find a less-brittle way to do the conversion that's required

PR #2986, opened 2023-05-16
Fix samples tester suite to update arm references for all resources

super-harsh:

Found this minor bug while recording `LoadBalancersInbountNatRule`, where setting the resource references was getting skipped.


v2/internal/testcommon/samples_tester.go
```diff
@@ -211,12 +211,13 @@ func (t *SamplesTester) setOwnershipAndReferences(samples map[string]genruntime.
...
-			var err error
 			sample = setOwnersName(sample, ownersName)
-			sample, err = t.updateARMReferencesForTest(sample)
-			if err != nil {
-				return err
-			}
+		}
+
+		var err error
+		sample, err = t.updateARMReferencesForTest(sample)
```
super-harsh, line 218: Putting it outside the check. Since the method was not called when the owner is not RG and neither we're using random names.

super-harsh: We were only updating references for resources owned by RG _and_ were using random names.

PR #2994, opened 2023-05-19
Refactor matcher to be reusable

matthchr:
* Moved out of generator and into v2/internal so it can be used across both generator and controller.
* Replaced panics with error returns.
* New constructor that doesn't support multi (with ';').

This is in preparation for implementing CRD filtering as discussed in #2973.

**If applicable**:
- [x] this PR contains tests


v2/internal/util/match/field_matcher.go
```diff
@@ -17,10 +17,14 @@ type FieldMatcher struct {
...
 var _ StringMatcher = &FieldMatcher{}
 
-func NewFieldMatcher(field string) FieldMatcher {
-	return FieldMatcher{
-		actual: NewStringMatcher(field),
+func NewFieldMatcher(field string) (FieldMatcher, error) {
+	matcher, err := NewStringMultiMatcher(field)
+	if err != nil {
+		return FieldMatcher{}, err
 	}
```
theunrepentantgeek, line 24: Missing line after close of err scope. #minor

v2/internal/util/match/glob_matcher.go
```diff
@@ -3,46 +3,42 @@
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to create regexp for glob %q", glob)
 	}
 
 	return &globMatcher{
 		glob:       glob,
+		regex:      regex,
 		candidates: make(set.Set[string]),
-	}
+	}, nil
```
theunrepentantgeek, line 38: The panic was correct.

This factory method should remain private (`newGlobMatcher`), and only be called when the predicate `HasWildCards()` returns true. It's a programming bug if the glob string has no wildcards, not an error.

v2/internal/util/match/glob_matcher.go
```diff
@@ -78,20 +74,21 @@ func (gm *globMatcher) IsRestrictive() bool {
...
 }
 
-func (gm *globMatcher) createRegex() {
-	g := regexp.QuoteMeta(gm.glob)
+// String returns the literal we match
+func (gm *globMatcher) String() string {
+	return gm.glob
+}
+
+func newGlobRegex(glob string) (*regexp.Regexp, error) {
```
theunrepentantgeek, line 82: With this construction, it's not possible for the regex to panic; no error return required.

v2/internal/util/match/literal_matcher.go
```diff
@@ -3,26 +3,28 @@
...
-	advisor *TypoAdvisor
+	advisor *typo.Advisor
 }
 
 var _ StringMatcher = &literalMatcher{}
 
-// newLiteralMatcher returns a new literalMatcher for the given string
-func newLiteralMatcher(literal string) *literalMatcher {
+// NewLiteralMatcher returns a new literalMatcher for the given string
+func NewLiteralMatcher(literal string) StringMatcher {
```
theunrepentantgeek, line 24: Should be private. The public API for matchers is through `StringMatcher` and `NewStringMatcher`

```suggestion
// newLiteralMatcher returns a new literalMatcher for the given string
func newLiteralMatcher(literal string) StringMatcher {
```

v2/internal/util/match/multi_matcher.go
```diff
@@ -12,29 +12,33 @@ import (
...
 	matchers []StringMatcher
 }
 
-var _ StringMatcher = &MultiMatcher{}
+var _ StringMatcher = &multiMatcher{}
 
-// newMultiMatcher returns a new matcher for multiple strings
-func newMultiMatcher(matcher string) *MultiMatcher {
+// NewMultiMatcher returns a new matcher for multiple strings
+func NewMultiMatcher(matcher string) (StringMatcher, error) {
```
theunrepentantgeek, line 23: Private.

```suggestion
// newMultiMatcher returns a new matcher for multiple strings
func newMultiMatcher(matcher string) (StringMatcher, error) {
```

v2/internal/util/match/string_matcher.go
```diff
@@ -0,0 +1,47 @@
...
+	WasMatched() error
+	// IsRestrictive returns true if the matcher is populated and will restrict matches
+	IsRestrictive() bool
+}
+
+// NewStringMatcher returns a matcher for the specified string
+// Different strings may return different implementations:
+// o If the string contains '*' or '?' a globbing wildcard matcher
+// o Otherwise a case-insensitive literal string matcher
+func NewStringMatcher(matcher string) (StringMatcher, error) {
```
theunrepentantgeek, line 28: An empty string should return an AlwaysMatcher

v2/internal/util/match/string_matcher.go
```diff
@@ -0,0 +1,47 @@
...
+// o If the string contains ';', a multi-matcher of sub-matches, one for each item in the string separated by ';'
+// o If the string contains '*' or '?' a globbing wildcard matcher
+// o Otherwise a case-insensitive literal string matcher
+func NewStringMultiMatcher(matcher string) (StringMatcher, error) {
+	if HasMultipleMatchers(matcher) {
+		return NewMultiMatcher(matcher)
+	}
+
+	return NewStringMatcher(matcher)
+}
```
theunrepentantgeek, line 47: Do we really need multiple entry points like this? Under what circumstances would a multimatcher be a problem?

v2/internal/util/typo/typo_advisor.go
```diff
@@ -3,7 +3,7 @@
  * Licensed under the MIT license.
  */
 
-package config
+package typo
```
theunrepentantgeek, line 6: Nice.

v2/tools/generator/internal/config/always_matcher.go
```diff
@@ -1,36 +0,0 @@
-/*
- * Copyright (c) Microsoft Corporation.
- * Licensed under the MIT license.
- */
-
-package config
-
-// alwaysMatcher is a matcher that always matches
-type alwaysMatcher struct{}
```
theunrepentantgeek, line 9: Why delete this?

Haven't you changed the behaviour for matching empty strings?

theunrepentantgeek (changes requested):
Some of the changes to the Matcher API are an issue - consumers shouldn't be aware of what kind of matcher is returned (shouldn't have access to know at all); everything should be funnelled through `NewStringMatcher()`. 

It's also unclear why MultiMatcher needs to be called out for special treatment with a different entry point.

Mostly good though. :-)

v2/internal/util/match/glob_matcher.go
matthchr: r.e. private, I think that makes sense.

> It's a programming bug if the glob string has no wildcards, not an error.

"myString" is a valid glob. It's a glob with no wildcards but that's ok, it works just fine? How is it a programming bug?

v2/internal/util/match/literal_matcher.go
matthchr: agreed

v2/internal/util/match/multi_matcher.go
matthchr: agreed

v2/internal/util/match/string_matcher.go
matthchr: why?

It's not used anywhere in the config today (note its removal here but CI passes), and that's not the behavior we want/need for CRD globbing - an empty string should match nothing, not everything. Otherwise it's too easy for users to accidentally include everything. They should need to be explicit and pass "*"

matthchr: I could've kept it and have split initialization that bypassed the `alwaysMatcher` possibility for the CRD globbing case but it didn't seem worth it seeing as this is never actually used in our config today. (I tested w/ a breakpoint in it, it's never hit)

v2/internal/util/match/string_matcher.go
matthchr: The `--crd-patterns` cmdline argument was originally documented as comma delimited, and this felt weird: `"foo/*;bar/*,baz/*"` -- but I suppose we could switch it to be `;` delimited in which case we could have a single path here? Thoughts?

v2/tools/generator/internal/config/always_matcher.go
matthchr: It's never used. If the behavior had been changed CI would break and/or there would've been generated code changes. I tested w/ a breakpoint too and this path is never taken. See comment above about why.

v2/internal/util/match/string_matcher.go
theunrepentantgeek: Maybe an empty string should return a `NeverMatcher`. 

My thinking was that matching an empty string using a literal matcher was brittle (programming by coincidence) and that being explicit was better. 

But maybe it's ok.

v2/internal/util/match/string_matcher.go
theunrepentantgeek: I'd favour changing `MultiMatcher` so that it accepted either `;` or `,` (but not both), giving users flexibility without allowing things that would be confusing (e.g. should `"foo/*;bar/*,baz/*"` be treated as `"foo/*` + `bar/*,baz/*"` or `"foo/*;bar/*` + `baz/*"` or `"foo/*` + `bar/*` + `baz/*"`)

I think then we could go back to a single entry point. 

I don't think it matters if we document the patterns as `,` separated and then happen to accept `;` as well - it's a case of Postel's law.

matthchr: Let me look into this some more and see what makes sense here. Will update shortly

v2/tools/generator/internal/config/always_matcher.go
matthchr: Having this conversation in the other comment thread so resolving this one

v2/internal/util/match/glob_matcher.go
theunrepentantgeek: It's the construction of the regex string that makes panic impossible. It also makes an error impossible.

`regexp.QuoteMeta` returns a valid regex that matches the exact string given - all regex operators are fully escaped. This regex will compile without panic or error.

The first replacement turns quoted `*` wildcards into matches. This regex will compile without issues.
The second replacement turns quoted `?` wildcards into matches. This regex will also compile without issues.

v2/internal/util/match/glob_matcher.go
theunrepentantgeek: A wildcard matcher should only ever be used with a match string containing wildcards. 

If it's called with anything else, that's an error in programming. The way this _was_ ensured was that the factory method was only ever called from one place, and that call protected by the `HasWildcards()` predicate.

v2/internal/util/match/glob_matcher.go
matthchr: > It also makes an error impossible.

How do we actually know this is true? I looked at [regex parse](https://go.dev/src/regexp/syntax/parse.go) and there seem to be (to me) at least a few error codes which could occur even with the protection we have here.
1. `ErrInvalidUTF8`
2. `ErrLarge`

There might be others as well (I haven't examined the implementation of regexp in detail, nor should we really have to in any case because the impl might change).

IMO the choices here are:
1. Use `MustCompile`, which could panic. AFAIK the only panics would be because of wonky user input, most things will work.
2. Use `Compile` and return the error, which is what I am doing here, primary to avoid the panic and give the caller the chance to attach additional details before deciding if it needs to exit or not.

Are you arguing for 1? Or are you arguing for some other option I don't see?

v2/internal/util/match/glob_matcher.go
matthchr: Now that this is private again (I had made it public in an earlier refactor and then when I moved things into the new `match` package forgotten to make them private again), the above is true again.

I've brought the `HasWildCards` check back at the top of this constructor as well along with the panic.

v2/internal/util/match/string_matcher.go
matthchr: The advantage of using `LiteralMatcher` is that:
1. `""` is a literal
2. `LiteralMatcher` has suggestions built in and so gives a better error experience than a `NeverMatcher` would (unless we implemented the same suggestions in `NeverMatcher`, I suppose).

v2/internal/util/match/string_matcher.go
matthchr: Ok so I dug into this a bit more and here's what I've got:

Here's the codepath as I have it locally:
1. User specifies `--crd-pattern=a,b,c,d`. This is parsed by [string_slice_flags](https://github.com/Azure/azure-service-operator/blob/main/v2/internal/util/flags/string_slice_flags.go) into a `[]string`.
2. This flows into a function `filterCRDsByPatterns` `in `crdmanagement` (new in my local) which translates each user specified pattern into a `match.StringMatcher`. So it's a `[]string` -> `[]match.StringMatcher` conversion.
3. Each matcher is checked individually, and each CRD it matches is cataloged.
4. An error is raised if any matcher doesn't match anything

We could re-use the `MultiMatcher` implementation and go from `[]string` -> `match.MultiMatcher` (or possibly from `string` -> `match.MultiMatcher`, but this presents one challenge:

1. I want to log which pattern matched each CRD.
    a. This could probably be fixed by extending the `match.StringMatcher` interface to have some sort of `bool, string` result where it included textual details about which pattern matched each string, although that interface extension is a bit awkward for most of the other `match` implementations because they only ever can be matched by a single pattern so it's sorta obvious.


I'll investigate changing this and see how it goes.

v2/internal/util/match/string_matcher.go
matthchr: I can dig in a bit more to why TypeMatcher is still matching everything/working

v2/internal/util/match/field_matcher.go
```diff
@@ -17,10 +17,15 @@ type FieldMatcher struct {
 
 var _ StringMatcher = &FieldMatcher{}
 
-func NewFieldMatcher(field string) FieldMatcher {
-	return FieldMatcher{
-		actual: NewStringMatcher(field),
+func NewFieldMatcher(field string) (FieldMatcher, error) {
```
matthchr, line 20: Move this back into the config package

v2/internal/util/match/string_matcher.go
matthchr: Notes: What I have is pretty good but 1 more change which is to return a result type rather than a bool directly, and avoid 2x methods.

matthchr: Fixed

v2/internal/util/match/glob_matcher.go
matthchr: Changed this

v2/internal/util/match/string_matcher.go
matthchr: So it looks like, if the field is not specified at all in YAML, the `FieldMatcher` YAML deserialization override isn't even called. Actual just becomes a `nil` `StringMatcher` and we end up taking the `if dm.actual == nil {` codepath in `FieldMatcher`.

This means that there _is_ a difference between not specified and specified as `""` in the YAML. The more I think about it though, that feels like it might be a feature not a bug. Explicit "" attempting to match "" seems reasonable?

v2/internal/util/match/glob_matcher.go
```diff
@@ -78,20 +82,23 @@ func (gm *globMatcher) IsRestrictive() bool {
...
+
+func newGlobRegex(glob string) *regexp.Regexp {
+	g := regexp.QuoteMeta(glob)
 	g = strings.ReplaceAll(g, "\\*", ".*")
 	g = strings.ReplaceAll(g, "\\?", ".")
 	g = "(?i)(^" + g + "$)"
-	gm.regex = regexp.MustCompile(g)
+
+	// We use MustCompile here because QuotaMeta above removes almost all possibility of a panic.
+	// While it's technically still possible to panic it should be extremely rare.
```
theunrepentantgeek, line 97: 😁

v2/internal/util/match/literal_matcher.go
```diff
@@ -3,47 +3,76 @@
...
+		MatchingPattern: "",
 	}
 }
 
-// Matches returns true if the passed value is a case-insensitive match with our configured literal
-func (lm *literalMatcher) Matches(value string) bool {
+// MatchesDetailed returns true if the provided value is matched by the matcher
+// If there was a match, the second return value contains the pattern that matched.
+// If there was not a match, the second return value contains ""
+func (lm *literalMatcher) MatchesDetailed(value string) (bool, string) {
```
theunrepentantgeek, line 59: Isn't `Matches()` now already returning this information?

v2/internal/util/match/multi_matcher_test.go
```diff
@@ -15,8 +15,9 @@ func TestMultiMatcher_GivenDefinition_MatchesExpectedStrings(t *testing.T) {
 	t.Parallel()
 
 	type expectation struct {
-		value string
-		match bool
+		value                   string
+		match                   bool
+		expectedMatchingPattern string
```
theunrepentantgeek, line 20: With factories for `Result`, could combine these into a single field, and the expectations could use the factories.

v2/internal/util/match/string_matcher.go
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package match
+
+import (
+	"fmt"
+)
+
+type Result struct {
+	Matches         bool
```
theunrepentantgeek, line 13: Nitpick on naming.

```suggestion
	Matched         bool
```

(Since, by the time anyone reads the field, the value indicates whether something matched in the past.)

theunrepentantgeek (approved):
Looks good. Still need to remove `MatchesDetailed()` and I recommend renaming `Result.Matches` to `Result.Matched`. My other suggestions are more stylistic, feel free to ignore and get this merged.

theunrepentantgeek (approved):
Looks good. Still need to remove `MatchesDetailed()` and I recommend renaming `Result.Matches` to `Result.Matched`. My other suggestions are more stylistic, feel free to ignore and get this merged.

v2/internal/util/match/literal_matcher.go
matthchr: Doh, yes. I missed deleting this method, good catch

v2/internal/util/match/glob_matcher.go
```diff
@@ -3,62 +3,66 @@
...
 			// First time we match, clear out our candidates as we won't be needing them
 			gm.matched = true
 			gm.candidates.Clear()
 		}
 
-		return true
+		return Result{
+			Matches:         true,
+			MatchingPattern: gm.glob,
+		}
```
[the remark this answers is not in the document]
matthchr: Done

v2/internal/util/match/literal_matcher.go
```diff
@@ -3,47 +3,76 @@
...
+
+	if !lm.matched {
+		// Still collecting potential suggestions
+		lm.advisor.AddTerm(value)
+	}
+
+	return Result{
+		Matches:         false,
+		MatchingPattern: "",
 	}
```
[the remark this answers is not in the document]
matthchr: Swapped to `matchNotFound`, but otherwise done.

v2/internal/util/match/multi_matcher_test.go
matthchr: Done

Issue #2989, opened 2023-05-18
Bug: Helm chart not deploy all CRDs available


javierdompablo:
**Version of Azure Service Operator**
mcr.microsoft.com/k8s/azureserviceoperator:v2.0.0,gcr.io/kubebuilder/kube-rbac-proxy:v0.13.1

**Describe the bug**
I am trying to deploy the DnsZonesCNAMERecord object that is found from version v1api20180501 but the operator does not install that version of crds

**To Reproduce**

❯ kubectl api-versions | grep network.azure.com
network.azure.com/v1api20180901
network.azure.com/v1api20180901storage
network.azure.com/v1api20200601
network.azure.com/v1api20200601storage
network.azure.com/v1api20201101
network.azure.com/v1api20201101storage
network.azure.com/v1api20220701
network.azure.com/v1api20220701storage
network.azure.com/v1beta20180901
network.azure.com/v1beta20180901storage
network.azure.com/v1beta20201101
network.azure.com/v1beta20201101storage

I am applying the repository example to create this object

``` yaml
apiVersion: network.azure.com/v1api20180501
kind: DnsZonesCNAMERecord
metadata:
  name: aso-sample-record-cname
  namespace: default
spec:
  CNAMERecord:
    cname: asosample.com
  owner:
    name: aso-sample-dz.com
  TTL: 3600
```

and the error says you don't have this resource

``` bash
error: resource mapping not found for name: "aso-sample-record-cname" namespace: "default" from "DNSrecordError.yaml": no matches for kind "DnsZonesCNAMERecord" in version "network.azure.com/v1api20180501"
ensure CRDs are installed first
```

**Expected behavior**
The operator install al the API version available

matthchr:
This is actually a documentation issue more than a CRD install issue. We've added `DNSZoneCNAME` in `main` but it's not released yet. You can sorta see this alluded to on [this page](https://azure.github.io/azure-service-operator/reference/#network) in the "Supported From" column.

Note that for that resource, "Supported From" says: "v2.1.0", which is actually not released yet.

I realize that's confusing though... we should see if we can come up with a way to make it a bit clearer. cc @theunrepentantgeek  and @super-harsh.

v2.1.0 is tentatively planned for release sometime next week, at which point that above CRDs should be available.

theunrepentantgeek:
A quick update - we've reorganized our resource index pages to more clearly highlight resources that are new in the next release vs those that are already available.

[...]

PR #3000, opened 2023-05-22
Split resources indexes into groups

### LINKED ISSUE #2997: Improve clarity of documentation for unreleased resources

We list resources in our documentation as soon as the relevant PR is merged, well before any release of ASO includes it.
This has resulted in some users attempting, and failing, use the resource. 
See #2989 for one example - it surely has affected other users who simply didn't report it.

**Describe the improvement**
Improve our documentation to make it clearer that the resource hasn't yet been released.

theunrepentantgeek:

As the number of supported resources grows, the tables of resources in each group are becoming unwieldy. 

We're also discovering that some users are trying to make use of unreleased resources (e.g. #2989), highlighting that our documentation is not sufficiently clear. 

This PR splits the list of resources for each group into (up to) three sections: _Next Release_, _Released_, and _Deprecated_, with suitable commentary under each.

Closes #2997 

Wording is split out into fragments under `/docs/v2/azure/supported-resources` to make it easy to update.

gif

**If applicable**:
- [x] this PR contains tests


v2/tools/generator/internal/codegen/pipeline/report_resource_versions.go
```diff
@@ -307,42 +298,115 @@ func (report *ResourceVersionsReport) WriteGroupResourcesReportToBuffer(
...
+	// Pull out any deprecated resources
+	deprecatedResources := releasedResources.Where(report.isDeprecatedResource)
+	releasedResources = releasedResources.Except(deprecatedResources)
+
+	// Pull out any prerelease resources
+	prereleaseResources := releasedResources.Where(report.isUnreleasedResource)
+	releasedResources = releasedResources.Except(prereleaseResources)
+
+	// First list prerelease reources, if any
+	if len(prereleaseResources) > 0 {
```
super-harsh, line 321: can consider moving `len(resources) > 0` into the `writeSection` method

super-harsh (approved):

v2/tools/generator/internal/codegen/pipeline/report_resource_versions.go
theunrepentantgeek: Fair call. Done.

docs/v2/azure/supported-resources/deprecated.md
```diff
@@ -0,0 +1 @@
+These resource versions are deprecated and will be removed in an upcoming ASO release. Migration to newer versions is advised. See [Breaking Changes](https://azure.github.io/azure-service-operator/guide/breaking-changes/) for more information.
```
matthchr, line 1: Where are these files displayed? They seem pretty bare.

matthchr: Or are they just injected into the documentation but not actually displayed themselves? If so that makes sense.

v2/azure-arm.yaml
```diff
@@ -31,6 +31,8 @@ supportedResourcesReport:
   # resourcePath is a template used for generating a file path for checking whether docs for a resource have been generated
   # specified relative to the directory of outputPath, and supports the same placeholders
   resourcePathTemplate: "{group}/{version}.md"
+  # latestRelease is the version of our current release (resources introduced in newer releases will be shown as unreleased)
+  latestRelease: v2.0.0
```
matthchr, line 35: The version is injected into each binary as part of the build process with this cmd from taskfile:
`-ldflags "-X {{.PACKAGE}}/internal/version.BuildVersion={{.VERSION}}"`

You should be able to read the `BuildVersion` variable, which will be of the form `v2.0.0-rev.72`. If you just cut off the "additional stuff" after the `-`, you'll have the most recent release.

That way you don't need to hard-code this, it can be determined dynamically.

v2/tools/generator/internal/config/supported_resources_report.go
```diff
@@ -22,6 +22,8 @@ type SupportedResourcesReport struct {
 	// ResourcePathTemplate is a template used for generating a file path for checking whether docs for a resource have been generated
 	// specified relative to the directory of outputPath
 	ResourcePathTemplate string `yaml:"resourcePathTemplate"`
+	// LatestRelease is the version of the current release. Newer resources will be shown in a separate section.
+	LatestRelease string `yaml:"latestRelease"`
```
matthchr, line 26: As mentioned, this is unneeded you should be able to use the `BuildVersion` variable. Just add a helper in that `internal/version` library that gets the latest released version by parsing the version.

docs/v2/azure/supported-resources/deprecated.md
theunrepentantgeek: This directory contains fragments that are injected into the generated files where specified.

v2/azure-arm.yaml
super-harsh: Good point! This way we don't have to hardcode the version manually on each release.

theunrepentantgeek: I tried this out - but the variable is only defined when `generator` is build by the `Taskfile` - so it's not available when running tests (e.g. the tests that the report works, nor if you're running the generator by hand. I'm open to making this change, but there seem to be a number of thorns, so I'll push it out to another PR.

theunrepentantgeek: Completed in this PR. Uses `version.BuildVersion` if available, falls back to using `git` commands if not.

PR #3007, opened 2023-05-23
Allow users to configure a subset of CRDs to install

### LINKED ISSUE #1433: Install Selected CRDs only by default
Instead of blithely installing all custom resource definitions (CRDs) when the operator is deployed, we should have an approach that allows users to install only the CRDs they need.

This will help to avoid any performance issues that might be caused by installing hundreds of CRDs by only installing the dozen or so needed for a specific context.

Suggested approach is some kind of configuration file that gives control. Some CRDs will be enabled by default, but most will be disabled.

We likely would want some dependency information included - e.g. if you want to use SQL Server, you probably want Networking and AAD as well.

### LINKED ISSUE #2920: kube-apiserver persistently high memory usage with large number of CRDs

matthchr:
Fixes #1433.
Fixes #2920.

* By default, no CRDs are installed.
* If no CRDs are installed, the operator pod will exit with an error stating that there are no CRDs.
* Operator pod --crd-pattern command-line argument will now accept more than '*'.
* --crd-pattern means NEW CRDs to install. Existing CRDs in the cluster will always be upgraded. This means that upgrading an existing ASO installation without specifying any new CRDs will upgrade all of the existing CRDs and install no new CRDs.

Documentation updates will come in a separate PR.

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/charts/azure-service-operator/values.yaml
+++ b/v2/charts/azure-service-operator/values.yaml
@@ -92,6 +92,25 @@ metrics:
 # the operator will only reconcile that subset if those are the only CRDs it finds when the pod starts.
 installCRDs: true
 
+# crdPattern is a semicolon delimited string containing the CRD patterns for the operator to install.
+# Setting this has no effect if installCRDs is false.
+# This defines what new CRDs will be installed by ASO. Will always upgrade any existing CRDs even if no
+# crdPattern is defined. Leaving this field unspecified for a fresh install of ASO will result in the operator pod
+# exiting with an error saying no CRDs are configured. Leaving this field unspecified during an upgrade of ASO preserves
+# the existing set of CRDs already in the cluster. The existing CRDs will be upgraded to the latest version but no
+# new CRDs will be installed.
+# Values can be globs utilizing * or ?. The pattern is compared against the "{group}/{kind}" string of each CRD.
+# Patterns are case-insensitive.
+# Example: "resources.azure.com/*" would match the "resources.azure.com/ResourceGroup" resource.
+# Example: "compute.azure.com/*" would match all compute CRDs
+# Example: "resources.azure.com/*;compute.azure.com/*" would match all resources.azure.com resources as well as all
+# compute resources.
+# We strongly recommend including entire groups such as "dbformysql.azure.com/*". Individual CRDs such as
+# "dbformysql.azure.com/FlexibleServer" can be listed, but there are often other resources in the group which pair
+# together to enable other scenarios, such as dbformysql.azure.com/FlexibleServersFirewallRules, and it's generally
+# easier to just include the whole group.
+crdPattern: ''
+
 # podAnnotations contain the pod annotations for Azure Service Operator
 podAnnotations: {}
 
--- a/v2/cmd/controller/main.go
+++ b/v2/cmd/controller/main.go
@@ -26,6 +26,8 @@ func main() {
 		os.Exit(1)
 	}
 
+	setupLog.Info("Launching with flags", "flags", flgs.String())
+
 	if flgs.PreUpgradeCheck {
 		err = app.SetupPreUpgradeCheck(ctx)
 		if err != nil {
--- a/v2/internal/crdmanagement/crd_installation_instruction.go
+++ b/v2/internal/crdmanagement/crd_installation_instruction.go
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package crdmanagement
+
+import (
+	"fmt"
+
+	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+)
+
+type DiffResult string
+
+const (
+	NoDifference     = DiffResult("NoDifference")
+	SpecDifferent    = DiffResult("SpecDifferent")
+	VersionDifferent = DiffResult("VersionDifferent")
+)
+
+func (i DiffResult) DiffReason(crd apiextensions.CustomResourceDefinition) string {
+	switch i {
+	case NoDifference:
+		return fmt.Sprintf("No difference between existing and goal CRD %q", makeMatchString(crd))
+	case SpecDifferent:
+		return fmt.Sprintf("The spec was different between existing and goal CRD %q", makeMatchString(crd))
+	case VersionDifferent:
+		return fmt.Sprintf("The version was different between existing and goal CRD %q", makeMatchString(crd))
+	default:
+		return fmt.Sprintf("Unknown DiffResult %q", i)
+	}
+}
+
+type FilterResult string
+
+const (
+	MatchedExistingCRD = FilterResult("MatchedExistingCRD")
+	MatchedPattern     = FilterResult("MatchedPattern")
+	Excluded           = FilterResult("Excluded")
+)
+
+type CRDInstallationInstruction struct {
+	CRD apiextensions.CustomResourceDefinition
+
+	// FilterResult contains the result of if the CRD was considered for installation or not
+	FilterResult FilterResult
+	// FilterReason contains a user-facing reason for why the CRD was or was not considered for installation
+	FilterReason string
+
+	// DiffResult contains the result of the diff between the existing CRD (if any) and the goal CRD. This may
+	// be NoDifference if the CRD was filtered from consideration before the diff phase.
+	DiffResult DiffResult
+}
+
+func (i *CRDInstallationInstruction) ShouldApply() (bool, string) {
+	excluded := i.FilterResult == Excluded
+	if excluded {
+		return false, i.FilterReason
+	}
+
+	isSame := i.DiffResult == NoDifference
+	if isSame {
+		return false, i.DiffResult.DiffReason(i.CRD)
+	}
+
+	return true, i.DiffResult.DiffReason(i.CRD)
+}
+
+func IncludedCRDs(instructions []*CRDInstallationInstruction) []*CRDInstallationInstruction {
+	// Prealloc false positive: https://github.com/alexkohler/prealloc/issues/16
+	//nolint:prealloc
+	var result []*CRDInstallationInstruction
+	for _, instruction := range instructions {
+		if instruction.FilterResult == Excluded {
+			continue
+		}
+		result = append(result, instruction)
+	}
+
+	return result
+}
--- a/v2/internal/crdmanagement/manager.go
+++ b/v2/internal/crdmanagement/manager.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
+	"golang.org/x/exp/maps"
 	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -23,6 +24,7 @@ import (
 
 	. "github.com/Azure/azure-service-operator/v2/internal/logging"
 	"github.com/Azure/azure-service-operator/v2/internal/util/kubeclient"
+	"github.com/Azure/azure-service-operator/v2/internal/util/match"
 )
 
 // ServiceOperatorVersionLabel is the label the CRDs have on them containing the ASO version. This value must match the value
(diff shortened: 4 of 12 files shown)


v2/charts/azure-service-operator/values.yaml
```diff
@@ -92,6 +92,25 @@ metrics:
...
+# Values can be globs utilizing * or ?. The pattern is compared against the "{group}/{kind}" string of each CRD.
+# Patterns are case-insensitive.
+# Example: "resources.azure.com/*" would match the "resources.azure.com/ResourceGroup" resource.
+# Example: "compute.azure.com/*" would match all compute CRDs
+# Example: "resources.azure.com/*;compute.azure.com/*" would match all resources.azure.com resources as well as all
+# compute resources.
+# We strongly recommend including entire groups such as "dbformysql.azure.com/*". Individual CRDs such as
+# "dbformysql.azure.com/FlexibleServer" can be listed, but there are often other resources in the group which pair
+# together to enable other scenarios, such as dbformysql.azure.com/FlexibleServersFirewallRules, and it's generally
+# easier to just include the whole group.
```
theunrepentantgeek, line 111: Does this also need to include a link to explain why we don't just install everything?

v2/cmd/controller/main.go
```diff
@@ -26,6 +26,8 @@ func main() {
 		os.Exit(1)
 	}
 
+	setupLog.Info("Launching with flags", "flags", flgs.String())
```
theunrepentantgeek, line 29: Is this one of those loggers that accepts name/value pairs? if so, maybe we want to log the parameters individually.

v2/internal/crdmanagement/crd_installation_instruction.go
```diff
@@ -0,0 +1,82 @@
...
+type CRDInstallationInstruction struct {
+	CRD apiextensions.CustomResourceDefinition
+
+	// FilterResult contains the result of if the CRD was considered for installation or not
+	FilterResult FilterResult
+	// FilterReason contains a user-facing reason for why the CRD was or was not considered for installation
+	FilterReason string
+
+	// DiffResult contains the result of the diff between the existing CRD (if any) and the goal CRD. This may
+	// be nil if the CRD was filtered from consideration before the diff phase.
```
theunrepentantgeek, line 39:
```suggestion
	// be empty if the CRD was filtered from consideration before the diff phase.
```

v2/internal/crdmanagement/crd_installation_instruction.go
```diff
@@ -0,0 +1,82 @@
...
+	case NoDifference:
+		return fmt.Sprintf("No difference between existing and goal CRD %q", makeMatchString(i.CRD))
+	case SpecDifferent:
+		return fmt.Sprintf("The spec was different between existing and goal CRD %q", makeMatchString(i.CRD))
+	case VersionDifferent:
+		return fmt.Sprintf("The version was different between existing and goal CRD %q", makeMatchString(i.CRD))
+	default:
+		return fmt.Sprintf("Unknown DiffResult %q", i.DiffResult)
+	}
+}
```
theunrepentantgeek, line 68: You could make this a method on `DiffResult` directly and avoid the type-envy smell:

```suggestion
func (d DiffResult) Reason(crd string) string {
	switch d {
	case NoDifference:
		return fmt.Sprintf("No difference between existing and goal CRD %q", crd)
	case SpecDifferent:
		return fmt.Sprintf("The spec was different between existing and goal CRD %q", crd)
	case VersionDifferent:
		return fmt.Sprintf("The version was different between existing and goal CRD %q", crd)
	default:
		return fmt.Sprintf("Unknown DiffResult %q", d)
	}
}
```

v2/internal/crdmanagement/crd_installation_instruction.go
```diff
@@ -0,0 +1,82 @@
...
+
+func IncludedCRDs(instructions []*CRDInstallationInstruction) []*CRDInstallationInstruction {
+	// Prealloc false positive: https://github.com/alexkohler/prealloc/issues/16
+	//nolint:prealloc
+	var result []*CRDInstallationInstruction
```
theunrepentantgeek, line 73: Is this really a false positive? Why not just allocate sufficient capacity?

```suggestion
	result := make([]*CRDInstallationInstruction, 0, len(instructions))
```

v2/charts/azure-service-operator/values.yaml
```diff
@@ -92,6 +92,25 @@ metrics:
...
 installCRDs: true
 
+# crdPattern is a semicolon delimited string containing the CRD patterns for the operator to install.
+# Setting this has no effect if installCRDs is false.
+# This defines what new CRDs will be installed by ASO. Will always upgrade any existing CRDs even if no
+# crdPattern is defined. Leaving this field unspecified for a fresh install of ASO will result in the operator pod
+# exiting with an error saying no CRDs are configured. Leaving this field unspecified during an upgrade of ASO preserves
+# the existing set of CRDs already in the cluster. The existing CRDs will be upgraded to the latest version but no
+# new CRDs will be installed.
+# Values can be globs utilizing * or ?. The pattern is compared against the "{group}/{kind}" string of each CRD.
```
theunrepentantgeek, line 102: What about making `/*` optional and allowing a pattern that's just `{group}` as well?

v2/internal/crdmanagement/manager.go
```diff
@@ -154,70 +155,142 @@ func (m *Manager) FindNonMatchingCRDs(
...
+		result.DiffResult = VersionDifferent
 	}
 
-	if len(crdsToApply) > 0 {
-		m.logger.V(Status).Info("Will apply CRDs", "count", len(crdsToApply))
+	// Collapse result to a slice
+	results := make([]*CRDInstallationInstruction, 0, len(resultMap))
+	for _, result := range resultMap {
+		results = append(results, result)
+	}
```
theunrepentantgeek, line 223: Use `maps.Values()`.

v2/internal/crdmanagement/manager.go
```diff
@@ -154,70 +155,142 @@ func (m *Manager) FindNonMatchingCRDs(
...
+		i += 1
+		toApply := &apiextensions.CustomResourceDefinition{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: instruction.CRD.Name,
+			},
+		}
+		m.logger.V(Verbose).Info(
+			"Applying CRD",
+			"Progress", fmt.Sprintf("%d/%d", i, len(instructionsToApply)),
+			"crd", instruction.CRD.Name)
```
theunrepentantgeek, line 272: Names should both Capitalized or both lowercase.

```suggestion
			"progress", fmt.Sprintf("%d/%d", i, len(instructionsToApply)),
			"crd", instruction.CRD.Name)
```

theunrepentantgeek (approved):
Really good!

v2/charts/azure-service-operator/values.yaml
matthchr: I think there's value in it being "a glob" because people at least in theory understand that from shell-days. If we did this it would be "a glob but with some exceptions"

Maybe if there's actually user confusion or complaints about this we could revisit? I think for now though the simpler "it's a glob" has the advantage of being both consistent and also implemented in this PR. We would need to change things up possibly significantly in order to allow non-glob shortcuts as it's not clear what `matcher` would make sense with them, so we'd probably have to filter them out first and then handle them outside of the `matcher` context?

v2/internal/crdmanagement/manager.go
matthchr: I don't know how the pedant in me missed this, usually this sort of thing is super obvious (if obviously pretty minor). Thanks for the catch.

v2/charts/azure-service-operator/values.yaml
matthchr: Yeah probably. I'll update it as part of the docs PR I am going to send tomorrow. Right now there isn't a great doc to link to (could link directly to the issue I suppose).

v2/charts/azure-service-operator/values.yaml
```diff
@@ -92,6 +92,25 @@ metrics:
...
+# Patterns are case-insensitive.
+# Example: "resources.azure.com/*" would match the "resources.azure.com/ResourceGroup" resource.
+# Example: "compute.azure.com/*" would match all compute CRDs
+# Example: "resources.azure.com/*;compute.azure.com/*" would match all resources.azure.com resources as well as all
```
super-harsh, line 106: Not sure if this makes sense -  Do we need some sort of version filtering as well?

v2/internal/crdmanagement/manager.go
```diff
@@ -154,70 +155,142 @@ func (m *Manager) FindNonMatchingCRDs(
 	return m.FindMatchingCRDs(existing, goal, invertedComparators...)
 }
 
-func (m *Manager) ApplyCRDs(
-	ctx context.Context,
+// DetermineCRDsToInstallOrUpgrade examines the set of goal CRDs and installed CRDs to determine the set which should
+// be installed or upgraded.
+func (m *Manager) DetermineCRDsToInstallOrUpgrade(
```
super-harsh, line 160: Looks good, the only thing I can think of is the performance issues and delayed pod start-up due to a lot of checks and loops, which is gonna increase as we plan to add more resources. One of the improvements I can think of is, rather than storing all the CRDs by name in a Map, we can have something like `groupString:[]CRDs` map to avoid looping through each CRD and we can filter out/exclude the groups totally without iterating through each CRD, if not specified at all. The above should be good for now, and further we can do this filtering per group in parallel using semaphores.

v2/internal/crdmanagement/manager_test.go
```diff
@@ -260,6 +224,108 @@ func Test_FindNonMatchingCRDs_CRDsWithDifferentConversionsCompareAsEqual(t *test
...
+				g.Expect(instructions[0].DiffResult).To(Equal(crdmanagement.VersionDifferent))
+				apply, _ := instructions[0].ShouldApply()
+				g.Expect(apply).To(BeTrue())
+			},
+		},
+		{
+			name:     "Pattern matches but CRD already installed, nothing to do",
+			goal:     []apiextensions.CustomResourceDefinition{makeBasicCRD("test")},
+			existing: []apiextensions.CustomResourceDefinition{makeBasicCRD("test")},
+			patterns: "testrp.azure.com/*",
```
super-harsh, line 300: Consider adding another test with 2 CRDs in a group where `pattern == testrp.azure.com/test` and the other one is ignored

v2/charts/azure-service-operator/values.yaml
theunrepentantgeek: Can't do version filtering. The conversion of versions to/from the hub version requires all versions to be available - if one is missing, the chain is broken:

image

We could potentially make changes to do this, but I'm not sure it's worth doing so.

v2/internal/crdmanagement/manager.go
theunrepentantgeek: What's the actual runtime of this check? 
If it's on the order of milliseconds, I think we're fine. If it's a few seconds, we need to keep an eye on it - and if it's minutes we really need to fix. I suspect it's a few seconds at most.

v2/charts/azure-service-operator/values.yaml
matthchr: I've made a note of this and will fix in a subsequent PR

v2/internal/crdmanagement/manager.go
matthchr: Done

v2/internal/crdmanagement/manager_test.go
matthchr: Good idea. Added

v2/internal/crdmanagement/crd_installation_instruction.go
matthchr: It is a false positive, see: https://github.com/alexkohler/prealloc/issues/16

The way the linter is supposed to work, is that if you have a `continue` in your loop, it doesn't fail the linter, because that means that the size of the collection is not equal to the size of the input collection.

v2/charts/azure-service-operator/values.yaml
matthchr: Version filtering is hard as well because the versions are _in_ the CRD so we would have to actually modify the CRD contents. That's significantly harder than what we're doing now which is just filtering specific CRDs in or out.

v2/cmd/controller/main.go
matthchr: Done

matthchr: Actually I take it back: I tried to do this, but actually making it work with the logger is quite awkward.

You need a method like this:
```
func (f Flags) Values() []any {
	return []any{
		"MetricsAddr", f.MetricsAddr,
		"HealthAddr", f.HealthAddr,
		"EnableLeaderElection", "%t", f.EnableLeaderElection,
		"CRDPatterns", f.CRDPatterns,
		"PreUpgradeCheck", "%t", f.PreUpgradeCheck,
	}
}
```

and then you also _must_ use it like:
`setupLog.Info("Launching with flags", flgs.Values()...)`

if you miss the `...` trailing, there's a runtime panic.

Given the above, I've moved back to logging it as a string like we do for `config.Vars`. That's consistent and it avoids the risk of programmer mistake leading to panic.

v2/internal/crdmanagement/manager.go
matthchr: The actual breakdown of time from a local run on a kind cluster is:
Loading CRDs from filesystem: ~3s
Listing CRDs from k8s APIServer: ~0.5s
Filtering CRDs locally: ~0.5s
Applying CRDs: ~14s

Total time: ~18s, of which this PR adds at most ~0.5s.

theunrepentantgeek: If anything, this PR is more likely to reduce startup time by skipping any CRDs that are both uninstalled and unrequested.

theunrepentantgeek: That said, 14s to apply CRDs is a relatively long time. Is that single-threaded? If not, maybe we should look at introducing some concurrency.

PR #3022, opened 2023-05-29
Improve link checking for our docs

theunrepentantgeek:

We have been using [markdown-link-check](https://github.com/tcort/markdown-link-check) but it has some issues specific to our context:

* Doesn't natively handle the way Hugo transmogrifies things
    * We got close by configuring some replacement patterns, but didn't achieve full coverage
* Slow
    * This wouldn't be a problem for smaller sites, but ours is large

In this PR, we switch to using [htmltest ](https://github.com/wjdp/htmltest) to verify the final HTML that's generated, instead of trying to second-guess [Hugo](https://gohugo.io/). 

(Until [PR#215](https://github.com/wjdp/htmltest/pull/215) is merged, we're using a private fork.  We'll switch once a new version is released.)

Running `htmltest` identified 57 broken links; fixes for those are included in this PR.

We don't need to guess at the links generated by Hugo - the built in **relref** function allows lookup of a page by name and will generate the appropriate link. No path is needed if the name of the page is unique (and ours mostly are). In most cases you can also omit the file extension, but it seems to be necessary in some cases.

gif

**If applicable**:


.devcontainer/install-dependencies.sh
```diff
@@ -150,7 +150,7 @@ go-install kustomize sigs.k8s.io/kustomize/kustomize/v4@v4.5.7
 
 # for docs site
 go-install hugo -tags extended github.com/gohugoio/hugo@v0.88.1
-go-install htmltest github.com/wjdp/htmltest@v0.15.0
+go-install htmltest github.com/theunrepentantgeek/htmltest@latest
```
super-harsh, line 153: Leave a todo here to replace and reason

theunrepentantgeek: Done.

theunrepentantgeek: I also created #3024 so we don't lose track of the need to update.

Taskfile.yml
```diff
@@ -89,21 +89,8 @@ tasks:
...
-      NODE_PATH:
-        sh: npm root -g
-
-  check-docs-site-links:
-    desc: Verify documentation site links
-    dir: docs/hugo
-    cmds:
-      # Excluding './content/reference/*' path as it contains all auto-generated API docs.
-      - find . -type f -name "*.md" -not -path "./content/reference/*" -not -path "./node_modules/*" -print0 | xargs -0 -n1 markdown-link-check -c link-checker.json
+      - htmltest
```
super-harsh, line 93: how does this work? do we not require to provide any path?

theunrepentantgeek: The task runs in `docs/hugo` (see line 90) and there's a `.htmltest.yml` file in that folder that has the rest of the configuration.

docs/hugo/.htmltest.yml
```diff
@@ -1,3 +1,13 @@
 DirectoryPath: public
 CheckExternal: false
```
super-harsh, line 2: We do want to check the external links as well?

super-harsh (approved):
LGTM! just a few minors. Also, get rid of [link-checker.json](https://github.com/Azure/azure-service-operator/blob/main/docs/hugo/link-checker.json) file as we don;t use it anymore.

docs/hugo/.htmltest.yml
theunrepentantgeek: I'll do that in a separate PR.

PR #3029, opened 2023-05-30
Fix DevContainer Builds

theunrepentantgeek:

Previous attempt to upgrade from deprecated versions of the docker actions failed, so trying again.

It appears the _required_ use of `docker/setup-buildx-action@v2` ... isn't.

Manually tested by pushing the branch and triggering the workflow.

gif

### DIFF
--- a/.github/workflows/build-devcontainer-image.yml
+++ b/.github/workflows/build-devcontainer-image.yml
@@ -25,8 +25,8 @@ jobs:
       contents: read
 
     steps:
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v2
+      - name: Checkout code
+        uses: actions/checkout@v3
 
       - name: Log in to GitHub Docker Registry
         uses: docker/login-action@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/build-devcontainer-image.yml
```diff
@@ -25,8 +25,8 @@ jobs:
       contents: read
 
     steps:
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v2
+      - name: Checkout code
+        uses: actions/checkout@v3
```
matthchr, line 29: The issue was that somehow `setup-buildx-action` is actually not needed and breaks the `build-push-action` below?

matthchr (approved):

.github/workflows/build-devcontainer-image.yml
theunrepentantgeek: Seems to be. I reverted back to the version that worked prior to Apr 26 and made changes one by one to move it forward. 

According to the docs, the `setup-buildx-action` is optional, but recommended. Leaving it out seems to be the key though.

PR #3034, opened 2023-05-31
Fix secret override not applying to Status

matthchr:
This is a breaking change, although it's a pretty safe one since these fields are all never returned from the service anyway (so always empty string)

Removed the following Status properties, which were never set (always empty string):
MachineLearningServices:
- UserAccountCredentials_STATUS.AdminUserPassword
- UserAccountCredentials_STATUS.AdminUserSshPublicKey
- VirtualMachineSshCredentials_STATUS.Password

Synapse:
-  Workspace_STATUS.SqlAdministratorLoginPassword

**If applicable**:
- [ ] this PR contains tests


v2/tools/generator/internal/codegen/pipeline/add_secrets.go
```diff
@@ -59,6 +60,17 @@ func applyConfigSecretOverrides(config *config.Configuration, definitions astmod
 			}
 		}
 
+		// For secrets, setting this on the spec should also apply to the status
+		if typeName.IsStatus() {
```
super-harsh, line 64: Rather than re-iterating, should we look into merging this into the loop above with some branches?

super-harsh (approved):

v2/tools/generator/internal/codegen/pipeline/add_secrets.go
matthchr: Yeah, agreed that's better. Fixed.

PR #3041, opened 2023-06-01
Improve our documentation for Dev Setup

theunrepentantgeek:

The documentation lacked clarity and had some errors.

For example:
> **Note**: in Windows, it is important to clone directly into a container instead of cloning first and then loading that with the `Remote Containers` extension, as the tooling performs a lot of file I/O, and if this is performed against a volume mounted in WSL then it is unusably slow.

The performance issues occur if you mount a Windows folder as a volume in the devcontainer, as the folder gets exposed as a network share and things are slow.

If you instead mount a Linux folder from inside WSL as a volume in the devcontainer, performance is just fine. 

The above note conflates the two scenarios and is misleading.

Our docs also had some information that wasn't consistently repeated across potential environments.

I've broken the Dev Setup stuff out into a separate page.

gif

**If applicable**:


docs/hugo/content/contributing/_index.md
```diff
@@ -11,56 +11,6 @@ cascade:
 description: "How to contribute new resources to Azure Service Operator v2"
 ---
 
-## Related pages
```
matthchr, line 14: I feel like we should somehow be linking or highlighting the developer setup page here at least? It's IMO the most important page for people to see. More important even than Directory structure of the operator (which is now the top thing here on the index).

Wondering if we should split the contributing section up entirely and move the other bits to subpages as well and just have index be a TOC-esque thing that links to the various sub-pages?

theunrepentantgeek: I've restored the `Related Pages` section, for now. 

It feels out of place to me, especially given we already have a list of related pages in the lefthand sidebar, plus a table of content in the righthand sidebar. That said, I think we need to restructure this area and I'm happy to leave the section in until we do that.

super-harsh (approved):
LGTM

docs/hugo/content/contributing/developer-setup.md
```diff
@@ -0,0 +1,174 @@
...
+
+## CLI on Linux
+
+If you are using Linux, instead of using VS Code you can run the `dev.sh` script in the root of the repository. This will install all required tooling into the `hack/tools` directory and then start a new shell with the `PATH` updated to use it.
+
+## CLI on MacOS
+
+Development of ASO on MacOS is possible (one of our team does so), but things are less automated.
+
+You'll need to manually install the tools as listed by `.devcontainer/install-dependencies.sh`.
```
super-harsh, line 85: If we don't have it already - Should we have a list of tools/dependencies somewhere? 

Just makes it easier for the users, so that they don't have to dig through the `.sh` files

theunrepentantgeek: Good idea. I'll log a separate task to track doing this.

PR #3048, opened 2023-06-06
Allow PropertyConversion to return errors

### LINKED ISSUE #2967: PropertyConversion should allow for returning error

The func type `conversions.PropertyConversion` doesn't allow for implementations to return an error.

**Describe the improvement**

**Additional context**

theunrepentantgeek:

One of the stepping stones required for us to move away from a `panic()` when predictable things go wrong to returning useful errors, this PR modifies the signature of `conversions.PropertyConversion` to also return an error.

Closes #2967 

gif

**If applicable**:
- [x] this PR contains tests


v2/tools/generator/internal/conversions/property_conversions.go
```diff
@@ -252,7 +253,12 @@ func writeToBagItem(
 	_, sourceIsMap := astmodel.AsMapType(actualSourceType)
 	_, sourceIsSlice := astmodel.AsArrayType(actualSourceType)
 
-	return func(reader dst.Expr, _ func(dst.Expr) []dst.Stmt, knownLocals *astmodel.KnownLocalsSet, generationContext *astmodel.CodeGenerationContext) []dst.Stmt {
+	return func(
+		reader dst.Expr,
+		_ func(dst.Expr) []dst.Stmt,
+		knownLocals *astmodel.KnownLocalsSet,
+		generationContext *astmodel.CodeGenerationContext,
+	) ([]dst.Stmt, error) {
```
super-harsh, line 261: Not sure if its right to return nil as an error entirely throughout the function, but seems like we have to do it to make the function in line with the function type.

super-harsh (approved):

v2/tools/generator/internal/conversions/property_conversions.go
theunrepentantgeek: Yeah, it's a common wart, caused when only some implementations of an interface need to return errors.

PR #3052, opened 2023-06-06
Fix redis samples

super-harsh:

Found out while scenario testing, `RedisLinkedServer` has `location` property on it which should not be there.

### DIFF
--- a/v2/samples/cache/v1api/v1api20201201_redislinkedserver.yaml
+++ b/v2/samples/cache/v1api/v1api20201201_redislinkedserver.yaml
@@ -11,5 +11,5 @@ spec:
   linkedRedisCacheReference:
     group: cache.azure.com
     kind: Redis
-    name: linked-sampleredis
+    name: linked-sample-redis
   serverRole: Secondary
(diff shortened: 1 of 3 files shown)


v2/samples/cache/v1api/v1api20201201_redislinkedserver.yaml
```diff
@@ -2,15 +2,14 @@ apiVersion: cache.azure.com/v1api20201201
 kind: RedisLinkedServer
 metadata:
   # This name needs to be the same as the secondary redis server.
-  name: sampleredis2-alpha
+  name: linked-sampleredis
```
theunrepentantgeek, line 5: What's the length limit for these names? If there's room, adding an additional separator would make the easier to read.

```suggestion
  name: linked-sample-redis
```

v2/samples/cache/v1api/v1api20201201_redislinkedserver.yaml
```diff
@@ -2,15 +2,14 @@ apiVersion: cache.azure.com/v1api20201201
...
 spec:
-  location: westcentralus
   owner:
     name: sampleredis1
   linkedRedisCacheLocation: westcentralus
   linkedRedisCacheReference:
     group: cache.azure.com
     kind: Redis
-    name: sampleredis2-alpha
+    name: linked-sampleredis
```
theunrepentantgeek, line 14:
```suggestion
    name: linked-sample-redis
```

theunrepentantgeek (approved):
A suggestion on naming, but otherwise looks good.

theunrepentantgeek (approved):
A suggestion on naming, but otherwise looks good.

PR #3055, opened 2023-06-08
Include hand crafted resources in documentation indexes

### LINKED ISSUE #3054: Bug: Hand coded resources are omitted from CRD Documentation
**Describe the bug**

Hand written resources - like users for PostgreSQL databases - aren't included in our CRD documentation; they should be.

**To Reproduce**

Look at the [documentation for PostgreSQL](https://azure.github.io/azure-service-operator/reference/dbforpostgresql/) and observe that only API versions are listed. The `v1` version used for the hand-written resources does not appear.

**Expected behavior**

All resources should be documented, regardless of whether they are generated or hand-written.

**Screenshots**

image

theunrepentantgeek:

Our index pages need to include any hand-crafted types that we add to ASO; without them, the documentation (which we _are_ generating) can't be easily found.

Closes #3054 

gif

**If applicable**:
- [x] this PR contains tests


v2/azure-arm.yaml
```diff
@@ -938,6 +938,9 @@ objectModelConfiguration:
         $exportAs: Database
         $supportedFrom: v2.0.0-beta.1
   dbformysql:
+    v1: # Handcrafted
```
matthchr, line 941: Do we have a way to detect if we've forgotten to add this?

v2/tools/generator/internal/codegen/pipeline/report_resource_versions.go
```diff
@@ -98,19 +105,23 @@ func NewResourceVersionsReport(
...
 
 	err = result.loadFragments()
 	if err != nil {
 		return nil, errors.Wrapf(err, "Unable to load report fragments")
 	}
 
-	result.summarize(definitions)
+	err = result.summarize(definitions)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Unable to summarize resources")
```
matthchr, line 122: super super minor: I think Go prefers errs not uppercase first letter. Who knows why...

v2/tools/generator/internal/codegen/pipeline/report_resource_versions.go
```diff
@@ -521,10 +543,21 @@ func (report *ResourceVersionsReport) FindSampleLinks(group string) (map[string]
...
+		name:          name,
+		armType:       armType,
+		armVersion:    armVersion,
+		supportedFrom: report.supportedFrom(name),
+	}
+}
 
-	name := rsrc.Name()
+// generateApiLink returns a link to the API definition for the given resource
+func (report *ResourceVersionsReport) generateApiLink(name astmodel.TypeName) string {
```
matthchr, line 560: minor: `API` preferred I think

matthchr (commented):
Overall LGTM. One question is if there's a way for us to detect that somebody has forgotten to include a handcrafted resource in the `azure-arm.yaml` file?

Doesn't necessarily have to be part of the code generator (in fact in some ways that'd be sorta weird). Maybe some way to write a taskfile target that verifies this?

The more I think about it the more I wonder if we should split the doc generation out into a different tool that does something based more on the filesystem (which would then work more easily for handcrafted resources as they also exist on the filesystem)... something to think about longer term possibly.

v2/azure-arm.yaml
theunrepentantgeek: I'm not sure at the moment. I'll give that some thought though.

v2/tools/generator/internal/codegen/pipeline/report_resource_versions.go
theunrepentantgeek: Fixed.

v2/tools/generator/internal/codegen/pipeline/report_resource_versions.go
theunrepentantgeek: Fixed.

matthchr (approved):

PR #3058, opened 2023-06-08
Fix bug preventing operator from installing CRDs into a different namespace

### LINKED ISSUE #3057: Bug: Invalid cert-manager annotation on CRD
v2.1.0

**Describe the bug**
When deploying to a namespace other than `azureserviceoperator-system`, the `cert-manager.io/inject-ca-from` annotation on CRDs contains the wrong namespace, causing webhook errors.

**To Reproduce**
* Deploy ASO to a namespace other than `azureserviceoperator-system` with CRD installation enabled
* Inspect the installed CRD resources
* The `cert-manager.io/inject-ca-from` annotation value is `azureserviceoperator-system/azureserviceoperator-serving-cert`

**Expected behavior**
The `cert-manager.io/inject-ca-from` annotation should reflect the namespace that ASO was installed in.

**Additional context**
The controller already tries to address this, but it is looking for a label named `cert-manager.io/inject-ca-from` instead of an annotation.

https://github.com/Azure/azure-service-operator/blob/3598bd628d55cb7ea47e4587e2fb1f4f5a9f33f3/v2/internal/crdmanagement/manager.go#L392-L400

matthchr:
We were fixing up the cert-manager label when really it's an annotation that should have been fixed.

- Adds a unit test.
- Fixes #3057.

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/internal/crdmanagement/manager.go
+++ b/v2/internal/crdmanagement/manager.go
@@ -390,11 +390,11 @@ func fixCRDNamespace(crd apiextensions.CustomResourceDefinition, namespace strin
 	}
 
 	// Set cert-manager.io/inject-ca-from
-	if len(result.Labels) > 0 {
-		if injectCAFrom, ok := result.Labels[certMgrInjectCAFromAnnotation]; ok {
+	if len(result.Annotations) > 0 {
+		if injectCAFrom, ok := result.Annotations[certMgrInjectCAFromAnnotation]; ok {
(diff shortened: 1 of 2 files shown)


theunrepentantgeek (approved):

matthchr:
Looks like we're having an issue with fork CI -- need to investigate what is going on. CI is actually passing but it's failing to report that it has done so

super-harsh (approved):

PR #3060, opened 2023-06-09
Opt-in to sending explicit payloads to ARM

### LINKED ISSUE #2914: Bug: Storage Account resource does not detect drift in networkAcls if the original is empty

theunrepentantgeek:

As identified in #2914, some Azure Resource Providers work more as a PATCH than a PUT, allowing unwarrented drift to occur without correction by ASO. 

This PR adds the option `$payloadType` so we can opt-in certain groups for being more explicit with the payloads they send to ARM. 

Closes #2914

I've left the door open for configuration of this at the object and property level, but given the behaviour we've seen so far, configuration at the group level seemed appropriate.

gif

**Copilot Summary**

copilot:summary

**Copilot Walkthrough**

copilot:walkthrough


theunrepentantgeek:
Let's see if CoPilot works ... 

copilot:summary

copilot:walkthrough

v2/azure-arm.yaml
```diff
@@ -527,7 +527,23 @@ status:
...
 #
-# Available 'type' modifiers:
+# ===================
+#   Group modifiers
+# ===================
+#
+# $payloadType: <string>
+#     Specifies the type of the payload to generate for the group. This is used to compensate for 
+#     some Azure Resource Providers using PATCH semantics instead of PUT semantics for updates.
+#     We control this at the group level because RPs tend to be consistent across all their resources.
```
super-harsh, line 537: ``
```suggestion
#     We control this at the group level because RPs tend to be consistent across all their resources and versions.
```

v2/tools/generator/internal/codegen/pipeline/create_arm_types.go
```diff
@@ -8,6 +8,8 @@ package pipeline
 import (
 	"context"
 
+	"github.com/Azure/azure-service-operator/v2/tools/generator/internal/config"
```
matthchr, line 11: minor: ordering, should be at the bottom with the other v2/tools imports.

v2/tools/generator/internal/config/property_configuration.go
```diff
@@ -232,14 +232,15 @@ func (pc *PropertyConfiguration) UnmarshalYAML(value *yaml.Node) error {
 
 		// $importConfigMapMode: <string>
 		if strings.EqualFold(lastId, importConfigMapModeTag) && c.Kind == yaml.ScalarNode {
-			switch c.Value {
+			switch strings.ToLower(c.Value) {
```
matthchr, line 235: Not sold on the value of accepting any case here.

I think it's better to force us to be consistently cased and return an error than to allow anything. With that said... I like your `UpperCamelCase` more than my `lowercase`, so possibly we could drop this case-insensitivity but fix this `importConfigMapMode` to accept `Optional|Required` rather than `optional|required`? Then it's forced to be consistent with your new enum (which IMO we should also not ignore-case on?)

matthchr: I'd be happy to do a follow up PR to fix this enums case if you didnt' want to do it here because it increased PR size.

v2/tools/generator/internal/config/group_configuration.go
```diff
@@ -169,6 +203,22 @@ func (gc *GroupConfiguration) UnmarshalYAML(value *yaml.Node) error {
 			continue
 		}
 
+		// $payloadType: <string>
+		if strings.EqualFold(lastId, payloadTypeTag) && c.Kind == yaml.ScalarNode {
```
matthchr, line 207: See comment below about uncertainty on value of being case-insensitive here.

v2/tools/generator/internal/codegen/pipeline/create_arm_types.go
```diff
@@ -448,6 +496,66 @@ func (c *armTypeCreator) convertObjectPropertiesForARM(t *astmodel.ObjectType, i
...
 }
 
+func (c *armTypeCreator) visitARMTypeName(this *astmodel.TypeVisitor, it astmodel.TypeName, ctx interface{}) (astmodel.Type, error) {
+	// Allow json type to pass through.
+	if it == astmodel.JSONType {
+		return it, nil
+	}
+
+	// Look up the definition
+	def, ok := c.definitions[it]
```
matthchr, line 506: minor: I think there's a `c.definitions.GetDefinition` that gets you a standard error here, rather than you having to craft it yourself.

v2/tools/generator/internal/codegen/pipeline/create_arm_types.go
```diff
@@ -375,17 +396,44 @@ func (c *armTypeCreator) createConfigMapReferenceProperty(prop *astmodel.Propert
...
 	newType, err := c.createARMTypeIfNeeded(prop.PropertyType())
 	if err != nil {
 		return nil, err
 	}
-	return prop.WithType(newType), nil
+
+	// Return a property with (potentially) a new type
+	result := prop.WithType(newType)
+
+	removeOmitEmpty := false
```
matthchr, line 411: very minor: I'm not sure if this would help readability at all, but would it make sense to put this removing omitEmpty bit into its own function?

happy if you think it won't, I'm on the fence.

matthchr (approved):

v2/tools/generator/internal/config/property_configuration.go
theunrepentantgeek: As discussed offline, we'll leave this as-is for the moment, as most of our configuration is already case-insensitive.

v2/tools/generator/internal/codegen/pipeline/create_arm_types.go
theunrepentantgeek: Fixed.

v2/tools/generator/internal/codegen/pipeline/create_arm_types.go
theunrepentantgeek: Thanks!

v2/tools/generator/internal/codegen/pipeline/create_arm_types.go
theunrepentantgeek: I don't think it would - but I restructured the code to be clearer another way.

PR #3076, opened 2023-06-14
Documentation fixes

### LINKED ISSUE #3067: Add current version to landing page

There's no indication on our [landing page](https://azure.github.io/azure-service-operator/) of the current version of ASO, nor any details of our history of releases, or a link to our GH project. These are details that people use to evaluate the health of a tool before adoption.

**Describe the improvement**

Add details of the current release (and possibly a couple prior releases) to the page, along with links to the GH project.

### LINKED ISSUE #3070: Bug: ASOv2 publishing workflow broken by automatic docs updates
**Describe the bug**
When publishing a new version of ASOv2, the release job runs into issues because resources move from "upcoming" to the "released" section. This causes a change in the docs that breaks CI and causes the release to tag the docker image incorrectly.

**Expected behavior**
The release job correctly tags the docker image.

### LINKED ISSUE #3074: Bug: CRD API generation breaks existing PRs
**Describe the bug**

When we merge an update to our CRD API documentation (as generated by our [scheduled workflow](https://github.com/Azure/azure-service-operator/actions/workflows/api-docs-repo.yaml)), existing PRs that might have passed CI builds will fail.

**To Reproduce**

Prerequisities:
* A recently merged PR#1 with new resources that are not linked to docs in our index pages
* An existing PR#2 ready for review where the CI checks all pass
* An automatically scheduled PR#3 containing updated CRD API docs for the new resources recently merged by PR#1

When the scheduled PR#3 is merged, the CI build for the PR#2 will fail because the index pages will now include links to the CRD API documentation for the new resources, even though PR#2 has nothing to do with those resources.

**Expected behavior**

Merging scheduled PR#3 should not break any other builds.

theunrepentantgeek:

Fixes for a number of documentation issues.

* Fixes hand-crafted resources showing as Deprecated, instead of released
* CRD API workflow now regenerates the index files as well.
* Instead of basing the available versions report on the latest tag (which breaks other tag based release activities), use some configuration, though this will require us to manually update it for each release.
* Release details added to both the home page for the repo (for both v1 and v2) and on the home page for v2 (listing current and prior).

Fixes #3074
Fixes #3070
Closes #3067

gif

**If applicable**:


v2/README.md
```diff
@@ -11,6 +11,11 @@ Instead of deploying and managing your Azure resources separately from your Kube
 ## Project Status
 This project is stable. We follow the [Kubernetes definition of stable](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-stages).
 
+| Release                                                                       | Date       |                 |
```
matthchr, line 14: Can you also update the docs/hugo/content/contributing/create-a-new-release.md documentation to mention we should add the new release here?

matthchr, line 14: Is this intended to be a list of... recent releases? Releases in the last year? All releases?

Might help to have a header clarifying what we are showing here?

v2/azure-arm.yaml
```diff
@@ -31,6 +31,7 @@ supportedResourcesReport:
   # resourcePath is a template used for generating a file path for checking whether docs for a resource have been generated
   # specified relative to the directory of outputPath, and supports the same placeholders
   resourcePathTemplate: "{group}/{version}.md"
+  currentRelease: v2.1.0
```
matthchr, line 34: Update docs/hugo/content/contributing/create-a-new-release.md documentation to mention we need to update this. Clarify when it should happen, before or after the actual release?

README.md
```diff
@@ -20,10 +20,11 @@ There are two major versions of Azure Service Operator: v1 and v2. Consult the b
...
+| ASO Version | Lifecycle stage | Current Release                                                                     | Development status                | Installation options                                                                                                                                                                                    |
+| ----------- | --------------- | ----------------------------------------------------------------------------------- | --------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| v2          | Stable          | [v2.1.0](https://github.com/Azure/azure-service-operator/releases/tag/v2.1.0)       | Under active development.         | [Helm chart](/v2/charts), [GitHub release 2.x](https://github.com/Azure/azure-service-operator/releases). See [installation](https://azure.github.io/azure-service-operator/#installation) for example. |
```
matthchr, line 25: I'm concerned we're putting specific links here that will become out of date when we forget to update them as part of the release process. If we document the need to update this in the release process then it makes the release process more manual.

I accept the need for _some_ links/updates, but every one we add adds more burden on the releaser.

Can we just link https://github.com/Azure/azure-service-operator/releases here instead for v2.0 maybe? The latest release is always at the top of that page so it's pretty much as good as linking to it. OTOH, we can use https://github.com/Azure/azure-service-operator/releases/latest, which is an alias for the latest release, and avoid mentioning the specific version here so we don't have to come update it every release?

Putting the actual version into our readme seems unnecessary since it's already on the front page of the repo: 
image

v2/README.md
matthchr: As mentioned in my other comment: This information is already available in a standardized place on GitHub, the [releases tab](https://github.com/Azure/azure-service-operator/releases). Since it's part of the GitHub standard UI, folks know to navigate there to check out releases. There's also already a shortcut to it on the main page in a standardized place.

What do you think about just linking to that location here, rather than making our own table that just mirrors what's already automatically included in the releases UI?

matthchr: I took a look around at a few other projects and I don't see a list of releases in the readme of CAPZ, Azure Workload Identity, cert-manager, or Kubernetes.

cert-manager does have a list [here](https://cert-manager.io/docs/installation/supported-releases/), although the context is a bit different because there it's saying which versions are supported whereas in our case I think our only officially supported version is whatever the latest release is.

I suspect Kubernetes also has a manually maintained list somewhere although in a few minutes of searching I can't find it.

theunrepentantgeek: Part of the issue I'm trying to address is caused by our reuse this README as the landing page of our website as well, where there is not a convenient list of existing versions/releases.

Maybe a link would be good enough? I'm unsure.

I very nearly titled the first column **Supported Releases**, but wasn't sure about doing that.

v2/azure-arm.yaml
theunrepentantgeek: I'll either do that, or update the workflow to update `currentRelease` automatically.

v2/README.md
matthchr: I guess my point is, none of the repos I checked have a list of versions on the front page of their site either. Obviously that doesn't prevent us from putting a list there, but is it what people are expecting? I would feel differently if we actually did have a list of supported versions, but I think by most definitions of supported we currently only support the latest release.

I just don't want us to have to manually maintain something that we then end up forgetting to update, as I think having a list and having it be out of date is a worse look than just not having a list at all, and there are already a number of "standard" ways to get versions such as:
1. The GitHub releases page
2. Helm commands to list releases (if you're using Helm).

theunrepentantgeek: Done.

theunrepentantgeek: Removed.

README.md
theunrepentantgeek: Removed.

v2/azure-arm.yaml
super-harsh, line 34: Also, can consider adding a yq cmd like below in `generate-types` task target: 

`yq eval -i  ".currentRelease=\"{LATEST_VERSION_TAG}\"" ./v2/azure-arm.yaml`

theunrepentantgeek: Thanks for the command, it's pretty close to what I want to put into the CRD Docs workflow.

But wouldn't the path need to be `.supportedResourcesReport.currentRelease` ?

docs/hugo/content/contributing/create-a-new-release.md
```diff
@@ -12,7 +12,7 @@ title: Creating a new release
 8. Ensure that the action associated with your release finishes successfully.
 9. Create a new PR to update our documentation and move resources listed under "Next Release" to the heading "Released" by checking out `main`, running `task` and creating a PR of the results.
```
matthchr, line 13: Should this entry link to the new section you added below (#update-documentation)?

theunrepentantgeek: Yes. Done.

PR #3087, opened 2023-06-19
Retry Workspace if storage account not ready

### LINKED ISSUE #3086: Bug: MachineLearningServices Workspace can fail to reconcile
**Describe the bug**

Reconciliation of [`MachineLearningServices.WorkSpace`](https://azure.github.io/azure-service-operator/reference/machinelearningservices/v1api20210701/#machinelearningservices.azure.com/v1api20210701.Workspace) will fail if the required StorageAccount and KeyVault are not already provisioned.

**To Reproduce**

Create a single yamlfile containing the `Workspace` along with the `StorageAccount` and the `KeyVault` as well and deploy it to ASO.

**Expected behavior**

All the resources are successfully reconciled. 

What will actually happen is that the `Workspace` will end up in a permanent failure state because provisioning is attempted before the `StorageAccount` and `KeyVault` are fully deployed. 

**Additional context**
ASO has extension points to allow customization of behaviour that will let us fix this.

Discovered while setting up testing for another resource. We didn't uncover the issue sooner because our samples tests deploy all the pre-requisites first (e.g. the `StorageAccount` and the `KeyVault`) before deploying the Workspace itself.

theunrepentantgeek:

If a required storage account is not yet provisioned when a `MachineLearningServices.Workspace` is deployed, that deployment will fail. We need to retry when that happens to allow `StorageAccount` provisioning to complete.

Fixes #3086 

We didn't catch this with our tests because they provisioned the `StorageAccount` first and only then attempted to provision the `Workspace` after that was complete.

gif

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/api/machinelearningservices/customizations/workspace_extension.go
+++ b/v2/api/machinelearningservices/customizations/workspace_extension.go
@@ -7,11 +7,12 @@ package customizations
 
 import (
 	"context"
+	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/machinelearning/armmachinelearning"
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/conversion"
 
@@ -20,6 +21,8 @@ import (
 	. "github.com/Azure/azure-service-operator/v2/internal/logging"
 	"github.com/Azure/azure-service-operator/v2/internal/util/to"
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/core"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/extensions"
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/secrets"
 )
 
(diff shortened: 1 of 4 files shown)


v2/api/machinelearningservices/customizations/workspace_extension.go
```diff
@@ -141,3 +145,38 @@ func getContainerRegCreds(keysResp armmachinelearning.ListWorkspaceKeysResult) (
...
+
+	return details, nil
+}
+
+func isRetryableWorkspaceError(err *genericarmclient.CloudError) bool {
+	if err == nil {
+		return false
+	}
+
+	if err.Code() == "BadRequest" && strings.Contains(err.Message(), "StorageAccountIsNotProvisioned") {
```
super-harsh, line 176: Do we do same for the KeyVault as well?

super-harsh (approved):

v2/internal/controllers/crd_machinelearningservices_test.go
```diff
@@ -27,16 +27,12 @@ func Test_MachineLearning_Workspaces_CRUD(t *testing.T) {
 	rg := tc.CreateTestResourceGroupAndWait()
 
 	sa := newStorageAccount(tc, rg)
-
-	tc.CreateResourceAndWait(sa)
-
-	kv := newVault(tc, rg)
-	tc.CreateResourceAndWait(kv)
+	kv := newVault("kv", tc, rg)
```
matthchr, line 30: curious: Why did we need the name change here? Was it because there was an old KeyVault somewhere that we had a hard time finding to purge?

matthchr (approved):

v2/internal/controllers/crd_machinelearningservices_test.go
theunrepentantgeek: Yeah - after spending 20+m trying to find it, I figured a name change was easier.

v2/api/machinelearningservices/customizations/workspace_extension.go
theunrepentantgeek: I couldn't make it fail and couldn't find what the error code might be. The obvious `KeyVaultIsNotProvisioned` didn't turn up any hits.

v2/internal/controllers/crd_machinelearningservices_test.go
theunrepentantgeek: I did it this way to avoid changing the other recordings.

PR #3088, opened 2023-06-19
Allow samples tester to detect unknown fields

### LINKED ISSUE #3069: Samples tests don't catch unknown properties
If a property is in the wrong location or is typoed, the way our samples tests deserialize the sample YAML discards this unknown property.

Ideally we would instead error, to ensure that our samples schema conforms exactly to the expected shape of the API.

super-harsh:
Closes #3069 

Replacing `runtime.Decoder` with json.Decoder to detect unknown fields and return errors. Found a few invalid samples, which have also been fixed in this PR


matthchr (approved):
Nice!

PR #3094, opened 2023-06-20
Generate RoleAssignment AzureName based on GK+Namespace+Name

### LINKED ISSUE #2352: Support autogenerating RoleAssignment GUID for AzureName
mcr.microsoft.com/k8s/azureserviceoperator:v2.0.0-beta.0

**Describe the bug**
When I try to add a RoleAssignment to a ServiceBus have got an error about the azureName.
This field is used to put the role assignment to Azure as ID, but for a RoleAssignment it should be a GUID.

ASO use the resource name as azureName by default, so it doesn' work.

**To Reproduce**
- Apply this resource definition
```yaml
apiVersion: authorization.azure.com/v1alpha1api20200801preview
kind: RoleAssignment
metadata:
  name: test-role-ass
  namespace: monitoring
spec:
  location: "East US"
  owner: 
    group: servicebus.azure.com
    name: cdm-device-monitoring-sb
    kind: Namespace
  # This is the Principal ID of the AAD identity to which the role will be assigned
  principalId: c9ad1210-fbe8-49ad-8d92-XXXXXXXXX
  roleDefinitionReference:
    # This ARM ID represents "Contributor" - you can read about other built in roles here: https://docs.microsoft.com/en-us/azure/role-based-access-control/built-in-roles
    armId: /subscriptions/9e7dbf8b-6a81-4beb-9434-8113e1c7a0a4/providers/Microsoft.Authorization/roleDefinitions/b24988ac-6180-42a0-ab88-20f7382dd24c
```
- Received this error:
```text
 The role assignment ID 'test-role-ass' is not valid. The role assignment ID must be a GUID.: PUT https://management.azure.com/subscriptions/XXXXX-XXXX-XXX-XXXXXX/resourceGroups/cdm-device-monitoring-rg/providers/Microsoft.ServiceBus/namespaces/cdm-device-monitoring-sb/providers/Microsoft.Authorization/roleAssignments/test-role-ass
      RESPONSE 400: 400 Bad Request
      ERROR CODE: InvalidRoleAssignmentId
      {
        "error": {
          "code": "InvalidRoleAssignmentId",
          "message": "The role assignment ID 'test-role-ass' is not valid. The role assignment ID must be a GUID."
        }
      }
```

**Expected behavior**
The RoleAssignment is working. Either, the azureName field is mandatory with a correct description, either this field is autogenerated with a correct description in the documentation.

matthchr:
- Add $defaultAzureName configuration option to azure-arm.yaml. This allows disabling the standard defaultAzureName() webhook which defaults AzureName to the Kubernetes resource name.
 - Disable defaultAzureName for RoleAssignment and SQLRoleAssignment.
 - Add a handcrafted defaultAzureName for RoleAssignment and SQLRoleAssignment which generates a UUID based on Owner.Group + Owner.Kind + Owner.Name + Group + Kind + Namespace + Name.

Fixes #2352.

**If applicable**:
- [x] this PR contains tests


v2/api/authorization/v1api20200801preview/role_assignment_defaults.go
```diff
@@ -0,0 +1,66 @@
...
+func (assignment *RoleAssignment) defaultAzureName() {
+	if assignment.AzureName() == "" {
+		_, err := uuid.Parse(assignment.Name)
+		if err == nil {
+			// Name is a UUID, so just use that for AzureName
+			assignment.Spec.AzureName = assignment.Name
+			return
+		}
+
+		uniqueStr := randextensions.MakeUniqueResourceString(GroupVersion.Group, fmt.Sprintf("%T", RoleAssignment{}), assignment.Namespace, assignment.Name)
```
theunrepentantgeek, line 62: Why use `%T` here? Might that change between compiler releases? 
Using `RoleAssignment{}.GetType()` would return `Microsoft.Authorization/roleAssignments` which is more stable.

v2/api/authorization/v1api20200801preview/role_assignment_types_gen.go
```diff
@@ -81,15 +81,8 @@ func (assignment *RoleAssignment) Default() {
...
-// defaultAzureName defaults the Azure name of the resource to the Kubernetes name
-func (assignment *RoleAssignment) defaultAzureName() {
-	if assignment.Spec.AzureName == "" {
-		assignment.Spec.AzureName = assignment.Name
-	}
-}
-
 // defaultImpl applies the code generated defaults to the RoleAssignment resource
-func (assignment *RoleAssignment) defaultImpl() { assignment.defaultAzureName() }
+func (assignment *RoleAssignment) defaultImpl() {}
```
theunrepentantgeek, line 85: Don't you still want this?

v2/api/authorization/v1beta20200801preview/role_assignment_defaults.go
```diff
@@ -0,0 +1,66 @@
...
+	if assignment.AzureName() == "" {
+		_, err := uuid.Parse(assignment.Name)
+		if err == nil {
+			// Name is a UUID, so just use that for AzureName
+			assignment.Spec.AzureName = assignment.Name
+			return
+		}
+
+		uniqueStr := randextensions.MakeUniqueResourceString(GroupVersion.Group, fmt.Sprintf("%T", RoleAssignment{}), assignment.Namespace, assignment.Name)
+		azureName := uuid.NewSHA1(randextensions.Namespace, []byte(uniqueStr))
```
theunrepentantgeek, line 63: Can the majority of this be pulled out into a helper method?

theunrepentantgeek (approved):
I think the code can be simplified with another helper but looks good otherwise.

v2/api/authorization/v1beta20200801preview/role_assignment_defaults.go
matthchr: Probably

v2/internal/util/randextensions/uuid5.go
```diff
@@ -0,0 +1,18 @@
...
+	"fmt"
+
+	"github.com/google/uuid"
+)
+
+// Namespace is the ASOv2 UUIDv5 namespace UUID
+var Namespace = uuid.Must(uuid.Parse("9a329043-7ad7-4b1c-812f-9c7a93d6392a"))
+
+// MakeUniqueResourceString generates a string that uniquely identifies a cluster resource.
+func MakeUniqueResourceString(group string, kind string, namespace string, name string) string {
```
super-harsh, line 16: Not sure if GK + Namespace + Name is the best way? There could be collisions if there are two clusters, having same namespace and using similar name. Maybe we want to make it to be based on RGName + ResourceName, which makes more sense as per Azure perspective.

v2/api/authorization/v1api20200801preview/role_assignment_defaults.go
matthchr: > Why use %T here? Might that change between compiler releases?

It won't change at least between versions of Go 1.x due to Go's compatibility guarantees. I chose `%T` here because `RoleAssignment{}.GetType()` isn't the `Kind` and I was trying to base it on Group + Kind + Namespace + Name (the Kubernetes constructs). Your comment made me realize though that I can just do this:
```
gvk := assignment.GroupVersionKind()
uniqueStr := randextensions.MakeUniqueResourceString(gvk.Group, gvk.Kind, assignment.Namespace, assignment.Name)
```

Which is much cleaner.

v2/internal/util/randextensions/uuid5.go
theunrepentantgeek: I think adding more content to make the GUID _"more unique"_ is the wrong way to think about it.

If we wanted every instance to be unique, we'd just generate a unique GUID and be done with it.

We're also talking about a _useful default_ - which means that users do have the capability to override.

The question here is: when do we want the default GUID to be the same vs when should it be different?

I think Namespace + Name + Group + Kind is more than sufficient.

HS> There could be collisions if there are two clusters, having same namespace and using similar name.

Same namespace and _similar_ name wouldn't collide - that's in the nature of GUIDs, even ones generated from a seed value. The only collision would be same namespace and _same_ name, in which case it's working as designed.

matthchr: After discussion in person I've made this change and re-recorded the tests.

theunrepentantgeek (approved):

super-harsh (approved):

PR #3097, opened 2023-06-21
Eliminate most configuration boilerplate

theunrepentantgeek:

We previously needed to create a lot of boilerplate for each new configuration item.

For example, to allow the property configuration `$iIsSecret`, we needed to first add two methods to `PropertyConfiguration`:

``` go
// IsSecret looks up a property to determine if it's a secret
func (pc *PropertyConfiguration) IsSecret() (bool, error) {
	isSecret, ok := pc.isSecret.read()
	if !ok {
		msg := fmt.Sprintf(isSecretTag+" not specified for property %s", pc.name)
		return false, NewNotConfiguredError(msg)
	}

	return isSecret, nil
}

// VerifyIsSecretConsumed returns an error if our configuration as a secret was not consumed.
func (pc *PropertyConfiguration) VerifyIsSecretConsumed() error {
	if pc.isSecret.isUnconsumed() {
		v, _ := pc.isSecret.read()
		return errors.Errorf("property %s: "+isSecretTag+": %t not consumed", pc.name, v)
	}

	return nil
}
```

and then additional support on `ObjectModelConfiguration` to provide easy access:

``` go
// IsSecret looks up a property to determine whether it is a secret.
func (omc *ObjectModelConfiguration) IsSecret(name astmodel.TypeName, property astmodel.PropertyName) (bool, error) {
	var result bool
	visitor := newSinglePropertyConfigurationVisitor(
		name,
		property,
		func(configuration *PropertyConfiguration) error {
			isSecret, err := configuration.IsSecret()
			result = isSecret
			return err
		})

	err := visitor.Visit(omc)
	if err != nil {
		return false, err
	}

	return result, nil
}

// VerifyIsSecretConsumed returns an error if any IsSecret configuration was not consumed
func (omc *ObjectModelConfiguration) VerifyIsSecretConsumed() error {
	visitor := newEveryPropertyConfigurationVisitor(
		func(configuration *PropertyConfiguration) error {
			return configuration.VerifyIsSecretConsumed()
		})
	return visitor.Visit(omc)
}
```

We can now achieve the same effect by adding a new field to `PropertyConfiguration` and initalizing it:

``` go
type PropertyConfiguration struct {
        // ... elided ...
        IsSecret configurable[bool] // Specify whether this property is a secret
        // ... elided ...
}

func NewPropertyConfiguration(name string) *PropertyConfiguration {
	scope := "property " + name
	return &PropertyConfiguration{
        	// ... elided ...
		IsSecret:                       makeConfigurable[bool](isSecretTag, scope),
        	// ... elided ...
	}
}
```

and then by adding another field to `ObjectModelConfiguration` and initializing that:

``` go
type ObjectModelConfiguration struct {
        // ... elided ...
	IsSecret                       propertyAccess[bool]
        // ... elided ...
}

 func NewObjectModelConfiguration() *ObjectModelConfiguration {
        // ... elided ...
	result.IsSecret = makePropertyAccess[bool](
		result, func(c *PropertyConfiguration) *configurable[bool] { return &c.IsSecret })
        // ... elided ...
}
```

gif

**If applicable**:
- [x] this PR contains tests


v2/tools/generator/internal/config/configurable.go
```diff
@@ -5,11 +5,55 @@
...
+	if c.isUnconsumed() {
+		return errors.Errorf("%s specified for %s but not consumed", c.tag, c.scope)
+	}
+
+	return nil
+}
+
+// MarkUnconsumed resets the consumption flag so that the value can be reused
+func (c *configurable[T]) MarkUnconsumed() {
+	c.markUnconsumed()
```
matthchr, line 51: Maybe just get rid of the `markUnconsumed` and change all callers to use `MarkUnconsumed`? It seems weird to have this alias here when the actual function is 1 line long.

v2/tools/generator/internal/config/configurable.go
```diff
@@ -5,11 +5,55 @@
...
+}
+
+// Set sets the value configured (used in tests)
+func (c *configurable[T]) Set(v T) {
+	c.write(v)
```
matthchr, line 56: Same here maybe swap call calls of `write` to the public `Set` so we don't need both?

v2/tools/generator/internal/config/object_model_configuration.go
```diff
@@ -29,14 +29,92 @@ import (
...
+	TypeNameInNextVersion    typeAccess[string]
+
+	// Property access fields here (alphabetical, please)
+	ARMReference                   propertyAccess[bool]
+	ImportConfigMapMode            propertyAccess[ImportConfigMapMode]
+	IsSecret                       propertyAccess[bool]
+	ResourceLifecycleOwnedByParent propertyAccess[string]
+}
+
+type groupAccess[T any] struct {
```
matthchr, line 57: Go supports multiple type parameters, right?

Could we have a single type: `type accessor[TConfiguration any, TProperty any]`? I think the issue here would be that there are slightly different Lookup implementations for each of the group/type/property variants, so I guess this isn't possible unless you also had `accessor` take a lambda that did the lookup (making it more awkward to construct), or possibly made `TConfiguration` require to implement an interface where that interface impl had some of the lookup logic?

Obviously pretty minor but just when I was looking at these types there' basically identical in shape here except for the first parameter of `accessor`

matthchr (approved):

v2/tools/generator/internal/config/configurable.go
theunrepentantgeek: Done.

v2/tools/generator/internal/config/configurable.go
theunrepentantgeek: I was mistakenly trying to avoid making wider changes by renaming the method. Changed now.

v2/tools/generator/internal/config/object_model_configuration.go
theunrepentantgeek: Different Lookup() signatures too, so hard to abstract. I did try!

PR #3111, opened 2023-06-29
[Not for merge] Prototype for how we might make recordings less fragile

theunrepentantgeek:

Based on a Teams discussion earlier today, one of the reasons our recording tests are fragile is that while the reconciler is goal-seeking, the recordings are transactional - so trivial changes that would have no impact on real use will break the tests.

In this PR, I've prototyped using a hash of the request body to disambiguate PUT requests only; GET requests are unchanged.

When a duplicate PUT occurs, the body will be the same, so the same response will be returned (goal seeking).
If that response includes a reference to a long running operation, the existing sequence of GETs for that operation can be reused.

This is a hack to prove the concept and open discussion; I intend to close this PR after discussion.
If we go ahead, I'll rewrite and create a new PR.

gif


theunrepentantgeek:
Abandoning in favour of https://github.com/Azure/azure-service-operator/pull/3753

PR #3115, opened 2023-06-30
chore: use time.Since instead of time.Now().Sub

testwill:
Closes #[issue number]

use time.Since instead of time.Now().Sub

gif

**If applicable**:
- [ ] this PR contains tests


matthchr (approved):
This looks kinda like a bot to me, but the change is good so why not.

matthchr:
I realized that this is targeting ASOv1 code which is in maintenance mode (bugfixes/security patches only), so going to close this

PR #3131, opened 2023-07-10
Fix to handle empty property ARM conversion for embedded resources

super-harsh:

This PR adds a fix to notice and ignore the properties for ConvertToARM conversion. These properties are pruned by `pruneResourcesWithLifecycleOwnedByParentStage` but still do have an ARM type.


v2/tools/generator/internal/codegen/pipeline/prune_resources_with_lifecycle_owned_by_parent.go
```diff
@@ -69,19 +75,65 @@ func PruneResourcesWithLifecycleOwnedByParent(configuration *config.Configuratio
...
+
+		tempResult := make(astmodel.TypeDefinitionSet)
+		for _, def := range defs {
+			updatedDef, err := emptyObjectVisitor.VisitDefinition(def, armDef.Name())
+			if err != nil {
+				return nil, errors.Wrapf(err, "failed to visit definition %s", def.Name())
+			}
+			tempResult.Add(updatedDef)
+		}
+		defs = tempResult
```
theunrepentantgeek, line 98: Visiting a type is relatively expensive - a visitor walks the entire passed type, making precise edits where required. It would be substantially more efficient to visit each type exactly once.

v2/tools/generator/internal/codegen/pipeline/prune_resources_with_lifecycle_owned_by_parent.go
```diff
@@ -69,19 +75,65 @@ func PruneResourcesWithLifecycleOwnedByParent(configuration *config.Configuratio
...
+	prop, ok := it.Properties().Find(func(prop *astmodel.PropertyDefinition) bool {
+		name, ok := astmodel.ExtractTypeName(prop.PropertyType())
+		if !ok {
+			return false
+		}
+		return name.Name() == typeName.Name()
+	})
+
+	if ok {
+		prop = prop.WithTag(armconversion.ConversionTag, armconversion.NoARMConversionValue)
```
theunrepentantgeek, line 132: Do these tags show up in the generated code?

v2/tools/generator/internal/codegen/pipeline/prune_resources_with_lifecycle_owned_by_parent.go
```diff
@@ -69,19 +75,65 @@ func PruneResourcesWithLifecycleOwnedByParent(configuration *config.Configuratio
 	return stage
 }
 
+func checkPrunedEmptyProperties(defs astmodel.TypeDefinitionSet, emptyPrunedProps []astmodel.TypeName) (astmodel.TypeDefinitionSet, error) {
```
theunrepentantgeek, line 78: Is `check` the best verb here? This doesn't really seem like a validation method. Maybe `flagPrunedEmptyProperties`?

theunrepentantgeek (commented):
This came out more simply than expected - a few comments to address, then I'll approve.

v2/tools/generator/internal/codegen/pipeline/prune_resources_with_lifecycle_owned_by_parent.go
super-harsh: Yep, makes sense. I've refactored the code

v2/tools/generator/internal/codegen/pipeline/prune_resources_with_lifecycle_owned_by_parent.go
super-harsh: Yep.

v2/tools/generator/internal/codegen/pipeline/prune_resources_with_lifecycle_owned_by_parent.go
super-harsh: `Flag` was the word I was looking for. Done!

v2/tools/generator/internal/codegen/pipeline/prune_resources_with_lifecycle_owned_by_parent.go
super-harsh: It makes sense to have them?

theunrepentantgeek: The annotation is only consumed within the generator, it has no runtime function - so I think it would be ideal to clean up so they don't show in the generated code. Not sure on the ROI of the effort though.

theunrepentantgeek (approved):

v2/tools/generator/internal/codegen/pipeline/prune_resources_with_lifecycle_owned_by_parent.go
```diff
@@ -69,19 +75,66 @@ func PruneResourcesWithLifecycleOwnedByParent(configuration *config.Configuratio
...
+}
+
 type misbehavingEmbeddedTypePruner struct {
-	configuration *config.Configuration
+	configuration         *config.Configuration
+	emptyPrunedProperties astmodel.TypeNameSet
 }
 
-func newMisbehavingEmbeddedTypeVisitor(configuration *config.Configuration) astmodel.TypeVisitor {
+func newMisbehavingEmbeddedTypeVisitor(configuration *config.Configuration) (astmodel.TypeVisitor, *misbehavingEmbeddedTypePruner) {
```
matthchr, line 107: hmm, maybe refactor this to just return `pruner` and have `pruner` have a property `visitor`?

It's a bit cyclic but would result in a single return type rather than two, which feels cleaner?

v2/tools/generator/internal/codegen/pipeline/prune_resources_with_lifecycle_owned_by_parent.go
matthchr: I think we have another tag that's only on the ARM type which we also have left in the code for now.

I see the ARM types as pretty internal so I'm OK if it's just on the ARM type. If it's going to be on the k8s type too we should remove it.

My 2c is one of these options makes sense:
1. Just leave it.
2. Leave it but file an issue to clean it (and any other tags) up.

I think I tend towards option 1 but I'm not married to it.

matthchr (approved):
LGTM

v2/tools/generator/internal/codegen/pipeline/prune_resources_with_lifecycle_owned_by_parent.go
super-harsh: Done!

PR #3132, opened 2023-07-10
ADR for updating Package References

theunrepentantgeek:

As discussed recently, if we want to move the ARM specific implementations into a subpackage, we'll need a new implementation of `PackageReference` to support that use (all our current implementations presume a package structure ending in `group/version`). 

To support this, I propose we make some minor changes to our domain model in that area to make it more closely match our needs.

**Special notes for your reviewer:**

Trying a new approach, using a Hubo content bundle for the ADR. This is the fancy term Hugo uses for putting everything into a folder together.

gif

**If applicable**:


docs/hugo/content/design/ADR-2023-07-PackageReferences/_index.md
```diff
@@ -0,0 +1,175 @@
...
+
+Breaking the implementation link between `ExternalPackageReference` and `PackageReference` means it's no longer possible for a `TypeDefinition` to be destined for an external package; nor is it possible for a StoragePackageReference to be based on the same. 
+
+The new implementation `SubPackageReference` declares a `parent` package for nesting.
+
+* PRO: Makes illegal states impossible to represent.
+* PRO: Removes need for type downcasts.
+* PRO: Removes need for checks for illegal states.
+* PRO: Makes the model easier to understand.
+* CON: Requires slightly more rework of existing code.
```
matthchr, line 127: There are other cons here IMO:
1. It requires `ExternalTypeName` to basically exist as a duplicate of `TypeName`. I _think_ that most any modification that would apply to `TypeName` would likely also apply to `ExternalTypeName`.
2. I also think that "slightly more rework" may be understating the challenge. TypeName is somewhat "special" already and often has special treatment across the codebase. For example in convert_from_arm_function_builder.go, we do:
    ```
    ownerNameType, ok := astmodel.AsTypeName(toProp.PropertyType())
    ownerNameType == astmodel.KnownResourceReferenceType
    ```
Which I think could just be replaced with `AsExternalTypeName`, but it would need to be fixed up. We would basically need to go through every explicit `TypeName` reference and ask ourselves "is this always a type from the generated code or could be also be an external type".
Many of these places may just be able to move to reference either a `TypeName` or an `ExternalTypeName`, but not all of them can.

For example, places such as `NewInterfaceImplementation` (from `interface_implementation.go`).  I _think_ we use `NewInterfaceImplementation` to implement both autogenerated interfaces (in the packages) _and_ interfaces from `genruntime` and elsewhere. So we would either need to somehow update that to be generic and work across both, or duplicate the code. Even if we don't do this today, it might be nice to do it in the future. How could we update `interface_implementation.go` to work with both shapes?

Possibly we could update it to work across both with some sort of `TypeNamer` interface, although it's not clear to me how that would differ from the existing `Type` interface (and many of these things cannot be dropped onto any type, they're specific to `TypeName`, so we wouldn't want to use `Type` instead I don't think, as then we're just letting illegal states sneak in someplace else)

In summary: while I definitely think that there are many places where it's an advantage to treat `ExternalTypeName` and `TypeName` differently, I think there are also places where it's going to be a disadvantage (I called out 2 above, the impl itself and `interface_implementation.go`, there are likely others as I didn't look exhaustively). This feels to me like there's a certain amount of inherent complexity here and this decision looks 60-40 either way. **Certainly we should do something**, but I still think I tend to err in the direction of option 2 as:
1. It's a better version of what we have now, so easier to implement
2. It more closely matches actual the actual Go ast/module structure (where there is similar local vs external distinction of type names), which feels more likely to fit better with the rest of astmodel and the code-generator where we often write things based on our knowledge of Go's AST structure (a type with a name can implement an interface).

Not that I'm totally against option3/4, I'm just worried that by diverging from how Go actually models its AST, we're setting ourselves up for pain either in the refactor (where at least it'd be obvious) or later as we try to evolve the generator further and then run afoul of the situation where "these are different but actually I need to treat them the same in this instance". Maybe there's some solution to that problem (that doesn't involve opening up to the whole `Type` interface?) that I am not seeing though?

docs/hugo/content/design/ADR-2023-07-PackageReferences/_index.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Option 4
+
+As for Option 4, but we also rename types to better reflect the new semantics. Given the introduction of `ExternalPackageReference`, maybe we use `InternalPackageReference`:
```
matthchr, line 131: What about TypeName, do we want to rename to `InternalTypeName`?

theunrepentantgeek: I think I showed that on the diagram, so yes.

docs/hugo/content/design/ADR-2023-07-PackageReferences/_index.md
theunrepentantgeek: > It requires `ExternalTypeName` to basically exist as a duplicate of `TypeName`.

I think there's a bunch of stuff that we could drop, making `ExternalTypeName` a smaller type.

> We would basically need to go through every explicit TypeName reference and ask ourselves "is this always a type from the generated code or could be also be an external type".

There are three possibilities:

* It's always the name of a type we're generating
* It's always the name of an external type
* It could be either

Much of the benefit of the change comes from simplifying the first two cases.

I think the last of those is rare, but I think we can support that with a common interface. If we rename `TypeName` to `InternalTypeName` then we could call the interface `TypeName`.

> I'm just worried that by diverging from how Go actually models its AST, we're setting ourselves up for pain

I agree that we need to be cautious here.

matthchr (approved):

PR #3134, opened 2023-07-10
added corner cases for comments.go in astbuilder package

khareyash05:
Increase test coverage of `astbuillder/comments.go` to add corner cases of empty comments and text width =0 

**If applicable**:
- [x] this PR contains tests


v2/tools/generator/internal/astbuilder/comments_test.go
```diff
@@ -48,6 +49,9 @@ func TestDocumentationCommentFormatting(t *testing.T) {
 	}
 
 	for _, c := range cases {
+		if c.comment == "" {
+			continue
+		}
```
theunrepentantgeek, line 54: All this does is to suppress the new test case - it's never run at all, so it adds nothing to the coverage.

v2/tools/generator/internal/astbuilder/comments_test.go
```diff
@@ -70,9 +74,13 @@ func TestWordWrap(t *testing.T) {
 		{"this is a simple line of text", 16, []string{"this is a simple ", "line of text"}},
 		{"this is a simple line of text", 20, []string{"this is a simple ", "line of text"}},
 		{"this is a simple line of text", 21, []string{"this is a simple line ", "of text"}},
+		{"", 0, []string{}},
 	}
 
 	for _, c := range cases {
+		if c.width == 0 {
+			continue
+		}
```
theunrepentantgeek, line 83: Ditto here; all you're doing is skipping the test if the passed width is zero.

theunrepentantgeek (changes requested):

v2/tools/generator/internal/astbuilder/comments_test.go
khareyash05: Will update the same

v2/tools/generator/internal/astbuilder/comments_test.go
```diff
@@ -70,6 +71,8 @@ func TestWordWrap(t *testing.T) {
 		{"this is a simple line of text", 16, []string{"this is a simple ", "line of text"}},
 		{"this is a simple line of text", 20, []string{"this is a simple ", "line of text"}},
 		{"this is a simple line of text", 21, []string{"this is a simple line ", "of text"}},
+		{"", 0, []string{}},
+		{"this is a sample text", 0, []string{}},
```
theunrepentantgeek, line 75: We should never throw away content like this - if someone inadvertently passes a width of `0` we want to return the content anyway. For consistency with existing behaviour when width is `1`, this should be:

```suggestion
		{"this is a sample text", 0, []string{"this ", "is ", "a ", "sample ", "text"}},
```

theunrepentantgeek (changes requested):
Adding coverage of edge cases is valuable, but don't lose sight of the context the methods are used within.

theunrepentantgeek (approved):

theunrepentantgeek:
Thanks @khareyash05.

PR #3136, opened 2023-07-11
Fix networking resources deletion of child resources during adoption

### LINKED ISSUE #3077: Bug: VNET and Subnet created in ASO adopting existing VNET/Subnet will always delete the subnets and recreate them
ASO V2: 

```azureserviceoperator-controller-manager   1/1     1            1           40m   manager,kube-rbac-proxy   mcr.microsoft.com/k8s/azureserviceoperator:v2.1.0,gcr.io/kubebuilder/kube-rbac-proxy:v0.13.1   control-plane=controller-manager```

**Describe the bug**
Create a VNET with ASOV2, added subnets and configured AKS in those subnets - this is done via a locally running Kind instance running on my laptop.

Kind was then destroyed and recreated and the manifests reapplied - all resources reconciled healthy except the VNET (and subsequent children resource dependent on the VNET)

Now it complains with the following error:

```
Subnet podsubnet is in use by /subscriptions/28848791-fcd6-43a7-aeef-767b1a66fede/resourceGroups/rg-vnet-ab-devtest/providers/Microsoft.Network/virtualNetworks/vnet/subnets/podsubnet/serviceAssociationLinks/AzureKubernetesService and cannot be deleted. In order to delete the subnet, delete all the resources within the subnet. See aka.ms/deletesubnet.: PUT https://management.azure.com/subscriptions/28848791-fcd6-43a7-aeef-767b1a66fede/resourceGroups/rg-vnet-ab-devtest/providers/Microsoft.Network/virtualNetworks/vnet -------------------------------------------------------------------------------- RESPONSE 400: 400 ERROR CODE: InUseSubnetCannotBeDeleted -------------------------------------------------------------------------------- { "error": { "code": "InUseSubnetCannotBeDeleted", "message": "Subnet podsubnet is in use by /subscriptions/28848791-fcd6-43a7-aeef-767b1a66fede/resourceGroups/rg-vnet-ab-devtest/providers/Microsoft.Network/virtualNetworks/vnet/subnets/podsubnet/serviceAssociationLinks/AzureKubernetesService and cannot be deleted. In order to delete the subnet, delete all the resources within the subnet. See aka.ms/deletesubnet.", "details": [] } } --------------------------------------------------------------------------------
```

**Expected behavior**
VNET should not be complaining about subnets being used based on this bug fix: #2169

### PR DESCRIPTION
**Pull request by matthchr**
Networking resources have a structure where a PUT on the parent also can delete child resources (or add them). ASO attempts to allow users to manage these networking child resources independent of the owner. In order to do that, when building the PUT payload for the parent resource we used the `ownerReferences.uid` field to look up the child resources owned by the parent resource we were about to send and include them all in the PUT payload of the owner so that they are not deleted.

The issue is that with adoption, parent + child already exist in Azure, but `ownerReferences.uid` isn't set yet in Kubernetes, which means the PUT of the parent doesn't include any reference to its children and so PUT parent ends up deleting the children.

This issue impacts the following resources:
* VirtualNetwork
* LoadBalancer
* RouteTable
* NetworkSecurityGroup

We investigated using `.spec.owner` instead of `ownerReferences`, which works, but inconsistently as there is still a race in Kubernetes client caching when the parent is created during adoption. Sometimes it realizes that it has children and sometimes it doesn't.

Instead, we rely on Azure itself, issuing a GET parent while building the PUT payload. We use the exact contents of the GET response as the PUT payload for the child resource field (`subnets` for VNET, `rules` for LoadBalancer, etc). This ensures that we don't ever delete any child resource when PUT-ing the parent, even if we're adopting that parent for the first time.

Note that _technically_ GET + PUT is subject to a dual-writer race, which we could use ETag and `If-Match` headers to prevent. We didn't implement that here because:
1. Some of the networking APIs support this paradigm but it's not universally available.
2. It would require moderate changes to the reconciler to pipe the etag through.
3. ASO expects to own the resource entirely, so if there is another parallel writer their changes will inevitably lose to ASO as it will keep updating the Azure resource to its desired state.

Given the above it didn't make sense to bother with optimistic concurrency.

Closes #3077

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/api/network/customizations/load_balancer_extension.go
+++ b/v2/api/network/customizations/load_balancer_extension.go
@@ -45,9 +45,15 @@ func (extension *LoadBalancerExtension) ModifyARMResource(
 
 	inboundNatRuleGVK := getInboundNatRuleGVK(obj)
 
+	// We use namespace + owner name here rather than something slightly more specific like actual owner UUID
+	// in order to account for cases where the subresources + owner were just created and so the subsresources haven't
+	// actually been assigned to the owner yet. See https://github.com/Azure/azure-service-operator/issues/3077 for more
+	// details.
+	matchingOwner := client.MatchingFields{".spec.owner.name": obj.GetName()}
+	matchingNamespace := client.InNamespace(obj.GetNamespace())
+
 	inboundNatRules := &network.LoadBalancersInboundNatRuleList{}
-	matchingFields := client.MatchingFields{".metadata.ownerReferences[0]": string(obj.GetUID())}
-	err := kubeClient.List(ctx, inboundNatRules, matchingFields)
+	err := kubeClient.List(ctx, inboundNatRules, matchingOwner, matchingNamespace)
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed listing InboundNatRules owned by LoadBalancer %s/%s", obj.GetNamespace(), obj.GetName())
 	}
--- a/v2/api/network/customizations/load_balancer_extension.go
+++ b/v2/api/network/customizations/load_balancer_extension.go
@@ -4,15 +4,16 @@ package customizations
 
 import (
 	"context"
+	"net/http"
 	"reflect"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
-	"k8s.io/apimachinery/pkg/runtime/schema"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/conversion"
 
 	network "github.com/Azure/azure-service-operator/v2/api/network/v1api20201101storage"
+	"github.com/Azure/azure-service-operator/v2/internal/genericarmclient"
 	. "github.com/Azure/azure-service-operator/v2/internal/logging"
 	"github.com/Azure/azure-service-operator/v2/internal/resolver"
 	"github.com/Azure/azure-service-operator/v2/internal/util/kubeclient"
@@ -28,12 +29,14 @@ var _ extensions.ARMResourceModifier = &LoadBalancerExtension{}
 
 func (extension *LoadBalancerExtension) ModifyARMResource(
 	ctx context.Context,
+	armClient *genericarmclient.GenericClient,
 	armObj genruntime.ARMResource,
 	obj genruntime.ARMMetaObject,
 	kubeClient kubeclient.Client,
 	resolver *resolver.Resolver,
 	log logr.Logger,
 ) (genruntime.ARMResource, error) {
+
 	typedObj, ok := obj.(*network.LoadBalancer)
 	if !ok {
 		return nil, errors.Errorf("cannot run on unknown resource type %T, expected *network.LoadBalancer", obj)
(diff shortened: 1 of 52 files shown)

### CODE REVIEW

File: v2/api/network/customizations/load_balancer_extension.go [hunk 1]
```diff
@@ -157,17 +254,8 @@ func fuzzySetInboundNatRule(inboundNatRule genruntime.ARMResourceSpec, embeddedI
...
-	embeddedInboundNatRuleJSON, err := json.Marshal(embeddedInboundNatRule.Interface())
-	if err != nil {
-		return errors.Wrap(err, "unable to check that embedded inboundNatRule is the same as inboundNatRule")
-	}
-
-	err = fuzzyEqualityComparison(embeddedInboundNatRuleJSON, inboundNatRuleJSON)
-	if err != nil {
-		return errors.Wrap(err, "failed during comparison for embeddedInboundNatRuleJSON and inboundNatRuleJSON")
-	}
+	// TODO: Can't do a trivial fuzzyEqualityComparison here because we don't know which fields are readonly
```
matthchr, line 257: This is my least favorite part of this change. I _think_ it's pretty safe though because before we were using the child resource shape and the parent resource shape and the assertion here was effectively "are they actually the same".

_Now_ we're using the parent resource shape (from GET) and putting that into the parent resource shape (for PUT). Since it's the same resource w/ the same API version, the properties are guaranteed to line-up structurally. Properties that exist on GET but not on PUT are read-only, so it's OK for them to be omitted.

v2/internal/controllers/controller_resources.go
```diff
@@ -242,7 +213,7 @@ func getControllerName(obj client.Object) (string, error) {
 	return fmt.Sprintf("%sController", typ.Name()), nil
 }
 
-func indexOwner(rawObj client.Object) []string {
+func indexOwnerByUID(rawObj client.Object) []string {
```
matthchr, line 216: todo: Possibly just delete both of these index methods as they're unused now

v2/api/network/customizations/load_balancer_extension.go
```diff
@@ -43,55 +46,45 @@ func (extension *LoadBalancerExtension) ModifyARMResource(
...
-	armInboundNatRules := make([]genruntime.ARMResourceSpec, 0, len(inboundNatRules.Items))
-	for _, inboundNatRule := range inboundNatRules.Items {
-		inboundNatRule := inboundNatRule
-
-		var transformed genruntime.ARMResourceSpec
-		transformed, err = transformToARM(ctx, &inboundNatRule, inboundNatRuleGVK, kubeClient, resolver)
-		if err != nil {
-			return nil, errors.Wrapf(err, "failed to transform InboundNatRules %s/%s", inboundNatRule.GetNamespace(), inboundNatRule.GetName())
+	// Get the raw resource
+	raw := make(map[string]any)
```
theunrepentantgeek, line 61: Does this need an initial capacity for the allocation?

v2/api/network/customizations/load_balancer_extension.go
```diff
@@ -120,33 +113,137 @@ func fuzzySetInboundNatRules(lb genruntime.ARMResourceSpec, inboundNatRules []ge
...
+
+func getNameField(natValue reflect.Value) (ret reflect.Value, err error) {
+	defer func() {
+		if x := recover(); x != nil {
+			err = errors.Errorf("caught panic: %s", x)
+		}
+	}()
+
+	nameField := natValue.FieldByName("Name")
+	if (nameField == reflect.Value{}) {
```
theunrepentantgeek, line 134: There's a method for checking whether a Value is the zero value or not. Confusingly, it's not `IsZero()` - that's different.

```suggestion
	if nameField.IsValid() {
```

v2/api/network/customizations/load_balancer_extension.go
```diff
@@ -120,33 +113,137 @@ func fuzzySetInboundNatRules(lb genruntime.ARMResourceSpec, inboundNatRules []ge
...
+		return reflect.Value{}, errors.Errorf("couldn't find name field")
```
theunrepentantgeek, line 135: Just return `nameField`, it saves creating a new empty instance.

v2/api/network/customizations/load_balancer_extension.go
```diff
@@ -120,33 +113,137 @@ func fuzzySetInboundNatRules(lb genruntime.ARMResourceSpec, inboundNatRules []ge
...
+		return nil, errors.Errorf("couldn't find inboundNatRules field on LoadBalancer")
+	}
+
+	inboundNatRulesSlice, ok := inboundNatRules.([]any)
+	if !ok {
+		return nil, errors.Errorf("inboundNatRules field wasn't a slice")
+	}
+
+	return inboundNatRulesSlice, nil
+}
```
theunrepentantgeek, line 165: Can we avoid the repetition here? What about defining a `dynamicObject` as an alias for `map[string]any` as a container for some helper methods?

v2/api/network/customizations/load_balancer_extension.go
```diff
@@ -120,33 +113,137 @@ func fuzzySetInboundNatRules(lb genruntime.ARMResourceSpec, inboundNatRules []ge
...
+
+			if existingName.String() == newRuleName.String() {
+				foundExistingRule = true
+				break
+			}
+		}
+
+		if !foundExistingRule {
+			resultSlice = reflect.Append(resultSlice, inboundNatRule)
+		}
```
theunrepentantgeek, line 240: Is order significant? This puts all the extras at the end of the slice, so any that are known by ASO will be promoted in front of the others.

v2/api/network/customizations/load_balancer_extension_test.go
```diff
@@ -35,15 +34,105 @@ func Test_FuzzySetLoadBalancers(t *testing.T) {
...
+
+	err = setInboundNatRules(lb, azureInboundNatRules)
+	g.Expect(err).ToNot(HaveOccurred())
+	g.Expect(lb.Location).To(Equal(to.Ptr("westus")))
+	g.Expect(lb.Properties).ToNot(BeNil())
+	g.Expect(lb.Properties.InboundNatRules).To(HaveLen(2))
+	g.Expect(lb.Properties.InboundNatRules[0].Properties).ToNot(BeNil())
+	g.Expect(lb.Properties.InboundNatRules[0].Name).To(Equal(to.Ptr("myrule")))
+	g.Expect(lb.Properties.InboundNatRules[0].Properties.BackendPort).ToNot(BeNil())
+	g.Expect(lb.Properties.InboundNatRules[0].Properties.FrontendPort).ToNot(BeNil())
```
theunrepentantgeek, line 73: Suggest making it explicit that we're doing a bunch of checks on the first rule by pulling it out into a local.

```suggestion
	rule0 := lb.Properties.InboundNatRules[0]
	g.Expect(rule0.Properties).ToNot(BeNil())
	g.Expect(rule0.Name).To(Equal(to.Ptr("myrule")))
	g.Expect(rule0.Properties.BackendPort).ToNot(BeNil())
	g.Expect(rule0.Properties.FrontendPort).ToNot(BeNil())
```

Ditto for `rule1`.

v2/internal/controllers/crd_machinelearningservices_test.go
```diff
@@ -17,24 +17,25 @@ import (
...
-
 	tc := globalTestContext.ForTest(t)
 
+	tc.AzureRegion = to.Ptr("westus3")
 	rg := tc.CreateTestResourceGroupAndWait()
 
 	sa := newStorageAccount(tc, rg)
 	kv := newVault("kv", tc, rg)
 
-	// Have to use 'eastus' location here as 'ListKeys' API is unavailable/still broken for 'westus2'
```
theunrepentantgeek, line 36: Should you keep an updated variant of this comment?

v2/internal/controllers/crd_networking_virtualnetwork_test.go
```diff
@@ -127,3 +127,46 @@ func newVNet(tc *testcommon.KubePerTestContext, owner *genruntime.KnownResourceR
...
+	originalVnet := vnet.DeepCopy()
+	originalSubnet := subnet.DeepCopy()
+
+	networkInterface := newVMNetworkInterface(tc, testcommon.AsOwner(rg), subnet)
+	secret := createVMPasswordSecretAndRef(tc)
+	vm := newVirtualMachine20201201(tc, rg, networkInterface, secret)
+	tc.CreateResourcesAndWait(vm, vnet, subnet, networkInterface)
+
+	//tc.Expect(subnet.Status.Id).ToNot(BeNil())
+	//subnetId := *subnet.Status.Id
```
theunrepentantgeek, line 149: Do you need this code?

theunrepentantgeek (approved):
Overall looks good. A few comments, including a possibility to simplify the code using reflection.

super-harsh (approved):

v2/api/network/customizations/load_balancer_extension.go
matthchr: No it's totally overwritten by the HTTP GET payload (note that it's passed as a ptr below to allow for that), and there's no good way to know what the capacity would need to be as we don't know what we're going to get back from Azure.

v2/api/network/customizations/load_balancer_extension.go
matthchr: Refactored things to share a bunch more code.

LoadBalancer is a bit different because it needs to merge rather than overwrite, and that got more complicated with this refactor because now we're looking at Azure state which means that we need to deal with conflicts consistently (previously we merged state from LB in ASO and Routes in ASO, now we're merging LB from ASO w/ LB from Azure, and since LB from Azure will have the routes defined on LB in ASO we have to make sure we choose the ASO ones). So specifically for loadbalancer it doesn't use all of the shared logic it has a bit of its own for that, but the rest of the examples have significantly reduced duplication and even LB makes use of a good bit of that.

v2/api/network/customizations/load_balancer_extension_test.go
matthchr: Done

v2/internal/controllers/crd_machinelearningservices_test.go
matthchr: No as far as I can tell it's no longer correct.

v2/internal/controllers/crd_networking_virtualnetwork_test.go
matthchr: Nope, good eye. Removed.

v2/api/network/customizations/load_balancer_extension.go
matthchr: Done

v2/api/network/customizations/load_balancer_extension.go
matthchr: Done and elsewhere in the shared `getChildCollectionField` function as well

v2/api/network/customizations/load_balancer_extension.go
matthchr: Order doesn't seem to matter. The networking RP seems to return the collection in order of creation time, so the oldest NatRules are first and the newer ones are last. Whatever order you supply rules to it in, that doesn't change.

For our purposes because we're doing a merge here based on name, it doesn't make a different to us I don't think, so the order we end up with here should be OK.

Issue #3142, opened 2023-07-16
Improve documentation for crdPattern


theunrepentantgeek:
**Describe the current behavior**

We require users to specify which CRDs they want to have installed. There are a number of ways to do this, but it's not obvious what values to include in the parameter.

**Describe the improvements**

1. Add docs for each group on our [Supported Resources](https://azure.github.io/azure-service-operator/reference/) page giving the exact addition required to bring in those resources, so that someone reading the list of supported resources has the required information available up front.

2. Break out examples of helm installation using different shells, as there are subtle details different for bash vs cmd.

We may also want to consider making the configuration tolerant of extra quotes.

**Additional context**

It might look something like this:

## AppConfiguration

To install these resources, your ASO configuration must include `appconfiguration.azure.com/*` as a part of the `crdPattern` parameter. See [CRD Management in ASO](https://azure.github.io/azure-service-operator/guide/crd-management/) for detailed instruction for both [Helm](https://azure.github.io/azure-service-operator/guide/crd-management/#helm) and [YAML](https://azure.github.io/azure-service-operator/guide/crd-management/#yaml) based installation.

### Released
These resource(s) are available for use in the current release of ASO. Different versions of a given resource reflect different versions of the Azure ARM API.

... elided ...

theunrepentantgeek:
Generating `crdPattern` docs for each resource group was added in #3147.

PR #3147, opened 2023-07-19
Add CRD pattern docs for each group

theunrepentantgeek:

Adds detail for each group identifying the CRD pattern parameter required for installation of those resources.

Partially addresses #3142 

Implemented by adding a new documentation fragment to show for each group, and by modifying the handling for all fragments to treat them as templates, allowing substitution of values.

gif

**If applicable**:


super-harsh (approved):
LGTM! Although, I'm not sure how many users would know where to find this pattern during installation. A suggestion is to add a Note on [crd management doc page](https://azure.github.io/azure-service-operator/guide/crd-management/) on how and where to find the group pattern.

theunrepentantgeek:
I hope that users will spot the additional docs as they skim the list of supported resources, and then they'll know to loop back and update their installation. Adding more information to the CRD management page is also a good idea, I'll add that to my list.

PR #3157, opened 2023-07-20
Fix sequencing of asoctl builds for release

theunrepentantgeek:

The Taskfile target `controller:run-kustomize` deletes and regenerates all the `zz_generated.deepcopy.go` files, resulting in build errors for `asoctl` unless we ensure the regeneration runs first.

This PR enforces that ordering so the builds work.

gif

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -105,16 +105,7 @@ tasks:
     deps:
       - controller:run-kustomize
       - controller:make-multitenant-files
-      - task: asoctl:build
-        vars: {GOOS: "linux", GOARCH: "amd64"}
-      - task: asoctl:build
-        vars: {GOOS: "linux", GOARCH: "arm64"}
-      - task: asoctl:build
-        vars: {GOOS: "darwin", GOARCH: "amd64"}
-      - task: asoctl:build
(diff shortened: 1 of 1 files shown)


Taskfile.yml
```diff
@@ -172,13 +163,32 @@ tasks:
       - if [ "{{.ARCHIVETYPE}}" = ".zip" ]; then zip -j -r {{.ARCHIVE}} {{.EXECUTABLE}}; fi
       - if [ "{{.ARCHIVETYPE}}" = ".gz" ]; then gzip -v -c {{.EXECUTABLE}} > {{.ARCHIVE}} ; fi
     vars:
+      GOOS: '{{default OS .GOOS}}'
```
matthchr, line 166: minor: Why reorder GOOS to the top?

Feels like it should be next to GOARCH?

matthchr (approved):

Taskfile.yml
theunrepentantgeek: VARs are evaluated sequentially - so we need it set first in case OS is used to override the value. Current usage doesn't run in to the problem, but fixing just in case.

PR #3168, opened 2023-07-27
Fix potential use of wrong statements

theunrepentantgeek:

Something has been nagging at the back of my brain about this prior fix and when I stumbled across the code today, I realized what it was. It's subtle, but worth fixing.

When the `noarmconversion` tag is found on the property, we want to return an empty set of statements (`[]dst.Stmt`), because no conversion is required.

What we've actually been doing is returning _whatever set of statements the first conversion in the list happened to return_.

In practice, this has been an empty set of statements, so the code has worked, but nothing has required this to be the case. 

Put another way, the contract for `propertyConversionHandler` is:
> If I return **false**, don't use the slice of statements.

But we were.

gif

### DIFF
--- a/v2/tools/generator/internal/armconversion/shared.go
+++ b/v2/tools/generator/internal/armconversion/shared.go
@@ -37,18 +37,13 @@ const (
 	TypeKindStatus
 )
 
-const (
-	ConversionTag        = "conversion"
-	NoARMConversionValue = "noarmconversion"
-)
-
 func (builder conversionBuilder) propertyConversionHandler(
 	toProp *astmodel.PropertyDefinition,
 	fromType *astmodel.ObjectType,
 ) []dst.Stmt {
 	for _, conversionHandler := range builder.propertyConversionHandlers {
 		stmts, matched := conversionHandler(toProp, fromType)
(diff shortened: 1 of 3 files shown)


v2/tools/generator/internal/armconversion/shared.go
```diff
@@ -194,3 +189,20 @@ func removeEmptyStatements(stmts []dst.Stmt) []dst.Stmt {
 
 	return result
 }
+
+const (
+	ConversionTag        = "conversion"
+	NoARMConversionValue = "noarmconversion"
+)
+
+func skipNonConvertablePropertyHandler(
```
super-harsh, line 198: I'd suggest having a method name which makes more sense in respect to skipping ARM conversion. Something like  `skipArmConversionPropertyHandler` or `noArmConversionPropertyHandler` ?

super-harsh (approved):

v2/tools/generator/internal/armconversion/shared.go
theunrepentantgeek: I agree on improving the name, came up with `skipPropertiesFlaggedWithNoARMConversion`, inspired by your suggestions.

PR #3171, opened 2023-07-31
Refactor constants from `internal/..` to `pkg/common` package.

### LINKED ISSUE #3149: Expose user-facing constants
Currently, some constant variables that would be useful to import to other Go projects can't be because they are defined in `internal` packages.

e.g.
https://github.com/Azure/azure-service-operator/blob/0eb60541b2ac8c3b8d41ada52f9165cb2cbb8bd6/v2/internal/reconcilers/reconcile_policy.go#L16-L37
https://github.com/Azure/azure-service-operator/blob/0eb60541b2ac8c3b8d41ada52f9165cb2cbb8bd6/v2/internal/reconcilers/predicates.go#L18
https://github.com/Azure/azure-service-operator/blob/0eb60541b2ac8c3b8d41ada52f9165cb2cbb8bd6/v2/internal/config/vars.go#L17-L36

**Describe the improvement**
It would be helpful to projects like CAPZ if these definitions were either moved or mirrored to packages that external projects can import.

**Additional context**

super-harsh:
Closes #3149 

Refactoring constants from `internal/..` to `pkg/common` package.


v2/pkg/common/constants.go
```diff
@@ -0,0 +1,56 @@
...
+// =================== Reconcile Policy Constants ===================
+
+// ReconcilePolicyAnnotation describes the reconcile policy for the resource in question.
+// A reconcile policy describes what action (if any) the operator is allowed to take when
+// reconciling the resource.
+// If no reconcile policy is specified, the default is "run"
+const ReconcilePolicyAnnotation = "serviceoperator.azure.com/reconcile-policy"
+
+const (
+	// ReconcilePolicyManage instructs the operator to manage the resource in question.
```
super-harsh, line 44: Should we keep these comments here, in `reconcile_policy.go` or both places?

super-harsh: Have to do it this way to avoid cyclic imports. Also, would be good to expose these values to external users with comments to have an idea of what the value does.

v2/cmd/controller/app/setup.go
```diff
@@ -37,6 +37,7 @@ import (
 	. "github.com/Azure/azure-service-operator/v2/internal/logging"
 	asometrics "github.com/Azure/azure-service-operator/v2/internal/metrics"
 	armreconciler "github.com/Azure/azure-service-operator/v2/internal/reconcilers/arm"
+	"github.com/Azure/azure-service-operator/v2/pkg/common"
```
theunrepentantgeek, line 40: I'll beat @matthchr to this (because he always catches me out) - move this into the import group for local packages, just below.

v2/internal/reconcilers/arm/arm_client_cache_test.go
```diff
@@ -141,7 +141,7 @@ func Test_ARMClientCache_ReturnsPerResourceScopedClientOverNamespacedClient(t *t
 	g.Expect(err).ToNot(HaveOccurred())
 
 	rg := newResourceGroup("test-namespace")
-	rg.Annotations = map[string]string{reconcilers.PerResourceSecretAnnotation: perResourceCredentialName.Name}
+	rg.Annotations = map[string]string{common.PerResourceSecretAnnotation: perResourceCredentialName.Name}
```
theunrepentantgeek, line 144: I'd prefer if we spread these literal maps across multiple lines, as we already typically do elsewhere.

```suggestion
	rg.Annotations = map[string]string{
		common.PerResourceSecretAnnotation: perResourceCredentialName.Name,
	}
```

v2/internal/reconcilers/reconcile_policy.go
```diff
@@ -11,29 +11,24 @@ import (
...
 	// ReconcilePolicySkip instructs the operator to skip all reconciliation actions. This includes creating
 	// the resource.
-	ReconcilePolicySkip = ReconcilePolicy("skip")
+	ReconcilePolicySkip = ReconcilePolicy(common.ReconcilePolicySkip)
 
 	// ReconcilePolicyDetachOnDelete instructs the operator to skip deletion of resources in Azure. This allows
 	// deletion of the resource in Kubernetes to go through but does not delete the underlying Azure resource.
-	ReconcilePolicyDetachOnDelete = ReconcilePolicy("detach-on-delete")
+	ReconcilePolicyDetachOnDelete = ReconcilePolicy(common.ReconcilePolicyDetachOnDelete)
 )
```
theunrepentantgeek, line 33: I suggest moving all of this into `common` - and get rid of the string typed constants. Consumers can then use the strongly typed `ReconcilePolicyManage` and cast it to a string if they must.

theunrepentantgeek: This also gets rid of the need to document them in two places.

v2/pkg/common/constants.go
```diff
@@ -0,0 +1,56 @@
...
+	ClientCertificateVar = "AZURE_CLIENT_CERTIFICATE"
+	// #nosec
+	ClientCertificatePasswordVar = "AZURE_CLIENT_CERTIFICATE_PASSWORD"
+	TargetNamespacesVar          = "AZURE_TARGET_NAMESPACES"
+	OperatorModeVar              = "AZURE_OPERATOR_MODE"
+	SyncPeriodVar                = "AZURE_SYNC_PERIOD"
+	ResourceManagerEndpointVar   = "AZURE_RESOURCE_MANAGER_ENDPOINT"
+	ResourceManagerAudienceVar   = "AZURE_RESOURCE_MANAGER_AUDIENCE"
+	AzureAuthorityHostVar        = "AZURE_AUTHORITY_HOST"
+	PodNamespaceVar              = "POD_NAMESPACE"
```
theunrepentantgeek, line 23: We should remove the `Var` suffix from all of these, it's kinda implied.

Might pay to put a line or two of documentation on each one, maybe even a link to our docs.

v2/pkg/common/constants.go
theunrepentantgeek: I think we move the `ReconcilePolicy` type here too, retaining type safety for clients who want it (and the others can cast it if they need to).

theunrepentantgeek (approved):

v2/internal/reconcilers/reconcile_policy.go
matthchr: Agree with @theunrepentantgeek

v2/pkg/common/constants.go
```diff
@@ -0,0 +1,56 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package common
+
+// =================== Config variable constants ===================
+
+const (
+	// #nosec
+	ClientSecretVar      = "AZURE_CLIENT_SECRET"
```
matthchr, line 10: I think we should make some effort to differentiate what these constants are for. Previous, when they were in their own separate packages, it was "obvious" which ones were for the Configuration secret and which were used elsewhere.

Now, it's not so obvious. It will be even less obvious if we remove `Var` suffix from them (not that I disagree with the plan to remove that suffix).

I wonder if we should either put the config specific constants into a common subpackage, like `common/config`, so that it clearly explains what goes into the configuration secret, or else we could give them a `Config` suffix (I like that less). We could also at least put them into their own file, although that won't really help clarify for users who are using them as they don't see our file structure just our package structure.

All in all, I think I tend towards `common/config`

matthchr: We could also have a `common/annotations` although I'm less certain that's a good idea.

theunrepentantgeek: I'm still not used to the way Go projects sometimes deliberately have very small packages purely for naming reasons - but I think both `common/config` and `common/annotations` have merit.

v2/internal/identity/credential_provider.go
```diff
@@ -95,7 +98,7 @@ func NewCredentialProvider(
 // If no matching credential can be found, an error is returned.
 func (c *credentialProvider) GetCredential(ctx context.Context, obj genruntime.MetaObject) (*Credential, error) {
 	// Resource annotation
-	cred, err := c.getCredentialFromAnnotation(ctx, obj, common.PerResourceSecretAnnotation)
+	cred, err := c.getCredentialFromAnnotation(ctx, obj, annotations.PerResourceSecretAnnotation)
```
theunrepentantgeek, line 101: Given the ~namespace~ package it's now in, can remove the suffix

```suggestion
	cred, err := c.getCredentialFromAnnotation(ctx, obj, annotations.PerResourceSecret)
```

v2/pkg/common/annotations/annotations.go
```diff
@@ -0,0 +1,12 @@
...
+
+package annotations
+
+const PerResourceSecretAnnotation = "serviceoperator.azure.com/credential-from"
+
+// ReconcilePolicyAnnotation describes the reconcile policy for the resource in question.
+// A reconcile policy describes what action (if any) the operator is allowed to take when
+// reconciling the resource.
+// If no reconcile policy is specified, the default is "run"
+const ReconcilePolicyAnnotation = "serviceoperator.azure.com/reconcile-policy"
```
theunrepentantgeek, line 12: Suffix mirrors the package name and can be removed.

```suggestion
const ReconcilePolicy = "serviceoperator.azure.com/reconcile-policy"
```

v2/pkg/common/config/config.go
```diff
@@ -0,0 +1,58 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package config
+
+const (
```
theunrepentantgeek, line 6: These read really nicely now.

v2/pkg/common/reconciler/reconcile_policy.go
```diff
@@ -0,0 +1,31 @@
...
+
+package reconciler
+
+type ReconcilePolicy string
+
+const (
+	// ReconcilePolicyManage instructs the operator to manage the resource in question.
+	// This includes issuing PUTs to update it and DELETE's to delete it from Azure if deleted in Kubernetes.
+	// This is the default policy when no policy is specified.
+	ReconcilePolicyManage = ReconcilePolicy("manage")
```
theunrepentantgeek, line 12: The common prefix here mirrors the name of the package, so can probably be removed for the type and all values.

```suggestion
	PolicyManage = Policy("manage")
```

theunrepentantgeek (approved):

v2/pkg/common/reconciler/reconcile_policy.go
matthchr, line 12: This is an annotation value, so shouldn't it just go in annotations?

v2/pkg/common/annotations/annotations.go
matthchr: Hmm actually, one interesting thing here is key vs value. This is the reconcile policy annotation key, but there's also a reconcile policy annotation value (which is currently in `reconciler` but IMO should be here.

We should adapt a common naming pattern for both of these.

`annotations.ReconcilePolicy` (for the annotation) and `annotations.ReconcilePolicyValue` (the enum type), and `annotations.ReconcilePolicySkip` or `annotations.ReconcilePolicyManage` for the actual values?

v2/pkg/common/annotations/annotations.go
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package annotations
```
matthchr, line 4: minor: Consider putting each annotation + value pair into its own file. So this could be called `reconcile_policy.go`

matthchr (approved):

PR #3183, opened 2023-08-09
Make TypeNameSet generic

theunrepentantgeek:

Much of our use of `TypeNameSet` is only with `InternalTypeName`; introducing generics allows us to remove some methods from the `TypeName` interface. 

Also included moving to `InternalTypeName` where it's easy to do so. 

Changes to `TypeVisitor` and `TypeWalker` will follow in another PR, increasing coverage and removing many of the casts.

gif

**If applicable**:
- [x] this PR contains tests


theunrepentantgeek:
Abandoned in favour of a different approach.

PR #3184, opened 2023-08-10
Add two new labels

### LINKED ISSUE #3165: Deployment should have an additional label
We currently only have `control-plane: controller-manager` but we probably should also have `app: azure-service-operator-v2` on the pod too like we do for the deployment.

**Describe the improvement**
Add the new label and also use it.

matthchr:
- Add the app.kubernetes.io/name and app.kubernetes.io/version labels.
 - Remove serviceoperator.azure.com/version in favor of the standard app.kubernetes.io/version label.
 - Update taskfile to use new labels in various places.

This fixes #3165.

**If applicable**:
- [ ] this PR contains tests


super-harsh (approved):

super-harsh:
Would be good to open an issue to remove the old label at some point?

PR #3187, opened 2023-08-13
Add pod identity support for namespaces and per-resource scoped auth

### LINKED ISSUE #3167: Allow multi-tenant environments to authenticate with different Managed Identities through AAD Pod Identity
Currently, ASO allows authenticating to Azure with a user-assigned managed identity + AAD Pod Identity, but only by referencing the managed identity's client ID in the global `aso-controller-settings` Secret as documented here: https://azure.github.io/azure-service-operator/guide/authentication/credential-format/#deprecated-managed-identity-aad-pod-identity.

When trying to configure the same credentials in a per-resource Secret that looks like:
```yaml
apiVersion: v1
kind: Secret
metadata:
 name: my-resource-secret
 namespace: default
stringData:
 AZURE_SUBSCRIPTION_ID: "$AZURE_SUBSCRIPTION_ID"
 AZURE_TENANT_ID: "$AZURE_TENANT_ID"
 AZURE_CLIENT_ID: "$IDENTITY_CLIENT_ID"
```

I see an error that seems to indicate ASO is trying to authenticate with Workload Identity:
```yaml
- lastTransitionTime: "2023-07-25T19:38:36Z"
    message: |-
      WorkloadIdentityCredential authentication failed
      POST https://login.microsoftonline.com/72f988bf-86f1-41af-91ab-2d7cd011db47/oauth2/v2.0/token
      RESPONSE 401 Unauthorized
      {
        "error": "unauthorized_client",
        "error_description": "AADSTS70021: No matching federated identity record found for presented assertion. Assertion Issuer: 'https://aso-auth-test-dns-1vkelzym.hcp.westus2.azmk8s.io/'. Assertion Subject: 'system:serviceaccount:capz-system:azureserviceoperator-default'. Assertion Audience: 'api://AzureADTokenExchange'. https://docs.microsoft.com/en-us/azure/active-directory/develop/workload-identity-federation\r\nTrace ID: 5df76890-4f45-4eb2-9954-e5cd2e302400\r\nCorrelation ID: 8d3a2801-8496-4176-94aa-f4bea2755ad1\r\nTimestamp: 2023-07-25 19:38:36Z",
        "error_codes": [
          70021
        ],
        "timestamp": "2023-07-25 19:38:36Z",
        "trace_id": "5df76890-4f45-4eb2-9954-e5cd2e302400",
        "correlation_id": "8d3a2801-8496-4176-94aa-f4bea2755ad1",
        "error_uri": "https://login.microsoftonline.com/error?code=70021"
      }
      To troubleshoot, visit https://aka.ms/azsdk/go/identity/troubleshoot#workload
    observedGeneration: 1
    reason: UnknownError
    severity: Warning
    status: "False"
    type: Ready
```

**Describe the improvement**
ASO should allow resources to set a `credential-from` that refers to a Secret whose `AZURE_CLIENT_ID` refers to a managed identity.

**Additional context**
CAPZ supports this use case as documented here: https://capz.sigs.k8s.io/topics/multitenancy.html#user-assigned-managed-identity. We'd like to be able to support this with our ASO integration as well to help smooth the transition for users to Workload ID.

super-harsh:
Closes #3167

This PR adds support for allowing single-operator multi-tenant environments to authenticate with different Managed Identities through AAD Pod Identity by adding another secret variable(USE_POD_IDENTITY_AUTH) for scoped credentials to instruct the operator to use Pod Identity authentication.

**If applicable**:
- [ ] this PR contains tests


docs/hugo/content/guide/authentication/credential-format.md
```diff
@@ -494,20 +516,121 @@ stringData:
...
+apiVersion: v1
+kind: Secret
+metadata:
+ name: my-resource-secret
+ namespace: my-namespace
+stringData:
+ AZURE_SUBSCRIPTION_ID: "$AZURE_SUBSCRIPTION_ID"
+ AZURE_TENANT_ID:       "$AZURE_TENANT_ID"
+ AZURE_CLIENT_ID:       "$IDENTITY_CLIENT_ID"
+ USE_POD_IDENTITY_AUTH: "true"
```
super-harsh, line 615: We already use a boolean for `USE_WORKLOAD_IDENTITY` for global secret and omit that for scoped credentials. Suggestions on using the same here for consistency and making forcing the Managed Identity to be a default in absence of ClientSecret?

super-harsh: My thoughts on not doing the above : 
1. Users using Workload Identity would have to update their secrets to add a boolean
2. Pod Identity is deprecated, so did not want to use it as default.

theunrepentantgeek: The problem with having multiple boolean configurations like this is that we end up with oddities - what should happen if both `USER_POD_IDENTITY_AUTH` and `USE_WORKLOAD_IDENTITY` are **true** at the same time?

I'd like us to consider using a single new configuration setting and deprecating `USE_WORKLOAD_IDENTITY`. 

What if we introduced `IDENTITY_AUTH` with permitted values of `pod`, `workload` or `scoped`, with `scoped` as the default if not set. 
For back compat, `USE_WORKLOAD_IDENTITY: "true"` would change the default to `workload`.

matthchr: > The problem with having multiple boolean configurations like this is that we end up with oddities - what should happen if both USER_POD_IDENTITY_AUTH and USE_WORKLOAD_IDENTITY are true at the same time?

AFAIK we don't actually support `USE_POD_IDENTITY_AUTH` and `USE_WORKLOAD_IDENTITY_AUTH` on the same secrets.
`USE_POD_IDENTITY_AUTH` is allowed only in the scoped credential secret (namespace or per-resource). It's not allowed at the global secret level from what I can tell.

`USE_WORKLOAD_IDENTITY` is only at the global secret level, it's not supported at the per-resource/per-namespace secret (@super-harsh correct me if I am wrong here). Instead, it's always inferred if the user omits `AZURE_CLIENT_SECRET` that they want workload identity.

So there's no _technical_ ambiguity here even if there may be some user-ambiguity. It would be better if things were exactly the same between the different secret formats probably, but I do think how we got here is somewhat reasonable because:

1. The global secret used the "if no `AZURE_CLIENT_SECRET`, assume AAD Pod Identity" approach, since it previously supported AAD Pod Identity. When we added workload identity support we needed a way to differentiate between the existing supported AAD Pod Identity and Workload Identity. We possibly should've added an enum instead of a boolean here, but the choice of "evolve without breaking" seems right to me.
2. The per-namespace and per-resource secrets didn't yet exist, so didn't support AAD Pod Identity, so didn't need a boolean until now. Now, we want to add support for AAD Pod Identity in a nonbrekaing way, so we need to differentiate between workload identity (already supported, chosen by omitting the `AZURE_CLIENT_SECRET`) and this new (legacy) mode of AAD Pod Identity.

I think the direction we want to head (eventually) is that AAD Pod Identity is fully deprecated and we can drop all these flags and just have the "if client_secret omitted, workload identity" - though maybe we want an enum just to allow future evolution. I don't see how we get:
1. Global and per-namespace/per-resource secrets are consistently shaped (support the same fields/flags).
2. No breaking changes from what we have today.

Without doing something like what @super-harsh has done here (though agree maybe boolean worse than enum)

v2/internal/identity/credential_provider.go
```diff
@@ -34,6 +34,17 @@ const (
...
 
+type IdentityAuthModeOption string
+
+const (
+	podIdentity      IdentityAuthModeOption = "pod"
+	workloadIdentity IdentityAuthModeOption = "workload"
+
+	// IdentityAuthMode enum is used to determine if we're using Pod Identity or Workload Identity
+	//authentication for namespace and per-resource scoped credentials
+	IdentityAuthMode = "IDENTITY_AUTH_MODE"
```
matthchr, line 45: Wondering if it would make more sense to call this:
`AUTH_MODE` and have the values be `workloadidentity`, `podidentity`, or (possibly in the future) `sp` or `cert`? Calling it `IDENTITY_AUTH_MODE` might limit us?

v2/internal/identity/credential_provider.go
```diff
@@ -34,6 +34,17 @@ const (
...
 )
 
+type IdentityAuthModeOption string
+
+const (
+	podIdentity      IdentityAuthModeOption = "pod"
+	workloadIdentity IdentityAuthModeOption = "workload"
+
+	// IdentityAuthMode enum is used to determine if we're using Pod Identity or Workload Identity
+	//authentication for namespace and per-resource scoped credentials
```
matthchr, line 44:
```suggestion
	// authentication for namespace and per-resource scoped credentials
```

docs/hugo/content/guide/authentication/credential-format.md
```diff
@@ -494,20 +516,121 @@ stringData:
...
+```
 
-helm upgrade --install --devel aso2 aso2/azure-service-operator \
-     --create-namespace \
-     --namespace=azureserviceoperator-system \
-     --set azureSubscriptionID=$AZURE_SUBSCRIPTION_ID \
-     --set aadPodIdentity.enable=true \
-     --set aadPodIdentity.azureManagedIdentityResourceId=${IDENTITY_RESOURCE_ID} \
-     --set azureClientID=${IDENTITY_CLIENT_ID} \
-     --set crdPattern='resources.azure.com/*;containerservice.azure.com/*;keyvault.azure.com/*;managedidentity.azure.com/*;eventhub.azure.com/*'
```
theunrepentantgeek, line 510: I don't see this helm command present below this point - should it have been removed?

v2/internal/identity/credential_provider.go
```diff
@@ -300,3 +334,10 @@ func (c *credentialProvider) getSecret(ctx context.Context, namespace string, se
 func getSecretNameFromAnnotation(credentialFrom string, resourceNamespace string) types.NamespacedName {
 	return types.NamespacedName{Namespace: resourceNamespace, Name: credentialFrom}
 }
+
+func authModeOrDefault(mode string) AuthModeOption {
+	if mode == string(podIdentity) {
+		return podIdentity
+	}
+	return workloadIdentity
+}
```
theunrepentantgeek, line 343: This treats anything other than `podIdentity` as `workloadIdentity` - including `foo`, `bang`, and `oh-oh`. 

A misconfigured value should at least trigger a log warning, if not an actual error. 

I'm also a fan of being case insensitive for configuration values, though @matthchr doesn't always agree.

```suggestion
func authModeOrDefault(mode string) (AuthModeOption, error) {
	if strings.EqualFold(mode, string(podIdentity)) {
		return podIdentity, nil
	}
	
	if strings.EqualFold(mode, string(workloadIdentity)) {
		return workloadIdentity, nil
	}
	
	return "", errors.Errorf("authorization mode %q not valid", mode)
}
```

theunrepentantgeek (approved):
A nitpick on validation of configuration, but otherwise looks good. Thanks for making the changes, much appreciated.

v2/internal/identity/credential_provider.go
matthchr: > I'm also a fan of being case insensitive for configuration values, though @matthchr doesn't always agree.

I don't have a philosophical objection here one way or the other. I do think that @theunrepentantgeek is correct we should have a warning or error if the user gets it wrong. I think from a consistency perspective, being case-sensitive here might make sense given other fields (such as enums in the CRDs themselves) are also case-sensitive, and case-sensitivity seems to be "the standard" in the Kubernetes world (look at labels which are case-sensitive, or names which only allow lowercase and thus are case-sensitive).

My dislike of case-insensitivity comes primarily from REST, where attempting to be case-insensitive is awkward because JSON keys are stereotypically case-sensitive (so the expectation is not necessarily case-insensitivity) and REST entities are both set by the user and returned to the user. In the "returned to the user" scenario, case insensitivity gets awkward because you want to preserve what they've sent you, but internally you want to store it in a single canonical form. Accomplishing this ends up being:
1. More work in your backend (need to remember their case)
2. Confusing. Which of their cases do you remember? The latest PUT? Or the first PUT? Or something else?

Postel's law ends up not _really_ applying because usually in REST you have to return exactly what you accept (the expectation is that what you PUT is what you GET, basically), so you can't actually be liberal in what you accept but conservative in what you return. They are one and the same. You have to pick a winner and I tend to favor picking the easier to implement (and simpler to explain) approach and just give an error in all other cases.

v2/internal/identity/credential_provider.go
matthchr: Not actually updated?

v2/internal/identity/credential_provider.go
```diff
@@ -34,6 +34,17 @@ const (
 	FederatedTokenFilePath = "/var/run/secrets/tokens/azure-identity"
 )
 
+type AuthModeOption string
+
+const (
+	podIdentity      AuthModeOption = "pod"
```
matthchr, line 40: better to call it `podidentity` and `workloadidentity`?

docs/hugo/content/guide/authentication/credential-format.md
super-harsh: Have moved it above on[ line 455](https://github.com/Azure/azure-service-operator/pull/3187/files#diff-53bcbae618fd3a35f4071344c37ff00c04398a3c03c3bbb4965c9b72f24ee241R455)

v2/internal/identity/credential_provider.go
```diff
@@ -34,6 +34,17 @@ const (
 	FederatedTokenFilePath = "/var/run/secrets/tokens/azure-identity"
 )
 
+type AuthModeOption string
+
+const (
```
nojnhuh, line 39: Could these constants be made available to import like #3171?

theunrepentantgeek (approved):

v2/internal/identity/credential_provider.go
```diff
@@ -300,3 +326,15 @@ func (c *credentialProvider) getSecret(ctx context.Context, namespace string, se
 func getSecretNameFromAnnotation(credentialFrom string, resourceNamespace string) types.NamespacedName {
 	return types.NamespacedName{Namespace: resourceNamespace, Name: credentialFrom}
 }
+
+func authModeOrDefault(mode string) (config.AuthModeOption, error) {
+	if strings.EqualFold(mode, string(config.WorkloadIdentityAuthMode)) || mode == "" {
```
matthchr, line 331: I do think it's more consistent if we're case-sensitive (see my other comment), but approving as-is as not gonna block it on this.

matthchr (approved):

Issue #3195, opened 2023-08-18
Bug: asoctl import of StorageAccount doesn't pull all sub resources


comptonad:
**Version of Azure Service Operator**

ASO version: n/a

```sh
✗ kubectl get deployment -n azureserviceoperator-system azureserviceoperator-controller-manager -o wide
NAME                                      READY   UP-TO-DATE   AVAILABLE   AGE     CONTAINERS                IMAGES                                                                                         SELECTOR
azureserviceoperator-controller-manager   1/1     1            1           3h31m   manager,kube-rbac-proxy   mcr.microsoft.com/k8s/azureserviceoperator:v2.1.0,gcr.io/kubebuilder/kube-rbac-proxy:v0.13.1   control-plane=controller-manager
```

asoctl version:

```sh
✗ asoctl version
asoctl v2.2.0 darwin
```

I recognize the version difference as I had to upgrade `asoctl` due to another bug but I don't believe ASO version matters here.

**Describe the bug**

When running `asoctl import azure-resource "<storage-account-resource-id>"` only 1 of the sub service resources (`StorageAccountsTableService`, `StorageAccountsQueueService`, `StorageAccountsBlobService`, `StorageAccountsFileService`) is imported with it. Running it repeatedly seems random which one it gets. I actually haven't even seen `StorageAccountsFileService` after more than 2 dozen tries but I'm assuming it would be imported as a blank resource like the others.

Wanted to note that when a service is imported (e.g. StorageAccountsTableService), all of its sub resources are also imported (e.g. `StorageAccountsTableServicesTable`) as expected.

**To Reproduce**

1. Create a StorageAccount with only required fields and default values. Tested on existing created via console and new one created via ASO and behavior is the same.
2. Grab Resource ID from `JSON View` button in the console
3. Run command with said Resource ID `asoctl import azure-resources "<id>"`
4. Repeat step 3 a bunch of times

**Expected behavior**

Expect all service resources to be imported, not just 1.

**Screenshots**

Here is some example output to show different outputs. (not shown) I have gotten the same results 5 times in a row before.

```sh
➜  ✗ asoctl import azure-resource "/subscriptions/<id>/resourceGroups/<rg>/providers/Microsoft.Storage/storageAccounts/<name>" -o storageaccount.yaml
14:16:02 INF Imported kind=StorageAccount.storage.azure.com name=comptest123a
14:16:03 INF Imported kind=StorageAccountsTableService.storage.azure.com name=default
[...]

PR #3204, opened 2023-08-23
Use idiomatic Go Concurrency

theunrepentantgeek:

While working on the fix for #3195 I became concerned by the complexity of state management in `ResourceImporter` - race conditions breed in such circumstances. After discussion with @babbageclunk, I decided the fix would be to rework the concurrent aspects of the importer using native Go idioms.

gif


v2/cmd/asoctl/internal/importing/resource_importer.go
```diff
@@ -160,94 +121,220 @@ func (ri *ResourceImporter) Import(
...
+// This ensures each distinct resource is only imported once, regardless of how many times we
+// encounter it. We also proactively buffer the pending channel to avoid blocking. Any fixed size channel
+// would risk deadlock for a sufficiently large resource graph.
+func (ri *ResourceImporter) queueUniqueImporters(
+	candidates chan ImportableResource,
+	pending chan ImportableResource,
+	progress chan progressDelta,
+) {
+	seen := set.Make[string]()
+	var queue []ImportableResource
```
matthchr, line 139: I understand what you're doing here I think: making an infinitely buffered channel.

What I see is:
1. All the candidates, pending, and progress channels are unbuffered.
2. The candidate is written to the `pending` channel from the buffer in `queueUniqueImporters`. As far as I understand, this will block until `importWorker` reads from that channel.
3. `importWorker` reads the `pending` channel and writes to the `completed` channel.
4. `collateResults` reads the `completed` channel and may write back to `candidates`. This is presumably the beginning of a deadlock risk.

Because `collateResults` can block trying to write to `candidates`, but `queueUniqueImporters` may be blocked trying to write to `pending` and `importWorker` will be blocked waiting for `collateResults` to read from `completed`, we're in a deadlock.

As you allude to above, there is no amount of buffering that solves this problem as you don't know how many resources can be discovered in the importResult and must be written into `candidates`.

Is another solution to this problem to move the actual writing of the candidates in `collateResults` into a goroutine?
```
		for _, p := range importResult.pending {
			candidates <- p
		}
```
becomes:
```
		go func() {
			for _, p := range importResult.pending {
				candidates <- p
			}
		}
```

Wondering if this is cleaner than what you have here? Even if not, it might help to explain a bit more about where exactly the deadlock you're avoiding is coming from by highlighting that the problem is that once a resource is completed it tries to enqueue an arbitrarily large number of other resources into `candidates`, which may block because `queueUniqueImporters` is blocking waiting for `importWorker` which is blocking waiting for `collateResults` which is blocked waiting for `queueUniqueImporters`.

matthchr (approved):

v2/cmd/asoctl/internal/importing/resource_importer.go
```diff
@@ -160,94 +121,220 @@ func (ri *ResourceImporter) Import(
...
+	}
+}
 
-		ri.log.Error(err,
-			"Failed",
-			"kind", gk,
-			"name", rsrc.Name())
+func (ri *ResourceImporter) importWorker(
+	ctx context.Context,
+	pending chan ImportableResource,
```
super-harsh, line 177: Do these channels need to be bi-directional?

v2/cmd/asoctl/internal/importing/resource_importer.go
```diff
@@ -76,82 +74,45 @@ func (ri *ResourceImporter) AddARMID(armID string) error {
...
+	// Dedupe candidates so we import each distinct resource only once
+	go ri.queueUniqueImporters(candidates, pending, progress)
 
-				// Update our global progress too
-				processed++
-				globalBar.SetTotal(processed+int64(len(ri.queue)), false)
-				globalBar.SetCurrent(processed)
+	// Create workers to run the import
+	for i := 0; i < workers; i++ {
+		go ri.importWorker(ctx, pending, completed, progress)
```
super-harsh, line 96: It's been a while since I've worked with channels and go concurrency.

If I'm not wrong, having multiple sources pushing into a channel at the same time and not having enough capacity to read from a channel would result in data loss. Would suggest using buffered channel for progress in case of concurrent writes.

v2/cmd/asoctl/internal/importing/importable_arm_resource.go
```diff
@@ -117,23 +129,29 @@ func (i *importableARMResource) Import(ctx context.Context, bar *mpb.Bar) ([]Imp
...
-	bar.SetCurrent(1)
+	progress <- progressDelta{complete: 1, total: len(childTypes)} // resource done, children pending
 
+	// We can't abort early, need to iterate all the child types so we don't miss anything
+	var errs []error
 	for _, subType := range childTypes {
 		subResources, err := i.importChildResources(ctx, ref, subType)
-		if err != nil {
+		if ctx.Err() != nil {
+			// Aborting, don't do anything
```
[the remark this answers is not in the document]
theunrepentantgeek: Yes - and we have to ensure we process all the available child types; aborting part way through means one problematic child type prevents the import of other, problem-free, child resources.

v2/cmd/asoctl/internal/importing/resource_importer.go
theunrepentantgeek: @babbageclunk convinced me that unbuffered channels are preferable, partly because they make the code more deterministic. Adding a buffer can mask (but won't eliminate) race conditions. There's no data loss with this approach.

v2/cmd/asoctl/internal/importing/resource_importer.go
theunrepentantgeek: You've described the deadlock problem well. It boils down to the presence of a cycle in the dependency tree between goroutines.

The suggestion of buffering goroutine was made by @babbageclunk and localizes the funky bit in one place.

Termination of the import is done by tracking whether all pending work has been completed. Making this reliable requires adding all child candidates before marking the parent as complete. I suspect using a goroutine as you suggest would add a race condition causing premature exit, but haven't reasoned through this fully.

v2/cmd/asoctl/internal/importing/resource_importer.go
theunrepentantgeek: Yeah, they probably should be. Will change.

v2/cmd/asoctl/internal/importing/resource_importer.go
```diff
@@ -160,94 +121,220 @@ func (ri *ResourceImporter) Import(
...
-	// Import it
-	gk := rsrc.GroupKind()
-	name := fmt.Sprintf("%s %s", gk, rsrc.Name())
-	bar := ri.progress.AddBar(
-		0, // zero total because we don't know how much work will be needed
-		mpb.BarRemoveOnComplete(),
-		mpb.PrependDecorators(
-			decor.Name(name, decor.WCSyncSpaceR)),
-		mpb.AppendDecorators(decor.Percentage(decor.WC{W: 5})))
+// queueUniqueImporters reads from the candicates channel, putting each importer onto pending exactly once.
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

super-harsh (approved):

PR #3193, opened 2023-08-18
Pare down TypeName by using specific types

theunrepentantgeek:
**What this PR does / why we need it**:

Continuing the refactoring and cleanup.

* Change return types of some methods on `InternalTypeName` to expose more functionality
* Change owner of `ResourceType` to an `InternalTypeName`
* Restrict `InternalTypeName` to using `InternalPackageReference`
* Remove unused methods from `TypeName` and `ExternalTypeName`
* Use `ExternalTypeName` for more external references
* Remove lots of casts that are now extraneous

**Prerequisites**

- [x] PR #3191 
- [x] PR #3192 

**How does this PR make you feel**:
gif

**If applicable**:
- [x] this PR contains tests

PR #3192, opened 2023-08-17
Improve Type Definition Set

theunrepentantgeek:
**What this PR does / why we need it**:

Changes the key of `TypeDefinitionSet` to `InternalTypeName`, and updates consuming code to match.

Changes `TypeAssociation` to use `InternalTypeName` for both Key and Value, updates consuming code to match, and extends use to more locations.

Changes receiver parameter used by the `Function` interface to use an `InternalTypeName` and updates consuming code to match.

Removes `Singular()` and `Plural()` from the TypeName interface and implementations from `ExternalTypeName` as all consumption is now via `InternalTypeName`.

**Prerequisites**

- [x] #3191

**Special notes for your reviewer**:

**How does this PR make you feel**:
gif

**If applicable**:
- [x] this PR contains tests

PR #3191, opened 2023-08-17
Split TypeVisitor and use generics for context

theunrepentantgeek:
**What this PR does / why we need it**:

Splits the TypeVisitor handling of `InternalTypeName` and `ExternalTypeName` into two separate code paths. Also updates TypeWalker to match.

Changes the _context_ parameter of `TypeVisitor` from `any` (or `interface{}`) to a generic type parameter, eliminating the need for runtime type casting and allowing the compiler to enforce type safety at compile time.

Other, minor, changes include

* Using `any` instead of `interface{}` 
* Removes the superfluous internal TypeVisitor flag `visitTypeNameIsIdentity` in favour of detecting a nil delegate.
* Adds helper method `VisitInternalTypeName` for visiting an InternalTypeName directly, taking care of the resulting cast.
* Changes `APIVersion` to an `InternalTypeName`

**Special notes for your reviewer**:

This came out pretty clean.

**How does this PR make you feel**:
gif

**If applicable**:
- [x] this PR contains tests

PR #3205, opened 2023-08-23
Add InternalTypeNameSet 

theunrepentantgeek:

Creates `InternalTypeNameSet` and uses that where possible, use of `InternalTypeName` without casting.

Uses `InternalPackageReference` where it's known that an external reference is semantically incorrect, allowing use without casting.

Also includes substantial cleanup of `PackageReference`, removing methods that don't make sense at that generic level.

**Prerequisites**

- [x] #3191 
- [x] #3192 
- [x] #3193 

gif


v2/tools/generator/internal/astmodel/external_package_reference.go
```diff
@@ -64,11 +59,6 @@ func (pr ExternalPackageReference) String() string {
 	return pr.packagePath
 }
 
-// TryGroupVersion returns the group and version of this external reference.
-func (pr ExternalPackageReference) TryGroupVersion() (string, string, bool) {
-	return "", "", false
-}
-
 // GroupVersion triggers a panic because external references don't have a group or version
 func (pr ExternalPackageReference) GroupVersion() (string, string) {
```
matthchr, line 63: Does it make sense to remove this too?

v2/tools/generator/internal/astmodel/external_package_reference.go
```diff
@@ -79,8 +69,3 @@ func (pr ExternalPackageReference) ImportAlias(style PackageImportStyle) string
 	msg := fmt.Sprintf("cannot create import alias for external package reference %s", pr.packagePath)
```
matthchr, line 69: and this?

v2/tools/generator/internal/astmodel/internal_type_name_set.go
```diff
@@ -0,0 +1,116 @@
...
+
+package astmodel
+
+import (
+	"fmt"
+)
+
+// InternalTypeNameSet stores type names in no particular order without
+// duplicates.
+type InternalTypeNameSet map[InternalTypeName]struct{}
```
matthchr, line 14: Is there value in this (and the current `type_name_set.go` over and aobve what we already have in `v2/internal/set/set.go`? Wondering if we can just do do away with all of them in favor of the generic one?

v2/tools/generator/internal/astmodel/package_reference.go
```diff
@@ -28,22 +28,11 @@ type PackageReference interface {
 	// IsPreview returns true if this package reference has a suffix indicating it's a preview
 	// release, false otherwise
 	IsPreview() bool
-	// TryGroupVersion returns the group and version of this reference.
-	// Returns true if the reference has a group and version, false otherwise.
-	TryGroupVersion() (string, string, bool)
-	// GroupVersion returns the group and version of this reference, triggering a panic if they aren't available
-	GroupVersion() (string, string)
 	// ImportAlias returns the import alias to use for this package reference
 	ImportAlias(style PackageImportStyle) string
```
matthchr, line 32: I think this `ImportAlias` function is also specific to the `InternalPackageReference`, so consider removing from here possibly?

v2/tools/generator/internal/astmodel/package_reference.go
```diff
@@ -28,22 +28,11 @@ type PackageReference interface {
 	// IsPreview returns true if this package reference has a suffix indicating it's a preview
 	// release, false otherwise
 	IsPreview() bool
```
matthchr, line 30: This is also (at least with our current implementation) specific to internal refs, but possibly makes sense to leave here as it could in theory apply to external refs one day too...

v2/tools/generator/internal/astmodel/type_definition_test.go
```diff
@@ -33,8 +33,7 @@ func Test_MakeTypeDefinition_GivenValues_InitializesProperties(t *testing.T) {
 	g.Expect(objectDefinition.Name().Name()).To(Equal(name))
 	g.Expect(objectDefinition.Type()).To(Equal(objectType))
 
-	actualGroup, actualVersion, ok := objectDefinition.Name().PackageReference().TryGroupVersion()
-	g.Expect(ok).To(BeTrue())
+	actualGroup, actualVersion := objectDefinition.Name().InternalPackageReference().GroupVersion()
```
matthchr, line 36: minor: Not sure if this was changed in this PR or an earlier one and I just noticed it, but IMO it'd be OK to call this method `PackageReference()` and just have it return an `InternalPackageReference`, since all internal type names by definition would have to be internal anyway, having the extra clarification in the method name doesn't necessarily disambiguate anything.

matthchr (commented):
Not approving yet primarily because of the outstanding question about `Set[T]` vs `InternalTypeNameSet`.

v2/tools/generator/internal/astmodel/external_package_reference.go
theunrepentantgeek: If memory serves, there were a couple of outstanding references - and I thought this PR had gotten large enough. Removing this is a goal, but not in this PR.

v2/tools/generator/internal/astmodel/external_package_reference.go
theunrepentantgeek: As above; will go, but not in this PR.

v2/tools/generator/internal/astmodel/internal_type_name_set.go
theunrepentantgeek: I started doing that and the scope of changes grew a bit too much for a single PR. It's a good idea, and one we should pursue as a follow-up.

v2/tools/generator/internal/astmodel/package_reference.go
theunrepentantgeek: I'll do a check to see if any additional methods can be removed, but I suspect there are lingering references.

v2/tools/generator/internal/astmodel/type_definition_test.go
theunrepentantgeek: Can't use co-variance on interface implementations in Go. 

The `PackageReference()` method on the `TypeName` interface must return a `PackageReference`, so I had to add a different method to return an `InternalPackageReference`. 

We may be able to get rid of `TypeName.PackageReference()` later, I'm not sure.

matthchr (approved):
Approving on the understanding we'll look into swapping to `Set[T]` later.

PR #3206, opened 2023-08-24
Support new versions of cache RP

matthchr:
**If applicable**:
- [x] this PR contains tests


theunrepentantgeek (approved):

v2/samples/cache/v1api/v1api20230401_redis.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: cache.azure.com/v1api20230401
```
super-harsh, line 1: I'd suggest separating these samples out in their own folders like `v1api20230401` and `v1api20201201`

v2/internal/controllers/crd_cache_redis_20201201_test.go
```diff
@@ -0,0 +1,188 @@
...
+	tc.DeleteResourcesAndWait(redis1, redis2)
+
+	// Ensure that the resource was really deleted in Azure
+	exists, retryAfter, err := tc.AzureClient.HeadByID(tc.Ctx, armId, string(cache.APIVersion_Value))
+	tc.Expect(err).ToNot(HaveOccurred())
+	tc.Expect(retryAfter).To(BeZero())
+	tc.Expect(exists).To(BeFalse())
+}
+
+func makeRedis20201201(tc *testcommon.KubePerTestContext, rg *resources.ResourceGroup, prefix string) *cache.Redis {
```
super-harsh, line 83: `newRedis20201201` maybe?

super-harsh (approved):
LGTM! just a couple minors comments.

v2/samples/cache/v1api/v1api20230401_redis.yaml
matthchr: Not every group does this actually, and because there's a bunch of API versions coexisting (because of redis + redisenterprise) it felt better to put it all into 1 folder like we do for compute so that people don't have to guess which API versions support which resources. Basically I think people are resource-centric and making them choose API versions first doesn't help in sample discovery. It's fine when every resource is in every API version (pick any one and you're right), but in this case not every resource is in every APIversion so it's harder.

v2/internal/controllers/crd_cache_redis_20201201_test.go
matthchr: Fixed.

PR #3222, opened 2023-09-01
Make helm chart version SemVer 2 compatible

### LINKED ISSUE #3189: Publish chart versions that are SemVer 2 compatible
Currently chart versions are being published with a prefix, "v" (example v2.2.0) which is not SemVer 2 compatible.  Such charts like aad-pod-identity are not including "v" in the version.

**Describe the improvement**
Consider dropping the "v" from publish charts so that the helm manifest provides SemVer 2 compatible versions.

**Additional context**
I believe the change could be as straightforward as changing - https://github.com/Azure/azure-service-operator/blob/main/v2/charts/azure-service-operator/Chart.yaml#L3 - to `version: 2.2.0`

super-harsh:
Closes #3189 

This PR removes `v` from chart versions to make ASOv2 helm chart version SemVer 2 compatible.
The change is made in the `generate-helm-manifest.sh` so that the next time we release the helm chart, we'll have the helm chart version in a new format.

Tested it out and upgrades work fine : 
image

### DIFF
--- a/scripts/v2/generate-helm-manifest.sh
+++ b/scripts/v2/generate-helm-manifest.sh
@@ -40,7 +40,7 @@ ${SCRIPT_DIR}/kustomize-build.sh -v "$VERSION" -k operator -o "$GEN_FILES_DIR"
 rm "$GEN_FILES_DIR"/*_namespace_* # remove namespace as we will let Helm manage it
 
 # Chart replacements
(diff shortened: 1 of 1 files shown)


theunrepentantgeek (approved):

matthchr (approved):
You tested it should have no issue w/ 2.3.0 coexisting with `v2.2.0`?

PR #3245, opened 2023-09-01
Support arbitrary ARM ID owners

### LINKED ISSUE #2357: Allow owners to arbitrary ARM IDs
Currently, the type of the owner field is `ArbitraryOwnerReference`. It's only support another resource defined in the cluster.

**Describe the improvement**
It would be great if I could define an `armId` to be able to add a role to an existing resource outside of the cluster.

Thanks

matthchr:
This closes #2357.

Nearly all resources now support being owned by an arbitrary ARM ID. Only database user resources (MySQLUser, PostgreSQLUser) do not allow ARM ID ownership. This is because those resources must extract certain connection information from their parents and so must have access to their parents via the Kubernetes API.

ARM ID ownership behaves the following way:
 * Resources owned by an ARM ID will continue to attempt to reconcile if that ID doesn't exist. When the owning resource cannot be found, the resource will report a Ready=false with a "Warning" and additional details. If possible, avoid deleting the owning ARM resource without also deleting the resources in Kubernetes.
 * The credential being used for the resource must have the same subscription ID as the owning ARM resource. This means if your global credential is for Subscription A you cannot have ARM ID owners from subscription B unless you also create a serviceoperator.azure.com/credential-from for subscription B.
 * Resources that set both owner.name and owner.armID will be rejected by webhook.
 * Resources that attempt to set an ARM ID of the "wrong kind" (should be a storage account, gave it a virtual machine) will be rejected by webhook.

Documentation will come in a follow-up PR.

**If applicable**:
- [x] this PR contains tests


v2/internal/resolver/resource_hierarchy.go
```diff
@@ -173,13 +266,56 @@ func (h ResourceHierarchy) fullyQualifiedARMIDImpl(subscriptionID string, origin
...
+	}
+
+	provider, rootResourceTypes, err := getResourceTypeAndProvider(resource)
+	if err != nil {
+		return err
+	}
+	resourceTypesIncludedInARMID := rootResourceTypes[:len(rootResourceTypes)-1]
+
+	// Ensure that the ARM ID actually has a suffix containing the resource types we expect
+	// TODO: make this a webhook instead (or in addition to?)
```
matthchr, line 291: Need to do this

v2/tools/generator/internal/codegen/testdata/ArmResource/Arm_oneof_resource_conversion_on_arm_type_only_azure.golden
```diff
@@ -173,7 +169,16 @@ func (resource *FakeResource) updateValidations() []func(old runtime.Object) (ad
 		func(old runtime.Object) (admission.Warnings, error) {
 			return resource.validateResourceReferences()
 		},
-		resource.validateWriteOnceProperties}
+		resource.validateWriteOnceProperties,
+		func(old runtime.Object) (admission.Warnings, error) {
+			return resource.validateOwnerReference()
+		},
```
theunrepentantgeek, line 175: Does this need to be wrapped? or, could you use `resource.validateOwnerReference` directly?

v2/tools/generator/internal/functions/kubernetes_admissions_validations.go
```diff
@@ -92,6 +101,42 @@ func validateResourceReferencesBody(codeGenerationContext *astmodel.CodeGenerati
...
+	receiverType := receiver.AsType(codeGenerationContext)
+
+	fn := &astbuilder.FuncDetails{
+		Name:          methodName,
+		ReceiverIdent: receiverIdent,
+		ReceiverType:  astbuilder.PointerTo(receiverType),
+		Body:          validateOwnerReferencesBody(codeGenerationContext, receiverIdent),
+	}
+
+	fn.AddReturn(astbuilder.QualifiedTypeName(codeGenerationContext.MustGetImportedPackageName(astmodel.ControllerRuntimeAdmission), "Warnings"))
```
theunrepentantgeek, line 115: Prefer putting the call to `MustGetImportedPackageName()` onto a separate line assigning a new variable, largely so any stack trace directly isolates the relevant call.

v2/pkg/genruntime/core/errors.go
```diff
@@ -187,6 +187,40 @@ func (e *ConfigMapNotFound) Format(s fmt.State, verb rune) {
...
+		cause: cause,
+	}
+}
+
+var _ error = &SubscriptionMismatch{}
+var _ causer = &SubscriptionMismatch{}
+
+func (e *SubscriptionMismatch) Error() string {
+	return e.cause.Error()
+}
```
theunrepentantgeek, line 206: Does `SubscriptionMismatch` need a custom error message as well?

v2/pkg/genruntime/kubernetes_resource.go
```diff
@@ -91,3 +93,17 @@ func GetAPIVersion(metaObject ARMMetaObject, scheme *runtime.Scheme) (string, er
 
 	return rsrc.GetAPIVersion(), nil
 }
+
+// GetResourceTypeAndProvider returns the provider and the array of resource types which represent the resource.
+// For example: Microsoft.Compute/virtualMachineScaleSets would return ("Microsoft.Compute", []string{"virtualMachineScaleSets"}, nil)
+func GetResourceTypeAndProvider(res ARMMetaObject) (string, []string, error) {
```
theunrepentantgeek, line 99: I think we already have this code somewhere. We should either reuse the original, or change the other case to use this one. Maybe as a separate PR though.

v2/internal/resolver/resolver.go
```diff
@@ -177,14 +183,64 @@ func (r *Resolver) ResolveReference(ctx context.Context, ref genruntime.Namespac
...
+	OwnerNotExpected = ResolveOwnerResult("OwnerNotExpected")
+)
+
+type OwnerDetails struct {
+	Result ResolveOwnerResult
+	Owner  genruntime.ARMMetaObject
+	ARMID  string
+}
+
+func (det OwnerDetails) FoundKuberentesOwner() bool {
```
theunrepentantgeek, line 203: Spelling

```suggestion
func (det OwnerDetails) FoundKubernetesOwner() bool {
```

v2/internal/resolver/resolver.go
```diff
@@ -177,14 +183,64 @@ func (r *Resolver) ResolveReference(ctx context.Context, ref genruntime.Namespac
...
+		Result: OwnerNotExpected,
+	}
+}
+
 // ResolveOwner returns the MetaObject for the given resource's owner. If the resource is supposed to have
 // an owner but doesn't, this returns an ReferenceNotFound error. If the resource is not supposed
-// to have an owner (for example, ResourceGroup), returns nil.
-func (r *Resolver) ResolveOwner(ctx context.Context, obj genruntime.ARMOwnedMetaObject) (genruntime.ARMMetaObject, error) {
+// to have an owner (for example, ResourceGroup) or the owner points to a raw ARM ID, returns nil.
+func (r *Resolver) ResolveOwner(ctx context.Context, obj genruntime.ARMOwnedMetaObject) (OwnerDetails, error) {
```
theunrepentantgeek, line 235: Seems odd that `ResolveOwner` returns nil even if it has an owner, just because it's an ARM ID.

Maybe rename to `ResolveClusterOwner()` or `ResolveKubernetesOwner()`?

v2/internal/resolver/resource_hierarchy.go
```diff
@@ -151,14 +176,82 @@ func (h ResourceHierarchy) fullyQualifiedARMIDImpl(subscriptionID string, origin
 		// Join them together
 		interleaved := genruntime.InterleaveStrSlice(resourceTypes, azureNames)
 		return genericarmclient.MakeTenantScopeARMID(provider, interleaved...)
+	case ResourceHierarchyRootARMID:
+		// TODO: Possibly refactor this huge method into sub-functions?
```
theunrepentantgeek, line 180: Good idea.

theunrepentantgeek (approved):
I think I understand what's going on here; will come back for another read through but don't want to block.

v2/tools/generator/internal/codegen/testdata/ArmResource/Arm_oneof_resource_conversion_on_arm_type_only_azure.golden
matthchr: It's wrapped to discard the `old` parameter because it's not needed but I wanted to share a single impl for update/create. Since the signature matches for create it's called directly:
```
// createValidations validates the creation of the resource
func (zone *DnsZone) createValidations() []func() (admission.Warnings, error) {
	return []func() (admission.Warnings, error){zone.validateResourceReferences, zone.validateOwnerReference}
}
```

but for update because the signature doesn't match exactly it's wrapped. This is the same behavior as w/ `validateResourceReferences` so I think it makes sense to do the same thing here. LMK if you see a cleaner way.

v2/pkg/genruntime/kubernetes_resource.go
matthchr: Not sure if you saw but I just moved it from where it was elsewhere. Previously it was unexpected in `resource_hierarchy.go`. Unless you're saying it's _already_ in two different places in the code? One in `resource_hierarchy.go` and one someplace else?

v2/internal/resolver/resolver.go
matthchr: Ah it doesn't actually return `nil` anymore. I refactored the method and added the `OwnerDetails` return type because I also didn't like this. Now it always has a return and that return has the `Result` set to either:
1. `OwnerFoundKubernetes` (if the owner was found in k8s)
2. `OwnerFoundARM` (if the owner was found in ARM)
3. `OwnerNotExpected` (if the owner was not found but also not expected)
4. an error (if an owner was not found but _was_ expected).

matthchr: I've fixed this comment.

v2/pkg/genruntime/core/errors.go
matthchr: I've changed how this is implemented so that it makes more sense. The error format is now part of this error type rather than just hardcoded in the file that called `NewSubscriptionMismatchError`

v2/internal/resolver/resource_hierarchy.go
matthchr: In the interest of time I'll file an item for this and do it later.

matthchr: See #3254

PR #3250, opened 2023-09-03
Update Generator Dependencies

theunrepentantgeek:

Dependabot highlighted a bunch of outdated dependencies, but doesn't produce PRs that build (they all need a go mod tidy pass), plus updating things one at a time is time consuming.

This PR updates all dependencies for generator.

gif


v2/tools/generator/internal/codegen/pipeline/export_generated_code.go
```diff
@@ -8,13 +8,14 @@ package pipeline
...
 	"fmt"
-	"golang.org/x/exp/slices"
-	"golang.org/x/sync/errgroup"
 	"os"
 	"path/filepath"
 	"sync"
 	"time"
 
+	"golang.org/x/exp/slices"
+	"golang.org/x/sync/errgroup"
```
matthchr, line 17: minor: I think this is correct since these aren't stdlib imports, but they can be grouped with the group below. (there should just be 3 groups total AFAIK: stdlib, external, internal)

matthchr (approved):

PR #3258, opened 2023-09-05
Update procedure for a new release

theunrepentantgeek:

While preparing the v2.3.0 release, I discovered some of our release docs were stale; this PR updates them.

Biggest change: we no longer publish a CRD Helm chart, so I switched to comparing the CRD YAML files. 

gif

**If applicable**:


docs/hugo/content/contributing/create-a-new-release.md
```diff
@@ -17,21 +22,88 @@ title: Creating a new release
...
+   wget https://github.com/Azure/azure-service-operator/releases/download/v2.2.0/azureserviceoperator_v2.2.0.yaml
+   ```
+
+2. Download the CRD definitions for the new release (in this example, v2.3.0):
+
+   ``` bash
+   wget https://github.com/Azure/azure-service-operator/releases/download/v2.3.0/azureserviceoperator_v2.3.0.yaml
+   ```
+
+   This will only be available when the GitHub action trigged by publishing the release has finished.
```
matthchr, line 37: We should do the diff against a local copy rather than the published copy IMO, since the intent is to catch diffs we didn't expect, and/or help crafting the release notes, both of which you have to do before the release itself.

I think you can do `task controller:run-kustomize` and then look at the `bin/azureserviceoperator_customresourcedefinitions_{{.VERSION}}.yaml` file.

That file will _only_ have the CRDs, so it compares with the `azureserviceoperator_customresourcedefinitions_v2.2.0.yaml` included in each release.

matthchr (approved):

docs/hugo/content/contributing/create-a-new-release.md
```diff
@@ -17,21 +22,88 @@ title: Creating a new release
...
+   kubectl get all -n azureserviceoperator-system
+   ```
+
+8. Create a resource group and a vnet in it (the vnet is to check that conversion webhooks are working, since there aren't any for RGs):
+
+   ``` bash
+   kubectl apply -f v2/samples/resources/v1api/v1api20200601_resourcegroup.yaml
+   kubectl apply -f v2/samples/network/v1api20201101/v1api20201101_virtualnetwork.yaml
    ```
 8. Make sure they deploy successfully - check in the portal as well.
```
matthchr, line 109: Can you also fix up the "Creating and testing Helm chart for new release" to refer to CRDpattern in the install step?
I think it just needs the default CRD pattern applied... maybe with networking too?

theunrepentantgeek: Done.

PR #3274, opened 2023-09-11
Move storage types into sub-packages

theunrepentantgeek:

Our storage types are essentially an implementation detail, but at the moment they're exposed at the same level as all our API types. Moving these into subpackages makes the API cleaner, removing distraction for those directly consuming our API.

Only `network` resources are actually relocated in this PR, partly to keep the change size down, and so we can prove the relocation works. I'll do followup PRs once this is merged, each relocating a small number of groups.

Includes some fixes to make this work:
* Changing areas that assumed group/version was the same as the package name
* We were missing the `// +versionName` tag in our `groupversion_info_gen.go` files. Without that, `controller-gen` assumes you want the package name as the version, which no longer holds.

gif

**If applicable**:
- [x] this PR contains tests


v2/api/network/v1api20220401/storage/groupversion_info_gen.go
```diff
@@ -5,12 +5,13 @@ Licensed under the MIT license.
 
 // Code generated by azure-service-operator-codegen. DO NOT EDIT.
 
-// Package v1api20220401storage contains API Schema definitions for the network v1api20220401storage API group
```
matthchr, line 8: minor: Should these comments continue to specify the version in them?

v2/tools/generator/internal/functions/get_extended_resources_function.go
```diff
@@ -6,10 +6,9 @@
 package functions
 
 import (
-	"reflect"
-	"sort"
-
 	"github.com/dave/dst"
+	"golang.org/x/exp/slices"
```
matthchr, line 10: minor: import ordering

v2/tools/generator/internal/astmodel/package_definition.go
```diff
@@ -175,6 +178,19 @@ func (p *PackageDefinition) writeTestFile(
 	return nil
 }
 
+func (p *PackageDefinition) createVersion(ref InternalPackageReference) string {
```
matthchr, line 181: minor: No actual usage of `p` receiver in this function, you could instead make this a pure function and change `NewPackageDefinition` above to do the following:
```
	return &PackageDefinition{
		PackageName: ref.PackageName(),
		GroupName:   ref.Group(),
		Version: createVersion(ref),
		Path:        ref.FolderPath(),
		definitions: make(TypeDefinitionSet),
	}
```

or, you could make this a method on `InternalPackageReference`, though I know it's an interface so I guess you can't easily put a method on it (unless you promoted Version() to be on the interface). You could make it a public `astmodel` method instead if you prefer: `PackageVersion(ref InternalPackageReference)` or similar as well.

It feels like it doesn't really make sense on `PackageDefinition` though., and having it on PackageDefinition means you end up having to do the awkward "construct a partial package definition, then fill the rest out" in the `New` function, which is _fine_ (it is `New` after all so the contract is just that it's fully specified when it returns) but it still feels more awkward than it needs to be.

matthchr (approved):

v2/tools/generator/internal/astmodel/package_definition.go
theunrepentantgeek: I prefer it with the receiver, because that acts as a namespace to restrict access. Moving it to a pure function is inviting use from other locations, which is inappropriate.

v2/api/network/v1api20220401/storage/groupversion_info_gen.go
theunrepentantgeek: I'll tweak to use the folder path in a followup PR:

``` go
// Package v1api20220401/storage contains API Schema definitions for the network v1api20220401/storage API group
```
or something similar.

PR #3286, opened 2023-09-12
Treat CustomKubeletIdentityMissingPermissionError as retryable

### LINKED ISSUE #3267: Have ManagedClusters reconcile again when hitting error code CustomKubeletIdentityMissingPermissionError
When creating `UserAssignedIdentity`, `RoleAssignment`, and a `ManagedCluster` all at once that reference each other the resources can become deadlocked. The only remedy is manual intervention to modify the `ManagedCluster` or to delete the aso pod to force another reconcile.

Here is the status of the `ManagedCluster` when this happens

```
status:
  conditions:
  - lastTransitionTime: "2023-09-06T21:24:00Z"
    message: |
      The cluster using user-assigned managed identity must be granted 'Managed Identity Operator' role to assign kubelet identity. You can run 'az role assignment create --assignee <control-plane-identity-principal-id> --role 'Managed Identity Operator' --scope <kubelet-identity-resource-id>' to grant the permission. See https://learn.microsoft.com/en-us/azure/aks/use-managed-identity#add-role-assignment: PUT https://management.azure.com/subscriptions/{subid}/resourceGroups/{rgname}providers/Microsoft.ContainerService/managedClusters/{clustername}
      RESPONSE 400: 400 Bad Request
      ERROR CODE: CustomKubeletIdentityMissingPermissionError
      {
        "code": "CustomKubeletIdentityMissingPermissionError",
        "details": null,
        "message": "The cluster using user-assigned managed identity must be granted 'Managed Identity Operator' role to assign kubelet identity. You can run 'az role assignment create --assignee \u003ccontrol-plane-identity-principal-id\u003e --role 'Managed Identity Operator' --scope \u003ckubelet-identity-resource-id\u003e' to grant the permission. See https://learn.microsoft.com/en-us/azure/aks/use-managed-identity#add-role-assignment",
        "subcode": ""
      }
    observedGeneration: 1
    reason: CustomKubeletIdentityMissingPermissionError
    severity: Error
    status: "False"
    type: Ready
```

**Describe the improvement**
When the specific code `CustomKubeletIdentityMissingPermissionError` occurs retry the reconcile function. In this case, the `RoleAssignment` granting the necessary permissions just needed additional time to propagate.

**Additional context**
In the above scenario we're apply a large number of resources via flux and ideally they would eventually all reconcile into their ready states. This specific case is for creating a cluster with workload identity enabled. Will note that this is still not entirely possible without #2950.

theunrepentantgeek:

Treat the error `CustomKubeletIdentityMissingPermissionError` as a retryable error, to avoid a race condition where slow creation of a `UserAssignedIdentity` might cause creation of a `ManagedCluster` to fail with a permanent error.

Closes #3267 

gif

### DIFF
--- a/v2/api/containerservice/customizations/managed_cluster_extensions.go
+++ b/v2/api/containerservice/customizations/managed_cluster_extensions.go
@@ -8,7 +8,6 @@ package customizations
 import (
 	"context"
 	"fmt"
-	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/core"
 	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice"
@@ -25,6 +24,7 @@ import (
 	"github.com/Azure/azure-service-operator/v2/internal/resolver"
 	"github.com/Azure/azure-service-operator/v2/internal/set"
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/core"
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/extensions"
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/secrets"
 )
@@ -201,7 +201,6 @@ func (ext *ManagedClusterExtension) ClassifyError(
 		return core.CloudErrorDetails{}, err
 	}
 
-	// Override is to treat Conflict as retryable for Redis, if the message contains "try again later"
 	if isRetryableClusterError(cloudError) {
 		details.Classification = core.ErrorRetryable
 	}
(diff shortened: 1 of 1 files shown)


v2/api/containerservice/customizations/managed_cluster_extensions.go
```diff
@@ -8,6 +8,7 @@ package customizations
 import (
 	"context"
 	"fmt"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/core"
```
matthchr, line 11: minor: ordering

v2/api/containerservice/customizations/managed_cluster_extensions.go
```diff
@@ -181,3 +182,43 @@ func clusterProvisioningStateBlocksReconciliation(provisioningState *string) boo
...
+	apiVersion string,
+	log logr.Logger,
+	next extensions.ErrorClassifierFunc,
+) (core.CloudErrorDetails, error) {
+	details, err := next(cloudError)
+	if err != nil {
+		return core.CloudErrorDetails{}, err
+	}
+
+	// Override is to treat Conflict as retryable for Redis, if the message contains "try again later"
```
matthchr, line 204: comment is wrong

matthchr (approved):

PR #3326, opened 2023-09-20
Add support for TrafficManagerProfile and TrafficManagerProfileEndpoints

### LINKED ISSUE #1265: Feature: Support Traffic Manager Profiles and Traffic Manager Endpoints
Added support for Traffic Manager Profiles, Traffic Manager Endpoints and DNS zone CNAME records would be great.
This way the service operator would be able to operate it's cluster in a "self sustained" load balanced environment (ie. failover clusters, load multi-region deployments and similar scenarios) and act as a replacement for the ExternalDNS component often used for single cluster installations.

As the traffic manager profiles obviously will be shared with other Kubernetes clusters (and/or other webserveres) this needs to be taken into consideration during provisioning. We could easily get a situation where multiple clusters (with their own operators) might try to create/update the same profile. The same applys to the DNS CNAME record which would point to a single traffic manager profile.

Each cluster should have it's own endpoint configuration - pointing to the ingress IP of choice. Being able to automatically fetch this from a reference to a namespace+ingress would be a huge time saver in terms of usability and generalization.

super-harsh:
Closes #1265 

This PR adds support for TrafficManagerProfile, TrafficManagerProfilesNestedEndpoints, TrafficManagerProfilesAzureEndpoints, and TrafficManagerProfilesExternalEndpoints.

**If applicable**:
- [x] this PR contains tests


v2/internal/controllers/crd_networking_trafficmanagerprofile_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+	// Perform a simple patch
+	old := tmp.DeepCopy()
+	tmp.Spec.Tags = map[string]string{
+		"foo": "bar",
+	}
+	tc.PatchResourceAndWait(old, tmp)
+	tc.Expect(tmp.Status.Tags).To(HaveKey("foo"))
+
+	tc.RunSubtests(
```
matthchr, line 66: Can we use parallelSubtests here? Or no?

matthchr (approved):

PR #3339, opened 2023-09-25
Move CRD tests to better location

theunrepentantgeek:

All our CRD tests were accumulating in the `controllers` package when they weren't really about the controllers, they were tests of the custom resources themselves.

This PR moves them out into a dedicated folder that's focussed precisely on testing.

gif

**If applicable**:
- [x] this PR contains tests


v2/internal/crdtests/aks_managedcluster_20230102preview_test.go
```diff
@@ -3,7 +3,7 @@ Copyright (c) Microsoft Corporation.
 Licensed under the MIT license.
 */
 
-package controllers_test
+package crdtests
```
matthchr, line 6: minor: I _think_ we want this to be a `_test` package as the compiler treats those specially

v2/internal/controllers/user_assigned_identities_test.go
```diff
@@ -6,13 +6,15 @@ Licensed under the MIT license.
 package controllers_test
 
 import (
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
```
matthchr, line 9: ordering?

v2/internal/controllers/compute_test.go
```diff
@@ -0,0 +1,188 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+package controllers_test
+
+import (
+	"fmt"
+	"github.com/Azure/azure-service-operator/v2/internal/util/to"
```
matthchr, line 10: minor: ordering

v2/internal/controllers/user_assigned_identities_test.go
```diff
@@ -111,3 +113,26 @@ func Test_MySQLServerWithKubernetesReferenceUserAssignedIdentity(t *testing.T) {
 		tc.Expect(armID).To(testcommon.EqualsIgnoreCase(*mi.Status.Id))
 	}
 }
+
+func createPasswordSecret(
```
matthchr, line 117: Why not move this test?

v2/internal/controllers/network_test.go
```diff
@@ -0,0 +1,120 @@
...
+import (
+	network "github.com/Azure/azure-service-operator/v2/api/network/v1api20201101"
+	resources "github.com/Azure/azure-service-operator/v2/api/resources/v1api20200601"
+	"github.com/Azure/azure-service-operator/v2/internal/genericarmclient"
+	"github.com/Azure/azure-service-operator/v2/internal/testcommon"
+	"github.com/Azure/azure-service-operator/v2/internal/util/to"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+)
+
+func newLoadBalancerForVMSS(
```
matthchr, line 17: Is this only exposing helper methods now? Why not move this? Who is consuming these?

v2/internal/controllers/compute_test.go
```diff
@@ -0,0 +1,188 @@
...
+
+	compute2020 "github.com/Azure/azure-service-operator/v2/api/compute/v1api20201201"
+	network "github.com/Azure/azure-service-operator/v2/api/network/v1api20201101"
+	resources "github.com/Azure/azure-service-operator/v2/api/resources/v1api20200601"
+
+	"github.com/Azure/azure-service-operator/v2/internal/testcommon"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+)
+
+func newVirtualMachine20201201(
```
matthchr, line 23: Is this only exposing helper methosd now? Why not move this too?

super-harsh: Makes sense to have them helper methods in the test files

super-harsh:
IMO, we can consider moving all the tests and recordings under `internal/controllers` to `internal/controllers/controllers/controllers_tests`

super-harsh:
We might also need to make changes to Taskfile(for running tests from the directory) and Documentation [here](https://azure.github.io/azure-service-operator/contributing/add-a-new-code-generated-resource/#write-a-crud-test-for-the-resource) and other places in the same doc.

v2/internal/crdtests/aks_managedcluster_20230102preview_test.go
theunrepentantgeek: The only special behaviour that I'm aware of is that a `_test` package can coexist in the same folder as a normal one. This is useful to stop tests accessing internals, etc; not something we need here.

I'll fix the name.

v2/internal/controllers/user_assigned_identities_test.go
theunrepentantgeek: I moved all the files that adhered to our naming convention (ie `crd_*.go`); can move other files if neede, but I'd like to do that in a later PR.

v2/internal/controllers/compute_test.go
theunrepentantgeek: Strong disagree. A helper belongs in a test file if it's only used by that test - having tests cross-wired makes things brittle as changes can be made for one test that inadvertently impact on other tests. These helpers are in separate methods to help convey that they're used from multiple places.

v2/internal/controllers/network_test.go
theunrepentantgeek: After moving the majority of the tests, I discovered some cross-wiring where other tests were mis-reusing helper methods that should have been private to other tests. I brought back into this package just the required helpers.

v2/internal/controllers/user_assigned_identities_test.go
matthchr: I think every test in this folder should move. This folder isn't _just_ CRD tests it's all end to end tests and I think they should be together regardless of if they are for CRUD for a single CRD, testing edge cases, testing other features (configmap export, etc).

To that effect a better folder name might be `endtoendtests` but I'm fine w/ crdtests too.

matthchr: Otherwise if we don't move (as @super-harsh pointed out) it makes the taskfile more complicated because the envtest tests are now split across directories and we need to run both. It removes some of the parallelism we've added within an envtest run, because if split across directories two tests that share the same envtest configuration will not run on the same envtest instance but instead 2 instances.

In fact, they won't run on 2 instances in parallel today unless we modify the taskfile to launch both `go test` commands in parallel, which IIRC we don't currently do for envtest runs to avoid noisy neighbor effects.

tl;dr: IMO we should move every test in this folder or none of them, not partially. They are all the same type of test (end to end). Some are CRD CRUD and some are other types of end to end tests but they belong together.

v2/internal/controllers/user_assigned_identities_test.go
theunrepentantgeek: Fixed.

v2/internal/controllers/compute_test.go
theunrepentantgeek: Fixed.

v2/internal/controllers/compute_test.go
matthchr: I don't mind the helpers being in their own file, but:
1. I thought you can't import functions from `_test` files from other packages, so if these need to be used in the other directory they cannot be, can they? In fact, I think in this PR now these same functions are duplicated in both places (old folder and new folder), which feels bad?
2. The name of the file they're in now suggests to me that it contains tests, when it actually doesn't. TBH I am not sure that there is a good way to solve this other than pulling them out into their own package `testcommon` or equivalent. I'm not sure that's worth doing so willing to live with this.

Item 1 above (especially the fact that these are now duplicated) is awkward - again suggesting that we should move the whole folder not just part of it?

theunrepentantgeek:
Abandoning, we'll leave these tests where they are.

PR #3348, opened 2023-09-28
Add reflection helper SetProperty

theunrepentantgeek:

As a part of writing an extension for goal-seeking KeyVaults, I need the ability to set a specific property via reflection. 
I've pulled this out into a separate helper with thorough testing.

gif

**If applicable**:
- [x] this PR contains tests


v2/internal/reflecthelpers/reflect_helpers.go
```diff
@@ -282,3 +282,74 @@ func getItemsField(listPtr client.ObjectList) (reflect.Value, error) {
...
 	return itemsField, nil
 }
+
+// SetProperty sets the property on the provided object to the provided value.
+// obj is the object to modify.
+// propertyPath is a dot-separated path to the property to set.
+// value is the value to set the property to.
+// Returns an error if any of the properties in the path do not exist, if the property is not settable,
+// or if the value provided is incompatible.
+func SetProperty(obj any, propertyPath string, value any) error {
```
matthchr, line 292: There might now be some duplicate (less generic version) of this code in the `setChildCollection` methods of some of the extensions.

Not clear it's actually fully gonna work with this because it needs to construct items in a collection but figured I'd highlight it anyway.

v2/internal/reflecthelpers/reflect_helpers.go
```diff
@@ -282,3 +282,74 @@ func getItemsField(listPtr client.ObjectList) (reflect.Value, error) {
...
+		subject = subject.Elem()
+	}
+
+	// Check we have a struct
+	if subject.Kind() != reflect.Struct {
+		return errors.Errorf("provided object was not a struct, was %s", subject.Kind())
+	}
+
+	// Get the field we need to modify
+	field := subject.FieldByName(propertyPath[0])
```
matthchr, line 312: Do we need to protect against the possibility of an empty property path? Otherwise a panic here.

v2/internal/reflecthelpers/reflect_helpers.go
```diff
@@ -282,3 +282,74 @@ func getItemsField(listPtr client.ObjectList) (reflect.Value, error) {
...
+	// If this is not the last property in the path, we need to recurse
+	if len(propertyPath) > 1 {
+		if field.Kind() == reflect.Ptr {
+			// Field is a pointer; initialize it if needed, then pass the pointer recursively
+			if field.IsNil() {
+				newValue := reflect.New(field.Type().Elem())
+				field.Set(newValue)
+			}
+
+			return errors.Wrapf(
```
matthchr, line 328: minor: Does `Wrapf` return nil if there was no error?

This reads pretty wonkily to me even if it is probably correct, it might be clearer as:

```
err := setPropertyCore(field.Interface(), propertyPath[1:], value)
if err != nil {
    return // error wrapping here
}

return nil
```

I don't feel super strongly about this but I did double-take here.

v2/internal/reflecthelpers/reflect_helpers.go
```diff
@@ -282,3 +282,74 @@ func getItemsField(listPtr client.ObjectList) (reflect.Value, error) {
...
+			propertyPath[0])
+	}
+
+	// If this is the last property in the path, we need to set the value, if we can
+	if !field.CanSet() {
+		return errors.Errorf("field %s was not settable", propertyPath[0])
+	}
+
+	// Catch any panic and turn it into an error return
+	defer func() {
```
matthchr, line 347: Don't you need to put this at the front of the function?
https://go.dev/play/p/nF13opHFOfO

Otherwise I don't think it'll be reached if anything above it panics?

v2/internal/reflecthelpers/reflect_helpers.go
matthchr: I guess it's not actually really possible in the calling paths we have and this is internal. I think if you move your panic protection up to the top of the function that's enough protection, no need for anything explicit probably.

v2/internal/reflecthelpers/reflect_helpers_test.go
```diff
@@ -408,3 +408,157 @@ func Test_SetObjectListItems(t *testing.T) {
...
+
+	subject := &ResourceWithReferences{
+		Spec: ResourceWithReferencesSpec{
+			AzureName: "azureName",
+		},
+	}
+
+	err := reflecthelpers.SetProperty(subject, "Spec.AzureName", 40)
+	g.Expect(err).To(HaveOccurred())
+	g.Expect(err.Error()).To(ContainSubstring("Spec"))
```
matthchr, line 559: minor, style: There's a `g.Expect(err).To(Match(ContainsSubstring())` which I think is what gomega wants you to use, as it then knows its an error context and gives a slightly better error message.

v2/internal/reflecthelpers/reflect_helpers.go
```diff
@@ -282,3 +282,85 @@ func getItemsField(listPtr client.ObjectList) (reflect.Value, error) {
...
+		return errors.Errorf("property path was empty")
+	}
+
+	steps := strings.Split(propertyPath, ".")
+	return setPropertyCore(obj, steps, value)
+}
+
+func setPropertyCore(obj any, propertyPath []string, value any) (err error) {
+	// Get the underlying object we need to modify
+	subject := reflect.ValueOf(obj)
```
super-harsh, line 303: do we need nil protection for `obj` here?

theunrepentantgeek: Not here, it's safe when recursing. But we do need a check on the entry point, I'll add that.

matthchr (approved):

pkg/resourcemanager/mysql/mysqlhelper.go
```diff
@@ -9,9 +9,9 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"github.com/hashicorp/go-multierror"
```
matthchr, line 12: This seems like a weird file to have included here? And I think the change you have here is actually not the recommended ordering anyway (it seemed correct before?)

v2/internal/reflecthelpers/reflect_helpers.go
theunrepentantgeek: As discussed, I've moved the recover to the top of the method.

PR #3355, opened 2023-10-01
Support multi architecture docker images using BuildX

### LINKED ISSUE #3321: Publish multi-arch image manifest for controller-manager including arm64
ASO's official images are currently only built for linux/amd64:
```
% docker manifest inspect mcr.microsoft.com/k8s/azureserviceoperator:v2.3.0
{
        "schemaVersion": 2,
        "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
        "config": {
                "mediaType": "application/vnd.docker.container.image.v1+json",
                "size": 3100,
                "digest": "sha256:c21d4b8ad4a63ed679622b1c2d7adaabb5df495db340062c1e715a256d0e6481"
        },
        "layers": [
                {
                        "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
                        "size": 83932,
                        "digest": "sha256:dd5ad9c9c29f04b41a0155c720cf5ccab28ef6d353f1fe17a06c579c70054f0a"
                },
                ...
        ]
}
```

Compared to:
```
% docker manifest inspect registry.k8s.io/cluster-api-azure/cluster-api-azure-controller:v1.11.0
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 2403,
         "digest": "sha256:c270d77a0cd5b9b3e59037b5915f3b68ef9658df7983908516f8182a8dc4c81c",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 2403,
         "digest": "sha256:3ef1aedb72804e5391c0cc7ceb5c36c2b56cbe8cef060c377b55d7fbaea038a9",
         "platform": {
            "architecture": "arm64",
            "os": "linux",
            "variant": "v8"
         }
      },
      ...
   ]
}
```

**Describe the improvement**
ASO should also build a linux/arm64 image and publish a multi-arch manifest referenced by the Helm chart and YAML manifests.

**Additional context**

super-harsh:
Closes #3321 

This PR adds support for publishing multi-architecture docker images using buildx.
Changes include: 
- Adding task for docker login
- Adding task for build, tag, and push docker image using buildx
- Moving controller build inside docker container to support multi-architecture build
- Updating devcontainer to install buildx


.devcontainer/install-dependencies.sh
```diff
@@ -210,6 +210,15 @@ if should-install "$TOOL_DEST/cmctl"; then
     curl -L "https://github.com/jetstack/cert-manager/releases/latest/download/cmctl-${os}-${arch}.tar.gz" | tar -xz -C "$TOOL_DEST"
 fi
 
+BUILDX_DEST=$HOME/.docker/cli-plugins
+write-verbose "Checking for $BUILDX_DEST/docker-buildx"
+if should-install "$BUILDX_DEST/docker-buildx"; then
+    write-info "Installing buildx-${os}_${arch}…"
```
theunrepentantgeek, line 216: The variables `os` and `arch` aren't defined when I check using **bash** under Ubuntu - so unless there's some magic going on, it looks like these will only work when building the devcontainer - but this script is supposed to work for local setup as well.

.github/workflows/create-release.yml
```diff
@@ -49,13 +42,18 @@ jobs:
...
-        # note that all creds are on host and never passed into devcontainer
-        uses: docker/login-action@v2.1.0
-        with:
-          registry: ${{ secrets.REGISTRY_LOGIN }}
-          username: ${{ secrets.AZURE_CLIENT_ID }}
-          password: ${{ secrets.AZURE_CLIENT_SECRET }}
+      - name: Docker login
+        run: |
+          container_id=${{env.container_id}}
+          docker exec -e AZURE_CLIENT_ID -e AZURE_CLIENT_SECRET -e DOCKER_REGISTRY "$container_id" task docker-login
```
theunrepentantgeek, line 48: Now that these secrets are being passed in to the devcontainer, is there any possibility of them being logged or left behind as a part of the published image? If so, this opens up the chance of someone harvesting those secrets.

theunrepentantgeek (commented):
A couple of comments, but I'm not yet convinced that this enables the container change we want without cutting off entire development scenarios.

.devcontainer/install-dependencies.sh
super-harsh: Aren't these vars (os and arch) are set above on line 205?

.github/workflows/create-release.yml
super-harsh: I don't think so - We already do pass-in secrets for live validation and az-login in other workflows as well.

Taskfile.yml
```diff
@@ -377,21 +377,20 @@ tasks:
       - "pkg/**/*.go"
       - "cmd/controller/**/*.go"
     cmds:
-      - CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build {{.VERSION_FLAGS}} -o ./bin/{{.CONTROLLER_APP}} ./cmd/controller/
+      - CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags {{.VERSION_FLAGS}} -o ./bin/{{.CONTROLLER_APP}} ./cmd/controller/
 
   controller:docker-build:
```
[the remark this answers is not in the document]
matthchr: Oh wait, buildx actually isn't using `CONTROLLER_DOCKER_IMAGE`? Its checking it in status but never actually producing it as far as I can tell?

Taskfile.yml
```diff
@@ -424,6 +423,23 @@ tasks:
...
+    cmds:
+      - 'if [ -z "{{.DOCKER_PUSH_TARGET}}" ]; then echo "Error: DOCKER_PUSH_TARGET must be set"; exit 1; fi'
+      - docker buildx create --use
+      - docker buildx build --push 
+        --build-arg VERSION_FLAGS={{.VERSION_FLAGS}}ga Ωx
+        --build-arg CONTROLLER_APP={{.CONTROLLER_APP}}
+        --platform linux/amd64,linux/arm64 
+        --tag "{{.DOCKER_PUSH_TARGET}}/{{.CONTROLLER_DOCKER_IMAGE}}" .
+    status:
+      - "docker manifest inspect {{.LOCAL_REGISTRY_CONTROLLER_DOCKER_IMAGE}} > /dev/null"
```
matthchr, line 441: This `status` check doesn't seem to mesh with the commands being run. What is producing the `LOCAL_REGISTRY_CONTROLLER_DOCKER_IMAGE`? If nothing, this status check is wrong?

It should do a manifest inspect on an actual produced image maybe?

matthchr: or maybe we can delete this since we have `run: always`? Might need to read the taskfile docs on that one

Taskfile.yml
```diff
@@ -1030,6 +1046,13 @@ tasks:
       - az login --service-principal -u {{.AZURE_CLIENT_ID}} -p {{.AZURE_CLIENT_SECRET}} --tenant {{.AZURE_TENANT_ID}} > /dev/null
       - az account set --subscription {{.AZURE_SUBSCRIPTION_ID}}
 
+  docker-login:
+    desc: Docker login
+    deps:
```
matthchr, line 1051: If no deps I think can delete this section

v2/Dockerfile
```diff
@@ -1,10 +1,29 @@
-# Note: This Dockerfile assumes that the binary has been built using the top-level Taskfile.yml
```
matthchr, line 1: Note that this Dockerfile still assumes that the `bundle-crds` taskfile target has been run, for line 27 to do the right thing. Might be worth keeping a comment either here at the top or on line 27 for the CRD copy?

v2/.dockerignore
```diff
@@ -0,0 +1,11 @@
+**
+
+# Ignore everything except below
+!/go.mod
+!/go.sum
+!/*.go
+!/internal
+!/pkg
```
matthchr, line 8: aren't internal and pkg going to be included because of the `!/*.go` above? Do we really need the called out explicitly?

matthchr: Why does the container need `api` folder?

.github/workflows/create-release.yml
```diff
@@ -49,13 +42,18 @@ jobs:
           file: "v2/out/release/*"
           file_glob: true
 
-      - name: Login to registry
```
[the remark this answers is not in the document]
super-harsh: Yep you're right below on - docker context on host and docker context in devcontainer are different. I tried this out in my test workflow and I was getting a 401 error

v2/.dockerignore
super-harsh: > Why does the container need `api` folder?

Initially I tried without `api` folder and build fails with error below: 
```
[controller:docker-push-multiarch]  > [linux/arm64 builder 7/7] RUN CGO_ENABLED=0 go build -ldflags "-X github.com/Azure/azure-service-operator/v2/internal/version.BuildVersion=v2.3.0-rev.71-dirty" -o ./bin/aso-controller ./cmd/controller/:
[controller:docker-push-multiarch] 1.197 internal/controllers/controller_resources_gen.go:213:2: no required module provides package github.com/Azure/azure-service-operator/v2/api/web/customizations; to add it:
[controller:docker-push-multiarch] 1.197 	go get github.com/Azure/azure-service-operator/v2/api/web/customizations
[controller:docker-push-multiarch] 1.197 internal/controllers/controller_resources_gen.go:214:2: no required module provides package github.com/Azure/azure-service-operator/v2/api/web/v1api20220301; to add it:
[controller:docker-push-multiarch] 1.197 	go get github.com/Azure/azure-service-operator/v2/api/web/v1api20220301
[controller:docker-push-multiarch] 1.197 internal/controllers/controller_resources_gen.go:215:2: no required module provides package github.com/Azure/azure-service-operator/v2/api/web/v1api20220301storage; to add it:
[controller:docker-push-multiarch] 1.197 	go get github.com/Azure/azure-service-operator/v2/api/web/v1api20220301storage
[controller:docker-push-multiarch] 1.197 internal/controllers/controller_resources_gen.go:216:2: no required module provides package github.com/Azure/azure-service-operator/v2/api/web/v1beta20220301; to add it:
[controller:docker-push-multiarch] 1.197 	go get github.com/Azure/azure-service-operator/v2/api/web/v1beta20220301
[controller:docker-push-multiarch] 1.197 internal/controllers/controller_resources_gen.go:217:2: no required module provides package github.com/Azure/azure-service-operator/v2/api/web/v1beta20220301storage; to add it:
[controller:docker-push-multiarch] 1.197 	go get github.com/Azure/azure-service-operator/v2/api/web/v1beta20220301storage
```

super-harsh: > aren't internal and pkg going to be included because of the `!/*.go` above? Do we really need the called out explicitly?

Actually, `!/*.go` doesn't make sense, I'll remove it.

.github/workflows/create-release.yml
```diff
@@ -49,13 +42,18 @@ jobs:
...
+          DOCKER_REGISTRY: ${{ secrets.REGISTRY_LOGIN }}
+          AZURE_CLIENT_ID: ${{ secrets.AZURE_CLIENT_ID }}
+          AZURE_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
 
-      - name: Push docker image
-        run: docker push --all-tags ${{ secrets.REGISTRY_PUBLIC }}/azureserviceoperator
+      - name: Build, tag and push docker image
+        run: |
+          container_id=${{env.container_id}}
+          docker exec -e DOCKER_PUSH_TARGET "$container_id" task controller:docker-push-multiarch
```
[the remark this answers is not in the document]
super-harsh: We don't have any other docker push target instead of `docker-push-local`. I'm letting the local push target be there for reasons: 
- For local builds, we don't need multi-arch images as they take time to build and occupy space.
- Buildx does not have direct support for pushing to local registries. We need to add in a few extra steps to enable it. Mentioned in [Build cannot export to registry on localhost](https://github.com/docker/buildx/issues/94#top)

.devcontainer/install-dependencies.sh
```diff
@@ -210,6 +210,15 @@ if should-install "$TOOL_DEST/cmctl"; then
     curl -L "https://github.com/jetstack/cert-manager/releases/latest/download/cmctl-${os}-${arch}.tar.gz" | tar -xz -C "$TOOL_DEST"
 fi
 
+BUILDX_DEST=$HOME/.docker/cli-plugins
+write-verbose "Checking for $BUILDX_DEST/docker-buildx"
+if should-install "$BUILDX_DEST/docker-buildx"; then
```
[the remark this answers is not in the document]
theunrepentantgeek: We have the exact same issue with other installations, which is why this script has the `--skip-installed` flag so that installation can be forced.

(It's that way around so someone running the script with no parameters gets everything installed; `dev.sh` uses the flag to make setup quick if things are present.)

.github/workflows/create-release.yml
theunrepentantgeek: Checking a `git blame` of this file, it was @Porges who wrote the comment

> `# note that all creds are on host and never passed into devcontainer`

Working from the precept that smart people do things for good reasons, I'm worried that we're missing something here. To be more specific on my earlier concern, I'm not worried about disclosure during CI, I'm worried about the credentials being left lying around _inside the container image_, available for anyone to extract if they go nosing around inside.

Taskfile.yml
```diff
@@ -35,7 +35,7 @@ vars:
   LATEST_VERSION_TAG:
     sh: git describe --tags $(git rev-list --tags=v2* --max-count=1)
 
-  VERSION_FLAGS: -ldflags "-X {{.PACKAGE}}/internal/version.BuildVersion={{.VERSION}}"
+  VERSION_FLAGS: '"-X {{.PACKAGE}}/internal/version.BuildVersion={{.VERSION}}"'
```
theunrepentantgeek, line 38: Why remove `-ldflags` - it's one of the flags required to set the version in the executable. Doesn't make sense to me to remove this here, then restate is manually everywhere else.

If you have a need for part of this value elsewhere, introduce a different variable for that purpose rather than redefining this one and making things awkward.

Taskfile.yml
```diff
@@ -424,6 +423,23 @@ tasks:
...
+    desc: Builds, tags and pushes the multi architecture controller container image to repository
+    deps:
+      - controller:bundle-crds
+    dir: "{{.CONTROLLER_ROOT}}"
+    run: always
+    cmds:
+      - 'if [ -z "{{.DOCKER_PUSH_TARGET}}" ]; then echo "Error: DOCKER_PUSH_TARGET must be set"; exit 1; fi'
+      - docker buildx create --use
+      - docker buildx build --push 
+        --build-arg VERSION_FLAGS={{.VERSION_FLAGS}}ga Ωx
```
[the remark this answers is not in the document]
theunrepentantgeek: Good question! #confused

Taskfile.yml
theunrepentantgeek: As I understand `go-task`, if you have `run: always` then it will always be run, regardless of the status checks. Either remove `run:` or `status:`, no need to have both.

theunrepentantgeek (approved):
I'm tentatively convinced that this doesn't prohibit development and debugging approaches that currently work, so I'm approving. Once the comments are addressed, merge away.

Taskfile.yml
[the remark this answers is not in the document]
super-harsh: That shouldn't be there. Thanks for spotting.

v2/.dockerignore
matthchr: doh, I mistook `api` for `config`. `api` contains the actual API Go files, which obviously do need to be included.

> Actually, !/*.go doesn't make sense, I'll remove it.

Can you expand on this? Based on my read of dockerignore syntax, `!/*.go` should include any go file (weirdly * seems multi-directory-recursive in .dockerignore even though it's not usually).

.github/workflows/create-release.yml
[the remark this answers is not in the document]
matthchr: We should probably document this on the taskfile targets, for why we have two and why different ones are used in different places.

There are some risks with this too (we're not actually testing the image we're creating), but I see that _lots_ of people have this problem, as documented:
https://github.com/docker/buildx/issues/166
https://github.com/docker/buildx/issues/1152

There _does_ seem to be a way to push buildx image to local registry now though, so I am wondering if we should try swapping the `kind` steps to use the multiarch image?

Taskfile.yml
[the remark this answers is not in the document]
super-harsh: yep its used here at [line 439](https://github.com/Azure/azure-service-operator/pull/3355/files/8030ada8af97fcb8b0cdd3a7333984eeb7e81d17#diff-cd2d359855d0301ce190f1ec3b4c572ea690c83747f6df61c9340720e3d2425eR439). We do want to use it for tagging the image right?

.github/workflows/create-release.yml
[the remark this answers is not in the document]
super-harsh: We can surely swap kind steps to use buildx, but not sure about the multiarch images if thats a good idea. As I noticed, for cross-platform images buildx takes around about 10-15 mins to build. Which would increase the CI run time.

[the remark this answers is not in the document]
theunrepentantgeek: CI shouldn't be building the image every time - it's supposed to be cached. We even have [a workflow](https://github.com/Azure/azure-service-operator/blob/main/.github/workflows/build-devcontainer-image.yml )that runs weekly to keep it current.

So if switching CI to use buildx slows it down, we need to fix that.

v2/.dockerignore
super-harsh: > !/*.go should include any go file

Is that documented? I thought its just gonna include `.go` files in the same directory as `Dockerfile` is.

.github/workflows/create-release.yml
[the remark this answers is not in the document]
matthchr: I don't think CI caches this image @theunrepentantgeek - it caches the .devcontainer image, not the controller image (being built here). It must build the controller image every time because that's the image that:
- Contains the code changes for the PR in question.
- Runs in kind and is verified by the tests.

My understanding of what `buildx` does is that it basically calls N docker builds with different architectures and then automatically merges the manifests. So it may be OK to not use it for the local build (see the issues I linked above for where this is talked about some and people suggest using buildx w/ a single version argument to get local working, which anyway is going to be different than what we do for release)

Taskfile.yml
```diff
@@ -377,52 +378,79 @@ tasks:
...
     desc: Builds the {{.CONTROLLER_APP}} Docker image.
     dir: "{{.CONTROLLER_ROOT}}"
     deps:
-      - controller:build
       - controller:bundle-crds
     run: always
     sources:
+      - "go.mod"
+      - "go.sum"
+      - "*.go"
```
matthchr, line 392: I think you can replace a lot of these w/ a `**/*.go` pattern?

Taskfile.yml
```diff
@@ -377,52 +378,79 @@ tasks:
...
     desc: Pushes the controller container image to a local registry
     deps: 
-      - controller:docker-build
+      - controller:bundle-crds
     dir: "{{.CONTROLLER_ROOT}}"
     run: always
+    sources:
+      - "go.mod"
+      - "go.sum"
+      - "*.go"
```
matthchr, line 418: same **/*.go comment, I think can reduce this

Taskfile.yml
```diff
@@ -377,52 +378,79 @@ tasks:
...
+      - Dockerfile
     cmds:
-      - docker tag {{.CONTROLLER_DOCKER_IMAGE}} {{.LOCAL_REGISTRY_CONTROLLER_DOCKER_IMAGE}}
-      - docker push {{.LOCAL_REGISTRY_CONTROLLER_DOCKER_IMAGE}}
-    status:
-      - "docker manifest inspect {{.LOCAL_REGISTRY_CONTROLLER_DOCKER_IMAGE}} > /dev/null"
+      - docker buildx create --driver-opt network=host --use
+      - docker buildx build --push
+        --build-arg VERSION_FLAGS={{.VERSION_FLAGS}}
+        --build-arg CONTROLLER_APP={{.CONTROLLER_APP}}
```
matthchr, line 428: minor: Leave a comment here about how we don't use multiarch platform options here because we dont need them for local testing and they don't work for local registries? (Can link the GH issues I linked earlier)

v2/.dockerignore
matthchr: Ah, you're right. It's docuemnted here: https://docs.docker.com/engine/reference/builder/#dockerignore-file

They support `**` though:
> Beyond Go's filepath.Match rules, Docker also supports a special wildcard string ** that matches any number of directories (including zero). For example, **/*.go will exclude all files that end with .go that are found in all directories, including the root of the build context.

So `!**/*.go` should cover all of the Go files without needing to remember to come update this folder list if we add new directories.

v2/Dockerfile
```diff
@@ -1,10 +1,30 @@
...
+# Note: This Dockerfile still assumes that the "bundle-crds" taskfile target has been run already
```
matthchr, line 1:
```suggestion
# Note: This Dockerfile assumes that the "bundle-crds" taskfile target has been run already
```

.github/workflows/create-release.yml
super-harsh: > By default, Docker looks for the native binary on each of the platforms, i.e. "osxkeychain" on macOS, "wincred" on windows, and "pass" on Linux. A special case is that on Linux, Docker will fall back to the "secretservice" binary if it cannot find the "pass" binary. If none of these binaries are present, it stores the credentials (i.e. password) in base64 encoding in the config files described above.

Here's the [reference](https://docs.docker.com/engine/reference/commandline/login/#default-behavior) to how docker stores the credentials and auth tokens.

Taskfile.yml
```diff
@@ -372,57 +373,72 @@ tasks:
       # excluding the ./apis directory here
```
matthchr, line 373: I'm not sure this is true anymore, it's not excluded. I don't know why it should be excluded though because I do think that the controller does depend on the api folder... so maybe just delete this comment?

matthchr (approved):

PR #3369, opened 2023-10-02
Log disk stuff

matthchr:
Closes #[issue number]

gif

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -55,6 +55,9 @@ jobs:
           fetch-depth: 0 # required to access tags
           submodules: 'true'
 
+      - name: Force docker to SSD
+        run: sudo scripts/v2/linux-docker-use-ssd.sh --containerd true
+
       - name: check-changes
         id: check-changes
         run: scripts/v2/check-changes.sh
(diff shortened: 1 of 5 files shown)


.github/workflows/pr-validation.yml
```diff
@@ -55,6 +55,9 @@ jobs:
           fetch-depth: 0 # required to access tags
           submodules: 'true'
 
+      - name: Force docker to SSD
```
super-harsh, line 58:
```suggestion
      - name: Force docker to use SSD
```

.github/workflows/pr-validation.yml
```diff
@@ -121,6 +124,9 @@ jobs:
         
       - name: Build docker image & build configuration YAML
         run: |
+          df -H
+          docker info
+          docker image list
```
super-harsh, line 129: Do we still need these logs here to be merged?

super-harsh (approved):

.github/workflows/pr-validation.yml
theunrepentantgeek: For now, it is probably good to keep these. Maybe we remove them later.

theunrepentantgeek (approved):

.github/workflows/pr-validation.yml
super-harsh: In that case, I'd suggest leaving a TODO here so we know this is just test stuff and can be removed later.

matthchr: I've removed them now

PR #3357, opened 2023-10-02
Goal Seeking Keyvaults

theunrepentantgeek:
**What this PR does / why we need it**:

Implements goal seeking for KeyVaults as described in our [ADR](https://github.com/Azure/azure-service-operator/blob/main/docs/hugo/content/design/ADR-2023-07-Goal-Seeking-KeyVaults/_index.md).

We add two additional options to `createMode`:

* `createOrRecover` - creates a new KeyVault, unless there's a soft-deleted one to recover
* `purgeThenCreate` - any existing soft-deleted KeyVault is purged, then a new one created.

Closes #1415 

**Special notes for your reviewer**:

The option `purgeThenCreate` is dangerous because it results in ASO turning a soft-delete into a hard delete without warning and without the possibility of recovery. Our documentation will call this out.

**Prerequisites**

* [x] #3348 

**How does this PR make you feel**:
gif

**If applicable**:
- [x] this PR contains tests

PR #3370, opened 2023-10-03
Use createOrRecover for sample keyvaults

theunrepentantgeek:

Updates our samples with KeyVaults to use `createMode` set to `createOrRecover` so they are somewhat more resilient.

**Prerequisites**

- [x] #3357 

Recovering a KeyVault only works if the soft-deleted KeyVault is in the same subscription as where you're trying to create a new one, so this isn't a universal fix.

gif

**If applicable**:
- [x] this PR contains tests


v2/samples/keyvault/v1api/v1api20210401preview_vault.yaml
```diff
@@ -43,4 +44,3 @@ spec:
             - "get"
           storage:
             - "get"
-
```
super-harsh, line 46: Minor: New line

super-harsh (approved):

v2/samples/keyvault/v1api/v1api20210401preview_vault.yaml
theunrepentantgeek: The diff is interesting. I've _removed_ a _missing EOLN_ by adding one. 😉

PR #3375, opened 2023-10-04
Change CRD doc automation to build index files as well

theunrepentantgeek:

Ensures that we build index files as well as the CRD docs so that everything gets updated together.

Looks like someone (probably me) had an earlier go at fixing this, but the extra build step was having no effect.

gif


super-harsh (approved):

.github/workflows/api-docs-repo.yaml
```diff
@@ -71,10 +71,11 @@ jobs:
           docker exec "$container_id" task controller:gen-crd-docs
 
       # After generating the API docs (above), regenerate our index files as now new CRDs can be linked to their docs
+      # Have to force this to ensure the index files are updated
       - name: Generate Index files
         run: |
           container_id=${{ env.container_id }}
-          docker exec "$container_id" task controller:generate-types
+          docker exec "$container_id" task controller:generate-types --force
```
matthchr, line 78: The fact we have to force this indicates that there is a problem in the `generate-types` task definition where it is not reading files that it ought to be.

Do we need to update its:
```
    sources:
      - ./bin/{{.GENERATOR_APP}}
      - azure-arm.yaml
```

section?

Sorry for the slow review on this.

PR #3380, opened 2023-10-05
Update Go-Get, take II

theunrepentantgeek:

Often the `go get` step in our CI builds will fail.

This is due to delays in the default proxy used by the `go` CLI, [specifically](https://proxy.golang.org/):

> In order to improve our services' caching and serving latencies, new versions may not show up right away. If you want new code to be immediately available in the mirror, then first make sure there is a semantically versioned tag for this revision in the underlying source repository. Then explicitly request that version via go get module@version. The new version should be available within one minute. 

Bypassing the proxy would invalidate the scenario we want to test, so I've juiced up the retries.

gif

### DIFF
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -41,15 +41,19 @@ jobs:
             sha='${{ github.sha }}'
           fi
 
-          # Sometimes go get will fail initially, so we retry it up to 5 times before giving up
+          # Give GitHub some time to make the commit available
+          sleep 10
+
+          # Sometimes it can take https://proxy.golang.org up to a minute for the latest commit to be available.
+          # See the FAQ at https://proxy.golang.org/
           FAILED=1
           for i in {1..5}; do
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-validation.yml
```diff
@@ -41,15 +41,19 @@ jobs:
...
           FAILED=1
           for i in {1..5}; do
             if go get "$repo/v2@$sha" ; then
               FAILED=0
               break
             fi
-            echo "Failed to 'go get' $repo/v2@$sha, retrying in 10 seconds…"
-            sleep 10
+            echo "Failed to 'go get' $repo/v2@$sha, retrying in 60 seconds…"
+            sleep 60
```
matthchr, line 56: We should monitor how long this takes, because if it actually is taking anywhere near 5m we should probably just strip this bit out into its own async check rather than putting it here. Adding +5m to the CI run would ideally be avoided.

matthchr (approved):

PR #3405, opened 2023-10-11
Add documentation about TypeTransformers

theunrepentantgeek:

Adds documentation to `azure-arm.yaml` to explain how TypeTransformers work.

Including the documentation in this file ensures it is right to hand when needed.

gif

**If applicable**:


v2/azure-arm.yaml
```diff
@@ -130,8 +130,45 @@ typeFilters:
...
+# Selection Properties
+#
+# group    - select types or properties in the specified group or groups
+# version  - select types or properties in the specified version or version(s)
+# name     - select types or properties with the specified name or name(s)
+# property - select properties with the given name or name(s)
+#
+# All of these properties are matchers:
+#
+# o They wllow multiple values, semicolon (;) separated
```
super-harsh, line 153:
```suggestion
# o They allow multiple values, semicolon (;) separated
```

super-harsh: At first, I thought that `o` was a character there by mistake and not a bullet point

super-harsh (approved):

v2/azure-arm.yaml
theunrepentantgeek: Normally when I write bullet points like this, I indent like so:

``` go
#
# All of these properties are matchers:
#
#   o   They wllow multiple values, semicolon (;) separated
```

which makes it more obvious. However, our ASO linting objects, so this is my compromise.

PR #3443, opened 2023-10-18
When loading CRDs, handle multi-documents

vincepri:
Closes #[issue number]

This is a small improvements to the current CRD loading logic from disk. If there are multiple documents per file, the loading logic doesn't take that into account today.

gif

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/v2/internal/crdmanagement/manager.go
+++ b/v2/internal/crdmanagement/manager.go
@@ -4,8 +4,11 @@
 package crdmanagement
 
 import (
+	"bufio"
+	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"reflect"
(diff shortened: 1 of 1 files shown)


matthchr:
> This is a small improvements to the current CRD loading logic from disk. If there are multiple documents per file, the loading logic doesn't take that into account today.

I'm not sure I understand how there could be multiple documents per file? 

- The container image is produced by our build process which writes a single CRD to a single file.
- The pod is launched with [a readonly filesystem](https://github.com/Azure/azure-service-operator/blob/main/v2/config/manager/manager.yaml#L73)

Somebody could base an image off of our image and update the files and run that instead, and there's probably other tricks/things they could do as well, but I'm not sure any of those things are really officially supported.

Can you describe the scenario you're enabling here?

PR #3445, opened 2023-10-19
Add a flag to control CRD management

vincepri:
Closes #[issue number]

Currently, there is no way to decide if the ASO manager should control all CRDs, or just use the ones already installed in the cluster without any modifications.

This PR adds a flag called `enable-crd-management`, enabled by default (to retain backward compatibility).

More information was captured in the upstream slack thread: https://kubernetes.slack.com/archives/C046DEVLAQM/p1697665840941649.

gif

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/v2/cmd/controller/app/flags.go
+++ b/v2/cmd/controller/app/flags.go
@@ -20,18 +20,20 @@ type Flags struct {
 	WebhookPort          int
 	WebhookCertDir       string
 	EnableLeaderElection bool
+	CRDManagementMode    string
 	CRDPatterns          string // This is a ; delimited string containing a collection of patterns
 	PreUpgradeCheck      bool
 }
 
 func (f Flags) String() string {
 	return fmt.Sprintf(
-		"MetricsAddr: %s, HealthAddr: %s, WebhookServerPort: %d, WebhookServerCertDir: %s, EnableLeaderElection: %t, CRDPatterns: %s, PreUpgradeCheck: %t",
+		"MetricsAddr: %s, HealthAddr: %s, WebhookPort: %d, WebhookCertDir: %s, EnableLeaderElection: %t, CRDManagementMode: %s, CRDPatterns: %s, PreUpgradeCheck: %t",
 		f.MetricsAddr,
 		f.HealthAddr,
 		f.WebhookPort,
 		f.WebhookCertDir,
 		f.EnableLeaderElection,
+		f.CRDManagementMode,
 		f.CRDPatterns,
 		f.PreUpgradeCheck)
 }
@@ -46,6 +48,7 @@ func ParseFlags(args []string) (Flags, error) {
 	var webhookPort int
 	var webhookCertDir string
 	var enableLeaderElection bool
+	var crdManagementMode string
 	var crdPatterns string
 	var preUpgradeCheck bool
 
@@ -56,6 +59,8 @@ func ParseFlags(args []string) (Flags, error) {
 	flagSet.StringVar(&webhookCertDir, "webhook-cert-dir", "", "The directory the webhook server's certs are stored.")
 	flagSet.BoolVar(&enableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controllers manager. Enabling this will ensure there is only one active controllers manager.")
+	flagSet.StringVar(&crdManagementMode, "crd-management", "auto",
+		"Instructs the operator on how it should manage the Custom Resource Definitions. One of 'auto', 'none'")
 	flagSet.StringVar(&crdPatterns, "crd-pattern", "", "Install these CRDs. CRDs already in the cluster will also always be upgraded.")
 	flagSet.BoolVar(&preUpgradeCheck, "pre-upgrade-check", false,
 		"Enable pre upgrade check to check if existing crds contain helm 'keep' policy.")
@@ -68,6 +73,7 @@ func ParseFlags(args []string) (Flags, error) {
 		WebhookPort:          webhookPort,
 		WebhookCertDir:       webhookCertDir,
 		EnableLeaderElection: enableLeaderElection,
+		CRDManagementMode:    crdManagementMode,
 		CRDPatterns:          crdPatterns,
 		PreUpgradeCheck:      preUpgradeCheck,
 	}, nil
--- a/v2/cmd/controller/app/setup.go
+++ b/v2/cmd/controller/app/setup.go
@@ -13,9 +13,6 @@ import (
 	"regexp"
 	"time"
 
-	"sigs.k8s.io/controller-runtime/pkg/metrics/server"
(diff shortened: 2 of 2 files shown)


v2/cmd/controller/app/setup.go
```diff
@@ -156,39 +155,44 @@ func SetupControllerManager(ctx context.Context, setupLog logr.Logger, flgs Flag
...
 	}
 
-	// We only apply CRDs if we're in webhooks mode. No other mode will have CRD CRUD permissions
-	if cfg.OperatorMode.IncludesWebhooks() {
-		var installationInstructions []*crdmanagement.CRDInstallationInstruction
-		installationInstructions, err = crdManager.DetermineCRDsToInstallOrUpgrade(goalCRDs, existingCRDs, flgs.CRDPatterns)
+	// By default, assume the existing CRDs are the goal CRDs. If CRD management is enabled, we will
+	// load the goal CRDs from disk and apply them.
+	goalCRDs := existingCRDs
+	if flgs.EnableCRDManagement {
```
theunrepentantgeek, line 167: This will lead to some very hard to diagnose failure modes.

The CRD versions in the cluster **must exactly** match the version of ASO that's being run. Upgrading either one in isolation will create an environment where some, if not all resources, will not work.

The underlying issue is to do with which version of each resource is tagged as the storage version - if an ASO release introduces a new version of a resource, the storage version changes (e.g. from `v1api20200401storage` to `v120230801storage`); a mismatch of versions between the CRDs and the running ASO operator is a fatal error.

I think we need to retain the check, but provide an option on what to do if a CRD mismatch is discovered. I can think of three options:

* auto-upgrade - the existing behaviour
* abort - the operator refuses to run, with suitable logging to inform users
* degrade - the operator runs, but only supporting correctly versioned resources

theunrepentantgeek: We considered using a central "hub" storage version during design, but that had [some](https://azure.github.io/azure-service-operator/design/versioning/#alternative-fixed-storage-version) [problems](https://azure.github.io/azure-service-operator/design/versioning/case-studies/fixed-storage-version/).

The approach we've taken is [chained storage versions](https://azure.github.io/azure-service-operator/design/versioning/case-studies/chained-storage-versions/); that case study goes into a great deal of information.

theunrepentantgeek: After discussions on Slack, I better understand the usefulness of this flag. 

Perhaps we rename the flag to better indicate that the user is _disabling_ standard functionality, rather than _enabling_ optional functionality?

For example, instead of `--enable-crd-management` with values `true` (the default) and `false`, what if it was `--crd-compatibility-check` with values of `auto-upgrade` (the default), `verify` and `none` ?

v2/cmd/controller/app/flags.go
```diff
@@ -46,6 +48,7 @@ func ParseFlags(args []string) (Flags, error) {
 	var webhookPort int
 	var webhookCertDir string
 	var enableLeaderElection bool
+	var enableCRDmanagement bool
```
matthchr, line 51:
```suggestion
	var enableCRDManagement bool
```

v2/cmd/controller/app/flags.go
```diff
@@ -20,18 +20,20 @@ type Flags struct {
...
 	EnableLeaderElection bool
+	EnableCRDManagement  bool
 	CRDPatterns          string // This is a ; delimited string containing a collection of patterns
 	PreUpgradeCheck      bool
 }
 
 func (f Flags) String() string {
 	return fmt.Sprintf(
-		"MetricsAddr: %s, HealthAddr: %s, WebhookServerPort: %d, WebhookServerCertDir: %s, EnableLeaderElection: %t, CRDPatterns: %s, PreUpgradeCheck: %t",
+		"MetricsAddr: %s, HealthAddr: %s, Webhook=Port: %d, WebhookCertDir: %s, EnableLeaderElection: %t, EnableCRDManagement: %t, CRDPatterns: %s, PreUpgradeCheck: %t",
```
matthchr, line 30: Was this `=` intentional? Maybe just `WebhookPort`?

v2/cmd/controller/app/setup.go
matthchr: I agree with the reversed phrasing, (about disabling a feature versus enabling one), if we were leaving this a boolean. I'm not sold on the `crd-compatibility-check` naming, as IMO "upgrading the CRDs" isn't really a compatibility check and that's the default mode.

Maybe `crd-management-mode` `auto|none`, which the default being `auto`? I also agree that a `verify` might make sense but not sure we need to add it in this PR.

v2/cmd/controller/app/flags.go
vincepri: Leftover, fixed

v2/cmd/controller/app/setup.go
theunrepentantgeek: I like `crd-management-mode` with the options of `auto|none` - if we were to add `verify` in a future PR, that wouldn't be a breaking change.

theunrepentantgeek (approved):

v2/cmd/controller/app/flags.go
```diff
@@ -20,18 +20,20 @@ type Flags struct {
...
 	EnableLeaderElection bool
+	CRDManagementMode    string
 	CRDPatterns          string // This is a ; delimited string containing a collection of patterns
 	PreUpgradeCheck      bool
 }
 
 func (f Flags) String() string {
 	return fmt.Sprintf(
-		"MetricsAddr: %s, HealthAddr: %s, WebhookServerPort: %d, WebhookServerCertDir: %s, EnableLeaderElection: %t, CRDPatterns: %s, PreUpgradeCheck: %t",
+		"MetricsAddr: %s, HealthAddr: %s, WebhookPort: %d, WebhookCertDir: %s, EnableLeaderElection: %t, CRDManagementMode: %t, CRDPatterns: %s, PreUpgradeCheck: %t",
```
theunrepentantgeek, line 30: `CRDManagementMode` is now a string.

```suggestion
		"MetricsAddr: %s, HealthAddr: %s, WebhookPort: %d, WebhookCertDir: %s, EnableLeaderElection: %t, CRDManagementMode: %s, CRDPatterns: %s, PreUpgradeCheck: %t",
```

matthchr (approved):

matthchr:
@vincepri, looks like this is failing a golint check:
> [controller:lint] cmd/controller/app/setup.go:169:7: shadow: declaration of "err" shadows declaration at line 108 (govet)
> [controller:lint] 		var err error
> [controller:lint] 		    ^

Can you fix this and then we're happy to merge.

vincepri:
@matthchr done!

PR #3465, opened 2023-10-25
Upgrade TypeTransforms and rename AutoscaleSetting resource

### LINKED ISSUE #3403: Fix casing for Autoscalesetting to AutoscaleSetting
This should be done before the `2.4.0` release and is dependent on #3402.

theunrepentantgeek:

Renames the resource `Autoscalesetting` to `AutoscaleSetting` and the object `AutoscaleSetting` to `AutoscaleSettingProperties`.

In addition to the configuration changes required to do this, there are changes to type transformations to make it possible.

My first approach was to add configuration to `ObjectModelConfiguration` to support renames, but the case-insensitive nature of our configuration meant that both `Autoscalesettings` and `AutoscaleSettings` picked up the same configuration. Modifying the type transformations was necessary as they can select on _type_ as well as _name_.

Closes #3403 

The type `TransformTarget` was an example of _reuse by shape_, used both to select types for transformation, and to define the desired result afterwards. This dual nature made the code for it harder to understand, and contributed to the problems multiple maintainers have had understanding the code.

I've separated the two concerns, replacing `TransformTarget` with `TransformSelector` and `TransformResult`. These have the same serialization shape as `TransformTarget` (and thus exhibit backward compatibility) but are considerably simpler and easier to understand because they only have a single meaning.

gif

**If applicable**:
- [x] this PR contains tests


super-harsh (approved):
Looks clean. The only thing that needs fixing is `Kind` in `AutoscaleSetting` sample

PR #3466, opened 2023-10-25
Add configuration to rename types

### LINKED ISSUE #3402: Add ability to rename internal types

Currently, we don't have the ability to rename an internal type or a property which can be useful in cases where we want to rename a conflicting type or want to correct the name for a property

**Describe the improvement**
Should be able to rename any type within a resource

**Additional context**

Comment : [On AutoscaleSettings PR](https://github.com/Azure/azure-service-operator/pull/3340#discussion_r1338931722)

The AutoscaleSetting resource is named `Autoscalesetting` and we can not rename it as the resource already has `AutoscaleSetting` name reserved internally for properties object. Which makes generator run into an error below when I try to use `exportAs: AutoscaleSetting`. 

> error="error generating code: failed to execute stage 20: Apply export filters to reduce the number of generated types: a definition for github.com/Azure/azure-service-operator/v2/api/insights/v1api20221001/AutoscaleSetting already exists"

theunrepentantgeek:

We've had the ability to rename _resources_ for some time, using `$exportAs` but that doesn't support renaming other types. This PR adds configuration and functionality to do that.

Given that `$exportAs` implies `$export: true`, I didn't want to reuse it for non-resource types; the new configuration option is `$renameTo`.

Closes #3402 

This came out of my first attempt at closing #3403 but wasn't actually used for that.

gif

### DIFF
--- a/v2/tools/generator/internal/config/type_configuration.go
+++ b/v2/tools/generator/internal/config/type_configuration.go
@@ -39,6 +39,7 @@ type TypeConfiguration struct {
 	IsResource               configurable[bool]
 	ManualConfigs            configurable[[]string]
 	NameInNextVersion        configurable[string]
+	RenameTo                 configurable[string]
 	ResourceEmbeddedInParent configurable[string]
 	SupportedFrom            configurable[string]
 }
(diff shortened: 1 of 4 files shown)


v2/tools/generator/internal/config/type_configuration.go
```diff
@@ -272,6 +278,18 @@ func (tc *TypeConfiguration) UnmarshalYAML(value *yaml.Node) error {
 			continue
 		}
 
+		// $renameTo: <string>
```
matthchr, line 281: Can we document this in the `azure-arm.yaml`? Specifically interested in the "does it only work on resources" versus does it work on all types?

I think it works on all types, but we also have `exportAs`, which I think only works on resources? We should clarify in the documentation when users should use one versus the other and we should make sure that we give a reasonable error if they try to do it wrong.

I'm also interested in what the difference between `exportAs: foo` and `renameTo: foo` + `export: true` are.

theunrepentantgeek: Docs added.

`$export: true` + `$renameTo: <string>` are very close to `$exportAs` but with the renaming happening at a different pipeline stage, so the results may be subtly different.

matthchr (approved):

Issue #3433, opened 2023-10-16
Feature: Goal-seeking creation of Azure API Managament


ross-p-smith:
Azure APIM supports soft deletion by default. We need to support goal seeking behaviour so that the Operator can `createOrRestore` if required. The Resource Provider will currently error if you supply `Restore: true` if an apim has not previously been created. We would like the Operator to succeed in this scenario.

ross-p-smith:
Deleted APIM services are unrecoverable within 48 hours. If you try and recover an apim after this, you will get the following error.

```text
[controller:test-integration-envtest]     crd_apimanagement_20220801_test.go:60: 
[controller:test-integration-envtest]         No future change is possible.  Bailing out early after 10.009s.
[controller:test-integration-envtest]         Expected
[controller:test-integration-envtest]             <conditions.Condition>: {
[controller:test-integration-envtest]                 Type: "Ready",
[controller:test-integration-envtest]                 Status: "False",
[controller:test-integration-envtest]                 Severity: "Error",
[controller:test-integration-envtest]                 LastTransitionTime: {
[controller:test-integration-envtest]                     Time: 2023-10-25T22:19:54+01:00,
[controller:test-integration-envtest]                 },
[controller:test-integration-envtest]                 ObservedGeneration: 1,
[controller:test-integration-envtest]                 Reason: "InvalidOperation",
[controller:test-integration-envtest]                 Message: "Unable to undelete service 'asotest-apim-sczlzm'.: PUT https://management.azure.com/subscriptions/d5a106d2-ebae-4327-b2e8-c21962d4aaf5/resourceGroups/asotest-rg-pwokuh/providers/Microsoft.ApiManagement/service/asotest-apim-sczlzm\n--------------------------------------------------------------------------------\nRESPONSE 400: 400 Bad Request\nERROR CODE: InvalidOperation\n--------------------------------------------------------------------------------\n{\n  \"error\": {\n    \"code\": \"InvalidOperation\",\n    \"message\": \"Unable to undelete service 'asotest-apim-sczlzm'.\",\n    \"details\": null,\n    \"innerError\": null\n  }\n}\n--------------------------------------------------------------------------------\n",
```

ross-p-smith:
Whilst adding the new v2 APIM SKU in, the error message you get if you re-run any of the envtest a second time, is as follows

```text
[...]

PR #3467, opened 2023-10-25
APIM: Delete subscriptions when deleting products

### LINKED ISSUE #3408: APIM: Product should delete any Subscriptions it owns when it deletes
When you create an APIM Product it creates a Subscription. When you come to delete the Product, you will get an error saying you must delete the Subscription first

The ApiManagementDeleteProducts.json shows that an extra parameter is passed in the body 
`"deleteSubscriptions": true`

Map that field into the Code Generation

ross-p-smith:
This pull request fixes two APIM issues

# Goal Seeker behaviour
My previous work on APIM left a setting of `Recover: true` when creating the apim instance. This will cause an error if you were to delete the recording and not have an apim instance in a deleted state.

It would be better not to have this setting defined and allow it to error if there was a deleted APIM in your resource group - which is a smaller probability!

This would eventually be handled by having a goal seeking APIM creation #3433.

# Test deletion of products
When we create an APIM Product, it creates a Subscription for that Product. This causes problems when you delete the Product; however there is a flag to pass deleteSubscriptions to true which will fix this. Override the default delete behaviour and use the go sdk to carry out the deletion. This will close #3408

gif

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/v2/api/apimanagement/customizations/product_extensions.go
+++ b/v2/api/apimanagement/customizations/product_extensions.go
@@ -71,12 +71,3 @@ func (extension *ProductExtension) Delete(
 	// Is it safe to proceed as we have already deleted it?
 	return next(ctx, log, resolver, armClient, obj)
 }
-
-func getProductID(typedObj *storage.Product) (string, bool) {
-	// Get the subscription ID
-	if typedObj.Status.Id == nil || *typedObj.Status.Id == "" {
-		return "", false
-	}
-
-	return *typedObj.Status.Id, true
-}
--- a/v2/api/apimanagement/customizations/product_extensions.go
+++ b/v2/api/apimanagement/customizations/product_extensions.go
@@ -57,6 +57,7 @@ func (extension *ProductExtension) Delete(
 		return ctrl.Result{}, errors.Wrapf(err, "failed to create new apimClient")
 	}
 
+	// This is a synchronous operation
 	_, err = clientFactory.NewProductClient().Delete(
 		ctx,
 		id.ResourceGroupName,
@@ -70,6 +71,5 @@ func (extension *ProductExtension) Delete(
 		return ctrl.Result{}, errors.Wrapf(err, "failed to delete product %q", productName)
 	}
 
-	// Is it safe to proceed as we have already deleted it?
 	return next(ctx, log, resolver, armClient, obj)
 }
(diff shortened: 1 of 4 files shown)


v2/api/apimanagement/customizations/product_extensions.go
```diff
@@ -0,0 +1,73 @@
...
+		parentName,
+		productName,
+		"*",
+		&armapimanagement.ProductClientDeleteOptions{DeleteSubscriptions: to.Ptr(true)})
+	if err != nil {
+		return ctrl.Result{}, errors.Wrapf(err, "failed to delete product %q", productName)
+	}
+
+	// Is it safe to proceed as we have already deleted it?
+	return next(ctx, log, resolver, armClient, obj)
```
theunrepentantgeek, line 72: Yes, this is safe. ASO will attempt the deletion, then update all necessary bookkeeping.

v2/api/apimanagement/customizations/product_extensions.go
```diff
@@ -0,0 +1,73 @@
...
+	parentName := id.Parent.Name
+
+	// Using armClient.ClientOptions() here ensures we share the same HTTP connection, so this is not opening a new
+	// connection each time through
+	clientFactory, err := armapimanagement.NewClientFactory(id.SubscriptionID, armClient.Creds(), armClient.ClientOptions())
+	if err != nil {
+		return ctrl.Result{}, errors.Wrapf(err, "failed to create new apimClient")
+	}
+
+	_, err = clientFactory.NewProductClient().Delete(
```
theunrepentantgeek, line 60: Is `Delete` synchronous? That is, finished by the time the call returns? Or does it return a long running operation you should monitor until the delete is complete?

theunrepentantgeek (changes requested):
Only needs a couple of minor updates, then will be ready to merge.

v2/api/apimanagement/customizations/product_extensions.go
ross-p-smith: My understanding is that Delete is synchronous. According to https://learn.microsoft.com/en-us/azure/developer/go/management-libraries#long-running-operations they all begin with `Begin`.

Documentation of creating the actual APIM Service is a long running operation and is called out as one e.g. https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/apimanagement/armapimanagement#ServiceClient.BeginCreateOrUpdate

theunrepentantgeek (approved):

PR #3468, opened 2023-10-26
Fix: Resource doesn't re-reconcile after completed update

### LINKED ISSUE #3451: Bug: Resource doesn't re-reconcile after completed update

v2.3.0

**Describe the bug**
After making a change to the `spec` of a resource while that resource is already being updated, the new change does not apply until after the next periodic resync.

**To Reproduce**
(This is the particular case I was hitting, not exactly sure how general this applies.)
- Create a ManagedClustersAgentPool with `spec.orchestratorVersion = 1.27.1` and `spec.count = 1`
- Update `spec.orchestratorVersion` to `1.27.3`
- While the upgrade is happening, update `spec.count` to 2
- After the upgrade finishes, `status.orchestratorVersion` is `1.27.3` as expected, but `status.count` is stuck at 1 and there are no signs of a scaling event occurring in the portal until ASO's next periodic resync.

**Expected behavior**
A scaling operation begins shortly after the Kubernetes version upgrade completes.

**Screenshots**

**Additional context**

super-harsh:
Closes #3451 

This PR adds a latest-reconciled-generation annotation to the resource every time the reconciler receives a createOrUpdate event and further is checked in `MonitorResourceCreation`. We need this check in `MonitorResourceCreation` since when a subsequent update is received, it always has a `resume-poller-token` annotation which makes the reconciler think that it's monitoring the resource and the update never happened. 

The generation check runs once the poller is done or runs into an error, and requeues the resource if generation set by API server does not match with the annotation or else returns successfully.


theunrepentantgeek (approved):
That came out really cleanly.

v2/internal/reconcilers/arm/annotations.go
```diff
@@ -32,3 +35,16 @@ func ClearPollerResumeToken(obj genruntime.MetaObject) {
 	genruntime.RemoveAnnotation(obj, PollerResumeTokenAnnotation)
 	genruntime.RemoveAnnotation(obj, PollerResumeIDAnnotation)
 }
+
+func SetLatestReconciledGeneration(obj genruntime.MetaObject) {
```
matthchr, line 39: I thought we could just use the Generation on the status? Does it not work?

super-harsh: Ah! it does not. Since we whenever we're polling and polling is not finished, we requeue the resource with `retryAfter` option. Which then requeues the resource and we always set Reconciling condition in `generic_reconciler.go/createOrUpdate` method which updates the observedGeneration on status. Which then is never picked up by the checker we want to add.

matthchr:
Hmm, the added test seems to fail?

v2/internal/controllers/latest_reconciled_generation_test.go
```diff
@@ -65,7 +67,7 @@ func Test_Latest_Reconciled_Generation_Reconciles_AllEvents(t *testing.T) {
 	old := agentPool.DeepCopy()
 	agentPool.Spec.OrchestratorVersion = to.Ptr("1.27.1")
 
-	tc.Patch(old, agentPool)
+	tc.PatchResourceAndWaitForState(old, agentPool, metav1.ConditionFalse, conditions.ConditionSeverityInfo)
```
matthchr, line 70: Did you confirm that doing this was still able to trigger the test on the behavior we wanted (that 2 PUTs happen before the first reconcile is finished)?

super-harsh: Yep, It does take that - generation mismatch path on re-runs.

matthchr (approved):

PR #3469, opened 2023-10-27
Allow renaming of properties

### LINKED ISSUE #3385: Bug: Can't import resources with existing Conditions property
**Describe the bug**

As discussed [on Slack](https://kubernetes.slack.com/archives/C046DEVLAQM/p1696700048253479), the CDN Profile Rule resource has an existing `conditions` property which conflicts with the ASO code generator's need to inject an identically named property into the status of the resource.

The error from the generator:
```
[controller:generate-types] 17:31:40 ERR failed to execute root command error="
error generating code: failed to execute stage 45: 
Add the property 'Conditions' to all status types and implements genruntime.Conditioner on all resources: 
couldn't add Conditions condition to status \"github.com/Azure/azure-service-operator/v2/api/cdn/v1api20230501/Profiles_RuleSets_Rule_STATUS\": 
failed to inject property \"Conditions\" into \"github.com/Azure/azure-service-operator/v2/api/cdn/v1api20230501/Profiles_RuleSets_Rule_STATUS\": 
visit of type of \"github.com/Azure/azure-service-operator/v2/api/cdn/v1api20230501/Profiles_RuleSets_Rule_STATUS\" failed: 
already has property named \"Conditions\""
task: Failed to run task "asoctl:quick-checks": task: Failed to run task "controller:generate-types": exit status 1
```

See [afdx.json](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cdn/resource-manager/Microsoft.Cdn/stable/2023-05-01/afdx.json) for details.

theunrepentantgeek:

As discussed in #3385, there are some resources we can't import because they have properties with names that collide with our requirement to be a well-behaved Kubernetes resource - such as `Conditions`.

This PR adds the ability to rename CRD properties to avoid collisions - new configuration and a pipeline stage to use it. Also includes a few miscellaneous improvements.

Closes #3385 

Tested by reproducing the scenario noted in #3385, renaming `Conditions` to `DeliveryRuleConditions`, and then manually reviewing the generated code:

* Many subordinate types are moved to different output files.
* The ARM types retain `Conditions` with serialization to `conditions`, maintaining the required ARM payload.
* Code to populate the ARM types correctly copies values between the two properties despite having different names.

gif


v2/tools/generator/internal/codegen/code_generator.go
```diff
@@ -301,7 +302,8 @@ func (generator *CodeGenerator) Generate(
 			stageDescription,
 			"elapsed", duration,
 			"added", len(defsAdded),
-			"removed", len(defsRemoved))
+			"removed", len(defsRemoved),
+			"count", len(newState.Definitions()))
```
matthchr, line 306: minor: Would `total` be any clearer here? TBH if I couldn't see the variable names it might not be super clear to me what these numbers were - added/removed/count of what? Maybe we should consider putting a suffix like `addedDefs` or similar?

v2/tools/generator/internal/codegen/pipeline/rename_properties.go
```diff
@@ -0,0 +1,121 @@
...
+
+	stage.RequiresPostrequisiteStages(
+		AddStatusConditionsStageID, // Must rename other properties before we try to introduce the `Conditions` property
+		AddOperatorSpecStageID,     // Must rename other properties before we try to introduce the `OperatorSpec` property
+	)
+
+	return stage
+}
+
+func createPropertyRenamingTypeVisitor(
```
matthchr, line 73: very minor: Could be the slightly shorter `createPropertyRenamingVisitor`, maybe?

v2/tools/generator/internal/codegen/pipeline/rename_properties.go
```diff
@@ -0,0 +1,121 @@
...
+
+	return builder.Build()
+}
+
+func renamePropertiesInObjectType(
+	typeName astmodel.InternalTypeName,
+	ot *astmodel.ObjectType,
+	cfg *config.ObjectModelConfiguration,
+) (astmodel.Type, error) {
+
```
matthchr, line 94: minor: Inconsistent newline - you don't have it for the above method but do here, not clear why the difference?

v2/tools/generator/internal/codegen/pipeline/rename_properties.go
```diff
@@ -0,0 +1,121 @@
...
+	properties := make([]*astmodel.PropertyDefinition, 0, ot.Properties().Len())
+	err := ot.Properties().ForEachError(
+		func(prop *astmodel.PropertyDefinition) error {
+			name, err := cfg.RenamePropertyTo.Lookup(typeName, prop.PropertyName())
+			if err != nil {
```
matthchr, line 99: I was talking with Artur and Christian, and realized that `errors.New` and similar `errors` functions (such as `Wrapf`) capture stacktraces that we're not logging.

I think in this case we don't actually wrap the returned error before we check it here (it's returned directly from `Lookup` which doesn't use `errors.New` or any similar such things.

It does seem like it might be an easy mistake to make to end up wrapping an error that's being used for control-flow, triggering a ton of stacktraces to be gathered. I'm wondering if we should think of other ways to implement this pattern rather than using an error? Or maybe an error is fine and we should just be careful about wrapping it (I think we are in general).

Just something to think about.

v2/tools/generator/internal/codegen/pipeline/rename_properties.go
```diff
@@ -0,0 +1,121 @@
...
+				// If error is just that there's no configured rename, proceed without issue
+				if !config.IsNotConfiguredError(err) {
+					return errors.Wrapf(err, "looking up rename for %s", prop.PropertyName())
+				}
+
+				properties = append(properties, prop)
+				return nil
+			}
+
+			updated := prop.
```
matthchr, line 109: How exactly is this working with the ARM conversion logic? Don't we need a name remapping set up for the `convert_to_arm_function_builder` so that when it actually goes to generate conversions between the properties, it uses the `DeliveryRuleConditions` to source the conversion from rather than `Conditions`?

I think with the way the ARM conversion code is set up now, it does property matching by name (see `propertiesWithSameNameHandler` and its usage).

Do we need to expand that with some sort of tag (similar to `OptionalConfigMapPairTag` or `ARMReferenceTag`) on the property that indicates it's renamed?

I'm not convinced that this code running when it does generates the right thing w.r.t. the actual ARM shape. Can you share a draft branch with the sample code?

theunrepentantgeek: I manually checked the code and in the ARM situation you end up with a "flattened from" list containing one item (that is, the degenerate case) which results in the correct assignment. I don't have a draft branch, but will create one to confirm.

v2/tools/generator/internal/astmodel/type_visitor.go
```diff
@@ -109,14 +109,17 @@ func (tv *TypeVisitor[C]) VisitInternalTypeName(name InternalTypeName, ctx C) (I
 // VisitDefinition invokes the TypeVisitor on both the name and type of the definition
 // NB: this is only valid if visitTypeName returns a TypeName and not generally a Type
 func (tv *TypeVisitor[C]) VisitDefinition(td TypeDefinition, ctx C) (TypeDefinition, error) {
-	visitedName, err := tv.VisitInternalTypeName(td.Name(), ctx)
+	name := td.Name()
+	visitedName, err := tv.VisitInternalTypeName(name, ctx)
 	if err != nil {
-		return TypeDefinition{}, errors.Wrapf(err, "visit of %q failed", td.Name())
+		return TypeDefinition{}, errors.Wrapf(
+			err, "visit of %s/%s failed", name.InternalPackageReference().FolderPath(), name.Name())
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

v2/tools/generator/internal/astmodel/type_visitor.go
```diff
@@ -109,14 +109,17 @@ func (tv *TypeVisitor[C]) VisitInternalTypeName(name InternalTypeName, ctx C) (I
...
 	}
 
 	visitedType, err := tv.Visit(td.Type(), ctx)
 	if err != nil {
-		return TypeDefinition{}, errors.Wrapf(err, "visit of type of %q failed", td.Name())
+		return TypeDefinition{}, errors.Wrapf(
+			err, "visit of type of %s/%s failed", name.InternalPackageReference().FolderPath(), name.Name())
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

v2/tools/generator/internal/codegen/code_generator.go
theunrepentantgeek: Changed to `totalDefs`; I don't think `added` and `removed` need a suffix too, the implied context should be enough.

v2/tools/generator/internal/codegen/pipeline/rename_properties.go
theunrepentantgeek: Agreed.

v2/tools/generator/internal/codegen/pipeline/rename_properties.go
theunrepentantgeek: Debris. 😄 #removed

v2/tools/generator/internal/codegen/pipeline/rename_properties.go
theunrepentantgeek: I wrote `Lookup()` to return `(value, error)` because returning a triple of `(value, bool, error)` seemed more than a bit nasty. We could change it to return `(LookupResult, error)` where `LookupResult` took care of returning the `NotFound` outcome.

theunrepentantgeek: I've logged #3545 to capture this.

v2/tools/generator/internal/codegen/pipeline/rename_properties.go
theunrepentantgeek: I've added explicit tests for property renaming to cover both the normal and the flattened cases. This found a couple of issues that I've addressed.

v2/tools/generator/internal/codegen/pipeline/rename_properties.go
matthchr: Nice. Resolving this comment as you've definitely addressed it as much as appropriate in this PR.

v2/tools/generator/internal/armconversion/convert_to_arm_function_builder.go
```diff
@@ -537,13 +537,24 @@ func (builder *convertToARMBuilder) buildToPropInitializer(
...
 	fromType *astmodel.ObjectType,
 ) (propertyConversionHandlerResult, error) {
 
-	fromProp, ok := fromType.Property(toProp.PropertyName())
-	if !ok {
+	fromProp, found := fromType.Property(toProp.PropertyName())
+	if !found {
+		// No direct match by name, look for renames
+		// Look for a candidate fromProp whose original name matches toProp
+		fromType.Properties().ForEach(func(p *astmodel.PropertyDefinition) {
```
matthchr, line 549: This came out pretty clean. 

You also could have done this w/ property tags as well if you wanted. How you did it works for me though.

matthchr (approved):

super-harsh (commented):
Looks good! Do we also need some sort of guide/doc with an example on how to use the new configuration?

theunrepentantgeek:
> Do we also need some sort of guide/doc with an example on how to use the new configuration?

Most of our config documentation is within the yaml file at the moment, so it's local. I'll create an issue to track whether we want to do more.

PR #3442, opened 2023-10-18
Allow to customize webhook server port and cert dir

vincepri:
Closes #[issue number]

**What this PR does / why we need it**:

Allows users to customize the port and certificate directory. Aligns with Cluster API flags.

**If applicable**:
- [ ] this PR contains documentation
- [ ] this PR contains tests

PR #3470, opened 2023-10-27
Support setting the webhook port and cert dir in the Helm chart

### LINKED ISSUE #3444: Support setting the webhook port in the Helm chart
This goes along with the behavior which is already available via cmdline option here: #3442.

In order to make this work obviously it would need to go into values.yaml for Helm but we also would need to update our CRD management to flow the right port to the conversion webhook section of the CRDs.

adamdougal:
Closes #3444

Provides the ability to set the webhook port and cert dir when deploying using the helm chart. These options were added in https://github.com/Azure/azure-service-operator/pull/3442.

- In the linked issue it's mentioned that the port needs to be set in the conversion webhook section of the CRDs, however, these reference the service and the service port is unchanged.
- This chart does not work with the latest **released** version of the app and Docker image as they do not support these flags, so the release of this will need to wait until a new app release is cut. Alternatively, the implementation of this PR could be changed to only set these flags if they need to be overridden, rather than every time. I considered this option but favoured a simpler helm template.

gif

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/scripts/v2/generate-helm-manifest.sh
+++ b/scripts/v2/generate-helm-manifest.sh
@@ -84,8 +84,12 @@ flow_control "aadpodidbinding" "aadpodidbinding" "$IF_TENANT" "$GEN_FILES_DIR"/*
 flow_control "--enable-leader-election" "--enable-leader-election" "$IF_TENANT" "$GEN_FILES_DIR"/*_deployment_*
 
 # TODO: This bit is tricky to exclude kube-rbac-proxy and webhook stuff.
-flow_control "serving-certs" "name: https" "$IF_CLUSTER" "$GEN_FILES_DIR"/*_deployment_*
+flow_control "mountPath: \/tmp\/k8s-webhook-server\/serving-certs" "name: https" "$IF_CLUSTER" "$GEN_FILES_DIR"/*_deployment_*
 flow_control "- name: cert" "secretName" "$IF_CLUSTER" "$GEN_FILES_DIR"/*_deployment_*
+flow_control "--webhook-cert-dir=" "--webhook-cert-dir=" "$IF_CLUSTER" "$GEN_FILES_DIR"/*_deployment_*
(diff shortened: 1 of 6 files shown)


super-harsh (commented):
@adamdougal #3444 is for ASOv2 helm chart and the changes in your PR are for ASOv1 helm chart. We have ASOv2 helm chart [here](https://github.com/Azure/azure-service-operator/tree/main/v2/charts/azure-service-operator) for which we generate files using task target [controller:gen-helm-manifest](https://github.com/Azure/azure-service-operator/blob/main/Taskfile.yml#L667)

adamdougal:
@super-harsh Dang, I missed the charts dir under `v2`. I'll get this fixed up. Thanks

scripts/v2/generate-helm-manifest.sh
```diff
@@ -84,8 +84,12 @@ flow_control "aadpodidbinding" "aadpodidbinding" "$IF_TENANT" "$GEN_FILES_DIR"/*
 flow_control "--enable-leader-election" "--enable-leader-election" "$IF_TENANT" "$GEN_FILES_DIR"/*_deployment_*
 
 # TODO: This bit is tricky to exclude kube-rbac-proxy and webhook stuff.
-flow_control "serving-certs" "name: https" "$IF_CLUSTER" "$GEN_FILES_DIR"/*_deployment_*
+flow_control "mountPath: \/tmp\/k8s-webhook-server\/serving-certs" "name: https" "$IF_CLUSTER" "$GEN_FILES_DIR"/*_deployment_*
```
adamdougal, line 87: This had to be made more specific so not to clash with the new arg

matthchr (approved):
Thanks for the contribution, this looks good to me. I've kicked off E2E testing to ensure that passes but from my examination of the resulting Helm chart it should.

As you highlighted, we will hold off publishing a new chart until we also publish a new container image (probably later this week).

super-harsh (approved):

Issue #3398, opened 2023-10-10
APIM: Generate types that have no body (e.g. ProductAPI)


ross-p-smith:
There are a few APIM resources that are created with a PUT and no body. This causes them to not be generated.

Examples can be found in `/workspaces/azure-service-operator/v2/specs/azure-rest-api-specs/specification/apimanagement/resource-manager/Microsoft.ApiManagement/stable/2022-08-01/apimproducts.json`

- ProductApi
- ProductGroup
- ProductSubscription
- ProductPolicy
- ProductTag
- ProductWiki

We could change the generator to generate these types.

matthchr:
I'm looking into this. I suspect the fix in the generator should be relatively easy.

matthchr:
I have this fixed locally but realized that one gotcha is that these resources don't actually have a GET, which we rely on in the reconciler currently. We need to fix that too. 

Need some way to indicate at codegen time what operations a resource supports and then hook the operator in to looking at that. Shouldn't be too hard, will work on this as well.

tomkerkhove:
https://github.com/Azure/azure-service-operator/issues/2747 depends on this

PR #3525, opened 2023-11-07
Support resource types that aren't quite "really" resources

matthchr:
They have a HEAD rather than a GET and no body parameters

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/v2/tools/generator/internal/codegen/embeddedresources/remove_empty_objects.go
+++ b/v2/tools/generator/internal/codegen/embeddedresources/remove_empty_objects.go
@@ -40,6 +40,10 @@ func findEmptyObjectTypes(
 ) astmodel.TypeNameSet {
 	result := astmodel.NewTypeNameSet()
 
+	// Get all top level spec and status types
+	specs := astmodel.FindSpecDefinitions(definitions)
+	statuses := astmodel.FindStatusDefinitions(definitions)
+
 	for _, def := range definitions {
 		ot, ok := astmodel.AsObjectType(def.Type())
 		if !ok {
@@ -52,6 +56,11 @@ func findEmptyObjectTypes(
 		}
 
 		if astmodel.DoNotPrune.IsOn(def.Type()) {
+			continue
+		}
+
+		// Don't prune top level status or spec types, they are allowed to be empty
+		if specs.Contains(def.Name()) || statuses.Contains(def.Name()) {
 			continue
 		}
 
--- a/v2/tools/generator/internal/jsonast/swagger_type_extractor.go
+++ b/v2/tools/generator/internal/jsonast/swagger_type_extractor.go
@@ -138,8 +138,8 @@ func (extractor *SwaggerTypeExtractor) ExtractResourceTypes(ctx context.Context,
 	}
 
 	for rawOperationPath, op := range extractor.swagger.Paths.Paths {
-		// a resource must have both PUT and GET
-		if op.Put == nil || op.Get == nil {
+		// a resource must have a PUT and one of GET or HEAD
+		if op.Put == nil || (op.Get == nil && op.Head == nil) {
 			continue
 		}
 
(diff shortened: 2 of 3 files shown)


matthchr:
This is related to #3398

v2/tools/generator/internal/codegen/embeddedresources/remove_empty_objects.go
```diff
@@ -55,6 +55,11 @@ func findEmptyObjectTypes(
 			continue
 		}
 
+		// Don't prune top level status or spec types, they are allowed to be empty
+		if def.Name().IsStatus() || def.Name().IsSpec() {
```
theunrepentantgeek, line 59: These methods only check based on naming convention - which is extended past the top level types in some cases.

Examples:
* In `apimanagement` - [BackendServiceFabricClusterProperties_STATUS](https://azure.github.io/azure-service-operator/reference/apimanagement/v1api20220801/#apimanagement.azure.com/v1api20220801.BackendServiceFabricClusterProperties_STATUS)
* In `cache` - [Persistence_STATUS](https://azure.github.io/azure-service-operator/reference/cache/v1api20210301/#cache.azure.com/v1api20210301.Persistence_STATUS)

You need another way to detect the top level spec/status.

theunrepentantgeek (changes requested):
Your check for top level Spec and Status types is flawed, but otherwise this is good.

theunrepentantgeek (commented):
Removing my _Request Changes_ so you're not blocked addressing this before I'm available tomorrow morning.

v2/tools/generator/internal/codegen/embeddedresources/remove_empty_objects.go
matthchr: Good eye. Will fix!

matthchr: Fixed

v2/tools/generator/internal/jsonast/swagger_type_extractor.go
```diff
@@ -118,7 +118,7 @@ func (extractor *SwaggerTypeExtractor) ExtractResourceTypes(ctx context.Context,
 
 	for rawOperationPath, op := range extractor.swagger.Paths.Paths {
 		// a resource must have both PUT and GET
```
theunrepentantgeek, line 120:
```suggestion
		// a resource must have a PUT and one of GET or HEAD
```

v2/tools/generator/internal/jsonast/swagger_type_extractor.go
```diff
@@ -191,7 +191,10 @@ func (extractor *SwaggerTypeExtractor) extractOneResourceType(
 	var resourceSpec astmodel.Type
 	if specSchema == nil {
 		// nil indicates empty body
-		resourceSpec = astmodel.NewObjectType()
+		// Fabricate a TypeName here because that's what all other specs are (typename pointing to parameters type)
+		name := astmodel.MakeInternalTypeName(resourceName.InternalPackageReference(), fmt.Sprintf("%sCreateParameters", resourceName.Name()))
```
theunrepentantgeek, line 195: Should this use an IdFactory instead?

v2/tools/generator/internal/jsonast/swagger_type_extractor.go
```diff
@@ -313,7 +319,10 @@ func (extractor *SwaggerTypeExtractor) findARMResourceSchema(op spec.PathItem, r
 	isResource := false
 
 	var foundStatus *Schema
-	if op.Get.Responses != nil {
+	// The previous assumption that all ARM resources have a GET is not correct, see for example
+	// https://learn.microsoft.com/en-us/azure/templates/microsoft.apimanagement/service/products/groups?pivots=deployment-language-arm-template
+	// which has a head not a GET
```
theunrepentantgeek, line 324:
```suggestion
	// which has HEAD and not GET
```

v2/tools/generator/internal/jsonast/swagger_type_extractor.go
```diff
@@ -328,6 +337,11 @@ func (extractor *SwaggerTypeExtractor) findARMResourceSchema(op spec.PathItem, r
 		}
 	}
 
+	if isResource == false && op.Head != nil {
+		// assume this is a resource?
```
theunrepentantgeek, line 341: Do you know the answer to this yet?

theunrepentantgeek (approved):

v2/tools/generator/internal/jsonast/swagger_type_extractor.go
matthchr: yeah I think this is OK, at least for now. If this is problematic it's because it may over-classify things

I've updated this comment to say:
> 		// assume this is a resource. It's possible this is too permissive and classifies some things as resources
>		// when they really aren't, but that's OK because anyway we only generate the resources we name in the configuration
>		// and classifying something as a resource here just makes it a candidate for mention in the config.

v2/tools/generator/internal/jsonast/swagger_type_extractor.go
matthchr: I don't think it's required, as we're really just appending a suffix here to an already generated name. This is similar to how we append names for `StatusSuffix`, which does:
`resourceName.WithName(resourceName.Name() + astmodel.StatusSuffix)`

Issue #3435, opened 2023-10-16
Support collection of SecretReferences


super-harsh:
**Describe the current behavior**

Currently `isSecret` field for the generator, only lets us specify the string fields to mark as a secret reference.

**Describe the improvement**

We should be able to mark collections(map, slice) as a secret reference

**Additional context**

Prerequisite for #2934

tomkerkhove:
I see it is mentioned that this is a requirement for https://github.com/Azure/azure-service-operator/pull/3528, but it has been merged.

Does that mean that this has been completed, @matthchr? I presume it is not?

matthchr:
It's not done, we worked around it in #3528.

I'm working on this now though.

PR #3528, opened 2023-11-07
Add support for Microsoft.KubernetesConfiguration/extensions

### LINKED ISSUE #2934: Feature: Support for Azure container marketplace offerings (Microsoft.KubernetesConfiguration/extensions)
Add ability to install any extension such as [Azure container marketplace offers](https://learn.microsoft.com/en-us/azure/aks/deploy-marketplace) or Arc Extension (e.g. Flux v2) onto AKS clusters.

[Rest API is here](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/kubernetesconfiguration/resource-manager/Microsoft.KubernetesConfiguration/stable/2022-11-01/extensions.json)

super-harsh:
Closes #2934 

This PR adds support for Microsoft.KubernetesConfiguration/extensions. 

_Also includes a generator fix to identify the extension types_

Removing the `ConfigurationProtectedSettings` configuration from the resource for now as it's dependent on #3435 being resolved. Once resolved, we can revert this field to be on the resource.


super-harsh:
TODO: Need to add sample and record sample test

v2/api/kubernetesconfiguration/v1api20221101/structure.txt
```diff
@@ -0,0 +1,267 @@
+// Code generated by azure-service-operator-codegen. DO NOT EDIT.
+github.com/Azure/azure-service-operator/v2/api/kubernetesconfiguration/v1api20221101
+├── APIVersion: Enum (1 value)
+│   └── "2022-11-01"
```
matthchr, line 4: There's a 2023-05-01 API version, we should probably use that instead?

v2/api/kubernetesconfiguration/v1api20221101/structure.txt
```diff
@@ -0,0 +1,267 @@
...
+│   │   │   ├── PromotionCode: *string
+│   │   │   ├── Publisher: *string
+│   │   │   └── Version: *string
+│   │   ├── ReleaseTrain: *string
+│   │   ├── Scope: *Object (2 properties)
+│   │   │   ├── Cluster: *Object (1 property)
+│   │   │   │   └── ReleaseNamespace: *string
+│   │   │   └── Namespace: *Object (1 property)
+│   │   │       └── TargetNamespace: *string
+│   │   ├── Statuses: Object (5 properties)[]
```
matthchr, line 31: Hmm, can this field actually be set? Given its name I am wondering if it's really supposed to be marked readonly?

v2/azure-arm.yaml
```diff
@@ -1587,6 +1599,11 @@ objectModelConfiguration:
       VirtualNetworkRule:
         Id:
           $armReference: true
+  kubernetesconfiguration:
+    2022-11-01:
```
matthchr, line 1603: As mentioned above, 2023-05-01 is a newer API version than this -- should we use that instead?

v2/internal/controllers/samples_test.go
```diff
@@ -39,6 +39,7 @@ var randomNameExclusions = []string{
 	"compute",
 	"documentdb",
 	"insights",
+	"kubernetesconfiguration",
```
matthchr, line 42: minor: Why does this need to avoid random name exclusions?

super-harsh: Cz the owner exists as a reference object, if we use the random names, there's no way the sample can find the owner if we use random names.

super-harsh: Now if I think more of it.. its a tech-debt - I'll open an issue for this and add a todo on the exclusion.

v2/api/kubernetesconfiguration/v1api20221101/structure.txt
matthchr: Checking on this

v2/internal/controllers/samples_test.go
matthchr: Missing a TODO here?

v2/api/kubernetesconfiguration/v1api20221101/structure.txt
super-harsh: Done!

v2/internal/controllers/samples_test.go
super-harsh: Done!

matthchr (approved):
LGTM!

PR #3530, opened 2023-11-08
Controller should use HEAD if GET is not available

### LINKED ISSUE #3398: APIM: Generate types that have no body (e.g. ProductAPI)

matthchr:
This fixes #3398.

Note that there are no tests for this (yet), but they will come when we add the APIM resources which don't have a body (which will come in a future PR). I've already confirmed that these tests pass.

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
+++ b/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
@@ -559,18 +559,35 @@ func (r *azureDeploymentReconcilerInstance) getStatus(ctx context.Context, id st
 	}
 
 	// Get the resource
-	retryAfter, err := r.ARMConnection.Client().GetByID(ctx, id, apiVersion, armStatus)
-	if err != nil {
-		return nil, retryAfter, errors.Wrapf(err, "getting resource with ID: %q", id)
-	}
-
-	if r.Log.V(Debug).Enabled() {
-		statusBytes, marshalErr := json.Marshal(armStatus)
-		if marshalErr != nil {
-			return nil, zeroDuration, errors.Wrapf(marshalErr, "serializing ARM status to JSON for debugging")
+	if genruntime.ResourceOperationGet.IsSupportedBy(r.Obj) {
+		var retryAfter time.Duration
+		retryAfter, err = r.ARMConnection.Client().GetByID(ctx, id, apiVersion, armStatus)
+		if err != nil {
+			return nil, retryAfter, errors.Wrapf(err, "getting resource with ID: %q", id)
 		}
 
-		r.Log.V(Debug).Info("Got ARM status", "status", string(statusBytes))
+		if r.Log.V(Debug).Enabled() {
+			statusBytes, marshalErr := json.Marshal(armStatus)
+			if marshalErr != nil {
+				return nil, zeroDuration, errors.Wrapf(marshalErr, "serializing ARM status to JSON for debugging")
+			}
+
+			r.Log.V(Debug).Info("Got ARM status", "status", string(statusBytes))
+		}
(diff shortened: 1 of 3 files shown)


v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -559,18 +560,29 @@ func (r *azureDeploymentReconcilerInstance) getStatus(ctx context.Context, id st
...
-		r.Log.V(Debug).Info("Got ARM status", "status", string(statusBytes))
+		if r.Log.V(Debug).Enabled() {
+			statusBytes, marshalErr := json.Marshal(armStatus)
+			if marshalErr != nil {
+				return nil, zeroDuration, errors.Wrapf(marshalErr, "serializing ARM status to JSON for debugging")
+			}
+
+			r.Log.V(Debug).Info("Got ARM status", "status", string(statusBytes))
+		}
+	} else if operations.IsSupported(genruntime.ResourceOperationHead, r.Obj.GetSupportedOperations()) {
```
theunrepentantgeek, line 578: Why not make `IsSupported()` a member function of Operation, and add an interface for `GetSupportedOperations()` so that you can write this:

```suggestion
	} else if genruntime.ResourceOperationHead.IsSupportedBy(r.Obj) {
```

which I think would be clearer and more intention revealing.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -559,18 +560,29 @@ func (r *azureDeploymentReconcilerInstance) getStatus(ctx context.Context, id st
...
+		var retryAfter time.Duration
+		retryAfter, err = r.ARMConnection.Client().CheckExistenceByID(ctx, id, apiVersion)
+		if err != nil {
+			return nil, retryAfter, errors.Wrapf(err, "getting resource with ID: %q", id)
+		}
+	} else {
+		return nil, zeroDuration, errors.Errorf("resource must support one of GET or HEAD, but it supports neither")
```
theunrepentantgeek, line 585: Should this include details (e.g. GVK) of _which_ resource has a problem? Or are those details added elsewhere?

theunrepentantgeek (approved):
A couple of minor comments. Structure is good.

super-harsh (approved):

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
matthchr: The ARM ID gets included in the wrapping caller (`updateStatus`). It also eventually gets logged by the logger at the top of the reconcile loop, which has been constructed to have the resource name/type, so I think it's probably OK to omit the name here (and in general in all errors).

The logs should have the context, and the user will also see this error exposed as a condition on the resource, but saying the name of the resource in that context seems needless, as obviously they're already look at the resource to see the condition.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
matthchr: Yeah this is definitely clearer. Done.

PR #3546, opened 2023-11-13
Add guards against using distinct subscriptions between owner and child resources

### LINKED ISSUE #3244: Bug: Child resource and parent resource can attempt to use different subscriptions when using serviceoperator.azure.com/credential-from
ASOv2

**Describe the bug**
If a child resource has the `serviceoperator.azure.com/credential-from` annotation pointing to a secret which specifies a different subscription than its parent, it tries to use that subscription rather than the parent subscription and deployment of the child resource will fail.

**Expected behavior**
Probably we should return a clear error in this case instead.

super-harsh:
Closes #3244 

Adding check to match resource credential subscription with owner's subscription.

- [x] this PR contains tests

### DIFF
--- a/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
+++ b/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
@@ -810,7 +810,7 @@ func (r *azureDeploymentReconcilerInstance) deleteResource(
 		return ctrl.Result{}, nil
 	}
 
-	err := r.checkSubscription(resourceID) // TODO: Possibly we should pass this in as a parameter?
+	err := r.checkSubscription(resourceID)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
--- a/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
+++ b/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
@@ -293,7 +293,7 @@ func (r *azureDeploymentReconcilerInstance) BeginCreateOrUpdateResource(
 
 	resourceID := genruntime.GetResourceIDOrDefault(r.Obj)
 	if resourceID != "" {
-		err := r.checkSubscription(resourceID)
+		err = r.checkSubscription(resourceID)
 		if err != nil {
 			return ctrl.Result{}, err
 		}
--- a/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
+++ b/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
@@ -293,7 +293,7 @@ func (r *azureDeploymentReconcilerInstance) BeginCreateOrUpdateResource(
 
 	resourceID := genruntime.GetResourceIDOrDefault(r.Obj)
 	if resourceID != "" {
-		err = checkSubscription(resourceID, r.ARMConnection.SubscriptionID())
+		err := r.checkSubscription(resourceID)
 		if err != nil {
 			return ctrl.Result{}, err
 		}
@@ -303,7 +303,6 @@ func (r *azureDeploymentReconcilerInstance) BeginCreateOrUpdateResource(
 	if err != nil {
 		return ctrl.Result{}, err
 	}
-
 	// Use conditions.SetConditionReasonAware here to override any Warning conditions set earlier in the reconciliation process.
 	// Note that this call should be done after all validation has passed and all that is left to do is send the payload to ARM.
 	conditions.SetConditionReasonAware(r.Obj, r.PositiveConditions.Ready.Reconciling(r.Obj.GetGeneration()))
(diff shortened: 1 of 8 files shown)


v2/internal/resolver/resolver_test.go
```diff
@@ -89,6 +90,9 @@ func createResourceGroup(name string) *resources.ResourceGroup {
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
 			Namespace: testNamespace,
+			Annotations: map[string]string{
+				genruntime.ResourceIDAnnotation: fmt.Sprintf("/subscriptions/1234/resourceGroups/%s", name),
```
theunrepentantgeek, line 94: Suggest using the zero GUID because that's the format for all subscription IDs. 

```suggestion
				genruntime.ResourceIDAnnotation: fmt.Sprintf("/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/%s", name),
```

v2/test/single_operator_multitenant_test.go
```diff
@@ -133,6 +134,43 @@ func Test_Multitenant_SingleOperator_PerResourceCredential(t *testing.T) {
...
+	tc.Expect(err).To(BeNil())
+
+	tc.CreateResource(secret)
+
+	rg := tc.CreateTestResourceGroupAndWait()
+
+	acct := newStorageAccount(tc, rg)
+	acct.Annotations = map[string]string{annotations.PerResourceSecret: secret.Name}
+
+	// Creating new storage account in with distinct subscription, per resource secret should fail.
```
theunrepentantgeek, line 152:
```suggestion
	// Create a new storage account with a distinct subscription; per resource secret should fail because it does not match
```

v2/test/single_operator_multitenant_test.go
```diff
@@ -133,6 +134,43 @@ func Test_Multitenant_SingleOperator_PerResourceCredential(t *testing.T) {
...
+	tc.CreateResourceAndWaitForFailure(acct)
+	tc.Expect(acct.Status.Conditions[0].Message).To(ContainSubstring("does not match parent subscription"))
+	tc.Expect(acct.Status.Conditions[0].Severity).To(Equal(conditions.ConditionSeverityError))
+
+	// Deleting the per-resource credential annotation would default to applying the global credential with all permissions
```
theunrepentantgeek, line 157:
```suggestion
	// Deleting the per-resource credential annotation should default to applying the global credential with all permissions
```

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -809,7 +810,7 @@ func (r *azureDeploymentReconcilerInstance) deleteResource(
 		return ctrl.Result{}, nil
 	}
 
-	err := checkSubscription(resourceID, r.ARMConnection.SubscriptionID()) // TODO: Possibly we should pass this in as a parameter?
+	err := r.checkSubscription(resourceID) // TODO: Possibly we should pass this in as a parameter?
```
matthchr, line 813: minor: Probably can remove this TODO? I am not sure what it was referencing but the code seems OK to me now

v2/internal/resolver/resource_hierarchy.go
```diff
@@ -261,6 +274,18 @@ func (h ResourceHierarchy) fullyQualifiedARMIDImpl(subscriptionID string, origin
 	}
 }
 
+func (h ResourceHierarchy) matchOwnerSubscription(subscriptionID string, ownerRID *arm.ResourceID) error {
```
matthchr, line 277: minor:
1. This method doesn't need to be on `ResourceHierarchy`, as it doesn't seem to use `h` at all. It could just be a package scoped function.
2. The name `matchOwnerSubscription` seems too specific to me. In reality this is just checking if the provided subscription matches the passed in ResourceID, so possibly we could rename to something like `checkARMIDHasSubscription` or `checkARMIDMatchesSubscription` or similar?

v2/internal/resolver/resource_hierarchy.go
```diff
@@ -202,13 +219,9 @@ func (h ResourceHierarchy) fullyQualifiedARMIDImpl(subscriptionID string, origin
 		if err != nil {
```
matthchr, line 219: If you take the `checkARMIDHasSubscription` suggestion below, also consider renaming this `armID` variable to `ownerARMID` so the check reads: `checkARMIDHasSubscription(subscriptionID, ownerARMID)`

matthchr (approved):
Some naming comments/minor refactoring suggestions, but LGTM overall

v2/internal/resolver/resource_hierarchy.go
super-harsh: Yep, I like it. Considering to move the method into `genruntime/arm_id.go` for re-usability.

theunrepentantgeek (approved):

PR #3552, opened 2023-11-14
APIM: ProductPolicy and ProductApi

ross-p-smith:
Relates #3398

This implements ProductPolicy and ProductApi which was not possible to generate until #3530 was complete.

CRD sample tests for this are disabled because it takes too long to build.

gif

**If applicable**:
- [x] this PR contains tests


matthchr:
This won't work in the actual controller until I merge this: https://github.com/Azure/azure-service-operator/pull/3530

theunrepentantgeek:
We've merged a fix for the broken link that's breaking CI in #3595 - merge `main` into your branch and that should pass.

v2/azure-arm.yaml
```diff
@@ -883,6 +883,12 @@ objectModelConfiguration:
       Service_Product:
         $exportAs: Product
         $supportedFrom: v2.4.0
+      Service_Products_Api:
+        $exportAs: ProductApi
+        $supportedFrom: v2.4.0
```
theunrepentantgeek, line 888: Assuming we target getting this into the upcoming release, this needs to be `v2.5.0`

```suggestion
        $supportedFrom: v2.5.0
```

v2/azure-arm.yaml
```diff
@@ -883,6 +883,12 @@ objectModelConfiguration:
...
+      Service_Products_Policy:
+        $exportAs: ProductPolicy
+        $supportedFrom: v2.4.0
```
theunrepentantgeek, line 891:
```suggestion
        $supportedFrom: v2.5.0
```

v2/internal/controllers/crd_apimanagement_20220801_test.go
```diff
@@ -111,6 +111,18 @@ func Test_ApiManagement_20220801_CRUD(t *testing.T) {
...
+			Test: func(tc *testcommon.KubePerTestContext) {
+				APIM_Product_Api_CRUD(tc, &service)
+			},
+		},
+		testcommon.Subtest{
+			Name: "APIM Product Policy CRUD",
+			Test: func(tc *testcommon.KubePerTestContext) {
+				APIM_Product_Policy_CRUD(tc, &service)
+			},
+		},
```
theunrepentantgeek, line 125: Are these safe to run in parallel with the other tests? The degree of parallelism has me worrying about race conditions.

theunrepentantgeek (commented):
Not approving, because this lacks samples. Looking good though!

v2/internal/controllers/crd_apimanagement_20220801_test.go
ross-p-smith: These should be safe to create in parallel as they create different 'named' products to hang the sub-types from.

tomkerkhove: Correct. As long as the service is different we should be good to go

super-harsh:
PR looks good to me! The only thing missing is a test recording for samples. Here's the [guide to record samples test](https://azure.github.io/azure-service-operator/contributing/add-a-new-code-generated-resource/#run-test-for-added-sample-and-commit-the-recording).

v2/samples/apimanagement/v1api20220801/v1api20220801_productapi.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: apimanagement.azure.com/v1api20220801
+kind: ProductApi
+metadata:
+  name: sampleproductapi
+  namespace: default
+spec:
+  azureName: sampleapi
+  owner:
+    name: sampleapimproduct
```
super-harsh, line 9: We're excluding samples test for APIM/Product [here](https://github.com/Azure/azure-service-operator/blob/main/v2/internal/testcommon/samples_tester.go#L64), hence why the samples test would be failing. Since you've done a fix for deleteSubscription in #3467, should be alright to include the sample back in and record.

ross-p-smith: I think further investigation is needed as to why these samples don't run in the tests. Product has little use without ProductApi and ProductPolicy, but I would like to re-enable these tests when v2 of APIM is ready as this will cut the tests down from +1.5 hours to a few minutes.

For teh time being we are now not running Sample Tests for Product/Subscription/ProductPolicy/ProductApi.

However, if we get this PR, @ansulgoenka can uncomment these tests

What say you, @super-harsh ? :-)

super-harsh (approved):

tomkerkhove (approved):

tomkerkhove:
@super-harsh It looks like something is blocking auto-merge, what do we need to do?

Issue #1776, opened 2021-09-02
Feature: Handle properties that skip versions


theunrepentantgeek:
We need to design for the case where a particular property is removed in one version of a resource but reappears in a later version.

**TL;DR:** The crux of the problem is that we can currently end up with two different versions (shapes) of the same property serialised in a property bag on the same intermediate version.

To clarify, consider the following example. First we'll show the current behaviour, and then how that causes a problem. Finally, we'll discuss some constraints on any potential solution.

## Version 4

Consider a CRM system containing details of people. In v3 of the system, we capture each person's residental address, but in v4 we have dropped that property.

![image](https://user-images.githubusercontent.com/1272094/131927352-41115234-95d3-4f71-9adc-95b0d5cff244.png)

When we convert v3 -> v4, the `ResidentialAddress` property gets serialized into the `PropertyBag` on the v4 Person. Note that the bag contains a **v3** `Address` in a serialized form.

Conversion in the other direction (v4 -> v3) works fine, as we can deserialize the v3 `Address` from the `PropertyBag`.

## Version 5

In v5, the `ResidentialAddress` is reintroduced, but with a different shape. Instead of being the single field `Label`, it's now multiple fields.

![image](https://user-images.githubusercontent.com/1272094/131927399-d122a259-9c76-4737-aff2-d9d63e7945df.png)

When we convert v5 -> v4, again the `ResidentialAddress` property gets serialized into the `PropertyBag` on the v4 Person. Note that this time the bag contains a **v5** `Address` in a serialized form.

Conversion in the other direction (v4 -> v5) works fine, as we can deserialize the v5 `Address` from the property bag.

## The Problem

Observe how we can end up with two different kinds of v4 `Person`. We can end up with one where the property bag contains a **v3** `Address` or we can end up with one where the property bag contains a **v5** `Address`.

Here's where we run into problems.

![image](https://user-images.githubusercontent.com/1272094/131927447-30bf9eb5-493f-429b-8eab-e412173ec58d.png)

When we convert v3 -> v4 -> v5 we end up stuck part way through. 

After our first conversion step, we have a v4 `Person` that contains a v3 `Address` in the property bag. 

Conversion to the v5 `Person` will fail when we try to deserialize a **v5** `Address` from the **v3** version in the property bag. The two are not compatible, and the conversion will fail. This will cause the operator to fail.

### We can't see the future

[...]

PR #3555, opened 2023-11-14
Design to handle skipping properties

theunrepentantgeek:

We have a long-standing issue (see #1776) where our conversions between versions will break if a complex (object valued) property is reintroduced with a different shape in a later version after being removed in an earlier one.

Design of a solution has become a priority as this seems to be currently blocking an update of ManagedCluster to the latest preview and stable versions. 

gif

**If applicable**:


docs/hugo/content/design/ADR-2023-09-Skipping-Properties/_index.md
```diff
@@ -0,0 +1,174 @@
...
+[v3.Person]--[v3.Address|Label string]
+[v4.Person|FullName string;FamilyName string; KnownAs string|PropertyBag PropertyBag]..[v4.Address|Label string {bg:cornsilk}]
+[v5.Person|FullName string;FamilyName string; KnownAs string;ResidentialAddress|PropertyBag PropertyBag] 
+[v5.Person]--[v5.Address|Street string; Suburb string; City string|PropertyBag PropertyBag]
+-->
+
+The invariant we need to maintain here is that the **v5/compat** version of `Address` needs to have the same exact shape as the **v3** version of `Address`.
+
+Pros:
+* Addresses the central inconsistency problem, ensuring we always serialize the same shape into intermediate property bags
```
matthchr, line 150: and that shape is the old shape (but possibly with a propertybag of its own, to hold new properties that can't be converted to the old shape?)

docs/hugo/content/design/ADR-2023-09-Skipping-Properties/_index.md
```diff
@@ -0,0 +1,174 @@
...
+* Preserves the serialization format for existing resource versions (e.g. **v3** and **v4**)
+* Does not change the object model for released code
+* Isolates types introduced for compatibility purposes in a dedicated subpackage, avoiding the problem of circular dependencies.
```
matthchr, line 153: Can you expand on the mechanics of actually:
1. Autogenerating this package
2. handcrafting (I assume that would be required) the extensions to it such that it can convert to/from the `v3` shape?

You've already got the example laid about above - walk through exactly what that would look like with this proposal.

matthchr (commented):
I think needs some expanded details around how exactly the option 2 compat package would work.

docs/hugo/content/design/ADR-2023-09-Skipping-Properties/_index.md
```diff
@@ -0,0 +1,174 @@
...
+
+<!-- yuml.me
+[v3.Person|FullName string;FamilyName string; KnownAs string;ResidentialAddress]
+[v3.Person]--[v3.Address|Label string]
+[v4.Person|FullName string;FamilyName string; KnownAs string|PropertyBag PropertyBag]..[v4.Address|Label string {bg:cornsilk}]
+[v5.Person|FullName string;FamilyName string; KnownAs string;ResidentialAddress|PropertyBag PropertyBag] 
+[v5.Person]--[v5.Address|Street string; Suburb string; City string|PropertyBag PropertyBag]
+-->
+
+The invariant we need to maintain here is that the **v5/compat** version of `Address` needs to have the same exact shape as the **v3** version of `Address`.
```
super-harsh, line 147: Would be good to expand a bit more on how would option 2 work

super-harsh: Also, would be good to have an understanding of what is `compat` subpackage

docs/hugo/content/design/ADR-2023-09-Skipping-Properties/_index.md
```diff
@@ -0,0 +1,174 @@
...
+* When there are multi-generational skips, other questions arise: Do we introduce the new type into every intermediate version, only the last, or both first and last? Each approach has advantages and drawbacks.
+
+Fatal:
+* The **v4** package already has a reference to **v5**. The Go compiler's rules prohibiting circular package dependencies preclude adding a reference to **v4** from **v5**.
+
+### Option 2: Introduce Conversion Type
+
+As for Option 1, but instead of introducing the new type into **v4** where it can't be referenced by **v5**, use a dedicated **v5/compat** subpackage.
+
+{{< figure src="./option-2.png" >}}
```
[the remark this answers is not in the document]
theunrepentantgeek: I've improved all the diagrams, and I think I fixed this.

docs/hugo/content/design/ADR-2023-09-Skipping-Properties/_index.md
theunrepentantgeek: Yes.

docs/hugo/content/design/ADR-2023-09-Skipping-Properties/_index.md
theunrepentantgeek: Expanded, majorly.

docs/hugo/content/design/ADR-2023-09-Skipping-Properties/_index.md
theunrepentantgeek: The `compat` subpackage is literally a place to stash an extra type or two that's out of the way. Nothing more. 

I've expanded how this would work - though it's now Option 4.

docs/hugo/content/design/ADR-2023-09-Skipping-Properties/_index.md
```diff
@@ -0,0 +1,418 @@
...
+[v3.Person]->[v3.Address|Label multilineString|Properties PropertyBag]
+
+[v4.Person|FullName string;FamilyName string; KnownAs string|Properties PropertyBag]
+
+[v5.Person|FullName string;FamilyName string; KnownAs string;ResidentialAddress|Properties PropertyBag]
+[v5.Person]--[v5.Address|Street string; Suburb string; City string; Country string|Properties PropertyBag]
+
+-->
+
+{{< figure src="./conversion-v3-v4-v5-class.png" >}}
```
matthchr, line 145: I think in the class diagram, the some of the Address fields are missing a type specifier.

docs/hugo/content/design/ADR-2023-09-Skipping-Properties/_index.md
```diff
@@ -0,0 +1,418 @@
...
+[v3.Person]->[v3.Address|Label multilineString|Properties PropertyBag]
+
+[v4.Person|FullName string;FamilyName string; KnownAs string;ResidentialAddress v4.Address|Properties PropertyBag]
+[v4.Person]->[v4.Address|Label multilineString|Properties PropertyBag]
+
+[v5.Person|FullName string;FamilyName string; KnownAs string;ResidentialAddress v5.Address|Properties PropertyBag]
+[v5.Person]->[v5.Address|Street string; Suburb string; City string; Country string|Properties PropertyBag]
+-->
+
+{{< figure src="./option-1-class.png" >}}
```
matthchr, line 245: Some address fields in this diagram missing a type specifier (I think)

docs/hugo/content/design/ADR-2023-09-Skipping-Properties/_index.md
```diff
@@ -0,0 +1,418 @@
...
+Pros:
+* Works the way existing conversions do.
+* We have an existing contract for augmenting conversions with custom code, allowing us to handcode the conversion between **v4** and **v5** to handle the mismatch in the shapes of `Address`.
+* Addresses the central inconsistency problem, ensuring we always serialize the same shape into intermediate property bags
+
+Cons:
+* Changes the serialization format for existing **v4** resources. To avoid a breaking change, we'd need to code generate conversions for several edge cases, including:
+  * Conversion of a _new style_ **v4** `Person` to **v3** person, handling both the case where the address is in the property bag, and the case where it is a property reference.
+  * Conversion of a _new style_ **v4** `Person` to **v5** person, handling both the case where the address is in the property bag (requiring conversion from the **v3** shape), and the case where it is a property reference (requiring conversion from the **v4** shape).
+* Changes the object model of released resources, a breaking change for code consumers of the object model.
```
matthchr, line 258: Another con is that it seems to be an end-run around `PropertyBag`? If we took this approach, would we be using PropertyBag for anything?  I guess it would still be used for v5 adds a property v4 doesn't have, then when converting v5->v4 that property goes into the property bag?

But when going forward (v3 -> v4), if v4 removed a property that v3 had, then it would be resurrected as a property on v4 anyway?

I think it ends up adding more confusion to an already somewhat complex topic, because no longer is the answer "it's always in the propertybag" now it depends.

docs/hugo/content/design/ADR-2023-09-Skipping-Properties/_index.md
```diff
@@ -0,0 +1,418 @@
...
+
+[v4.Person|FullName string;FamilyName string; KnownAs string|Properties PropertyBag]
+[v4.Person]-.-[v4.Address {bg:cornsilk}|Label multilineString|Properties PropertyBag]
+
+[v5.Person|FullName string;FamilyName string; KnownAs string;ResidentialAddress v5.Address|Properties PropertyBag]
+[v5.Person]->[v5.Address|Street string; Suburb string; City string; Country string|Properties PropertyBag]
+
+-->
+
+{{< figure src="./option-2-class.png" >}}
```
matthchr, line 278: You inconsistently call the `v4` address `compat.Address` and `v4.Address` in the two option-2 diagrams.

I _think_ you mean to call it `v4.Address` in this (new) option 2?

docs/hugo/content/design/ADR-2023-09-Skipping-Properties/_index.md
```diff
@@ -0,0 +1,418 @@
...
+The process for conversion from **v5** to **v4** to **v3** is the reverse. We convert the **v5** `Address` into a **v4** `Address` before storing it in the property bag, and then rehydrate a **v4** `Address` from the property bag and convert that to a **v3** `Address`.
+
+While this might seem to just _kick the can down the road_, our conversion framework already has support for customization of the conversions between adjacent versions of a type, allowing the mismatch between **v4** and **v5** of `Address` to be handled by augmenting the conversion with a hand crafted conversion.
+
+Pros:
+* Addresses the central inconsistency problem, ensuring we always serialize the same shape into intermediate property bags
+* Preserves the serialization format for existing resource versions (e.g. **v3** and **v4**)
+
+Cons:
+* Changes the object model of released resources, a breaking change for consumers of the object model
```
matthchr, line 295: Does it actually do this? It seems to just add _types_ to existing packages, which is OK?

Those types aren't actually referenced anywhere explicitly though, because they're only ever put into the property bag?

I think the issue is that we have existing `v4` types that have a `v3` thing inside of their Property Bag? But if v3 and v4 always have the same structure then that maybe doesn't matter either?

I am not following why this reverses our current direction. Wouldn't we implement conversion for this new type exactly the same way we implemented it for all the other types? It's just it wouldn't be referred to anywhere as a property? For example why couldn't we continue having the conversion code to convert to the `v4.Address` on the `v3.Address`? (and same for `v5` from `v4`)?

docs/hugo/content/design/ADR-2023-09-Skipping-Properties/_index.md
```diff
@@ -0,0 +1,418 @@
...
+
+-->
+
+{{< figure src="./option-4-class.png" >}}
+
+Conversion from **v4** to **v5** would work as follows:
+
+* Check the property bag for a `ResidentialAddress` property.
+* If found, deserialize it as a **v4/compat** `Address`.
+* Recursively invoke a conversion method to convert the **v4/compat** `Address` to a **v5** `Address` (this mechanism already exists in our property conversion framework).
```
matthchr, line 356: As I alluded to before, once the `v4.Address` (or `v4/compat.Address`) type exists the conversion logic for going to/from `v5` would live on that type, in the same way that `v4` references `v5` for doing conversion today, so I don't understand the need for the subpackage.

Am I correct that the only difference between Option 2 and Option 4 is the subpackage?

docs/hugo/content/design/ADR-2023-09-Skipping-Properties/_index.md
```diff
@@ -0,0 +1,418 @@
...
+
+
+The code for this would look like:
```
matthchr, line 359: Which package is this code in? In `v4` right?

matthchr: Or maybe more specifically `v4storage`

docs/hugo/content/design/ADR-2023-09-Skipping-Properties/_index.md
```diff
@@ -0,0 +1,418 @@
...
+Pros:
+* Addresses the central inconsistency problem, ensuring we always serialize the same shape into intermediate property bags
+* Conversions are still driven by the earlier version, avoiding any issues with circular dependencies or needing to move conversion code to the later version.
+* Isolates the compatibility type in a dedicated subpackage, where it won't inadvertently be picked up by documentation tools or other code consumers.
+* Allows a hand driven conversion to be written to convert between **v4** and **v5** using the augmentation mechanism already in place for conversions between adjacent versions.
+* Conversions between **v3** and **v4** remain unchanged
+* Serialization format of **v4** remains unchanged, ensuring compatibility with existing releases of ASO.
+
+Cons:
+* Requires modification of the conversion-graph, which to this point has been immutable once constructed (we can't delay construction until after we inject the compatibility package as we need it to identify skipping properties).
```
matthchr, line 395: Not sure I understand what this con is saying

matthchr (approved):

PR #3556, opened 2023-11-14
Allow AsDeclarations to return an error

### LINKED ISSUE #2969: Type.AsDeclarations should allow for returning error

The method `AsDeclarations()` on the interface `astmodel.Type`does not allow for implementations to return an error.

**Describe the improvement**

**Prerequisites**

- [x] #2966 
- [x] #2967 
- [ ] #2968 
- [x] #2971 

**Additional context**

theunrepentantgeek:

Changes `Type.AsDeclarations()` to allow for an error return, and removes use of **panic** for when things go predictably awry.

Part of a sequence of changes designed to remove the use of **panic** for errors that are expected because they can be triggered by external factors.

Closes #2969 

gif

### DIFF
--- a/v2/tools/generator/internal/astmodel/allof_type.go
+++ b/v2/tools/generator/internal/astmodel/allof_type.go
@@ -119,7 +119,7 @@ func (allOf *AllOfType) AsDeclarations(
 	codeGenerationContext *CodeGenerationContext,
 	declContext DeclarationContext,
 ) ([]dst.Decl, error) {
-	return nil, errors.New(allOfFailureMsg)
+	panic(errors.New(allOfFailureMsg))
 }
 
 // AsZero always panics; AllOf cannot be represented by the Go AST and must be
--- a/v2/tools/generator/internal/astmodel/allof_type.go
+++ b/v2/tools/generator/internal/astmodel/allof_type.go
@@ -105,30 +105,33 @@ func (allOf *AllOfType) References() TypeNameSet {
 	return result
 }
 
-var allOfPanicMsg = "AllOfType should have been replaced by generation time by 'convertAllOfAndOneOf' phase"
+var allOfFailureMsg = "AllOfType should have been replaced by generation time by 'convertAllOfAndOneOf' phase"
 
 // AsType always panics; AllOf cannot be represented by the Go AST and must be
 // lowered to an object type
 func (allOf *AllOfType) AsType(_ *CodeGenerationContext) dst.Expr {
-	panic(errors.New(allOfPanicMsg))
+	panic(errors.New(allOfFailureMsg))
 }
 
 // AsDeclarations always panics; AllOf cannot be represented by the Go AST and must be
 // lowered to an object type
-func (allOf *AllOfType) AsDeclarations(_ *CodeGenerationContext, _ DeclarationContext) []dst.Decl {
-	panic(errors.New(allOfPanicMsg))
+func (allOf *AllOfType) AsDeclarations(
+	codeGenerationContext *CodeGenerationContext,
+	declContext DeclarationContext,
+) ([]dst.Decl, error) {
+	return nil, errors.New(allOfFailureMsg)
 }
 
 // AsZero always panics; AllOf cannot be represented by the Go AST and must be
 // lowered to an object type
 func (allOf *AllOfType) AsZero(_ TypeDefinitionSet, _ *CodeGenerationContext) dst.Expr {
-	panic(errors.New(allOfPanicMsg))
+	panic(errors.New(allOfFailureMsg))
 }
 
 // RequiredPackageReferences always panics; AllOf cannot be represented by the Go AST and must be
 // lowered to an object type
 func (allOf *AllOfType) RequiredPackageReferences() *PackageReferenceSet {
-	panic(errors.New(allOfPanicMsg))
+	panic(errors.New(allOfFailureMsg))
 }
 
 // Equals returns true if the other Type is a AllOf that contains
--- a/v2/tools/generator/internal/astmodel/allof_type_test.go
+++ b/v2/tools/generator/internal/astmodel/allof_type_test.go
@@ -103,6 +103,16 @@ func TestAllOfAsTypePanics(t *testing.T) {
 	}).To(PanicWith(MatchError(expectedAllOfPanic)))
 }
 
+func TestAllOfAsDeclarationsPanics(t *testing.T) {
+	t.Parallel()
+	g := NewGomegaWithT(t)
+
+	x := AllOfType{}
+	g.Expect(func() {
+		x.AsDeclarations(&CodeGenerationContext{}, DeclarationContext{})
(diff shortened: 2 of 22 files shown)


v2/tools/generator/internal/astmodel/allof_type.go
```diff
@@ -105,30 +105,33 @@ func (allOf *AllOfType) References() TypeNameSet {
...
 
 // AsDeclarations always panics; AllOf cannot be represented by the Go AST and must be
 // lowered to an object type
-func (allOf *AllOfType) AsDeclarations(_ *CodeGenerationContext, _ DeclarationContext) []dst.Decl {
-	panic(errors.New(allOfPanicMsg))
+func (allOf *AllOfType) AsDeclarations(
+	codeGenerationContext *CodeGenerationContext,
+	declContext DeclarationContext,
+) ([]dst.Decl, error) {
+	return nil, errors.New(allOfFailureMsg)
```
matthchr, line 122: Is this actually _expected_?

I don't think it is? Or maybe put another way, there's nothing in type configuration or otherwise that should cause this, right?

Should this stay a panic (even though it _could_ be an error now)?

v2/tools/generator/internal/astmodel/allof_type_test.go
```diff
@@ -103,16 +103,6 @@ func TestAllOfAsTypePanics(t *testing.T) {
 	}).To(PanicWith(MatchError(expectedAllOfPanic)))
 }
 
-func TestAllOfAsDeclarationsPanics(t *testing.T) {
```
matthchr, line 106: minor: Either keep (if my above comment is correct and this should still panic), or transform into a test that ensures that it returns the right message when it errors?

v2/tools/generator/internal/astmodel/oneof_type.go
```diff
@@ -149,29 +149,32 @@ func (oneOf *OneOfType) References() TypeNameSet {
...
 
 // AsDeclarations always panics; OneOf cannot be represented by the Go AST and must be
 // lowered to an object type
-func (oneOf *OneOfType) AsDeclarations(_ *CodeGenerationContext, _ DeclarationContext) []dst.Decl {
-	panic(errors.New(oneOfPanicMsg))
+func (oneOf *OneOfType) AsDeclarations(
+	codeGenerationContext *CodeGenerationContext,
+	declContext DeclarationContext,
+) ([]dst.Decl, error) {
+	return nil, errors.New(oneOFailureMsg)
```
matthchr, line 166: Same q as allof

v2/tools/generator/internal/astmodel/oneof_type_test.go
```diff
@@ -56,16 +56,6 @@ func TestOneOfAsTypePanics(t *testing.T) {
 	}).To(PanicWith(MatchError(expectedOneOfPanic)))
 }
 
-func TestOneOfAsDeclarationsPanics(t *testing.T) {
```
matthchr, line 59: Same comment as AllOf

v2/tools/generator/internal/conversions/property_bag_member_type.go
```diff
@@ -65,8 +66,11 @@ func (b *PropertyBagMemberType) AsType(ctx *astmodel.CodeGenerationContext) dst.
 }
 
 // AsDeclarations panics because this is a metatype that will never be rendered
-func (b *PropertyBagMemberType) AsDeclarations(_ *astmodel.CodeGenerationContext, _ astmodel.DeclarationContext) []dst.Decl {
-	panic("should never try to render a PropertyBagMemberType as declarations")
+func (b *PropertyBagMemberType) AsDeclarations(
+	codeGenerationContext *astmodel.CodeGenerationContext,
+	declContext astmodel.DeclarationContext,
+) ([]dst.Decl, error) {
+	return nil, errors.New("should never try to render a PropertyBagMemberType into declarations")
```
matthchr, line 73: Similar Q as AllOf/OneOf.

v2/tools/generator/internal/astmodel/allof_type.go
theunrepentantgeek: I've been applying the rubric to use errors for things that are most likely to be triggered by external factors, and panic for things most likely to be triggered by developer errors.

At this point, I'm fairly confident that our code is working properly. 

I figure the most likely case where things go wrong with an AllOf or OneOf type is if a Product Group does something new with their OpenAPI spec definitions that we haven't seen before.

matthchr: > At this point, I'm fairly confident that our code is working properly.

But, wouldn't it be a developer error if all `AllOf` types weren't removed by the time we get to the declarations stage of the pipeline?

I thought the contract we had was that `OneOf` and `AllOf` had to be turned into `ObjectType` before the end of the pipeline. I don't think it's possible for _any_ configuration to cause them to stick around? Wouldn't that be a developer error?

theunrepentantgeek: I'm not thinking about ASO configuration - I'm thinking about what happens when someone uses an AllOf in a new way that we haven't seen before. That's an external factor, where I don't think we should panic - because external data shouldn't be trusted.

theunrepentantgeek: That said, I'm not fervent on this, happy to change it back.

matthchr: > I'm not thinking about ASO configuration - I'm thinking about what happens when someone uses an AllOf in a new way that we haven't seen before.

You mean in our code or in the Swagger? If in the Swagger, I'm 90% sure that the problem won't come here, it will come from the `ConvertAllOfAndOneOfToObjects` stage (which does not call `AsDeclarations`). If you mean in our code, that's developer error right?

I've approved as I am also not fervent on this, but I do tend towards leaving this as a panic as I'm still not understanding what non-developer configuration could possibly cause this code to be hit. My assertion is that if this code is ever hit we have a code bug in the `ConvertAllOfAndOneOfToObjects` that somehow missed converting AllOf or OneOf types, which I don't think could happen unless somehow the visitor failed to visit them.

Also happy to chat and understand how I'm wrong about the above.

matthchr (approved):

v2/tools/generator/internal/astmodel/allof_type.go
theunrepentantgeek: Changed back to panic.

v2/tools/generator/internal/astmodel/oneof_type.go
theunrepentantgeek: Reverted to panic.

v2/tools/generator/internal/conversions/property_bag_member_type.go
theunrepentantgeek: Restored the panic.

PR #3567, opened 2023-11-16
Improve test summary tool

### LINKED ISSUE #3495: Bug: Test summary tool sometimes omits results
**Describe the bug**

Discovered during CI of PR #3488, it appears the tool we wrote to sumarize the output of `go test` sometimes doesn't include all the expected results. 

In that case, the test `Test_TwoObjectsWriteSameSecret_WarningConditionIsSetOnSecond` had a **panic** but was not listed as failed in the summary document.

The output from that test includes:

```
=== CONT  Test_TwoObjectsWriteSameSecret_WarningConditionIsSetOnSecond\n"}
panic: no logger registered for aso-test-configmapindifferentnamespace-configmapnotfound: asotest-assignment-zlbfgy [recovered]
    panic: no logger registered for aso-test-configmapindifferentnamespace-configmapnotfound: asotest-assignment-zlbfgy
```

**To Reproduce**

I think all we need for a repro is a test that throws a panic.

**Expected behavior**

All tests should be listed in the output, even those that panic.

theunrepentantgeek:

The `mangle-test-json` tool that we use to generate summaries of test results has a bug - it only lists the outcome for tests that have an explicit outcome listed in the test output.

Other tests were omitted, including:
* Tests that panic
* Tests that were paused at the time of test suite termination
* Tests that were still running at the time of test suite termination

Of these, the first (omission of tests that panic) was crucial because we had test failures where the identity of the failing test was suppressed.

Instead of tracking information about test execution and only creating a `TestRun` summary once we know the outcome of the test, we instead need to start tracking each test up front, pruning those that aren't _interesting_ once we're finished parsing.

Closes #3495

The use of three parallel maps, all tracking information for specific tests, was a significant code smell that prompted the non-trivial refactoring. 

Now we have a single map tracking all test runs.

The `TestRun` abstraction already existed - I've just relocated it into a dedicated file and given it functionality that was originally elsewhere. The main application then becomes a loop creating and updating these TestRuns.

gif

### DIFF
--- a/v2/tools/mangle-test-json/main.go
+++ b/v2/tools/mangle-test-json/main.go
@@ -25,14 +25,6 @@ type JSONFormat struct {
 	Output  string    `json:"Output"`
 }
 
-type TestRun struct {
-	Action  string
-	Package string
-	Test    string
-	Output  []string
-	RunTime time.Duration
-}
-
 func main() {
 	log := CreateLogger()
 
--- a/v2/tools/mangle-test-json/main.go
+++ b/v2/tools/mangle-test-json/main.go
@@ -28,6 +28,11 @@ type JSONFormat struct {
 func main() {
 	log := CreateLogger()
 
+	if len(os.Args) <= 1 {
+		log.Info("No log file specified on command line.")
+		return
+	}
+
 	for _, testOutputFile := range os.Args[1:] {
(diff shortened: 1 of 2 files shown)


super-harsh (approved):
Looks clean!

PR #3593, opened 2023-11-23
Add latest version of PostgreSQL

### LINKED ISSUE #3294: Import stable versions of preview resources

We have a small number of resources where the latest available API version we support is a preview version despite there being multiple later stable releases. 

Some users are hesitant to use preview versions and would prefer to use stable API versions only.

**Describe the improvement**

Check each resource where the latest version we support is a preview and identify stable versions we should have on our backlog.

Groups to update

- [x] Authorization
- [x] ContainerService - @theunrepentantgeek 
- [x] DBforPostgreSQL
- [x] Insights - @super-harsh 
- [x] KeyVault
- [x] ManagedIdentity
- [x] ServiceBus
- [x] Storage - @matthchr

theunrepentantgeek:

Add support for the 2022-12-01 API version of PostgreSQL.

One of the updates tracked in #3294

gif

**If applicable**:
- [x] this PR contains tests


super-harsh (approved):
LGTM! 
Do you think its worth making the controller test for older version to use the latest version we support(since not adding controller test)?

PR #3607, opened 2023-11-28
Dependabot updates

super-harsh:

This PR contains all ./v2 dependabot updates.


theunrepentantgeek (approved):
Approved, but we mostly don't need to create custom branches for these now. Usually we can just push a tweak to the dependabot branch and let it take care of the rest of the details. Most commonly, the required tweak is a `go mod tidy` for either `asoctl` or `generator`.

PR #3608, opened 2023-11-28
Add owner-name label on resources

### LINKED ISSUE #3190: Support finding child resources easier with automatic label additions?
AFAIK Kubernetes doesn't support automatically finding owned resources, see
https://github.com/kubernetes/kubernetes/issues/54498
https://github.com/kubernetes/client-go/issues/1085

As a workaround, we could possibly automatically add labels to child resources so that it's easier to query them from the parents.

super-harsh:
Closes #3190 

This PR enables controller to add `serviceoperator.azure.com/owner-name: <OwnerName>` label at the time of adding the initial resource state

### DIFF
--- a/v2/pkg/common/annotations/annotations.go
+++ b/v2/pkg/common/annotations/annotations.go
@@ -3,15 +3,4 @@
 
 package annotations
 
-import "github.com/Azure/azure-service-operator/v2/pkg/genruntime"
-
-const (
-	PerResourceSecret = "serviceoperator.azure.com/credential-from"
-	OwnedByAnnotation = "serviceoperator.azure.com/owned-by"
-)
-
-func SetOwnedByAnnotation(obj genruntime.ARMMetaObject) {
-	if obj.Owner() != nil && obj.Owner().Name != "" {
-		genruntime.AddAnnotation(obj, OwnedByAnnotation, obj.Owner().Name)
-	}
-}
+const PerResourceSecret = "serviceoperator.azure.com/credential-from"
--- a/v2/pkg/common/annotations/annotations.go
+++ b/v2/pkg/common/annotations/annotations.go
@@ -3,4 +3,15 @@
 
 package annotations
 
-const PerResourceSecret = "serviceoperator.azure.com/credential-from"
+import "github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+
+const (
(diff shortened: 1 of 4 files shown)


theunrepentantgeek (approved):

v2/pkg/common/annotations/annotations.go
```diff
@@ -3,4 +3,15 @@
 
 package annotations
 
-const PerResourceSecret = "serviceoperator.azure.com/credential-from"
+import "github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+
+const (
+	PerResourceSecret = "serviceoperator.azure.com/credential-from"
+	OwnedByAnnotation = "serviceoperator.azure.com/owned-by"
```
matthchr, line 10: I think `owner-name` or `owner` would be clearer than `owned-by`.

Do we also need to think about `owner-group-kind` or something as well, for extension resources such as roleassignment?

matthchr (changes requested):
Users can't filter w/ annotations I don't think, only with labels?

So I think this needs to be a label not an annotation.

v2/pkg/common/annotations/annotations.go
theunrepentantgeek: Not having `owner-group-kind` forces the user to do another query to discover the owner type if they don't already know it. I think there's value in making it easy/obvious.

v2/pkg/genruntime/base_types.go
```diff
@@ -84,21 +84,40 @@ type ARMOwnedMetaObject interface {
...
+// AddLabel adds the specified label to the object.
+// Empty string labels are not allowed. Attempting to add a label with a value
+// of empty string will result in the removal of that label.
+func AddLabel(obj MetaObject, k string, v string) {
+	labels := obj.GetLabels()
+	labels = AddToMap(labels, k, v)
+	obj.SetLabels(labels)
+}
+
+func AddToMap(labels map[string]string, k string, v string) map[string]string {
```
matthchr, line 105: minor: Consider not exporting this, and changing the parameter name from `labels` or `annotations` to something generic like `m`?

v2/pkg/common/annotations/annotations.go
super-harsh: Sure! and that applies to only resources which are KubernetesReferences?

matthchr (approved):

PR #3614, opened 2023-11-29
Implement repair for skipping properties

### LINKED ISSUE #1776: Feature: Handle properties that skip versions

theunrepentantgeek:

Convert the pipeline stage `DetectSkippingProperties` into `RepairSkippingProperties` - instead of identifying the problem, it now repairs the object model by cloning relevant type definitions into a `compat` subpackage.

Modify code generation for PropertyBags to look for, and use, the `compat` subpackage objects if present.

Also
* Use factory method NewState() with definitions instead of immediately calling WithDefinitions()
* Add helper method State.WithOverlaidDefinitions() to simplify updating the state after each pipeline stage.

Implements our [agreed design](https://azure.github.io/azure-service-operator/design/adr-2023-09-skipping-properties/),

Closes #1776

gif

**If applicable**:
- [x] this PR contains tests


v2/tools/generator/internal/codegen/pipeline/repair_skipping_properties.go
```diff
@@ -184,46 +201,87 @@ func (detector *skippingPropertyDetector) findChains() *astmodel.PropertyReferen
...
+	def := detector.definitions[tn]
+	defs, err := astmodel.FindConnectedDefinitions(detector.definitions, astmodel.MakeTypeDefinitionSetFromDefinitions(def))
+	if err != nil {
+		return nil, errors.Wrapf(
+			err,
+			"failed to find connected definitions from %s",
+			tn)
+	}
+
+	compatPkg := astmodel.MakeCompatPackageReference(lastMissing.DeclaringType().InternalPackageReference())
```
super-harsh, line 267: A few comments here would be helpful

v2/tools/generator/internal/astmodel/compat_package_reference.go
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package astmodel
+
+func MakeCompatPackageReference(pkg InternalPackageReference) SubPackageReference {
```
super-harsh, line 8: Function documentation here would be helpful here for future ref - on why and where is this used

v2/tools/generator/internal/codegen/pipeline/repair_skipping_properties.go
```diff
@@ -43,47 +43,49 @@ const DetectSkippingPropertiesStageID = "detectSkippingProperties"
...
 			}
 
-			err := detector.CheckForSkippedProperties()
+			defs, err := repairer.RepairSkippedProperties()
+			if err != nil {
+				return nil, err
+			}
 
-			return state, err
+			return state.WithOverlaidDefinitions(defs), err
```
super-harsh, line 69: Do we need pre/post requisite check for `CreateConversionGraph` stage here?

super-harsh (approved):
LGTM! Just a few minor comments

v2/tools/generator/internal/codegen/pipeline/repair_skipping_properties.go
```diff
@@ -184,46 +201,87 @@ func (detector *skippingPropertyDetector) findChains() *astmodel.PropertyReferen
...
+	if !ok {
+		// Should never fail, given the way findBreak() works
+		panic(fmt.Sprintf("failed to find type for property %s", lastObserved))
+	}
+
+	tn, ok := astmodel.AsInternalTypeName(lastObservedPropertyType)
+	if !ok {
+		// If not a type name, defer to our existing property conversion logic
+		// Continue checking the rest of the chain
+		return detector.repairChain(reintroduced)
```
matthchr, line 255: Won't this call the logic even though `!typesSame`?

Is that desired? I assume that this could/would happen in cases where the removed property was say, `*int`, and then it was added back as `*float`.

Do we already deal with primitive types here?

v2/tools/generator/internal/codegen/pipeline/repair_skipping_properties_test.go
```diff
@@ -0,0 +1,308 @@
...
+	// Pkg2021storage.Person.FullName => Person2022storage.Person.FullName
+	// Pkg2022storage.Person.FullName => EmptyReference
+	g.Expect(repairer.links).To(HaveLen(4))
+	AssertLinkExists(g, repairer, person2020, test.FullNameProperty.PropertyName())
+	AssertLinkExists(g, repairer, person2020s, test.FullNameProperty.PropertyName())
+	AssertLinkExists(g, repairer, person2021s, test.FullNameProperty.PropertyName())
+	AssertLinkExists(g, repairer, person2022s, test.FullNameProperty.PropertyName())
+}
+
+func TestSkippingPropertyRepairer_findBreak_returnsExpectedResults(t *testing.T) {
```
matthchr, line 62:
```suggestion
func TestSkippingPropertyRepairer_FindBreak_ReturnsExpectedResults(t *testing.T) {
```

I think?

v2/tools/generator/internal/codegen/pipeline/repair_skipping_properties_test.go
```diff
@@ -0,0 +1,308 @@
...
+	g.Expect(err).To(BeNil())
+	g.Expect(finalState.definitions).To(HaveLen(11)) // Our five original resources, plus the storage variants, plus one more
+
+	expected := astmodel.MakeInternalTypeName(
+		astmodel.MakeCompatPackageReference(test.Pkg2021s),
+		"Residence")
+	g.Expect(finalState.definitions).To(HaveKey(expected))
+}
+
+// Test_RepairSkippingProperties checks that the pipeline stage correctly injects new types with the right functions
```
matthchr, line 243: Name doesn't match test name?

v2/tools/generator/internal/codegen/pipeline/repair_skipping_properties_test.go
```diff
@@ -0,0 +1,308 @@
...
+// when properties of different types skip versions.
+func TestGolden_RepairSkippingProperties_WhenPropertyTypesDiffer_GeneratesExpectedCode(t *testing.T) {
```
matthchr, line 245: minor: Wondering if it would be clearer if we also had a `repair_skipping_properties_golden_test.go` and put this there, since this file is otherwise pure unit tests?

Don't feel super strongly about this, was just initially confused that you had golden files at all as I didn't see this test when I scanned this file initially and I was wondering where the golden files came from. Only on re-read did I realize that there was one golden test at the end of the file.

Obviously minor.

v2/tools/generator/internal/codegen/pipeline/repair_skipping_properties_test.go
```diff
@@ -0,0 +1,308 @@
...
+		CreateConversionGraph(cfg, "v"), // Then, RECREATE the conversion graph showing relationships
+		InjectPropertyAssignmentFunctions(cfg, idFactory, logr.Discard()),
+	)
+	g.Expect(err).To(BeNil())
+
+	// Assert - check that we get the expected code changes
+	test.AssertPackagesGenerateExpectedCode(
+		t,
+		finalState.Definitions(),
+		test.DiffWith(comparisonState.Definitions().AsSlice()...))
```
matthchr, line 293: Do we actually want `DiffWith` here rather than just looking at the full set of generated code?

Or maybe we do and what I am looking for is a comment or something on this test indicating which bits we should check to ensure that the code looks right?

I looked through the code and can see that conversion for residence is now using the compat package, which makes sense, but it wasn't immediately obvious to me if the other code around it was correct or not.

matthchr (approved):
Had some minor comments but approving to unblock you from merging if you fix those comments on your Monday.

v2/tools/generator/internal/codegen/pipeline/repair_skipping_properties.go
theunrepentantgeek: Done.

v2/tools/generator/internal/codegen/pipeline/repair_skipping_properties.go
theunrepentantgeek: Our checks can't handle requiring the same stage both before and after; I've added a prerequisite check.

v2/tools/generator/internal/codegen/pipeline/add_api_version_enums.go
```diff
@@ -30,12 +30,11 @@ func AddAPIVersionEnums() *Stage {
...
 					version := apiVersions.Get(name.InternalPackageReference())
 					def = def.WithType(rt.WithAPIVersion(version.name, version.value))
+					newDefs.Add(def)
 				}
-
-				newDefs.Add(def)
 			}
 
-			return state.WithDefinitions(newDefs), nil
+			return state.WithOverlaidDefinitions(newDefs), nil
```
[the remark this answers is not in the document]
theunrepentantgeek: Mostly that it's more declarative - we create _these_ new types and add them to the ones we already have. With the old structure you have to read the loop to work out that all the existing types are retained.

v2/tools/generator/internal/codegen/pipeline/add_operator_spec.go
```diff
@@ -55,9 +55,8 @@ func AddOperatorSpec(configuration *config.Configuration, idFactory astmodel.Ide
 				return nil, err
 			}
 
-			result = defs.OverlayWith(result)
-
-			return state.WithDefinitions(result).WithGeneratedConfigMaps(exportedTypeNameConfigMaps), nil
+			return state.WithOverlaidDefinitions(result).
```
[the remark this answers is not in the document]
theunrepentantgeek: I generally prefer chained method calls on the same type to be on different lines, but it's minor. Reverted.

PR #3634, opened 2023-12-06
Augment conversions for UpgradeOverrideSettings

### LINKED ISSUE #3632: Conversion augmentation interfaces not generated for `compat` package
**Describe the bug**

We now inject a `compat` subpackage in selected cases, to address the _skipping property_ problem. 

But, we're not generating the interfaces that allow augmentation of the conversions, exactly when those conversions are likely to be most important.

**Expected behavior**

Types in the `compat` package should get the same interfaces to allow for augmentation of the conversions.

theunrepentantgeek:

Fixes the generator bug where we failed to generate interfaces allowing hand-coded augmentation of inter-version object conversions, and implements a conversion for `ManagedCluster`.

Closes #3632

The fix to generate the augment interfaces was to remedy a bug in `IsStorageReference` which should have been recursive, but was not.

Also tweaks the way we select root objects for writing generated go files so that we don't end up with every type in a separate file in `compat` packages.

gif

**If applicable**:
- [x] this PR contains tests


v2/api/containerservice/v1api20230201/storage/compat/istio_components_types_gen.go
```diff
@@ -1,105 +0,0 @@
-// Code generated by azure-service-operator-codegen. DO NOT EDIT.
-// Copyright (c) Microsoft Corporation.
-// Licensed under the MIT license.
-package compat
-
-import (
-	v20231001s "github.com/Azure/azure-service-operator/v2/api/containerservice/v1api20231001/storage"
```
matthchr, line 7: Any idea why these files were deleted?

v2/api/containerservice/v1api20230201/storage/compat/cluster_upgrade_settings_types.go
```diff
@@ -0,0 +1,38 @@
...
+		settings.ControlPlaneOverrides = append(settings.ControlPlaneOverrides, ignoreKubernetesDeprecations)
+	}
+
+	return nil
+}
+
+func (settings *UpgradeOverrideSettings) AssignPropertiesTo(dest *v20231001s.UpgradeOverrideSettings) error {
+	// If the preview version has "IgnoreKubernetesDeprecations" in ControlPlaneOverrides, the GA version needs to have
+	// ForceUpgrade true, otherwise false.
+	dest.ForceUpgrade = to.Ptr(false)
```
matthchr, line 29: Does this work for round-trip cases? Do we need to deal w/ `nil`? So something like:
- If nil, then `ForceUpgrade` nil.
- If empty, false
- If populated w/ IgnoreKubernetesDeprecations, true

Not sure it's needed just checking.

matthchr (approved):

v2/api/containerservice/v1api20230201/storage/compat/istio_components_types_gen.go
theunrepentantgeek: Allocating types to files is based on root types - and we were only treating Resources + ARM Status + ARM Spec as available roots, so every type in the `compat` packge ended up in a separate file. My tweaks in `reference_graph.go` add "otherwised unreferenced types" as potential roots, consolidating things in to fewer files.

v2/api/containerservice/v1api20230201/storage/compat/cluster_upgrade_settings_types.go
theunrepentantgeek: And I wrote unit tests to check all the nil cases.

PR #3641, opened 2023-12-12
Add support for discontinuous property conversions 

theunrepentantgeek:

We've had the assumption that two properties needing conversion would have line-of-sight to each other in the storage conversion graph constructed by the generator. To this point this has been true, but with a recent attempt to import another preview version of `ManagedCluster` we discovered that this was not so.

To support these conversions, I've added another conversion handler, one that looks for a commonly visible _pivot_ type that can be used to build the conversion required.

I've also refactored creation of test cases for property conversions - the existing test case factory method was too large and complicated to comprehend; now the various groups of test cases are split up.

gif

**If applicable**:
- [x] this PR contains tests


v2/tools/generator/internal/conversions/property_conversions.go
```diff
@@ -1934,17 +1948,34 @@ func assignObjectsViaIntermediateObject(
...
-	intermediateName, err := conversionContext.FindNextType(earlierName)
-	if err != nil {
-		return nil, errors.Wrapf(
-			err,
-			"looking up next type for %s",
-			astmodel.DebugDescription(destinationEndpoint.Type()))
+	// Require a path from one name to the next, and work out an intermediate step to break down the conversion
+	var intermediateName astmodel.InternalTypeName
+	if conversionContext.PathExists(sourceName, destinationName) {
+		var err error
```
super-harsh, line 1954: Can move this(and below) var outside the `if` statements and to the error checking after the `if-else` blocks?

v2/tools/generator/internal/conversions/property_conversions.go
```diff
@@ -1934,17 +1948,34 @@ func assignObjectsViaIntermediateObject(
...
+				astmodel.DebugDescription(destinationEndpoint.Type()))
+		}
+	} else {
+		// No path between the two types, we can't handle the required conversion
+		return nil, nil
 	}
 
-	if intermediateName.IsEmpty() || astmodel.TypeEquals(intermediateName, sourceName) {
+	if intermediateName.IsEmpty() ||
+		astmodel.TypeEquals(intermediateName, sourceName) ||
```
matthchr, line 1977: minor: Should there be a comment here explaining the situations where the intermediate name is the source or destination?

v2/tools/generator/internal/conversions/property_conversions.go
```diff
@@ -2021,6 +2052,163 @@ func assignObjectsViaIntermediateObject(
...
+// var <otherlocal> <destinationType>
+// err := <pviot>.AssignPropertiesTo(<otherLocakl>)
+//
+//	if err != nil {
+//	   return errors.Wrap(err, "while calling <otherlocal>.AssignPropertiesFrom(<local>)")
+//	}
+//
+// Note the actual steps are generated by nested conversions; this handler works by finding the two conversions needed
+// given our intermediate type and chaining them together.
+func assignNoninlineObjectsViaPivotObject(
```
matthchr, line 2082: minor, should this be:
```suggestion
func assignNonInlineObjectsViaPivotObject(
```

Since I think `non-inline` would be cased like that?

matthchr (approved):

v2/tools/generator/internal/codegen/storage/conversion_graph.go
```diff
@@ -134,6 +134,33 @@ func (graph *ConversionGraph) FindHubAndDistance(
...
+) (astmodel.InternalTypeName, bool) {
+	current := start
+	for {
+		if predicate(current) {
+			return current, true
+		}
+
+		next, err := graph.FindNextType(current, nil)
+		if err != nil {
+			break
```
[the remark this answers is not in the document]
theunrepentantgeek: If there's an actual problem (other than _no next type_) that'll be reported elsewhere. So I think this one is safe to ignore.

v2/tools/generator/internal/codegen/storage/conversion_graph_test.go
```diff
@@ -385,3 +385,65 @@ func TestGolden_ConversionGraph_WhenCompatPackagePresent_HasExpectedTransitions(
...
+			func(t *testing.T) {
+				t.Parallel()
+				gg := NewGomegaWithT(t)
+
+				count := 0
+				_, found := graph.FindInPath(
+					c.start,
+					func(t astmodel.InternalTypeName) bool {
+						count++
+						return t == c.end
```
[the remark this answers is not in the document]
theunrepentantgeek: I don't think we need to - we're comparing `InternalTypeName` instances, which are value types, not reference - and we don't need any of the special support from `.Equals`.

v2/tools/generator/internal/conversions/property_conversions.go
theunrepentantgeek: We could, but I prefer keeping things tightly scoped. 

We need `intermediateName` after the **if**, but we don't need either of the potential errors later on. Moving the `err` variable outside implies a need to access the error again, which isn't correct.

v2/tools/generator/internal/conversions/property_conversions.go
theunrepentantgeek: Added.

v2/tools/generator/internal/conversions/property_conversions.go
theunrepentantgeek: Changed.

PR #3650, opened 2023-12-13
Add config for User-Agent suffix

### LINKED ISSUE #1827: Allow customization of the user agent
Provide the ability to add segments to the user-agent via command line args in the controller.

Originally suggested by @devigned in #1469

nojnhuh:
Closes #1827

**What this PR does / why we need it**: This PR adds an `AZURE_USER_AGENT_SUFFIX` configuration option to allow appending to the default User-Agent used for Azure HTTP clients.

Getting this wired up felt a little clunky, so any feedback on how to smooth that out is welcome.

I also wasn't sure exactly how best to add tests for this config, so any advice there would be appreciated. And tips on exactly where to look with ASO fully up and running to see if the user agent is really being set as expected.

gif

**If applicable**:
- [X] this PR contains documentation
- [ ] this PR contains tests

### DIFF
--- a/docs/hugo/content/guide/aso-controller-settings-options.md
+++ b/docs/hugo/content/guide/aso-controller-settings-options.md
@@ -150,6 +150,6 @@ AZURE_USER_AGENT_SUFFIX is appended to the default User-Agent for Azure HTTP cli
 
 **Format:** `string`
 
-**Example:** `" my-user-agent"`
+**Example:** `"my-user-agent"`
 
 **Required**: False
--- a/docs/hugo/content/guide/aso-controller-settings-options.md
+++ b/docs/hugo/content/guide/aso-controller-settings-options.md
@@ -143,3 +143,13 @@ Note that the resource manager audience is referred to as "activeDirectoryResour
 **Example:** `"https://management.core.chinacloudapi.cn/"`
 
 **Required**: False
+
+### AZURE_USER_AGENT_SUFFIX
+
+AZURE_USER_AGENT_SUFFIX is appended to the default User-Agent for Azure HTTP clients.
+
+**Format:** `string`
+
+**Example:** `" my-user-agent"`
+
+**Required**: False
--- a/v2/internal/genericarmclient/generic_client.go
+++ b/v2/internal/genericarmclient/generic_client.go
@@ -66,7 +66,7 @@ func NewGenericClient(
 
 	ua := options.UserAgent
 	if ua == "" {
-		ua = userAgent
+		ua = DefaultUserAgent
 	}
 
 	opts := &arm.ClientOptions{
--- a/v2/internal/genericarmclient/generic_client.go
+++ b/v2/internal/genericarmclient/generic_client.go
@@ -66,7 +66,7 @@ func NewGenericClient(
 
 	ua := options.UserAgent
 	if ua == "" {
-		ua = DefaultUserAgent
+		ua = userAgent
 	}
 
 	opts := &arm.ClientOptions{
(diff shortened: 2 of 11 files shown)


docs/hugo/content/guide/aso-controller-settings-options.md
```diff
@@ -143,3 +143,13 @@ Note that the resource manager audience is referred to as "activeDirectoryResour
...
 
 **Required**: False
+
+### AZURE_USER_AGENT_SUFFIX
+
+AZURE_USER_AGENT_SUFFIX is appended to the default User-Agent for Azure HTTP clients.
+
+**Format:** `string`
+
+**Example:** `" my-user-agent"`
```
nojnhuh, line 153: This leading whitespace is intentional since there's no other hard-coded separator between this and the default.

v2/internal/genericarmclient/generic_client.go
```diff
@@ -66,7 +66,7 @@ func NewGenericClient(
 
 	ua := options.UserAgent
 	if ua == "" {
-		ua = userAgent
+		ua = DefaultUserAgent
```
theunrepentantgeek, line 69: This will only be used if the code that calls `NewGenericClient()` doesn't specify a user agent in the `GenericClientOptions` struct passed in. I don't think that's ideal, as the configured setting wouldn't be included.

I haven't checked where `GenericClientIOptions.UserAgent` is set, but I'm wondering if the value used there should be used as a _part_ of the UA, not the whole thing.

v2/internal/reconcilers/arm/arm_client_cache.go
```diff
@@ -29,14 +29,16 @@ type ARMClientCache struct {
...
 }
 
 func NewARMClientCache(
 	credentialProvider identity.CredentialProvider,
 	kubeClient kubeclient.Client,
 	configuration cloud.Configuration,
 	httpClient *http.Client,
-	armMetrics *metrics.ARMClientMetrics) *ARMClientCache {
+	armMetrics *metrics.ARMClientMetrics,
+	userAgent string) *ARMClientCache {
```
theunrepentantgeek, line 41: Instead of plumbing the UA through here, what if there was a method in the `genericarmclient` package that could be called to _add_ a string to the UA for this process?

The code in `setup.go` would then be

``` 
genericarmclient.AddToUserAgent(cfg.UserAgentSuffix)
```

The variable `DefaultUserAgent` woudl revert to package internal and everything else would continue as before.

theunrepentantgeek (commented):
Generally looks good, but I have some ideas about an alternative way to plumb the value in.

v2/internal/reconcilers/arm/arm_client_cache.go
nojnhuh: Good call, that's a lot less involved.

docs/hugo/content/guide/aso-controller-settings-options.md
theunrepentantgeek: As far as I know, yes. 😁

v2/internal/genericarmclient/useragent_policy.go
```diff
@@ -41,3 +40,8 @@ func (p userAgentPolicy) Do(req *policy.Request) (*http.Response, error) {
 	req.Raw().Header.Set("user-Agent", newUserAgent)
 	return req.Next()
 }
+
+// AddToUserAgent appends the given suffix to the default user agent.
+func AddToUserAgent(suffix string) {
+	userAgent += " " + suffix
```
theunrepentantgeek, line 46: Suggest only adding the suffix if the value passed in is non-empty. As it stands, if nothing is configured, we add an empty space to the end of the UA.

theunrepentantgeek (approved):
One minor comment, otherwise looks good. I'll trigger testing and merge once that's addressed.

v2/internal/genericarmclient/useragent_policy.go
nojnhuh: Fixed, thanks!

PR #3672, opened 2023-12-21
Pull in latest ARM specs

ross-p-smith:
This PR updates the Git Submodule so that we can pull the APIM v2 SKU

I'm unsure why I have to define a commit id - shouldn't we just take the latest?

gif


theunrepentantgeek:
> I'm unsure why I have to define a commit id - shouldn't we just take the latest?

Submodules aways pull in a specific version so that you can always get a reproducible build.

Think of it this way ... if the December 2023 ASO build just pulled in the _latest_ version of specs, we wouldn't be able to go back to that exact version six months later - because then we'd have the December 2023 ASO code + the June 2024 specs.

So instead, the December 2023 ASO build pulls in (say) the October 2023 specs, and when we go back in six months time we still have the same source to build from.

It looks as though the 2021-06-01 API version for Synapse go updated; if you run the generator to update those files and commit them to this PR, everything should build and we can then merge.

(Aside: I'm on leave until mid-January, so responses will be delayed.)

ross-p-smith:
Closing this as it's in #3673

PR #3685, opened 2024-01-05
Stop hardcoding properties in newConvertToARMFunctionBuilder

### LINKED ISSUE #3549: Don't hardcode properties in newConvertToARMFunctionBuilder
`v2/tools/generator/internal/armconversion/convert_to_arm_function_builder.go` has the following code snippet

```
	var emptyCollectionProperties set.Set[string]
	if forceEmptyCollections {
		// TODO: These should not be hardcoded here, instead should be in the azure-arm.yaml config
		emptyCollectionProperties = set.Make(
			"Tags",
			"NodeLabels",
			"NodeTaints")
	}
```

**Describe the improvement**
The properties should be specified in the configuration file, not in code.

matthchr:
This fixes #3549.

This uses a property tag for a few reasons:
  1. Some types are flattened and other approaches are easily lost when flattening occurs. Putting the tag onto the property preserves it even through flattening.
  2. In many ways this behavior is like an augmented `json:omitempty`, so it (IMO) makes sense to have a tag just like for JSON. It makes it clearer when looking at the Go object that the serialization behavior of these fields is special.

This also fixes an issue with the configuration where hierarchical configurations didn't propagate the checks for usage.

gif

**If applicable**:
- [ ] this PR contains tests


v2/tools/generator/internal/config/object_model_configuration.go
```diff
@@ -62,8 +62,15 @@ func NewObjectModelConfiguration() *ObjectModelConfiguration {
 	}
 
 	// Initialize group access fields here (alphabetical, please)
```
super-harsh, line 64: Do we still need this comment here?

super-harsh, line 64: Do we still need this comment here?

v2/tools/generator/internal/codegen/pipeline/add_arm_conversion_interface.go
```diff
@@ -231,20 +237,13 @@ func (c *armConversionApplier) addARMConversionInterface(
...
-
 	addInterfaceHandler := func(t *astmodel.ObjectType) (astmodel.Type, error) {
 		result := t.WithInterface(armconversion.NewARMConversionImplementation(
 			armDef.Name(),
 			objectType,
+			kubeDef.Name(),
 			c.idFactory,
-			typeType,
-			payloadType))
+			typeType))
```
super-harsh, line 246: Should we consider renaming `typeType` to `typeKind`?

super-harsh (approved):

v2/tools/generator/internal/config/object_model_configuration.go
matthchr: I think so?

v2/tools/generator/internal/codegen/pipeline/add_arm_conversion_interface.go
matthchr: Good catch - done

v2/tools/generator/internal/config/object_model_configuration.go
theunrepentantgeek: Yes, as a guidepost for future maintainers that _here_ is where group access is configured.

PR #3687, opened 2024-01-06
Contributing: arm64 support

mehighlow:

- This PR fixes issues with packages for ARM64 architecture
- Updates `azwi` as macOS arm64 support has been introduced starting version 1.1.0
- Formats documentation

### DIFF
--- a/.devcontainer/install-dependencies.sh
+++ b/.devcontainer/install-dependencies.sh
@@ -255,7 +255,7 @@ fi
 # Install yq
 #doc# | YQ | v4.13.0 | https://github.com/mikefarah/yq/ |
 yq_version=v4.13.0
-yq_binary=yq_linux_amd64
+yq_binary=yq_${os}_${arch}
 write-verbose "Checking for $TOOL_DEST/yq"
 if should-install "$TOOL_DEST/yq"; then 
     write-info "Installing yq..."
--- a/.devcontainer/install-dependencies.sh
+++ b/.devcontainer/install-dependencies.sh
@@ -36,7 +36,7 @@ set -eu
 # Be sure to use include the `#doc#` prefix on each line.
 #
 
-VERBOSE=true
+VERBOSE=false
 SKIP=false
 DEVCONTAINER=false
 
@@ -87,7 +87,9 @@ if [ "$DEVCONTAINER" = true ]; then
     BUILDX_DEST=/usr/lib/docker/cli-plugins
 else
     TOOL_DEST=$(git rev-parse --show-toplevel)/hack/tools
+    CROSSPLANE_CONFIG_DIR=$(git rev-parse --show-toplevel)/hack/crossplane/config
     mkdir -p "$TOOL_DEST"
+    mkdir -p "$CROSSPLANE_CONFIG_DIR"
     KUBEBUILDER_DEST="$TOOL_DEST/kubebuilder"
     BUILDX_DEST=$HOME/.docker/cli-plugins
 fi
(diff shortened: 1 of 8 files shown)


ManuInNZ:
not really related but as I saw the macos arm64 and cross plane, I am trying my luck.

is there more details or content to explore/use the crossplane hack?

cheers

docs/hugo/content/contributing/developer-setup.md
```diff
@@ -10,11 +10,11 @@ description: "How to set up your developer environment for Azure Service Operato
...
-- Dev Container with VS Code on Linux
-- Dev Container with VS Code on Windows
-- Docker on Linux
-- CLI on Linux
-- CLI on MacOS
+- [Dev Container with VS Code on Linux](#dev-container-with-vs-code-on-linux)
+- [Dev Container with VS Code on Windows](#dev-container-with-vs-code-on-windows)
+- [Docker on Linux](#docker-on-linux)
+- [CLI on Linux](#cli-on-linux)
+- [CLI on MacOS](#cli-on-macos)
```
super-harsh, line 17: With this change, I reckon statement on line 84(You'll need to manually install the tools as listed by `.devcontainer/install-dependencies.sh`) proves to be false. Would be great to update that section as well.

mehighlow: I've updated macOS instructions. You can still use either `.devcontainer` approach or to install tools manually, executing `.devcontainer/install-dependencies.sh` script which downloads all necessary tools to the localhost.
I tested both scenarios on my apple silicon-based mac.

mehighlow:
> not really related but as I saw the macos arm64 and cross plane, I am trying my luck.
> 
> is there more details or content to explore/use the crossplane hack?
> 
> cheers

IDK, frankly speaking. It is better to ask this in the #azure-service-operator channel on the [Kubernetes Slack](https://communityinviter.com/apps/kubernetes/community)

.devcontainer/install-dependencies.sh
```diff
@@ -87,7 +87,9 @@ if [ "$DEVCONTAINER" = true ]; then
     BUILDX_DEST=/usr/lib/docker/cli-plugins
 else
     TOOL_DEST=$(git rev-parse --show-toplevel)/hack/tools
+    CROSSPLANE_CONFIG_DIR=$(git rev-parse --show-toplevel)/hack/crossplane/config
```
matthchr, line 90: This PR is amazing.

The only thing I'm not sure about is, this directory is created/managed by `task`, and I don't think should need to be pre-created in the devcontainer.

I think it might make more sense to just create that directory and include a dummy `config/default/kustomization.yaml` file, that way the path always exists.

I feel that's a better solution than fixing it in the dockerfile as in theory it should fix it for users running the build outside of a container too, and it keeps the Dockerfile clean(er), without knowledge of our specific build layout for the most part.

matthchr (approved):
LGTM with one minor nit.

ManuInNZ:
> > not really related but as I saw the macos arm64 and cross plane, I am trying my luck.
> > 
> > is there more details or content to explore/use the crossplane hack?
> > 
> > cheers
> 
> IDK, frankly speaking. It is better to ask this in the #azure-service-operator channel on the [Kubernetes Slack](https://communityinviter.com/apps/kubernetes/community)

thanks

.devcontainer/install-dependencies.sh
matthchr: I sent a PR to fix this.

PR #3699, opened 2024-01-12
feat: Provide experimental images built from main

### LINKED ISSUE #3682: Introduce experimental image
As of today, you have to wait for a new release or build the container image yourself to have a preview image.

**Describe the improvement**
Introduce a nightly image that gets built every day, or even for every main push, that can be pulled for dev/test/experimentation through official channels.

**Additional context**
Relates to https://github.com/Azure/azure-service-operator/issues/3648

tomkerkhove:

Provide experimental images built from main so that end-users can easily run the latest bits from main without having to build the images themselves.

Currently it is pushing to the same image repo / path, but we may want to change that to be more explicit in the usage scenarios.

The release creation is inspired by how [Higress](https://github.com/alibaba/higress/releases/tag/latest) does it, but I'm happy to adapt.

Examples:
- GitHub Release ([link](https://github.com/tomkerkhove/azure-service-operator/releases/tag/experimental))
- Image (not public though): `apimaso.azurecr.io/azureserviceoperator:experimental`

I am new to Make so if I'm doing things wrong/hacky, then I'm happy to improve

gif

**If applicable**:
- [ ] this PR contains tests

Closes #3682


.github/workflows/create-release-main.yml
```diff
@@ -0,0 +1,71 @@
...
+            This is the "latest" releasewhich contains the most recent commits from the main branch. (commit: ${{ github.sha }}).
+
+            **This release might not be stable.**
+
+            It is only intended for developers wishing to try out the latest feature, some of which may not be fully implemented.
+
+      - name: Docker login
+        run: |
+          container_id=${{env.container_id}}
+          docker exec -e AZURE_CLIENT_ID -e AZURE_CLIENT_SECRET -e DOCKER_REGISTRY "$container_id" task VERSION=${{ env.ARTIFACT_VERSION }} LATEST_VERSION_TAG=${{ env.ARTIFACT_VERSION }} docker-login
```
super-harsh, line 60: Docker login would only require `AZURE_CLIENT_ID` and `AZURE_CLIENT_SECRET` here.

.github/workflows/create-release-main.yml
```diff
@@ -0,0 +1,71 @@
+name: Create new preview release (main)
+
+on:
+  # run when pushed to main is published,
+  # which creates a new tag
+  push:
+    branches:
+      - main
```
super-harsh, line 8: IMO this should be every 24 hours or every weekend.

.github/workflows/create-release-main.yml
```diff
@@ -0,0 +1,71 @@
+name: Create new preview release (main)
```
super-harsh, line 1: Suggestion: Would be better to make it sound like its pushing image per day/week? 

something like - Build and push nightly/weekly ASO image

.github/workflows/create-release-main.yml
```diff
@@ -0,0 +1,71 @@
...
+          docker build --tag devcontainer:latest .devcontainer
+          mkdir -p $HOME/.docker # in case it doesn't exist
+          container_id=$(docker create -w /workspace -v $GITHUB_WORKSPACE:/workspace -v /var/run/docker.sock:/var/run/docker.sock devcontainer:latest)
+          docker start "$container_id"
+          echo "container_id=$container_id" >> $GITHUB_ENV
+
+      - name: Build required release files
+        run: |
+          container_id=${{env.container_id}}
+          docker exec "$container_id" task VERSION=${{ env.ARTIFACT_VERSION }} LATEST_VERSION_TAG=${{ env.ARTIFACT_VERSION }} make-release-artifacts
```
super-harsh, line 38: Do we need to generate artifacts here as well? if yes, not sure where do we publish them?

.github/workflows/create-release-main.yml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 # required to access tags
+          submodules: "true"
+
+      - name: Build & run devcontainer image
+        # this always builds a new image from scratch rather than from the build-devcontainer-image workflow output
+        # so that we pick up the latest versions of everything
+        # NB: if you update this also update live-validation.yml
```
super-harsh, line 26:
```suggestion
        # NB: if you update this also update live-validation.yml and create-release-official.yaml
```

super-harsh (commented):
Thanks for the PR. A few minor suggestions/questions below

.github/workflows/create-release-main.yml
tomkerkhove: That is what i thought, but it tries to resolve the version anyway for some reason.

Example:
- Commit: https://github.com/tomkerkhove/azure-service-operator/commit/6d5345a1debab58f736c9e4f19c8857545583bd2
- Failure because Git tags are not applicable here: https://github.com/tomkerkhove/azure-service-operator/actions/runs/7503829500/job/20429481194

.github/workflows/create-release-main.yml
tomkerkhove: They are attached to the release it pushes them to, that's enough for this scenario imo

.github/workflows/create-release-official.yml
```diff
@@ -1,59 +1,59 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 # required to access tags
+          submodules: "true"
+
+      - name: Build & run devcontainer image
+        # this always builds a new image from scratch rather than from the build-devcontainer-image workflow output
+        # so that we pick up the latest versions of everything
+        # NB: if you update this also update live-validation.yml
```
tomkerkhove, line 23:
```suggestion
        # NB: if you update this also update live-validation.yml and create-release-main.yaml
```

.github/workflows/create-release-main.yml
tomkerkhove: Let's figure out https://github.com/Azure/azure-service-operator/pull/3699#discussion_r1451794792 first and then we can align

.github/workflows/create-release-main.yml
tomkerkhove: Every 24h or every week is another option, but it feels like main images are more ideal given you don't want to wait to try something exciting that was contributed. But that is just my opinion of course, I am not a maintainer :) If we do not go with main, which is fairly common, then we should go with daily and not weekly as that is too long.

Additionally, I don't see the benefit of pushing a new version if there are no changes.

Manual trigger is something I'm planning on contributing once this gets merged, but from a non-main branch feels only reasonable scenario

theunrepentantgeek:
> I am new to Make so if I'm doing things wrong/hacky, then I'm happy to improve

I haven't reviewed the actual changes yet (doing that soon), but ASO v2 does not use Make - we use [Task](https://taskfile.dev/) for built coordination. (The existing Makefile is a remnant of ASO v1, and only used for maintenance of that.)

.github/workflows/create-release-main.yml
theunrepentantgeek: Manual trigger is useful for testing purposes, esp when someone is working on changing/fixing this Github Action.

Triggering on pushes to `main` is probably the best approach - I agree that weekly is too long, and daily has a bunch of downsides (two being new images even when there are no changes, enforced waiting to try things out).

However, lots of things push to `main` that wouldn't impact on the image, so I think we need some path filtering as well. E.g. we don't need a new image if our documentation is updated under `/docs/`, nor if there's a bug fix or security update for ASO v1.

.github/workflows/create-release-main.yml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 # required to access tags
+          submodules: "true"
+
+      - name: Build & run devcontainer image
+        # this always builds a new image from scratch rather than from the build-devcontainer-image workflow output
+        # so that we pick up the latest versions of everything
+        # NB: if you update this also update live-validation.yml and create-release-official.yaml
```
theunrepentantgeek, line 26: There are similar comments in `create-release.yml`, `live-validation.yml` and `pre-release-tests.yaml` - all four files should reference each other so that someone making a change is clued-in to the need to keep all four in sync.

.github/workflows/create-release-main.yml
```diff
@@ -0,0 +1,71 @@
...
+          tag: refs/tags/${{ env.ARTIFACT_VERSION }}
+          file: "v2/out/release/*"
+          file_glob: true
+          prerelease: true
+          overwrite: true
+          release_name: "Experimental"
+          body: |
+            This is the "latest" releasewhich contains the most recent commits from the main branch. (commit: ${{ github.sha }}).
+
+            **This release might not be stable.**
```
theunrepentantgeek, line 53:
```suggestion
            **This release might not be stable. Use at your own risk.** 
```

.github/workflows/create-release-main.yml
```diff
@@ -0,0 +1,71 @@
...
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          tag: refs/tags/${{ env.ARTIFACT_VERSION }}
+          file: "v2/out/release/*"
+          file_glob: true
+          prerelease: true
+          overwrite: true
+          release_name: "Experimental"
+          body: |
+            This is the "latest" releasewhich contains the most recent commits from the main branch. (commit: ${{ github.sha }}).
```
theunrepentantgeek, line 51: I think we should avoid using the term _latest_ as that may be misinterpreted.

```suggestion
            This is an experimental release which containing the most recent commits from the main branch as of commit: ${{ github.sha }}.
```

Also, is there a convenient timestamp we could include in this message as well, so it's easy for a reader to determine whether it's stale or not?

.github/workflows/create-release-official.yml
```diff
@@ -1,59 +1,59 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 # required to access tags
+          submodules: "true"
+
+      - name: Build & run devcontainer image
+        # this always builds a new image from scratch rather than from the build-devcontainer-image workflow output
+        # so that we pick up the latest versions of everything
+        # NB: if you update this also update live-validation.yml and create-release-main.yaml
```
theunrepentantgeek, line 23: As noted above, there are now four files that need to be updated together.

.github/workflows/create-release-main.yml
theunrepentantgeek: I think we need to be consistent with terminology - so far I've spotted _preview_, _experimental_ and _latest_ all used to describe this.

I suggest we standardize on _experimental_, as the other two terms have common meanings that don't suit (that is, latest = most recent release, preview = beta release).

I suggest
```
name: Create new release (experimental)
```

.github/workflows/create-release-main.yml
tomkerkhove: That is why I added the commit, but will see if I can add a timestamp although it would be the release's date:
image

Verifying with a new run in my fork to see if those are properly updated. (one would expect)

.github/workflows/create-release-main.yml
tomkerkhove: Agreed, I will align things!

.github/workflows/create-release-main.yml
tomkerkhove: > Manual trigger is useful for testing purposes, esp when someone is working on changing/fixing this Github Action.

As mentioned above, I was planning on doing that as a follow-up PR to make the change more explicit and easier to review given we may want to handle things differently, is it OK to do it in a new PR?

We (APIM team) have the same plans, was just not added yet.

> However, lots of things push to main that wouldn't impact on the image, so I think we need some path filtering as well. E.g. we don't need a new image if our documentation is updated under /docs/, nor if there's a bug fix or security update for ASO v1.

That definitely makes sense to me, but I was not 100% sure which filters to use. Would you mind sharing the ones you think make sense or shall I just exclude docs as the rest all feel related?

.github/workflows/create-release-main.yml
tomkerkhove: Release does not get updated, release assets do:

.github/workflows/create-release-main.yml
tomkerkhove: Updated all references and opened https://github.com/Azure/azure-service-operator/issues/3712

.github/workflows/create-release-main.yml
tomkerkhove: Aligned

.github/workflows/create-release-main.yml
tomkerkhove: Added filter to exclude docs for now.

tomkerkhove: Tracking build from non-main branch in https://github.com/Azure/azure-service-operator/issues/3713

theunrepentantgeek: > As mentioned above, I was planning on doing that as a follow-up PR to make the change more explicit

I think we may be talking about two different things. I was thinking about adding an additional trigger to your workflow:

``` yaml
on:
  workflow_dispatch:
  # no content, allows manual triggering
```

This would allow repo contributors to directly trigger the workflow as needed.

theunrepentantgeek: > Added filter to exclude docs for now.

I'd invert it to only rebuild the image if something under the `v2` folder is modified.

.github/workflows/create-release-main.yml
theunrepentantgeek: Good enough. Thanks for checking.

theunrepentantgeek (approved):

.github/workflows/create-release-main.yml
tomkerkhove: Valid point, I thought you wanted to have customer tags; added both!

theunrepentantgeek (approved):

.github/workflows/create-release-experimental.yml
```diff
@@ -0,0 +1,75 @@
...
+          docker exec -e AZURE_CLIENT_ID -e AZURE_CLIENT_SECRET -e DOCKER_REGISTRY "$container_id" task VERSION=${{ env.ARTIFACT_VERSION }} LATEST_VERSION_TAG=${{ env.ARTIFACT_VERSION }} docker-login
+        env:
+          DOCKER_REGISTRY: ${{ secrets.REGISTRY_LOGIN }}
+          AZURE_CLIENT_ID: ${{ secrets.AZURE_CLIENT_ID }}
+          AZURE_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
+
+      - name: Build, tag and push docker image
+        run: |
+          container_id=${{env.container_id}}
+          docker exec -e DOCKER_PUSH_TARGET "$container_id" task VERSION=${{ env.ARTIFACT_VERSION }} LATEST_VERSION_TAG=${{ env.ARTIFACT_VERSION }} controller:docker-push-multiarch
```
matthchr, line 73: re-tagging different docker images over and over is possible to cause problems, I think. See [the default image pull policy](https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting)

We currently set `imagePullPolicy` to `Always` (which TBH is probably wrong), but would mitigate this issue at least somewhat... though even then there are possible awkward situations with for example the following situations:
1. User installs experimental ASO via YAML.
2. We merge a new PR containing a new resource, including webhooks, Kubernetes RBAC roles, etc
3. Users pod restarts for whatever reason.
4. Users pod pulls the new image, _but_ the resource YAMLs (including RBAC roles, webhooks) are from the N-1st version of experimental. Pod will try to install the new CRD that is now embedded inside of it from the new resource (assuming configuration is `crdPattern=*`), but the pod will crash during launch because it will attempt to start a watch on the resource but it doesn't have permission to do so.

The workaround there is that you can delete the ASO namespace and re-install from latest and that should fix it. The number of times this happens will likely not be _super_ common... but we should probably document this behavior and _really really_ emphasize that people need to not run this in production. It literally won't work if they do it long enough because we will add resources and this will happen and break them.

It's much less of a problem for doing local sorta testing, kicking the tires, etc - but running for a week in a dev or test cluster is more likely to run into this issue.

Possible solutions are to pin the specific image hash...

.github/workflows/create-release-official.yml
```diff
@@ -1,59 +1,59 @@
...
-          DOCKER_REGISTRY: ${{ secrets.REGISTRY_LOGIN }}
-          AZURE_CLIENT_ID: ${{ secrets.AZURE_CLIENT_ID }}
-          AZURE_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
-
-      - name: Build, tag and push docker image
-        run: |
-          container_id=${{env.container_id}}
-          docker exec -e DOCKER_PUSH_TARGET "$container_id" task controller:docker-push-multiarch
-        env:
-          DOCKER_PUSH_TARGET: ${{ secrets.REGISTRY_PUBLIC }}
```
matthchr, line 59: Weird how GitHub is showing this diff -- from what I can tell you just changed the comments is that right?

matthchr (approved):

.github/workflows/create-release-official.yml
tomkerkhove: Correct, and I renamed the file

.github/workflows/create-release-experimental.yml
tomkerkhove: I see your point but it's more of a use-case scenario, tagging with the commit is a good addition to both scenarios which users can use depending on their needs.

Would that make sense? If so, I'll open a follow-up PR with docs + additional image pushed with the sha tag

.github/workflows/create-release-experimental.yml
```diff
@@ -0,0 +1,75 @@
+name: Create new experimental release
+
+on:
+  # run when pushed to main is published,
+  # which creates a new tag
+  push:
```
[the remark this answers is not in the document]
tomkerkhove: Correct!

.github/workflows/create-release-experimental.yml
timja: personally I think it would be good to have one with a commit, even if it's just built nightly so that not all commits are built.

If I need a change sooner than the current 2 month release schedule I would like to use a fixed image so I can get changes sooner and provide feedback (and continue running it until the release).

A moving image isn't very useful except for a quick test.

tomkerkhove: These are the two scenarios that I often see in the various projects that I maintain/interact with:
1. `:main`/`:experimental` - I want  to run the latest and greatest, don't ask me to check what the latest commit was
2. `:<sha>` - I want to test a given change and stick to it, if I want another change, then I will look it up

I believe ASO needs both and 2) is what Tim is looking for as well

matthchr: Do you actually need to push a specific SHA-tagged image? I don't think you do because you can always pull a sha-tagged image regardless.

I just tested this locally and it already works:
`docker pull mcr.microsoft.com/k8s/azureserviceoperator@sha256:d82f04240b3b72881342ae1cd0714e154cbaa00bc9f8fb226e7ec8c23fefe28f` - which is the sha for v2.3.0.

matthchr: The only thing is that is needed is that, the YAML in the `experimental` release will be referring to the YAML tag, so users will need to use `kustomize` or some other mechanism to update that to refer to a specific SHA... unless we modify the release to produce a YAML that refs a sha rather than the `experimental` tag directly, which possibly is worth doing but may need some build script changes.

tomkerkhove: Fair enough - We should be good to go then!

With regards to the YAML; if they want to use a SHA instead of experimental tag, then I'd suggest we leave it up to customers to do a find/replace after downloading the artifacts, I would not offer this OOTB personally.

.github/workflows/create-release-experimental.yml
matthchr, line 6: Some math: If we do push on each commit, the image we push is ~170mb compressed. acr starts charging for over 100GB consumed, and we merge roughly 2 commits a day (620 since 1 year ago), so that's a growth of ~300mb / day == 9GB/month.

We'll likely need to clean up old images eventually. I think that we can enable that feature in our ACR if we upgrade to premium.

How does a retention time for untagged images of 45d sound? We ship official releases probably around that often, and worst case if you're using a tagged image and it gets pruned you cna swap to a newer version of the image anyway?

theunrepentantgeek: I think 45d is reasonable; we can change if anyone complains. 

The core scenarios we're enabling here are for dev & test, and 45d should be enough for most cases.

matthchr: Before we merge this let me fix our ACR settings then.

matthchr: k, I've updated our ACR to be premium tier with a 45d retention and also protected all of the ASO images in it.

tomkerkhove: > Some math: If we do push on each commit, the image we push is ~170mb compressed. acr starts charging for over 100GB consumed, and we merge roughly 2 commits a day (620 since 1 year ago), so that's a growth of ~300mb / day == 9GB/month.

That's a good estimate, but in reality various layers of the images should be re-used and the net storage impact should be a lot lower, no?

PR #3717, opened 2024-01-18
Protect acr images from updation or deletion

matthchr:


theunrepentantgeek (approved):

scripts/v2/aso-acr-protect.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
```
super-harsh, line 2: Is this script suppose to be used in task target `controller:acr-protect-image` ?

matthchr: No, it's a standalone script to protect sections of the acr. I used it to backfill protection

PR #3734, opened 2024-01-22
Add permissions section to workflows

### LINKED ISSUE #3732: Add permissions blocks to action YAMLs
See https://docs.opensource.microsoft.com/github/apps/permission-changes/

matthchr:
This resolves #3732

### DIFF
--- a/.github/workflows/api-docs-repo.yaml
+++ b/.github/workflows/api-docs-repo.yaml
@@ -11,6 +11,9 @@ on:
 jobs:
   gen-crd-docs:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: read
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
--- a/.github/workflows/helm-chart-repo.yaml
+++ b/.github/workflows/helm-chart-repo.yaml
@@ -18,6 +18,11 @@ env:
 jobs:
   generate:
     runs-on: ubuntu-latest
+
+    permissions:
+      contents: read
+      packages: read
+
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
--- a/.github/workflows/create-release-experimental.yml
+++ b/.github/workflows/create-release-experimental.yml
@@ -17,6 +17,10 @@ env:
 jobs:
   build-and-push:
     runs-on: [self-hosted, 1ES.Pool=aso-1es-pool]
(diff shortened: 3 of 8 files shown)


.github/workflows/helm-chart-repo.yaml
```diff
@@ -93,7 +98,7 @@ jobs:
       - name: Create Pull Request
         uses: devops-infra/action-pull-request@v0.4.2
         with:
-          github_token: ${{ secrets.GH_PAT }}
+          github_token: ${{ secrets.GH_PAT }}  # TODO: Do we remember why we use GH_PAT here and not GITHUB_TOKEN?
```
matthchr, line 101: Was it because we didn't want to give `write` perms to `GITHUB_TOKEN`?

theunrepentantgeek: It's because we _can't_.

We earlier found that an action running with `GITHUB_TOKEN` _**cannot**_ create a PR, though we tried all sorts of things.

Root cause seems to be a combination of 
(a) The same permission that allows creation of PRs also allows approval, and automatic approval of PRs from uncontrolled public forks is hazardous
(b) Policies are tighter within the `Azure` organization

theunrepentantgeek: So we've been using PATs since then as a workaround. 

But it seems there's a third option we should investigate:

It seems we can set up an App to generate a token for us [[Link](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#authenticating-with-github-app-generated-tokens)]..

.github/workflows/api-docs-repo.yaml
```diff
@@ -11,6 +11,10 @@ on:
 jobs:
   gen-crd-docs:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: read
+      pull-requests: write
```
theunrepentantgeek, line 17: I don't think `pull-requests: write` will work.

matthchr: Ah it's using the PAT to create the PR so it's not needed too.

matthchr: Fixed

theunrepentantgeek (approved):

Issue #3091, opened 2023-06-19
Modify samples tests to create resources in parallel


theunrepentantgeek:
**Describe the current behavior**

At the moment the samples tests create all the refs in one go, then all the resources under test. We've identified a case where this concealed a sequencing bug, and it would be good to discover any other cases that lurk in ASO.

**Describe the improvement**

Instead of creating the resources in two tranches, do them all at once via a single call to `CreateResourcesAndWait()`.

PR #3736, opened 2024-01-23
Feature Enhancement: Enable Backup Instance creation for Microsoft.DataProtection 

### LINKED ISSUE #2345: Feature: Add support for backup vault for storage accounts
It would be nice to be able to manage backup vaults with ASO, and be able to configure backup for storageAccount within that backup vault.

mayankagg9722:

This pull request introduces enhancement for Backup Instance creation for Microsoft.DataProtection. Additionally, we have added the necessary test recording that covers the creation, and deletion of these resources, namely Backup Instance with thorough validation.

Fixes #2345.

**Special notes for your reviewer**: Feature enhancement on top of already existing Microsoft.DataProtection CRDs.

**If applicable**:
- [x] this PR contains tests


theunrepentantgeek (changes requested):
You'll need to fix the sample, and include the recording of the test run.

v2/internal/controllers/crd_dataprotection_backupinstance_test.go
```diff
@@ -0,0 +1,279 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+package controllers_test
+
+import (
+	// The testing package is imported for testing-related functionality.
```
theunrepentantgeek, line 9: This comment belongs on the `testing` import, not `fmt` - though I'd argue it could be skipped entirely.

v2/internal/controllers/crd_dataprotection_backupinstance_test.go
```diff
@@ -0,0 +1,279 @@
...
+	tc.CreateResourceAndWait(blobService)
+
+	blobContainer := &storage.StorageAccountsBlobServicesContainer{
+		ObjectMeta: tc.MakeObjectMeta("velero"),
+		Spec: storage.StorageAccounts_BlobServices_Container_Spec{
+			Owner: testcommon.AsOwner(blobService),
+		},
+	}
+
+	tc.CreateResourceAndWait(blobContainer)
```
theunrepentantgeek, line 90: You shouldn't be creating the resources in strict sequence like this - this doesn't reflect the way a YAML file containing the resources will be processed the cluster and is therefore unrealistic.

When the YAML file is applied to the cluster, all the resources will be created and ASO will start reconciling all of them at the same time. ASO knows about owner references (and other dependencies), and therefore should know in which order to create things in Azure.

To test this, declare ALL your resource variables and then have a single call `tc.CreateResourcesAndWait(cluster, acct, blobService, etc, etc, etc)`.

theunrepentantgeek (commented):
Generally looks good. Just need to switch the test to create all the resources in one go.

super-harsh:
Thanks for the PR. This one closes #2345

v2/samples/dataprotection/v1api/v1api20230101_backupvaultsbackupinstance.yaml
```diff
@@ -0,0 +1,36 @@
...
+spec:
+  owner:
+    name: asotestbackupvault
+  properties:
+    objectType: BackupInstance
+    friendlyName: asotestbackupinstance
+    dataSourceInfo:
+      objectType: Datasource
+      datasourceType: Microsoft.ContainerService/managedClusters
+      resourceUri: /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/rg/providers/Microsoft.ContainerService/managedClusters/aks
```
super-harsh, line 15: I'd suggest replacing `resourceGroups/rg` with `resourceGroups/aso-sample-rg` here and below in armIDs.

mayankagg9722: done the changes

v2/samples/dataprotection/v1api/v1api20230101_backupvaultsbackupinstance.yaml
```diff
@@ -0,0 +1,36 @@
...
+      resourceLocation: westus2
+      resourceType: Microsoft.ContainerService/managedClusters
+    policyInfo:
+      policyId: /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/rg/providers/Microsoft.DataProtection/backupVaults/asotestbackupvault/backupPolicies/test
+      policyParameters:
+        dataStoreParametersList:
+          - dataStoreType: OperationalStore
+            objectType: AzureOperationalStoreParameters
+            resourceGroupId: /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/rg
+      backupDatasourceParametersList:
```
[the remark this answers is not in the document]
mayankagg9722: done the changes

v2/internal/controllers/crd_dataprotection_backupinstance_test.go
mayankagg9722: done the changes

v2/internal/controllers/crd_dataprotection_backupinstance_test.go
mayankagg9722: I do need order in certain places as extension installation itself takes internal input in properties for blob and storage account details. Can I keep it intact for now, given mutiple azure permissions are also needed?

image

mayankagg9722:
> You'll need to fix the sample, and include the recording of the test run.

Test recordings are included:
https://github.com/mayankagg9722/azure-service-operator/blob/main/v2/internal/controllers/recordings/Test_Dataprotection_Backupinstace_CRUD.yaml

v2/internal/controllers/crd_dataprotection_backupinstance_test.go
mayankagg9722: Also for trusted access though we have owner ref as cluster but we do need backup  vault msi to be created and present before we create TA binding.

image

v2/samples/dataprotection/v1api/v1api20230101_backupvaultsbackupinstance.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: dataprotection.azure.com/v1api20230101
```
theunrepentantgeek, line 1: The folder this file is in should be `v1api20230101`, not `v1api`, plus you're still missing the recording of the sample being verified.

See [_Run test for added sample and commit the recording_](https://azure.github.io/azure-service-operator/contributing/add-a-new-code-generated-resource/#run-test-for-added-sample-and-commit-the-recording) in our contributor docs.

v2/internal/controllers/crd_dataprotection_backupinstance_test.go
theunrepentantgeek: _Apologies, but I've run out of time today to give your PR a thorough review; I'll do that tomorrow._ 

From a cursory review, this is my major concern:

If your test requires everything to be done in a certain order, how will an end-user create one of those resources?

A user will want to use `kubectl apply` to create all the resources at one go based on a single YAML file; they won't be creating resources one at a time, harvesting details from each one to hand-modify the next in turn.

In fact, the ability of a user to submit a single YAML file containing dozens of resources, relying on ASO (and other operators) to untangle everything and do the right thing is a fundamental design goal.

ASO knows about ownership references - it won't try to create a child resource until the owner exists.
ASO also knows about other kinds of references - if a resource needs a secret, ASO won't try to create the resource until the secret is available. Same for config maps.

If one resource needs to consume information generated by another, the usual approach is this:

* The first resource is configured to publish the information to a specific configmap (or to a secret, if it's sensitive)
* The second resource is configured to read that information from the same configmap (or secret)

As mentioned above, ASO knows about those links - so it will wait for the first resource to publish that information to the config map before trying to create the second resource.

Much of this is set up in the `azure-arm.yaml` file, configuring the generator:

`$azureGeneratedSecrets` - specify properties returned from Azure that can be published as secrets. Requires some straightforward code to extract the values.

`$manualConfigs` - specifies configmaps that can be published based on the information returned from Azure. Also requires a smidge of code.

`$generatedConfigs` - similar, but autopublishes specific properties to configmaps, no code required.

**Action**: Please review your resources to see how much of the wiring-up between resources that you've done manually in your test can be achieved using these. Let me know how you get on.

It's _possible_ your resource needs something new - we build out support for these as needed (@matthchr is currently working on references to entire collections of secrets), so if the API Management Resource Provider does something in a way that we haven't seen before, we may need to do some work behind the scenes for you.

(And sorry for the wall of text. There's a lot of information here, happy to answer any questions that come up.)

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
```diff
@@ -0,0 +1,142 @@
...
+				id, _ := genruntime.GetAndParseResourceID(backupInstance)
+				subscription := id.SubscriptionID
+				rg := id.ResourceGroupName
+				vaultName := id.Parent.Name
+
+				log.V(Debug).Info("########################## Starting NewBackupInstancesClient for Backup Instance ##########################")
+				clientFactory, err := armdataprotection.NewClientFactory(subscription, armClient.Creds(), armClient.ClientOptions())
+				if err != nil {
+					log.Error(err, "failed to create armdataprotection client")
+				}
```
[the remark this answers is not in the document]
theunrepentantgeek: Immediately after the `PostReconcileCheck` extension is invoked, the Ready condition of the resource is updated, as can be seen in this code from `azure_generic_arm_reconciler_instance.go`:

``` go
	if err != nil {
		impactingError, ok := conditions.AsReadyConditionImpactingError(err)
		if !ok {
			impactingError = conditions.NewReadyConditionImpactingError(
				err,
				conditions.ConditionSeverityWarning,
				conditions.ReasonFailed)
		}

		return impactingError
	}
```

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
```diff
@@ -0,0 +1,142 @@
...
+				log.V(Debug).Info("########################## Starting BeginSyncBackupInstance for Backup Instance ##########################")
+				poller, err := clientFactory.NewBackupInstancesClient().BeginSyncBackupInstance(ctx, rg, vaultName, backupInstance.AzureName(), parameters, nil)
+				if err != nil {
+					log.Error(err, "failed BeginSyncBackupInstance to get the result")
+				}
+
+				_, err = poller.PollUntilDone(ctx, nil)
+				if err != nil {
+					log.Error(err, "failed BeginSyncBackupInstance to poll the result")
+				}
```
theunrepentantgeek, line 112: And here.

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
```diff
@@ -0,0 +1,142 @@
...
+
+				var parameters armdataprotection.SyncBackupInstanceRequest
+				parameters.SyncType = to.Ptr(armdataprotection.SyncTypeDefault)
+				log.V(Debug).Info("########################## Starting BeginSyncBackupInstance for Backup Instance ##########################")
+				poller, err := clientFactory.NewBackupInstancesClient().BeginSyncBackupInstance(ctx, rg, vaultName, backupInstance.AzureName(), parameters, nil)
+				if err != nil {
+					log.Error(err, "failed BeginSyncBackupInstance to get the result")
+				}
+
+				_, err = poller.PollUntilDone(ctx, nil)
```
theunrepentantgeek, line 109: How long does this take to execute? 

If this is more than a few seconds, there is a problem - while ASO is waiting on this to happen, no other reconciliations will be happening. This line is putting all reconciliation of all resources _**on hold**_ while this runs, effectively breaking ASO.

You should follow the pattern used by other long running operations - stash the URL required for the poll in an annotation on the resource, and check for it next time around.

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
```diff
@@ -0,0 +1,142 @@
...
+				log.V(Debug).Info("########################## Starting NewBackupInstancesClient for Backup Instance ##########################")
+				clientFactory, err := armdataprotection.NewClientFactory(subscription, armClient.Creds(), armClient.ClientOptions())
+				if err != nil {
+					log.Error(err, "failed to create armdataprotection client")
+				}
+
+				var parameters armdataprotection.SyncBackupInstanceRequest
+				parameters.SyncType = to.Ptr(armdataprotection.SyncTypeDefault)
+				log.V(Debug).Info("########################## Starting BeginSyncBackupInstance for Backup Instance ##########################")
+				poller, err := clientFactory.NewBackupInstancesClient().BeginSyncBackupInstance(ctx, rg, vaultName, backupInstance.AzureName(), parameters, nil)
```
theunrepentantgeek, line 104: Beginning a sync on every reconciliation seems odd - shouldn't you be checking whether that's needed or not first? If the resource is already in the goal state, there's no reason to do this.

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
```diff
@@ -0,0 +1,142 @@
...
+)
+
+func (extension *BackupVaultsBackupInstanceExtension) PostReconcileCheck(
+	ctx context.Context,
+	obj genruntime.MetaObject,
+	owner genruntime.MetaObject,
+	_ *resolver.Resolver,
+	armClient *genericarmclient.GenericClient,
+	log logr.Logger,
+	_ extensions.PostReconcileCheckFunc,
```
theunrepentantgeek, line 59: Do not discard this parameter - if your `PostReconcileCheck` works, you must call the one passed to you to maintain the chain.

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func (ext *BackupVaultsBackupInstanceExtension) PreReconcileCheck(
+	ctx context.Context,
+	obj genruntime.MetaObject,
+	owner genruntime.MetaObject,
+	_ *resolver.Resolver,
+	armClient *genericarmclient.GenericClient,
+	log logr.Logger,
+	_ extensions.PreReconcileCheckFunc,
```
theunrepentantgeek, line 126: Do not discard this parameter - if your extension works, you must call the one passed to you to maintain the chain.

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	// Type assert that we are the hub type. This will fail to compile if
+	// the hub type has been changed but this extension has not
+	var _ conversion.Hub = backupInstance
+
+	log.V(Debug).Info("########################## Ending Pre-reconcilation for Backup Instance ##########################")
+
+	return extensions.ProceedWithReconcile(), nil
+}
```
theunrepentantgeek, line 142: Your implementation is "do nothing" - either add the missing functionality or remove it entirely.

theunrepentantgeek: (In fact, by not calling the passed instance of `extensions.PreReconcileCheckFunc`, this implementation isn't "do nothing", it is "break ASO".)

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
```diff
@@ -0,0 +1,142 @@
...
+	"Invalid",
+	"NotProtected",
+	"ProtectionConfigured",
+	"SoftDeleted",
+	"ProtectionStopped",
+	"BackupSchedulesSuspended",
+	"RetentionSchedulesSuspended",
+)
+
+var retryableStates = set.Make(
```
theunrepentantgeek, line 45: You never use this map - did you mean to? Remove if not required.

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
```diff
@@ -0,0 +1,142 @@
...
+		// return success for reconcilation if the state is non-retryable
+		if nonRetryableStates.Contains(protectionStatus) {
+			log.V(Debug).Info("########################## Returning PostReconcileCheckResultSuccess ##########################")
+			return extensions.PostReconcileCheckResultSuccess(), nil
+		}
+
+		// call sync api only when protection status is ProtectionError and error code is usererror
+		if protectionStatus == "ProtectionError" {
+			protectionStatusErrorCode := ""
+			protectionStatusErrorCode = strings.ToLower(*backupInstance.Status.Properties.ProtectionStatus.ErrorDetails.Code)
```
theunrepentantgeek, line 86: No need to declare and then immediately overwrite, just initialize it directly.
```suggestion
			protectionStatusErrorCode := strings.ToLower(*backupInstance.Status.Properties.ProtectionStatus.ErrorDetails.Code)
```

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
```diff
@@ -0,0 +1,142 @@
...
+		backupInstance.Status.Properties != nil && backupInstance.Status.Properties.ProtectionStatus != nil &&
+		backupInstance.Status.Properties.ProtectionStatus.Status != nil {
+
+		protectionStatus := *backupInstance.Status.Properties.ProtectionStatus.Status
+		log.V(Debug).Info(fmt.Sprintf("########################## Protection Status is  %q ##########################", protectionStatus))
+
+		// return success for reconcilation if the state is non-retryable
+		if nonRetryableStates.Contains(protectionStatus) {
+			log.V(Debug).Info("########################## Returning PostReconcileCheckResultSuccess ##########################")
+			return extensions.PostReconcileCheckResultSuccess(), nil
```
theunrepentantgeek, line 80: This is where you should be calling the passed `extensions.PostReconcileCheckFunc`, but only if the resource is in a non-error condition - which I think means only if the state is `ProtectionConfigured`, but I'm not 100% on that.

v2/internal/testcommon/samples_tester.go
```diff
@@ -69,6 +69,9 @@ var exclusions = []string{
 
 	// Excluding cdn secret as it requires KV secrets
 	"secret",
+
+	// Exclude backupvaultsbackupinstance as it requires role assignments to be created after backup instance is created to make it land into protection configured state.
```
theunrepentantgeek, line 73: Is this still the case?

theunrepentantgeek, line 73: Is this still true? 

Your test in `crd_dataprotection_backupinstance_test.go` creates everything all at once, including the Role Assignment.

v2/internal/controllers/crd_dataprotection_backupinstance_test.go
```diff
@@ -0,0 +1,302 @@
...
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	tc.CreateResourcesAndWait(cluster, acct, blobService, blobContainer, backupVault, backupPolicy, extension,
+		trustedAccessRoleBinding, extenstionRoleAssignment, clusterRoleAssignment, clusterMSIRoleAssignment, snapshotRGRoleAssignment, backupInstance)
```
theunrepentantgeek, line 271: Our coding standard is to have all parameters on one line, or one per line, to make it easier to scan through the list.

```suggestion
	tc.CreateResourcesAndWait(
		cluster, 
		acct, 
		blobService, 
		blobContainer, 
		backupVault, 
		backupPolicy, 
		extension,
		trustedAccessRoleBinding, 
		extenstionRoleAssignment, 
		clusterRoleAssignment, 
		clusterMSIRoleAssignment, 
		snapshotRGRoleAssignment, 
		backupInstance)
```

theunrepentantgeek: It's good to have all of these in one `CreateResources()` call , ensuring that ASO correctly manages the dependencies.

v2/samples/dataprotection/v1api20231101/refs/v1api20200601_resourcegroup.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: resources.azure.com/v1api20200601
+kind: ResourceGroup
```
theunrepentantgeek, line 2: Shouldn't need to have a resource group explicitly given in the sample - the testing infrastructure creates one for you, I think. @super-harsh to confirm.

theunrepentantgeek: I think the standard name is `aso-sample-rg`, which will be redirected to the actual RG if necessary.

v2/samples/dataprotection/v1api20231101/refs/v1api20220401_clustersnapshotrgroleassignment.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: authorization.azure.com/v1api20220401
+kind: RoleAssignment
+metadata:
+  name: aso-sample-clustersnapshotrg
+  namespace: default
+spec:
+  # This resource can be owner by any resource. In this example we've chosen a resource group for simplicity
```
theunrepentantgeek, line 7:
```suggestion
  # This resource can be owned by any resource. In this example we've chosen a resource group for simplicity
```

v2/samples/dataprotection/v1api20231101/refs/v1api20220401_extensionsaroleassignment.yaml
```diff
@@ -0,0 +1,18 @@
...
+kind: RoleAssignment
+metadata:
+  name: aso-sample-extsa
+  namespace: default
+spec:
+  # This resource can be owner by any resource. In this example we've chosen a resource group for simplicity
+  owner:
+    name: asodpsamplekubestorage
+    group: storage.azure.com
+    kind: StorageAccount
```
theunrepentantgeek, line 11: Comment and `kind:` are inconsistent.

v2/samples/dataprotection/v1api20231101/refs/v1api20220401_vaultclusterroleassignment.yaml
```diff
@@ -0,0 +1,18 @@
...
+kind: RoleAssignment
+metadata:
+  name: aso-sample-vaultcluster
+  namespace: default
+spec:
+  # This resource can be owner by any resource. In this example we've chosen a resource group for simplicity
+  owner:
+    name: aso-sample-aks
+    group: containerservice.azure.com
+    kind: ManagedCluster
```
theunrepentantgeek, line 11: Commend and `kind:` are inconsistent.

v2/samples/dataprotection/v1api20231101/refs/v1api20220401_vaultsnapshotrgroleassignment.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: authorization.azure.com/v1api20220401
+kind: RoleAssignment
+metadata:
+  name: aso-sample-vaultsnapshotrg
+  namespace: default
+spec:
+  # This resource can be owner by any resource. In this example we've chosen a resource group for simplicity
```
theunrepentantgeek, line 7:
```suggestion
  # This resource can be owned by any resource. In this example we've chosen a resource group for simplicity
```

v2/samples/dataprotection/v1api20231101/v1api20231101_backupvaultsbackupinstance.yaml
```diff
@@ -0,0 +1,61 @@
+# Perform all the required Prerequisites before deploying backupvaultbackupinstnace yaml.
+# Prerequisites: https://learn.microsoft.com/en-us/azure/backup/azure-kubernetes-service-cluster-backup-concept
+# Permissions needed: https://learn.microsoft.com/en-us/azure/backup/azure-kubernetes-service-cluster-backup-concept#required-roles-and-permissions
```
theunrepentantgeek, line 3: How many of those prerequisites are already handled by the sample? 
If they're already handled, consider rephrasing as "Read here for more information about the prerequisites set up by this sample" instead of as a command requiring those steps first.

Also
* Spelling
* Azure standard is to omit the language from URL links so that it will auto-redirect to the users language, if available.

```suggestion
# Perform all the required Prerequisites before deploying backupvaultbackupinstance yaml.
# Prerequisites: https://learn.microsoft.com/azure/backup/azure-kubernetes-service-cluster-backup-concept
# Permissions needed: https://learn.microsoft.com/azure/backup/azure-kubernetes-service-cluster-backup-concept#required-roles-and-permissions
```

theunrepentantgeek (changes requested):
Looking good and mostly complete, but with a few changes required before we can consider merging. Also, check the CI builds to make sure those pass.

v2/azure-arm.yaml
```diff
@@ -1695,9 +1698,39 @@ objectModelConfiguration:
...
+      DatasourceSet:
+        ResourceID:
+          $armReference: true
+        ResourceUri:
+          $armReference: true
+      Datasource:
+        ResourceID:
+          $armReference: true
+        ResourceUri:
+          $armReference: true
```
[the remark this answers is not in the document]
matthchr: Yes.

Though I also wonder, does the user set both `ResourceUri` _and_ `ResourceId`? Only one is documented as required in the API. When must the user specify `ResourceUri` and what will it look like? Do we have an example?

Or is `ResourceUri` actually a derived field (and thus readonly?)

v2/internal/testcommon/samples_tester.go
super-harsh: I think yes, since in samples test suite, refs apply before the actual resources. Currently, we don't apply resources all at once. We have a tech debt tracking(#3091) this which I should be adding in into 2.7.0.

Maybe add a comment and mention the issue number?

v2/samples/dataprotection/v1api20231101/refs/v1api20200601_resourcegroup.yaml
super-harsh: Yep, We don't need any resource groups to be created in samples. Just use `aso-sample-rg` and that gets replaces with the "test generated resourceGroup"

v2/internal/controllers/crd_dataprotection_backupinstance_test.go
```diff
@@ -0,0 +1,264 @@
...
+			Owner:         tc.AsExtensionOwner(cluster),
+			Scope: &kubernetesconfiguration.Scope{
+				Cluster: &kubernetesconfiguration.ScopeCluster{
+					ReleaseNamespace: to.Ptr("dataprotection-microsoft"),
+				},
+			},
+			ConfigurationSettings: map[string]string{
+				"configuration.backupStorageLocation.bucket":                blobContainer.AzureName(),
+				"configuration.backupStorageLocation.config.resourceGroup":  rg.AzureName(),
+				"configuration.backupStorageLocation.config.storageAccount": acct.AzureName(),
```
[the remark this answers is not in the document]
mayankagg9722: We are using .Name instead of AzureName as we disucssed offline.

image

v2/internal/controllers/crd_dataprotection_backupinstance_test.go
```diff
@@ -0,0 +1,264 @@
...
+			ExtensionType: to.Ptr("microsoft.dataprotection.kubernetes"),
+			Owner:         tc.AsExtensionOwner(cluster),
+			Scope: &kubernetesconfiguration.Scope{
+				Cluster: &kubernetesconfiguration.ScopeCluster{
+					ReleaseNamespace: to.Ptr("dataprotection-microsoft"),
+				},
+			},
+			ConfigurationSettings: map[string]string{
+				"configuration.backupStorageLocation.bucket":                blobContainer.AzureName(),
+				"configuration.backupStorageLocation.config.resourceGroup":  rg.AzureName(),
```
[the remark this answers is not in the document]
mayankagg9722: image
we have discussed offline for this

v2/samples/dataprotection/v1api20231101/v1api20231101_backupvaultsbackupinstance.yaml
mayankagg9722: all samples are there in the directory. corrected terms and remove lang spec from url

v2/azure-arm.yaml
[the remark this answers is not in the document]
mayankagg9722: Yes, both are set by user and in our cases we have resouce id and resource uri both as same and its an arm resource.

v2/samples/dataprotection/v1api/v1api20230101_backupvaultsbackupinstance.yaml
mayankagg9722: Excluding sample recording for [backupvaultsbackupinstance.yaml] as role assignment creation is not working as expected in samples. Discussed and closed with harsh.

v2/azure-arm.yaml
```diff
@@ -1695,9 +1698,39 @@ objectModelConfiguration:
...
+        $supportedFrom: v2.2.0
+    2023-11-01: 
+      BackupVault: 
+        $export: true 
         $supportedFrom: v2.2.0 
+        $generatedConfigs:
+          PrincipalId: $.Status.Identity.PrincipalId
+      BackupVaults_BackupPolicy: 
+        $exportAs: BackupVaultsBackupPolicy 
+        $supportedFrom: v2.2.0
```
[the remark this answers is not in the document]
mayankagg9722: changed this to v2.7.0.

docs/hugo/content/reference/_index.md
```diff
@@ -257,10 +257,13 @@ To install the CRDs for these resources, your ASO configuration must include `da
...
+| Resource                                                                                                                                                                           | ARM Version | CRD Version   | Supported From | Sample                                                                                                                                                  |
+|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------|---------------|----------------|---------------------------------------------------------------------------------------------------------------------------------------------------------|
+| BackupVault                                                                                                                                                                        | 2023-11-01  | v1api20231101 | v2.2.0         | [View](https://github.com/Azure/azure-service-operator/tree/main/v2/samples/dataprotection/v1api20231101/v1api20231101_backupvault.yaml)                |
```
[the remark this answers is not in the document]
mayankagg9722: used v2.7.0.

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
```diff
@@ -0,0 +1,142 @@
...
+func (extension *BackupVaultsBackupInstanceExtension) PostReconcileCheck(
+	ctx context.Context,
+	obj genruntime.MetaObject,
+	owner genruntime.MetaObject,
+	_ *resolver.Resolver,
+	armClient *genericarmclient.GenericClient,
+	log logr.Logger,
+	_ extensions.PostReconcileCheckFunc,
+) (extensions.PostReconcileCheckResult, error) {
+	log.V(Debug).Info("########################## Starting Post-reconcilation for Backup Instance ##########################")
```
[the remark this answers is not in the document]
mayankagg9722: right, I am using it for testing purpose.

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
```diff
@@ -0,0 +1,142 @@
...
+	backupInstance, ok := obj.(*dataprotection.BackupVaultsBackupInstance)
+	if !ok {
+		return extensions.PostReconcileCheckResult{},
+			errors.Errorf("cannot run on unknown resource type %T, expected *dataprotection.BackupVaultsBackupInstance", obj)
+	}
+
+	var _ conversion.Hub = backupInstance
```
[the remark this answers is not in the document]
mayankagg9722: added comments on top of this line

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
```diff
@@ -0,0 +1,142 @@
...
+
+	if owner != nil && backupInstance != nil && backupInstance.Status.Id != nil &&
+		backupInstance.Status.Properties != nil && backupInstance.Status.Properties.ProtectionStatus != nil &&
+		backupInstance.Status.Properties.ProtectionStatus.Status != nil {
```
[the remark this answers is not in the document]
mayankagg9722: made changes to have one condition per line

v2/samples/dataprotection/v1api20231101/refs/v1api20200601_resourcegroup.yaml
mayankagg9722: No this is a different RG, needed while creation backup instance. This RG is known as "Snapshot RG" which is provded in Backup Instance payload. This RG is given explicitly by customer so that all the PV(volume) snapshots are stored within it.

v2/internal/testcommon/samples_tester.go
mayankagg9722: added issue number in the comment

v2/azure-arm.yaml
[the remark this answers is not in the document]
mayankagg9722: though resource uri is optional

v2/internal/controllers/crd_dataprotection_backupinstance_test.go
mayankagg9722: changes indentation

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
mayankagg9722: yes, we can remove this now. as I changed the impl.

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
mayankagg9722: NO, If I remove this Post reconciliation flow is not getting called. I have seen that behaviour in ASO and that why I am keeping it explicitly

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
mayankagg9722: actually I needed to move it above if condition

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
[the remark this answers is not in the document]
mayankagg9722: got it, made the necessary changes

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
mayankagg9722: changed

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
```diff
@@ -0,0 +1,142 @@
...
+					log.Error(err, "failed to create armdataprotection client")
+				}
+
+				var parameters armdataprotection.SyncBackupInstanceRequest
+				parameters.SyncType = to.Ptr(armdataprotection.SyncTypeDefault)
+				log.V(Debug).Info("########################## Starting BeginSyncBackupInstance for Backup Instance ##########################")
+				poller, err := clientFactory.NewBackupInstancesClient().BeginSyncBackupInstance(ctx, rg, vaultName, backupInstance.AzureName(), parameters, nil)
+				if err != nil {
+					log.Error(err, "failed BeginSyncBackupInstance to get the result")
+				}
```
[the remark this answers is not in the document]
mayankagg9722: sure, changed

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
mayankagg9722: I need to make these changes to put the url and reuse it to check the state

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
mayankagg9722: thats alredy been done as part of above if block

if protectionStatus == "ProtectionError"

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
mayankagg9722: I have verified it mutiple times to check on this in my local testing

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
mayankagg9722: I am not able to see any example for this. Please point me to the same.

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
mayankagg9722: I am not able to see any example for this or places where its been done. Please point me to the same.

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
mayankagg9722: sure, please check once

v2/azure-arm.yaml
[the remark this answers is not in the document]
mayankagg9722: verified from the swagger and team internally, only resourceId is the only arm reference

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
mayankagg9722: made the changes for these

mayankagg9722:
image
We need to increase the timeout for our test case as BI sync will take time

theunrepentantgeek:
> We need to increase the CI timeout for our test case as BI sync will take time. Otherwise its running fine and test recording is also included.

Why? CI builds run an accelerated replay of the recordings and don't interact with Azure at all.

mayankagg9722:
> of the recordings and don't interact with Azure at all.

Okay, sure. I will check once again if I am able to find anything useful from CI logs about this step failing.

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
theunrepentantgeek: The `next` parameter is provided to give the extension flexibility to augment or replace functionality as it chooses - it's a more powerful approach than providing a `PostReconcileCheck` that then always does the standard thing regardless.
 
The most usual approach is to handle special cases, then delegate to `next()` for the standard cases:

``` go
if specialCaseApplies() {
    // custom behaviour
} else {
    next()
}
```

Sometimes, an extension wants to do something after the standard behaviour has executed:

``` go
result := next()
if specialCaseApplies(result) {
    // Override the result
}

return result
```

Very very rarely, an extension wants to completely take over and skip the default behaviour. 

``` go
// Don't call 
// next()
SpecialCase()
```

In this last case, the extension author is stating "I know exactly what to do here, this will always be correct, _even if the default behaviour changes in the future_".

There are cases where this is true. We have some in ASO. But they aren't a precedent for other extensions skipping next().

Unless you have a really compelling reason otherwise, you should always call `next()`. 

I have a TODO task to write up more information about the extensions, and this guidance will be a significant part of it.

theunrepentantgeek (commented):
This is getting close to complete; once you've addressed the last couple of comments and rerecorded the test (after merging main, as discussed offline), we should be able to merge.

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
mayankagg9722: Thanks, I am get this removed

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
mayankagg9722: using this below now

v2/samples/dataprotection/v1api20231101/refs/v1api20230501_extension.yaml
```diff
@@ -0,0 +1,26 @@
...
+    kind: ManagedCluster
+    name: aso-sample-aks
+  scope:
+    cluster:
+      releaseNamespace: dataprotection-microsoft
+  configurationSettings:
+    configuration.backupStorageLocation.bucket: "asodpsamplecontainer"
+    configuration.backupStorageLocation.config.resourceGroup: "aso-sample-rg"
+    configuration.backupStorageLocation.config.storageAccount: "asodpsamplekubestorage"
+    configuration.backupStorageLocation.config.subscriptionId: "f0c630e0-2995-4853-b056-0b3c09cb673f"
```
super-harsh, line 25: replace with `00000000-0000-0000-0000-000000000000`

v2/samples/dataprotection/v1api20231101/refs/v1api20230501_extension.yaml
```diff
@@ -0,0 +1,26 @@
...
+    credentials.tenantId: "72f988bf-86f1-41af-91ab-2d7cd011db47"
```
super-harsh, line 26: replace with `00000000-0000-0000-0000-000000000000`

v2/samples/dataprotection/v1api20231101/v1api20231101_backupvaultsbackupinstance.yaml
```diff
@@ -0,0 +1,55 @@
...
+        name: aso-sample-aks
+        group: containerservice.azure.com
+        kind: ManagedCluster
+    dataSourceSetInfo:
+      objectType: DatasourceSet
+      datasourceType: Microsoft.ContainerService/managedClusters
+      resourceName: aso-sample-aks
+      resourceLocation: westus2
+      resourceType: Microsoft.ContainerService/managedClusters
+      resourceUri: /subscriptions/f0c630e0-2995-4853-b056-0b3c09cb673f/resourceGroups/aso-sample-rg/providers/Microsoft.ContainerService/managedClusters/aso-sample-aks
```
super-harsh, line 32: replace the guid with `00000000-0000-0000-0000-000000000000`

v2/samples/dataprotection/v1api20231101/v1api20231101_backupvaultsbackupinstance.yaml
```diff
@@ -0,0 +1,55 @@
...
+  properties:
+    objectType: BackupInstance
+    friendlyName: aso-sample-backupinstance
+    dataSourceInfo:
+      objectType: Datasource
+      datasourceType: Microsoft.ContainerService/managedClusters
+      resourceName: aso-sample-aks
+      resourceLocation: westus2
+      resourceType: Microsoft.ContainerService/managedClusters
+      resourceUri: /subscriptions/f0c630e0-2995-4853-b056-0b3c09cb673f/resourceGroups/aso-sample-rg/providers/Microsoft.ContainerService/managedClusters/aso-sample-aks
```
super-harsh, line 21: replace the guid with `00000000-0000-0000-0000-000000000000`

super-harsh: Result should be `/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/aso-sample-rg/providers/Microsoft.ContainerService/managedClusters/aso-sample-aks`

mayankagg9722: changed this

v2/samples/dataprotection/v1api20231101/v1api20231101_backupvaultsbackupinstance.yaml
mayankagg9722: changed this

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
```diff
@@ -0,0 +1,158 @@
...
+					_, err := poller.Result(ctx)
+					if err != nil {
+						return extensions.PostReconcileCheckResultFailure("couldn't create PUT resume token for resource"), err
+					}
+				}
+				log.V(Debug).Info("Polling is in-progress")
+			}
+		}
+	}
+	return extensions.PostReconcileCheckResultFailure("Backup Instance is in non terminal state"), nil
```
super-harsh, line 157: I can't see anywhere that `PostReconcileCheck` is passing?

v2/samples/dataprotection/v1api20231101/refs/v1api20200601_resourcegroup.yaml
super-harsh, line 2: This should be deleted, and everything should be in the test RG.

matthchr:
This fixes #2345

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
mayankagg9722: // return success for reconcilation if the state is non-retryable
		if nonRetryableStates.Contains(protectionStatus) {
			log.V(Debug).Info("Returning PostReconcileCheckResultSuccess")
			return next(ctx, obj, owner, resolver, armClient, log)
		}

v2/internal/controllers/crd_dataprotection_backupinstance_20231101_test.go
```diff
@@ -0,0 +1,314 @@
...
+				Key:  backupVaultPrincipalIdKey,
+			},
+			RoleDefinitionReference: &genruntime.ResourceReference{
+				ARMID: fmt.Sprintf("/subscriptions/%s/providers/Microsoft.Authorization/roleDefinitions/acdd72a7-3385-48ef-bd42-f606fba81ae7", tc.AzureSubscription), // This is Reader Role
+			},
+		},
+	}
+
+	// give cluster msi access over snapshot rg for pv creation
+	clusterMSIRoleAssignmentAssignmentGUID, err := tc.Namer.GenerateUUID()
```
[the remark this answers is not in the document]
mayankagg9722: Thanks for the suggestions will make this improvement.

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
super-harsh: Yep, I saw that. My concern was mainly, what happens when `protectionStatus` is a success?

super-harsh: Reconciler would still do a postReconcileCheck, and be always falling on line 157?

v2/api/dataprotection/customizations/backup_vaults_backup_instance_extensions.go
```diff
@@ -0,0 +1,158 @@
...
+		backupInstance.Status.Id != nil &&
+		backupInstance.Status.Properties != nil &&
+		backupInstance.Status.Properties.ProtectionStatus != nil &&
+		backupInstance.Status.Properties.ProtectionStatus.Status != nil {
+
+		protectionStatus := *backupInstance.Status.Properties.ProtectionStatus.Status
+		log.V(Debug).Info(fmt.Sprintf("Protection Status is  %q", protectionStatus))
+
+		// return success for reconcilation if the state is non-retryable
+		if nonRetryableStates.Contains(protectionStatus) {
```
super-harsh, line 92: Would suggest inversing this logic here. We can always return success, if` protectionStatus != "ProtectionError"`

super-harsh: that way, we're not checking the list each time during the reconcile.

theunrepentantgeek (approved):

PR #3738, opened 2024-01-24
Delete the Experimental release and then re-upload it

matthchr:


.github/workflows/create-release-experimental.yml
```diff
@@ -53,7 +59,6 @@ jobs:
           file: "v2/out/release/*"
           file_glob: true
           prerelease: true
-          overwrite: true
```
theunrepentantgeek, line 56: Does `overwrite` not work?

theunrepentantgeek (approved):

.github/workflows/create-release-experimental.yml
matthchr: It works but should only be needed if the release already exists. Since I'm deleting it now it won't exist and we don't need overwrite.

The issue is that overwrite will overwrite the release binaries but it doesn't update the release date, which means that it looks as if the release was published N days ago when it was really updated yesterday.

I think we want the date on the release to match the time the files were updated, and it seems the only way to do that is to delete and republish the release.

PR #3749, opened 2024-01-29
Make crdPattern a required Helm value

matthchr:


theunrepentantgeek (approved):

matthchr:
Our tests keeping me honest - I forgot that in _existing_ clusters, `crdPattern` isn't required because we also source from the set of installed CRDs.

So this change is problematic in those cases... going to abandon it for now at least...

PR #3752, opened 2024-01-31
Add ConfigurationProtectedSettings to extensions resource

matthchr:
**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/api/kubernetesconfiguration/v1api20230501/structure.txt
+++ b/v2/api/kubernetesconfiguration/v1api20230501/structure.txt
@@ -3,13 +3,14 @@ github.com/Azure/azure-service-operator/v2/api/kubernetesconfiguration/v1api2023
 ├── APIVersion: Enum (1 value)
 │   └── "2023-05-01"
 ├── Extension: Resource
-│   ├── Spec: Object (12 properties)
+│   ├── Spec: Object (13 properties)
 │   │   ├── AksAssignedIdentity: *Object (1 property)
 │   │   │   └── Type: *Enum (2 values)
 │   │   │       ├── "SystemAssigned"
 │   │   │       └── "UserAssigned"
 │   │   ├── AutoUpgradeMinorVersion: *bool
 │   │   ├── AzureName: string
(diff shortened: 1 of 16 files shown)


super-harsh (approved):

v2/api/kubernetesconfiguration/v1api20230501/structure.txt
```diff
@@ -53,6 +54,7 @@ github.com/Azure/azure-service-operator/v2/api/kubernetesconfiguration/v1api2023
 │       │       └── "UserAssigned"
 │       ├── AutoUpgradeMinorVersion: *bool
 │       ├── Conditions: conditions.Condition[]
+│       ├── ConfigurationProtectedSettings: map[string]string
```
theunrepentantgeek, line 57: Should this exist on the Status? Won't it reveal secrets?

matthchr: Comment in add_secrets.go, previous PR explains this:
```
// We only remove pure secret references here. For the case of secret maps, different services seem to treat them
// differently. Some services (such as Microsoft.KubernetesConfiguration/extensions) will return the keys of the map
// but not the values. Other services (such as APIM) will return certain keys and values that it knows are non-secret, but
// redact the ones that are secret. Since it's hard to know statically what will be returned for any given service, we
// default to having the map[string]string on the Status type and letting the service return what it wants.
```

Basically the problem here is that unlike pure string secrets, not every service has the same behavior. They all redact things but they all do it differently. It _seems_ like there's much less of a standard here for questions like: "Are the keys secret too" or "is _everything_ in this map secret or only some stuff" - so rather than try to guess at the services intent we just let them return what they want and display it.

The reasoning behind this (and feel free to disagree w/ this if you think it's wrong) is to let the service show us what it wants to show us. If the service is returning something on GET which _is_ a secret that's a service bug and would impact more than just ASO (as SDKs, etc also always have these fields, we only remove them in ASO as a convenience because we know they'll always be empty in the case of single-string secrets).

theunrepentantgeek (approved):

PR #3753, opened 2024-01-31
Improve resiliency of recorded tests

### LINKED ISSUE #3112: Modify Test Recordings

We've identified problems with the recordings used to verify ASO behaviour when testing our samples. 

Because of the goal-seeking nature of ASO, test replay will sometimes make an additional request that can't be satisfied from the recording, resulting in test failure. This happens primarily when a property is being populated from a configMap.

**Describe the improvement**

We're going to modify our test recording to make it more tolerant of these extra requests.

Currently we group HTTP requests & responses by the requested URL, with different requests to the same URL disambiguated by a sequence number.

The grouping will be extended to _also_ use a hash of the request body, and we're going to allow replays to satisfy a request that is just past the end of the expected sequence by returning the last item from the sequence.

theunrepentantgeek:

Adds resiliency to our recorded tests to counter the non-deterministic way the ASO runs when handling multiple resources. We do this by allowing extra replay of selected requests.

* The last PUT to a specific URL may be repeated one extra time
* The last GET of a specific URL may be repeated multiple times

The combination of these should allow an extra reconciliation loop for resources in goal state, while allowing other resources to progress.

Background: We've been experiencing a low level of flaky tests for some time, where the test will fail because the operator issues an extra PUT or GET for a resource. These happen when an extra reconciliation occurs for one resource that has alaready reached goal state, but the test is continuing to run because another resource is still processing.

Closes #3112 

Also includes an upgrade to version 3 of `go-vcr`. This version includes a change to the recording format; to avoid lots of churn, the version of each recording is checked and v1 of `go-vcr` will be used for older recordings. 

Only recordings made with `go-vcr` v3 will have the extra resiliency included during playback.

One change in `go-vcr` v3 is that it will always produce a recording file, so no need to manually create an empty file when nothing happens.

gif

**If applicable**:
- [x] this PR contains tests


matthchr:
> The last PUT to a specific URL may be repeated one extra time
> The last GET of a specific URL may be repeated multiple times

Any specific reason that you've chosen this breakdown (GETs are allowed multiple times but PUTs are allowed only 1x extra?) In practice I think the 1x extra will help quite a bit, and it's not like the tests are extremely flaky now, but strictly speaking there's nothing preventing more than 1x extra PUT in a row, although I don't think I've ever seen it.

v2/internal/testcommon/replay_roundtripper.go
```diff
@@ -0,0 +1,117 @@
...
+		if !errors.Is(err, cassette.ErrInteractionNotFound) {
+			return response, err
+		}
+
+		replayer.padlock.Lock()
+		defer replayer.padlock.Unlock()
+
+		// We didn't find an interaction, see if we have a cached response to return
+		if cachedResponse, ok := replayer.puts[request.URL.String()]; ok {
+			delete(replayer.puts, request.URL.String())
```
matthchr, line 102: minor: leave a comment on this line to the effect of "only allowing one duplicate PUT"

v2/internal/testcommon/test_context.go
```diff
@@ -48,7 +43,7 @@ type PerTestContext struct {
 	TestContext
 	T                     *testing.T
 	logger                logr.Logger
-	AzureClientRecorder   *recorder.Recorder
+	AzureClientRecorder   testRecorder
```
matthchr, line 46: Exported field whose type is unexported seenms wrong here?

v2/internal/testcommon/test_context.go
```diff
@@ -159,11 +154,11 @@ func (tc TestContext) ForTest(t *testing.T, cfg config.Values) (PerTestContext,
 		Namer:                 namer,
 		NoSpaceNamer:          namer.WithSeparator(""),
 		AzureClient:           globalARMClient,
-		AzureSubscription:     details.ids.subscriptionID,
-		AzureTenant:           details.ids.tenantID,
-		AzureBillingInvoiceID: details.ids.billingInvoiceID,
+		AzureSubscription:     details.Ids().subscriptionID,
```
matthchr, line 157: minor: I think should be `.IDs`?

v2/internal/testcommon/test_replayer_v1.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// IsReplaying returns true if we're replaying a recorded test, false if we're recording a new test
+func (r playerDetailsV1) IsReplaying() bool {
+	return r.recorder.Mode() == recorder.ModeReplaying
+}
+
+// CreateClient creates an HTTP client configured to record or replay HTTP requests.
+// t is a reference to the test currently executing.
+// TODO: Remove the reference to t to reduce coupling
```
matthchr, line 131: Is this something you're planning on doing in this PR?

v2/internal/testcommon/test_recorder.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	// If a cassette file exists in the old format, use the old player
+	v1Exists, err := cassetteFileV1Exists(cassetteName)
+	if err != nil {
+		return nil, errors.Wrapf(err, "checking existence of cassette %s", cassetteName)
+	}
+
+	if v1Exists {
+		return newTestPlayerV1(cassetteName, cfg)
```
matthchr, line 60: minor: Should this be `newTestReplayerV1`?

v2/internal/testcommon/test_recorder_v3.go
```diff
@@ -0,0 +1,239 @@
...
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-service-operator/v2/internal/config"
+	"github.com/Azure/azure-service-operator/v2/internal/genericarmclient"
+	"github.com/go-logr/logr"
+	"github.com/google/uuid"
+	"github.com/pkg/errors"
+
+	"gopkg.in/dnaeon/go-vcr.v3/cassette"
+	"gopkg.in/dnaeon/go-vcr.v3/recorder"
```
matthchr, line 23: minor: import ordering (ASO stuff should be at the bottom, go-vcr stuff should be in the middle)

v2/internal/testcommon/test_recorder_v3.go
```diff
@@ -0,0 +1,239 @@
...
+			panic(err)
+		}
+
+		r.Body = io.NopCloser(&b)
+		return b.String() == "" || hideRecordingData(b.String()) == i.Body
+	})
+
+	r.AddHook(redactRecording(azureIDs), recorder.BeforeSaveHook)
+
+	return recorderDetailsV3{
```
matthchr, line 114: minor: Wondering if, given this method is named `new...` and `recorderDetailsV3` contains various ptr-y things, we should return `&recorderDetailsV3` here, and have the receiver be on ptr too?

v2/internal/testcommon/test_replayer_v1.go
```diff
@@ -0,0 +1,136 @@
...
+		var b bytes.Buffer
+		if _, err := b.ReadFrom(r.Body); err != nil {
+			panic(err)
+		}
+
+		r.Body = io.NopCloser(&b)
+		return b.String() == "" || hideRecordingData(b.String()) == i.Body
+	})
+
+	return playerDetailsV1{
```
matthchr, line 95: same ptr question

v2/go.mod
```diff
@@ -95,6 +95,7 @@ require (
 	gomodules.xyz/jsonpatch/v2 v2.4.0 // indirect
 	google.golang.org/appengine v1.6.8 // indirect
 	google.golang.org/protobuf v1.32.0 // indirect
+	gopkg.in/dnaeon/go-vcr.v3 v3.1.2 // indirect
```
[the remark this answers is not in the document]
theunrepentantgeek: I don't think it should be indirect, that's odd. Will fix.

v2/internal/controllers/recordings/TestOperatorMode_Watchers.yaml
```diff
@@ -0,0 +1,3 @@
+---
+version: 2
```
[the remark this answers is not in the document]
theunrepentantgeek: It's the version of the file format, not the version of `go-vcr`. 

File format `2` is used by `go-vcr` v3 and higher.

v2/internal/testcommon/error_translating_roundtripper.go
```diff
@@ -12,12 +12,11 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/dnaeon/go-vcr/cassette"
-	"github.com/dnaeon/go-vcr/recorder"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/conditions"
 	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
 
-	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/conditions"
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

v2/internal/testcommon/error_translating_roundtripper_v1.go
```diff
@@ -0,0 +1,126 @@
+/*
+Copyright (c) Microsoft Corporation.
```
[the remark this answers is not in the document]
theunrepentantgeek: I'll give this a go.

v2/internal/testcommon/error_translating_roundtripper_v1.go
```diff
@@ -0,0 +1,126 @@
...
+package testcommon
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"testing"
+
+	cassettev1 "github.com/dnaeon/go-vcr/cassette"
```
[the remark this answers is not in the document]
theunrepentantgeek: It's about half of what we had already - the playback part only.

v2/internal/testcommon/error_translating_roundtripper_v1.go
```diff
@@ -0,0 +1,126 @@
...
+// Ideally we would panic on this error but we don't have a good way to deal with the following
+// problem at the moment:
+//   - during record the controller does GET (404), PUT, … GET (OK)
+//   - during playback the controller does GET (which now returns OK), DELETE, PUT, …
+//     and fails due to a missing DELETE recording
+func translateErrorsV1(r http.RoundTripper, cassetteName string, t *testing.T) http.RoundTripper {
+	return errorTranslationV1{r, cassetteName, nil, t}
+}
+
+type errorTranslationV1 struct {
```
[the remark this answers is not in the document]
theunrepentantgeek: I'll add some notes.

v2/internal/testcommon/kube_test_context_envtest.go
```diff
@@ -14,14 +14,14 @@ import (
...
 	"runtime"
-	"sigs.k8s.io/controller-runtime/pkg/metrics/server"
-	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"strings"
 	"sync"
 	"time"
 
+	"sigs.k8s.io/controller-runtime/pkg/metrics/server"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

v2/internal/testcommon/replay_roundtripper.go
```diff
@@ -0,0 +1,117 @@
...
+	// For other kinds of request, just pass through to the inner round tripper.
+	return replayer.inner.RoundTrip(request)
+}
+
+func (replayer *replayRoundTripper) roundTripGet(request *http.Request) (*http.Response, error) {
+	// First use our inner round tripper to get the response.
+	response, err := replayer.inner.RoundTrip(request)
+	if err != nil {
+		// We have an error - return it, unless it's from go-vcr
+		if !errors.Is(err, cassette.ErrInteractionNotFound) {
```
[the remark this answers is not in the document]
theunrepentantgeek: Count's are independent for GET vs PUT for the same URL (see the initialization of the variable `key` on line 38 of  `counting_roundtripper.go`).

That said, the `COUNT` header is irrelevant because this only triggers once we fall off the end of the recording - it doesn't matter that a subsequent GET specifies COUNT+1 because there are no more recorded responses for that combination of method and URL.

The sequence of requests from a single reconciler is stable, which is why we didn't have problems doing replay of simple resources, only when we got to more complex scenarios where multiple reconcilers were running in parallel.

Different reconcilers are handling different resources, which have different URLs, so I don't think the scenario you described above can happen.

We'd actually have the following:

1. PUT (COUNT=0) for create, served from the recording file
2. GET (COUNT=0) for monitoring the long running operation
3. GET (COUNT=1...n) for monitoring the LRO until it completes
4. PUT (COUNT=1) for update, served from the recording file
5. GET (COUNT=0) for monitoring a new long running operation
6. GET (COUNT=1...n) for monitoring the LRO until it completes
7. PUT (COUNT=2) for extra update, no available recording, replay from COUNT=1, returning the same LRO
8. GET (COUNT=n+1) for monitoring the LRO, no available recording, replay from COUNT=n, indicating the LRO has completed

I can add the specHash if that's needed (I have it in another working branch) but I'm not sure it is.

Have I covered off your concerns, or is there something I need to address?

v2/internal/testcommon/replay_roundtripper.go
theunrepentantgeek: Done.

v2/internal/testcommon/test_context.go
theunrepentantgeek: Will fix.

v2/internal/testcommon/test_context.go
theunrepentantgeek: Agreed. Will fix.

v2/internal/testcommon/test_recorder.go
theunrepentantgeek: No, because the v1 version doesn't do recording, only playback.

v2/internal/testcommon/test_recorder_v3.go
theunrepentantgeek: Fixed.

v2/internal/testcommon/test_recorder_v3.go
theunrepentantgeek: Yup.

v2/internal/testcommon/test_replayer_v1.go
theunrepentantgeek: Same _Yup_ answer.

v2/internal/testcommon/test_replayer_v1.go
theunrepentantgeek: I should, yes.

v2/internal/testcommon/test_recorder.go
matthchr: I missed this at the time but realized it later in the same review, but forgot to go back and delete this comment!

v2/internal/testcommon/replay_roundtripper.go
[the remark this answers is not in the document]
matthchr: > Different reconcilers are handling different resources, which have different URLs, so I don't think the scenario you described above can happen.

The scenario I described above is for the same resource, same name, same type, where the first PUT is a create, and the second PUT is an update.

Imagine I have test code like this:
```
// create my StorageAccount
acct := &storage.StorageAccount{
	ObjectMeta: tc.MakeObjectMetaWithName(tc.NoSpaceNamer.GenerateName("stor")),
	Spec: storage.StorageAccount_Spec{
		Location: tc.AzureRegion,
		Owner:    testcommon.AsOwner(rg),
		Kind:     to.Ptr(storage.StorageAccount_Kind_Spec_StorageV2),
		Sku: &storage.Sku{
			Name: to.Ptr(storage.SkuName_Standard_LRS),
		},
		// TODO: They mark this property as optional but actually it is required
		AccessTier: to.Ptr(storage.StorageAccountPropertiesCreateParameters_AccessTier_Hot),
	},
}

// Triggers duplicate reconcile during recordings 5% of the time and consumes 2 PUTs.
// The second PUT is for the update request below, and the LRO returned is for that as well, 
// so at the end of the LRO the acct variable 5% of the time will have the second LROs response (after update)
// and 95% of the time will have the first LROs response (before update).
tc.CreateResourceAndWait(acct)

// This assert will fail 5% of the time
tc.Expect(acct.Status.AccessTier).To(Equal(to.Ptr(storage.StorageAccountPropertiesCreateParameters_STATUS_AccessTier_Hot)))

old := acct.DeepCopy()
acct.Spec.AccessTier = to.Ptr(storage.StorageAccountPropertiesCreateParameters_AccessTier_Cold),
tc.PatchResourceAndWait(old, &acct)

tc.Expect(acct.Status.AccessTier).To(Equal(to.Ptr(storage.StorageAccountPropertiesCreateParameters_STATUS_AccessTier_Hot)))
```

You're right that I forgot about LROs but I don't think it saves us from the above?

Now, I don't know how often that the tests we have today actually do this pattern of:
```
// create
// assert propA has shape 1
// update
// assert propA has shape 2
```

but I do think there are a few, and it would be easy to write more that (mostly) passed but flaked occasionally.

v2/go.mod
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

v2/internal/testcommon/error_translating_roundtripper_v1.go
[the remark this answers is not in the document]
theunrepentantgeek: This worked out well, though there were some cascading effects. I used the package name `vcr` to avoid clashing with existing names like `recorder` and `record`.

v2/internal/testcommon/error_translating_roundtripper_v1.go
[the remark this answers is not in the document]
theunrepentantgeek: Added a README to the new `vcr/v1` folder.

v2/internal/testcommon/test_context.go
theunrepentantgeek: Fixed by introduction of subpackages.

v2/internal/testcommon/test_replayer_v1.go
theunrepentantgeek: I changed my mind; keeping the flexibility of using the standard library instead of passing a string.

v2/internal/testcommon/test_recorder_v3.go
theunrepentantgeek: Done.

v2/internal/testcommon/test_replayer_v1.go
theunrepentantgeek: Done.

v2/internal/testcommon/vcr/redact.go
```diff
@@ -0,0 +1,124 @@
...
+*/
+
+package vcr
+
+import (
+	"net/http"
+	"regexp"
+)
+
+var RequestHeadersToRemove = []string{
```
matthchr, line 13: minor: Do these need to be exported? (and below w/ `ResponseHeadersToRemove`)?

It means external callers could mutate them which seems ick. AFAIK we don't want/need them to be externally configurable as they're an implementation detail of the `vcr` package? Exporting `RedactRequestHeaders` is enough?

v2/internal/testcommon/vcr/v1/test_replayer.go
```diff
@@ -0,0 +1,138 @@
...
+		var b bytes.Buffer
+		if _, err := b.ReadFrom(r.Body); err != nil {
+			panic(err)
+		}
+
+		r.Body = io.NopCloser(&b)
+		return b.String() == "" || vcr.HideRecordingData(b.String()) == i.Body
+	})
+
+	return &playerDetails{
```
matthchr, line 97: minor: wonder if this could just be called `player` rather than `playerDetails`?

v2/internal/testcommon/vcr/v3/replay_roundtripper.go
```diff
@@ -0,0 +1,153 @@
...
+
+	// We have a response, cache it and return it
+	replayer.gets[request.URL.String()] = response
+	return response, nil
+}
+
+func (replayer *replayRoundTripper) roundTripPut(request *http.Request) (*http.Response, error) {
+	// Calculate a hash of the request body to use as a cache key
+	// We need this whether we are updating our cache or replaying
+	hash := replayer.hashOfPUTBody(request)
```
matthchr, line 100: minor: be consistent in how you're casing `PUT`.

I think it's fine to call it `Put` here

v2/internal/testcommon/vcr/v3/replay_roundtripper.go
```diff
@@ -0,0 +1,153 @@
...
+	defer replayer.padlock.Unlock()
+
+	// We have a response, cache it and return it
+	replayer.puts[hash] = response
+	return response, nil
+}
+
+// hashOfPUTBody calculates a hash of the body of a PUT request, for use as a cache key.
+// The body is santised before calculating the hash to ensure that the same request body always results in the same hash.
+func (replayer *replayRoundTripper) hashOfPUTBody(request *http.Request) string {
```
matthchr, line 134: You could possibly just call this `hashOfBody`

v2/internal/testcommon/vcr/v3/replay_roundtripper_test.go
```diff
@@ -0,0 +1,99 @@
...
+	g.Expect(resp.StatusCode).To(Equal(200))
+
+	// Assert - third request works by replaying the first
+	//nolint:bodyclose // there's no actual body in this response to close
+	resp, err = replayer.RoundTrip(req)
+	g.Expect(err).ToNot(HaveOccurred())
+	g.Expect(resp.StatusCode).To(Equal(200))
+}
+
+func TestReplayRoundTripperRoundTrip_GivenSinglePUT_ReturnsOnceExtra(t *testing.T) {
```
matthchr, line 62: Should we have a test that ensures that the hash-based lookup works?

Specifically, something where there are 2 puts w/ different hashes and the first one fails, assert we get the first ones response replayed (not the second), and then another PUT which should give the second response?

matthchr: Though actually now that I re-read how the PUT logic you have works, I am not actually sure such a test will pass because the hash isn't tightly integrated into the `replayer.inner.RoundTrip` - I think it needs to be for the above to work?

Happy to make this comment nonblocking though and we can come around again and fix that issue? wdyt?

v2/internal/testcommon/vcr/v3/test_recorder.go
```diff
@@ -0,0 +1,233 @@
...
+	"gopkg.in/dnaeon/go-vcr.v3/recorder"
+
+	"github.com/Azure/azure-service-operator/v2/internal/config"
+	"github.com/Azure/azure-service-operator/v2/internal/genericarmclient"
+	"github.com/Azure/azure-service-operator/v2/internal/testcommon/creds"
+	"github.com/Azure/azure-service-operator/v2/internal/testcommon/vcr"
+)
+
+// recorderDetails is an implementation of testRecorder using go-vcr v3.
+type recorderDetails struct {
```
matthchr, line 29: same comment as the v1 version - should this be `recorder`?

v2/internal/testcommon/vcr/v3/test_recorder.go
```diff
@@ -0,0 +1,233 @@
...
+// any subscription, so a contributor can update the tests against their own sub.
+func redactRecording(
+	azureIDs creds.AzureIDs,
+) recorder.HookFunc {
+	hide := func(s string, id string, replacement string) string {
+		return strings.ReplaceAll(s, id, replacement)
+	}
+
+	return func(i *cassette.Interaction) error {
+
```
matthchr, line 138: minor: Excess newline?

v2/internal/testcommon/vcr/v3/test_recorder_test.go
```diff
@@ -0,0 +1,127 @@
...
+	. "github.com/onsi/gomega"
+
+	"github.com/Azure/azure-service-operator/v2/internal/config"
+	"github.com/Azure/azure-service-operator/v2/internal/testcommon/vcr"
+)
+
+//nolint:paralleltest
+func TestRecorder_WhenRecordingAndRecordingDoesNotExist_MakesRecording(t *testing.T) {
+	// NB: Can't run tests using Setenv() in parallel
+	t.Setenv("AZURE_SUBSCRIPTION_ID", "00000000-0000-0000-0000-000000000000")
```
matthchr, line 25: Where are these env variables being read? If it's from `config.Values` someplace (I don't see exactly where...) can we not just pass dummy values into the `config.Values` struct directly? That would allow this test to be run in parallel and avoid this awkward env variable mutation.

v2/internal/testcommon/vcr/v1/test_replayer.go
theunrepentantgeek: Yes, that works.

v2/internal/testcommon/vcr/v3/replay_roundtripper.go
theunrepentantgeek: Renamed.

v2/internal/testcommon/vcr/v3/replay_roundtripper.go
theunrepentantgeek: Good idea.

v2/internal/testcommon/vcr/v3/replay_roundtripper_test.go
theunrepentantgeek: I added another test, and I think the functionality is working as you desire. Let me know and I can address in a future PR if not.

v2/internal/testcommon/vcr/v3/test_recorder.go
theunrepentantgeek: Can't - name collision with the imported go-vcr package.

v2/internal/testcommon/vcr/v3/test_recorder.go
theunrepentantgeek: Yup.

v2/internal/testcommon/vcr/v3/test_recorder_test.go
theunrepentantgeek: There's a check for these environment variables being present that's triggered when the recording file doesn't exist. I couldn't see an easy way to suppress, so I took this approach. Can revisit in a future PR.

v2/internal/testcommon/vcr/v3/replay_roundtripper_test.go
matthchr: What I was thinking we wanted was this:

- PUT1 (served from recording)
- PUT2 (served from hash-based cache)
- PUT3 (served from recording)
- PUT4 (served from recording)

and what we have is

- PUT1 (served from recording)
- PUT2 (served from hash-based cache)
- PUT3 (served from hash-based cache)
- PUT4 (served from hash-based cache)

The difference being that once the first miss happens we're in the "served from hash-based cache" for the rest of that PUT. BUT, you're right that what we're doing actually meshes with what you said we'd do which is allow 1 miss per PUT-URL (not per PUT-URL-Hash).

I've approved this PR - I had missed the fact that even though we're playing the other hashes from the cache after the first miss it will work. If you feel that this option is valuable to note as a future change if needed, feel free to leave a TODO or non-TODO comment in the roundtripper that alludes to this as an option if we end up needing it. I suspect we will not end up needing it as we don't see that many flakes now so even the limit of 1 per URL is likely gonna be OK.

matthchr (approved):
Thanks for doing this!

v2/internal/testcommon/vcr/v3/replay_roundtripper_test.go
theunrepentantgeek: > BUT, you're right that what we're doing actually meshes with what you said we'd do which is allow 1 miss per PUT-URL (not per PUT-URL-Hash).

What we're doing is allowing one miss per PUT body hash, not per URL (that changed when we switched to using body-hash as the key).

So the scenario you describe will work:

* PUT 1 - body1 (response served from recording)
* PUT 2 - body1 (response served from hash-based cache)
* PUT 3 - body2 (served from recording)
* PUT 4 - body2 (served from recording)

PR #3763, opened 2024-02-06
Allow samples with a hardcoded secret rather than an autogenerated one

matthchr:
* This has the advantage of the sample being clearer to users, although it does involve having secrets committed in the repo. I think this is OK for two reasons:
   * The resources in question aren't durably created and exist for sample purposes only.
   * For SecretMapReference, it will not be possible for the test infrastructure to generate the secret, as it doesn't understand the expected payload of the map (what keys should be set).

As part of this change, slightly harden the test helper functions to do the right thing if passed a non-ASO resource (so they don't try to monitor it's Ready condition)

Note: This just enables the capability in the framework, we can still choose to ask PRs to mostly use the secret-autogeneration function if we want to.

gif

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/v2/internal/testcommon/kube_per_test_context.go
+++ b/v2/internal/testcommon/kube_per_test_context.go
@@ -411,6 +411,11 @@ func (tc *KubePerTestContext) CreateResourceAndWait(obj client.Object) {
 
 	gen := obj.GetGeneration()
 	tc.CreateResource(obj)
+	// Only wait for ASO objects
+	// TODO: Consider making tc.Match.BeProvisioned(0) work for non-ASO objects with a Ready condition too?
+	if _, ok := obj.(genruntime.MetaObject); !ok {
+		return
+	}
 	tc.Eventually(obj).Should(tc.Match.BeProvisioned(gen))
 }
 
@@ -420,6 +425,10 @@ func (tc *KubePerTestContext) CreateResourceAndWaitWithoutCleanup(obj client.Obj
 	tc.T.Helper()
 	gen := obj.GetGeneration()
 	tc.CreateResourceUntracked(obj)
+	// Only wait for ASO objects
+	if _, ok := obj.(genruntime.MetaObject); !ok {
+		return
+	}
 	tc.Eventually(obj).Should(tc.Match.BeProvisioned(gen))
 }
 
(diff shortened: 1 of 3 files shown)


v2/internal/testcommon/kube_per_test_context.go
```diff
@@ -411,6 +411,10 @@ func (tc *KubePerTestContext) CreateResourceAndWait(obj client.Object) {
 
 	gen := obj.GetGeneration()
 	tc.CreateResource(obj)
+	// Only wait for ASO objects
+	if _, ok := obj.(genruntime.MetaObject); !ok {
+		return
+	}
```
theunrepentantgeek, line 417: Doesn't this create a race condition where sometimes the non-ASO resource will be created in time for a later step of the test, and sometimes it won't? This seems risky.

v2/internal/testcommon/kube_per_test_context.go
```diff
@@ -420,6 +424,10 @@ func (tc *KubePerTestContext) CreateResourceAndWaitWithoutCleanup(obj client.Obj
 	tc.T.Helper()
 	gen := obj.GetGeneration()
 	tc.CreateResourceUntracked(obj)
+	// Only wait for ASO objects
+	if _, ok := obj.(genruntime.MetaObject); !ok {
+		return
+	}
```
theunrepentantgeek, line 430: Ditto here.

Shouldn't we wait for the resource to be fully created, regardless of whether it's ASO or not?

theunrepentantgeek (changes requested):
I think this will cause flaky tests - let me know if I'm wrong.

v2/internal/testcommon/kube_per_test_context.go
matthchr: Most non-ASO resources (== core Kubernetes resources) such as Secret and ConfigMap don't have a notion of "readiness".

These functions previously panic'ed when called with a non-ASO resource, because right below here we call `tc.Eventually(obj).Should(tc.Match.BeProvisioned(gen))` which internally does `conditioner, ok := obj.(conditions.Conditioner)`, but not every Kubernetes resource implements `Conditioner` (it's ASO specific).

matthchr: Since these panic'ed, this change is safe to make (no callers). If there are callers in the future, it does (closer) to the right thing.

_Technically_ speaking, probably we should have a variant of `v2/internal/testcommon/verify.go` that works for resources that do have a Ready condition but don't implement our interface, such as `Pod`.

In practice, we don't actually make any of those resources in our tests at this time, and if we tried to we _would_ have the problem you're describing, but fixing it is a bigger item and I'm not sure that we really have a need to make Pods or Pod-like objects in our tests.

If we ever do, we can fix this. I'll leave a comment here and the other place I changed to that effect.

super-harsh: Nice to have protection here in `Wait` functions. 
We've always been using `CreateResource` function for non-ASO resources.

matthchr: > Not sure if that's what we want?

IMO it is - if there's nothing to wait for, then CreateResourceAndWait should == CreateResource. If we wanted to Wait for Pods or other resources (which we could in the future) we could add specific handling for that here as well (as I mentioned in above comment)

super-harsh (approved):

theunrepentantgeek (approved):

PR #3781, opened 2024-02-12
Fix local package aliases with "." in them

matthchr:
**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/tools/generator/internal/astmodel/local_package_reference.go
+++ b/v2/tools/generator/internal/astmodel/local_package_reference.go
@@ -146,6 +146,8 @@ func (pr LocalPackageReference) GroupVersion() (string, string) {
 
 // ImportAlias returns the import alias to use for this package reference
 func (pr LocalPackageReference) ImportAlias(style PackageImportStyle) string {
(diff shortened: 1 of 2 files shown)


theunrepentantgeek (changes requested):
Which types have a group containing a `.` ? 

I thought we had specific logic in the loading stages of the pipeline to prevent that from happening. If there's a bug in that, we should fix it there, not patch it at the end of the process.

matthchr:
> Which types have a group containing a . ?

`network.frontdoor` is a new group coming soon. This is added because of the [mappings we define in azure-arm](https://github.com/Azure/azure-service-operator/blob/main/v2/azure-arm.yaml#L622), combined with [this code](https://github.com/Azure/azure-service-operator/blob/main/v2/tools/generator/internal/codegen/pipeline/load_types.go#L794), produces groups containing `.` in their name.

I _think_ that this is expected and that we should tolerate it in package references? It's just until now we've not run up against it.

theunrepentantgeek (approved):

theunrepentantgeek:
I think you need to add some tests, but otherwise good.

super-harsh (approved):

PR #3782, opened 2024-02-12
Redact potentially sensitive annotations in the log

theunrepentantgeek:

Reduce the volume of logging by eliminating tokens from the list of annotations included.

gif

### DIFF
--- a/v2/internal/reconcilers/common.go
+++ b/v2/internal/reconcilers/common.go
@@ -28,9 +28,17 @@ func LogObj(log logr.Logger, level int, note string, obj genruntime.MetaObject) 
 	if log.V(level).Enabled() {
 		ourAnnotations := make(map[string]string)
 		for key, value := range obj.GetAnnotations() {
-			if strings.HasPrefix(key, "serviceoperator.azure.com") {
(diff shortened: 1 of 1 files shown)


v2/internal/reconcilers/common.go
```diff
@@ -28,9 +28,17 @@ func LogObj(log logr.Logger, level int, note string, obj genruntime.MetaObject)
...
 		ourAnnotations := make(map[string]string)
 		for key, value := range obj.GetAnnotations() {
-			if strings.HasPrefix(key, "serviceoperator.azure.com") {
-				ourAnnotations[key] = value
+			if !strings.HasPrefix(key, "serviceoperator.azure.com") {
+				// Skip annotations that aren't ours
+				continue
 			}
+
+			if strings.Contains(key, "token") {
```
matthchr, line 36: minor: Consider making this more specific?

I think we have a const for this annotation someplace?

theunrepentantgeek: I needed to move the annotations to avoid a circular import cycle to do this. Done.

matthchr (approved):

PR #3784, opened 2024-02-14
Add JSONToString and StringToJSON property conversion

super-harsh:

Added helper functions to support JSONToString and StringToJSON property conversion.

**If applicable**:
- [x] this PR contains tests


v2/pkg/genruntime/conversion_helpers.go
```diff
@@ -105,3 +107,19 @@ func GetOptionalIntValue(ptr *int) int {
...
+		return ""
+	}
+	
+	return string(json.Raw)
+}
+
+// ConvertStringToJSON returns the v1.JSON value of the given string.
+func ConvertStringToJSON(s string) v1.JSON {
+	return v1.JSON{
+		Raw: []byte(s),
```
theunrepentantgeek, line 123: Is there no factory function for `v1.JSON` values?

theunrepentantgeek (approved):
Looks good, just a query about whether a factory func exists for `v1.JSON`.

v2/pkg/genruntime/conversion_helpers.go
super-harsh: Done!

PR #3785, opened 2024-02-14
Add support for CDN AzureFrontDoor resources

### LINKED ISSUE #3382: Feature: Add support to CDN Profile AFD resources
Hi Team, We plan to use the ASO with flux for our resource deployments, The existing CDN Profile doesn't support AFD resources. 

Please add support to the CDN Profile for the following resources: 

- AFD Endpoints/Routes
- Origin Groups/Origins
- Custom Domains
- Rules Sets/Rules
- Secrets
- Security Policies

matthchr:
The following resources were added in API Version cdn.azure.com/v1api20230501
* AfdCustomDomain (Microsoft.Cdn/profiles/customDomains in Azure)
* AfdEndpoint (Microsoft.Cdn/profiles/afdEndpoints in Azure)
* AfdOriginGroup (Microsoft.Cdn/profiles/originGroups in Azure)
* AfdOrigin (Microsoft.Cdn/profiles/originGroups/origins in Azure)
* [new version, was already supported in cdn.azure.com/v1api2020601] Profile (Microsoft.Cdn/profiles in Azure)
* Route (Microsoft.Cdn/profiles/afdEndpoints/routes in Azure)
* RuleSet (Microsoft.Cdn/profiles/ruleSets in Azure)
* Rule (Microsoft.Cdn/profiles/ruleSets/rule in Azure)
* Secret (Microsoft.Cdn/profiles/secrets in Azure)
* SecurityPolicy (Microsoft.Cdn/profiles/securityPolicies in Azure)

Also added one AFD classic resource which is used in both AFD classic but also AFD standard/premium in API version network.frontdoor.azure.com/v1api20220501
* WebApplicationFirewallPolicy (Microsoft.Network/frontDoorWebApplicationFirewallPolicies in Azure)

Note that there are other resources in the network.frontdoor folder in the REST API specs repo, which correspond to AzureFrontDoor classic. The CDN provider contains resources for AzureFrontDoor Standard and Premium. As I understand it, AFD standard/premium are recommended over AFD classic, so for now only adding support for AFD standard/premium via the CDN provider.

This resolves #3382.

**If applicable**:
- [x] this PR contains tests


v2/azure-arm.yaml
```diff
@@ -1197,6 +1209,78 @@ objectModelConfiguration:
...
+#      Profiles_Endpoints_Origin:
+#        $exportAs: Origin
+#        $supportedFrom: v2.6.0
+#      Profiles_Endpoints_CustomDomain:
+#        $exportAs: CustomDomain
+#        $supportedFrom: v2.6.0
+      Profiles_CustomDomain:
+        $exportAs: AfdCustomDomain
+        $supportedFrom: v2.6.0
+      Profiles_AfdEndpoint:
```
super-harsh, line 1254: There is `Endpoint/Origin` and `Endpoint/OriginGroup` resources as well. Do we want them too? 
https://learn.microsoft.com/en-us/azure/templates/microsoft.cdn/profiles/endpoints/origins?pivots=deployment-language-arm-template

v2/internal/controllers/cdn_profile_afd_test.go
```diff
@@ -0,0 +1,350 @@
...
+	expectCDNDeleteSucceeds(tc, rule, *rule.Status.Id)
+	expectCDNDeleteSucceeds(tc, ruleSet, *ruleSet.Status.Id)
+	expectCDNDeleteSucceeds(tc, securityPolicy, *securityPolicy.Status.Id)
+	expectCDNDeleteSucceeds(tc, endpoint, *endpoint.Status.Id)
+	expectCDNDeleteSucceeds(tc, customDomain, *customDomain.Status.Id)
+
+	expectCDNDeleteSucceeds(tc, profile, *profile.Status.Id)
+}
+
+func expectCDNDeleteSucceeds(tc *testcommon.KubePerTestContext, obj genruntime.ARMMetaObject, id string) {
```
super-harsh, line 114: Would be good to add this in `testcommon` for future use as well. 
`func ensureDeleteSucceeds(tc *testcommon.KubePerTestContext, obj genruntime.ARMMetaObject, id string, apiVersion string)`

v2/internal/testcommon/test_context.go
```diff
@@ -177,6 +177,7 @@ func createTestNamespaceName(t *testing.T) string {
 
 	name := strings.ReplaceAll(strings.ToLower(t.Name()), "/", "-")
 	result := "aso-" + strings.ReplaceAll(name, "_", "-")
+	result = strings.ReplaceAll(result, ".", "-")
```
super-harsh, line 180: could use regexp(`[./_]`) here to match and replace all at once?

v2/samples/cdn/v1api20230501/refs/v1api20220501_webapplicationfirewallpolicy.yaml
```diff
@@ -0,0 +1,31 @@
...
+        matchVariable: RemoteAddr
+        negateCondition: true
+        operator: IPMatch
+      name: RateLimit
+      priority: 100
+      rateLimitDurationInMinutes: 5
+      rateLimitThreshold: 10
+      ruleType: RateLimitRule
+  location: global
+
```
super-harsh, line 26: Minor: extra line

v2/samples/cdn/v1api20230501/v1api20230501_profile.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Profile
+metadata:
+  name: aso-sample-cdnprofile
+  namespace: default
+spec:
+  owner:
+    name: aso-sample-rg
+  location: Global
+  sku:
+    name: Standard_AzureFrontDoor
```
super-harsh, line 11: Minor: Need new line

v2/samples/network.frontdoor/v1api/v1api20220501_webapplicationfirewallpolicy.yaml
```diff
@@ -0,0 +1,30 @@
...
+        matchVariable: RemoteAddr
+        negateCondition: true
+        operator: IPMatch
+      name: RateLimit
+      priority: 100
+      rateLimitDurationInMinutes: 5
+      rateLimitThreshold: 10
+      ruleType: RateLimitRule
+  location: global
+
```
super-harsh, line 25: Minor: Extra line

super-harsh (commented):
LGTM, just a few minor comments and thoughts below.

v2/internal/controllers/cdn_profile_afd_test.go
matthchr: Fair point - will add

v2/internal/testcommon/test_context.go
matthchr: Done

v2/azure-arm.yaml
matthchr: See comment in azure-arm.yaml for this:
```
      # DeepCreatedOriginGroup actually is a resource, for some CDN SKUs it's required to be specified on the CDN endpoint
      # while for other SKUs it seems to be optional on the endpoint (or may not be allowed to be set on the endpoint at all, I am not sure).
      # Given the requirement at least for classic CDNs (Sku == Standard_Microsoft) to specify Origin at the Endpoint scope
      # we can't treat this purely as a resource. Instead, I think this field is like AKS AgentPools - both embedded and not, and
      # the author of the ASO resource must choose which makes sense for their context.
      # This comment applies to all the "DeepCreated*" resources.
      DeepCreatedOriginGroup:
        $isResource: false
      DeepCreatedOriginGroup_STATUS:
        $isResource: false
      DeepCreatedOriginGroupProperties:
        $isResource: false
      DeepCreatedOriginGroupProperties_STATUS:
        $isResource: false
      DeepCreatedOrigin:
        $isResource: false
      DeepCreatedOrigin_STATUS:
        $isResource: false
```

It's not clear to me when they allow management via the child resource and when they don't. So I started with AFD which at least was clear to me. We can add these resources in as actual resources in the future as well but it may require merging with the `DeepCreated` variants on the parent objects - needs more testing.

v2/azure-arm.yaml
```diff
@@ -1197,6 +1209,78 @@ objectModelConfiguration:
...
+        Conditions:
+          $renameTo: RuleConditions # Renamed to not clash with our "Conditions"
+      Profiles_RuleSets_Rule_STATUS:
+        Conditions:
+          $renameTo: RuleConditions # Renamed to not clash with our "Conditions"
+      Profiles_SecurityPolicy:
+        $exportAs: SecurityPolicy
+        $supportedFrom: v2.6.0
+      Profiles_Secret:
+        $exportAs: Secret
```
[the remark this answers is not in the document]
matthchr: We're (unfortunately) not actually fully consistent here.

Sometimes we do parent + child (for example `RedisFirewallRule`) while sometimes we do just child (for example, APIM `Service_Product` becomes `Product`).

What I did here (and _probably_ what we should do going forward?) is named the objects the same as they are named in the Swagger, which is to say in [this section](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cdn/resource-manager/Microsoft.Cdn/stable/2023-05-01/afdx.json#L843)

```
      "put": {
        "tags": [
          "AFDEndpoints"
        ],
```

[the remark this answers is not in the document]
matthchr: Thoughts?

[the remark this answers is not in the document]
super-harsh: If we go with the above pattern, we'll have to be extra cautious with the names overlap. For example: `Route` resource here could be easily mis-read with `RouteTable/Route` if both have same kind name by the user having an infrastructure with many resources. IMO Adding owner's name makes the resource unique and easily identifiable within the cluster without going much in depth.

theunrepentantgeek (approved):

v2/azure-arm.yaml
```diff
@@ -841,6 +841,12 @@ status:
 #     The value should be the name of the parent resource which owns the lifecycle of the
 #     sub-resource.
 #
+# $renameTo: <string>
+#     Changes the name of the property.
+#
+#     The name in the serialized ARM payload does NOT change. This only changes the name of the property
+#     in the ASO types. The shape serialized to ARM uses the original name.
+#
```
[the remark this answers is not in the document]
matthchr: Unless it got somehow removed in the interim?

I went to add support and realized we already had it =)

v2/azure-arm.yaml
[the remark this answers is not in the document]
theunrepentantgeek: Our standard is to give child resources individual names - the resources that use `parentChild` naming either predate that decision or we didn't catch it prior to release and we now can't fix it.

VirtualNetworks_Subnet -> VirtualNetworksSubnet

This is one of those. It _should_ be `Subnet` - the added verbiage adds no value. This one predates the naming decision.

Names are unique within the group/version, they don't need to be (and can't be) globally unique.

> It helps user to easily identify what resource should be the owner.

Not every resource has a single potential owner.

theunrepentantgeek (approved):

v2/internal/testcommon/test_context.go
```diff
@@ -171,12 +171,13 @@ func (tc TestContext) ForTest(t *testing.T, cfg config.Values) (PerTestContext,
...
 
+var replaceRegex = regexp.MustCompile("[./_]")
+
 func createTestNamespaceName(t *testing.T) string {
 	const maxLen = 63
 	const disambigLength = 5
 
-	name := strings.ReplaceAll(strings.ToLower(t.Name()), "/", "-")
-	result := "aso-" + strings.ReplaceAll(name, "_", "-")
+	result := replaceRegex.ReplaceAllString(strings.ToLower(t.Name()), "-")
```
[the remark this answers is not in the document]
matthchr: Oops. That was 100% unintentional good catch

v2/api/cdn/customizations/security_policy_extension_types.go
```diff
@@ -0,0 +1,56 @@
...
+// isRetryableBadRequest checks the passed error to see if it is a retryable conflict, returning true if it is.
+func isSecurityPolicyRetryableBadRequest(err *genericarmclient.CloudError) bool {
+	if err == nil {
+		return false
+	}
+
+	// string matching for error detection is fragile but unfortunately there's no better way to determine this given the
+	// shape the API returns
+	return err.Code() == "BadRequest" &&
+		strings.Contains(err.Message(), "is not associated with the AFDX profile or is in a invalid provisioning state")
```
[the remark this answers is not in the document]
matthchr: done

v2/internal/controllers/networkfrontdoor_webapplicationfirewall_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	. "github.com/onsi/gomega"
+
+	frontdoor "github.com/Azure/azure-service-operator/v2/api/network.frontdoor/v1api20220501"
+	resources "github.com/Azure/azure-service-operator/v2/api/resources/v1api20200601"
+	"github.com/Azure/azure-service-operator/v2/internal/testcommon"
+	"github.com/Azure/azure-service-operator/v2/internal/util/to"
+)
+
+// https://github.com/Azure/azure-quickstart-templates/blob/master/quickstarts/microsoft.cdn/front-door-standard-premium-rate-limit/azuredeploy.json
```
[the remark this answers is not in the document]
matthchr: Added clarifying text surrounding this

PR #3787, opened 2024-02-15
Mark VM and VMSS extensions ProtectedSettings as map[string]string Se…

super-harsh:

Mark VM and VMSS extensions ProtectedSettings as map[string]string SecretReference

### DIFF
--- a/v2/azure-arm.yaml
+++ b/v2/azure-arm.yaml
@@ -279,10 +279,27 @@ typeTransformers:
     because: The enum restrictions on this type are incomplete. Making this type a plain string to allow for growth of supported VM sizes without a new API version
 
   - group: compute
-    name: VirtualMachineScaleSetExtensionProperties
+    name: VirtualMachineScaleSetExtensionProperties;VirtualMachineExtensionProperties
+    property: ProtectedSettings
+    ifType:
+      map:
+        key:
+          name: string
+        value:
+          name: any
+    target:
+      map:
+        key:
+          name: string
+        value:
+          name: string
+    because: ProtectedSettings is defined as map[string]v1.JSON and it should be a marked as secret.
+
+  - group: compute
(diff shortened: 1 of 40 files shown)


v2/azure-arm.yaml
```diff
@@ -279,10 +279,27 @@ typeTransformers:
...
+          name: string
+        value:
+          name: any
+    target:
+      map:
+        key:
+          name: string
+        value:
+          name: string
+    because: ProtectedSettings is defined as map[string]v1.JSON and it should be a marked as secret.
```
matthchr, line 296: Im actually not sure we've ever changed a fields type without changing its name before in the same version.

We should confirm that this isn't going to have problems in upgrade. I think we can do that by:
1. Create a VMSS w/ Extension that specifies ProtectedSettings in old version (map[string]JSON)
2. Upgrade to new version of ASO
3. GET the existing resource, what do you see
4. Edit the existing resource, does it reconcile successfully (no controller error logs?)

I'm worried that there might be a deserialization issue but I'm not 100% sure where it will show up or if it'll even happen.

super-harsh: Since its a breaking change, I think users would have to follow the steps to put `reconcile-policy: skip` and then re-create the resource with secret ref?

super-harsh: Tested scenario: 
 1. Create a VMSS w/ Extension that specifies ProtectedSettings in old version (map[string]JSON)
 2. Add `reconcile-policy: skip` on VMSS resource
 3. Upgrade to new version of ASO
 4. Add secret for protectedSettings
 5. Edit the resource to use secretMap
 6. Remove `reconcile-policy: skip` from the resource
 
 Outcome: Success

matthchr (approved):

PR #3820, opened 2024-02-23
Fix ACR protect cmd run during official release

matthchr:


Taskfile.yml
```diff
@@ -456,12 +456,13 @@ tasks:
       - 'if [ -z "${DOCKER_REGISTRY}" ]; then echo "Error: DOCKER_REGISTRY must be set"; exit 1; fi'
       - az acr repository update --name ${DOCKER_REGISTRY} --image "{{.DOCKER_PUSH_TARGET}}/{{.CONTROLLER_DOCKER_IMAGE}}" --write-enabled false --delete-enabled false
     vars:
+      # This strips the leftmost .*.azurecr.io/ from the PUSH_TARGET as the acr repository update command doesn't want that bit in the image parameter
       DOCKER_PUSH_TARGET:
         sh: |
           if [ ! -z "${DOCKER_PUSH_TARGET}" ]; then
-            echo "${DOCKER_PUSH_TARGET}"
+            echo "${DOCKER_PUSH_TARGET}" | sed -e "s|^.*.azurecr.io/||"
```
super-harsh, line 463: az cli automatically strips it out, I guess?

Taskfile.yml
```diff
@@ -456,12 +456,13 @@ tasks:
...
           else
-            echo "{{.DOCKER_PUSH_TARGET}}"
+            echo "{{.DOCKER_PUSH_TARGET}}" | sed -e "s|^.*.azurecr.io/||"
```
super-harsh, line 465: Should output an error if `DOCKER_PUSH_TARGET` is not specified?

Taskfile.yml
matthchr: That's being applied to the `--name ${DOCKER_REGISTRY}` argument, _not_ the `--image` argument (I've tested this locally)

Taskfile.yml
theunrepentantgeek, line 465: Do you need to do this with `sed` or could you do this within **task**?

```suggestion
            echo "{{.DOCKER_PUSH_TARGET | regexReplaceAllLiteral '^.*.azurecr.io' '' }}"
```

theunrepentantgeek (approved):
I think you can tidy up by avoiding `sed`, but that's minor - see my comment.

Taskfile.yml
matthchr: This will only work for the `.DOCKER_PUSH_TARGET` variable. AFAIK these task replacements don't work for env variables. The reason for the env variable is to allow a default in task + an override from the env, without requiring the env to be set.

I think it makes more sense to use sed for both, than to use `regexReplaceAllLiteral` for one and `sed` for the other.

PR #3825, opened 2024-02-25
Use gofumpt

theunrepentantgeek:

Tightens up our source code formatting by switching from `gofmt` to `gofumpt`. 

I've grouped each kind of change into a single commit, so they can be reviewed in isolation. They're pretty much the kinds of tidyups we've been doing manually, so I think this has worked well.

gif


.devcontainer/install-dependencies.sh
```diff
@@ -210,6 +210,10 @@ go-install gen-crd-api-reference-docs github.com/ahmetb/gen-crd-api-reference-do
 #doc# | setup-envtest | latest | https://book.kubebuilder.io/reference/envtest.html |
 go-install setup-envtest sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
 
+# Stricter GO formatting
+#doc# | gofumpt | latest | https://pkg.go.dev/mvdan.cc/gofumpt |
+go-install gofumpt mvdan.cc/gofumpt@latest
```
matthchr, line 215: Should we pin to a specific version to avoid breaking updates? or try latest for a while and then see?

v2/pkg/genruntime/core/errors.go
```diff
@@ -207,8 +213,10 @@ func NewSubscriptionMismatchError(expectedSub string, actualSub string) *Subscri
 	}
 }
 
-var _ error = &SubscriptionMismatch{}
```
matthchr, line 210: This particular usage of this pattern is a bit funky

matthchr (approved):
I'm OK moving to use this, given that `gofmt(gofumpt(code)) == gofumpt(code)`. I think we should fix our devcontainer to use it too though: https://github.com/mvdan/gofumpt?tab=readme-ov-file#visual-studio-code

.devcontainer/install-dependencies.sh
theunrepentantgeek: Try for a while.

v2/pkg/genruntime/core/errors.go
theunrepentantgeek: Agreed

super-harsh (approved):

PR #3832, opened 2024-02-28
Improve heuristic for detection of ARM Resource References

### LINKED ISSUE #3772: Bug: MySQL Flexible Server property sourceServerResourceId has wrong type
While responding to a [question on Slack](https://kubernetes.slack.com/archives/C046DEVLAQM/p1707656368127859), I spotted the property [FlexibleServer](https://azure.github.io/azure-service-operator/reference/dbformysql/v1api20210501/#dbformysql.azure.com/v1api20210501.FlexibleServer) `sourceServerResourceId` which looks to be an ARM resource ID, and therefore should be a `genruntime.ResourceReference` not a `string`.

We can't really change the existing version, but I think we should import the latest version(s) of the API (`2023-06-30` and maybe `2023-10-01-Preview`) and change it there.

theunrepentantgeek:

As detailed in #3772 we have a few properties that are ARM Resource References, but had slipped through both our heuristic and our manual checks.

This PR improves our heuristic to try and prevent future issues.

Closes #3772

gif


v2/tools/generator/internal/codegen/pipeline/apply_cross_resource_references_from_config.go
```diff
@@ -7,10 +7,10 @@ package pipeline
 
 import (
 	"context"
-
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	"regexp"
```
matthchr, line 13: minor: stdlib, goes at the top

v2/tools/generator/internal/codegen/pipeline/add_cross_resource_references.go
```diff
@@ -20,7 +20,7 @@ import (
 // TransformCrossResourceReferencesStageID is the unique identifier for this pipeline stage
 const TransformCrossResourceReferencesStageID = "transformCrossResourceReferences"
 
-var armIDDescriptionRegex = regexp.MustCompile("(?i)(.*/subscriptions/.*?/resourceGroups/.*|ARM ID|Resource ID|resourceId)")
```
matthchr, line 23: minor: Consider either colocating this regex in the pipeline stage, or putting the new regex you added in the pipeline stage here?

It feels like the `armIDDescriptionRegex` and the `armIDNameRegex` should be together.

matthchr (approved):

v2/tools/generator/internal/codegen/pipeline/add_cross_resource_references.go
theunrepentantgeek: Moved.

v2/tools/generator/internal/codegen/pipeline/apply_cross_resource_references_from_config.go
theunrepentantgeek: Fixed.

PR #3833, opened 2024-02-28
Remove dependency for kube-rbac-proxy and add secure-metrics feature 

### LINKED ISSUE #3741: Feature: Securely serve metrics using controller-runtime, instead of kube-rbac-proxy
Today, ASO securely serves metrics with kube-rbac-proxy. It can do this with controller-runtime itself, using a feature added in https://github.com/kubernetes-sigs/controller-runtime/pull/2407, and available since the v0.16.0 release. This allows ASO to drop its dependency on kube-rbac-proxy.

For reference, the cluster-api core project motivated the above work, which it integrated in https://github.com/kubernetes-sigs/cluster-api/pull/9264. Some years ago, cluster-api used to deploy kube-rbac-proxy, but removed it in https://github.com/kubernetes-sigs/cluster-api/pull/4640.

Related slack conversation: https://kubernetes.slack.com/archives/C046DEVLAQM/p1706203810505779

super-harsh:
Closes #3741 

This PR removes dependency for `kube-rbac-proxy` for securely serving metrics. 
Instead, we have a `secure-metrics` feature flag in ASO using which users can toggle to enable metrics via HTTPs.

Bonus: Have added `pprof` handlers as well. 

**If applicable**:
- [ ] this PR contains tests


docs/hugo/content/guide/metrics.md
```diff
@@ -7,7 +7,7 @@ The metrics exposed fall into two groups: Azure based metrics, and reconciler me
 
 ## Toggling the metrics
 
-By default, metrics for ASOv2 are turned on and can be toggled by the following options:
+By default, secure metrics for ASOv2 are turned on and can be toggled by the following options:
```
theunrepentantgeek, line 10: Are they really turned _**on**_, given the option below says `default: false` ?

docs/hugo/content/guide/metrics.md
```diff
@@ -29,8 +30,71 @@ By default, metrics for ASOv2 are turned on and can be toggled by the following
...
-   
+
+## Scraping Metrics Securely via HTTPs
+
+A ServiceAccount token is required to scrape metrics securely. The corresponding ServiceAccount needs permissions on the "/metrics" and "debug/pprof" paths. 
+This can be achieved e.g. by following the [Kubernetes documentation](https://kubernetes.io/docs/concepts/cluster-administration/system-metrics/).
+
+- Use the settings below in your deployment: 
+
+- #### ASOv2 Helm Chart
```
theunrepentantgeek, line 43: I don't think this bullet point needs to be a heading - and if it were a heading, it would be 3rd level, not 4th.

Suggest changing to plain text.

scripts/v2/generate-helm-manifest.sh
```diff
@@ -43,8 +42,8 @@ rm "$GEN_FILES_DIR"/*_namespace_* # remove namespace as we will let Helm manage
 sed -i "s/\(version: \)\(.*\)/\1${VERSION//v}/g" "$ASO_CHART"/Chart.yaml  # find version key and update the value with the current version
 
 # Deployment replacements
-grep -E $KUBE_RBAC_PROXY "$GEN_FILES_DIR"/*_deployment_* > /dev/null # Ensure that what we're about to try to replace actually exists (if it doesn't we want to fail)
-sed -i "s@$KUBE_RBAC_PROXY.*@{{.Values.image.kubeRBACProxy}}@g" "$GEN_FILES_DIR"/*_deployment_*
+#grep -E $KUBE_RBAC_PROXY "$GEN_FILES_DIR"/*_deployment_* > /dev/null # Ensure that what we're about to try to replace actually exists (if it doesn't we want to fail)
+#sed -i "s@$KUBE_RBAC_PROXY.*@{{.Values.image.kubeRBACProxy}}@g" "$GEN_FILES_DIR"/*_deployment_*
```
theunrepentantgeek, line 46: Commented out - either restore or delete.

v2/charts/azure-service-operator/values.yaml
```diff
@@ -92,6 +92,9 @@ image:
 # 'address' field defines the metrics binding address on which metrics
 metrics:
   enable: true
+  # secureMetrics controls whether metrics should be served via 'http' or 'https'.
+  # Flagging secureMetrics as true would use https
+  secureMetrics: false
```
theunrepentantgeek, line 97: The comment in this yaml file reinforces my earlier question about whether secure metrics are turned on by default.

theunrepentantgeek (approved):
Looks good, but needs a couple minor tweaks. Approving so you can merge when they're done.

v2/config/rbac/auth_proxy_service.yaml
```diff
@@ -1,14 +0,0 @@
-apiVersion: v1
-kind: Service
-metadata:
```
matthchr, line 3: We should possibly remove the auth_proxy role and rolebinding and just move this into our default role?
```
- apiGroups:
  - authentication.k8s.io
  resources:
  - tokenreviews
  verbs:
  - create
- apiGroups:
  - authorization.k8s.io
  resources:
  - subjectaccessreviews
  verbs:
  - create
```

That's what CAPI did here? https://github.com/kubernetes-sigs/cluster-api/pull/9264/files

v2/charts/azure-service-operator/values.yaml
```diff
@@ -92,6 +92,9 @@ image:
 # 'address' field defines the metrics binding address on which metrics
 metrics:
   enable: true
+  # secureMetrics controls whether metrics should be served via 'http' or 'https'.
```
matthchr, line 95: It's more than just HTTP vs HTTPS - if you use secure mode, it means that we will also be doing RBAC checks. Should we explain that here?

docs/hugo/content/guide/metrics.md
matthchr, line 10: Questions based on our chat:
1. Do we actually need to expose an `8080` service - it seems that CAPI doesn't? I think, you should be able to scrape the pods metrics from within the cluster without a service. Maybe we need to do a little bit more reading/learning on if we really should have a service exposed for metrics or not.
2. Is the right default, to have securemetrics= false - I sorta wonder if we should set the default to true. I acknowledge that this is _not_ what we were doing before, but it is I think what we _thought_ we were doing before. I think etiher way is breaking because before we were exposing 2 services, an 8443 AND an 8080.

v2/charts/azure-service-operator/values.yaml
matthchr: Yes you already document the required roles in the `metrics.md` - we can possibly link to it here, OR just include a little snippet w/ the names of the roles (or both).

docs/hugo/content/guide/metrics.md
matthchr: Just maybe doublecheck other projects doing this... which ones have services and which ones don't? Or we can talk with John Rusk or somebody on his team, they may have suggestions here.

v2/config/rbac/auth_proxy_service.yaml
super-harsh: I'd rather make this `metrics-role` role to keep the things more granular?

docs/hugo/content/guide/metrics.md
```diff
@@ -7,19 +7,20 @@ The metrics exposed fall into two groups: Azure based metrics, and reconciler me
...
--  ### ASOv2 Helm Chart
+### ASOv2 Helm Chart
 
     While installing the Helm chart, we can turn the metrics _**on**_ and _**off**_ and set the metrics expose address using the 
     below settings. Also, we can change the settings inside `values.yaml` file for ASOv2 Helm chart.
 
     ```
     --set metrics.enable=true/false (default: true)
-    --set metrics.address=0.0.0.0:8080 (default)    
+    --set metrics.secure-metrics=true/false (default: true)
```
matthchr, line 19: minor: I think call this just `metrics.secure`? The fact it's metrics is sorta implied given the path? Just like we don't call it `metrics.enableMetrics`?

docs/hugo/content/guide/metrics.md
```diff
@@ -29,8 +30,71 @@ By default, metrics for ASOv2 are turned on and can be toggled by the following
...
+
+A ServiceAccount token is required to scrape metrics securely. The corresponding ServiceAccount needs permissions on the "/metrics" and "debug/pprof" paths. 
+This can be achieved e.g. by following the [Kubernetes documentation](https://kubernetes.io/docs/concepts/cluster-administration/system-metrics/).
+
+- Use the settings below in your deployment: 
+
+    ### ASOv2 Helm Chart
+        ```
+        --set metrics.enable=true
+        --set metrics.secure-metrics=true
```
matthchr, line 46: fix here if changed above

docs/hugo/content/guide/metrics.md
```diff
@@ -29,8 +30,71 @@ By default, metrics for ASOv2 are turned on and can be toggled by the following
...
-   
+
+## Scraping Metrics Securely via HTTPs using RBAC
+
+A ServiceAccount token is required to scrape metrics securely. The corresponding ServiceAccount needs permissions on the "/metrics" and "debug/pprof" paths. 
+This can be achieved e.g. by following the [Kubernetes documentation](https://kubernetes.io/docs/concepts/cluster-administration/system-metrics/).
+
+- Use the settings below in your deployment: 
+
+    ### ASOv2 Helm Chart
```
matthchr, line 43:
```suggestion
### ASOv2 Helm Chart
```

docs/hugo/content/guide/metrics.md
```diff
@@ -29,8 +30,71 @@ By default, metrics for ASOv2 are turned on and can be toggled by the following
...
+         - --secure-metrics=true/false (default: true)
     ```
-   
+
+## Scraping Metrics Securely via HTTPs using RBAC
+
+A ServiceAccount token is required to scrape metrics securely. The corresponding ServiceAccount needs permissions on the "/metrics" and "debug/pprof" paths. 
+This can be achieved e.g. by following the [Kubernetes documentation](https://kubernetes.io/docs/concepts/cluster-administration/system-metrics/).
+
+- Use the settings below in your deployment:
```
matthchr, line 41:
```suggestion
Follow the steps below to scrape metrics securely 
```

docs/hugo/content/guide/metrics.md
```diff
@@ -29,8 +30,71 @@ By default, metrics for ASOv2 are turned on and can be toggled by the following
...
+- Use the settings below in your deployment: 
+
+    ### ASOv2 Helm Chart
+        ```
+        --set metrics.enable=true
+        --set metrics.secure-metrics=true
+        --set metrics.address=0.0.0.0:8443
+        ```
+    
+    ### Deployment YAML
```
matthchr, line 50:
```suggestion
### Deployment YAML
```

docs/hugo/content/guide/metrics.md
```diff
@@ -29,8 +30,71 @@ By default, metrics for ASOv2 are turned on and can be toggled by the following
...
+        ```
+        spec:
+          containers:
+           - args:
+             - --metrics-addr=0.0.0.0:8443  
+             - --secure-metrics=true 
+        ```
+
+- Deploy the following RBAC configuration:
```
matthchr, line 59:
```suggestion
- Deploy the following RBAC configuration. This creates a role that can scrape metrics.
```

docs/hugo/content/guide/metrics.md
```diff
@@ -29,8 +30,71 @@ By default, metrics for ASOv2 are turned on and can be toggled by the following
...
+      apiGroup: rbac.authorization.k8s.io
+      kind: ClusterRole
+      name: default-metrics
+    subjects:
+    - kind: ServiceAccount
+      name: default
+      namespace: default
+      EOT
+    ```
+- Open a port-forward
```
matthchr, line 87: These last two steps are only required for local testing, not actually for something like Prom scraping. Might be worth calling that out somehow?

maybe do something like:
```
- Test locally:
    -  Open a port forward: `kubectl port-forward deployments/azureserviceoperator-controller-manager -n azureserviceoperator-system 8443`
    - Create a ServiceAccount token and scrape metrics:
    ```
    TOKEN=$(kubectl create token default)
    curl https://localhost:8443/metrics --header "Authorization: Bearer $TOKEN" -k
    ```
```

v2/cmd/controller/app/flags.go
```diff
@@ -54,6 +56,8 @@ func ParseFlags(args []string) (Flags, error) {
 
 	// default here for 'MetricsAddr' is set to "0", which sets metrics to be disabled if 'metrics-addr' flag is omitted.
 	flagSet.StringVar(&metricsAddr, "metrics-addr", "0", "The address the metric endpoint binds to.")
+	flagSet.BoolVar(&secureMetrics, "secure-metrics", true, "Enable secure metrics. This will enable serving pprof endpoints and metrics securely using https")
```
matthchr, line 59: This probably should still be `secure-metrics` (the container cmdline arg). Just the Helm chart variable changes name?

matthchr, line 59:
```suggestion
	flagSet.BoolVar(&secureMetrics, "secure-metrics", true, "Enable secure metrics. This secures the pprof and metrics endpoints via Kubernetes RBAC and HTTPS")
```

v2/cmd/controller/app/setup.go
```diff
@@ -253,6 +254,30 @@ func SetupControllerManager(ctx context.Context, setupLog logr.Logger, flgs Flag
...
 
+func getMetricsOpts(flags Flags) server.Options {
+	var metricsOptions server.Options
+	if flags.SecureMetrics {
+		metricsOptions = server.Options{
+			BindAddress:    flags.MetricsAddr,
+			SecureServing:  true,
+			FilterProvider: filters.WithAuthenticationAndAuthorization,
+			// Note that pprof endpoints are meant to be sensitive and shouldn't be exposed publicly.
+			ExtraHandlers: map[string]http.Handler{
```
matthchr, line 265: We probably should have a specific cmdline flag that enables this and it should default to off.

This is what [apiserver does](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/) via the `--profiling` argument, though it defaults it to true.

I'd argue we should have both `metrics.secure` and `metrics.profiling` flags in Helm, and two cmdline args in ASO to control these two bits.

Secure metrics shouldn't (IMO) _require_ that you expose pprof.

v2/config/manager/manager_metrics_service.yaml
```diff
@@ -10,6 +10,6 @@ metadata:
 spec:
   ports:
     - name: metrics
-      port: 8080
+      port: 8443
```
matthchr, line 13: This is technically a breaking change but I think a good one.

We should remember to notify customers about it. Created #3846 to track this.

v2/config/rbac/crd_manager_role.yaml
```diff
@@ -14,3 +14,15 @@ rules:
       - patch
       - update
       - watch
+  - apiGroups:
```
matthchr, line 17: minor: Consider leaving a comment here that this is required for secure metrics?

v2/cmd/controller/app/setup.go
```diff
@@ -253,6 +254,37 @@ func SetupControllerManager(ctx context.Context, setupLog logr.Logger, flgs Flag
 	return mgr
 }
 
+func getMetricsOpts(flags Flags) server.Options {
+	var metricsOptions server.Options
+
+	if !flags.SecureMetrics {
```
matthchr, line 260: This code doesn't seem right, because it doesn't return here, it falls through and then overwrites the `metricsOptions` below?

I think you want something like:
```
var metricsOptions
if secure {
    // secure
} else {
    // insecure
}

if profiling {
    // profiling
}

return
```

v2/charts/azure-service-operator/values.yaml
```diff
@@ -86,14 +86,19 @@ webhook:
...
 
 # 'metrics' define settings for the metrics from controller.
 # 'address' field defines the metrics binding address on which metrics
 metrics:
   enable: true
+  # secure controls whether metrics should be served via 'http' or 'https'.
+  # Flagging secure as 'true' would use https
+  # Refer to https://azure.github.io/azure-service-operator/guide/metrics/ for more information
+  secure: true
+  # profiling endpoints are only enabled when serving metrics securely
```
matthchr, line 98: minor: Say what profiling is (it enables /pprof endpoints), in addition to its restrictions. Though, see below about this restriction because I am not sure it is needed?

v2/cmd/controller/app/setup.go
matthchr: Or if you want to only enable profiling if secure, put that bit into the secure bit, and/or (maybe better) add a check that secure is not off with profiling on and if so return an error and crash the pod.

v2/cmd/controller/app/flags.go
```diff
@@ -54,6 +60,9 @@ func ParseFlags(args []string) (Flags, error) {
 
 	// default here for 'MetricsAddr' is set to "0", which sets metrics to be disabled if 'metrics-addr' flag is omitted.
 	flagSet.StringVar(&metricsAddr, "metrics-addr", "0", "The address the metric endpoint binds to.")
+	flagSet.BoolVar(&secureMetrics, "secure-metrics", true, "Enable secure metrics. This secures the pprof and metrics endpoints via Kubernetes RBAC and HTTPS")
+	flagSet.BoolVar(&profilingMetrics, "profiling-metrics", true, "Enable pprof metrics, only enabled in conjunction with secure-metrics. This will enable serving pprof metrics endpoints")
```
matthchr, line 64: Default should probably be false

v2/cmd/controller/app/setup.go
super-harsh: Ah good catch. I missed return there. I'll update

docs/hugo/content/guide/metrics.md
```diff
@@ -29,8 +31,75 @@ By default, metrics for ASOv2 are turned on and can be toggled by the following
...
-   
+
+## Scraping Metrics Securely via HTTPs using RBAC
+
+A ServiceAccount token is required to scrape metrics securely. The corresponding ServiceAccount needs permissions on the "/metrics" and "debug/pprof" paths. 
+This can be achieved e.g. by following the [Kubernetes documentation](https://kubernetes.io/docs/concepts/cluster-administration/system-metrics/).
+
+Follow the steps below to scrape metrics securely.
+
+### ASOv2 Helm Chart
```
matthchr, line 45: very minor: Could use one of the table-selector thingies here, like we do for Windows versus Linux in the installation instructions?

matthchr (approved):
Approved with one minor style comment

PR #3836, opened 2024-03-01
Improve/vcr tracking

theunrepentantgeek:

@matthchr identified an issue with #3753 where an extra reconcile early in a recorded test (resulting in an extra PUT) would increment the `TEST-REQUEST-ATTEMPT` header value, resulting in all future PUT requests to that URL failing.

This PR fixes that by using a separate count sequence for each distinct PUT hash.

gif

### DIFF
--- a/v2/internal/testcommon/vcr/v3/test_recorder.go
+++ b/v2/internal/testcommon/vcr/v3/test_recorder.go
@@ -92,7 +92,7 @@ func NewTestRecorder(
 		}
 
 		// verify custom request count header (see counting_roundtripper.go)
-		if r.Header.Get(vcr.COUNT_HEADER) != i.Headers.Get(vcr.COUNT_HEADER) {
+		if r.Header.Get(COUNT_HEADER) != i.Headers.Get(COUNT_HEADER) {
 			return false
 		}
 
@@ -222,9 +222,9 @@ func (r *recorderDetails) IsReplaying() bool {
 func (r *recorderDetails) CreateClient(t *testing.T) *http.Client {
 	withReplay := NewReplayRoundTripper(r.recorder, r.log)
 	withErrorTranslation := translateErrors(withReplay, r.cassetteName, t)
-	withCountHeader := vcr.AddCountHeader(withErrorTranslation)
+	withTrackingHeaders := AddTrackingHeaders(withErrorTranslation)
 
 	return &http.Client{
-		Transport: withCountHeader,
+		Transport: withTrackingHeaders,
 	}
 }
--- a/v2/internal/testcommon/vcr/v3/test_recorder.go
+++ b/v2/internal/testcommon/vcr/v3/test_recorder.go
@@ -104,6 +104,7 @@ func NewTestRecorder(
 		}
 
 		if r.Body == nil {
+			// Empty bodies are stored by go-vcr as empty strings
 			return i.Body == ""
 		}
 
--- a/v2/internal/testcommon/vcr/v3/test_recorder.go
+++ b/v2/internal/testcommon/vcr/v3/test_recorder.go
@@ -6,8 +6,6 @@ Licensed under the MIT license.
 package v3
 
 import (
-	"bytes"
-	"io"
 	"net/http"
 	"strings"
 	"testing"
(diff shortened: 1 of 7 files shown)


v2/internal/testcommon/vcr/v3/tracking_roundtripper.go
```diff
@@ -0,0 +1,104 @@
...
+
+func AddTrackingHeaders(inner http.RoundTripper) *requestCounter {
+	return &requestCounter{
+		inner:       inner,
+		counts:      make(map[string]uint32),
+		countsMutex: sync.Mutex{},
+	}
+}
+
+const (
```
matthchr, line 39: super duper minor: Consider putting these consts at the top of the file

v2/internal/testcommon/vcr/v3/tracking_roundtripper.go
```diff
@@ -0,0 +1,104 @@
...
+	COUNT_HEADER string = "TEST-REQUEST-ATTEMPT"
```
matthchr, line 40: minor: Is the `string` required here? For some reason I thought you could elide the type in this instance.

v2/internal/testcommon/vcr/v3/tracking_roundtripper.go
```diff
@@ -0,0 +1,104 @@
...
+	HASH_HEADER  string = "TEST-REQUEST-HASH"
+)
+
+var _ http.RoundTripper = &requestCounter{}
+
+func (rt *requestCounter) RoundTrip(req *http.Request) (*http.Response, error) {
+	if req.Method == "PUT" || req.Method == "POST" {
```
matthchr, line 47: Not every POST is guaranteed to have a body, it's optional. I _think_ that'll cause an issue because they'll all have the same hash?

PUTs I think probably are guaranteed to have a body.

v2/internal/testcommon/vcr/v3/test_recorder.go
```diff
@@ -92,7 +92,7 @@ func NewTestRecorder(
 		}
 
 		// verify custom request count header (see counting_roundtripper.go)
-		if r.Header.Get(vcr.COUNT_HEADER) != i.Headers.Get(vcr.COUNT_HEADER) {
+		if r.Header.Get(COUNT_HEADER) != i.Headers.Get(COUNT_HEADER) {
```
matthchr, line 95: No need to check body hash here because we're actually checking the body matches exactly, is that right?

Is it worth leaving a comment to that effect?

v2/internal/testcommon/vcr/v3/tracking_roundtripper.go
```diff
@@ -0,0 +1,104 @@
...
+		rt.addHashHeader(req)
```
matthchr, line 48: minor: This adds a hash _and_ a count header. Consider reflecting that reality in the method name, either by saying `HashAndCount` or by saying something like `addTrackingHeaders`

matthchr (commented):
Missing a test (and "end to end unit test?" that demonstrates the problem is solved.

v2/internal/testcommon/vcr/v3/tracking_roundtripper.go
theunrepentantgeek: These were **var** and I changed them. Now simplified further.

v2/internal/testcommon/vcr/v3/tracking_roundtripper.go
theunrepentantgeek: Tweaked the condition.

v2/internal/testcommon/vcr/v3/test_recorder.go
theunrepentantgeek: Good catch. With the hash header, we don't need to read the entire body to do the match.

v2/internal/testcommon/vcr/v3/test_recorder.go
```diff
@@ -91,22 +89,25 @@ func NewTestRecorder(
...
+			if header != i.Headers.Get(HASH_HEADER) {
+				return false
+			}
 		}
 
-		var b bytes.Buffer
-		if _, err := b.ReadFrom(r.Body); err != nil {
-			panic(err)
+		if r.Body == nil {
+			return i.Body == ""
```
matthchr, line 107: Is this just a quirk of vcr that empty bodies are empty strings?

_Maybe_ worth a comment?

v2/internal/testcommon/vcr/v3/tracking_roundtripper.go
matthchr: Did you miss this comment?

matthchr (approved):
LGTM, few minor comments

v2/internal/testcommon/vcr/v3/test_recorder.go
theunrepentantgeek: Yes. Added a comment.

v2/internal/testcommon/vcr/v3/tracking_roundtripper.go
theunrepentantgeek: Had trouble coming up with a non-misleading name that was better than that. But I did, now updated.

super-harsh (approved):

PR #3847, opened 2024-03-06
Modify multitenant test to install operator in custom namespace

### LINKED ISSUE #3715: Bug: helm install -n <namespace-name> breaks operator
For ASOv2, it seems like if you specify a non-default namespace name, the operator doesn't work.

if we don't support the -n parameter, we shouldn't support it at all. If we _do_ support it, it should work.

super-harsh:
Closes #3715 

The above issue was for ASO beta.5 version and does is not true for current version. As a part of this PR, have modified a multitenant test to confirm ASO works fine in other namespaces as well.

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -887,7 +887,7 @@ tasks:
     desc: Waits for the operator and associated CRDs to be ready
     run: always
     cmds:
-      - "{{.SCRIPTS_ROOT}}/wait-for-operator-ready.sh {{.ARGS}}"
+      - "{{.SCRIPTS_ROOT}}/wait-for-operator-ready.sh -n '{{default \"azureserviceoperator-system\" .NAMESPACE}}' {{.ARGS}}"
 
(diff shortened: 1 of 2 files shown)


Taskfile.yml
```diff
@@ -887,7 +887,7 @@ tasks:
     desc: Waits for the operator and associated CRDs to be ready
     run: always
     cmds:
-      - "{{.SCRIPTS_ROOT}}/wait-for-operator-ready.sh {{.ARGS}}"
+      - "{{.SCRIPTS_ROOT}}/wait-for-operator-ready.sh -n '{{default \"azureserviceoperator-system\" .NAMESPACE}}' {{.ARGS}}"
```
matthchr, line 890: Can you not use `.ASO_NAMESPACE` as the default here?

matthchr (approved):
LGTM, 1 comment, PTAL before merging

PR #3863, opened 2024-03-12
Support renaming properties between versions

### LINKED ISSUE #1439: Feature: Support for renaming of properties between versions
Define the required YAML format for capturing renaming vs end-of-life for properties and integrate that into the generation process for the conversion routines.

Prerequisites: Azure/k8s-infra#353 & Azure/k8s-infra#354

theunrepentantgeek:

If a property gets renamed by an RP between versions, we want to hook things up so that intra-version conversions work properly. 

We've had partial support for this for some time - the ability to configure renames, but that configuration was never wired up. This PR does the wiring and tests that the conversion works.

Closes #1439 

gif

**If applicable**:
- [x] this PR contains tests


v2/tools/generator/internal/config/object_model_configuration.go
```diff
@@ -50,6 +50,7 @@ type ObjectModelConfiguration struct {
 	ARMReference                   propertyAccess[bool]
 	ImportConfigMapMode            propertyAccess[ImportConfigMapMode]
 	IsSecret                       propertyAccess[bool]
+	PropertyNameInNextVersion      propertyAccess[string]
```
super-harsh, line 53: Do we need documentation for how to use this property in azure-arm.yaml?

super-harsh (approved):
Looks good. Just a minor comment on documentation below.

v2/tools/generator/internal/config/object_model_configuration.go
theunrepentantgeek: We do - and we have:

``` yaml
# $nameInNextVersion: <string>
#     Gives the name this property has in the next version of the resource
#     Establishes a connection between the two properties allowing for proper
#     forward and backward conversion.
```

theunrepentantgeek: I think this stems from the fact this feature was half completed - we had the configuration, but never used it.

PR #3890, opened 2024-03-26
Update Go version

matthchr:
This was required as setup-envtest@latest no longer works with Go versions less than 1.22.

controller-tools was also updated as it panics if 0.13.0 is used with Go 1.22. 0.14.0 must be used instead.

Go version also moved to 1.21 to comply with new mod shape.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -91,7 +91,7 @@ tasks:
     desc: Ensure all code is formatted
     dir: v2
     cmds:
-      - gofumpt -l -s -w .
+      - gofumpt -l -w .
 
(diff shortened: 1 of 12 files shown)


Taskfile.yml
```diff
@@ -91,7 +91,7 @@ tasks:
     desc: Ensure all code is formatted
     dir: v2
     cmds:
-      - gofumpt -l -s -w .
+      - gofumpt -l -w .
```
theunrepentantgeek, line 94: 😁

theunrepentantgeek (approved):

PR #3892, opened 2024-03-27
Prepare to move ARM types to subpackages

theunrepentantgeek:

Moves ARM types to subpackages, subject to a deny list that maintains the status quo.

We can remove groups from the deny list gradually, allowing a step-by-step migration to the new structure.

The count of changed lines looks large, but is almost entirely due to testing with the `containerservice` group. I can take that out if that's preferable.

gif


matthchr:
We think that the problem with some of the tests is this test-specific function:
`func CreateFileDefinition(definitions ...astmodel.TypeDefinition) *astmodel.FileDefinition {`

It needs to take a collection of packages, so that the codeGenContext has information about the imported packages too. Otherwise it can't find the arm types and fails.

matthchr:
I also merged + pushed, we may need to re-record the golden file tests after my changes.

v2/pkg/genruntime/secrets/collector.go
```diff
@@ -115,11 +116,12 @@ func (c *Collector) Values() ([]*v1.Secret, error) {
...
-	sort.Slice(result, func(i, j int) bool {
-		left := result[i]
-		right := result[j]
+	slices.SortFunc(result, func(left, right *v1.Secret) int {
+		if c := cmp.Compare(left.Namespace, right.Namespace); c != 0 {
+			return c
+		}
 
-		return left.Namespace < right.Namespace || (left.Namespace == right.Namespace && left.Name < right.Name)
+		return cmp.Compare(left.Name, right.Name)
```
matthchr, line 124: Curious what the reason for the sortFunc change here is?

I'm not understanding what the difference is between what we had and what's here now, in terms of actual behavior.

v2/tools/generator/internal/astmodel/internal_type_name.go
```diff
@@ -186,6 +186,10 @@ func (tn InternalTypeName) IsStatus() bool {
 
 // IsARMType returns true if the TypeName identifies an ARM specific type, false otherwise.
 func (tn InternalTypeName) IsARMType() bool {
+	if IsARMPackageReference(tn.InternalPackageReference()) {
+		return true
+	}
+
 	return strings.HasSuffix(tn.Name(), ARMSuffix)
```
matthchr, line 193: Can this second bit be removed if we migrate all types to the new format?

v2/tools/generator/internal/astmodel/type_name.go
```diff
@@ -18,9 +20,58 @@ const (
...
+	"insights",
+	"keyvault",
+	"kubernetesconfiguration",
+	"machinelearningservices",
+	"managedidentity",
+	"monitor",
+	"network",
+	"network.frontdoor",
+	"operationalinsights",
+	//"person",
```
matthchr, line 56: Intentionally here?

matthchr (approved):
LGTM, few minor questions

v2/pkg/genruntime/secrets/collector.go
theunrepentantgeek: I think we've both changed this. You added the ordering by namespace for packages with identical names; I updated it to use the new API. I'll update it.

v2/tools/generator/internal/astmodel/internal_type_name.go
theunrepentantgeek: Once we've moved every group, I'll remove both this and the deny list.

v2/tools/generator/internal/astmodel/type_name.go
theunrepentantgeek: No, will remove.

PR #3910, opened 2024-04-05
Improve/astype signature

### LINKED ISSUE #2970: Type.AsType should allow for returning error

The method `AsType()` on the interface `astmodel.Type`does not allow for implementations to return an error.

**Describe the improvement**

**Prerequisites**

- [x] #2966 
- [x] #2967 
- [x] #2968 
- [x] #2969 
- [x] #2971 

**Additional context**

theunrepentantgeek:

Improves the signature of the `AsTypeExpr` method (formerly `AsType`), allowing it to return errors instead of forcing implementations to panic. 

Closes #2970 

The `AsTypeExpr` method is used in a lot of places!

gif

### DIFF
--- a/v2/tools/generator/internal/armconversion/convert_to_arm_function_builder.go
+++ b/v2/tools/generator/internal/armconversion/convert_to_arm_function_builder.go
@@ -32,8 +32,12 @@ func newConvertToARMFunctionBuilder(
 	codeGenerationContext *astmodel.CodeGenerationContext,
 	receiver astmodel.InternalTypeName,
 	methodName string,
-) *convertToARMBuilder {
-	receiverExpr := receiver.AsTypeExpr(codeGenerationContext)
+) (*convertToARMBuilder, error) {
+	receiverExpr, err := receiver.AsTypeExpr(codeGenerationContext)
+	if err != nil {
+		return nil, errors.Wrapf(err, "creating type expression for %s", receiver)
+	}
+
 	result := &convertToARMBuilder{
 		conversionBuilder: conversionBuilder{
 			methodName:            methodName,
@@ -80,7 +84,7 @@ func newConvertToARMFunctionBuilder(
 		result.propertiesByNameHandler,
 	}
 
-	return result
+	return result, nil
 }
 
 func (builder *convertToARMBuilder) functionDeclaration() (*dst.FuncDecl, error) {
--- a/v2/tools/generator/internal/armconversion/convert_to_arm_function_builder.go
+++ b/v2/tools/generator/internal/armconversion/convert_to_arm_function_builder.go
@@ -33,6 +33,7 @@ func newConvertToARMFunctionBuilder(
 	receiver astmodel.InternalTypeName,
 	methodName string,
 ) *convertToARMBuilder {
+	receiverExpr := receiver.AsTypeExpr(codeGenerationContext)
 	result := &convertToARMBuilder{
 		conversionBuilder: conversionBuilder{
 			methodName:            methodName,
@@ -41,7 +42,7 @@ func newConvertToARMFunctionBuilder(
 			destinationType:       c.armType,
 			destinationTypeName:   c.armTypeName,
 			receiverIdent:         c.idFactory.CreateReceiver(receiver.Name()),
-			receiverTypeExpr:      receiver.AsTypeExpr(codeGenerationContext),
+			receiverTypeExpr:      receiverExpr,
 			idFactory:             c.idFactory,
 			typeKind:              c.typeKind,
 			codeGenerationContext: codeGenerationContext,
@@ -95,7 +96,8 @@ func (builder *convertToARMBuilder) functionDeclaration() (*dst.FuncDecl, error)
 		Body:          body,
 	}
 
-	fn.AddParameter(resolvedParameterString, astmodel.ConvertToARMResolvedDetailsType.AsTypeExpr(builder.codeGenerationContext))
+	convertToARMResolvedDetailsExpr := astmodel.ConvertToARMResolvedDetailsType.AsTypeExpr(builder.codeGenerationContext)
+	fn.AddParameter(resolvedParameterString, convertToARMResolvedDetailsExpr)
 	fn.AddReturns("interface{}", "error")
 	fn.AddComments("converts from a Kubernetes CRD object to an ARM object")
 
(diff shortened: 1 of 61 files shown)


v2/tools/generator/internal/armconversion/convert_to_arm_function_builder.go
```diff
@@ -629,13 +652,26 @@ func (builder *convertToARMBuilder) convertUserAssignedIdentitiesCollection(
 	locals := params.Locals.Clone()
 
 	itemIdent := locals.CreateLocal("ident")
-	keyTypeAst := destinationType.KeyType().AsType(conversionBuilder.CodeGenerationContext)
-	valueTypeAst := destinationType.ValueType().AsType(conversionBuilder.CodeGenerationContext)
+	keyTypeExpr, kerr := destinationType.KeyType().AsTypeExpr(conversionBuilder.CodeGenerationContext)
```
matthchr, line 655: why not use `err` on the lhs here, rather than `kerr` and `verr`?

I thought linting/etc allowed that as long as the other variable in the tuple was new.

v2/tools/generator/internal/astmodel/conversion_function_builder.go
```diff
@@ -407,13 +417,20 @@ func IdentityConvertComplexMapProperty(
 		return astbuilder.InsertMap(lhs, dst.NewIdent(keyIdent), rhs)
 	}
 
-	keyTypeAst := destinationType.KeyType().AsType(builder.CodeGenerationContext)
-	valueTypeAst := destinationType.ValueType().AsType(builder.CodeGenerationContext)
+	keyTypeExpr, kerr := destinationType.KeyType().AsTypeExpr(builder.CodeGenerationContext)
```
matthchr, line 420: same kerr/verr question

v2/tools/generator/internal/astmodel/map_type.go
```diff
@@ -64,15 +65,25 @@ func NewStringMapType(value Type) *MapType {
...
+	return AsSimpleDeclarations(codeGenerationContext, declContext, m)
 }
 
 // AsType implements Type for MapType to create the abstract syntax tree for a map
-func (m *MapType) AsType(codeGenerationContext *CodeGenerationContext) dst.Expr {
-	return &dst.MapType{
-		Key:   m.key.AsType(codeGenerationContext),
-		Value: m.value.AsType(codeGenerationContext),
+func (m *MapType) AsTypeExpr(codeGenerationContext *CodeGenerationContext) (dst.Expr, error) {
+	keyExpr, kerr := m.key.AsTypeExpr(codeGenerationContext)
```
matthchr, line 73: also here

v2/tools/generator/internal/astmodel/property_definition_test.go
```diff
@@ -515,8 +517,8 @@ func Test_PropertyDefinitionAsAst_GivenValidField_ReturnsNonNilResult(t *testing
 		MakeRequired().
 		WithDescription(propertyDescription)
 
-	node := original.AsField(nil)
-
+	node, err := original.AsField(nil)
+	g.Expect(err).To(BeNil())
```
matthchr, line 521: `ToNot(HaveOccurred())` is I think more idiomatic as far as gomega is concerned and (I think) gives a better error message if there is an assertion failure.

v2/tools/generator/internal/functions/conditions_functions.go
```diff
@@ -8,6 +8,8 @@ package functions
 import (
 	"go/token"
 
+	"github.com/pkg/errors"
+
```
matthchr, line 12: super minor: I think can group these imports

v2/tools/generator/internal/functions/kubernetes_exporter_function.go
```diff
@@ -207,12 +210,41 @@ func (d *KubernetesExporterBuilder) exportKubernetesResources(
 			sliceToClientObjectSlice),
 	}
 
-	fn.AddParameter("_", astmodel.ContextType.AsType(codeGenerationContext))
```
matthchr, line 210: Curious as to your thinking on the `err` versus `panic` pattern here, I actually think the old code was significantly easier to read.

I sorta thought that our stance was something like:
1. Error if it's possibly a user-error.
2. Panic if it's a bug.

I've seen a few panics here and there from `AsType`, but don't recall many of them being the users fault. I've definitely seen some that were due to us not handling a particular Swagger well, although I would make that argument that those are all bugs in the generator because we _should_ handle valid Swagger correctly, or have pipeline stages that protect against invalid Swagger before it ever gets to the point where we try to generate types.

Most of the other panic -> error moves were (IIRC) either earlier in the stage of things or didn't have much of an impact on readability. This I think actually has a pretty big impact on readability (due to the ubiquity of `AsType`).

v2/tools/generator/internal/astmodel/property_definition_test.go
theunrepentantgeek: Changed.

v2/tools/generator/internal/functions/conditions_functions.go
theunrepentantgeek: Done.

v2/tools/generator/internal/functions/kubernetes_exporter_function.go
theunrepentantgeek: The approach I've been taking is to panic if it's a bug in the generator, error if it's caused by external factors. 

External factors includes OpenAPI/Swagger specs that we don't correctly support.

I'm driving this change based on experience. 

By it's nature, **panic** only gives a call stack - which is next to useless when the problem is specific to just one object from a specific group/version, because there's no information to tell you which type definition has the problem. Essentially, this _requires_ someone to run the generator inside an IDE in order to debug and troubleshoot. I don't see this as acceptable.

When returning an **error**, we're very good at including context, allowing the problem to be diagnosed and solved without having to resort to a debugger. In many cases, with errors, I can see exactly what needs to be addressed. 

I also this an error is a better user experience - they get told "here's what went wrong" instead of getting a very unfriendly stack dump which says nothing more than "it went bang".

theunrepentantgeek: > I've definitely seen some that were due to us not handling a particular Swagger well, although I would make that argument that those are all bugs in the generator because we should handle valid Swagger correctly

I agree that they're generator bugs. 

But which is preferable? Having to first spend an hour tracking down the source (not the line number) of the panic, or getting straight into fixing the problem because you know which new definition is causing grief.

matthchr: We could trap/wrap panics at a level that injected details about the type it was on, but then we're somewhat back to the `err` at least in the method impls (though it does keep the signatures cleaner). Go doesn't support preserving callstacks after `recover` though I don't think so that's awkward too...

I'm not 100% sold on this but maybe Go forces our hand here, and we certainly have the muscle memory around `err` more than we do around wrapping panics.

matthchr (approved):

PR #3926, opened 2024-04-09
Refactor to combine samples and refs in samples testing suite

### LINKED ISSUE #3091: Modify samples tests to create resources in parallel

super-harsh:
Closes #3091 

Refactor to combine samples and refs in samples testing suite


matthchr (approved):
Any idea why you only had to re-record these tests?

super-harsh:
@matthchr The re-recorded tests were the ones failing with go-vcr@v1 with "no recording found for an extra recording"

theunrepentantgeek (approved):

PR #3927, opened 2024-04-09
Exclude `v2` and `docs` paths from ADO pipeline PR trigger

super-harsh:
Excluding `v2` and `docs` path for ADO pipeline trigger on PRs. 

Tested out on commit [6ad7510](https://github.com/Azure/azure-service-operator/pull/3927/commits/6ad7510f3ac28459e274d9cf3bd7162874f2748c): 
image

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -14,7 +14,6 @@ pr:
     exclude:
       - docs/
       - v2/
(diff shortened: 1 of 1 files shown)


matthchr:
I think this check is marked required so we can't just exclude it. It should already pass immediately though?

Issue #3637, opened 2023-12-07
Bug: RoleAssignment UUID clashes


Roman-Galeev:
**Version of Azure Service Operator**
v2.4.0

**Describe the bug**
We have development, staging and production subscriptions for deployments, and each of them has a managed kubernetes service (AKS), so the algorithm used to calculate "stable" AzureName UUID produces clashes. This is the function used currently:

```
		assignment.Spec.AzureName = randextensions.MakeUUIDName(
			ownerGK,
			assignment.Spec.Owner.Name,
			gk,
			assignment.Namespace,
			assignment.Name)
```

I believe that "uniqueness" of the UUID could be easily improved by including into the seed string subscription ID (which is known to ASO anyways). However, this will not cover the case when there are more than one cluster in the subcription, but handling that is a more complex story, e.g. ASO could generate an instance ID on install, and use it as a seed.

matthchr:
> I'm open to moving to include subscription in this too, although we'd need to think about how to do so in a way which is backwards compatible as we won't want to orphan existing RoleAssignments.

Actually we wouldn't break existing RoleAssignments because defaulting only applies when they create a new one, it would break the case where they move RoleAssignments between two instances of ASO w/ the same name. Possibly we could document our way around that.

A (likely bigger) problem is that, at the point the webhook runs we don't actually know the subscription and it's actually possible that the subscription configuration doesn't exist yet, so it might be difficult for the webhook to take it into account.

Have you considered just using something like `kustomize` to make your dev/test/prod env roleassignments have slightly different names?

`-dev`, `-test` and `-prod` suffix on the RoleAssignment would solve the problem without you needing to fully manage the UUID yourselves.

Roman-Galeev:
...and another idea is including principalID into role UUID seed, instead of owner.

related: https://github.com/Azure/azure-service-operator/issues/3318

theunrepentantgeek:
There's an interesting conflict here between the two scenarios. In one case, generating a stable UUID for the role identity is crucial to allowing resources to be handed off between ASO instances; in the other, having a different UUID for each installation is needed to keep dev/test/prod environments separate.

I suspect there's no one-size-fits-all solution here; maybe we need to add some configuration to `RoleAssignment` to control how the stable UUIDs are generated. I'll drop this issue into our next milestone and tag as `high-priority` so we don't lose sight of it.

PR #3935, opened 2024-04-11
Add RoleAssignment UUID Generation ADR

super-harsh:

This PR adds ADR for RoleAssignment UUID clashes. ADR discusses about the issue with UUID clashes for RoleAssignment mentioned in #3637. 

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/docs/hugo/content/design/ADR-2024-04-RoleAssignment-UUID-Generation.md
+++ b/docs/hugo/content/design/ADR-2024-04-RoleAssignment-UUID-Generation.md
@@ -11,7 +11,7 @@ The aim for above approach was to:
 - include the owner group, kind, and name to avoid collisions between resources with the same name in different clusters that actually point to different Azure resources.
 
 However, the case where users have multiple ASO instances in multiple clusters with resources in the same namespace, in each cluster having the same name and pointing to different Azure resource is not supported without the user manually giving each RoleAssignment its own UUID.
-The above issue can still be avoided by overriding the AzureName property with random uuids, however it's not consistent and user-friendly.
+The above issue can be avoided by defaulting the AzureName property to a random UUID, however it will cause issues if the RoleAssignment is ever deleted and recreated (or migrated between clusters) as the old UUID will be orphaned.
 
 Issue: [RoleAssignment UUID clashes](https://github.com/Azure/azure-service-operator/issues/3637)
 
--- a/docs/hugo/content/design/ADR-2024-04-RoleAssignment-UUID-Generation.md
+++ b/docs/hugo/content/design/ADR-2024-04-RoleAssignment-UUID-Generation.md
@@ -21,12 +21,12 @@ Add a new property on RoleAssignment spec to control this behaviour.
 
 To use random uuid: 
 ```yaml
-uuidGeneration: Random
+uuidGeneration: random
 ```
 
 To use deterministic uuid(default)
 ```yaml
-uuidGeneration: Default
+uuidGeneration: stable
 ```
 
 **Pro**: Backward compatible
@@ -37,16 +37,17 @@ uuidGeneration: Default
 
 ### Option 2: Using annotations
 
-We can use annotation below to control the name generation behaviour
+We can use annotation below to control the name generation behaviour. 
+In this case, If user does not specify any annotation and does not specify AzureName, ASO sets the default annotation below and follows the default behaviour of stable UUID generation.  
 
 To use random uuid:
 ```yaml
-serviceoperator.azure.com/uuid-generation: Random 
+serviceoperator.azure.com/uuid-generation: random 
 ```
 
 To use deterministic uuid(default)
 ```yaml
-serviceoperator.azure.com/uuid-generation: Default
+serviceoperator.azure.com/uuid-generation: stable
 ```
 
 **Pro**: Backward compatible
--- a/docs/hugo/content/design/ADR-2024-04-RoleAssignment-UUID-Generation.md
+++ b/docs/hugo/content/design/ADR-2024-04-RoleAssignment-UUID-Generation.md
@@ -10,8 +10,7 @@ The aim for above approach was to:
 - include the group and kind to ensure that different kinds of resources get different UUIDs. This isn't entirely required by Azure, but it makes sense to avoid collisions between two resources of different types even if they have the same namespace and name.
 - include the owner group, kind, and name to avoid collisions between resources with the same name in different clusters that actually point to different Azure resources.
 
-However, the case where users have multiple ASO instances in multiple clusters with resources in the same namespace, in each cluster having the same name and pointing to different Azure resource is not supported without the user manually giving each RoleAssignment its own UUID.
-The above issue can be avoided by defaulting the AzureName property to a random UUID, however it will cause issues if the RoleAssignment is ever deleted and recreated (or migrated between clusters) as the old UUID will be orphaned.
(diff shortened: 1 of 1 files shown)


docs/hugo/content/design/ADR-2024-04-RoleAssignment-UUID-Generation.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Option 3: Using subscription ID as seed
+
+Adding subscriptionId as a seed string would make the resource names distinct across subscriptions
+
+**Pro**: User does not have to do anything
+
+**Con**: Moving resource with same name between multiple ASO instances would require a workaround
+**Con**: Webhooks don't have information about subscriptionID
+
```
matthchr, line 65: What option are you recommending?

docs/hugo/content/design/ADR-2024-04-RoleAssignment-UUID-Generation.md
```diff
@@ -0,0 +1,80 @@
...
+**Con**: Users will have to manage the AzureName by themselves while exporting/importing
+
+### Option 3: Using subscription ID as seed
+
+Adding subscriptionId as a seed string would make the resource names distinct across subscriptions
+
+**Pro**: User does not have to do anything
+
+**Con**: Moving resource with same name between multiple ASO instances would require a workaround
```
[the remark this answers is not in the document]
super-harsh: True, will update. 
The case is more when `Moving resource with same name between older to newer ASO version would require a workaround`

docs/hugo/content/design/ADR-2024-04-RoleAssignment-UUID-Generation.md
super-harsh: Have updated!

docs/hugo/content/design/ADR-2024-04-RoleAssignment-UUID-Generation.md
```diff
@@ -0,0 +1,86 @@
...
+
+To use random uuid:
+```yaml
+serviceoperator.azure.com/uuid-generation: Random 
+```
+
+To use deterministic uuid(default)
+```yaml
+serviceoperator.azure.com/uuid-generation: Default
+```
```
matthchr, line 50: I think you should talk here about how we actually do defaulting too:

There are two options:
- If not specified, automatically set `serviceoperator.azure.com/uuid-generation: Default` when a resource is created when defaulting its name.
- If not specified, don't set any annotation (but still generate a UUID as per the `Default` pattern.

I _think_ it's OK to default-in the `uuid-generation` annotation if the user hasn't set it (and also hasn't set the name?). I guess if the name _is_ set by the user, then we wouldn't add the annotation at all because we actualyl didn't do any generation?

and if the user sets it, we'll honor what they set (though it'll not do anything if they set it and the name)?

matthchr (approved):

matthchr:
Approved but had a few comments

docs/hugo/content/design/ADR-2024-04-RoleAssignment-UUID-Generation.md
```diff
@@ -0,0 +1,87 @@
+---
+title: '2024-04: RoleAssignments UUID Generation'
+---
+
+## Context
+
+In ASO, we have made several improvements on making the RoleAssignment's name user-friendly by auto-generating the UUID using UUIDv5 with a seed string based on the group, kind, namespace and name.
```
theunrepentantgeek, line 7:
```suggestion
In ASO, we made several improvements to make RoleAssignment user-friendly by auto-generating their AzureName (which must be a UUID) using UUIDv5 with a seed string based on group, kind, namespace, and name. 
```

docs/hugo/content/design/ADR-2024-04-RoleAssignment-UUID-Generation.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Context
+
+In ASO, we have made several improvements on making the RoleAssignment's name user-friendly by auto-generating the UUID using UUIDv5 with a seed string based on the group, kind, namespace and name. 
+The aim for above approach was to:
+- include the namespace and name to ensure no two RoleAssignments in the same cluster can end up with the same UUID.
+- include the group and kind to ensure that different kinds of resources get different UUIDs. This isn't entirely required by Azure, but it makes sense to avoid collisions between two resources of different types even if they have the same namespace and name.
+- include the owner group, kind, and name to avoid collisions between resources with the same name in different clusters that actually point to different Azure resources.
+
+However, the case where users have multiple ASO instances in multiple clusters with resources in the same namespace, in each cluster having the same name and pointing to different Azure resource is not supported without the user manually giving each RoleAssignment its own UUID.
```
theunrepentantgeek, line 13: I don't understand this sentence.

docs/hugo/content/design/ADR-2024-04-RoleAssignment-UUID-Generation.md
```diff
@@ -0,0 +1,87 @@
...
+
+To use deterministic uuid(default)
+```yaml
+serviceoperator.azure.com/uuid-generation: stable
+```
+
+**Pro**: Backward compatible
+**Pro**: Ease of use
+
+**Con**: Users will have to manage the AzureName by themselves while exporting/importing
```
theunrepentantgeek, line 56:
```suggestion
**Con**: Users will have to manage the AzureName by themselves while exporting/importing
**Con**: Pushes a crucial part of the resource definition into an annotation; the spec is no longer a complete definition of the resource.
**Con**: Annotations are far more easily modified by other tooling, which may have unexpected flow on effects.
```

docs/hugo/content/design/ADR-2024-04-RoleAssignment-UUID-Generation.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Option 3: Using subscription ID as seed
+
+Adding subscriptionId as a seed string would make the resource names distinct across subscriptions
+
+**Pro**: User does not have to do anything
+
+**Con**: Moving resource with same name between older to newer ASO version would require a workaround
```
theunrepentantgeek, line 64: This is a breaking change when someone upgrades the version of ASO in their cluster, as any automated deployments will suddenly experience different behaviour.

docs/hugo/content/design/ADR-2024-04-RoleAssignment-UUID-Generation.md
```diff
@@ -0,0 +1,87 @@
...
+**Con**: Webhooks don't have information about subscriptionID
```
theunrepentantgeek, line 65: Doesn't this automatically disqualify Option 3 from consideration?

docs/hugo/content/design/ADR-2024-04-RoleAssignment-UUID-Generation.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Decision
+
+Recommendation: Option 2 - Using annotations
+
+We retain how the resource shape looks like and suggest using annotation for the users running into the edge case.
```
theunrepentantgeek, line 71: Ergh. Do not like.

theunrepentantgeek (changes requested):

docs/hugo/content/design/ADR-2024-04-RoleAssignment-UUID-Generation.md
super-harsh: Kinda, however worth mentioning it here so we know why we didn't consider this option. Just in case if someone stumbles upon this in future

theunrepentantgeek (approved):

PR #3968, opened 2024-04-23
Add asoctl template command

### LINKED ISSUE #3719: Provide way to fill out crdpattern in raw release YAML
crdpattern is empty so the pod won't come up cleanly

**Describe the improvement**
crdpattern should be specified, at least for the experimental release. Possibly for all releases (though that begs the question what to default it to...), OR there should be some way for users to supply it if it's not going to be defaulted.

matthchr:
Fixes #3719

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/.github/workflows/create-release-experimental.yml
+++ b/.github/workflows/create-release-experimental.yml
@@ -80,8 +80,13 @@ jobs:
 
             > ⚠️ The provided YAML manifest does not configure any CRDs to install by default, but is required.
             > You must specify the CRDs that you want to use as part of `crdPattern`, for example `'resources.azure.com/*;containerservice.azure.com/*;keyvault.azure.com/*;managedidentity.azure.com/*;apimanagement.azure.com/*'`.
+            
+            The recommended way to supply `crdPattern` is using [asoctl template](https://azure.github.io/azure-service-operator/tools/asoctl/#template):
+            ```bash
+            asoctl template https://github.com/Azure/azure-service-operator/releases/download/experimental/azureserviceoperator_experimental.yaml --crd-pattern "<pattern>" | kubectl apply -f -`
+            ```
 
-            It is only intended for developers wishing to try out the latest feature, some of which may not be fully implemented.
+            This release is only intended for developers wishing to try out the latest features, some of which may not be fully implemented.
             
             It is not recommended to run the experimental release for a long period of time, as the docker image referenced by the deployment is 
             `mcr.microsoft.com/k8s/azureserviceoperator:experimental`, which is always being updated. Test what you want to test and then uninstall the operator. 
--- a/docs/hugo/content/tools/asoctl.md
+++ b/docs/hugo/content/tools/asoctl.md
@@ -17,11 +17,19 @@ Usage:
   asoctl [command]
 
 Available Commands:
+  clean       Clean Custom Resource Definitions (CRDs) prior to upgrade
   completion  Generate the autocompletion script for the specified shell
-  crd         Custom Resource Definition (CRD) related actions
+  export      Exports an ASO YAML file from a template
   help        Help about any command
-  import      imports ARM resources as YAML resource definitions
+  import      Imports ARM resources to YAML files containing ASO custom resource definitions
   version     Display version information
+
+Flags:
+      --quiet     Silence most logging
+      --verbose   Enable verbose logging
+  -h, --help      help for asoctl
+
+Use "asoctl [command] --help" for more information about a command.
 ```
 
 ## Installation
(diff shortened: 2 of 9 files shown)


.github/workflows/create-release-experimental.yml
```diff
@@ -80,8 +80,13 @@ jobs:
...
             > ⚠️ The provided YAML manifest does not configure any CRDs to install by default, but is required.
             > You must specify the CRDs that you want to use as part of `crdPattern`, for example `'resources.azure.com/*;containerservice.azure.com/*;keyvault.azure.com/*;managedidentity.azure.com/*;apimanagement.azure.com/*'`.
+            
+            The recommended way to supply `crdPattern` is using [asoctl template](https://azure.github.io/azure-service-operator/tools/asoctl/#template):
+            ```bash
+            asoctl template https://github.com/Azure/azure-service-operator/releases/download/experimental/azureserviceoperator_experimental.yaml --crd-pattern "<pattern>" | kubectl apply -f -`
+            ```
 
-            It is only intended for developers wishing to try out the latest feature, some of which may not be fully implemented.
+            This release is only intended for developers wishing to try out the latest feature, some of which may not be fully implemented.
```
theunrepentantgeek, line 89:
```suggestion
            This release is only intended for developers wishing to try out the latest features, some of which may not be fully implemented.
```

docs/hugo/content/tools/asoctl.md
```diff
@@ -94,6 +102,34 @@ Append that directory to your `PATH` if desired.
...
 
+## Template
+
+The `template` commands helps render a YAML file from the raw ASO published with each release.
+
+```bash
+template creates a YAML file from the specified ASO yaml template
+
+Usage:
+  asoctl template <path to ASO YAML> [flags]
```
theunrepentantgeek, line 113: Lots of thoughts.

Suggest the command should be `asoctl export template`, both to allow for exporting of other things, and to follow the approach of `asoctl <verb> <noun>` that we standardised on originally.

Given the number of potential URLs from which to download the required YAML is very low, would be nice to have less verbose options. As discussed offline, maybe

`--source <file|url>` to support the current behaviour
`--tag <tag>` to allow users to specify just the bit of interest (e.g. `asoctl export template --tag v2.7.0`)

theunrepentantgeek: Can `--crd-pattern` be repeated? If so, include an example of that too?

v2/cmd/asoctl/cmd/template.go
```diff
@@ -0,0 +1,79 @@
...
+			ctx := cmd.Context()
+			uri := args[0]
+
+			result, err := template.Get(ctx, uri)
+			if err != nil {
+				return err
+			}
+
+			if !options.raw {
+				// Render the CRDPattern w/ `;` separating them
```
theunrepentantgeek, line 46: Any chance of validating the patterns here, to reject typos/mistakes

```
$ asoctl export template --tag v2.9.0 --crd-pattern compoot/*
```

theunrepentantgeek: You can see how to approach creating a map to do the checks in [find_group_kind.go](https://github.com/Azure/azure-service-operator/blob/main/v2/cmd/asoctl/internal/importing/find_group_kind.go#L38).

v2/cmd/asoctl/cmd/template.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// newTemplateCommand creates a new cobra Command when invoked from the command line
+func newTemplateCommand() (*cobra.Command, error) {
+	var options templateOptions
+
+	cmd := &cobra.Command{
+		Use:   "template <path to ASO YAML>",
+		Short: "Template creates a YAML file from the specified ASO yaml template",
+		Example: `asoctl template https://github.com/Azure/azure-service-operator/releases/download/v2.6.0/azureserviceoperator_v2.6.0.yaml --crd-pattern "resources.azure.com/*;containerservice.azure.com/*;keyvault.azure.com/*;managedidentity.azure.com/*;eventhub.azure.com/*"
```
super-harsh, line 29: Would be good if user doesn't have to specify the whole url. We can just ask them for the version they want, and construct a URL in asoctl?

v2/cmd/asoctl/internal/template/template.go
```diff
@@ -0,0 +1,83 @@
+package template
```
super-harsh, line 1: Missing header file here, also the reason why CI failed.

docs/hugo/content/tools/asoctl.md
matthchr: > Suggest the command should be asoctl export template, both to allow for exporting of other things, and to follow the approach of asoctl <verb> <noun> that we standardised on originally.

I've done this but I actually think it's somewhat awkward how we've got it today. If you look for example at `asoctl import` it's documented as: "Imports ARM resources to YAML files containing ASO custom resource definitions" - which isn't really a "generic" import, it's the description of the operation inside `import`.

It's not clear to me what I should write for the `export` documentation... The command is supposed to be in theory generic (could export many things) but it seems difficult to write a useful description of it that doesn't have intimate knowledge about what the subcommands are.

matthchr: Added repeated `--crd-pattern` examples.

matthchr: swapped `--tag` to `--version` as I think that's slightly clearer? LMK what you think.

v2/cmd/asoctl/cmd/template.go
matthchr: Changed this

v2/cmd/asoctl/cmd/template.go
matthchr: I got partway through doing this but realized it won't all that well because asoctl isn't necessarily built with the same version as the target version. 

For example let's say that the user is attempting to install the `experimental` release to test a new CRD they've just added, if they try with an older asoctl, we would reject them because their group doesn't exist. They'd need to download the experimental asoctl and use that.

The problem works the opposite way too, if the user is using a newer asoctl but installing an older release, we'd allow crdpatterns that won't actually work for the operator.

We could maybe solve this problem by doing our best in asoctl and then having a `--novalidate` flag or something that bypassed validation for crdpattern if the user really knows what they're doing... but really the crdpattern validation seems somewhat at odds with letting the user select a version that may not match the asoctl version.

I think let's table this for now and we can investigate it further in a future release?

matthchr: https://github.com/Azure/azure-service-operator/issues/3969 tracking this, we can discuss next week.

super-harsh (approved):

docs/hugo/content/tools/asoctl.md
theunrepentantgeek: I'm happy with that. I guess `tag` is more of an implementation detail anyway.

v2/cmd/asoctl/cmd/template.go
```diff
@@ -0,0 +1,124 @@
...
+
+type templateOptions struct {
+	source  string
+	version string
+
+	raw        bool
+	crdPattern []string
+}
+
+var expectedTagRegex = regexp.MustCompile(`^v\d.\d.\d$`)
```
theunrepentantgeek, line 27: Those periods will match any one character - to match a literal period, you need to quote them:

```suggestion
var expectedTagRegex = regexp.MustCompile(`^v\d\.\d\.\d$`)
```

theunrepentantgeek (approved):

v2/cmd/asoctl/cmd/template.go
matthchr: Ahh good catch

PR #3974, opened 2024-04-28
Update k8s.io dependencies

super-harsh:
Fixes: #3954, #3955, #3957, #3958, #3959, #3961, #3962


v2/cmd/asoctl/go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/Azure/azure-service-operator/v2/cmd/asoctl
 
-go 1.21
+go 1.22.0
```
matthchr, line 3: We should maybe check with @nojnhuh on how much of a problem this would be for CAPZ.

matthchr (approved):

v2/cmd/asoctl/go.mod
nojnhuh: We're blocked on importing ASO v2.7 for other reasons that might not be resolved until we import v0.31 of our k8s.io dependencies, so don't wait for us.
https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4709#issuecomment-2080050652

theunrepentantgeek (approved):

PR #4003, opened 2024-05-08
Update local tests and CI tests to use az login token or managed identity, not service principal

### LINKED ISSUE #3930: Use Federated Identity Credentials rather than SP for GitHub Actions
Steps:
- [ ] Swap the `az-login` step in the Taskfile to use our interactive credentials rather than SP when run in interactive mode. This should mean we don't need a personal SP at all.
- [ ] Use OIDC Issuer + ManagedIdentity during CI (see [GH actions az-login](https://github.com/marketplace/actions/azure-login#login-with-openid-connect-oidc-recommended), [setting up GH OIDC](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-azure), and [Weinong's example](https://github.com/weinong/azure-federated-identity-samples/tree/feat/gh-oidc-env-var))
- [ ] Dynamically create the MULTITENANT credentials rather than having them hardcoded?

matthchr:
* Update Taskfile to use az cli token rather than SP 
  * Update az login step to use an interactive login, rather than an SP.
    Interactive login is good for 12 hours.
  * Remove SP-based kind cluster helpers and replace usages with
    workload-identity based.
  * Refactor make-mi-fic.sh to Python. This is required to manage the
    more limited permissions that an interactive token has with repsect
    to calling 'az role assignment list'.
* HTTP Recording tests now use CLI or ServicePrincipal credentials
  * CLI credentials are preferred over ServicePrincipal.
* Add unit tests with mocked azidentity calls 
  * This replaces actual live testing of the various
    authentication modes. We trust that, if we call the azidentity SDK with
    the correct parameters, it does the correct thing.
* Fix issue with credential format documentation 
  * We documented the AUTH_MODE: "workloadidentity" for workload identity,
    but in actuality the code doesn't require this be set for WI on any
    secret but the global secret.
* Remove use of MULTITENANT secret
  * Now using a dynamically created UserAssignedIdentity instead of a
    hardcoded static multitenant secret.
  * There are two advantages here:
    1. Less setup to run a test on a new subscription because fewer inputs
       to the CI job.
    2. No static secrets that can possibly be leaked or compromised.
* Use MSI for CI instead of service principal
  * Updated cipool BICEP to provision a managed identity and give it the
    required roles.
  * Updated Taskfile to support MSI-based az login when running in the
    context of a GitHub action.
  * Removed all reference to AZURE_CLIENT_ID and AZURE_CLIENT_SECRET from
    GitHub workflows.
* Update documentation to be more clear about authentication mode
  * Be explicit about which mode we're using in the various installation
    instructions, and provide links to the detailed authentication
    documentation so that users are more easily exposed to other
    authentication options (formats and scopes).

Closes #3930

**If applicable**:
- [x] this PR contains tests


Taskfile.yml
```diff
@@ -324,42 +327,55 @@ tasks:
...
+  controller:test-upgrade-setup:
+    desc: Setup for upgrade test
     dir: "{{.CONTROLLER_ROOT}}"
     cmds:
-      - task: controller:kind-create
+      - task: controller:kind-create-wi
+      - task: controller:create-mi-for-workload-identity
       - task: controller:install-cert-manager
       - task: controller:docker-push-local
       - task: controller:gen-helm-manifest
```
theunrepentantgeek, line 338: Do these have to be serial like this? We should be preferring `deps` over calling tasks like this, where possible.

Taskfile.yml
```diff
@@ -324,42 +327,55 @@ tasks:
...
       # We need the below to wait until cert-manager is up, otherwise the subsequent installation of webhooks fails. See https://cert-manager.io/next-docs/installation/verify/
       - "cmctl check api --wait=2m"
+
+  controller:test-upgrade-pre:
+    desc: Test upgrading {{.CONTROLLER_APP}} and helm chart for new release
+    dir: "{{.CONTROLLER_ROOT}}"
+    cmds:
       - "helm repo add asov2 https://raw.githubusercontent.com/Azure/azure-service-operator/main/v2/charts"
       - "helm repo update"
-      - "helm upgrade --install --devel aso2 asov2/azure-service-operator \
```
theunrepentantgeek, line 339: Don't we need the `--devel` flag here? I have some recollection that we needed it for installs to actually work.

Taskfile.yml
```diff
@@ -750,44 +769,54 @@ tasks:
...
     vars:
+      DIR: "{{default .KIND_WORKLOAD_IDENTITY_PATH .DIR}}"
       RESOURCE_GROUP:
-        sh: "cat ./out/workload-identity/azure/rg.txt"
+        sh: |
+          if [ ! -z "{{.RESOURCE_GROUP}}" ]; then
+            echo "{{.RESOURCE_GROUP}}"
+          else
+            cat ./{{.DIR}}/azure/rg.txt
+          fi
```
theunrepentantgeek, line 803: This can be done more easily using Taskfile templating. Looks something like this:

``` yaml
RESOURCE_GROUP_FILE: 
    sh: cat ./{{.DIR}}/azure/rg.txt
RESOURCE_GROUP: "{{default .RESOURCE_GROUP .RESOURCE_GROUP_FILE}}"
```

Taskfile.yml
```diff
@@ -943,7 +979,17 @@ tasks:
       # This below is added to test if ASO works fine in namespaces other than 'azureserviceoperator-system"
```
theunrepentantgeek, line 979: I thought we'd removed this option - do we need this?

Taskfile.yml
```diff
@@ -943,7 +979,17 @@ tasks:
...
+      - task: controller:create-mi-for-workload-identity
+        vars:
+          # Use the tenant1-system namespace as that's where the single tenant operator will be installed
+          SUBJECT: "system:serviceaccount:tenant1-system:azureserviceoperator-default"
+          # We have to set RESOURCE_GROUP and ISSUER here because if we don't, the cached variables
+          # will be used from the create-mi-for-workload-identity task in ci. See: https://github.com/go-task/task/issues/524
+          RESOURCE_GROUP:
+            sh: "cat ./{{.KIND_WORKLOAD_IDENTITY_PATH}}/azure/rg.txt #controller:kind-create-multitenant-cluster-helm"
+          ISSUER:
+            sh: "cat ./{{.KIND_WORKLOAD_IDENTITY_PATH}}/azure/saissuer.txt #controller:kind-create-multitenant-cluster-helm"
```
theunrepentantgeek, line 992: I think you can use `run: always` on the task to avoid this workaround.

.github/workflows/create-release-official.yml
```diff
@@ -68,11 +66,9 @@ jobs:
       - name: Protect image
         run: |
           container_id=${{env.container_id}}
-          docker exec -e DOCKER_PUSH_TARGET -e DOCKER_REGISTRY -e AZURE_TENANT_ID -e AZURE_CLIENT_ID -e AZURE_CLIENT_SECRET -e AZURE_SUBSCRIPTION_ID "$container_id" task controller:acr-protect-image
+          docker exec -e GITHUB_ACTIONS -e DOCKER_PUSH_TARGET -e DOCKER_REGISTRY -e AZURE_TENANT_ID -e AZURE_SUBSCRIPTION_ID "$container_id" task controller:acr-protect-image
```
super-harsh, line 69: Where's this `GITHUB_ACTIONS` var is sourced from?

Taskfile.yml
super-harsh: or we run the CI tasks with `--force` always?

Taskfile.yml
matthchr: https://helm.sh/docs/helm/helm_install/#options

It's for beta versions, since we're GA now we don't need it anymore.

Taskfile.yml
matthchr: The bits I added/swapped (king-create-wi and create-mi-for-workload-identity) do need to be run in serial, yes, as the first produces output consumed by the second.

install-cert-manager also must run after the first 2.

We could in theory move `docker-push-local` and `gen-helm-manifest` to deps, but my focus here was on changing auth not necessarily improving build parallelism and the parts that have been changed do need to run in sequence. Given the priority on this auth stuff I think it makes sense to focus on that for this PR as I am not 100% sure that moving other things around to deps/etc will trivially work.

Taskfile.yml
matthchr: No, I think we talked about it but then realized it actually worked and so instead of removing it we added a test to ensure that it works.

Taskfile.yml
matthchr: The issue isn't that the task doesn't run. The task runs. See https://github.com/go-task/task/issues/524 (linked in the above comment) for why this is needed.

.github/workflows/create-release-official.yml
matthchr: GitHub itself includes it in all the action contexts. It's how you can easily tell if you're running in the context of an action versus locally.

Taskfile.yml
```diff
@@ -802,6 +823,8 @@ tasks:
       - az-login
     cmds:
       - "az group delete --name {{.RESOURCE_GROUP_NAME}} -y"
+      - "{{.SCRIPTS_ROOT}}/delete-role-assignment.sh -d {{.AKS_WORKLOAD_IDENTITY_PATH}}"
+      - "rm -rf {{.AKS_WORKLOAD_IDENTITY_PATH}}"
```
theunrepentantgeek, line 827: Do we need `-rf` here? The combination can be dangerous if `AKS_WORKLOAD_IDENTITY_PATH` happens to get an odd value.

theunrepentantgeek (approved):

super-harsh (approved):

Taskfile.yml
matthchr: Unfortunately yes I think we do need `-rf` here. `-r` because there's 2 levels of subfolder to remove and `-f` because otherwise it'll prompt which doesn't work in unattended scenarios.

I did some digging though and realized that now rm's default is:
>       --preserve-root[=all]  do not remove '/' (default);
>                              with 'all', reject any command line argument
>                              on a separate device from its parent

Which means actually, you can't blat your whole system with this anymore so it's much safer than it used to be. You _can_ still accidentally delete directories you didn't mean to, but in this case I think we're pretty safe because we've hardcoded it to a specific variable `.AKS_WORKLOAD_IDENTITY_PATH` and this path won't ever be unset unless we delete that variable.

matthchr: and if the variable is just deleted, we'll get this:

> rm: missing operand
> Try 'rm --help' for more information.

PR #4030, opened 2024-05-22
Use Azure Storage static web serving rather than public access

matthchr:
**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/scripts/v2/create-kind-wi-storage.sh
+++ b/scripts/v2/create-kind-wi-storage.sh
@@ -33,7 +33,9 @@ fi
 mkdir -p "${DIR}/azure"
 RESOURCE_GROUP="${PREFIX}-rg-wi$(openssl rand -hex 6)"
 AZURE_STORAGE_ACCOUNT="asowi$(openssl rand -hex 6)"
-AZURE_STORAGE_CONTAINER="oidc-test"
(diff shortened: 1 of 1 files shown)


scripts/v2/create-kind-wi-storage.sh
```diff
@@ -42,24 +44,26 @@ if [ -f "$DIR/azure/rg.txt" ]; then
   exit 0
 fi
 
+ISSUER_URL="https://${AZURE_STORAGE_ACCOUNT}.z22.web.core.windows.net/"
```
nojnhuh, line 47: Is this `z22` deterministic? When I just tried out these steps manually mine happens to be `z13`.

matthchr: Thats a good question... it may not be fixed let me see if I can get it dynamically

nojnhuh: Here's a command that shows me my URL if that's helpful:
```
az storage account show --name "${AZURE_STORAGE_ACCOUNT}" -o json | jq -r .primaryEndpoints.web
```

matthchr: Fixed this, thanks!

super-harsh (approved):

PR #4043, opened 2024-05-28
Simplify helm chart build process and add validation check

### LINKED ISSUE #2778: Improve chart and release YAML build automation
We currently use Kustomize but have some scripts wrapping it for various purposes. This is somewhat counter to how Kustomize _should_ be used, although we were led into it somewhat by the default `kubebuilder` Kustomize structure.

We use Kustomize for 3 things:
- To produce CRDs to embed into the operator pod (see #2769)
- To produce deployment, service, rbac, etc files for usage in the raw YAML we release. There's a small amount of `sed` replace that happens after this to replace the `manager:latest` image with an actual image from MCR.
- To produce those same deployment, service, rbac, etc files for usage in Helm. Currently this is done by running Kustomize and then doing a bunch of `sed`/`perl` replacements, which works but is awkward

**Describe the improvement**
We need a less ad-hoc process for this whole thing. There are [some articles](https://jfrog.com/blog/power-up-helm-charts-using-kustomize-to-manage-kubernetes-deployments/) that talk about using Kustomize to generate Helm templates - that's a possible route we can investigate. We should also investigate other tools in this space.

super-harsh:
Closes #2778 

This PR includes the steps to simplify helm build automation by enabling to manage helm chart with no dependency on generation script and have the helm chart templates checked into the repo. 
This introduces a need for validation script which runs against the Kustomize generated YAML to make sure that helm chart is in sync with the same.


scripts/v2/validate_helm.py
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env python3
+import subprocess
+
+subprocess.run("pip install deepdiff".split(" "))
```
super-harsh, line 4: Should this be in `install-dependencies.sh` instead?

v2/charts/azure-service-operator/templates/generated/apps_v1_deployment_azureserviceoperator-controller-manager.yaml
```diff
@@ -0,0 +1,203 @@
...
+        kubectl.kubernetes.io/default-container: manager
+      labels:
+   {{- if or (eq .Values.multitenant.enable false) (eq .Values.azureOperatorMode "watchers") }}
+        aadpodidbinding: aso-manager-binding
+   {{- end }}
+        app.kubernetes.io/name: azure-service-operator
+        app.kubernetes.io/version: v2.7.0
+        control-plane: controller-manager
+    spec:
+      {{- with .Values.tolerations }}
```
matthchr, line 30: minor: Fix indentation here (the `- with`) now that it's handcrafted/maintained?

v2/config/manager/manager.yaml
```diff
@@ -38,7 +38,7 @@ spec:
       serviceAccountName: default
       containers:
       - args:
-        - --metrics-addr=:8443
+        - --metrics-addr=0.0.0.0:8443
```
matthchr, line 41: minor: Can you explain why we need this change?

matthchr: Ah, it's because that's the default in the Helm chart, right?

Taskfile.yml
```diff
@@ -677,8 +676,11 @@ tasks:
     dir: "v2/"
     cmds:
       - mkdir -p bin # in case it doesn't exist
-      - "{{.SCRIPTS_ROOT}}/kustomize-build.sh -k operator -v {{.VERSION}} | sed -e 's@{{.LOCAL_REGISTRY_CONTROLLER_DOCKER_IMAGE}}@{{.PUBLIC_REGISTRY}}{{.CONTROLLER_DOCKER_IMAGE}}@g' > bin/azureserviceoperator_{{.VERSION}}.yaml"
-      - "{{.SCRIPTS_ROOT}}/kustomize-build.sh -k crd -v {{.VERSION}} > bin/azureserviceoperator_customresourcedefinitions_{{.VERSION}}.yaml"
+      - "{{.SCRIPTS_ROOT}}/kustomize-build.sh -k operator -v {{.VERSION_TAG}} | sed -e 's@{{.LOCAL_REGISTRY_CONTROLLER_DOCKER_IMAGE}}@{{.PUBLIC_REGISTRY}}{{.CONTROLLER_IMAGE}}@g' > bin/azureserviceoperator_{{.VERSION_TAG}}.yaml"
+      - "{{.SCRIPTS_ROOT}}/kustomize-build.sh -k crd -v {{.VERSION_TAG}} > bin/azureserviceoperator_customresourcedefinitions_{{.VERSION_TAG}}.yaml"
+    vars:
+      VERSION_TAG: "{{default .VERSION .VERSION_TAG}}"
```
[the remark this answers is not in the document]
theunrepentantgeek: Yes, variables for the task can override tasks from the environment.

scripts/v2/package-helm-manifest.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+# This script generates helm manifest and replaces required values in helm chart.
+
+set -e
+
+LOCAL_REGISTRY_CONTROLLER_DOCKER_IMAGE=$1
+PUBLIC_REGISTRY=$2
+VERSION=$3
+DIR=$4
```
theunrepentantgeek, line 10: Do a proper parsing loop to handle commandline arguments and provide a usage message if the user gets it wrong.

Examples
* [clean-orphan-role-assignments.sh](https://github.com/Azure/azure-service-operator/blob/main/scripts/v2/clean-orphan-role-assignments.sh#L23)
* [create-kind-wi-storage.sh](https://github.com/Azure/azure-service-operator/blob/main/scripts/v2/create-kind-wi-storage.sh#L18)
* [delete-kind-wi-storage.sh](https://github.com/Azure/azure-service-operator/blob/main/scripts/v2/delete-kind-wi-storage.sh#L17)

v2/charts/azure-service-operator/templates/generated/apps_v1_deployment_azureserviceoperator-controller-manager.yaml
super-harsh: Did you mean `{{ - with .Values.tolerations }}` or the yaml indentation level?

v2/config/manager/manager.yaml
super-harsh: Yep

Taskfile.yml
```diff
@@ -335,7 +335,6 @@ tasks:
       - task: controller:create-mi-for-workload-identity
       - task: controller:install-cert-manager
       - task: controller:docker-push-local
-      - task: controller:gen-helm-manifest
```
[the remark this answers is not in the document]
super-harsh: Nice catch! I've replaced this with `controller:package-helm-manifest`.

Taskfile.yml
```diff
@@ -1003,14 +1016,25 @@ tasks:
       # Wait for the other operator to be ready - not using controller:wait-for-operator-ready because that also waits for CRDs, etc which this doesn't install
       - "kubectl wait --for=condition=ready --timeout=2m pod -n tenant1-system -l control-plane=controller-manager"
 
+  controller:kind-create-helm:
```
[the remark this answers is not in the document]
super-harsh: Ah! this wasn't supposed to be there. I think some merge conflicts might've added it back. Since I made some changes, and it was removed in another PR. 
PS. I've removed it now. Thanks for pointing out.

scripts/v2/requirements.txt
```diff
@@ -0,0 +1,3 @@
+
+deepdiff
+pyyaml
```
matthchr, line 3: minor: missing newline at the end of requirements.txt

.devcontainer/install-dependencies.sh
```diff
@@ -308,7 +314,12 @@ if [ "$VERBOSE" == true ]; then
     echo "Installed tools: $(ls "$TOOL_DEST")"
 fi
 
-if [ "$DEVCONTAINER" == true ]; then 
+# python packages to be installed for helm validation script
+# using `--break-system-packages` here since the python3 env is externally managed through the OS, so we have to pass this to install the dependency for local environment.
+pip3 install deepdiff --break-system-packages
```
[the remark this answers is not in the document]
matthchr: @super-harsh pointed out a few challenges with requirements.txt in the Dockerfile.

1. Docker build can't easily access files outside of the build context (`.devcontainer` folder).
2. Docker build doesn't share context with the host easily, which means if we were to create a venv here it probably wouldn't exist on the host filesystem. It would be nice if it could exist on the host filesystem though as that way it works in builds and for IDEs (for those like me using the host for their IDE)

I think the right thing to do here is to split the management of the Python dependencies required into two sections.
**Section 1:** In the Dockerfile/setup.sh, install virtualenv - this you have already and is required to use venv on either host or docker container so it makes sense to be in the main Dockerfile setup.
**Section 2:** Put requirements.txt in `v2/scripts` and use it _in the taskfile_, not in the Dockerfile. I think this makes sense because we could in theory have multiple Python projects with multiple sets of dependencies so doing a single set of Python dependency management in the Dockerfile is wrong. In the same way that we don't install the Go dependencies for our projects in the Dockerfile - we let the Go tooling do that when we run `go build` via the Taskfile.

I think accomplishing this boils down to making a Taskfile target that does something like:
```
  scripts:build-venv:
    desc: Creates a Python virtual env
    dir: {{.SCRIPTS_ROOT}}
    cmds:
      - python3 -m virtualenv venv
      - source venv/bin/activate
      - pip install -r requirements.txt
      - deactivate
```

...
  controller:validate-helm:
    desc: Validates and compares helm against the ASO yaml generated by kustomize and ensures that they match
    dir: "{{.CONTROLLER_ROOT}}"
    deps:
      - task: controller:run-kustomize
        vars:
          VERSION_TAG: "{{.LATEST_VERSION_TAG}}"
          CONTROLLER_IMAGE: "azureserviceoperator:{{.LATEST_VERSION_TAG}}"
      - task: scripts:build-venv
    cmds:
      - defer:
          deactivate  # deactivate the venv even if the script fails
      - source venv/bin/activate # activate the venv
      - "{{.SCRIPTS_ROOT}}/validate_helm.py --helm-dir=$PWD/charts/azure-service-operator --yaml-path=$PWD/bin/azureserviceoperator_{{.LATEST_VERSION_TAG}}.yaml"
```

File: v2/charts/azure-service-operator/templates/generated/apps_v1_deployment_azureserviceoperator-controller-manager.yaml (hunk 2 above)
Comment (matthchr) [reply v2/charts/azure-service-operator/templates/generated/apps_v1_deployment_azureserviceoperator-controller-manager.yaml]: This: `{{- if .Values.podAnnotations }}` and the `{{- end }}` don't match the indentation of the other ifs in this file, basically is what I was getting at.

Some other files have similar wonkiness.

Since these are now handcrafted rather than generated via script I basically am wondering if it's worth doing a pass through them and moving all of the `if`s to have no indentation at all (always left-hand justified)? Or making them all line up with the indentation of the thing they're surrounding, but I think that's a bit harder to read so probably no indentation at all is better?

I'm not sure which way is idiomatic Helm and I don't know that it matters all that much which we pick, but I think we should be consistent, either:
```
   {{- if not .Values.multitenant.enable }}
  name: azureserviceoperator-manager-rolebinding
   {{ else }}
```

OR

```
{{- if not .Values.multitenant.enable }}
  name: azureserviceoperator-manager-rolebinding
{{ else }}
```

I have a slight preference towards the second but have a strong preference towards consistency of some kind (even if it's the first).

Comment (super-harsh) [reply v2/charts/azure-service-operator/templates/generated/apps_v1_deployment_azureserviceoperator-controller-manager.yaml]: Yep, sounds good. I can have a take on fixing the indentations in the following PR with #3837 if that's alright?

Comment (matthchr) [reply v2/charts/azure-service-operator/templates/generated/apps_v1_deployment_azureserviceoperator-controller-manager.yaml]: sgtm

File: Taskfile.yml [hunk 10]
```diff
@@ -719,13 +723,34 @@ tasks:
...
-    deps: 
+    deps:
+      - task: controller:run-kustomize
+        vars:
+          VERSION_TAG: "{{.LATEST_VERSION_TAG}}"
+          CONTROLLER_IMAGE: "azureserviceoperator:{{.LATEST_VERSION_TAG}}"
+      - task: scripts:build-python-venv
+    defer:
+    cmds:
+      - "{{.SCRIPTS_ROOT}}/venv/bin/python3 {{.SCRIPTS_ROOT}}/validate_helm.py --helm-dir=$PWD/charts/azure-service-operator --yaml-path=$PWD/bin/azureserviceoperator_{{.LATEST_VERSION_TAG}}.yaml"
```
matthchr, line 744: Nice this works pretty well.

scripts/v2/requirements.txt
```diff
@@ -0,0 +1,3 @@
+
```
matthchr, line 1: minor: excess newline at start of file.

matthchr (approved):

PR #4055, opened 2024-06-01
Improve package import aliasing

theunrepentantgeek:

While working on moving the ARM types into a subpackage called `arm`, I noticed that referencing ARM types as `v20240531a.Account` instead of `Account_ARM` was considerably less clear.

This PR changes the way subpackages are imported to use the name of the subpackage (e.g. `arm`) if that works without causing conflicts. This will allow ARM types to be referenced as `arm.Account` which is nice and clear, and has the welcome side effect of changing many of our current references to `storage`, adding clarity to the existing code.

gif

**If applicable**:
- [x] this PR contains tests


v2/tools/generator/internal/astmodel/function.go
```diff
@@ -15,6 +15,8 @@ type Function interface {
 	// (You can't have two functions with the same name on the same object or resource)
 	Name() string
 
+	// RequiredPackageReferences returns a set of packages imports required by this type.
```
matthchr, line 18: hah. nice - I've gotten confused by this too.

v2/tools/generator/internal/astmodel/package_import_set.go
```diff
@@ -184,23 +186,62 @@ func (set *PackageImportSet) orderImports(i PackageImport, j PackageImport) bool
...
+func (set *PackageImportSet) tryAssignImportAliases(
+	style PackageImportStyle,
+) bool {
+	// aliased keeps track of every import that we've assigned an alias to, keyed by the new alias.
+	// This allows us to check for duplicates.
+	aliased := make(map[string]PackageImport, len(set.imports))
+
+	// groupConflicts keeps track of any group for which we have conflicted aliases.
+	// This allows us to add aliases for groups where we don't have any conflicts.
+	groupConflicts := make(map[string]bool, 100)
```
matthchr, line 201: minor, I think you can avoid having a default szie here and the linter will be OK with it because the addition to the map is guarded by 1 or more complex ifs. If you want to have a default size, 100 seems kinda big? Do we really expect anywhere close to 100 conflicts in an average ASO file? I'd think something like 10 would be P95?

matthchr, line 201: minor: Can you use our `set.Set[T]` here rather than a map to bool?

matthchr (approved):

v2/tools/generator/internal/astmodel/package_import_set.go
theunrepentantgeek: I'll remove the default sizes and see if the linter complains.

theunrepentantgeek: Done - with a workaround to avoid the receivers on this type also being `set`.

theunrepentantgeek: Linter seems happy.

PR #4063, opened 2024-06-05
Fix issue where ok-to-test couldn't comment on PR comments

matthchr:


.github/workflows/ok-to-test.yml
```diff
@@ -9,6 +9,8 @@ on:
 jobs:
   ok-to-test:
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write  # Required to comment on PR comments
```
theunrepentantgeek, line 13: Does this work? I thought our `GITHUB_TOKEN` was locked down to readonly access.

theunrepentantgeek (approved):

.github/workflows/ok-to-test.yml
matthchr: We specify other `write` perms, just not `pull-requests: write` atm. I think it's readonly _by default_, but you can elevate. We elevate frequently for things like `contents: write` to upload releases.

PR #4065, opened 2024-06-06
Fix breaking changes documentation ordering to be consistent

matthchr:

**If applicable**:
- [ ] this PR contains tests


theunrepentantgeek (commented):
Won't this make them sequential, with the latest at the bottom?

I thought we wanted them in reverse order, latest at the top.

matthchr:
> Won't this make them sequential, with the latest at the bottom?
> I thought we wanted them in reverse order, latest at the top.

It would be nice to have them backwards-ordered. Right now the issue is they're incorrectly ordered, which is worse than either forwards or backwards... let me see if I can make them ordered the way we want but in a way that's harder to mess up.

theunrepentantgeek (approved):

PR #4066, opened 2024-06-07
Deterministic Testcase ordering

theunrepentantgeek:

Turns out that our test case ordering wasn't deterministic.

When calling `NewTestFileDefinition()` to generate a `_test.go` file, the slice of `TypeDefinition` passed in comes directly from a map (via `maps.Values()`) , and therefore has non-deterministic ordering. 

I'm surprised we didn't observe this before, I'm guessing there's a change in `go 1.22.2` as I just updated to that version (was running `go 1.22.1` previously.

gif


matthchr (approved):
Code generation is funny.

PR #4067, opened 2024-06-07
Add support for RoleDefinition resource

### LINKED ISSUE #2570: Feature: Create RoleDefinitions
The resource is `Microsoft.Authorization/roleDefinitions`.

An example ARM template is here: https://learn.microsoft.com/en-us/azure/role-based-access-control/custom-roles-template

matthchr:
This fixes #2570.

Closes #[issue number]

gif

**If applicable**:
- [ ] this PR contains tests


v2/internal/util/randextensions/uuid5.go
```diff
@@ -13,27 +13,38 @@ import (
...
+// * Namespace
+// * Name
+func MakeUniqueResourceScopedString(gk schema.GroupKind, namespace string, name string) string {
+	return fmt.Sprintf("%s/%s:%s/%s", gk.Group, gk.Kind, namespace, name)
 }
 
-// TODO: Fix name
-// MakeUniqueResourceString generates a string that uniquely identifies a cluster resource.
-func MakeUniqueResourceString2(ownerGK schema.GroupKind, ownerName string, gk schema.GroupKind, namespace string, name string) string {
+// TODO: This method has a bug where it is called with an empty owner gk when the owner is an ARM ID.
```
matthchr, line 26: I'm not 100% sure what we should do about this bug. We can fix it (which could cause people to leak resources? Maybe not because our validation will protect us possibly? I need to do some testing...)

v2/internal/util/randextensions/uuid5.go
```diff
@@ -13,27 +13,38 @@ import (
...
-// MakeUniqueResourceString generates a string that uniquely identifies a cluster resource.
-func MakeUniqueResourceString(group string, kind string, namespace string, name string) string {
-	return fmt.Sprintf("%s/%s:%s/%s", group, kind, namespace, name)
+// MakeUniqueResourceScopedString generates a string that uniquely identifies a cluster resource. It includes the
+// following distinguishing parts:
+// * Group
+// * Kind
+// * Namespace
+// * Name
+func MakeUniqueResourceScopedString(gk schema.GroupKind, namespace string, name string) string {
```
matthchr, line 22: Not clear to me if we want to just not consider the owner ID in the case of RoleDefinition or not... the more I think about it the more I think maybe we should be considering owner here too.

theunrepentantgeek (approved):

v2/internal/util/randextensions/uuid5.go
matthchr: I filed #4079 to track this in 2.9.0 and fixed this bug but _only for the new resource RoleDefinition_. We can track fixing it for existing resources as part of 2.9.0 with the above bug.

theunrepentantgeek (approved):

PR #4091, opened 2024-06-16
Fix for missed secret configuration 

### LINKED ISSUE #4088: Question  ManagedCluster - Can credential for Windows & Linux node be passed as secret instead of plain text
Can credential for Windows & Linux node be passed as secret instead of plain text ?

**Current supported**
```
# Linux node public key profile
      linuxProfile:
        adminUsername: "azureuser"
        ssh:
          publicKeys:                                        # SSH Key for node
          - keyData: |
              ssh-rsa xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
# Windows node credential profile
      windowsProfile:
        adminUsername: "azureuser"
        adminPassword: "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
```

**Expectation**

```
# Linux node public key profile
      linuxProfile:
        adminUsername: "azureuser"
        ssh:
          publicKeys:                                        # SSH Key for node
          - keyData:
               secret:
                  key:
                  value:
# Windows node credential profile
      windowsProfile:
        adminUsername: "azureuser"
        adminPassword:
           secret:
              key:
              value:

```

theunrepentantgeek:

Adds configuration for `ManagedClusterWindowsProfile` to mark `adminPassword` as a secret.

Fixes #4088

Also adds a heuristic to flag potential fields that should be explicitly configured with `$isSecret`, to ensure we don't miss one in the future.

gif


matthchr (approved):
I did some checking and don't see much usage of this property through ASOv2 so I think this is OK to change in a breaking way.

PR #4107, opened 2024-06-17
Expose default primary PrivateEndpoint privateIpAddress

### LINKED ISSUE #3901: Bug: `ipConfigurations` field is not populated in `PrivateEndpoint` status

ASO V2 `v2.6.0`

**Describe the bug**
When a private endpoint provisioned by ASO doesn't populate the `ipConfigurations` field in its status.  We expect this field to be present so that we can get the private ip of the provisioned endpoint.

```
apiVersion: network.azure.com/v1api20220701
kind: PrivateEndpoint
metadata:
  annotations:
    serviceoperator.azure.com/latest-reconciled-generation: "1"
    serviceoperator.azure.com/operator-namespace: azureserviceoperator-system
    serviceoperator.azure.com/resource-id: /subscriptions/0000-00000-000000/resourceGroups/debug-rg/providers/Microsoft.Network/privateEndpoints/test
  creationTimestamp: "2024-03-28T07:06:26Z"
  finalizers:
  - serviceoperator.azure.com/finalizer
  generation: 1
  name: test
  namespace: default
  resourceVersion: "11865582"
  uid: fbf76c6d-3593-4f03-83c7-e8f43b66dc1b
spec:
  azureName: luketestpe
  location: central india
  owner:
    armId: /subscriptions/0000-00000-000000/resourceGroups/debug-rg
  privateLinkServiceConnections:
  - name: debug-service
    privateLinkServiceReference:
      armId: /subscriptions/0000-00000-000000/resourceGroups/debug-rg/providers/Microsoft.Network/privateLinkServices/debug-service
  subnet:
    reference:
      armId: /subscriptions/0000-00000-000000/resourceGroups/debug-rg/providers/Microsoft.Network/virtualNetworks/debug-vpc/subnets/debug-subnet
status:
  conditions:
  - lastTransitionTime: "2024-04-01T06:40:39Z"
    observedGeneration: 1
    reason: Succeeded
    status: "True"
    type: Ready
  customNetworkInterfaceName: ""
  etag: W/"a21197e5-51ed-4142-93f3-f8dc27ebdb32"
  id: /subscriptions/0000-00000-000000/resourceGroups/dev-azr-inc-aks-a-rg/providers/Microsoft.Network/privateEndpoints/test
  location: centralindia
  name: luketestpe
  networkInterfaces:
  - id: /subscriptions/0000-00000-000000/resourceGroups/debug-rg/providers/Microsoft.Network/networkInterfaces/test.nic.437f5635-6ee4-4e52-8aef-985b7c4a6017
  privateLinkServiceConnections:
  - etag: W/"a21197e5-51ed-4142-93f3-f8dc27ebdb32"
    id: /subscriptions/3f59ac6a-d708-49ec-9490-a3c786d63455/resourceGroups/dev-azr-inc-aks-a-rg/providers/Microsoft.Network/privateEndpoints/luketestpe/privateLinkServiceConnections/dev-azr-inc-mgmt-aks-demosrc
    name: debug-service
    privateLinkServiceConnectionState:
      actionsRequired: None
      description: Approved
      status: Approved
    privateLinkServiceId: /subscriptions/0000-00000-000000/resourceGroups/debug-rg/providers/Microsoft.Network/privateLinkServices/debug-service
    provisioningState: Succeeded
    type: Microsoft.Network/privateEndpoints/privateLinkServiceConnections
  provisioningState: Succeeded
  subnet:
    id: /subscriptions/0000-00000-000000/resourceGroups/debug-rg/providers/Microsoft.Network/virtualNetworks/debug-vpc/subnets/debug-subnet
  type: Microsoft.Network/privateEndpoints
```

**To Reproduce**
Create a privatelink through ASO can check its status in the yaml output.

**Expected behavior**
ipConfigurations can be retrieved so that we can know the IP info of the private endpoint.

super-harsh:
Closes #3901 

This PR adds a PrivateEndpoint extension to expose the privateIpAddress generated by Azure for PrivateEndpoint. This private IP Address is found on the NIC attached to the PrivateEndpoint.  

Only exporting this one since this is the only address generated by Azure.

gif

**If applicable**:
- [x] this PR contains tests


v2/api/network/customizations/private_endpoints_extensions.go
```diff
@@ -60,3 +67,99 @@ func (extension *PrivateEndpointExtension) PostReconcileCheck(
...
+	var _ conversion.Hub = endpoint
+
+	hasIpConfiguration := configMapSpecified(endpoint)
+	if !hasIpConfiguration {
+		log.V(Debug).Info("no configmap retrieval to perform as operatorSpec is empty")
+		return nil, nil
+	}
+
+	if endpoint.Status.NetworkInterfaces == nil || len(endpoint.Status.NetworkInterfaces) == 0 {
+		log.V(Debug).Info("no configmap retrieval to perform as there is no NetworkInterfaces attached")
```
matthchr, line 95: minor:
```suggestion
		log.V(Debug).Info("no configmap retrieval to perform as there are no NetworkInterfaces attached")
```

v2/api/network/customizations/private_endpoints_extensions.go
```diff
@@ -60,3 +67,99 @@ func (extension *PrivateEndpointExtension) PostReconcileCheck(
...
+	endpoint, ok := obj.(*network.PrivateEndpoint)
+	if !ok {
+		return nil, errors.Errorf("cannot run on unknown resource type %T, expected *network.PrivateEndpoint", obj)
+	}
+
+	// Type assert that we are the hub type. This will fail to compile if
+	// the hub type has been changed but this extension has not
+	var _ conversion.Hub = endpoint
+
+	hasIpConfiguration := configMapSpecified(endpoint)
```
matthchr, line 88: very minor: 
```suggestion
	hasIPConfiguration := configMapSpecified(endpoint)
```

v2/azure-arm.yaml
```diff
@@ -2779,6 +2779,8 @@ objectModelConfiguration:
       PrivateEndpoint:
         $export: true
         $supportedFrom: v2.0.0
+        $manualConfigs:
+          - PrimaryNicPrivateIPAddress
```
matthchr, line 2783: minor, not consistent casing of property name. I think we want:
`PrimaryNicPrivateIpAddress` -- note that this isn't the recommended Go casing but I think it's the casing that other properties use because we build their names from JSON so it's consistent with our other stuff I think

v2/api/network/customizations/private_endpoints_extensions.go
matthchr: Actually maybe just consider 
```suggestion
	hasConfigMaps := configMapSpecified(endpoint)
```

or else rename the `configMapSpecified` to `ipAddressSpecified`?

v2/api/network/customizations/private_endpoints_extensions.go
```diff
@@ -60,3 +67,99 @@ func (extension *PrivateEndpointExtension) PostReconcileCheck(
...
+		log.V(Debug).Info("no configmap retrieval to perform as operatorSpec is empty")
+		return nil, nil
+	}
+
+	if endpoint.Status.NetworkInterfaces == nil || len(endpoint.Status.NetworkInterfaces) == 0 {
+		log.V(Debug).Info("no configmap retrieval to perform as there is no NetworkInterfaces attached")
+		return nil, nil
+	}
+
+	nicId, err := arm.ParseResourceID(*endpoint.Status.NetworkInterfaces[0].Id)
```
matthchr, line 99: No guarantee that this ptr-dereference is safe, I think you need to guard it and early return to avoid a panic?

matthchr, line 99:
```suggestion
	nicID, err := arm.ParseResourceID(*endpoint.Status.NetworkInterfaces[0].Id)
```

v2/api/network/customizations/private_endpoints_extensions.go
```diff
@@ -60,3 +67,99 @@ func (extension *PrivateEndpointExtension) PostReconcileCheck(
...
+	if err != nil {
+		return nil, err
+	}
+
+	// The default primary ip configuration for PrivateEndpoint is on NetworkInterfaceController. Hence, we fetch it from there.
+	var interfacesClient *armnetwork.InterfacesClient
+	interfacesClient, err = armnetwork.NewInterfacesClient(nicId.SubscriptionID, armClient.Creds(), armClient.ClientOptions())
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to create new PrivateEndpointsClient")
```
matthchr, line 108:
```suggestion
		return nil, errors.Wrapf(err, "failed to create new NetworkInterfacesClient")
```

v2/api/network/customizations/private_endpoints_extensions.go
```diff
@@ -60,3 +67,99 @@ func (extension *PrivateEndpointExtension) PostReconcileCheck(
...
+	}
+
+	var resp armnetwork.InterfacesClientGetResponse
+	resp, err = interfacesClient.Get(ctx, nicId.ResourceGroupName, nicId.Name, nil)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed getting PrivateEndpoint")
```
matthchr, line 114:
```suggestion
		return nil, errors.Wrapf(err, "failed getting NIC")
```

v2/api/network/customizations/private_endpoints_extensions.go
```diff
@@ -60,3 +67,99 @@ func (extension *PrivateEndpointExtension) PostReconcileCheck(
...
+	}
+
+	return configmaps.SliceToClientObjectSlice(configs), nil
+}
+
+func configByName(resp armnetwork.InterfacesClientGetResponse) map[string]string {
+	result := make(map[string]string)
+
+	if resp.Properties != nil && resp.Properties.IPConfigurations != nil && len(resp.Properties.IPConfigurations) > 0 {
+		result["PrimaryNicPrivateIPAddress"] = *resp.Properties.IPConfigurations[0].Properties.PrivateIPAddress
```
matthchr, line 130: Need to guard this deref too?

v2/api/network/customizations/private_endpoints_extensions.go
```diff
@@ -60,3 +67,99 @@ func (extension *PrivateEndpointExtension) PostReconcileCheck(
...
+	configsByName := configByName(resp)
+	configs, err := configMapToWrite(endpoint, configsByName)
+	if err != nil {
+		return nil, err
+	}
+
+	return configmaps.SliceToClientObjectSlice(configs), nil
+}
+
+func configByName(resp armnetwork.InterfacesClientGetResponse) map[string]string {
```
matthchr, line 126: minor: Could you combine the `configByName` and `configMapToWrite` sections here?

I think in some secrets handlers we do this pattern because the actual data comes from a `map[string]string` to start, but here it's well-typed so it feels like just passing `armnetwork.InterfacesClientGetResponse` to `configMapToWrite` is OK?

Though actually, you shouldn't pass the `GetResponse`, if you look inside it there should be a resource type embedded and that's what you wanna pass. You don't need the HTTP junk that's with it on the `GetResponse`.

v2/api/network/customizations/private_endpoints_extensions.go
```diff
@@ -60,3 +67,99 @@ func (extension *PrivateEndpointExtension) PostReconcileCheck(
...
+
+func configMapToWrite(obj *network.PrivateEndpoint, configs map[string]string) ([]*v1.ConfigMap, error) {
+	operatorSpecConfigs := obj.Spec.OperatorSpec.ConfigMaps
+	if operatorSpecConfigs == nil {
+		return nil, errors.Errorf("unexpected nil operatorspec")
+	}
+
+	collector := configmaps.NewCollector(obj.Namespace)
+
+	primaryNicPrivateIPAddress, ok := configs["PrimaryNicPrivateIPAddress"]
```
matthchr, line 144:
```suggestion
	primaryNICPrivateIPAddress, ok := configs["PrimaryNicPrivateIPAddress"]
```

matthchr: (but as suggested above maybe do away with the map entirely here)

v2/api/network/customizations/private_endpoints_extensions.go
```diff
@@ -60,3 +67,99 @@ func (extension *PrivateEndpointExtension) PostReconcileCheck(
...
+	}
+
+	hasIpConfiguration := false
+	configMaps := endpoint.Spec.OperatorSpec.ConfigMaps
+
+	if configMaps != nil && configMaps.PrimaryNicPrivateIPAddress != nil {
+		hasIpConfiguration = true
+	}
+
+	return hasIpConfiguration
```
matthchr, line 164:
```suggestion
	return hasIPConfiguration
```

v2/api/network/customizations/private_endpoints_extensions.go
super-harsh: I followed this approach to make the code easily extendable in the future if we plan on adding more config exports. That way it's cleaner to just have all the fetching logic in one method and writing logic in another.

v2/api/network/customizations/private_endpoints_extensions.go
super-harsh: I'd rather prefer `primaryNicPrivateIPAddress` here to keep it consistent with the operatorSpec property name?

v2/api/network/customizations/private_endpoints_extensions.go
matthchr: The thing I think could be improved about the way it is now is that you fetch from the typed structure into a `map[string]string`, with a somewhat magical key of `primaryNICPrivateIPAddress`, and then you write by reading that somewhat magical key again.

My argument is that it's no clearer to read it from a `map[string]string` than it is to read it from a `network.PrivateEndpoint`, and if in the future we need to read other things from the same `network.PrivateEndpoint` there's no advantage. Only if we were reading from multiple places might there be a slight advantage, but even then it's only slight because while you avoid needing N parameters for the `configMapToWrite` function, you still need N parameters for the `configByName` function so it's not really all that much cleaner/clearer?

It feels like the `map[string]string` is adding a step that doesn't really have a ton of value now. I'd argue rather than position us for a possible future where things are more complicated we just keep it simple now and could split into 2 functions in the future if/when the need arises. YAGNI and all that, I guess.

Alternatively, if you want the structure of the split maybe you could accomplish it by still having 1 function, but it constructs the `collector` and passes it to a function that extracts the values and writes it - that way there's no intermediate map it's just straight from the struct to the collector? Then if there were other sets to collect you'd have another function for those and pass the same collector down to it too?

v2/api/network/customizations/private_endpoints_extensions.go
```diff
@@ -60,3 +67,109 @@ func (extension *PrivateEndpointExtension) PostReconcileCheck(
...
+		log.V(Debug).Info("no configmap retrieval to perform as operatorSpec is empty")
+		return nil, nil
+	}
+
+	if endpoint.Status.NetworkInterfaces == nil || len(endpoint.Status.NetworkInterfaces) == 0 {
+		log.V(Debug).Info("no configmap retrieval to perform as there are no NetworkInterfaces attached")
+		return nil, nil
+	}
+
+	if endpoint.Status.NetworkInterfaces[0].Id == nil {
```
matthchr, line 99: Is the default network interface always `[0]`? Or do you need to scan them for an `.IsDefault` property or something?

matthchr (approved):
Few more comments but mostly LGTM

v2/api/network/customizations/private_endpoints_extensions.go
matthchr: I added a scan for this at the IPConfiguration level but left the assumption that there's only 1 NIC because as far as I can tell there's no dual-NIC'ed PEs.

Issue #3805, opened 2024-02-20
Bug: `asoctl import`-ing a ManagedCluster produces an invalid resource


nojnhuh:
**Version of Azure Service Operator**

v2.5.0

**Describe the bug**
`asoctl import`-ing an AKS cluster created with `az aks create` with only the required parameters produces invalid values that trigger validation errors from the AKS API.

I had to make the following tweaks to the generated YAML to get the ManagedCluster to reconcile successfully:

`spec.networkProfile.loadBalancerProfile.backendPoolType`: nodeIPConfiguration -> NodeIPConfiguration

`spec.networkProfile.loadBalancerSku`: Standard -> standard

**To Reproduce**

```
az group create -n adopt-test -l eastus && az aks create -g adopt-test -n adopt1
asoctl import azure-resource /subscriptions/<sub>/resourceGroups/adopt-test -o managedcluster.yaml
# Delete all but the ResourceGroup and ManagedCluster resources, then
kubectl apply -f managedcluster.yaml
```

**Expected behavior**
`asoctl import` produces valid YAML that doesn't require any additional changes to reconcile successfully.

**Screenshots**

**Additional context**

theunrepentantgeek:
I suspect this is an example of a wider issue. 

In _theory_, Azure Resource Providers are supposed to be _case insensitive, case preserving_ - it doesn't matter what letter case you use, and the system preserves what you send it.

In _practice_, this is a goal, but one that isn't achieved by all RPs, including it seems, `ContainerService`.

I suspect that what we need is a _case correcting_ conversion when populating the `Spec` based on the `Status`.

matthchr:
We can work around this now in ASO (if we want), but FWIW this is actually on my list to fix in the AKS RP. Just not sure when that fix will land.

theunrepentantgeek:
To handle this purely within ASO - which I suspect we want to do to avoid upstream delays with RP changes - we'd need to change our conversion logic for **enum** values.

Currently we convert enum values using direct casts:

``` go
backendPoolType := string(source.BackendPoolType)
destination.BackendPoolType = BackendPoolType(backendPoolType)
```

This is case preserving, but assumes the source casing we have is correct.

We could instead code generate a conversion map, indexed by lowercase string value, allowing case correction as we go.

``` go
// Package global
var backendPoolTypeConversion = map[string]BackendPoolType {
        "nodeip": BackendPoolType_NodeIP,
[...]

PR #4108, opened 2024-06-18
Fix spec initializing copying wrong property

theunrepentantgeek:

In #3805, @nojnhuh noticed that `asoctl` was initializing the `type` field on an imported AgentPool with the wrong value.

Investigation leads to the observation that properties with the same name don't necessarily have the same providence - not every `Type` field can be copied into another `Type` field.

There are two fixes. 

The first is to check the origin of each property, skipping the _assign-identically-named-properties_ step if the properties came from different places

The second fix is to use the path-of-origin of the remaining properties to match things up. This is complicated by the possibility that the property may be configurable - so we would have both `Type` and `TypeFromConfig` as candidates. We disambiguate these by doing a fast and loose compatibility check on the types, which seems good enough.

gif

**If applicable**:
- [x] this PR contains tests


v2/tools/generator/internal/conversions/typed_conversion_endpoint.go
```diff
@@ -54,3 +62,20 @@ func (endpoint *TypedConversionEndpoint) IsBagItem() bool {
 	_, result := AsPropertyBagMemberType(endpoint.Type())
 	return result
 }
+
+func (endpoint *TypedConversionEndpoint) WithPath(
+	path []astmodel.PropertyName,
+) *TypedConversionEndpoint {
+	var builder strings.Builder
+	for i, name := range path {
```
matthchr, line 70: minor: You could probably have used `strings.Join` here, though it would require transforming the path to strings first.

v2/tools/generator/internal/functions/property_assignment_function_builder.go
```diff
@@ -369,10 +370,87 @@ func (*PropertyAssignmentFunctionBuilder) selectIdenticallyNamedProperties(
...
+
+// selectIdenticallyPathedProperties matches up properties with paths for conversion.
+// This serves to match up properties that were flattened from the original location,
+// even if they've ended up with different names.
+// sourceProperties is a set of endpoints that can be read from.
+// destinationProperties is a set of endpoints that can be written to.
+// assign is a function that will be called for each matching property, with the source and destination endpoints
+// for that property.
+// Returns an error if any of the assignments fail.
+func (builder *PropertyAssignmentFunctionBuilder) selectIdenticallyPathedProperties(
```
matthchr, line 402: extremely minor: 
```suggestion
func (builder *PropertyAssignmentFunctionBuilder) selectPropertiesWithIdenticalPaths(
```

(avoids `IdenticallyPathed` which read weird to me. Not sure it's really all that much better.)

matthchr (approved):
Nice!

v2/tools/generator/internal/conversions/typed_conversion_endpoint.go
theunrepentantgeek: Changed.

matthchr:
Seems some unit test failures @theunrepentantgeek

v2/tools/generator/internal/functions/property_assignment_function_builder.go
theunrepentantgeek: Done.

Issue #3399, opened 2023-10-10
Allow programatic reuse of import functionality


theunrepentantgeek:
Expose the `import azure-resource` functionality from **asoctl** as a public library so that it can be reused.

Currently the functionality is in `internal` packages and can't be reused.

* Need to refactor to remove dependencies on console tools such as the multiple-progress-bar.

ashetonh-ad:
this would help us immensely as we want to wrap the import process programmatically to allow for custom validation and formatting for our kube manifests and where they live.

matthchr:
We're still interested in this.

PR #4109, opened 2024-06-18
Update InitializeSpec to use enumeration conversions and maintain letter case

### LINKED ISSUE #3805: Bug: `asoctl import`-ing a ManagedCluster produces an invalid resource

theunrepentantgeek:

The case-correcting conversions previously introduced weren't being used for the Status->Spec conversion required by `asoctl`, resulting in the generated YAML files containing the incorrect values (which is an issue given Kubernetes validation is case-strict, not case-forgiving).

We update the property conversions to handle the new case and regenerate the api files.

Closes #3805 

gif

**If applicable**:
- [x] this PR contains tests


v2/tools/generator/internal/functions/testdata/TestGolden_PropertyAssignmentFunction_AsFunc/ConvertBetweenRequiredEnumAndRequiredEnum.golden
```diff
@@ -13,7 +16,7 @@ type Person struct {
 func (person *Person) AssignProperties_From_Person(source *vnext.Person) error {
 
 	// Release
-	release := Bucket(source.Release)
+	release := genruntime.ToEnum(string(source.Release), bucket_Values)
```
matthchr, line 19: where is `bucket_Values` actually defined?

matthchr (approved):
Possibly related to https://github.com/Azure/azure-service-operator/issues/3399 but should we make a note someplace to add some tests for the import functionality so we can catch that we missed this in the future?

(I think that's what was missing for us to notice this?)

v2/tools/generator/internal/functions/testdata/TestGolden_PropertyAssignmentFunction_AsFunc/ConvertBetweenRequiredEnumAndRequiredEnum.golden
theunrepentantgeek: Internal to the package, next to the definition of the enum and its values. 

Here's an example from `v2/api/compute/v1api20220301
/virtual_machine_scale_set_types_gen.go`:

``` go
// +kubebuilder:validation:Enum={"Delete","Detach"}
type VirtualMachineScaleSetNetworkConfigurationProperties_DeleteOption string

const (
	VirtualMachineScaleSetNetworkConfigurationProperties_DeleteOption_Delete = VirtualMachineScaleSetNetworkConfigurationProperties_DeleteOption("Delete")
	VirtualMachineScaleSetNetworkConfigurationProperties_DeleteOption_Detach = VirtualMachineScaleSetNetworkConfigurationProperties_DeleteOption("Detach")
)

// Mapping from string to VirtualMachineScaleSetNetworkConfigurationProperties_DeleteOption
var virtualMachineScaleSetNetworkConfigurationProperties_DeleteOption_Values = map[string]VirtualMachineScaleSetNetworkConfigurationProperties_DeleteOption{
	"delete": VirtualMachineScaleSetNetworkConfigurationProperties_DeleteOption_Delete,
	"detach": VirtualMachineScaleSetNetworkConfigurationProperties_DeleteOption_Detach,
}
```

PR #4127, opened 2024-06-24
Fix bug where KeyVault with ARMID owner can't be recovered

### LINKED ISSUE #4117: Bug: KeyVault createMode and armID owner doesn't work
Azure Service Operator 2.7.0

**Describe the bug**
When trying to create KeyVault and set creationMode spec to anything other than default while using armID as resource owner an error is raised.

**To Reproduce**
Just replace the id values on code bellow. 
```
apiVersion: keyvault.azure.com/v1api20230701
kind: Vault
metadata:
  annotations:
    serviceoperator.azure.com/credential-from: aso-credential-microsoft-sub
  name: test-keyvault
  namespace: azureserviceoperator-system
spec:
  location: eastus
  owner: 
    armId: /subscriptions/mysub/resourceGroups/myrg
  properties:
    createMode: createOrRecover
    tenantId: mytenant
    sku:
      family: A
      name: standard
    accessPolicies:
      - objectId: "objid"
        tenantId: "tenantid"
        permissions:
          certificates:
            - "get"
          keys:
            - "get"
          secrets:
            - "get"
          storage:
            - "get"

```
**Expected behavior**
KeyVault created with expected createMode set and selected resource group as its owner.

**Additional context**
Error message is the following:
`failed to add initial resource state: failed to modify ARM payload: unable to find owner of KeyVault argocd-keyvault: owner of KeyVault test-keyvault is not within the cluster`

matthchr:
Fixes #4117.

Also fix an issue where we gave a confusing and non-fatal error when attempting to recover a KeyVault in a different ResourceGroup than it was originally in. Changed the error to be clearer and also immediately fatal so that no retries happen.

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/api/keyvault/customizations/vault_extensions.go
+++ b/v2/api/keyvault/customizations/vault_extensions.go
@@ -8,9 +8,11 @@ package customizations
 import (
 	"context"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/keyvault/armkeyvault"
 	"github.com/go-logr/logr"
@@ -24,7 +26,9 @@ import (
 	"github.com/Azure/azure-service-operator/v2/internal/reflecthelpers"
 	"github.com/Azure/azure-service-operator/v2/internal/resolver"
 	"github.com/Azure/azure-service-operator/v2/internal/util/kubeclient"
+	"github.com/Azure/azure-service-operator/v2/internal/util/to"
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/conditions"
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/extensions"
 )
 
@@ -64,15 +68,9 @@ func (ex *VaultExtension) ModifyARMResource(
 		return armObj, nil
 	}
 
-	// Get the owner of the KeyVault, we need this resource group to determine the subscription
-	owner, ownerErr := ex.getOwner(ctx, kv, resolver, log)
-	if ownerErr != nil {
-		return nil, errors.Wrapf(ownerErr, "unable to find owner of KeyVault %s", kv.Name)
-	}
-
 	// Parse the ID of the owner
 	// (Can't use the KeyVault as we do this before the KV exists)
-	id, err := genruntime.GetAndParseResourceID(owner)
+	id, err := ex.getOwner(ctx, kv, resolver)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get and parse resource ID from KeyVault owner")
 	}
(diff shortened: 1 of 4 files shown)


v2/api/keyvault/customizations/vault_extensions.go
```diff
@@ -19,12 +21,13 @@ import (
 
 	keyvault "github.com/Azure/azure-service-operator/v2/api/keyvault/v1api20230701/storage"
 	resources "github.com/Azure/azure-service-operator/v2/api/resources/v1api20200601/storage"
-
```
theunrepentantgeek, line 22: nit: I like separating named imports from regular imports.

v2/api/keyvault/customizations/vault_extensions.go
```diff
@@ -184,25 +186,36 @@ func (ex *VaultExtension) handlePurgeThenCreate(
 	return nil
 }
 
+type deletionDetails struct {
+	Exists       bool
+	DeletedVault *armkeyvault.DeletedVault
+}
+
+func deletionDetailsExist(deletedVault armkeyvault.DeletedVault) deletionDetails {
```
theunrepentantgeek, line 194: Maybe use a name that's more declarative at the point of use?

```suggestion
func deletedKeyVaultFound(vault armkeyvault.DeletedVault) deletionDetails {
```

v2/api/keyvault/customizations/vault_extensions.go
```diff
@@ -184,25 +186,36 @@ func (ex *VaultExtension) handlePurgeThenCreate(
...
+	return deletionDetails{
+		Exists:       true,
+		DeletedVault: &deletedVault,
+	}
+}
+
+func deletionDetailsDontExist() deletionDetails {
```
theunrepentantgeek, line 201:
```suggestion
func deletedKeyVaultNotFound() deletionDetails {
```

v2/api/keyvault/customizations/vault_extensions.go
```diff
@@ -184,25 +186,36 @@ func (ex *VaultExtension) handlePurgeThenCreate(
 	return nil
 }
 
+type deletionDetails struct {
```
theunrepentantgeek, line 189: I like the use of a result struct.

theunrepentantgeek (approved):

v2/api/keyvault/customizations/vault_extensions.go
```diff
@@ -113,23 +110,37 @@ func (ex *VaultExtension) handleCreateOrRecover(
 	ctx context.Context,
 	kv *keyvault.Vault,
 	vc *armkeyvault.VaultsClient,
-	resolver *resolver.Resolver,
+	ownerID *arm.ResourceID,
 	log logr.Logger,
 ) (string, error) {
-	exists, err := ex.checkForExistenceOfDeletedKeyVault(ctx, kv, resolver, vc, log)
+	delDetails, err := ex.checkForExistenceOfDeletedKeyVault(ctx, kv, vc, log)
```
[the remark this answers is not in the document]
matthchr: Yeah I didn't really like `delDetails` either, I just wanted a name that didn't shadow the type... I do like this a bit better. Went with `deletedKeyVault` though just to expand the `KV` as I think it reads slightly better.

v2/api/keyvault/customizations/vault_extensions.go
matthchr: Much better.

v2/api/keyvault/customizations/vault_extensions.go
matthchr: Yeah I wrote it without one (partially) initially and the nil-checking was pretty awkward so went with this instead.

v2/api/keyvault/customizations/vault_extensions.go
matthchr: I brought these back but AFAIK that's not a common pattern so tools (gofump/goimports/goland import ordering) don't have an option to do this. They preserve them if there but AFAIK the standard ordering is either stdlib / everything else (2 groups) or stdlib / external / internal (3 groups) and this results in 4.

PR #4128, opened 2024-06-25
Fix pre-release-test task target

super-harsh:

This PR fixes the TaskFile's `controller:test-upgrade-apply-prerelease-chart` target to replace the old target `controller:gen-helm-manifest` to `controller:package-helm-manifest`

Also, fixes `create-kind-wi-storage.sh` to set --allow-blob-public-access==false for StorageAccount creation.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -374,7 +374,7 @@ tasks:
     dir: "{{.CONTROLLER_ROOT}}"
     cmds:
       - task: controller:docker-push-local
-      - task: controller:gen-helm-manifest
+      - task: controller:package-helm-manifest
       - task: controller:install-helm-wi
         vars:
           AZURE_MI_CLIENT_ID:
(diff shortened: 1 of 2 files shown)


theunrepentantgeek (approved):

matthchr:
> Also, fixes create-kind-wi-storage.sh to set --allow-blob-public-access==false for StorageAccount creation.

Was there an error someplace? IIRC this is the default, it shouldn't need to be specified explicitly?

super-harsh:
@matthchr - `allow-blob-public-access` is blocked by the policy and is set to True by default.

super-harsh:
actually, default is True for versions prior to `2023-01-01`. What version does az pick up by default? any idea?  :-?

matthchr:
It should pick the latest API (depending on version of CLI). The docker container at least is using a new-ish version of the CLI so should have this as the default, but being explicit doesn't hurt.

matthchr (approved):

PR #4136, opened 2024-06-25
Group updates from all 3 packages into a single dependabot PR

matthchr:
This should avoid the issues we've had in the past with dependencies between /v2 and /v2/cmd/asoctl requiring merges in a certain order.

**If applicable**:
- [ ] this PR contains tests


theunrepentantgeek (approved):
If this works as advertised, it'll be most welcome!

PR #4142, opened 2024-06-27
Fix issue with aks-create Taskfile target failing due to no specified DOCKER_REGISTRY

matthchr:
Also fix unrelated warning from Dockerfiles for using FROM and as with different cases.

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -456,7 +456,9 @@ tasks:
     desc: Builds, tags and pushes the multi architecture controller container image to repository
     deps:
       - controller:bundle-crds
-      - docker-login
+      - task: docker-login
+        vars:
+          DOCKER_REGISTRY: "{{.DOCKER_REGISTRY}}"
     dir: "{{.CONTROLLER_ROOT}}"
     run: always
     cmds:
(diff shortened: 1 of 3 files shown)


Taskfile.yml
```diff
@@ -1226,6 +1229,14 @@ tasks:
...
       - 'if [ -z "{{.DOCKER_REGISTRY}}" ]; then echo "Error: DOCKER_REGISTRY must be set"; exit 1; fi'
       - az acr login --name {{.DOCKER_REGISTRY}}
+    vars:
+      DOCKER_REGISTRY:
+        sh: |
+          if [ ! -z "${DOCKER_REGISTRY}" ]; then
+            echo "${DOCKER_REGISTRY}"
+          else
+            echo "{{.DOCKER_REGISTRY}}"
+          fi
```
theunrepentantgeek, line 1239: I don't think you need to do that. Task [will look in the environment](https://taskfile.dev/usage/#variables) for `.DOCKER_REGISTRY` if it's not already in scope.

theunrepentantgeek (approved):
Overall looks good. I think a couple of things can be simplified though.

Taskfile.yml
```diff
@@ -891,11 +893,12 @@ tasks:
           DIR: "{{.AKS_WORKLOAD_IDENTITY_PATH}}"
           RESOURCE_GROUP: "{{.HOSTNAME}}-aso-rg"
       - task: controller:install-cert-manager
-      - "az acr login --name {{.ACR_NAME}}"
+      # - "az acr login --name {{.ACR_NAME}}"
```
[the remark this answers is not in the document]
matthchr: Deleted actually, fixed.

This is now done as part of the docker-push-multiarch

Taskfile.yml
matthchr: Tested without this and confirmed it works.

PR #4144, opened 2024-07-01
Catalog all known resources, regardless of whether ASO exports them

theunrepentantgeek:

To produce a report listing supported resources that need upgrading, we first need to have a catalog of all known resources.

In this PR we add that catalog to our pipeline state, unblocking the future report.

As this is the _third_ piece of specialized (non-global) data that we need to store in the state, I've switched up the storage to a more generic approach where we don't need to make extensive customizations to `State` as new data needs are identified.

gif

### DIFF
--- a/v2/tools/generator/internal/codegen/pipeline/add_kubernetes_exporter.go
+++ b/v2/tools/generator/internal/codegen/pipeline/add_kubernetes_exporter.go
@@ -24,9 +24,9 @@ func AddKubernetesExporter(idFactory astmodel.IdentifierFactory) *Stage {
 			defs := state.Definitions()
 			updatedDefs := make(astmodel.TypeDefinitionSet)
 
-			mappings, ok := GetStateInfo[*ExportedTypeNameProperties](state, ExportedConfigMaps)
-			if !ok {
-				return nil, errors.New("exported config maps not found")
+			mappings, err := GetStateInfo[*ExportedTypeNameProperties](state, ExportedConfigMaps)
+			if err != nil {
+				return nil, errors.Wrapf(err, "couldn't find exported config maps")
 			}
 
 			for _, def := range astmodel.FindResourceDefinitions(defs) {
--- a/v2/tools/generator/internal/codegen/pipeline/add_kubernetes_exporter.go
+++ b/v2/tools/generator/internal/codegen/pipeline/add_kubernetes_exporter.go
@@ -24,7 +24,10 @@ func AddKubernetesExporter(idFactory astmodel.IdentifierFactory) *Stage {
 			defs := state.Definitions()
 			updatedDefs := make(astmodel.TypeDefinitionSet)
 
-			mappings := state.GeneratedConfigMaps()
+			mappings, ok := GetStateInfo[*ExportedTypeNameProperties](state, ExportedConfigMaps)
+			if !ok {
+				return nil, errors.New("exported config maps not found")
+			}
 
 			for _, def := range astmodel.FindResourceDefinitions(defs) {
 				resourceType, ok := astmodel.AsResourceType(def.Type())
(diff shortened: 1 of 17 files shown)


v2/tools/generator/internal/codegen/pipeline/state.go
```diff
@@ -131,10 +109,43 @@ func (s *State) CheckFinalState() error {
...
+		stagesSeen:     s.stagesSeen,
+		stagesExpected: s.stagesExpected,
+		stateInfo:      maps.Clone(s.stateInfo),
+	}
+}
+
+// StateWithInfo returns a new state with the given information included.
+// Any existing information with the same key is replaced.
+// Has to be written as a standalone function because methods can't introduce new generic variation.
+func StateWithInfo[I any](
```
matthchr, line 122: minor, style: prefer `T` for generic type variable, rather than I?

v2/tools/generator/internal/codegen/pipeline/state.go
```diff
@@ -131,10 +109,43 @@ func (s *State) CheckFinalState() error {
...
 	}
+
+	result.stateInfo[key] = info
+	return result
+}
+
+// GetStateInfo returns the information stored in the state under the given key.
+// If no information is stored under the key, or if it doesn't have the expected type, the second
+// return value is false.
+// Has to be written as a standalone function because methods can't introduce new generic variation.
```
matthchr, line 139:
```suggestion
// Has to be written as a standalone function because methods can't introduce new generic variables.
```
?

I wish this weren't the case, annoying gap.

v2/tools/generator/internal/codegen/pipeline/state.go
```diff
@@ -30,6 +29,14 @@ type State struct {
  *  that identifier.
  */
 
+type StateInfo string
```
matthchr, line 32: We could call this a `StateItem` or `StateElement` if we wanted?

Not sure if that's necessarily a lot clearer than `StateInfo` but might be _slightly_ less generic?

v2/tools/generator/internal/codegen/pipeline/add_kubernetes_exporter.go
```diff
@@ -24,7 +24,10 @@ func AddKubernetesExporter(idFactory astmodel.IdentifierFactory) *Stage {
 			defs := state.Definitions()
 			updatedDefs := make(astmodel.TypeDefinitionSet)
 
-			mappings := state.GeneratedConfigMaps()
+			mappings, ok := GetStateInfo[*ExportedTypeNameProperties](state, ExportedConfigMaps)
```
matthchr, line 27: minor: If the pattern is always:
```
result, ok := GetStateInfo(state, key)
if !ok {
    return ???, err
}
```

would it be better if we just made `GetStateInfo` return an error?

I'm not sold it's better but figured I'd ask. OK with either answer.

matthchr (approved):
LGTM, few minor comments on naming/style.

v2/tools/generator/internal/codegen/pipeline/state.go
theunrepentantgeek: I used `I` for info, mirroring the way the Golang libraries use single capital letters, e.g. the `maps` package uses `K` for the type of key, `V` for the type of value.

v2/tools/generator/internal/codegen/pipeline/add_kubernetes_exporter.go
theunrepentantgeek: I modeled the API on the way maps work, returning a **bool** to say whether it worked. I like the idea of standardizing the error message though, so I'll make a change.

v2/tools/generator/internal/codegen/pipeline/state.go
theunrepentantgeek: Renamed to `StateDataKey`, along with adjusting the names of the associated functions.

Issue #2555, opened 2022-10-19
Feature: Support dynamic secret/configmap exports


matthchr:
Sometimes the consuming application wants a value like `myredis.redis.cache.windows.net:6380`, but the problem is that we export the pieces of that string separately today, like so:

```
  operatorSpec:
    secrets:
      ...
      hostName:
        name: redis-secret
        key: hostName
      sslPort:
        name: redis-secret
        key: port
```

There's no way for the application to get the value in the format it needs. Sometimes this can be worked around by the application making a code change, but that only works if the user controls the application. Some applications (such as DAPR which the above example is from) we cannot expect the user to change. We have to support getting them the values they want in a format that makes sense.

We should consider supporting something like `formattedSecrets` (or `formattedValues`?), which would be used like so:
```
  operatorSpec:
    secrets:
      ...
      hostName:
        name: redis-secret
        key: hostName
      sslPort:
        name: redis-secret
        key: port
      formattedValues:
      - name: redis-secret
        key: hostPort
        value: "${hostName}:${sslPort}"
```

Terraform supports this more generically with [format](https://www.terraform.io/language/functions/format), which is what [ARM templates](https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-string#format) does as well.

matthchr:
Is there an existing operator we could use to do this instead of doing it ourselves.

matthchr:
We want to do this still, but need a design

jonasnorlund:
In general we need to have a way to get values from generated resources in a dynamic approach. Let me give one example, if we create a private endpoint we would like to get new ip address of the newly created resource so we can create a network policy that opens up communication to that ip address in the next step. 

@matthchr, this is just an example, is there any roadmap item to support fetching properties from created resources in a dynamic way, not static as it is done today? Similar to Terraform or Bicep output. This becomes important when there are resources that are dependent on each others values.

matthchr:
This is going to be our solution to exporting things like connections strings, so closed #3929 in favor in this

PR #4148, opened 2024-07-02
Add Design for Dynamic export of ConfigMap/Secret values

matthchr:
Related to #2555.

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/docs/hugo/content/design/ADR-2024-07-Dynamic-Export-To-Secret-Or-ConfigMap.md
+++ b/docs/hugo/content/design/ADR-2024-07-Dynamic-Export-To-Secret-Or-ConfigMap.md
@@ -1,0 +1,471 @@
+---
+title: '2024-07: Dynamic Export to Secret or ConfigMap'
+---
+
+## Context
+
+It is a common ask for users to request a new value be exportable to `ConfigMap`, or for the values
+that are currently exported to a `ConfigMap` or `Secret` to be customizable in some way.
+
+Outstanding issues related to this topic:
+1. [#2920: Feature: Support export of storage account connection strings](https://github.com/Azure/azure-service-operator/issues/3929).
+2. [#2555: Feature: Support dynamic secret/configmap exports](https://github.com/Azure/azure-service-operator/issues/2555)
+3. [#3711: PublicIPPrefix: Allow exporting IP range as ConfigMap](https://github.com/Azure/azure-service-operator/issues/3711)
+4. [#3671: DatabaseAccount CRD: Add support to store the account Id in a config map](https://github.com/Azure/azure-service-operator/issues/3671)
+
+(there may be more)
+
+Digging into [#2920](https://github.com/Azure/azure-service-operator/issues/3929):
+
+We don't currently support Azure storage connection strings, which is something that ASOv1 has
+supported for quite some time.
+
+The format of a connection string is:
+`DefaultEndpointsProtocol=https;AccountName=%s;AccountKey=%s;EndpointSuffix=%s`
+
+Users would like the ability to export a `Secret` formatted like this from a `StorageAccount`. While we could
+manually support this shape in the `OperatorSpec` of the `StorageAccount`, a better solution would be one that
+scaled to other connection string formats and worked for other resources without manual effort required for each one.
+
+## Requirements
+
+* Support composite exports that contain a combination of static text and variable data.
+* Support variable data from from spec or status.
+* Support variable data from operatorSpec.secrets (only to secrets).
+  * If no secrets are used, no secret APIs should be called. 
+* Mechanism of export is safe in a multitenant environment.
+
+## Options
+
+There are 3 aspects to delivering this feature.
+
+1. How to express the desired output string format. This is the [template language](#template-language).
+2. Where to signal the desire for one or more of these dynamic outputs, and where they should go. This is 
+   the [location](#location).
+3. How to include secret data.
+
+### Template Language
+
+#### Option 1: Go Text Templates (text/template)
+
+Templates based on the Go stdlib [text/template](https://pkg.go.dev/text/template) package.
+
+The format of these templates is: 
+`DefaultEndpointsProtocol=https;AccountName={{.Spec.AzureName}};AccountKey={{.Spec.OperatorSpec.Secrets.Key1}};EndpointSuffix=core.windows.net`
+
+The Go code involved in parsing this template would look something like this:
+```go
+func renderTemplate(resource any, template string) string {
+	tmpl, err := template.New("test").Parse(template)
+	if err != nil { 
+		panic(err) 
+	}
+	builder := strings.Builder{}
+	err = tmpl.Execute(builder, resource)
+	if err != nil { 
+		panic(err) 
+	}
+
+	return builder.String()
+}
+```
+
+**Pros**
+
+* Likely familiar to users of Helm, which uses the same engine, as do many other projects like go-task.
+* Simple templates are simple, string interpolation is supported.
+
+**Cons**
+
+* `text/template` is powerful and not very configurable. It supports loops (`range`), if-statements, and some builtin
+  functions (`printf`, etc) out of the box without any way to restrict these capabilities. This puts the operator at 
+  risk in a multi-tenant environment because there's nothing stopping namespace A from submitting a template 
+  `range int64max`.
+
+#### Option 2: Google Common Expression Language (CEL)
+
+You can read more about it in the 
+[language specification](https://github.com/google/cel-spec/blob/master/doc/langdef.md#overview) and at
+[cel-go](https://github.com/google/cel-go).
+
+The format of these templates is: 
+`'DefaultEndpointsProtocol=https;AccountName=%s;AccountKey=%s;EndpointSuffix=core.windows.net'.format([resource.spec.azureName, resource.spec.operatorSpec.secrets.key1])`
+
+You can read more about other CEL string extensions and the format function specifically in 
+[cel-go string extensions](https://github.com/google/cel-go/blob/master/ext/strings.go#L81).
+
+If we really wanted string interpolation we could possibly write a helper for it, see 
+[here](https://github.com/google/cel-go/blob/a555792723a2ae8d35cec381e8b222b5b20cb1d9/ext/strings.go#L468).
+
+The Go code involved in parsing this template would look something like this:
+```go
+// renderTemplate produces a variable from the given template
+func renderTemplate(resource any, template string) string {
+	// The environment can be customized by providing the options cel.EnvOption to the call.
+	// Those options are able to disable macros, declare custom variables and functions, etc.
+	env, err := cel.NewEnv(
+		ext.Strings(),
+		cel.Declarations(
+			decls.NewVar("resource", decls.Dyn)),
+	)
+	if err != nil {
+		panic(err) // Simple panic-based error handling for example purposes.
+	}
+
+	ast, iss := env.Compile(template)
+	if iss.Err() != nil {
+		panic(iss.Err())
+	}
+	// Check the output type is a string.
+	if !reflect.DeepEqual(ast.OutputType(), cel.StringType) {
+		panic(fmt.Sprintf(
+			"Got %v, wanted %v output type",
+			ast.OutputType(), cel.StringType,
+		)
+	}
+
+	program, err := env.Program(ast)
+	if err != nil {
+		panic(err)
+	}
+
+
+    input := map[string]any{
+        "resource": resource,
+    }
+	
+    out, _, err := program.Eval(input)
+    if err != nil {
+        panic(err)
+    }
+    outStr := out.Value().(string)
+    return outStr
+}
+```
+
+**Pros**
+
+* We have a significant amount of control over what functions are accessible in the CEL expressions, 
+  including defining our own if we want.
+* Security: Users have little/no flow-control constructs, and CEL is hardened for use in multitenant environments such 
+  as Kubernetes APIServer.
+* Power: We can parse the CEL expression and examine the AST ourselves to extract useful information such as 
+  which secrets they're attempting to use, which enables us to deliver on the "no secret APIs called if no secrets used"
+  goal.
+
+**Cons**
+
+* May not be as familiar to users.
+* No string interpolation, which makes the templates not quite as clean as they otherwise could be.
+  * Mitigation: We could write our own, although then that function will _definitely_ not be familiar to users. 
+
+#### Option 3: Write our own simple text-replacement engine
+
+Supporting a format similar to `text/template`, but without any of the other looping/flow-control features.
+`DefaultEndpointsProtocol=https;AccountName={{.Spec.AzureName}};AccountKey={{.Spec.OperatorSpec.Secrets.Key1}};EndpointSuffix=core.windows.net`
+
+**Pros**
+
+* Theoretically simple
+
+**Cons**
+
+* More work than using something off the shelf. Including things like getting errors right, etc.
+* Not as easily extensible as something like CEL if we wanted to add more capabilities.
+* Trusting our security/hardening is scarier than trusting a "real" product.
+
+### Location
+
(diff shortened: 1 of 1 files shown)


docs/hugo/content/design/ADR-2024-07-Dynamic-Export-To-Secret-Or-ConfigMap.md
```diff
@@ -0,0 +1,428 @@
...
+This implementation is not suitable to supporting the new approach because it returns `[]client.Object`, 
+but to supply the CEL-expressions we need a return type of an object or a map whose
+properties/keys line up with the secret names. Alternatively we could delegate evaluation of the CEL expression
+to the extension itself but this would require relatively complex code autogenerated for each type, since every
+type will support the new `operatorSpec.formattedConfigMaps` field.
+
+Since `ExportKubernetesResources` is already used for a variety of purposes, rather than change its signature I propose
+we add a new sister interface to it called `SecretGetter` for resources that have Secret APIs to call.
+This interface must be implemented manually alongside the manual implementation of the
+`KubernetesExporter` resource.
```
theunrepentantgeek, line 339: If it must be implemented alongside - that is, if it's mandatory - why not extend the existing `KubernetesExporter` interface with a second method, to force this.

docs/hugo/content/design/ADR-2024-07-Dynamic-Export-To-Secret-Or-ConfigMap.md
```diff
@@ -0,0 +1,428 @@
...
+implementation can use these hints to determine which (if any) APIs to call. Empty `secretHints` means no secrets
+were requested and no secret APIs should be invoked.
+
+Q: How will `secretHints` be determined, since they're buried in an expression the users authored?
+
+A: The easy (but probably bad) way is to do some basic string-matching for the keywords we're aware of, for example because
+the user will be forced to do something like `secret.foo` in their expression to use the `foo` secret, we can parse
+that out of the expression string manually and build a catalog of what secrets they've requested. The harder but probably
+better way is to examine the CEL AST and find all instances of `SelectExpr` and extract out those that correspond to
+`secret.foo`.
```
theunrepentantgeek, line 406: :+1: for using the CEL AST

docs/hugo/content/design/ADR-2024-07-Dynamic-Export-To-Secret-Or-ConfigMap.md
```diff
@@ -0,0 +1,428 @@
...
+* More complicated, especially setting up watches on N resource types so that changes are correctly monitored and
+  propagated quickly. That additional complexity doesn't necessarily bring all that many wins.
+* Access to secret data is more difficult to implement. We would need to have a static mapping of resource types and
+  the `Exporter` would need to call each referenced resources `ExportKubernetesResources`, and we'd likely need to 
+  refactor `ExportKubernetesResources` some to support the new pattern (though we may need to do this anyway).
+
+### Access to secret data
+
+Getting access to data to export to ConfigMaps is pretty easy, we'll just pass the whole `resource` to
+the expression. As shown elsewhere fields will be accessible in expressions like `resource.status.id`.
```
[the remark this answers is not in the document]
matthchr: There's also metadata though so I think just passing the whole top level object might be cleaner?

docs/hugo/content/design/ADR-2024-07-Dynamic-Export-To-Secret-Or-ConfigMap.md
matthchr: Not every KubernetesExporter produces secrets. If all it produces is configmaps we don't need to have the secret specific one

docs/hugo/content/design/ADR-2024-07-Dynamic-Export-To-Secret-Or-ConfigMap.md
```diff
@@ -0,0 +1,428 @@
...
+        name: redis-secret
+        key: hostName
+      sslPort:
+        name: redis-secret
+        key: port
+      formattedValues:
+        - name: redis-secret # Name of the destination secret
+          key: hostPort # Name of the key in the secret
+          value: "'%s:%s'.format([secret.hostName, secret.sslPort])" # Value (format) of the secret
+```
```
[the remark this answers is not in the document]
matthchr: See later proposal. Closing this

docs/hugo/content/design/ADR-2024-07-Dynamic-Export-To-Secret-Or-ConfigMap.md
```diff
@@ -0,0 +1,428 @@
...
+
+**ConfigMaps**
+```yaml
+  operatorSpec:
+    configMaps:
+      formattedValues:
+        - name: account-data # Name of the destination configMap
+          key: id # Name of the key in the configMap
+          value: "resource.status.id" # Value (format) of the configMap entry
+```
```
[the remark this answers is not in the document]
matthchr: See later proposal =)

docs/hugo/content/design/ADR-2024-07-Dynamic-Export-To-Secret-Or-ConfigMap.md
```diff
@@ -0,0 +1,428 @@
...
+  operatorSpec:
+    secrets:
+      hostName:
+        name: redis-secret
+        key: hostName
+      sslPort:
+        name: redis-secret
+        key: port
+    formattedSecrets:
+      values:
```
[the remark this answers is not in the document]
matthchr: My reasoning for `values` was that it gives us more extension flexibility in the future if this is an object rather than a map, but maybe that's not a big deal because we have operatorSpec already.

docs/hugo/content/design/ADR-2024-07-Dynamic-Export-To-Secret-Or-ConfigMap.md
matthchr: Yes I looked into how to do it using the CEL AST and it's not very difficult.

matthchr: I'll update this item to just say CEL AST is the answer. It's one of the advantages of using CEL as well.

docs/hugo/content/design/ADR-2024-07-Dynamic-Export-To-Secret-Or-ConfigMap.md
```diff
@@ -0,0 +1,428 @@
...
+        - name: redis-secret # Name of the destination secret
+          key: hostPort # Name of the key in the secret
+          value: "'%s:%s'.format([secret.hostName, secret.sslPort])" # Value (format) of the secret
+```
+
+**ConfigMaps**
```
super-harsh, line 240: Would we be able to export collections? Most preferably `Maps` since we don't support exporting slices yet.

docs/hugo/content/design/ADR-2024-07-Dynamic-Export-To-Secret-Or-ConfigMap.md
```diff
@@ -0,0 +1,428 @@
...
+### FAQ
+
+Q: How does this new way avoid calling the secrets APIs if the user didn't ask for any secrets to be exported? 
+Keep in mind that some resources such as ManagedCluster (at v2/api/containerservice/customizations/managed_cluster_extensions.go)
+require the ability to differentiate _which_ secrets are being requested as well because there are multiple
+secrets APIs.
+
+A: For `FormattedSecrets`, the function signature to discover secrets will have `secretHints` passed to it, 
+which will contain details about the specific secrets requested in the formatting expression. The resource-specific 
+implementation can use these hints to determine which (if any) APIs to call. Empty `secretHints` means no secrets
```
super-harsh, line 397: Do we specify these APIs for each resource somewhere? How'd the operator know which API to call for each resource?

docs/hugo/content/design/ADR-2024-07-Dynamic-Export-To-Secret-Or-ConfigMap.md
matthchr: Yes, though the user might need to write a CEL expression to explain what exactly they want to gather from the map

docs/hugo/content/design/ADR-2024-07-Dynamic-Export-To-Secret-Or-ConfigMap.md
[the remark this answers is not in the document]
matthchr: I've changed this to remove the intermediate `values`

docs/hugo/content/design/ADR-2024-07-Dynamic-Export-To-Secret-Or-ConfigMap.md
```diff
@@ -0,0 +1,428 @@
...
+        - name: redis-secret # Name of the destination secret
+          key: hostPort # Name of the key in the secret
+          value: "'%s:%s'.format([secret.hostName, secret.sslPort])" # Value (format) of the secret
+```
+
+**ConfigMaps**
+```yaml
+  operatorSpec:
+    formattedConfigMaps:
+      values:
```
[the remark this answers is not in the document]
matthchr: Removed.

docs/hugo/content/design/ADR-2024-07-Dynamic-Export-To-Secret-Or-ConfigMap.md
```diff
@@ -0,0 +1,428 @@
...
+        - name: account-data # Name of the destination configMap
+          key: id # Name of the key in the configMap
+          value: "resource.status.id" # Value (format) of the configMap entry
+```
+
+**Pros**
+
+* Fits into our existing model relatively well.
```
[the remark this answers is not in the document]
matthchr: Added.

docs/hugo/content/design/ADR-2024-07-Dynamic-Export-To-Secret-Or-ConfigMap.md
matthchr: I've clarified this wording some and pointed out we could add the method to the existing interface, but I think a separate interface will ultimately be cleaner. Not married to making a decision here though we can also discuss during implementation.

docs/hugo/content/design/ADR-2024-07-Dynamic-Export-To-Secret-Or-ConfigMap.md
```diff
@@ -0,0 +1,428 @@
...
+
+* More work than using something off the shelf. Including things like getting errors right, etc.
+* Not as easily extensible as something like CEL if we wanted to add more capabilities.
+* Trusting our security/hardening is scarier than trusting a "real" product.
+
+### Location
+
+#### Option 1: Part of operatorSpec.secrets and operatorSpec.configMap
+
+A new `formattedValues` option in `operatorSpec.secrets` and `operatorSpec.configMaps`
```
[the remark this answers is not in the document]
matthchr: Swapped this to `dynamic` everywhere

docs/hugo/content/design/ADR-2024-07-Dynamic-Export-To-Secret-Or-ConfigMap.md
matthchr: The operator will call the existing handcrafted functions (they need some modifications) for this to work.

docs/hugo/content/design/ADR-2024-07-Dynamic-Export-To-Secret-Or-ConfigMap.md
matthchr: I will update the documentation here to be explicit about how this might look in an actual CEL expression

matthchr: Updated the FAQ documentation. It now says:

> Q: Does this work with arbitrary arrays and maps?
>
> A: Yes! The syntax for maps is: `request.myMap['hello']`, and for arrays `request.myArray[0]`. CEL has a concept of
macros and there are some useful ones such as `filter`, `exists`, and `all` 
[built in](https://github.com/google/cel-spec/blob/master/doc/langdef.md#macros), so you can craft more complex
expressions like `request.slice.exists_one(i, i.name=='foo') ? request.slice.filter(i, i.name == 'foo')[0].value : 0`
which checks if the slice contains exactly one element whose name is `foo` and if so, gets that elements value, 
otherwise returns 0. A similar construct can be done for maps using the same `exists_one` and `filter` macros.

matthchr: Ah, but what about if you want the _whole_ collection...

matthchr: Ok we can support whole collections too. Here's how:
> Q: Can this be used to export entire collections of secret or configmap values?
> 
> A: Yes. We will support two flavors of entry into the `dynamicSecrets` and `dynamicConfigMaps` collections. One
flavor is the one we've seen above, where `value` is a CEL expression that returns a `string` and the `name` and `key`
fields specify which configMap or secret and what key within that to export.
>
> The other kind of entry is:
> ```
>  operatorSpec:
>    dynamicConfigMaps:
>      - name: account-data # Name of the destination configMap
>        valueMap: "{"key1": "foo", "key2": "bar"}" # CEL expression that returns a map[string]string
> ```
>
>The `valueMap` fields result can be directly saved as keys + values in the resulting configmap or secret, so the author
of the CEL expression has direct control over what values go into their secret/configMap.
>


Great question @super-harsh.

docs/hugo/content/design/ADR-2024-07-Dynamic-Export-To-Secret-Or-ConfigMap.md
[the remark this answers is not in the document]
theunrepentantgeek: I think the tension here is between _making things possible_ and _making common things easy_. 

Passing just the top level object makes things _possible_ but it forces every usage to incur a little bit of overhead. Passing multiple objects is more complex to implement, but has the benefit of making most usage simpler.

theunrepentantgeek (approved):

super-harsh (approved):

docs/hugo/content/design/ADR-2024-07-Dynamic-Export-To-Secret-Or-ConfigMap.md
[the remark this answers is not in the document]
matthchr: Still not sure I agree. `resource.spec.whatever` isn't much more complicated than `spec.whatever`.


The advantages to us using top level names we choose are:
1. It disambiguates between the resource and its secrets, since for secret usage we'll support `secrets.myKey`. If we control the top-level name of all the inputs (the main 2 right now being `resource` and `secrets`), then we can ensure there are no collisions because we own the top of the path for each input. If we instead snip that part out (for just resource? or for secrets too?) then we run the risk (though it's small) that a top level field such as `secrets` collides.
2. It puts us in a better position to grow in the future by adding other top-level types going forward, such as `operator` (holding operator configuration?) or `cluster` (holding cluster details?). I don't know if we're ever going to want to add these but if we did, disambiguating between them at the top level helps with clarity.
3. It means we're consistent between `resource.x` and `secrets.y` in terms of the pattern, as opposed to doing `x` and `secrets.y`. I think including `secrets` as a prefix for secret values which are not on the resource is very useful from a security/safety point of view because we're making it really obvious the value you're about to output is a secret. Obviously we only support this in the context of saving to secrets so it's not like there's risk of exposing a secret value as plaintext but still the clarity seems a win for something sensitive like secrets.

I'll update the document to discuss this and leave it as an option question at the end of the document.

PR #4149, opened 2024-07-02
Add an ADR template

matthchr:
Not sure if this is super useful or not, but I thought it might be so here it is.

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/docs/hugo/content/design/ADR-(YEAR)-(MONTH)-(TITLE).md
+++ b/docs/hugo/content/design/ADR-(YEAR)-(MONTH)-(TITLE).md
@@ -1,0 +1,59 @@
+---
+title: 'YEAR-MONTH: ADR TEMPLATE'
+toc_hide: true
+---
+
+## Context
+
+## Requirements
+
+<!-- 
+## Options (optional) 
+
+### Option 1: Foo
+
+**Pros:**
+
+- Pro 1
+
+**Cons:**
+
+- Con 1
+
+### Option 2: Bar
+
+**Pros:**
+
+- Pro 1
+
+**Cons:**
+
+- Con 1
+
+-->
+
+## Decision
+
+<!--
+### FAQ 
+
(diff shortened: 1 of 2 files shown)


docs/hugo/content/design/ADR-Template.md
```diff
@@ -0,0 +1,76 @@
+---
+title: 'ADR Template'
+---
```
matthchr, line 3: Maybe I should force this early in the order?

theunrepentantgeek:
We already had guidance on [writing an ADR](https://azure.github.io/azure-service-operator/design/) - should we beef that up instead?

matthchr:
> We already have guidance on [writing an ADR](https://azure.github.io/azure-service-operator/design/) - should we beef that up instead?

Hmm, I wasn't aware of that although I think the template solves a slightly different problem. Basically I wrote the template because I always find myself copying an existing ADR to write a new one and then just deleting the items in each section, which is what made me think a template would be useful.

Even now that I am aware of the instructions in `_index.md` it's not actually markdown formatted so I can't easily just copy+paste it to get started.

What do you think about possibly...
- Moving the template to a subfolder (so it's more out of the way)
- Linking to it from the index
- Possibly moving some of the description and optional structure into the index document as opposed to the template? (So the template is just a template without much explanatory text)

theunrepentantgeek:
Yeah, I like that approach.

matthchr:
I'll update the PR

matthchr:
I did the above, but realized that putting it into a folder doesn't actually make it less discoverable on the UI because folders and file are both just rendered as TOC entries, so I swapped to using `toc_hide: true` instead, so the template is right there but it doesn't show up to readers of the site on the nav.

super-harsh (approved):

theunrepentantgeek (approved):

PR #4152, opened 2024-07-08
Add support for Azure RedhatOpenshift Clusters

### LINKED ISSUE #2802: Feature: Support Azure RedHat OpenShift cluster creation through ASO
The Swagger is here: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/stable/2022-09-04/redhatopenshift.json

super-harsh:
Closes #2802 

This PR adds support for 2023-11-22 Azure RedhatOpenshift Clusters.

Have excluded samples testing and controller tests from running in live mode since we require Service Principal credentials to run them. 

**If applicable**:
- [x] this PR contains tests


v2/internal/testcommon/vcr/redact.go
```diff
@@ -131,6 +132,13 @@ func hidePasswords(s string) string {
 	return passwordMatcher.ReplaceAllLiteralString(s, "\"{PASSWORD}\"")
 }
 
+var secretMatcher = regexp.MustCompile("\".*[Ss]ecret.*\"")
```
super-harsh, line 135: TODO: Need to fix this.

matthchr:
Looks like an APIManagement test flake possibly? May need to re-run test-generator.

v2/internal/testcommon/vcr/redact.go
```diff
@@ -131,6 +132,18 @@ func hidePasswords(s string) string {
 	return passwordMatcher.ReplaceAllLiteralString(s, "\"{PASSWORD}\"")
 }
 
+var (
+	secretMatcher  = regexp.MustCompile(`"([A-Za-z]+)?[Ss]ecret":".*"`)
```
matthchr, line 136: Are we sure this doesn't match other fields that aren't secrets and don't need to be redacted? Should we be looking specifically for `clientSecret`?

Also did the `keyMatcher` not work for this?

matthchr, line 136: If we do need this, consider leaving a comment describing what it does and why we need it in addition to the other matchers. See for example the matchers here: toolkit/testhttprecorder/recorder.go

In fact yours maybe should go there too?

v2/azure-arm.yaml
```diff
@@ -2839,6 +2839,29 @@ objectModelConfiguration:
       ResourceGroup:
         $export: true
         $supportedFrom: v2.0.0-alpha.1
+  redhatopenshift:
+    2023-11-22:
+      OpenShiftCluster:
+        $export: true
+        $supportedFrom: v2.8.0
```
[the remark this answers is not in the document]
super-harsh: True, I had this from pre 2.8.0. Thanks for pointing out, will update.

v2/internal/controllers/crd_redhatopenshift_openshiftcluster_20230401_test.go
```diff
@@ -0,0 +1,129 @@
...
+			Location: tc.AzureRegion, // Not available in westus2
+			Owner:    testcommon.AsOwner(rg),
+			ApiserverProfile: &aro.APIServerProfile{
+				Visibility: to.Ptr(aro.Visibility_Private),
+			},
+			ClusterProfile: &aro.ClusterProfile{
+				Domain:  to.Ptr("aro-example.com"),
+				Version: to.Ptr("4.14.16"),
+				// This is RG below is created by the RP.
+				ResourceGroupId:      to.Ptr(fmt.Sprintf("/subscriptions/%s/resourceGroups/%s", tc.AzureSubscription, tc.Namer.GenerateName("cluster-rg"))),
```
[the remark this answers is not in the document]
super-harsh: We have to supply this.

v2/internal/controllers/crd_redhatopenshift_openshiftcluster_20230401_test.go
```diff
@@ -0,0 +1,129 @@
...
+	roleAssingmentToVnet := newRoleAssignment(tc, vnet, "roleassingment", contributorRoleId)
+	roleAssingmentToVnet.Spec.PrincipalId = to.Ptr(principalId)
+
+	roleAssingmentFromRPtoVnet := newRoleAssignment(tc, vnet, "rollassginmentrp", contributorRoleId)
+	roleAssingmentFromRPtoVnet.Spec.PrincipalId = to.Ptr(azureRedHadOpenshiftRPIdentityPrincipalId)
+
+	cluster := &aro.OpenShiftCluster{
+		ObjectMeta: tc.MakeObjectMeta("aro-cluster"),
+		Spec: aro.OpenShiftCluster_Spec{
+			Location: tc.AzureRegion, // Not available in westus2
```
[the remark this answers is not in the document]
super-harsh: My bad, will clean this up

v2/internal/controllers/crd_redhatopenshift_openshiftcluster_20230401_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+	tc := globalTestContext.ForTest(t)
+	rg := tc.CreateTestResourceGroupAndWait()
+
+	secretName := "aro-secret"
+	secretKey := "client-secret"
+	// TODO: Replace the principalID, clientSecret and clientId vars below with principalId, clientSecret and clientId of your SP
+	principalId := "5c6be76c-5fc4-4817-992d-22027b44c402"
+	clientId := "5b7a18b9-8aec-4456-a4c3-0865cbfa1512"
+	clientSecret := "your-client-secret-here"
```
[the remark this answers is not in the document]
super-harsh: Nice idea for the clientSecret, although the ClientId and PrincipalId don't get redacted in the tests hence I've left them here. Maybe we need another matcher to match guids and replace them?

v2/internal/testcommon/vcr/redact.go
super-harsh: keyMatcher does not work for this. I've added a comment and and updated the regex to match `([A-Za-z]+)[Ss]ecret":".*`. Let me know wdyt about that?

v2/internal/controllers/crd_redhatopenshift_openshiftcluster_20230401_test.go
```diff
@@ -0,0 +1,139 @@
...
+		t.Skip("ARO cluster test requires ServicePrincipal creation and referenced in the cluster object.")
+	}
+
+	tc := globalTestContext.ForTest(t)
+	rg := tc.CreateTestResourceGroupAndWait()
+
+	secretName := "aro-secret"
+	secretKey := "client-secret"
+	// TODO: Replace the principalID, clientSecret and clientId vars below with principalId, clientSecret and clientId of your SP
+	principalId := "5c6be76c-5fc4-4817-992d-22027b44c402"
```
matthchr, line 39: I think these should be env variables too, or hardcoded to 00000's GUIDs?

Is the issue with changing these that the test recordings don't pass then?

v2/internal/controllers/crd_redhatopenshift_openshiftcluster_20230401_test.go
```diff
@@ -0,0 +1,172 @@
...
+	tc.WithLiteralRedaction(details.clientId, "{REDACTED}")
+
+	serviceEndpoints := []network.ServiceEndpointPropertiesFormat{
+		{
+			Service: to.Ptr("Microsoft.ContainerRegistry"),
+		},
+	}
+
+	vnet := newVNet(tc, testcommon.AsOwner(rg), []string{"10.100.0.0/15"})
+	masterSubnet := newSubnet(tc, vnet, "10.100.76.0/24")
```
matthchr, line 63: There's a bunch of vnet addressspaces here... how did you pick them all?
If it was from a specific example, maybe include the link in a comment?

v2/internal/testcommon/resource_namer.go
```diff
@@ -145,17 +145,24 @@ func (n ResourceNamer) GeneratePassword() string {
...
 
 	// This pass + <content> + pass pattern is to make it so that matchers can reliably find and prune
 	// generated passwords from the recordings. If you change it make sure to change the passwordMatcher
 	// in test_context.go as well.
 	return "pass" + n.makeSecureStringOfLength(length, runes) + "pass"
 }
 
+// GenerateSecretOfLength generates and returns a non-deterministic secret.
+// This method does not use any seed value, so the returned password is never stable.
+func (n ResourceNamer) GenerateSecretOfLength(length int) string {
```
matthchr, line 162: Is this required because the pass + pass wrapping is rejected by ARO?

v2/internal/testcommon/kube_per_test_context.go
```diff
@@ -495,6 +495,33 @@ func (tc *KubePerTestContext) PatchResourceAndWaitForState(
 	tc.Eventually(new).Should(tc.Match.BeInState(status, severity, gen))
 }
 
+func (tc *KubePerTestContext) CreateSecret(
```
matthchr, line 498: Wondering if this should be called `CreateSimpleSecret`, and should have method comments explaining that if data isn't specified it'll be autogenerated?

matthchr: Though TBH I think you could maybe just drop this bit:
```
	if secretData == "" {
		secretData = tc.Namer.GeneratePasswordOfLength(40)
	}
```

because even with this the usage is 

`tc.CreateSecret("mysecret", "mykey", "")` which I am not sure is clearer than `tc.CreateSecret("mysecret", "mykey", tc.GeneratePasswordOfLength(40)` and even though the second _is_ a bit longer it's also (IMO) clearer about what's happening compared to the first one.

matthchr (approved):

v2/internal/controllers/crd_redhatopenshift_openshiftcluster_20230401_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+	if id := os.Getenv("ARO_OBJECT_ID"); id != "" {
+		objectId = id
+	}
+
+	if secret := os.Getenv("ARO_CLIENT_SECRET"); secret != "" {
+		clientSecret = secret
+	}
+
+	tc.WithLiteralRedaction(clientId, nilUUID)
```
matthchr, line 163: I think you don't want to add these redactions if there's nothing to redact (input is 0000's already), as it's causing the tests to fail.

v2/internal/testcommon/resource_namer.go
super-harsh: yep

v2/internal/controllers/crd_redhatopenshift_openshiftcluster_20230401_test.go
super-harsh: I picked them randomly

PR #4158, opened 2024-07-09
Feature/ratelimit relax

### LINKED ISSUE #4150: Remove or relax self throttling, or make it configurable
We self-throttle to avoid ARM throttling:
https://github.com/Azure/azure-service-operator/blob/807a2bb52d0cbc9f4db424fb0fc30ca2bcbacda9/v2/internal/reconcilers/generic/generic_reconciler.go#L278-L295

**Describe the improvement**
ARM has since changed how they throttle, and we have moved to support multiple identities/multiple subscriptions, which means that this simple bucket-based throttling is no-longer sufficient.

We should either remove the self-throttling entirely, OR make it larger and possibly configurable. I tend towards removing it given that:
- ARMs new limits are harder to predict (based on subscription size and other things).
- Throttling is per identity (I believe) which means if the user is using per-namespace identities they'll get more throughput, which we now allow but didn't in 2022.

matthchr:
* Add MAX_CONCURRENT_RECONCILES option to Helm chart and pod
* Remove default ratelimiter and add ratelimiter configuration

Fixes #4150

**If applicable**:
- [x] this PR contains tests


docs/hugo/content/guide/aso-controller-settings-options.md
```diff
@@ -156,3 +156,72 @@ AZURE_USER_AGENT_SUFFIX is appended to the default User-Agent for Azure HTTP cli
...
+
+**Example:** `2`
+
+**Required**: False
+
+### RATE_LIMIT_MODE
+
+RateLimitMode configures the internal rate-limiting mode.
+
+ * disabled: No ASO-controlled rate-limiting occurs. ASO will attempt to communicate with Azure and
```
theunrepentantgeek, line 185: Suggest bolding the values and indicating the default.

```suggestion
 * **disabled** (default): No ASO-controlled rate-limiting occurs. ASO will attempt to communicate with Azure and
```

docs/hugo/content/guide/aso-controller-settings-options.md
```diff
@@ -156,3 +156,72 @@ AZURE_USER_AGENT_SUFFIX is appended to the default User-Agent for Azure HTTP cli
...
+   kube-apiserver as much as needed based on load. It will back off based on throttling from
+   either kube-apiserver or Azure, but will not artificially limit its throughput.
+ * bucket: Uses a token-bucket algorithm to rate-limit reconciliations. Note that this limits how often
```
theunrepentantgeek, line 188:
```suggestion
 * **bucket**: Uses a token-bucket algorithm to rate-limit reconciliations. Note that this limits how often
```

theunrepentantgeek (approved):

PR #4168, opened 2024-07-31
Read aso-controller-settings ConfigMap via volume rather than env variables

matthchr:
Moving from a model where we have to explicitly set each secret in the pod env to a model where we just mount the whole secret into the pod. This has the added benefit of allowing us to autodetect secret changes and restart the pod AND means that users can totally omit the secret now as opposed to having to create the secret with empty values.

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/v2/internal/config/watcher.go
+++ b/v2/internal/config/watcher.go
@@ -1,0 +1,81 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package config
+
+import (
+	"context"
+	"os"
+
+	"github.com/fsnotify/fsnotify"
+	"github.com/go-logr/logr"
+
+	. "github.com/Azure/azure-service-operator/v2/internal/logging"
+)
+
+type Watcher struct {
+	path string
+
+	ctx    context.Context
+	cancel context.CancelFunc
+	logger logr.Logger
+}
+
+func NewWatcher(logger logr.Logger) *Watcher {
+	logger = logger.WithName("configwatcher")
+
+	return &Watcher{
+		path:   MountPath,
+		logger: logger,
+	}
+}
+
+func (w *Watcher) Start(ctx context.Context) error {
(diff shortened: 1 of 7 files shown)


theunrepentantgeek (dismissed):

matthchr:
I may hold off on merging this until we re-enable KIND tests as they're the primary test of this change. While I've tested it locally and it works it may be safer to wait to merge until those tests are re-enabled.

v2/internal/config/watcher.go
```diff
@@ -0,0 +1,80 @@
...
+		var ok bool
+		var event fsnotify.Event
+
+		select {
+		case <-w.ctx.Done():
+			w.logger.Error(err, "Aborting watcher")
+			return
+		case event, ok = <-watcher.Events:
+			if !ok {
+				continue
```
super-harsh, line 66: Should return here rather than continue? 
Since if a channel return `bool==false` that means channel is closed and cannot receive anymore?

v2/internal/config/watcher.go
```diff
@@ -0,0 +1,80 @@
...
+			}
+			if event.Has(fsnotify.Write | fsnotify.Create | fsnotify.Remove | fsnotify.Rename) {
+				w.logger.V(Info).Info("Secret updated, exiting...")
+				// TODO: Ideally we would give up the lease here too
+				os.Exit(0)
+			}
+		case err, ok = <-watcher.Errors:
+			if !ok {
+				continue
+			}
```
super-harsh, line 76: same here as above

v2/internal/config/watcher.go
matthchr: Good catch - fixed

matthchr:
This has some possible disadvantages during upgrades that change the configmap, as there's just one ConfigMap which means the old pod may exit early (due to configmap change).

Need to do some testing around that.

matthchr:
Closing this due to above issues

PR #4170, opened 2024-07-31
Flatten ARM type references

theunrepentantgeek:

In preparation for moving the ARM types into subpackages, we need to eliminate reuse of CRD types by the ARM types (if we don't, we'll get circular reference errors and things won't work). 

This PR takes the first step, replacing use of simple primitive type aliases with the underlying types.

Currently only handling single values, will have a followup PR to handle collections.

gif

### DIFF
--- a/v2/tools/generator/internal/codegen/pipeline/create_arm_types.go
+++ b/v2/tools/generator/internal/codegen/pipeline/create_arm_types.go
@@ -95,6 +95,7 @@ func newARMTypeCreator(
 
 	result.visitor = astmodel.TypeVisitorBuilder[any]{
 		VisitInternalTypeName: result.visitARMTypeName,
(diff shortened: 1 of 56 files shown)


v2/tools/generator/internal/codegen/pipeline/create_arm_types.go
```diff
@@ -520,6 +521,28 @@ func (c *armTypeCreator) visitARMTypeName(
 		return astmodel.CreateARMTypeName(def.Name()), nil
 	}
 
+	// If the name is an alias for a primitive type, we look through it to the underlying type
```
matthchr, line 524: minor: Consider expanding this comment from the "what" we are doing, to the "why" (we flatten the types to break the circular dependency between ARM types and Kube types)

matthchr (approved):

PR #4173, opened 2024-08-01
Add support to hide custom recording data

### LINKED ISSUE #4172: Add support to hide custom recording data while recording tests

Currently, we only support pre-defined regex matchers to redact/hide recordings data and don't have support to hide custom test data while recording tests dynamically

**Describe the improvement**

Add a function to support redaction of custom data like ClientIds, PrincipalIds etc in the test so that no sensitive data is added into the recordings. This can be configured dynamically depending upon the needs of the test.

super-harsh:
Closes #4172 

Currently, we only support pre-defined regex matchers to redact/hide recordings data and don't have support to hide custom test data while recording tests dynamically. This PR introduces methods on TestContext to `AddLiteralRedaction` and `AddRegexpRedaction`. Also includes some refactoring to add `Redactor` struct.


v2/internal/genericarmclient/generic_client_test.go
```diff
@@ -36,7 +36,7 @@ func Test_NewResourceGroup(t *testing.T) {
 	cfg, err := config.ReadFromEnvironment()
 	g.Expect(err).ToNot(HaveOccurred())
 
-	testContext, err := testContext.ForTest(t, cfg)
+	testContext, err := testContext.ForTest(t, cfg, nil)
```
super-harsh, line 39: TODO: Add a test

theunrepentantgeek: Forcing consumers to pass a nil value for things they aren't interested in is a bad API design - especially given most consumers won't care.

v2/internal/testcommon/test_context.go
```diff
@@ -87,11 +88,11 @@ func NewTestContext(
 	}
 }
 
-func (tc TestContext) ForTest(t *testing.T, cfg config.Values) (PerTestContext, error) {
+func (tc TestContext) ForTest(t *testing.T, cfg config.Values, hideCustomData map[string]string) (PerTestContext, error) {
```
theunrepentantgeek, line 91: Using a map here makes this complex to set up when it is needed.

Why not keep the original signature for the factory method, but add a `WithRedaction(regex string, placeholder string)` method that can be used to add any required additional rules. It'd return the `TestContext` to allow chaining.

v2/internal/testcommon/test_recorder.go
```diff
@@ -16,12 +16,7 @@ import (
 )
 
 // createTestRecorder returns an instance of testRecorder to allow recording and playback of HTTP requests.
-func createTestRecorder(
-	cassetteName string,
-	cfg config.Values,
-	recordReplay bool,
-	log logr.Logger,
-) (vcr.Interface, error) {
+func createTestRecorder(cassetteName string, cfg config.Values, recordReplay bool, log logr.Logger, hideCustomData map[string]string) (vcr.Interface, error) {
```
theunrepentantgeek, line 19: Please keep long signatures with one parameter per row, it's a lot easier to read.

v2/internal/testcommon/vcr/redact.go
```diff
@@ -128,7 +136,7 @@ var passwordMatcher = regexp.MustCompile("\"pass[^\"]*?pass\"")
 
 // hidePasswords hides anything that looks like a generated password
 func hidePasswords(s string) string {
-	return passwordMatcher.ReplaceAllLiteralString(s, "\"{PASSWORD}\"")
+	return passwordMatcher.ReplaceAllLiteralString(s, "\"{REDACTED}\"")
```
theunrepentantgeek, line 139: I think we want this to stay as `PASSWORD` so that we can see what the value was in the recording.

Plus, doing this will break many of the existing recordings.

v2/internal/testcommon/vcr/redact.go
```diff
@@ -147,7 +155,7 @@ func hideKubeConfigs(s string) string {
 var keyMatcher = regexp.MustCompile("(?:[A-Za-z0-9+/]{4}){10,}(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)")
 
 func hideKeys(s string) string {
-	return keyMatcher.ReplaceAllLiteralString(s, "{KEY}")
+	return keyMatcher.ReplaceAllLiteralString(s, "{REDACTED}")
```
theunrepentantgeek, line 158: Ditto here - we used `KEY` for a reason, so we know what kind of value was removed.

v2/internal/testcommon/vcr/v3/recordings/TestRecorder_WhenRecordingAndRecordingDoesNotExist_MakesRecording.yaml
```diff
@@ -0,0 +1,66 @@
+---
+version: 2
```
theunrepentantgeek, line 2: Given the title of this test, I don't think this file should be committed.

v2/internal/testcommon/vcr/v3/test_recorder.go
```diff
@@ -30,11 +30,7 @@ type recorderDetails struct {
 	log          logr.Logger
 }
 
-func NewTestRecorder(
-	cassetteName string,
-	cfg config.Values,
-	log logr.Logger,
-) (vcr.Interface, error) {
+func NewTestRecorder(cassetteName string, cfg config.Values, log logr.Logger, hideCustomData map[string]string) (vcr.Interface, error) {
```
theunrepentantgeek, line 33: Please keep parameters on separate lines for ease of reading.

theunrepentantgeek (changes requested):
I think the API needs work - have made comments indicating how it could be improved.

Also, I don't think you should need to be passing `hideCustomData` around that much, it feels like something that should be stored on the TestContext or maybe the TestRecorder so that's available to methods when needed. 

The current code has a few too many cargo parameters - places where something is just carried through, not consumed.

v2/internal/testcommon/test_context.go
```diff
@@ -33,9 +33,10 @@ import (
 var DefaultTestRegion = "westus2" // Could make this an env variable if we wanted
 
 type TestContext struct {
-	AzureRegion  *string
-	NameConfig   *ResourceNameConfig
-	RecordReplay bool
+	AzureRegion     *string
+	NameConfig      *ResourceNameConfig
+	RecordReplay    bool
+	customRedactMap map[string]string
```
matthchr, line 39: minor: consider something like `recordingRedactions`?

v2/internal/testcommon/test_recorder.go
```diff
@@ -21,7 +21,7 @@ func createTestRecorder(
 	cfg config.Values,
 	recordReplay bool,
 	log logr.Logger,
-) (vcr.Interface, error) {
+	customRedactMap map[string]string) (vcr.Interface, error) {
```
matthchr, line 24:
```suggestion
	redactions map[string]string) (vcr.Interface, error) {
```

or maybe just `redact`?

v2/internal/testcommon/vcr/redact.go
```diff
@@ -109,6 +109,17 @@ func HideRecordingData(azureIDs creds.AzureIDs, s string) string {
 	return s
 }
 
+func HideRecordingDataWithCustomRedaction(azureIDs creds.AzureIDs, s string, redact map[string]string) string {
+	// Replace and hide all the custom data
+	if redact != nil {
```
matthchr, line 114: It looks guaranteed to me now that redact is not nil? Even if it is nil, it's OK to `range` over a nil map it shouldn't panic right?

v2/internal/testcommon/vcr/v3/error_translating_roundtripper.go
```diff
@@ -33,13 +33,19 @@ import (
...
+		t:               t,
+	}
 }
 
 type errorTranslation struct {
-	recorder     http.RoundTripper
-	cassetteName string
+	recorder        http.RoundTripper
+	cassetteName    string
+	customRedactMap map[string]string
```
matthchr, line 48: minor, nmaing: redactions or redact might be a better name here too?

v2/internal/testcommon/vcr/redact.go
```diff
@@ -109,6 +109,17 @@ func HideRecordingData(azureIDs creds.AzureIDs, s string) string {
...
+		for k, v := range redact {
+			s = strings.ReplaceAll(s, k, v)
+		}
+	}
+
+	return HideRecordingData(azureIDs, s)
```
matthchr, line 120: Do we ever just call `HideRecordingData` now? If not make it unexported possibly? Or just inline it in here and combine these two methods if nobody ever calls it except this?

v2/internal/testcommon/vcr/v3/test_recorder.go
```diff
@@ -22,18 +22,20 @@ import (
...
-	recorder     *recorder.Recorder
-	cfg          config.Values
-	log          logr.Logger
+	cassetteName    string
+	creds           azcore.TokenCredential
+	ids             creds.AzureIDs
+	recorder        *recorder.Recorder
+	cfg             config.Values
+	log             logr.Logger
+	customRedactMap map[string]string
```
matthchr, line 31: same comment about possibly cleaner name

v2/internal/testcommon/vcr/v3/tracking_roundtripper.go
```diff
@@ -34,17 +34,19 @@ const (
 type requestCounter struct {
 	inner http.RoundTripper
 
-	countsMutex sync.Mutex
-	counts      map[string]uint32
-	azureIDs    creds.AzureIDs
+	countsMutex     sync.Mutex
+	counts          map[string]uint32
+	azureIDs        creds.AzureIDs
+	customRedactMap map[string]string
```
matthchr, line 40: and here

v2/internal/testcommon/vcr/redact.go
super-harsh: We do call it for v1.vcr recorder. Since we don't use it anymore for recordings so no point supporting redactions there.

v2/internal/testcommon/vcr/v3/tracking_roundtripper.go
super-harsh: I've replaced the map with a redactor struct and renamed wherever possible.

v2/internal/testcommon/vcr/redact.go
theunrepentantgeek: `v1.vcr` doesn't have a recorder - BUT the v1 player does need access to redactions, because we redact URLs. If we don't have those, we can't match the URLs for playback from the recording.

theunrepentantgeek (changes requested):
Ideally we'll solve the fragility problem before merging this.

v2/internal/testcommon/vcr/v3/test_recorder.go
```diff
@@ -28,12 +28,14 @@ type recorderDetails struct {
...
 	cfg          config.Values
 	log          logr.Logger
+	redactions   map[string]string
 }
 
 func NewTestRecorder(
 	cassetteName string,
 	cfg config.Values,
 	log logr.Logger,
+	redactions map[string]string,
```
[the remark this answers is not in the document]
super-harsh: Makes sense, in my latest commit I've solved this problem by implementing facade pattern to update the map within `Redactor` struct via TestContext.

v2/internal/testcommon/vcr/redact.go
```diff
@@ -16,6 +16,22 @@ import (
 	"github.com/Azure/azure-service-operator/v2/internal/testcommon/creds"
 )
 
+type Redactor struct {
```
theunrepentantgeek, line 19: Creating an explicit `Redactor` to capture all the desired redactions is a good idea. It makes the coupling visible, which is much less fragile.

v2/internal/testcommon/vcr/redact.go
```diff
@@ -109,6 +128,15 @@ func HideRecordingData(azureIDs creds.AzureIDs, s string) string {
 	return s
 }
 
+func (r *Redactor) HideRecordingDataWithCustomRedaction(s string) string {
```
theunrepentantgeek, line 131: I'd make this internal, and just call it from all the existing entry points; that way we can guarantee these custom redactions are applied everywhere. It'd be unfortunate to redact something from the body of a request, but leave it present in headers, for example.

This would also mean consumers wouldn't need to change as they'd still be calling `HideRecordingData()`

v2/internal/testcommon/vcr/redact.go
```diff
@@ -16,6 +16,22 @@ import (
 	"github.com/Azure/azure-service-operator/v2/internal/testcommon/creds"
 )
 
+type Redactor struct {
+	azureIDs   creds.AzureIDs
+	redactions map[string]string
```
theunrepentantgeek, line 21: Suggest renaming this to `literalRedactions` - better indicates what it does, and leaves the door open for adding `regexRedactions` later (which I suspect we'll need).

v2/internal/testcommon/vcr/redact.go
```diff
@@ -16,6 +16,22 @@ import (
 	"github.com/Azure/azure-service-operator/v2/internal/testcommon/creds"
 )
 
+type Redactor struct {
+	azureIDs   creds.AzureIDs
```
theunrepentantgeek, line 20: Not for this PR, but we might want to get rid of this - and just use the value in `NewRedactor` to preinitialize values in `redactions`.

v2/internal/testcommon/vcr/recorder.go
```diff
@@ -26,6 +27,9 @@ type Interface interface {
 	// IDs returns the available Azure resource IDs for the test
 	IDs() creds.AzureIDs
 
+	// Redactor returns the redactor associated with the recorder
+	Redactor() *Redactor
```
theunrepentantgeek, line 31: Instead of providing direct access to the redactor (which is kinda an implementation detail, albeit an important one), how about changing this method to add a redaction?

```suggestion
	AddRedaction(redactionValue string, replacementValue string)
```

theunrepentantgeek (commented):
Great approach - makes the coupling explicit and clear. I have a few suggestions on improvements, but they're just adding polish, not suggesting a different direction.

v2/internal/testcommon/vcr/redact.go
super-harsh: Actually, I've got rid of the map finally, and using a struct + supporting regexp replacements too. Took suggestions from the pointers you shared.

v2/internal/testcommon/vcr/redact.go
super-harsh: yep, sounds good. I've taken the suggestion and calling the `hideRecordingDataWithCustomRedaction` from other methods(HideURL, HideHeaders) as well

v2/internal/testcommon/vcr/redact.go
super-harsh: Nice! I've taken this too

v2/internal/testcommon/vcr/redact.go
```diff
@@ -8,14 +8,52 @@ package vcr
...
+	return redactor
+}
+
+func (r *Redactor) AddLiteralRedaction(redactionValue string, replacementValue string) {
+	pattern := regexp.QuoteMeta(redactionValue)
+	r.AddRegexRedaction(pattern, replacementValue)
+}
+
+func (r *Redactor) AddRegexRedaction(regex string, replacementValue string) {
+
```
matthchr, line 48: minor: Excess newline

matthchr (approved):
Nice!

theunrepentantgeek (approved):
This is now looking really good.

PR #4179, opened 2024-08-06
Use crypto/rand while generating random passwords

### LINKED ISSUE #4061: Fix code scanning alert - Randomness of passwords generated for tests
The generated passwords are only ever transient, but we still think this should be fixed.

Tracking issue for:
- [ ] https://github.com/Azure/azure-service-operator/security/code-scanning/2

super-harsh:
Closes #4061 

This PR adds a method `makeSecureStringOfLength` to use `crypto/rand` package for generating secure password strings.


v2/internal/testcommon/resource_namer.go
```diff
@@ -103,6 +105,18 @@ func (n ResourceNamer) makeRandomStringOfLength(num int, runes []rune) string {
 	return string(result)
 }
 
+func (n ResourceNamer) makeSecureStringOfLength(num int, runes []rune) string {
```
matthchr, line 108: Before, these passwords were deterministic, and now they are not.

Should we seed the crypto randomizer? 
I guess there's no way to do that in the crypto API?

Might be worth making a note on this method and maybe the password method too that it's not stable...

v2/internal/testcommon/resource_namer.go
```diff
@@ -132,7 +146,7 @@ func (n ResourceNamer) GeneratePasswordOfLength(length int) string {
 	// This pass + <content> + pass pattern is to make it so that matchers can reliably find and prune
```
matthchr, line 146: I thought about suggesting we also expand the collection of runes to choose from here (in the spirit of making things more secure) but realized that's likely to cause problems as at least some applications we use this for (SQL, etc) may be picky about the charset of passwords...

For now I think this is probably OK.

v2/internal/controllers/crd_apimanagement_20230501preview_test.go
```diff
@@ -38,7 +38,7 @@ func Test_ApiManagement_20230501preview_CRUD(t *testing.T) {
 	// Create an APIM instance. APIM has a soft delete feature; if you find that you
 	// hit this problem add the `restore` back in to resurrect it
 	service := apim.Service{
-		ObjectMeta: tc.MakeObjectMetaWithName(tc.Namer.GenerateName("apimv2")),
+		ObjectMeta: tc.MakeObjectMetaWithName(tc.Namer.GenerateName("apimanagementv2")),
```
super-harsh, line 41: Had to update these since the resource with generated name already existed somewhere(not in our sub).

matthchr (approved):

PR #4196, opened 2024-08-19
Add support for RoleAssignments random UUID name generation

### LINKED ISSUE #3637: Bug: RoleAssignment UUID clashes

super-harsh:
Closes #3637 

This PR adds in a custom UUIDGeneration `OperatorSpecProperty` to determine whether to use `random` or `deterministic` approach for RoleAssignment name generation.


v2/azure-arm.yaml
```diff
@@ -1132,6 +1132,10 @@ objectModelConfiguration:
         $export: true
         $supportedFrom: v2.4.0
         $defaultAzureName: false
+        $operatorSpecProperties:
+          - name: UUIDGeneration
```
theunrepentantgeek, line 1136: Is this the best name for the property? 

I used `NamingConvention` in my demo of the OperatorSpec implementation; maybe there's something better than either?

v2/api/authorization/v1api20220401/role_assignment_defaults.go
```diff
@@ -43,13 +45,18 @@ func (assignment *RoleAssignment) defaultAzureName() {
...
-		gk := assignment.GroupVersionKind().GroupKind()
-		assignment.Spec.AzureName = randextensions.MakeUUIDName(
-			assignment.Name,
-			randextensions.MakeUniqueOwnerScopedStringLegacy(
-				assignment.Owner(),
-				gk,
-				assignment.Namespace,
-				assignment.Name))
+		if assignment.Spec.OperatorSpec != nil &&
+			strings.ToLower(*assignment.Spec.OperatorSpec.UUIDGeneration) == "random" {
```
theunrepentantgeek, line 49: Use `strings.EqualFold()` instead.

v2/api/authorization/v1api20220401/role_assignment_defaults.go
```diff
@@ -43,13 +45,18 @@ func (assignment *RoleAssignment) defaultAzureName() {
...
+			assignment.Spec.AzureName = randextensions.MakeRandomUUID()
+		} else {
```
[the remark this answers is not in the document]
theunrepentantgeek: If the user supplied value isn't one of the enum, we won't assign `AzureName` and we'll get a natural error.

Also, update the doc-comment above to reflect the new behaviour.

v2/api/authorization/v1api20220401/role_definition_defaults.go
```diff
@@ -35,6 +37,10 @@ func (definition *RoleDefinition) defaultAzureName() {
 	}
 
 	if definition.AzureName() == "" {
+		if definition.Spec.OperatorSpec != nil &&
+			strings.ToLower(*definition.Spec.OperatorSpec.UUIDGeneration) == "random" {
+			definition.Spec.AzureName = randextensions.MakeRandomUUID()
+		}
```
theunrepentantgeek, line 43: Whatever is set here will be immediately overwritten by line 45

v2/azure-arm.yaml
theunrepentantgeek: Also, don't we need to specify this on both versions of `RoleAssignment`, and both versions of `RoleDefinition`?

theunrepentantgeek (changes requested):

v2/azure-arm.yaml
super-harsh: Bit unsure, should we specify on both versions?

v2/api/authorization/v1api20220401/role_assignment_defaults.go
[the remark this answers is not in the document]
super-harsh: Wouldn't that be a breaking change? 
Thought would be good if we do the default behavior if nothing is mentioned? 

Also, I'd update `deterministic` to be `default`

v2/azure-arm.yaml
super-harsh: Do you have a suggestion for naming? or something from the list below you prefer? 
- `GenerationType`
- `NamingConvention`
- `AzureNameGeneration`

v2/api/authorization/v1api20220401/role_assignment_defaults.go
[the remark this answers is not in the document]
super-harsh: or I think `stable` and `random` would be good too. WDYT?

v2/azure-arm.yaml
theunrepentantgeek: I like `NamingConvention` (obviously!) - also like `AzureNameGeneration` too.

theunrepentantgeek: Naming anything is hard. Naming about naming, doubly so! 😁

v2/api/authorization/v1api20220401/role_assignment_defaults.go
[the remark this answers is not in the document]
theunrepentantgeek: My concern is that if someone set the property to `Jamaican` we would accept that and do something with it! I really don't think we should.

I'd retain the existing behavour if someone leaves it blank (`""`) _or_ requests it explicitly (`"stable"`), and the new behaviour if someone sets it to `"random"`.

I don't really favour `default` because it's not clear what that means, other than it's the default. I like `stable` though, that's good.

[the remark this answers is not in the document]
super-harsh: We have the validations to take care of that case if the user sets a disallowed value. I think that should take care of that?

theunrepentantgeek (approved):

PR #4201, opened 2024-08-20
Make testing on an AKS cluster a bit easier

matthchr:
Adding a DIR override to some tarets so we can point them at the AKS directory rather than the default KIND directory, in case we want to run the tests against an AKS cluster.

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -356,8 +356,9 @@ tasks:
         --set crdPattern=*"
       - task: controller:wait-for-operator-ready
     vars:
+      DIR: "{{.DIR | default .KIND_WORKLOAD_IDENTITY_PATH}}"
       AZURE_MI_CLIENT_ID:
-        sh: "cat {{.KIND_WORKLOAD_IDENTITY_PATH}}/azure/miclientid.txt"
+        sh: "cat {{.DIR}}/azure/miclientid.txt"
 
   controller:test-upgrade-pre:
     desc: Test upgrading {{.CONTROLLER_APP}} and helm chart for new release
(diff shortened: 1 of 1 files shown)


Taskfile.yml
```diff
@@ -356,8 +356,9 @@ tasks:
         --set crdPattern=*"
       - task: controller:wait-for-operator-ready
     vars:
+      DIR: "{{default .KIND_WORKLOAD_IDENTITY_PATH .DIR}}"
```
theunrepentantgeek, line 359: I find it clearer to use `|` syntax with `default` as the argument order trips me up otherwise (having the fallback value first is weird, even though required by the way templates work).

```suggestion
      DIR: "{{.DIR | default .KIND_WORKLOAD_IDENTITY_PATH}}"
```

Taskfile.yml
```diff
@@ -368,8 +369,9 @@ tasks:
       - "kubectl create namespace pre-release"
       - "go test -timeout 15m -count=1 -v -run Test_Pre_Release_ResourceCanBeCreated_BeforeUpgrade ./test/pre-release"
     vars:
+      DIR: "{{default .KIND_WORKLOAD_IDENTITY_PATH .DIR}}"
```
theunrepentantgeek, line 372:
```suggestion
      DIR: "{{.DIR | default .KIND_WORKLOAD_IDENTITY_PATH}}"
```

Taskfile.yml
```diff
@@ -379,8 +381,9 @@ tasks:
       - task: controller:package-helm-manifest
       - task: controller:install-helm-wi
         vars:
+          DIR: "{{default .KIND_WORKLOAD_IDENTITY_PATH .DIR}}"
```
theunrepentantgeek, line 384:
```suggestion
          DIR: "{{.DIR | default .KIND_WORKLOAD_IDENTITY_PATH}}"
```

Taskfile.yml
```diff
@@ -962,8 +965,9 @@ tasks:
       - "{{.SCRIPTS_ROOT}}/deploy-testing-secret.sh workloadidentity"
     env:
       # Override the AZURE_CLIENT_ID env variable here
+      DIR: "{{default .KIND_WORKLOAD_IDENTITY_PATH .DIR}}"
```
theunrepentantgeek, line 968:
```suggestion
      DIR: "{{.DIR | default .KIND_WORKLOAD_IDENTITY_PATH}}"
```

theunrepentantgeek (approved):

PR #2816, opened 2023-03-28
Import Azure Resources

theunrepentantgeek:
**What this PR does / why we need it**:

Adds the verb `asoctl import azure-resource` for importing resources from Azure Resource Manager.

Refactored extensively based on feedback given to #2728, to the point where those old comments had little relevance.

Sample usage (redacted)
```
$ asoctl import azure-resource /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/<rg>/providers/Microsoft.DBforPostgreSQL/flexibleServers/pg
---
apiVersion: dbforpostgresql.azure.com/v1beta20210601
kind: FlexibleServer
metadata:
  creationTimestamp: null
  name: pg
spec:
  administratorLogin: admin
  availabilityZone: "1"
  backup:
    backupRetentionDays: 7
    geoRedundantBackup: Disabled
  highAvailability:
    mode: Disabled
  location: East US
  maintenanceWindow:
    customWindow: Disabled
    dayOfWeek: 0
    startHour: 0
    startMinute: 0
  network: {}
  sku:
    name: Standard_B1ms
    tier: Burstable
  storage:
    storageSizeGB: 32
  version: "14"
status: {}
---
```
**Special notes for your reviewer**:

Closes #2805 

**How does this PR make you feel**:

gif

**If applicable**:
- [x] this PR contains tests

PR #4204, opened 2024-08-21
Reactivate some previously removed linters

matthchr:
Also fix the various errors reactivation found

**If applicable**:
- [ ] this PR contains tests


.golangci.yml
```diff
@@ -15,11 +17,16 @@ linters:
...
     - musttag # Extremely slow, at least on CI machines
 linters-settings:
   govet:
     check-shadowing: true
+  unused:
+    #field-writes-are-uses: false
+    generated-is-used: true
+  gosimple:
+    checks:
+      - "all"
```
matthchr, line 30: minor: Should leave link here...

.golangci.yml
```diff
@@ -15,11 +17,17 @@ linters:
...
-    - staticcheck
-    - gosimple
     - stylecheck
-    - unused
     - musttag # Extremely slow, at least on CI machines
 linters-settings:
   govet:
     check-shadowing: true
+  unused:
+    #field-writes-are-uses: false
```
matthchr, line 26: Fix this

theunrepentantgeek (approved):

v2/cmd/controller/main.go
```diff
@@ -17,7 +17,7 @@ import (
 
 func main() {
 	setupLog := ctrl.Log.WithName("setup")
-	ctrl.SetLogger(klogr.New())
+	ctrl.SetLogger(klogr.New()) //nolint: staticcheck
```
matthchr, line 20: We have a separate item tracking fixing this, and doing so in this PR was nontrivial so I skipped it.

super-harsh (approved):

v2/tools/generator/pkg/versions/versions.go
theunrepentantgeek, line 1: Pretty sure this is used by `asoctl` and by external consumers of `v2/api`.

theunrepentantgeek (changes requested):

v2/tools/generator/internal/codegen/pipeline/remove_status_property_validations.go
```diff
@@ -43,12 +43,10 @@ func RemoveStatusValidations() *Stage {
 				return nil, err
 			}
 
-			/* TODO(donotmerge)
 			err = errorIfSpecStatusOverlap(result, state.Definitions())
 			if err != nil {
 				return nil, err
 			}
-			*/
```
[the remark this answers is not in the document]
matthchr: My interpretation of this was the donotmerge was the fact it was commented out, not the fact the check was there. The check should be made and indeed passes as soon as I uncommented it.

v2/tools/generator/internal/codegen/pipeline/convert_allof_and_oneof_to_objects.go
```diff
@@ -399,9 +399,6 @@ func (s synthesizer) getOneOfName(t astmodel.Type, propIndex int) (propertyNames
 		// Try unwrapping the meta type and basing the name on what's inside
 		return s.getOneOfName(concreteType.Unwrap(), propIndex)
 
-	case *astmodel.OptionalType:
-		return s.getOneOfName(concreteType.Element(), propIndex)
-
 	case *astmodel.InterfaceType:
```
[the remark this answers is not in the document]
matthchr: OptionalType is covered by the `MetaType` check right above, which has the same impl.

v2/tools/generator/internal/config/group_configuration.go
```diff
@@ -125,8 +125,6 @@ func (gc *GroupConfiguration) findVersion(ref astmodel.PackageReference) *Versio
 		return gc.findVersion(r.Base())
 	case astmodel.LocalPackageReference:
 		return gc.findVersionForLocalPackageReference(r)
-	case astmodel.SubPackageReference:
-		return gc.findVersion(r.Parent())
```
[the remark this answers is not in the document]
matthchr: SubPackageReference impls `DerivedPackageReference`, which has the same impl (above), meaning this code can't ever be hit (I was impressed the linter noticed this)

v2/tools/generator/pkg/versions/versions.go
matthchr: There aren't any usages of it in our code I can find. Is someplace supposed to be using this?

Based on git blame you added this but I haven't traced if it was used at one point and we refactored things... can do that tomorrow.

v2/tools/generator/internal/config/group_configuration.go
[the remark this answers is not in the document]
theunrepentantgeek: Ok, that makes sense.

v2/tools/generator/pkg/versions/versions.go
theunrepentantgeek: If memory serves (it might not!) this is (was?) used to select the appropriate version of a resource to use when asoctl does an import.

v2/tools/generator/internal/config/group_configuration.go
```diff
@@ -210,21 +208,3 @@ func (gc *GroupConfiguration) UnmarshalYAML(value *yaml.Node) error {
 
 	return nil
 }
-
-// configuredVersions returns a sorted slice containing all the versions configured in this group
-func (gc *GroupConfiguration) configuredVersions() []string {
```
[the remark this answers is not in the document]
matthchr: History seems to be that this was added [here](https://github.com/Azure/azure-service-operator/commit/e9e311188d0cbee78b0d0cb6e22a27d15b8a594b#diff-0e39ae1050ef60ddd2fe7383bd101a4c3387d7226bbc2cef00a982158cd111e0R168) and called in `findVersion`. Then the logic seems to have been eventually refactored into `findVersionForLocalPackageReference`, which was then changed [here](https://github.com/Azure/azure-service-operator/commit/e07bb554fca4d5cbff24ae725b3875282fdf08b5#diff-0e39ae1050ef60ddd2fe7383bd101a4c3387d7226bbc2cef00a982158cd111e0) to no-longer call that method. 

The PR that removed that call specifically says:
"Remove NotConfiguredError" and "Remove error return from find*() methods", which I think means that this change was intentional and we're now doing error reporting a different way?

You wrote those PRs though, can you confirm the above?

v2/tools/generator/pkg/versions/versions.go
matthchr: It was used in this PR: https://github.com/Azure/azure-service-operator/pull/2816/files

In `v2/tools/asoctl/internal/importing/importable_resource.go` in the `selectLatestVersion` function. That's been refactored now and the logic seems to be in https://github.com/Azure/azure-service-operator/blob/main/v2/cmd/asoctl/internal/importing/importable_resource.go#L74, which doesn't call these helpers anymore.

So I think this is safe to delete?

v2/tools/generator/internal/config/group_configuration.go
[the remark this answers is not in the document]
theunrepentantgeek: Oh, I remember this now. The error return values from methods like `findVersionForLocalPackageReference()` (and friends) were being discarded, in favour of more informative errors constructed further up the call chain.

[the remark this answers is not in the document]
theunrepentantgeek: I don't remember details, but looks as though we moved the version decision into the code generator, so there should only ever be one candidate when `asoctl` does its thing.

v2/tools/generator/pkg/versions/versions.go
theunrepentantgeek: I don't remember details but looks as though we moved the version decision into the code generator, so there should only ever be one candidate when asoctl does its thing.

theunrepentantgeek (approved):

PR #4207, opened 2024-08-22
feat(helm): allow hardcodes to be configurable

t3mi:
Allows to override some hardcodes for deployment object.

Comparing with default values, these changes bring nothing except one default.
```diff
...
---
# Source: azure-service-operator/templates/apps_v1_deployment_azureserviceoperator-controller-manager.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/name: azure-service-operator
    app.kubernetes.io/version: v2.8.0
    control-plane: controller-manager
  name: azureserviceoperator-controller-manager
  namespace: default
spec:
  replicas: 1
+ revisionHistoryLimit: 10
  selector:
    matchLabels:
      control-plane: controller-manager
  template:
    metadata:
      annotations:
        kubectl.kubernetes.io/default-container: manager
      labels:
-       aadpodidbinding: aso-manager-binding
        app.kubernetes.io/name: azure-service-operator
        app.kubernetes.io/version: v2.8.0
        control-plane: controller-manager
...
```

gif

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/v2/charts/azure-service-operator/templates/apps_v1_deployment_azureserviceoperator-controller-manager.yaml
+++ b/v2/charts/azure-service-operator/templates/apps_v1_deployment_azureserviceoperator-controller-manager.yaml
@@ -1,6 +1,10 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
+  {{- with .Values.deploymentAnnotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
   labels:
     app.kubernetes.io/name: azure-service-operator
     app.kubernetes.io/version: v2.9.0
@@ -9,23 +13,27 @@ metadata:
   namespace: {{ .Release.Namespace }}
 spec:
   replicas: 1
+  revisionHistoryLimit: {{ .Values.revisionHistoryLimit }}
   selector:
     matchLabels:
       control-plane: controller-manager
   template:
     metadata:
       annotations:
-        {{- if .Values.podAnnotations }}
-        {{ toYaml .Values.podAnnotations }}
+        {{- with .Values.podAnnotations }}
+        {{- toYaml . | nindent 8}}
         {{- end }}
         kubectl.kubernetes.io/default-container: manager
       labels:
-        {{- if or (eq .Values.multitenant.enable false) (eq .Values.azureOperatorMode "watchers") }}
+        {{- if .Values.aadPodIdentity.enable }}
         aadpodidbinding: aso-manager-binding
         {{- end }}
         app.kubernetes.io/name: azure-service-operator
         app.kubernetes.io/version: v2.9.0
         control-plane: controller-manager
+        {{- with .Values.podLabels }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
     spec:
       {{- with .Values.tolerations }}
       tolerations:
@@ -191,16 +199,14 @@ spec:
             path: /readyz
             port: 8081
           initialDelaySeconds: 60
+        {{- with .Values.resources }}
         resources:
-          limits:
-            cpu: 500m
-            memory: 512Mi
-          requests:
-            cpu: 200m
-            memory: 256Mi
+          {{- toYaml . | nindent 10 }}
+        {{- end }}
+        {{- with .Values.securityContext }}
         securityContext:
-          allowPrivilegeEscalation: false
-          readOnlyRootFilesystem: true
+          {{- toYaml . | nindent 10 }}
+        {{- end }}
         volumeMounts:
         - mountPath: /var/run/secrets/tokens
           name: azure-identity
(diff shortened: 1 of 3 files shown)


v2/charts/azure-service-operator/templates/apps_v1_deployment_azureserviceoperator-controller-manager.yaml
```diff
@@ -191,16 +199,14 @@ spec:
...
         resources:
-          limits:
-            cpu: 500m
-            memory: 512Mi
-          requests:
-            cpu: 200m
-            memory: 256Mi
+          {{- toYaml . | nindent 10 }}
+        {{- end }}
+        {{- with .Values.securityContext }}
```
matthchr, line 206: There's an issue related to this here: https://github.com/Azure/azure-service-operator/issues/4214

Can you expand what options you'd like to configure here? I'm somewhat of the opinion that we should be setting the most restrictive options we can here (I know we are not now), but I suppose opening it to allow users more control, as long as they're making things even more restrictive, makes some sense and is future-proofing in case we somehow miss something here.

v2/charts/azure-service-operator/values.yaml
```diff
@@ -155,6 +155,44 @@ networkPolicies:
...
+  requests:
+    cpu: 200m
+    memory: 256Mi
+
+# Number of old history to retain to allow rollback
+# Default Kubernetes value is set to 10
+revisionHistoryLimit: 10
+
+# Specify security settings for a Container
+# Ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container
```
matthchr, line 191: I think there should be a note here that it is _strongly_ recommended you do not increase this pods permissions, and also a call to action to raise an issue if you believe things could be more locked down than they are here.

Then obviously we should lock things down more as we're lacking some defaults like:
```
runAsNonRoot: true
```
(I can do this in a followup PR).

Maybe something like: 
```suggestion
# Ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container
# It is not recommended to reduce the restrictions in this list, but additional restrictions outside of the default set 
# can be applied. If you believe additional securityContext configuration should be specified by default
# please raise an issue.
```

v2/charts/azure-service-operator/templates/apps_v1_deployment_azureserviceoperator-controller-manager.yaml
t3mi: Sure, we're using options below - dropping all capabilities at container level and non root user ID at pod level probably due to presence of rbac proxy container in previous releases):
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  ...
  name: azureserviceoperator-controller-manager
spec:
  ...
  template:
    ...
    spec:
      ...
      containers:
        - name: manager
          ...
          securityContext:
            capabilities:
              drop:
                - ALL
            readOnlyRootFilesystem: true
            allowPrivilegeEscalation: false
      ...
      securityContext:
        runAsUser: 10000
        runAsNonRoot: true
```

v2/charts/azure-service-operator/values.yaml
```diff
@@ -155,6 +155,44 @@ networkPolicies:
...
+# adjust them as necessary
+resources:
+  limits:
+    cpu: 500m
+    memory: 512Mi
+  requests:
+    cpu: 200m
+    memory: 256Mi
+
+# Number of old history to retain to allow rollback
```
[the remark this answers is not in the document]
t3mi: This was misplaced in the wrong PR, wasn't it?

v2/charts/azure-service-operator/values.yaml
t3mi: I can add the values from the upper comment if you'd agree with that.

v2/charts/azure-service-operator/values.yaml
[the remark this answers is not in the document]
matthchr: No, it's just a similar comment to one I made on a different PR.

This specific value (revisionHistoryLimit) applied to the Job that was part of the pre-upgrade check.

That job is gone now which means this value doesn't do anything and should be removed. (specifically just the `revisionHistoryLimit` - rest is fine)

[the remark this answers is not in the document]
matthchr: Wait I misread where this was going. Ignore me!

v2/charts/azure-service-operator/templates/apps_v1_deployment_azureserviceoperator-controller-manager.yaml
matthchr: We should probably make some of these the default, but will leave it for another PR.

matthchr (approved):

matthchr (approved):

PR #4208, opened 2024-08-22
feat(helm): allow upgrade checker to be disabled

### LINKED ISSUE #3436: Remove pre-upgrade check in controller
This is preventing beta-> v2.1.0 upgrades.

We should consider removing this at some point.

t3mi:
Potentially closes #3436

The purpose of the pre-upgrade checker was to block upgrades between beta -> stable releases but in the same time it blocks upgrades between stable releases. Adding such toggle allows user to opt-out from such behaviour until it would be fully removed.

gif

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/v2/charts/azure-service-operator/templates/pre-release-hook/batch_v1_job_pre-upgrade-checker.yaml
+++ b/v2/charts/azure-service-operator/templates/pre-release-hook/batch_v1_job_pre-upgrade-checker.yaml
@@ -1,3 +1,4 @@
+{{- if not .Values.disablePreUpgradeCheck }}
 apiVersion: batch/v1
 kind: Job
 metadata:
@@ -25,3 +26,4 @@ spec:
           requests:
             cpu: 200m
             memory: 256Mi
+{{- end }}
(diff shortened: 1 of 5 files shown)


theunrepentantgeek:
Under what circumstances is the pre-upgrade-check causing upgrades to fail between stable versions?

I've just reread that code and it should not fire unless the old version in the cluster is a beta one. Is the check buggy?

I'm worried that bypassing a needed check will put the cluster into an unstable, potentially unusable, state.

matthchr:
We just don't need the check anymore (as per #3436). I think I'd rather just rip it all out though, rather than add a Helm option to disable it conditionally.

I've sent a PR to do that here: https://github.com/Azure/azure-service-operator/pull/4217

matthchr:
My PR removing the upgrade check entirely has merged, so I think it makes sense to close this.

Let me know if you disagree @t3mi

t3mi:
@theunrepentantgeek we were installing the chart using terraform and for some reason it was failing at the pre-upgrade check. Unfortunately, I haven't saved the logs as we decided to use forked version of the chart w/o the check for future upgrades.

superseded by #4217

PR #4211, opened 2024-08-22
feat(helm): allow to use existing service account

t3mi:
Allows to use existing service account and/or control the name of the created service account so that prerequisites (like federated identity setup) could be successfully passed during deployment without hardcoding service account name used in the chart.

gif

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/v2/charts/azure-service-operator/templates/_helpers.tpl
+++ b/v2/charts/azure-service-operator/templates/_helpers.tpl
@@ -1,0 +1,10 @@
+{{/*
+Create the name of the service account to use
+*/}}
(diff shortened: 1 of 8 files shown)


matthchr (approved):

matthchr:
I can't tell if the test that's failing is just flaky today for some reason or if there's something else going on. As far as I can tell there's nothing in this PR that would've changed this test though, and it's passing locally for me...

matthchr:
Oh wait I am silly I retried it the wrong way... trying again.

PR #4222, opened 2024-08-28
Upgrade controller logging

### LINKED ISSUE #1620: Feature: Allow configuration of json log format
Currently all logs are written as plain text, e.g. 

`2021-07-05T01:54:16.492Z        INFO    controller-runtime.controller   Starting Controller     {"controller": "postgresqldatabase"}`

As logs will be processed most likely by a log processor like fluent bit and will be parsed and stored in a tool like Elasticsearch, it would be good to be able to log in json format. This makes it easier to integrate the logs in such a logging stack.

### LINKED ISSUE #3529: Log stacktrace on errors
We dont log stacktrace

**Describe the improvement**
We could, if we did something like this: https://pkg.go.dev/github.com/pkg/errors#hdr-Retrieving_the_stack_trace_of_an_error_or_wrapper

### LINKED ISSUE #4191: Replace deprecated klogr with textlogger

The package `k8s.io/klog/v2/klogr` is deprecated. 

**Describe the improvement**

The `klogr` package has documentation indicating that users should switch to using `k8s.io/klog/v2/textlogger` instead.

**Additional context**

This should be a relatively lightweight change, given we already use the lightweight `logr.Logger` struct in most places.

theunrepentantgeek:
# What this PR does / why we need it

Refactors the startup of our controller to separate log initialization from app initialization. 
Removes use of the deprecated `klogr` library in favour of `textlogger`.
Introduces the new commandline flag `--json-logging` which switches the output in the obvious way.

# Special notes for your reviewer

The package `k8s.io/klog/v2/klogr` is deprecated in favour of `k8s.io/klog/v2/textlogger` so we need to switch to that.

Closes #4191 

Adding JSON logging as an option (by specifying `--json-logging` on the command line) enables better integration with structured logging tools.

Closes #1620 

By using `zap` for JSON formatted logging, we get stack traces included in the logs by default.

Closes #3529 

# Testing

Tested manually on the command line.

# How does this PR make you feel
gif

### DIFF
--- a/v2/cmd/controller/app/flags.go
+++ b/v2/cmd/controller/app/flags.go
@@ -37,7 +37,6 @@ func (f Flags) String() string {
 }
 
 func InitFlags(flagSet *flag.FlagSet) Flags {
-
 	// default here for 'MetricsAddr' is set to "0", which sets metrics to be disabled if 'metrics-addr' flag is omitted.
 	metricsAddr := flagSet.String("metrics-addr", "0", "The address the metric endpoint binds to.")
 	secureMetrics := flagSet.Bool("secure-metrics", true, "Enable secure metrics. This secures the pprof and metrics endpoints via Kubernetes RBAC and HTTPS")
--- a/v2/cmd/controller/app/flags.go
+++ b/v2/cmd/controller/app/flags.go
@@ -11,57 +11,46 @@ import (
 )
 
 type Flags struct {
-	MetricsAddr          *string
-	ProfilingMetrics     *bool
-	SecureMetrics        *bool
-	HealthAddr           *string
-	WebhookPort          *int
-	WebhookCertDir       *string
-	EnableLeaderElection *bool
-	CRDManagementMode    *string
-	CRDPatterns          *string // This is a ';' delimited string containing a collection of patterns
+	MetricsAddr          string
+	ProfilingMetrics     bool
+	SecureMetrics        bool
+	HealthAddr           string
+	WebhookPort          int
+	WebhookCertDir       string
+	EnableLeaderElection bool
+	CRDManagementMode    string
+	CRDPatterns          string // This is a ';' delimited string containing a collection of patterns
 }
 
 func (f Flags) String() string {
 	return fmt.Sprintf(
 		"MetricsAddr: %s, SecureMetrics: %t, ProfilingMetrics: %t, HealthAddr: %s, WebhookPort: %d, WebhookCertDir: %s, EnableLeaderElection: %t, CRDManagementMode: %s, CRDPatterns: %s",
-		*f.MetricsAddr,
-		*f.SecureMetrics,
-		*f.ProfilingMetrics,
-		*f.HealthAddr,
-		*f.WebhookPort,
-		*f.WebhookCertDir,
-		*f.EnableLeaderElection,
-		*f.CRDManagementMode,
-		*f.CRDPatterns)
+		f.MetricsAddr,
+		f.SecureMetrics,
+		f.ProfilingMetrics,
+		f.HealthAddr,
+		f.WebhookPort,
+		f.WebhookCertDir,
+		f.EnableLeaderElection,
+		f.CRDManagementMode,
+		f.CRDPatterns)
 }
 
-func InitFlags(flagSet *flag.FlagSet) Flags {
+func InitFlags(flagSet *flag.FlagSet) *Flags {
+	result := &Flags{}
+
 	// default here for 'MetricsAddr' is set to "0", which sets metrics to be disabled if 'metrics-addr' flag is omitted.
-	metricsAddr := flagSet.String("metrics-addr", "0", "The address the metric endpoint binds to.")
-	secureMetrics := flagSet.Bool("secure-metrics", true, "Enable secure metrics. This secures the pprof and metrics endpoints via Kubernetes RBAC and HTTPS")
-	profilingMetrics := flagSet.Bool("profiling-metrics", false, "Enable pprof metrics, only enabled in conjunction with secure-metrics. This will enable serving pprof metrics endpoints")
+	flagSet.StringVar(&result.MetricsAddr, "metrics-addr", "0", "The address the metric endpoint binds to.")
+	flagSet.BoolVar(&result.SecureMetrics, "secure-metrics", true, "Enable secure metrics. This secures the pprof and metrics endpoints via Kubernetes RBAC and HTTPS")
+	flagSet.BoolVar(&result.ProfilingMetrics, "profiling-metrics", false, "Enable pprof metrics, only enabled in conjunction with secure-metrics. This will enable serving pprof metrics endpoints")
+	flagSet.StringVar(&result.HealthAddr, "health-addr", "", "The address the healthz endpoint binds to.")
+	flagSet.IntVar(&result.WebhookPort, "webhook-port", 9443, "The port the webhook endpoint binds to.")
+	flagSet.StringVar(&result.WebhookCertDir, "webhook-cert-dir", "", "The directory the webhook server's certs are stored.")
+	flagSet.BoolVar(&result.EnableLeaderElection, "enable-leader-election", false, "Enable leader election for controllers manager. Enabling this will ensure there is only one active controllers manager.")
 
-	healthAddr := flagSet.String("health-addr", "", "The address the healthz endpoint binds to.")
-	webhookPort := flagSet.Int("webhook-port", 9443, "The port the webhook endpoint binds to.")
-	webhookCertDir := flagSet.String("webhook-cert-dir", "", "The directory the webhook server's certs are stored.")
-
-	enableLeaderElection := flagSet.Bool("enable-leader-election", false, "Enable leader election for controllers manager. Enabling this will ensure there is only one active controllers manager.")
-
-	crdManagementMode := flagSet.String("crd-management", "auto",
+	flagSet.StringVar(&result.CRDManagementMode, "crd-management", "auto",
 		"Instructs the operator on how it should manage the Custom Resource Definitions. One of 'auto', 'none'")
-	crdPatterns := flagSet.String("crd-pattern", "", "Install these CRDs. CRDs already in the cluster will also always be upgraded.")
+	flagSet.StringVar(&result.CRDPatterns, "crd-pattern", "", "Install these CRDs. CRDs already in the cluster will also always be upgraded.")
 
-	return Flags{
-		MetricsAddr:      metricsAddr,
-		SecureMetrics:    secureMetrics,
-		ProfilingMetrics: profilingMetrics,
-
-		HealthAddr:           healthAddr,
-		WebhookPort:          webhookPort,
-		WebhookCertDir:       webhookCertDir,
-		EnableLeaderElection: enableLeaderElection,
-		CRDManagementMode:    crdManagementMode,
-		CRDPatterns:          crdPatterns,
-	}
+	return result
 }
--- a/v2/cmd/controller/app/flags.go
+++ b/v2/cmd/controller/app/flags.go
@@ -8,77 +8,61 @@ package app
 import (
 	"flag"
 	"fmt"
-
-	"k8s.io/klog/v2"
-
-	"github.com/Azure/azure-service-operator/v2/internal/version"
 )
 
 type Flags struct {
-	MetricsAddr          string
-	ProfilingMetrics     bool
-	SecureMetrics        bool
-	HealthAddr           string
-	WebhookPort          int
-	WebhookCertDir       string
-	EnableLeaderElection bool
-	CRDManagementMode    string
-	CRDPatterns          string // This is a ';' delimited string containing a collection of patterns
+	MetricsAddr          *string
+	ProfilingMetrics     *bool
+	SecureMetrics        *bool
+	HealthAddr           *string
+	WebhookPort          *int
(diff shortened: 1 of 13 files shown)


v2/cmd/controller/app/flags.go
```diff
@@ -8,83 +8,66 @@ package app
...
-		"Enable leader election for controllers manager. Enabling this will ensure there is only one active controllers manager.")
-	flagSet.StringVar(&crdManagementMode, "crd-management", "auto",
+	enableLeaderElection := flagSet.Bool("enable-leader-election", false, "Enable leader election for controllers manager. Enabling this will ensure there is only one active controllers manager.")
+
+	crdManagementMode := flagSet.String("crd-management", "auto",
 		"Instructs the operator on how it should manage the Custom Resource Definitions. One of 'auto', 'none'")
-	flagSet.StringVar(&crdPatterns, "crd-pattern", "", "Install these CRDs. CRDs already in the cluster will also always be upgraded.")
-	flagSet.BoolVar(&preUpgradeCheck, "pre-upgrade-check", false,
+	crdPatterns := flagSet.String("crd-pattern", "", "Install these CRDs. CRDs already in the cluster will also always be upgraded.")
+	preUpgradeCheck := flagSet.Bool("pre-upgrade-check", false,
```
theunrepentantgeek, line 57: I'm expecting a conflic there once #4217 merges, but that should be easy to resolve.

v2/cmd/controller/logging/logging.go
```diff
@@ -0,0 +1,87 @@
...
+	verbose *bool
+
+	// useJson indicates whether we should output logs in JSON format.
+	// Default is no
+	useJson *bool
+)
+
+// Create returns a new logger, ready for use.
+// This can be called multiple times if required.
+func Create() logr.Logger {
```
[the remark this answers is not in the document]
matthchr: This would also more closely mirror what both `TextLogger` and `zap` seem to do, which is take their configuration at the instance level rather than the package level.

v2/internal/genericarmclient/suite_test.go
```diff
@@ -33,7 +34,9 @@ func setup() error {
 	format.TruncateThreshold = 4000 // Force a longer truncate threshold
 
 	// setup global logger for controller-runtime:
-	ctrl.SetLogger(klogr.New()) //nolint: staticcheck
+	cfg := textlogger.NewConfig(textlogger.Verbosity(3)) // Use verbose logging in tests
```
matthchr, line 35: Should we be using our predefined debug log-levels here rather than hardcoding 3?

v2/internal/testcommon/kube_test_context_envtest.go
```diff
@@ -88,10 +87,13 @@ func createSharedEnvTest(cfg testConfig, namespaceResources *namespaceResources)
 		Scheme: scheme,
 	}
 
-	// TODO: Switch to klogr.New() below the below if we want controller-runtime logs in the tests.
+	// Switch logger below if we want controller-runtime logs in the tests.
 	// By default we've disabled controller runtime logs because they're very verbose and usually not useful.
-	// ctrl.SetLogger(klogr.New())
-	ctrl.SetLogger(logr.New(ctrllog.NullLogSink{}))
+	// import (ctrl "sigs.k8s.io/controller-runtime")
+	// cfg := textlogger.NewConfig(textlogger.Verbosity(3)) // Use verbose logging in tests
```
matthchr, line 93: same q here about hardcoding 3 (though it's in a comment)

v2/test/multitenant/suite_test.go
```diff
@@ -37,7 +38,9 @@ func setup() error {
 	format.TruncateThreshold = 4000 // Force a longer truncate threshold
 
 	// setup global logger for controller-runtime:
-	ctrl.SetLogger(klogr.New()) //nolint: staticcheck
+	cfg := textlogger.NewConfig(textlogger.Verbosity(3)) // Use verbose logging in tests
```
matthchr, line 38: Same q here (not repeating it everywhere but in a few spots)

v2/cmd/controller/app/flags.go
theunrepentantgeek: Resolved.

v2/cmd/controller/app/flags.go
```diff
@@ -8,77 +8,61 @@ package app
...
-	MetricsAddr          string
-	ProfilingMetrics     bool
-	SecureMetrics        bool
-	HealthAddr           string
-	WebhookPort          int
-	WebhookCertDir       string
-	EnableLeaderElection bool
-	CRDManagementMode    string
-	CRDPatterns          string // This is a ';' delimited string containing a collection of patterns
+	MetricsAddr          *string
```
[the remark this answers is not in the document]
theunrepentantgeek: It's to do with the change from `ParseFlags` to `InitFlags`, below. Previously, we set up the flagset and did the parsing all in one go, but now we need to divide the process into two parts, allowing for other flags to be added before parsing is done.

The various methods like `flagSet.String()` and `flagSet.Bool()` return a pointer to a variable that will be updated when the parsing is done. So the `*string` is not an "optional string", it is a "reference to a string".

Plus, they'll always be populated, so there are no guard clauses needed.

All that said, if there's a better solution I'm all keen because I agree this seems a bit hacky (and not in a good way).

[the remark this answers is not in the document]
matthchr: Doesn't `flagSet` offer the exact same API surface as `flagset` did?

Can we not just do this?

```
	var metricsAddr string

	// default here for 'MetricsAddr' is set to "0", which sets metrics to be disabled if 'metrics-addr' flag is omitted.
	flagSet.StringVar(&metricsAddr, "metrics-addr", "0", "The address the metric endpoint binds to.")
```

I tried that out locally and it seems to at least compile.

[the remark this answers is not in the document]
matthchr: In fact this is what we were doing before already - I guess I am not seeing why it no longer works?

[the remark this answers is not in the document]
theunrepentantgeek: Using a local like that only works if the call to `flagSet.Parse()` happens in the same method, which it used to but no longer does.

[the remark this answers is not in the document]
theunrepentantgeek: The variable isn't updated by the call to `flagSet.StringVar()` - it keeps the pointer, and updates the variable later, when `Parse` is called.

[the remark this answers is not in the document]
matthchr: Ahh I had totally missed that.

What about this approach?

```
func InitFlags(flagSet *flag.FlagSet) *Flags {
	result := &Flags{}
	
	// default here for 'MetricsAddr' is set to "0", which sets metrics to be disabled if 'metrics-addr' flag is omitted.
	flagSet.StringVar(&result.MetricsAddr, "metrics-addr", "0", "The address the metric endpoint binds to.")
	//...

	return result
}
```

Basically, rather than make the ptr-ness at each entry in the struct, we could make the whole Flags result a ptr and then use StringVar. It I think maintains the same behavior that you have where the flags aren't actually populated until parse is called, but it avoids needing N pointers and then we could either change where we pass flags to be `*Flags`, _or_ (once parse is called) we could just do something like:
```
flagsPtr := app.InitFlags(...)
...
flagSet.Parse()
...

flgs := *flagsPtr
```

and the rest of the code could be unchanged.

v2/cmd/controller/logging/logging.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/go-logr/zapr"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"k8s.io/klog/v2/textlogger"
+)
+
+var (
+	// verbose indicates whether we should use verbose logging.
+	// Default is no
+	verbose *bool
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

v2/cmd/controller/logging/logging.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// useJson indicates whether we should output logs in JSON format.
+	// Default is no
+	useJson *bool
```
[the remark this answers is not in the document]
theunrepentantgeek: Changed.

v2/cmd/controller/logging/logging.go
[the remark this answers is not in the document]
theunrepentantgeek: Done.

v2/cmd/controller/logging/logging.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+// Create returns a new logger, ready for use.
+// This can be called multiple times if required.
+func Create() logr.Logger {
+	if useJson != nil && *useJson {
```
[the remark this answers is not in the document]
theunrepentantgeek: Simplified by refactoring.

v2/cmd/controller/logging/logging.go
```diff
@@ -0,0 +1,87 @@
...
+
+		return log
+	}
+
+	return createTextLogger()
+}
+
+func createTextLogger() logr.Logger {
+	opts := []textlogger.ConfigOption{}
+	if verbose != nil && *verbose {
```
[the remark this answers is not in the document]
theunrepentantgeek: No longer required.

v2/internal/genericarmclient/suite_test.go
theunrepentantgeek: Done.

v2/cmd/controller/app/flags.go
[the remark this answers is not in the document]
matthchr: This is nicer.

v2/cmd/controller/logging/logging.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/go-logr/zapr"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"k8s.io/klog/v2/textlogger"
+)
+
+type Config struct {
+	// verbose indicates the level of logging.
+	// Higher values indicate more logging.
+	verbose int
```
matthchr, line 21:
```suggestion
	verbosity int
```

v2/cmd/controller/logging/logging.go
```diff
@@ -0,0 +1,99 @@
...
+	level := zap.InfoLevel
+	if cfg != nil {
+		switch cfg.verbose {
+		case 0:
+			level = zap.ErrorLevel
+		case 1:
+			level = zap.WarnLevel
+		case 2:
+			level = zap.InfoLevel
+		case 3:
```
matthchr, line 64: \>= 3 can (I think?) be configured? Probably these should all be zap.DebugLevel, since users may ask to loglevel 10 (at least I think that's as high as controller-runtime goes)

v2/cmd/controller/logging/logging.go
```diff
@@ -0,0 +1,99 @@
...
+
+	return zapr.NewLogger(logger), nil
+}
+
+// InitFlags initializes the flags for the logging package
+func InitFlags(fs *flag.FlagSet) *Config {
+	result := &Config{}
+
+	fs.IntVar(&result.verbose, "verbose", 2, "Enable verbose logging")
+	fs.IntVar(&result.verbose, "v", 2, "Enable verbose logging")
```
matthchr, line 94: Just confirming that this is set via `--v` right? (AFAIK that's what we was used before)

v2/cmd/controller/logging/logging.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	return zapr.NewLogger(logger), nil
+}
+
+// InitFlags initializes the flags for the logging package
+func InitFlags(fs *flag.FlagSet) *Config {
+	result := &Config{}
+
+	fs.IntVar(&result.verbose, "verbose", 2, "Enable verbose logging")
```
matthchr, line 93: should we use our logger consts here too?

matthchr (approved):
Looks good, had a few minor comments

v2/internal/controllers/samples_test.go
```diff
@@ -187,10 +189,11 @@ func findRefsAndCreateSecrets(tc *testcommon.KubePerTestContext, resources []cli
 }
 
 func getTestName(group string, version string) string {
+	caser := cases.Title(language.English)
```
matthchr, line 192: I think this change may have broken tests, as I see the samples tests for some reason think they need a SUBSCRIPTION_ID, which means they thought they were recording rather than replaying.  I think that can happen if the test didn't find the file it was looking for (which is based on testName presumably defined here)

v2/cmd/controller/logging/logging.go
theunrepentantgeek: Done.

v2/cmd/controller/logging/logging.go
theunrepentantgeek: Changed.

v2/cmd/controller/logging/logging.go
theunrepentantgeek: Yes, that's right.

v2/cmd/controller/logging/logging.go
theunrepentantgeek: Done.

v2/internal/controllers/samples_test.go
theunrepentantgeek: Fixed in PR #4236

PR #4242, opened 2024-09-03
Improve security context restrictions

### LINKED ISSUE #4214: Customize pod security context via helm chart
Currently there is no way to customize the operator's pod security context. We have ASO running in an AKS cluster with Azure Defender for Cloud enabled for that cluster. Defender gives recommendations about pods not running as a root user (by having the runAsUser property set to a non-zero value or by having the runAsNonRoot property set to true). We would like to enforce that recommendation but are currently unable to because we are unable to set either property on the ASO's pod.

**Describe the improvement**
We would love to have the possibility to customize the ASO's pod's security context via the helm chart, or alternatively to set specific flags via the helm chart (depending on what suits the projects philosophy best).

matthchr:
Fixes #4214

**If applicable**:
- [ ] this PR contains tests

### DIFF
--- a/v2/charts/azure-service-operator/values.yaml
+++ b/v2/charts/azure-service-operator/values.yaml
@@ -189,12 +189,18 @@ revisionHistoryLimit: 10
 
 # Specify security settings for a Container
 # Ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container
-# It is not recommended to reduce the restrictions in this list, but additional restrictions outside of the default set 
+# It is not recommended to reduce the restrictions in this list, but additional restrictions outside the default set
 # can be applied. If you believe additional securityContext configuration should be specified by default
 # please raise an issue.
 securityContext:
+  runAsUser: 65532 # nonroot user from gcr.io/distroless/static:nonroot image
+  runAsGroup: 65532 # nonroot group from gcr.io/distroless/static:nonroot image
+  runAsNonRoot: true
   allowPrivilegeEscalation: false
   readOnlyRootFilesystem: true
+  capabilities:
+    drop:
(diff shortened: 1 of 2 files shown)


v2/charts/azure-service-operator/values.yaml
```diff
@@ -189,12 +189,17 @@ revisionHistoryLimit: 10
 
 # Specify security settings for a Container
 # Ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container
-# It is not recommended to reduce the restrictions in this list, but additional restrictions outside of the default set 
+# It is not recommended to reduce the restrictions in this list, but additional restrictions outside the default set
 # can be applied. If you believe additional securityContext configuration should be specified by default
 # please raise an issue.
 securityContext:
+  runAsUser: 65536
+  runAsNonRoot: true
```
neil-119, line 197: Will you be adding a runAsGroup here as well? Seems to be the same value:

image

neil-119: See here: https://github.com/GoogleContainerTools/distroless/blob/main/common/variables.bzl#L18

theunrepentantgeek (approved):

v2/charts/azure-service-operator/values.yaml
matthchr: yes, will add runAsGroup as well, and good callout on the ID. I thought it was any ID which is nonroot which is technically true, but actually 65532 is the designated nonroot user for distroless images so we should be using that.

We already _are_ in the sense we set USER nonroot:nonroot in our dockerfile, but we should use the same values in the Kuberentes configuration too.

neil-119 (approved):

PR #4243, opened 2024-09-04
Allow reuse of import from asoctl

### LINKED ISSUE #3399: Allow programatic reuse of import functionality

theunrepentantgeek:

Moves the import azure-resource functionality from asoctl into a library where it can be programmatically reused.

Includes abstraction of the progress bar display to a new interface (`Progress`) along with provision of two implementations, one preserving the current dynamic progress bar display and the other simply logging messages.

Closes #3399 

Also fixes a couple race condition bugs that would cause problems in rare edge cases.

* If the process exited too quickly after processing all required imports, some or all of the summary report would be omitted
* If the import aborted with an error it was possible for a deadlock to occur, waiting on a progress bar that would never be completed

gif


v2/cmd/asoctl/pkg/importing/resource_importer.go
```diff
@@ -186,35 +182,35 @@ func (ri *ResourceImporter) importWorker(
...
 		rsrc := importResult.resource
 		gk := rsrc.GroupKind()
 
-		// Enqueue any child resources we found
-		progress <- progressDelta{total: len(importResult.pending)}
+		// Enqueue any child resources we found; these will be marked as completed when we collate their results
+		progress.AddPending(len(importResult.pending))
 		for _, p := range importResult.pending {
 			candidates <- p
 		}
```
matthchr, line 216: It sorta feels like in the programmatic context, if `report.AddFailedImport` is called (because of a failed import) that we should end up returning an `err`?

What do you think?

v2/cmd/asoctl/pkg/importing/progress.go
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package importing
```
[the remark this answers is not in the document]
theunrepentantgeek: I had the same conversation with myself.

v2/cmd/asoctl/pkg/importing/resource_importer.go
```diff
@@ -79,30 +72,29 @@ func (ri *ResourceImporter) AddARMID(armID string) error {
 // Partial results are returned even in the case of an error.
 func (ri *ResourceImporter) Import(
 	ctx context.Context,
+	done chan struct{},
 ) (*ResourceImportResult, error) {
-	workers := 1
+	workers := 4
```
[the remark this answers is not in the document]
theunrepentantgeek: Yeah, I'll make it configurable in a followup PR. Logged as #4251

v2/cmd/asoctl/pkg/importing/resource_importer.go
theunrepentantgeek: The report collates and deals with the errors, so they're not lost. This method is triggered as a goroutine, so it can't return anything; the result is "returned" via the `publish` channel.

v2/cmd/asoctl/pkg/progress/log_progress.go
```diff
@@ -0,0 +1,94 @@
...
+type logProgress struct {
+	updates chan progressDelta // Channel for receiving updates (avoids concurrent access to progress)
+	log     logr.Logger        // Logger for updating progress
+	done    chan struct{}      // Channel for knowing when we're done
+}
+
+var _ importing.Progress = &logProgress{}
+
+// NewLog creates a new progress reporter using a log to report progress.
+func NewLog(
```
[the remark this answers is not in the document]
theunrepentantgeek: I see what you mean. I'll rename the factory methods and add a `DiscardReporter` too.

v2/cmd/asoctl/pkg/importing/resource_importer.go
```diff
@@ -243,39 +239,36 @@ func (ri *ResourceImporter) collateResults(
...
 		// We do this after everything else because it might indicate we're finished
-		progress <- progressDelta{complete: 1}
+		progress.Completed(1)
 	}
 
-	report.WriteToLog(ri.log)
+	publish <- report
 }
 
 func (ri *ResourceImporter) ImportResource(
```
matthchr, line 253: Do we want this exported? Or is the intent that the user always should use the `Add` + `Import` combination, in which case this could be made private?

v2/cmd/asoctl/pkg/importing/resource_importer.go
matthchr: I understand that, my point is that the `Import` method (which is the callers entrypoint to all of this) gives them the ability to:
1. Configure a `Progress` (for progress reporting)
2. Configure the resources to import.

but I don't see how they can consume the errors of the import programmatically. They're not lost (because they're written to log), but as far as I can tell the caller doesn't actually get a stack of errors that they can examine in code?

I'm thinking for scenarios where the user is writing their own tool on top of this SDK - they may want to exit with a nonzero exit code when certain resources (say, storage account) are failed to import, but bypass other failures as expected in their environment. 

It's not _100%_ clear to me they can't already do this with the interface provided (maybe by implementing their own `Progress`?) but it's not amazingly clear from the code. I think we should either:
1. Include those details on the `ResourceImportResult`
2. Return an `err` from `Import` along with `ResourceImportResult`, in keeping with the "// Partial results are returned even in the case of an error."
3. Document how the caller can accomplish the above in the documentation, if it can already be easily accomplished.

The fact that there's parallelism under the hood here shouldn't really matter from the perspective of the `Import` method because that's the wrapper for it all, the fan-out and join happen within that method only so from the callers perspective, they made a call, waited a while and got a result, and I think that result should include the ability to programmatically examine errors.

(I probably put this comment in a confusing spot because it's not clear I was talking about `Import` - I think I was tracing through to see if they could already do this with the existing inputs which is what led me to here where we catalog errors, I realize now that's probably confusing though - hopefully the above clarifies.)

v2/cmd/asoctl/pkg/importing/progress.go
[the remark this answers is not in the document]
theunrepentantgeek: I'll give this some thought and try to find a better name for the package - at the moment its internal and doesn't really matter, but after this PR it will be part of the API so we should really try and get it right.

v2/cmd/asoctl/pkg/progress/log_progress.go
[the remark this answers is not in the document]
theunrepentantgeek: Done. Let me know what you think of the new naming.

v2/cmd/asoctl/pkg/importing/progress.go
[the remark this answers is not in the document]
theunrepentantgeek: I renamed the package to `resourceimporter`.

v2/cmd/asoctl/pkg/importing/resource_importer.go
theunrepentantgeek: Ahh, now I understand your point. Yes, extending `ResourceImportResult` is the best way to go - that currently includes the set of successfully imported resources, and we can add to those details of any errors to allow a consumer to make decisions. I'll do that.

v2/cmd/asoctl/cmd/import_azure_resource.go
```diff
@@ -129,9 +129,9 @@ func importAzureResource(
 	}
 
 	done := make(chan struct{}) // signal that we're done
-	pb := progress.NewProgressBar("Import Azure Resources", progressBar, done)
+	pb := importreporter.NewBar("Import Azure Resources", progressBar, done)
 
-	importer := importing.NewResourceImporter(api.CreateScheme(), client, log, pb)
+	importer := importresources.NewResourceImporter(api.CreateScheme(), client, log, pb)
```
matthchr, line 134: minor: we could just make this `importresources.New` if we wanted to avoid repeating ourselves? Or maybe `NewImporter` (which is a bit of repetition still but less than we had before)

v2/cmd/asoctl/pkg/importresources/resource_importer.go
```diff
@@ -73,28 +73,28 @@ func (ri *ResourceImporter) AddARMID(armID string) error {
 func (ri *ResourceImporter) Import(
 	ctx context.Context,
 	done chan struct{},
-) (*ResourceImportResult, error) {
+) (*Result, error) {
```
matthchr, line 76: This is much nicer from a type naming perspective.

matthchr (approved):
Approved - I assume you're planning to follow up with updates to the `Result` type to report errors to the user as well?

v2/cmd/asoctl/pkg/importing/resource_importer.go
theunrepentantgeek: Made this private.

v2/cmd/asoctl/cmd/import_azure_resource.go
theunrepentantgeek: Done.

v2/cmd/asoctl/pkg/importresources/resource_import_result.go
```diff
@@ -22,26 +22,26 @@ import (
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
 )
 
-// ResourceImportResult represents the result of an import operation
-type ResourceImportResult struct {
+// Result represents the result of an import operation
+type Result struct {
```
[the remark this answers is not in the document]
theunrepentantgeek: Done.

PR #4245, opened 2024-09-04
Add some validation of crd-pattern to asoctl export template

### LINKED ISSUE #3969: asoctl export template --crd-pattern validation
The `--crd-pattern` parameter is not validated by `asoctl` at all.

**Describe the improvement**
It would be nice to validate the pattern, but there are some gotchas...

**Additional context**
I got partway through doing this but realized it won't work all that well because asoctl isn't necessarily built with the same version as the target version.

For example let's say that the user is attempting to install the experimental release to test a new CRD they've just added, if they try with an older asoctl, we would reject them because their group doesn't exist. They'd need to download the experimental asoctl and use that.

The problem works the opposite way too, if the user is using a newer asoctl but installing an older release, we'd allow crdpatterns that won't actually work for the operator.

We could maybe solve this problem by doing our best in asoctl and then having a --novalidate flag or something that bypassed validation for crdpattern if the user really knows what they're doing... but really the crdpattern validation seems somewhat at odds with letting the user select a version that may not match the asoctl version.

theunrepentantgeek:

When using `asoctl export template` the values passed for `--crd-pattern` aren't currently checked at all, so they might not do what the user wants.

Closes #3969 

## Special notes for your reviewer

Validation is done against the CRDs known to that version of `asoctl`, which might be different to the `--version` specified on the command-line. Not sure how to handle this - we should discuss.

gif

### DIFF
--- a/v2/cmd/asoctl/cmd/export_template.go
+++ b/v2/cmd/asoctl/cmd/export_template.go
@@ -31,7 +31,7 @@ func newTemplateCommand() *cobra.Command {
 	var options templateOptions
 
 	cmd := &cobra.Command{
-		Use:   "template [--source <string> |--version <string>] [--crd-pattern <string>|--raw]",
+		Use:   "template [--source <string> | --version <string>] [--crd-pattern <string> | --raw]",
 		Short: "Template creates a YAML file from the specified ASO yaml template",
 		Example: `asoctl export template --version v2.6.0 --crd-pattern "resources.azure.com/*" --crd-pattern "containerservice.azure.com/*"
 
--- a/v2/cmd/asoctl/cmd/export_template.go
+++ b/v2/cmd/asoctl/cmd/export_template.go
@@ -32,7 +32,7 @@ type templateOptions struct {
 	crdPattern []string
 }
 
-var expectedVersionRegex = regexp.MustCompile(`^v\d\.\d?\d\.\d$`)
+var expectedVersionRegex = regexp.MustCompile(`^v\d\.\d+\.\d+$`)
 
 // newTemplateCommand creates a new cobra Command when invoked from the command line
 func newTemplateCommand() *cobra.Command {
--- a/v2/cmd/asoctl/cmd/export_template.go
+++ b/v2/cmd/asoctl/cmd/export_template.go
@@ -24,7 +24,7 @@ type templateOptions struct {
 	crdPattern []string
 }
 
-var expectedVersionRegex = regexp.MustCompile(`^v\d\.\d\.\d$`)
+var expectedVersionRegex = regexp.MustCompile(`^v\d\.\d?\d\.\d$`)
 
 // newTemplateCommand creates a new cobra Command when invoked from the command line
 func newTemplateCommand() *cobra.Command {
--- a/v2/cmd/asoctl/cmd/export_template.go
+++ b/v2/cmd/asoctl/cmd/export_template.go
@@ -6,8 +6,11 @@
 package cmd
 
 import (
+	"fmt"
 	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"os"
+	"reflect"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


v2/cmd/asoctl/cmd/export_template.go
```diff
@@ -6,14 +6,22 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"os"
+	"reflect"
 	"regexp"
 	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
+	. "github.com/Azure/azure-service-operator/v2/internal/logging"
```
matthchr, line 20: minor: I think this line can/should be in the 3rd stanza below (lines 22-24) as it's from aso

v2/cmd/asoctl/cmd/export_template.go
```diff
@@ -24,14 +32,14 @@ type templateOptions struct {
 	crdPattern []string
 }
 
-var expectedVersionRegex = regexp.MustCompile(`^v\d\.\d\.\d$`)
+var expectedVersionRegex = regexp.MustCompile(`^v\d\.\d?\d\.\d$`)
```
matthchr, line 35: Ooh good catch

matthchr: Though maybe we just want:
```suggestion
var expectedVersionRegex = regexp.MustCompile(`^v\d\.(\d)+\.(\d)+$`)
```

v2/cmd/asoctl/cmd/export_template.go
```diff
@@ -64,13 +72,20 @@ asoctl export template --version v2.6.0 --crd-pattern "resources.azure.com/*;con
 				uri = template.URIFromVersion(options.version)
 			}
 
+			log := CreateLogger()
```
matthchr, line 75: What stream of stderr/stdout does this go to?

If it goes to stdout we can't have it at all (because we need to be able to pipe and this will break piping), if it goes to stderr it's OK but maybe should be behind a `--debug` flag, with the command producing no output by default?

v2/cmd/asoctl/cmd/export_template.go
```diff
@@ -122,3 +137,45 @@ asoctl export template --version v2.6.0 --crd-pattern "resources.azure.com/*;con
...
+				log.V(Status).Info("No CRDs matched pattern", "pattern", pattern)
+				warnings++
+			} else {
+				log.V(Info).Info("Pattern matched CRDs", "pattern", pattern, "count", count)
+			}
+		}
+	}
+
+	if warnings > 0 {
+		log.V(Status).Info("Possibly ineffective CRD patterns detected")
```
matthchr, line 162: Same comment about us needing to be careful due to piping. Fine if this goes to stderr, let's just confirm that's the case?

v2/cmd/asoctl/cmd/export_template.go
theunrepentantgeek: Logging goes to `stderr`; this is the same logging used by `import azure-resource` which supports piping the results to a file even though it has lots of progress updates beforehand.

v2/cmd/asoctl/cmd/export_template.go
theunrepentantgeek: It's a `.` import, so I separated it for better visibility so readers know it's special.

v2/cmd/asoctl/cmd/export_template.go
theunrepentantgeek: Why do we need the `(` parenthesis `)`? 

I'm thinking we could simplify it to `^v\d\.\d+\.\d+$`.

matthchr: hah, we don't - good point

theunrepentantgeek: Updated regex.

v2/cmd/asoctl/cmd/export_template.go
theunrepentantgeek: All logs go to stderr.

PR #4257, opened 2024-09-12
Allow number of simultaneous workers to be configured

### LINKED ISSUE #4251: Make the number of workers used by `asoctl` configurable

As of PR #4243, the number of workers will be hard coded at 4. 

## Describe the improvement

Making this configurable - both on the commandline, and via the programatic interface, would be useful flexibility.

theunrepentantgeek:

Does what it says on the box: makes the number of concurrent workers configurable. 

More workers can make the import process complete more quickly - but not in all cases. 

The default of 4 seems to be a good compromise, ensuring a single slow resource doesn't stall the entire process too much. Making this configurable gives consumers more control when required.

Closes #4251 

gif

### DIFF
--- a/v2/cmd/asoctl/cmd/import_azure_resource.go
+++ b/v2/cmd/asoctl/cmd/import_azure_resource.go
@@ -103,6 +103,13 @@ https://docs.microsoft.com/azure/active-directory/develop/authentication-nationa
 		nil,
 		"Add the specified annotations to the imported resources. Multiple comma-separated annotations can be specified (--annotation example.com/myannotation=foo,example.com/myannotation2=bar) or the --annotation (-a) argument can be used multiple times (-a example.com/myannotation=foo -a example.com/myannotation2=bar)")
 
+	cmd.Flags().IntVarP(
+		&options.workers,
+		"workers",
+		"w",
+		4,
+		"Specify the number of parallel workers to use when importing resources")
+
 	return cmd
 }
 
@@ -136,6 +143,7 @@ func importAzureResource(
 	pb := importreporter.NewBar("Import Azure Resources", progressBar, done)
 
 	importerOptions := importresources.ResourceImporterOptions{
+		Workers: options.workers,
 	}
 
 	importer := importresources.New(api.CreateScheme(), client, log, pb, importerOptions)
@@ -217,6 +225,7 @@ type importAzureResourceOptions struct {
 	namespace    string
 	annotations  []string
 	labels       []string
+	workers      int
 
 	readCloud               sync.Once
 	azureAuthorityHost      string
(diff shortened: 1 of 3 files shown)


v2/cmd/asoctl/cmd/import_azure_resource.go
```diff
@@ -86,19 +88,28 @@ https://docs.microsoft.com/azure/active-directory/develop/authentication-nationa
...
 		"a",
 		nil,
 		"Add the specified annotations to the imported resources. Multiple comma-separated annotations can be specified (--annotation example.com/myannotation=foo,example.com/myannotation2=bar) or the --annotation (-a) argument can be used multiple times (-a example.com/myannotation=foo -a example.com/myannotation2=bar)")
 
+	cmd.Flags().IntVarP(
+		&options.workers,
+		"workers",
+		"w",
+		4,
+		"Specify the number of parallel workers to use when importing resources")
```
matthchr, line 111:
```suggestion
		"The number of parallel workers to use when importing resources")
```

(Specify is somewhat redundant when dealing with a cmdline parameter that accepts an input)

v2/cmd/asoctl/pkg/importresources/resource_importer.go
```diff
@@ -282,3 +291,12 @@ func (ri *ResourceImporter) importResource(
 
 	return result
 }
+
+// desiredWorkers returns the number of workers to use for importing resources.
+func (ri *ResourceImporter) desiredWorkers() int {
```
matthchr, line 296: minor: This method isn't hurting but my understanding is that it's not actually possible to take this path unless the user explicitly passes a negative or nonzero value for the cmdline option, because if the user doesn't specify anything the default of 4 is used at the cmdline flag level (thus not hitting the default of 4 here).

I'm wondering if it's better to add a `validate` helper to `importAzureResourceOptions`, and then call that at the start of `importAzureResource` and fail the cmd if the user passes a negative value? Could also be extended to check the provided labels for validity (length, allowed number of characters?).

Oh wait I see you already parse the labels it's just down below after the resources are imported. Maybe move the parsing of the labels up front, and validate input of workers too to fail fast on bad user input, then run import (and use the actual parsed labels if configured?)

matthchr (approved):
Approved, but have reservations about ignoring invalid user input for `--workers` flag. I think rejecting would be better.

v2/cmd/asoctl/cmd/import_azure_resource.go
theunrepentantgeek: Updated description.

v2/cmd/asoctl/pkg/importresources/resource_importer.go
theunrepentantgeek: The default will always be `4` if used from the command-line, but that's not the only way to get here. 
With the relocation under `pkg` we've allowed for reuse. The value here is an option provided in the `ResourceImporterOptions` struct, which might not be set by a caller.

PR #4262, opened 2024-09-13
Allow importing of FrontDoor resources

### LINKED ISSUE #4260: Bug: asoctl can't create manifests for Azure Front Door and CDN profile
**Version of asoctl**
asoctl version: v2.9.0 linux

**Describe the bug**
It does not appear to be possible to use asoctl on a resource where the location is 'global' such as Front Door using the
`asoctl import azure-resource` command

**To Reproduce**
Run the following supplying the Resource ID of a Front Door and CDN profile in Azure:
asoctl import azure-resource /subscriptions/xxx.../fd-resource-id

The following error is reported multiple times:
"No registered resource provider found for location 'global' and API version '2023-05-01' for type 'extensions'. The supported api-versions are '2020-07-01-preview, 2021-05-01-preview, 2021-09-01, 2022-03-01, 2022-04-02-preview, 2022-07-01, 2022-11-01, 2023-05-01, 2019-11-01-preview, 2020-10-01-preview, 2021-03-01'. The supported locations are 'eastus, westeurope, westcentralus, westus2, westus3, southcentralus, eastus2, northeurope, uksouth, southeastasia, australiaeast, francecentral, centralus, northcentralus, westus, koreacentral, eastasia, japaneast, canadaeast, canadacentral, norwayeast, germanywestcentral, italynorth, israelcentral, polandcentral, swedencentral, switzerlandnorth, australiasoutheast, centralindia, southindia, japanwest, ukwest, francesouth, koreasouth, southafricanorth, brazilsouth, uaenorth, norwaywest, germanynorth, jioindiawest, southafricawest'.\"

Manifest files are  only generated for RoleDefinitions.

**Expected behavior**
Manifests for all Front Door child resources should be generated including:
- AFD Endpoints/Routes
- Origin Groups/Origins
- Custom Domains
- Rules Sets/Rules
- Secrets
- Security Policies

theunrepentantgeek:

When scanning for child/extension resources in the `cdn` group, `asoctl` runs into a `NoRegisteredProviderFound` error which is incorrectly treated as a fatal error. 

Changing this to a known case, where we simply skip the check for that particular kind of child resource, results in the import working correctly.

Closes #4260 

gif

### DIFF
--- a/v2/cmd/asoctl/pkg/importresources/find_child_resources_test.go
+++ b/v2/cmd/asoctl/pkg/importresources/find_child_resources_test.go
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package importresources
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+
+	cdn "github.com/Azure/azure-service-operator/v2/api/cdn/v1api20230501"
+	dbforpostgresql "github.com/Azure/azure-service-operator/v2/api/dbforpostgresql/v1api20221201"
+)
+
+func Test_FindChildResources_GivenParentResource_ReturnsExpectedChildResources(t *testing.T) {
+	t.Parallel()
+
+	postgreSQLServer := &dbforpostgresql.FlexibleServer{}
+	postgreSQLConfiguration := &dbforpostgresql.FlexibleServersConfiguration{}
+
+	ruleset := &cdn.RuleSet{}
+	rule := &cdn.Rule{}
+
+	cases := map[string]struct {
+		parentResourceType        string
+		expectedChildResourceType string
+	}{
+		"PostgreSQL servers have configurations": {
+			parentResourceType:        postgreSQLServer.GetType(),
+			expectedChildResourceType: postgreSQLConfiguration.GetType(),
+		},
+		"CDN rulesets have rules": {
+			parentResourceType:        ruleset.GetType(),
+			expectedChildResourceType: rule.GetType(),
+		},
+	}
+
+	for name, c := range cases {
+		c := c
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			g := NewGomegaWithT(t)
+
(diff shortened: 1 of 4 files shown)


v2/cmd/asoctl/pkg/importresources/importable_arm_resource.go
```diff
@@ -371,7 +371,8 @@ func (*importableARMResource) classifyError(err error) (string, bool) {
 
 		if responseError.StatusCode == http.StatusBadRequest {
 			if strings.Contains(responseError.Error(), "RequestUrlInvalid") ||
-				strings.Contains(responseError.Error(), "ValidationFailed") {
+				strings.Contains(responseError.Error(), "ValidationFailed") ||
+				strings.Contains(responseError.Error(), "NoRegisteredProviderFound") {
```
matthchr, line 375: This error could also in theory mean that the user hasn't registered this resource provider in their subscription? I guess even in that case the best thing we can do is just continue.

matthchr (approved):

matthchr:
It sounds like the user that reported this tested it and it helped but there was some other issue with importing routes, may be worth looking at that too

v2/cmd/asoctl/pkg/importresources/find_child_resources_test.go
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package importresources
```
matthchr, line 6: minor: Could be `importresources_test`, I think?

Cleaner as it avoids the temptation to test internals (which is sometimes what you want to do but sometimes not desirable).

matthchr (approved):

v2/cmd/asoctl/pkg/importresources/find_child_resources_test.go
theunrepentantgeek: I'll move it, but really prefer to keep things close unless there's a problem (like a circular reference) that can be resolved by using the test package naming.

PR #4263, opened 2024-09-13
Built-in role definitions should be skiped by asoctl

theunrepentantgeek:

They can't be created by ASO because they'd conflict - they're built in, so already present.

`asoctl` shouldn't include built in role definitions in the output.

gif

### DIFF
--- a/v2/api/authorization/customizations/role_definition_extensions.go
+++ b/v2/api/authorization/customizations/role_definition_extensions.go
@@ -28,6 +28,9 @@ func (extension *RoleDefinitionExtension) Import(
 		return extensions.ImportResult{}, err
 	}
 
+	// If this cast doesn't compile, update the `api` import to reference the now latest
+	// stable version of the authorization group (this will happen when we import a new
+	// API version in the generator.)
 	if definition, ok := rsrc.(*api.RoleDefinition); ok {
 		// If this role definition is built in, we don't need to export it
 		if definition.Spec.Type != nil {
--- a/v2/api/authorization/customizations/role_definition_extensions.go
+++ b/v2/api/authorization/customizations/role_definition_extensions.go
@@ -1,0 +1,41 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+package customizations
+
+import (
+	"context"
+	"strings"
+
+	api "github.com/Azure/azure-service-operator/v2/api/authorization/v1api20220401"
+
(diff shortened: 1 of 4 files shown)


v2/api/authorization/customizations/role_definition_extensions.go
```diff
@@ -0,0 +1,41 @@
...
+Licensed under the MIT license.
+*/
+
+package customizations
+
+import (
+	"context"
+	"strings"
+
+	api "github.com/Azure/azure-service-operator/v2/api/authorization/v1api20220401"
```
matthchr, line 12: Does this need to be the storage version? (I checked other `Import` impls and it seems it doesn't, I guess this is because we call import on the latest public API shape not the latest storage shape which makes some sense...)

Assuming above is true and this doesn't need to be the storage version, how do we catch when we update to support a newer API version and this cast is no longer correct? We can do that statically by asserting the type impls the `Hub` interface, but there's no such static assertion here so the code will compile even though it doesn't work.

Do we have a test that covers this case?

Wondering if at they very least we should do something like:
```
// If you change this, make sure to also change the api import reference.
storage "github.com/Azure/azure-service-operator/v2/api/authorization/v1api20220401/storage"

...


// Type assert that we are the hub type. This will fail to compile if
// the hub type has been changed but this extension has not
var _ conversion.Hub = &storage.RoleDefinition{}
```

matthchr: (I think the other imports could use this check too, or one like it - we could fabricate an interface on the latest API version of each resource and then avoid the weird indirection through the storage version for this if we prefer, but in some ways that feels even more awkward)

theunrepentantgeek: The interface is only ever generated for a single version of any one resource - usually the latest stable version, though a preview version will be used if no stable version is available.

If we import a new stable version of a resource and that interface moves, we get a compilation error because the case from `genruntime.ImportableResource` can never succeed. It will look something like this:

``` bash
# github.com/Azure/azure-service-operator/v2/api/authorization/customizations
..\..\api\authorization\customizations\role_assignment_extensions.go:31:23: impossible type assertion: rsrc.(*api.RoleAssignment)
        *v1api20200801preview.RoleAssignment does not implement genruntime.ImportableResource (missing method InitializeSpec)
```

To provoke this, I changed the import from `authorization/v1api20220401` (which works) to `authorization/v1api20200801preview`.

So while we don't have an explicit type assertion, we still get the same level of protection.

matthchr (approved):

PR #4270, opened 2024-09-16
Fix deadlock and ux bugs in asoctl

theunrepentantgeek:

Fixes a deadlock bug that occurs if an error is returned before the first progressbar is created.

Fixes a UX bug where completed progress bars are left on screen as debris.

gif

### DIFF
--- a/v2/cmd/asoctl/cmd/import_azure_resource.go
+++ b/v2/cmd/asoctl/cmd/import_azure_resource.go
@@ -7,12 +7,15 @@ package cmd
 
 import (
 	"context"
+	"github.com/go-logr/logr"
+	"io"
+	"os"
+	"sync"
+
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
-	"os"
-	"sync"
 
 	internalconfig "github.com/Azure/azure-service-operator/v2/internal/config"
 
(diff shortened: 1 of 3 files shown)


v2/cmd/asoctl/cmd/import_azure_resource.go
```diff
@@ -7,13 +7,12 @@ package cmd
...
 import (
 	"context"
-	"os"
-	"sync"
-
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"os"
```
matthchr, line 14: minor: avoid reordering these? stdlib imports in first group makes sense IMO.

v2/cmd/asoctl/cmd/import_azure_resource.go
```diff
@@ -128,6 +122,9 @@ func importAzureResource(
 		return errors.Wrapf(err, "failed to create ARM client")
 	}
 
+	//
```
matthchr, line 125: Missing comment?

v2/cmd/asoctl/cmd/import_azure_resource.go
```diff
@@ -139,6 +136,9 @@ func importAzureResource(
 		}
 	}
 
+	// Make sure all output is written when we're done
```
matthchr, line 139: minor: maybe leave a note that this shouldn't be `defer`ed until right before `Import` is called, because the `Import` call is what actually uses/writes to the progress bar and calls to `Wait` will deadlock if a bar was added but never written to?

Though... if that's actually the case, is it also possible to deadlock if `armIDs` is empty? We'd not take the for loop at all but `importreporter.NewBar` would still be created which is what actually triggers the deadlock?

matthchr: Is it possible to test this case somehow?

theunrepentantgeek: As long as one bar is created, nothing deadlocks. The call to `importer.Import()` will do that, so we should be safe. I'll refactor the code to tidy things up and add some docs.

v2/cmd/asoctl/cmd/import_azure_resource.go
theunrepentantgeek: Removed.

v2/cmd/asoctl/cmd/import_azure_resource.go
```diff
@@ -139,6 +136,11 @@ func importAzureResource(
 		}
 	}
 
+	// Make sure all output is written when we're done.
+	// This defer has to be immediately before the call the importer.Import();
```
super-harsh, line 140:
```suggestion
	// This defer has to be immediately before we call the importer.Import();
```

super-harsh (approved):

PR #4279, opened 2024-09-18
Reduce noise when importing PostgreSQL Flexible Server Configurations

### LINKED ISSUE #4278: Reducing noise when importing PostgreSQL Flexible Server

Pulled out of a comment on #4153:

> There is one more thing I want to ask: is it possible to make asoctl allow users to specify certain configuration items for import? I'm not sure if this would make sense for most users. When we tested importing an Azure PostgreSQL Flexible Server, the generated YAML file was about 15,000 lines long, which included all configurations and role definitions. It is quite tedious to modify or manage this file. In our case, we actually only need a few important parameters to be managed through ASO, such as max_connections and other similar settings.

**Describe the improvement**

Consider whether any of the child resources imported under a PostgreSQL Flexible Server could be suppressed - things like default values, etc, where there is no value in having them included.

theunrepentantgeek:
## What this PR does

When importing a PostgreSQL Flexible Server, `asoctl` was including more than 400 configuration options, many of which couldn't be changed by the user.

With this PR, we now
* Skip configurations that haven't been changed from their default
* Skip configurations marked readonly that the user cannot modify

Closes #4278 

### How does this PR make you feel

gif

### DIFF
--- a/v2/api/dbforpostgresql/customizations/flexible_servers_configuration_extensions.go
+++ b/v2/api/dbforpostgresql/customizations/flexible_servers_configuration_extensions.go
@@ -42,6 +42,13 @@ func (extension *FlexibleServersConfigurationExtension) Import(
 			*config.Status.IsReadOnly {
 			return extensions.ImportSkipped("readonly configuration can't be set"), nil
 		}
+
+		// Skip default values
+		if config.Status.DefaultValue != nil &&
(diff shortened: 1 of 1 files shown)


matthchr (approved):

matthchr:
Should check MySQL too, it may have similar behavior

PR #4237, opened 2024-09-02
Add support for 20240401 AML resources

super-harsh:
**What this PR does / why we need it**:

This PR includes latest version of AML resources. Consider this as a starting point to support further AML resources in next iteration.

**Special notes for your reviewer**:

This PR includes breaking changes for `WorkspacesCompute` 20210701 API version below: 
- SslConfiguration.Key - marked as secret
- SslConfiguration.Cert - marked as secret
- VirtualMachineSshCredentials.PrivateKeyData - marked as secret
- VirtualMachineSshCredentials.PublicKeyData - marked as secret
- DatabricksProperties.DatabricksAccessToken - marked as secret

**If applicable**:
- [ ] this PR contains documentation
- [x] this PR contains tests
- [x] this PR contains YAML Samples

PR #4289, opened 2024-09-22
Improve secret detection heuristics

### LINKED ISSUE #4269: Include "Token" in the generator secret check heuristic

Currently, our code generator's `mightBeSecretProperty` check heuristic doesn't include "Token" properties to be checked.

**Describe the improvement**
Should include "Token" into the secret check heuristic

### LINKED ISSUE #4369: Feature: Support for creating Smart Detector Alert Rules through ASO
This is a request to add [Smart Detector Alert Rules](https://learn.microsoft.com/en-us/rest/api/monitor/smart-detector-alert-rules) in `Support Resources" list of ASO.

theunrepentantgeek:

We've noticed a significant number of properties that should have been flagged as secrets, but which weren't noticed earlier. In this PR, we're improving our secret detection heuristics to try and prevent this from happening in the future.

Closes #4269
 
This detected a number of issues, which I think I've resolved correctly, but this needs a careful check. 

Blocked on merge of #4237 and this will likely require a rebase after that one merges.

gif


v2/azure-arm.yaml
```diff
@@ -1288,6 +1288,9 @@ objectModelConfiguration:
         $isResource: false
       DeepCreatedOriginProperties_STATUS:
         $isResource: false
+      DomainValidationProperties:
+        ValidationToken:
+          $isSecret: false # This might be a secret, @theunrepentantgeek is not sure
```
matthchr, line 1293: Also not sure... we should find+ read their docs?

v2/api/insights/customizations/component_extensions.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func secretsToWrite(obj *insights.Component) ([]*v1.Secret, error) {
+	operatorSpecSecrets := obj.Spec.OperatorSpec.Secrets
+	if operatorSpecSecrets == nil {
+		return nil, errors.Errorf("unexpected nil operatorspec")
+	}
+
+	collector := secrets.NewCollector(obj.Namespace)
+	collector.AddValue(operatorSpecSecrets.HockeyAppToken, to.Value(obj.Status.HockeyAppToken))
```
matthchr, line 60: Not following what this is doing - if it's on the status why is it a secret?

If this is just about making it exportable to a secret/configmap, the stuff I'm doing related to Dynamic export should cover it?

v2/azure-arm.yaml
```diff
@@ -2300,6 +2344,8 @@ objectModelConfiguration:
         $generatedConfigs:
           ConnectionString: $.Status.ConnectionString
           InstrumentationKey: $.Status.InstrumentationKey
+        $azureGeneratedSecrets:
+          - HockeyAppToken
```
matthchr, line 2348: Is this a secret or no? If it's on the status then it sortof seems like it can't be a secret (because we're storing it on Status which is unencrypted?)

v2/azure-arm.yaml
theunrepentantgeek: It's not a conventional secret.

If you're interacting with a domain name provider, this is the key you provide to prove ownership of the domain.

v2/api/insights/customizations/component_extensions.go
theunrepentantgeek: Removed.

v2/azure-arm.yaml
theunrepentantgeek: Removed.

v2/azure-arm.yaml
```diff
@@ -1782,6 +1818,12 @@ objectModelConfiguration:
         Id:
           $armReference: true
     2024-04-02-preview:
+      ContainerServiceSshPublicKey:
```
super-harsh, line 1821: Just a thought - can we exclude `PublicKey` from the secret check heuristic as I think we might encounter these in the future too.

v2/azure-arm.yaml
```diff
@@ -2554,6 +2598,9 @@ objectModelConfiguration:
       Databricks:
         ResourceId:
           $armReference: true
+      DatabricksProperties:
+        DatabricksAccessToken:
```
super-harsh, line 2602: Breaking change?

v2/azure-arm.yaml
```diff
@@ -2587,6 +2634,9 @@ objectModelConfiguration:
       DatabricksProperties:
         DatabricksAccessToken:
           $isSecret: true
+      VirtualMachineSshCredentials_STATUS:
+        PrivateKeyData:
```
super-harsh, line 2638: This shouldn't be included in the `STATUS`?

v2/azure-arm.yaml
super-harsh: I see below its already marked as `IsSecret: False` in the `SecretDetector`. So we can remove these now?

theunrepentantgeek: The property `KeyData` is triggering the heuristic - we don't currently use the object's name. Though maybe we should. 🤔

theunrepentantgeek: I'll think about doing this as a followup.

v2/azure-arm.yaml
theunrepentantgeek: Yup.

v2/azure-arm.yaml
theunrepentantgeek: Correct.

v2/azure-arm.yaml
matthchr: Where are the associated code-generated changes for this?

I only see 2 files in this PR now?

matthchr (commented):
Looks good except that I was expecting generated code changes which I don't see now?

Most of the changes are marking false-positives as not secrets but there is one new secret, right? Where is it in the generated code?

matthchr (approved):

v2/azure-arm.yaml
theunrepentantgeek: This isn't new - it's relocated from just below - we're supposed to have things in alphabetical order, so I moved it up into place. That's why there's no change to generated code.

PR #4295, opened 2024-09-26
Improve Type manipulation

theunrepentantgeek:

While convenience functions like `AsObjectType()` and `AsResourceType()` are, well, convenient, they can cause issues if the results are then used to create new types - because they will strip out and discard type structure information that should be retained.

While troubleshooting issues related to import of the `kusto` group of resources, we discovered a couple places where this was happening, resulting in issues. This PR addresses those issues.

Once merged, @super-harsh should be able to merge the changes into their working branch.

gif

### DIFF
--- a/v2/tools/generator/internal/astmodel/readonly_object_type.go
+++ b/v2/tools/generator/internal/astmodel/readonly_object_type.go
@@ -1,0 +1,25 @@
+package astmodel
+
+type ReadonlyObjectType interface {
+	// Properties returns all the property definitions
+	Properties() ReadOnlyPropertySet
+
+	// Property returns the details of a specific property based on its unique case-sensitive name
+	Property(name PropertyName) (*PropertyDefinition, bool)
+
+	// EmbeddedProperties returns all the embedded properties
+	// A sorted slice is returned to preserve immutability and provide determinism
+	EmbeddedProperties() []*PropertyDefinition
+
+	IsResource() bool
+	Resources() TypeNameSet
+
+	// Functions returns all the function implementations
+	// A sorted slice is returned to preserve immutability and provide determinism
+	Functions() []Function
+
+	// HasFunctionWithName determines if this object has a function with the given name
+	HasFunctionWithName(name string) bool
+
+	TestCases() []TestCase
+}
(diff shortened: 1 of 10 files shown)


v2/tools/generator/internal/astmodel/readonly_object_type.go
```diff
@@ -0,0 +1,33 @@
...
+
+// ReadonlyObjectType allows exposing an ObjectType for querying while prohibiting any modification.
+// This is useful as a convenience to consumers for working with ObjectTypes, while ensuring they
+// make any modifications using a TypeVisitor (or one of the Injectors based on TypeVisitor).
+type ReadonlyObjectType interface {
+	// Properties returns all the property definitions
+	Properties() ReadOnlyPropertySet
+
+	// Property returns the details of a specific property based on its unique case-sensitive name
+	Property(name PropertyName) (*PropertyDefinition, bool)
```
matthchr, line 16: Are we safe from the `PropertyDefinition` being modified here?

Maybe I can tell when I look at usage.

v2/tools/generator/internal/interfaces/kubernetes_resource_interface.go
```diff
@@ -119,38 +128,42 @@ func AddKubernetesResourceInterfaceImpls(
...
-	return r, nil
+	result := astmodel.MakeTypeDefinitionSetFromDefinitions(
+		updatedResource,
+		specDef,
+	)
+
+	return result, nil
 }
 
 // note that this can, as a side effect, update the resource type
```
matthchr, line 160: I think you've changed this now?

matthchr (approved):
LGTM, had a few minor comments

v2/tools/generator/internal/astmodel/readonly_object_type.go
theunrepentantgeek: Each `PropertyDefinition` is immutable, but yes, a new one could be created here.

That's not a problem because the new one would be created from the _whole_ existing property definition. The problem I'm solving there is that the `SpecType` previously exposed in the result struct was incomplete (any wrapper types already stripped) so using that as a base for further modification was flawed.

v2/tools/generator/internal/interfaces/kubernetes_resource_interface.go
theunrepentantgeek: Correct. Removed.

Issue #4298, opened 2024-09-27
Data encryption - PostgreSQL Flexible Server


nraooptum:
We have a policy / compliance requirement to use customer managed keys. I see that the PostgreSQL Flexible Server resource supports geoBackupKeyUri and primaryKeyUri which is great, but I don't see a way to generate those CMKs with ASOPv2. In terraform, we created one via `azurerm_key_vault_key` and gave the user-assigned managed identity access to it and nabbed the key URL from that. How can we achieve this with ASOP v2?

nraooptum:
Also, it seems we cannot supply a value from a config map for the key URIs either.

theunrepentantgeek:
We don't yet support creating Keys within Keyvaults, we're tracking that in #3188. I'll tag that for consideration when we triage things for our v2.11 release.

Enabling configuration for the Key URIs should be straightforward; I'll check into that and report back.

nraooptum:
I was able to work around that by deploying a custom (in-house) operator to augment the CMK functionality. I'll look at getting it over to ASO as well when I'm not so strapped for time. As for adding the missing config map support, looks like it's just a matter of adding `$importConfigMapMode: optional` to those fields in `azure-arm.yaml`?

theunrepentantgeek:
To make them configurable, yes - and I have that change in a working branch already. The most tedious bit was confirming the generated code is correct.

However, I'm wondering if those are actually ARM URLs, not arbitrary ones - which would instead involve tagging them with `$isARMResource: true` and they'd change from a string to a `KnownResourceReference`. Then, they could be configured with an ARM URL to reference a key in Azure, or with details of a Key within the cluster (once #3188 is done).

theunrepentantgeek:
We've confirmed that while these keys are URIs, they're not Azure Resource URIs - they come from the KeyVault dataplane API. PR #4311 will be merged shortly and released in the next version of ASO.

nraooptum:
Awesome -- looking forward to testing it. Yeah, from our tenants I can see it uses `*.vault.azure.net` for public cloud and `*.vault.usgovcloudapi.net` for azure gov cloud for example.

nraooptum:
@theunrepentantgeek is there anywhere I can go to get a nightly / early release version of the ASOP image? I would love to test it and give you my feedback. Plus, it would simplify quite a bit of logic in my custom operator.

nraooptum:
It's probably something I'm doing, but I'm getting:

[...]

PR #4311, opened 2024-10-01
Allow the data encryption keys for PostgreSQL Flexible server to be configured

theunrepentantgeek:
## What this PR does

Allows the values for `geoBackupKeyURI` and `primaryKeyURI` to be pulled from a configmap instead of requiring them to be hard coded in the resource.

Partially addresses #4298 

gif

### DIFF
--- a/v2/azure-arm.yaml
+++ b/v2/azure-arm.yaml
@@ -1974,6 +1974,8 @@ objectModelConfiguration:
           $armReference: true
     2022-12-01:
       DataEncryption:
+        PrimaryKeyURI:
+          $importConfigMapMode: optional
         PrimaryUserAssignedIdentityId:
           $armReference: true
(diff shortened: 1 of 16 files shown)


v2/azure-arm.yaml
```diff
@@ -1974,6 +1974,8 @@ objectModelConfiguration:
           $armReference: true
     2022-12-01:
       DataEncryption:
+        PrimaryKeyURI:
```
matthchr, line 1977: No GeoBackupKey in this API version?

matthchr (approved):

v2/azure-arm.yaml
theunrepentantgeek: Correct. I did check. ;-)

PR #4312, opened 2024-10-01
Retry if/when ScheduledQueryRules encounter BadRequest

### LINKED ISSUE #4293: Bug: Controller doesn't requeue ScheduledQueryRules
v2.8.0

**Describe the bug**
Our use-case is to deploy an Azure AKS scale-unit, consisting of multiple resources, including:
* AKS cluster
* Log Analytics Workspace, deployed via terraform
* ScheduledQueryRule, which ideally will be deployed by ASO.
When we create a new scale-unit, we do introduce a direct dependency for LAW to be created before the ScheduledQueryRule.
The first time ASO tries to create the resource it will fail, because it's waiting for the RG object to be created, but afterwards, it should fail and retry until the LAW object is created (eventual consistency).
But after the first failure of LAW missing, it won't requeue the object. We think that the Bad Request response got from Azure is not handled properly and goes into a permanent failure.

**To Reproduce**
* Deploy a ScheduledQueryRule, this should fail for LAW missing (expected)
* Deploy the missing LAW and reference it in the ScheduledQueryRule by ArmId
* The ScheduledQueryRule should reconcile now that the LAW is available, but it doesn't get requeued

**Expected behavior**
The controller should be able to retry missing resource requests, coded as Bad Requests.

**Screenshots**
These are the logs that we saw in the controller, but the object will not be reconciled.
```
I0918 13:38:25.265321       1 generic_reconciler.go:101] "msg"="Encountered error, re-queuing..." "logger"="controllers.ScheduledQueryRuleController" "name"="aso-flux-alerts" "namespace"="ns" "result"={"Requeue":false,"RequeueAfter":30000000000}
I0918 13:39:56.908898       1 generic_reconciler.go:101] "msg"="Encountered error, re-queuing..." "logger"="controllers.ScheduledQueryRuleController" "name"="aso-flux-alerts" "namespace"="ns" "result"={"Requeue":false,"RequeueAfter":0}
```

**Additional context**
We are also using CAPZ along with ASO, and we also have these kinds of resources there, which reference LAW via ArmId, and they will be eventually reconciled when the LAW is created. This should behave the same.

theunrepentantgeek:
# What this PR does

When reconciling an Insights `ScheduledQueryRule`, we can get a `BadRequest` response if the associated `Log Analytics Workspace` doesn't yet exist; we need to reclassify that as a retryable error.

Closes #4293

gif

### DIFF
--- a/v2/api/insights/customizations/scheduled_query_rule_extensions.go
+++ b/v2/api/insights/customizations/scheduled_query_rule_extensions.go
@@ -6,6 +6,8 @@ Licensed under the MIT license.
 package customizations
 
 import (
+	"strings"
+
 	"github.com/go-logr/logr"
 
 	"github.com/Azure/azure-service-operator/v2/internal/genericarmclient"
@@ -16,8 +18,8 @@ import (
 var _ extensions.ErrorClassifier = &ScheduledQueryRuleExtension{}
 
 // ClassifyError evaluates the provided error, returning whether it is fatal or can be retried.
-// A conflict error (409) is normally fatal, but ScheduledQueryRule resources may return 400 whilst a dependency is
-// being created, so we override for that case.
+// A badrequest (400) is normally fatal, but ScheduledQueryRule resources may return 400 whilst
+// a dependency is being created, so we override for that case.
 // cloudError is the error returned from ARM.
 // apiVersion is the ARM API version used for the request.
 // log is a logger than can be used for telemetry.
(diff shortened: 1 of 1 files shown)


v2/api/insights/customizations/scheduled_query_rule_extensions.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/Azure/azure-service-operator/v2/internal/genericarmclient"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/core"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/extensions"
+)
+
+var _ extensions.ErrorClassifier = &ScheduledQueryRuleExtension{}
+
+// ClassifyError evaluates the provided error, returning whether it is fatal or can be retried.
+// A conflict error (409) is normally fatal, but ScheduledQueryRule resources may return 400 whilst a dependency is
```
matthchr, line 19:
```suggestion
// A badrequest error (400) is normally fatal, but ScheduledQueryRule resources may return 400 whilst a dependency is
```

v2/api/insights/customizations/scheduled_query_rule_extensions.go
```diff
@@ -0,0 +1,55 @@
...
+	return details, nil
+}
+
+// isRetryableBadRequest checks the passed error to see if it is a retryable bad request, returning true if it is.
+func isRetryableBadRequest(err *genericarmclient.CloudError) bool {
+	if err == nil {
+		return false
+	}
+
+	if err.Code() == "BadRequest" {
```
matthchr, line 50: Should we look for a specific message? The `BadRequest` code may be used for multiple error messages?

matthchr (approved):

v2/api/insights/customizations/scheduled_query_rule_extensions.go
theunrepentantgeek: Updated the comment while addressing other issues.

v2/api/insights/customizations/scheduled_query_rule_extensions.go
theunrepentantgeek: I made it more restrictive, and added a comment explaining exactly what we're doing.

PR #4314, opened 2024-10-01
Add genruntime support for Dynamic secrets and configmap exports

matthchr:
Part of #2555.

See the design [here](https://azure.github.io/azure-service-operator/design/adr-2024-07-dynamic-export-to-secret-or-configmap/)

**If applicable**:
- [x] this PR contains tests


v2/pkg/genruntime/configmaps/validation.go
```diff
@@ -21,6 +22,11 @@ type keyPair struct {
 // ValidateDestinations checks that no two destinations are writing to the same configmap/key, as that could cause
 // those values to overwrite one another.
 func ValidateDestinations(destinations []*genruntime.ConfigMapDestination) (admission.Warnings, error) {
+	return ValidateDestinationsExt(destinations, nil)
+}
+
+// TODO: ValidateDestinationsExt will replace ValidateDestinations in a future PR.
+func ValidateDestinationsExt(destinations []*genruntime.ConfigMapDestination, destinationExpressions []*core.DestinationExpression) (admission.Warnings, error) {
```
theunrepentantgeek, line 29: Consider formatting these parameters on different lines to make the declaration easier to read.

```suggestion
func ValidateDestinationsExt(
        destinations []*genruntime.ConfigMapDestination, 
        destinationExpressions []*core.DestinationExpression,
) (admission.Warnings, error) {
```

v2/pkg/genruntime/configmaps/validation.go
```diff
@@ -39,6 +45,32 @@ func ValidateDestinations(destinations []*genruntime.ConfigMapDestination) (admi
 		locations.Add(pair)
 	}
 
+	if destinationExpressions == nil {
```
theunrepentantgeek, line 48: I don't think you need this guard clause; it doesn't change the behaviour.

v2/pkg/genruntime/configmaps/validation.go
```diff
@@ -39,6 +45,32 @@ func ValidateDestinations(destinations []*genruntime.ConfigMapDestination) (admi
...
+			continue
+		}
+
+		if dest.Key == "" {
+			// TODO: Key may be empty because of map[string]string supported exports.
+			// TODO: We should validate that in more depth but need a CEL parser to do so.
+			continue
+		}
+
+		pair := keyPair{
```
theunrepentantgeek, line 63: Consider adding a `createKeyPair()` method to `core.DestinationExpression()` (and to `genruntime.ConfigMapDestination`) to simplify these loops.

v2/pkg/genruntime/configmaps/validation.go
```diff
@@ -21,6 +22,11 @@ type keyPair struct {
 // ValidateDestinations checks that no two destinations are writing to the same configmap/key, as that could cause
 // those values to overwrite one another.
 func ValidateDestinations(destinations []*genruntime.ConfigMapDestination) (admission.Warnings, error) {
+	return ValidateDestinationsExt(destinations, nil)
+}
+
+// TODO: ValidateDestinationsExt will replace ValidateDestinations in a future PR.
```
theunrepentantgeek, line 28: Hasn't this replacement effectively already happened? given the code already on Line 25? Why not do the replacement now?

theunrepentantgeek (approved):
Looks like a good foundation. A couple of minor comments, keen to see the next step.

v2/pkg/genruntime/configmaps/validation.go
matthchr: Yes this is a holdover from when I had a different input type -- good catch

v2/pkg/genruntime/configmaps/validation.go
matthchr: No not yet, what I mean by this is that there is a future PR coming where a generator change also happens. When that happens, `ValidateDestinations` (1 arg) goes away and becomes `ValidateDestinations` (2 args).

The advantage of doing it like this was just that it let me split the PRs, because the generator PR is going to be pretty big.

v2/pkg/genruntime/configmaps/validation.go
matthchr: Given the simplicity of the structure and the fact that it's not exported, I opted to keep the pure struct-initialization pattern here rather than do a `new` or `make` method as it felt like it protected better against the main bug you could have here of getting `key` and `value` ordering mixed up.

PR #4321, opened 2024-10-07
ADR for changing handling of spec-one-of types

theunrepentantgeek:
## What this PR does

Two of our existing ASO design decisions are colliding and we need to decide how to resolve the conflict.

**TL;DR:** Our one-of handling requires common properties to be pushed down to the leaves; ARM spec generation requires `Name` to be on the root object of the spec.

gif

**If applicable**:

### DIFF
--- a/docs/hugo/content/design/ADR-2024-10-OneOf-Resources/_index.md
+++ b/docs/hugo/content/design/ADR-2024-10-OneOf-Resources/_index.md
@@ -1,6 +1,6 @@
 ---
 title: '2024-10 OneOf Resouces'
-toc_hide: true
+toc_hide: false
 ---
 
 ## Context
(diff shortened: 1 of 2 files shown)


docs/hugo/content/design/ADR-2024-10-OneOf-Resources/_index.md
```diff
@@ -0,0 +1,253 @@
...
+
+* Conceptually simpler
+
+### Conts
+
+* Changing already complex one-of handling to make it more complex.
+* Requires changing the way we serialize/deserialize one-of types in non-trivial ways.
+* Potentially large blast-radius if our changes impact on one-of types we've already generated and released.
+
+## Option 4: Special case root OneOf Types
```
matthchr, line 216: There's a variant of this where we just generate two totally different resources for them... it wouldn't involve changing JSON marshalling, instead we'd just pretend it wasn't a OneOf and was two totally separate resources that look normal.

The problem would be... what are they called and how does that interact with ownership (because they're not two different "kinds" of owners as per ARM, it might cause issues for child resources).

Not sure it's a great idea because of those complexities but figured I'd mention it as it would probably be doable.

docs/hugo/content/design/ADR-2024-10-OneOf-Resources/_index.md
```diff
@@ -0,0 +1,253 @@
...
+
+* Easy
+
+### Cons
+
+* Distasteful to decline to support a resource due to a technical limitation.
+* No guarantee that Kusto will be the only affected resource.
+* Risk that resources we already support will use this pattern in a new API version, rendering us unable to upgrade.
+
+## Option 2: Loosen the rules on Name
```
matthchr, line 184: I think I tend towards this option, which is awkward/weird but may make the most sense...

Resulting usage would be something like:
```
Database:
  spec:
    readWrite:
      name: ...
      azureName: ...
```

I think most usage of `Name` and `AzureName` by the generator is done through interfaces so as long as the `Database` type still implemented `AzureName()` we'd be OK? Making that function generate correctly will take some generator work though.

It _is_ a bit confusing for users but I'd argue it's about the same amount of confusing as using the Go SDK type (which takes an interface rather than a concrete type and thus requires the user to figure out what types impl that interface and then pick one)

matthchr (approved):

docs/hugo/content/design/ADR-2024-10-OneOf-Resources/_index.md
```diff
@@ -0,0 +1,459 @@
...
+```
+
+### Pros
+
+* Easily to implement using a new pipeline stage in the code generator (but: ease of implementation is only one factor of many).
+
+### Cons
+
+* Adds yet another level of nesting to affected resources when we've done a lot of prior work to flatten things out.
+* Potentially confusing to users expecting the structure to more or less match the ARM API
```
matthchr, line 437: I don't know (w.r.t. OneOf) how much of an issue this really is, because we're already different than the raw ARM API, as are some of the SDKs because most languages don't support true OneOf, at least not in a user-friendly way.

I think the biggest con here is just the extra level of nesting and what confusion(s) that might bring.

docs/hugo/content/design/ADR-2024-10-OneOf-Resources/_index.md
```diff
@@ -0,0 +1,459 @@
...
+
+* Conceptually simpler
+
+### Cons
+
+* Changing already complex one-of handling to make it more complex.
+* Requires changing the way we serialize/deserialize one-of types in non-trivial ways.
+* Potentially large blast-radius as these changes would impact on all one-of types we've already generated and released. We'd either need to take a breaking change on all these resources, or maintain two different sets of rules for one-of types.
+
+## Option 4: Special case root one-of types
```
matthchr, line 312:
```suggestion
## Option 4: Special case root one-of types to allow non-one-of properties
```

docs/hugo/content/design/ADR-2024-10-OneOf-Resources/_index.md
```diff
@@ -0,0 +1,459 @@
...
+
+### Pros
+
+* Limits the scope of impact of the change.
+
+### Cons
+
+* Need to change the generation of our JSON marshalling code to handle this special case.
+* Special casing is always a bit of a code smell.
+
```
matthchr, line 372: I think the other con here is that it's now less clear which properties are the "one-of" properties and which are not.

So when presented with an arbitrary resource spec like:
```
spec:
  foo:
  bar:
  baz:
  azureName:
  operatorSpec:
```

You just sort of need to know that foo/bar/baz, only one can be specified, but `azureName` and `operatorSpec` can be alongside them.

I suspect this is something that will trip people up once or twice and then they'll get the hang of it for a given resource, so it's not a huge problem, but it's really the biggest issue I see w/ this approach.

matthchr (approved):

PR #4330, opened 2024-10-10
Fix bug where some ConfigMap and Secret references could be missed

### LINKED ISSUE #4298: Data encryption - PostgreSQL Flexible Server

### LINKED ISSUE #4316: Bug: ConfigMap and Secret discovery doesn't work for properties that only exist on preview verisons
2.9.0

**Describe the bug**
When a `ConfigMapReference` field is exported _only_ in a preview version and not in the latest stable, the discovery mechanism we use to find the `ConfigMapReference` and look it up fails because it does the lookup on the storage version of the resource. The storage version of the resource has the `genruntime.ConfigMapReference` _in the PropertyBag_ of the storage version - but this means it's serialized to a string and isn't discoverable by reflection.

**To Reproduce**
Can be reproduced with PostgreSQLFlexibleServer GeoBackupKeyURI, see #4298

**Expected behavior**
ConfigMap export should work.

matthchr:
The ASO controller uses reflection to look through the resource and find the ConfigMapReference and SecretReference fields. This reflection could miss SecretReference and ConfigMapReference fields that had been serialized into the "PropertyBag" property on the storage type. This was most common when working with preview APIs that had added new ConfigMapReference or SecretReference fields that don't exist in the GA api version.

The fix is to use the converted version for reflection-based discovery.

Fixes #4298.
Fixes #4316.

**If applicable**:
- [x] this PR contains tests

### DIFF
--- a/v2/pkg/genruntime/convertible.go
+++ b/v2/pkg/genruntime/convertible.go
@@ -1,0 +1,105 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package genruntime
+
+import (
+	"reflect"
+
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/conversion"
+)
+
+func findStorageGVK(scheme *runtime.Scheme, gk schema.GroupKind) (schema.GroupVersionKind, error) {
+	interfaceType := reflect.TypeOf((*conversion.Hub)(nil)).Elem()
+
+	for gvk, rt := range scheme.AllKnownTypes() {
+		if gvk.Group != gk.Group || gvk.Kind != gk.Kind {
+			continue
+		}
+
+		if reflect.PointerTo(rt).Implements(interfaceType) {
+			return gvk, nil
+		}
+	}
+
+	return schema.GroupVersionKind{}, errors.Errorf("couldn't find hub type for group kind %s", gk.String())
+}
+
+// ObjAsOriginalVersion returns the obj as the original API version used to create it.
+func ObjAsOriginalVersion(obj ARMMetaObject, scheme *runtime.Scheme) (ARMMetaObject, error) {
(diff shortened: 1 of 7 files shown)


v2/pkg/genruntime/convertible.go
```diff
@@ -0,0 +1,105 @@
...
+	interfaceType := reflect.TypeOf((*conversion.Hub)(nil)).Elem()
+
+	for gvk, rt := range scheme.AllKnownTypes() {
+		if gvk.Group != gk.Group || gvk.Kind != gk.Kind {
+			continue
+		}
+
+		if reflect.PointerTo(rt).Implements(interfaceType) {
+			return gvk, nil
+		}
```
theunrepentantgeek, line 27: Why not use the existing `NewEmptyVersionedResourceFromGVK()` and then test using `if _, ok := empty.(conversion.Hub); ok`, avoiding the cryptic reflection code?

super-harsh (approved):

v2/pkg/genruntime/convertible.go
matthchr: Two reasons:
1. `scheme.AllKnownTypes` returns a `map[schema.GroupVersionKind]reflect.Type`, so it seemed to make some sense to stay in the "reflect world" when scanning it if I don't actually need an instance.
2. Since this is called in the (somewhat) hot path of the core reconcile loop logic, and can be hit at least a couple times through each loop (because there may be multiple versions of a resource for a given group + version) I was a bit concerned about performance/extra allocations. It _probably_ wouldn't matter but avoiding constructing 8-10 objects that are immediately thrown away when instead we can do a check via reflection that avoids those allocations.

I _almost_ just computed a map of this upfront once and then we could look it up whenever we wanted but that seemed overkill so instead settled on using a less-allocations approach using this reflection method that kept things "simple" (no extra variables to storage cached data that needed to be passed around) while still avoiding excess allocations.

PR #4339, opened 2024-10-14
Add support for AML Registries

### LINKED ISSUE #4041: Feature: add support for AML registry
Please add support for Azure Machine Learning [registries](https://learn.microsoft.com/en-us/azure/machine-learning/how-to-manage-registries). This would imply API version update for `Microsoft.MachineLearningServices` group at least to `2023-04-01`.

super-harsh:
Closes #4041 

This PR adds AML/Registry resource, samples, tests and recordings

**If applicable**:
- [x] this PR contains tests


v2/azure-arm.yaml
```diff
@@ -371,6 +371,18 @@ typeTransformers:
     remove: true
     because: it is supposed to be marked ReadOnly in Swagger
 
+  - group: machinelearningservices
+    name: SystemCreatedStorageAccount
+    property: ArmResourceId
+    remove: true
+    because: it is supposed to be marked ReadOnly in Swagger
```
theunrepentantgeek, line 378: Would we want to publish these as configmaps for people to consume?

If we do, maybe we need a new type-transformation to mark as read-only, so it ends up only on the `status`.

v2/api/machinelearningservices/v1api20240401/structure.txt
```diff
@@ -3,6 +3,316 @@ github.com/Azure/azure-service-operator/v2/api/machinelearningservices/v1api2024
 ------------------------------------------------------------------------------------
 APIVersion: Enum (1 value)
 └── "2024-04-01"
+Registry: Resource
+├── Owner: resources/v1apiv20191001.ResourceGroup
+├── Spec: Object (15 properties)
+│   ├── AzureName: Validated<string> (1 rule)
+│   │   └── Rule 0: Pattern: "^[a-zA-Z0-9][a-zA-Z0-9\\-_]{2,32}$"
+│   ├── DiscoveryUrl: *string
```
theunrepentantgeek, line 11: Is this another field that should be read-only but isn't?

theunrepentantgeek (approved):

v2/azure-arm.yaml
matthchr: I think this transformation actually does that, because by the time typeTransformers runs I'm pretty sure types are forked into STATUS and non-status. Removing it here on the non-status only removes it from the spec.

See for example:
```
  - group: machinelearningservices
    name: EncryptionProperty
    property: Status
    remove: true
    because: it is supposed to be marked ReadOnly in Swagger
```

but that field does exist on the status type still:

```
type EncryptionProperty_STATUS struct {
	// Identity: The identity that will be used to access the key vault for encryption at rest.
	Identity *IdentityForCmk_STATUS `json:"identity,omitempty"`

	// KeyVaultProperties: Customer Key vault properties.
	KeyVaultProperties *KeyVaultProperties_STATUS `json:"keyVaultProperties,omitempty"`

	// Status: Indicates whether or not the encryption is enabled for the workspace.
	Status *EncryptionProperty_Status_STATUS `json:"status,omitempty"`
}
```

matthchr (approved):

v2/api/machinelearningservices/v1api20240401/structure.txt
super-harsh: It can either set by the user or auto-generated if not

PR #4343, opened 2024-10-15
Loosen up Owner-Child resource subscription checks

### LINKED ISSUE #3754: Bug: remove subscription check between resource/owner for DNS records
2.4.0

**Describe the bug**
It's possible for clients to segregate management and location of the DNS zones to separate subscriptions. New alias records could be easily created to point to the azure resources in subscriptions other then the one containing DNS zone using Azure Portal with least privilege access. But current case is not working with ASO.

**To Reproduce**
Try to create any alias DNS record in different subscription from where ASO is running using `armId` value to point to the owner (DNS Zone) in different subscription and dummy traffic manager profile, e.g.
```yaml
# traffic manager profile create in Subscription-1
# ASO is running in AKS in Subscription-1
# DNS Zone is located in Subscription-2
apiVersion: network.azure.com/v1api20180501storage
kind: DnsZonesCNAMERecord
metadata:
  name: dummy-record
spec:
  owner:
    armId: >-
      /subscriptions/<Subscription-2>/resourceGroups/resourceGroup/providers/Microsoft.Network/dnszones/example.com
  targetResource:
    reference:
      group: network.azure.com
      kind: TrafficManagerProfile
      name: dummy-traffic-manager-profile
  TTL: 3600
```

**Expected behavior**
Record should be created successfully.

**Current behavior**
```
resource subscription "<guid>" does not match parent subscription "<<guid>>"
```

super-harsh:
Closes #3754 

We introduced this check a while ago with single-operator multitenancy credentials to prevent users from creating resources in a subscription that doesn't match the subscription of the owner resource. However, there could be scenarios where owner and child resources could be provisioned in distinct subscriptions. 

This PR is to loosen up check between Owner-Child resource subscription mismatch for _ARMID owner references_. From a security perspective Azure has access checks and the operation is not allowed, it'll be rejected at ARM level. 

_Will add documentation after we make a decision on taking the change_

**If applicable**:
- [x] this PR contains tests


theunrepentantgeek (commented):
This just removes the restriction - I thought we discussed yesterday that we needed it when the owner was within the cluster, but _possibly_ loosen it when the owner is referenced directly by ARM ID.

super-harsh:
@theunrepentantgeek I've made the change as per above.

theunrepentantgeek (approved):

v2/internal/controllers/owner_arm_id_negative_test.go
```diff
@@ -53,7 +53,7 @@ func Test_OwnerIsARMIDOfWrongType_Rejected(t *testing.T) {
 	tc.Expect(err).To(MatchError(ContainSubstring("expected owner ARM ID to be for a resource group, but was \"Microsoft.Compute/virtualMachines\"")))
 }
 
-func Test_OwnerIsARMIDFromDifferentSubscription_ResourceFails(t *testing.T) {
+func Test_OwnerIsNotARMIDFromDifferentSubscription_ResourceFails(t *testing.T) {
```
matthchr, line 56: minor: naming might be better as 
```suggestion
func Test_OwnerIsKubernetesResourceFromDifferentSubscription_ResourceFails(t *testing.T) {
```

(saying what the owner is, rather than what it is not)

v2/internal/resolver/resource_hierarchy.go
```diff
@@ -218,16 +218,6 @@ func (h ResourceHierarchy) fullyQualifiedARMIDImpl(subscriptionID string, origin
 			return "", err
 		}
```
matthchr, line 220: minor: Consider leaving a comment here explaining that we don't check that the owner matches the subscription on purpose?

Something like:

```suggestion
		// We used to have a check here ensuring that the owner ARM ID matched the referenced credential/secret 
		// ARM ID. That check was removed as with ARM ID-based owners the subscription the resource is being
		// deployed to is pretty obvious (it's directly in the ID) so we felt the chances of mistakenly deploying to 
		// an incorrect subscription with kube-based ownership.
```

matthchr (approved):
LGTM, with one minor comment

PR #4346, opened 2024-10-15
Add configuration to allow resolving name collisions when loading resources

### LINKED ISSUE #4337: Bug: Unable to import any resources from Insights version 2021-05-01-preview
## Describe the bug

When trying to import a new resource from Microsoft.Insights, the Load Types stage of the generator crashes with a panic. This is the error from the start of the panic (formatted for readability):

``` 
error generating code: 
failed to execute stage 1: Load all types from Swagger files: 
panic: 
While merging file /workspaces/azure-service-operator/v2/specs/azure-rest-api-specs/specification/monitor/resource-manager/Microsoft.Insights/preview/2021-05-01-preview/subscriptionDiagnosticsSettings_API.json: 
duplicate resource types generated
```

## To Reproduce

Add this to the `objectModelConfiguration` section for `insights`:

``` yaml
    2021-05-01-preview:
      DiagnosticSetting:
        $export: true
        $supportedFrom: v2.10.0
```

Also add this to the `typeFilters` (to allow the new preview version to be _seen_):

``` yaml
  - action: include
    group: insights
    version: v*20210501preview
    because: We're using DiagnosticSetting from this.
```

## Expected behavior

The specified resource should be imported.

If any error occurs, the error message should be useful to an end user.

## Additional context

Error reported [via Slack](https://kubernetes.slack.com/archives/C046DEVLAQM/p1727951816125199).

theunrepentantgeek:

When loading resources from `Microsoft.Insights` we are encountering a panic caused by a naming collision between two near identical resources, `DiagnosticsSettings`. 

To resolve this, we need to target specific renaming to the two types so they end up with different names. Our existing configuration doesn't operate this early in the pipeline, so we need to special case things.

Closes #4337 

## Special notes for your reviewer

I've left the configuration required for `insights` in our configuration file so things are ready to go for our contributer.

gif

### DIFF
--- a/v2/tools/generator/internal/jsonast/swagger_type_extractor.go
+++ b/v2/tools/generator/internal/jsonast/swagger_type_extractor.go
@@ -70,6 +70,7 @@ type ResourceDefinition struct {
 	ARMType             string // e.g. Microsoft.XYZ/resourceThings
 	ARMURI              string
 	SupportedOperations set.Set[astmodel.ResourceOperation]
+	Scope               astmodel.ResourceScope
 	// TODO: use ARMURI for generating Resource URIs (only used for documentation & ownership at the moment)
 }
 
@@ -278,6 +279,7 @@ func (extractor *SwaggerTypeExtractor) extractOneResourceType(
 			ARMType:             armType,
 			ARMURI:              operationPath,
 			SupportedOperations: supportedOperations,
+			Scope:               categorizeResourceScope(operationPath),
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


v2/tools/generator/internal/jsonast/swagger_type_extractor.go
```diff
@@ -953,3 +955,39 @@ func makeSchemaFromSimpleSchemaParam(param schemaAndValidations) *spec.Schema {
 
 	return schema
 }
+
+var (
+	resourceGroupScopeRegex = regexp.MustCompile(`(?i)^/subscriptions/[^/]+/resourcegroups/[^/]+/.*`)
+	locationScopeRegex      = regexp.MustCompile(`(?i)^/subscriptions/[^/]+/.*`)
+	resourceScopePrefixes   = []string{
```
matthchr, line 962: Shouldn't this be something like 
```suggestion
	extensionScopePrefixes   = []string{
```

matthchr: Also somewhat minor but in this instance you could probably get away with a single regex like:
`(?i)^/({scope}|{resourceUri})/.*` (might need to escape the `{`'s though)

Main advantage of this would just be to simplify the scope check at least for now.

matthchr (approved):

v2/tools/generator/internal/jsonast/swagger_type_extractor.go
theunrepentantgeek: I considered that but figured to take the more explicit approach.

PR #4310, opened 2024-10-01
Update container dependencies

theunrepentantgeek:
# What this PR does

Update Go version and versions of tools in the container. Staying up to date is useful. 

**How does this PR make you feel**:
gif

PR #4365, opened 2024-10-22
Fix incorrect Trivy download link

matthchr:
**If applicable**:
- [ ] this PR contains tests


theunrepentantgeek (approved):
How did we not notice?!?!?

matthchr:
I noticed when I checked the trivy output today - it only failed once and the issue was only added on Oct 17 in #4310 so it hasn't been broken long.

PR #4383, opened 2024-10-25
Update secrets interface to work for CEL export

matthchr:
* Interface is now ExportKubernetesSecrets.
* Now accepts additionalSecrets.
* Now returns KubernetesSecretExportResult including RawSecrets map[string]string, which can be passed to CEL.
* Also a new ExportConfigMaps interface for ConfigMaps.

Related to #2555 

- [x] this PR contains tests


v2/pkg/genruntime/kubernetes_config_exporter.go
```diff
@@ -14,12 +14,12 @@ import (
...
-// KubernetesExporter defines a resource which can create other resources in Kubernetes.
-type KubernetesExporter interface {
-	// ExportKubernetesResources provides a list of Kubernetes resource for the operator to create once the resource which
+// KubernetesConfigExporter defines a resource which can create configmaps in Kubernetes
+type KubernetesConfigExporter interface {
+	// ExportKubernetesConfigMaps provides a list of Kubernetes ConfigMaps for the operator to create once the resource which
 	// implements this interface is successfully provisioned. This method is invoked once a resource has been
 	// successfully created in Azure, but before the Ready condition has been marked successful.
-	ExportKubernetesResources(
+	ExportKubernetesConfigMaps(
```
matthchr, line 22: This function actually can still export any Kubernetes resource, we just have a special function for secrets.

I could keep this as `KubernetesExporter` rather than making it ConfigMap specific. Not sure if that's better?

v2/pkg/genruntime/kubernetes_secret_exporter.go
```diff
@@ -0,0 +1,45 @@
...
+type KubernetesSecretExportResult struct {
+	// Objs is the set of objects (secrets) to export.
+	// Only secrets defined on the operatorSpec.secrets are included here. Secrets referenced via a "secret expression"
+	// in operatorSpec.secretExpressions are returned in RawSecrets for later use.
+	Objs []client.Object
+
+	// RawSecrets contains the raw secret values from Azure.
+	// The keys are the "names" of the secrets as defined on operatorSpec.secrets (JSON-cased), and the
+	// values are the actual secrets. So for example ManagedCluster has "adminCredentials" and "userCredentials".
+	// This will ONLY contain secrets that were requested via additionalSecrets, NOT secrets requested via
```
theunrepentantgeek, line 27: This sentence seems incomplete.

v2/internal/set/set.go
```diff
@@ -107,3 +107,14 @@ func AsSortedSlice[T constraints.Ordered](set Set[T]) []T {
 	slices.Sort(result)
 	return result
 }
+
+// Union returns the union of the specified sets
+func Union[T comparable](sets ...Set[T]) Set[T] {
+	result := make(Set[T])
```
theunrepentantgeek, line 113: Should/could this be presized?

v2/internal/reflecthelpers/reflect_helpers.go
```diff
@@ -379,3 +379,47 @@ func setPropertyCore(obj any, propertyPath []string, value any) (err error) {
 	field.Set(reflect.ValueOf(value))
 	return nil
 }
+
+// PopulateStruct is far from complete/perfect and shouldn't be used in production code.
+// Test usage is OK.
+// deprecated: Don't use it in production code.
+func PopulateStruct(s any) {
```
theunrepentantgeek, line 386: I don't follow what this is for.

v2/internal/reconcilers/arm/kubernetes_resource_exporter.go
```diff
@@ -0,0 +1,256 @@
...
+			return nil, err
+		}
+
+		if value.Value != "" {
+			collector.AddValue(
+				&genruntime.ConfigMapDestination{
+					Name: expression.Name,
+					Key:  expression.Key,
+				}, value.Value)
+		} else if len(value.Values) > 0 {
```
theunrepentantgeek, line 69: Could you have both `value.Value != ""` and `len(value.Values) > 0` true at the same time? 

If so, the current behaviour is to use `value.Value` and ignore `value.Values` - is this the desired behaviour? Or should this be an error?

theunrepentantgeek (approved):

v2/internal/reconcilers/arm/kubernetes_resource_exporter.go
matthchr: It's not really possible to have both, if you look at the [CompileAndRun](https://github.com/Azure/azure-service-operator/blob/main/v2/internal/util/cel/cel.go#L207) function, it will only ever return either `Value` or `Values` and which of those it does is directly related to if the users CEL expression is of result type `string` or `map[string]string`.

I think here it's OK to treat the `ExpressionResult` from `CompileAndRun` as a discriminated union here (which is what it's intended to be). If there were a bug here, it would be in the `ComileAndRun` function returning two different types of output. As implemented at least today, that's not possible.

v2/pkg/genruntime/kubernetes_secret_exporter.go
matthchr: Fixed!

v2/internal/set/set.go
matthchr: Yeah it could be pre-sized. I initially didn't do that because it would be looping over the set of sets twice, but that's still probably more performant than doing extra allocations. I've updated to iterate the set of sets, get their length and preallocate.

Also added some tests.

v2/internal/reflecthelpers/reflect_helpers.go
matthchr: It's for testing, see the tests at v2/api/apimanagement/customizations/subscription_extensions_test.go for example.

Specifically what I wanted to test was: 

- Fill a given `XyzSecrets` object. For example for `apimanagement` above, the `SubscriptionOperatorSecrets` object has a `primaryKey` and `secondaryKey` field.
- Call the `secretsSpecified` helper on the above object, which returns the set of secrets requested.
- Ensure that the set of secret keys matches the set of JSON keys on the objects.

Or maybe put another way, I wanted to assert that this:
```
const (
	primaryKey   = "primaryKey"
	secondaryKey = "secondaryKey"
)
```

Equals the non-$propertyBag keys from this:
```
type SubscriptionOperatorSecrets struct {
	PrimaryKey   *genruntime.SecretDestination `json:"primaryKey,omitempty"`
	PropertyBag  genruntime.PropertyBag        `json:"$propertyBag,omitempty"`
	SecondaryKey *genruntime.SecretDestination `json:"secondaryKey,omitempty"`
}
```

This being in reflecthelpers is awkward though because we use that in production and this method is super hacky. I moved it to a new testcommon/reflect package which should make it more obvious it's just for test.

v2/internal/reconcilers/arm/kubernetes_resource_exporter.go
theunrepentantgeek: Fair enough.

PR #4398, opened 2024-10-31
Add support for dynamic expression Secret and ConfigMap outputs

### LINKED ISSUE #2555: Feature: Support dynamic secret/configmap exports

matthchr:
* Add generator support for dynamic configmap and secret export.
 * Add webhook validation for new operatorSpec fields.
 * Add generic controller integration to output expressions to
   ConfigMap/Secret.
 * Add CEL tests
 * Fix a few issues with CEL expression parsing.

Resolves #2555 

- [x] this PR contains tests


matthchr:
Recommend reviewers review by commit and they can bypass (or review last) the "generate APIs" commit.

v2/tools/generator/internal/functions/dynamic_exports.go
```diff
@@ -0,0 +1,148 @@
...
+	"strings"
+
+	"github.com/dave/dst"
+	"github.com/pkg/errors"
+
+	"github.com/Azure/azure-service-operator/v2/tools/generator/internal/astbuilder"
+	"github.com/Azure/azure-service-operator/v2/tools/generator/internal/astmodel"
+)
+
+// TODO: Docs
```
matthchr, line 19: todo

v2/pkg/genruntime/configmaps/validation.go
```diff
@@ -48,17 +45,39 @@ func ValidateDestinationsExt(
...
+		// Evaluator may be nil in some cases, if it is don't use it
+		if celEvaluator != nil {
+			// Validate that the expression is a valid CEL expression
+			outputType, err := celEvaluator.Check(dest.Value, self)
+			if err != nil {
+				// Error should already have enough context
+				return nil, err
+			}
+
+			// TODO: Need a test for this
```
matthchr, line 64: We have tests for these now , remove todo

v2/internal/controllers/cel_secret_export_test.go
```diff
@@ -0,0 +1,208 @@
...
+	managedidentity "github.com/Azure/azure-service-operator/v2/api/managedidentity/v1api20181130"
+	storage "github.com/Azure/azure-service-operator/v2/api/storage/v1api20230101"
+	"github.com/Azure/azure-service-operator/v2/internal/testcommon"
+	"github.com/Azure/azure-service-operator/v2/internal/util/to"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/conditions"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/core"
+)
+
+// TODO:
```
matthchr, line 24: This is done, remove

v2/pkg/genruntime/configmaps/validation.go
matthchr: Actually we didn't but I added them

theunrepentantgeek (approved):

docs/hugo/content/guide/expressions.md
```diff
@@ -0,0 +1,168 @@
...
+and the resulting output, still based on the example `Redis` above.
+
+| Description        | Expression                                             | Result                                                                                                  |
+|--------------------|--------------------------------------------------------|---------------------------------------------------------------------------------------------------------|
+| Hardcoded string   | `"helloworld"`                                         | `helloworld`                                                                                            |
+| Formatted string   | `"%s:%d".format([self.spec.location, 7])`              | `westus2:7`                                                                                             |
+| String math        | `self.metadata.namespace + ":" + self.metadata.name`   | `default:sampleredis1`                                                                                  |
+| Int output (error) | `self.spec.sku.capacity`                               | Error, expression "self.spec.sku.capacity" must return one of [string,map(string, string)], but was int |
```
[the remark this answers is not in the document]
matthchr: We technically could, but it would require us to coerce it in Go, not in CEL. What we could do is support more than just `string` and `map[string]string` outputs and do the coercion ourselves afterwards, effectively expanding the set of allowed types from `string` and `map[string]string` to include others.

The problem with that is: 
- What types do we covert? Everything or just ints? Just ints feels incomplete, but if we start converting other types like doubles too it gets more complicated as you then have to talk about precision and how many decimal points you keep in the string representation - different users may want different values, etc.
- Do we convert types in _all_ contexts or only in the "single-return" context. What I mean is, say we auto-converted ints. Do we also auto-convert map keys that are ints to be strings? What about map values?

I don't know in practice how often users are going to export ints. If you look at all of the existing fields that we export via ConfigMap every one of them seems to natively be a string. For secrets there are a few ints (Redis `Port` and `SSLPort` for example), but at least based on existing exports, pure strings are the 95% case.

All the above suggests to me that the best thing to do for user clarity is to keep it simple: The expression output type must be `string` or `map[string]string`, and if you're dealing with types _other_ than those it's up to you to convert them yourself using inbuilt CEL helpers to do so. This keeps our life easy, keeps it easy to explain how expressions work to users, and gives users the most control to do exactly what they want without relying on magic on our end.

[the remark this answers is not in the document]
theunrepentantgeek: Fair enough.

PR #4399, opened 2024-10-31
Undo force covecov to /mnt

matthchr:
There was a bug in the GitHub runner infrastructure we use that caused the build to happen on the wrong drive. Now that this bug is fixed we don't need this workaround anymore.

- [ ] this PR contains tests


theunrepentantgeek (approved):

matthchr:

matthchr:
(hoping above re-triggers CLA check since it seems stuck...

Issue #4418, opened 2024-11-05
Release notes tracking for 2.11


matthchr:
- [x] Announce a removal of some ManagedCluster API versions in 2.12 release, as the CRD has gotten too big. We should remove the following APIs:
  - 2021-05-01 (2.12)
  - 2023-11-02-preview (2.12)
  - 2023-02-01 (2.13?)
- [x] As part of the above, we should clarify our stance on how many `ManagedCluster` resources we support and what the migration path is. This is probably related to #4376 so that users don't need to constantly manage removing `storedVersions` via `asoctl`.

PR #4419, opened 2024-11-06
Add support for 2024-09-01 version of AKS API

### LINKED ISSUE #4164: Feature: Support for Microsoft.ContainerService/managedClusters/maintenanceConfigurations
The lack of this feature prevents the automatic upgrade configuration of AKS clusters.
Thank you!

### LINKED ISSUE #4327: Feature: Support newest AKS stable API
Current latest supported AKS API version is 2023-10-01
There have been a few other stable versions since then.  Request to update to latest 2024-05-01.

matthchr:
## What this PR does

- Fix issue in `RepairSkippingProperties` which could cause an issue when a type is reused among multiple resources, such as in a containerservice ManagedCluster and AgentPool.
- Update the rest-api-specs submodule, to get access to the hot off the presses 2024-09-01 version of the containerservice API.
- Add new `StripLargeDescriptions` pipeline stage, which can be used to strip descriptions off CRD types to reduce their size.
  - We don't want to do this in perpetuity, see #4418 which tracks removing some old versions of ManagedCluster so we can drop this in a future release.
 - Add support for 2024-09-01 version of AKS API.
 - Add support for MaintenanceConfiguration.

Fixes #4164.
Fixes #4327.

- [x] this PR contains tests


v2/specs/azure-rest-api-specs
theunrepentantgeek, line 1: We should do the update of this submodule in a separate PR, will be easy to review and will reduce the size of this one.

v2/tools/generator/internal/codegen/pipeline/strip_descriptions.go
```diff
@@ -0,0 +1,96 @@
...
+func createPropertyDescriptionRemovalVisitor() astmodel.TypeVisitor[any] {
+	visitor := astmodel.TypeVisitorBuilder[any]{
+		VisitObjectType: func(
+			this *astmodel.TypeVisitor[any],
+			it *astmodel.ObjectType,
+			ctx any,
+		) (astmodel.Type, error) {
+			result := it
+			for _, prop := range it.Properties().AsSlice() {
+				// If the property already has a description, use it
```
theunrepentantgeek, line 81:
```suggestion
				// If the property already has an empty description, use it
```

v2/azure-arm.yaml
```diff
@@ -1900,6 +1900,59 @@ objectModelConfiguration:
...
+          $payloadType: explicitEmptyCollections
+        Tags:
+          $payloadType: explicitEmptyCollections
+      ManagedClusterAgentPoolProfileProperties:
+        NodeLabels:
+          $payloadType: explicitEmptyCollections
+        NodeTaints:
+          $payloadType: explicitEmptyCollections
+        Tags:
+          $payloadType: explicitEmptyCollections
```
theunrepentantgeek, line 1937: You can put `$payloadType` at any level, so you could put it directly on `ManagedClusterAgentPoolProfileProperties` and simplify, if all collections should be treated the same way.

v2/specs/azure-rest-api-specs
matthchr: https://github.com/Azure/azure-service-operator/pull/4427

v2/tools/generator/internal/codegen/pipeline/strip_descriptions.go
matthchr: Fixed.

matthchr:
I've changed how I stripped descriptions to make it a configuration option in `objectModelConfiguration` and done it per resoruce version, and only stripped some of the older versions documentation. Agree that's better - I had missed that it would impact our reference docs, which as you say we definitely don't want to lose.

PR #4426, opened 2024-11-06
Move ARM types for resources

theunrepentantgeek:

Removes the legacy suppression for `resources` and regenerates the ARM files in their new location.

gif


v2/internal/genericarmclient/generic_client_test.go
```diff
@@ -12,13 +12,16 @@ import (
 	"net/http/httptest"
 	"testing"
 
+	. "github.com/onsi/gomega"
+
+	resources "github.com/Azure/azure-service-operator/v2/api/resources/v1api20200601"
```
matthchr, line 17: Where is this actually used now? As far as I can tell there isn't any actual usage of this import here.

theunrepentantgeek: Used in the test `Test_NewResourceGroup`, see line 63.

matthchr: Find in IE betrayed me.

matthchr (approved):

PR #4430, opened 2024-11-09
Deduplicate SubResource for networking

matthchr:
This fixes the SubResource from DnsResolver to match the SubResource from core networking so that we don't have a million different <Resource>SubResource's created.

The issue is that dnsresolver and networking are two different Swagger specifications from two different Swagger folders that both put resources into the networking group. The 2022-07-01 API just happened to be supported by both networking and dnsresolver, but many subsequent API versions of plain networking don't correspond to a dnsresolver version. This results in inconsistencies between type names in the two versions of core networking which causes issues adding a new networking version due to conversion problems.

- [ ] this PR contains tests


matthchr:
Planning to file an item to do this in a less hacky way, although I don't really know what a less hacky way would be...

theunrepentantgeek (approved):

PR #4433, opened 2024-11-11
Fix report case sensitivity

theunrepentantgeek:
## What this PR does

Prior to this point, every resource group imported by ASO has been consistent, with its resource URLs all defined with identical letter case. This results in all the resources within the group having an identical _provider_ name as the prefix of their ARM type.

For example, all the resources in the `storage` group have an ARMType() with the prefix `Microsoft.Storage`, including

* Microsoft.Storage/storageAccounts
* Microsoft.Storage/storageAccounts/blobServices
* Microsoft.Storage/storageAccounts/blobServices/containers
* Microsoft.Storage/storageAccounts/fileServices

and so on.

However, the `alertsmanagement` group is not consistent with its letter casing, resulting in differences between resources.

* microsoft.alertsManagement/smartDetectorAlertRules
* Microsoft.AlertsManagement/prometheusRuleGroups

This PR modifies report generation to produce consistent results even when the underlying resources are not. When looking up metadata, we prefer resources with PascalCasing (with provider names starting `Microsoft.`) over resources with other casing conventions.

gif

## Checklist

- [x] this PR contains tests

### DIFF
--- a/v2/tools/generator/internal/codegen/pipeline/report_resource_versions.go
+++ b/v2/tools/generator/internal/codegen/pipeline/report_resource_versions.go
@@ -175,14 +175,14 @@ func (report *ResourceVersionsReport) summarize(definitions astmodel.TypeDefinit
 		name := rsrc.Name()
 		pkg := name.PackageReference()
 
-		defType := astmodel.MustBeResourceType(rsrc.Type())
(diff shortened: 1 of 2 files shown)


babbageclunk (approved):
Would it be worth prioritising a provider with a title-cased suffix after `Microsoft.` as well?

PR #4440, opened 2024-11-12
Tidy import order

theunrepentantgeek:
## What this PR does

We have strong conventions on import ordering, but it's not always the case that we catch variance from those conventions during code reviews. 

This PR activates the linter [**gci** linter](https://github.com/daixiang0/gci) to check, and enforce, import ordering in the following order:

- standard (Go standard library)
- dot imports (typically gomega, in test files)
- aliased imports (separated for visibility)
- default (other third party libraries)
- prefix(github.com/Azure/azure-service-operator/) (everything internal)

I've used `prefix` instead of `localmodule` so that we can run a single pass of **golangci-lint**, instead of needing to invoke it once for each component, with the active directory containing the `go.mod` file to use each time.

I've excluded **gci** from all generated files, so this only impacts hand-written code. 

Cleanup was automated by using the built in fix:

``` bash
$ golangci-lint run --fix ./...
```

### Special Notes

Suggest reviewing by commit, as the actual changes are quite small. The `Update .*` commits are large, but entirely mechanical.

gif


.golangci.yml
```diff
@@ -4,31 +4,52 @@ run:
...
   exhaustive:
     default-signifies-exhaustive: true
+  gci:
+    custom-order: true
+    sections:
+      - standard
+      - dot
+      - alias
+      - default
+      - prefix(github.com/Azure/azure-service-operator/)
```
matthchr, line 34: I don't mind pulling out dot aliases because they're weird/uncommon and also have a different impact on import usage than standard impacts.

I'm not sold on pulling out just plain aliases though. Plain aliases are much more common and don't really change the usage of the package, nor do they really feel related to one another, at least not in nearly the same way that the standard/default/prefix groups are.

What do you think about just:

```suggestion
    sections:
      - standard
      - dot
      - default
      - prefix(github.com/Azure/azure-service-operator/)
```

theunrepentantgeek: I could be persuaded. I tend to pull the aliased imports out into a separate group because it looks messy with them in the `default` list.

matthchr: I prefer them inline. I took a quick look at some random other projects and determined:
... a lot of projects don't enforce any ordering, which isn't surprising I guess because the default Go tooling doesn't either.

the ones I found that seemed to follow a pattern follow the one I suggested (not pulling the `alias` out and allowing inline aliases):
- [msal1](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/main/apps/confidential/confidential_test.go), [msal2](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/main/apps/internal/base/internal/storage/items_test.go)
- [controller-runtime1](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/envtest/envtest_suite_test.go), [controller-runtime2](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/cache/cache.go)
- [client go](https://github.com/kubernetes/client-go/blob/master/openapi/groupversion_test.go)

I probably prefer that pattern because it's the same pattern Python had (which also allows inline aliases).

Issue #4277, opened 2024-09-17
Use a more specific selector than control-plane: controller-manager


matthchr:
**Describe the current behavior**
We use the kubebuilder default selector `control-plane: controller-manager`, which many other projects may also use. We should be good citizens and at least in theory support being installed into the same namespace, but if we're using the same labels as other deployments webhooks and other selectors may get confused.

See #4274 as an example.

**Describe the improvement**
Use a label unique to ASO.

matthchr:
The challenge here is that we have to delete the deployment in order to change this field.

We can do this with a Helm hook, but given it's a bit involved I'm going to kick this out to 2.11 for now.

matthchr:
The actual error is:
> * spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/name":"azure-service-operator", "control-plane":"controller-manager"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable

Note that this happens even if the deployment mode is `Recreate` - the only solution is to delete and recreate the deployment itself.

matthchr:
The consequence of not doing this, is that you cannot run multiple ASO deployments in the same namespace, because they all select one-another.

matthchr:
I still would like to do this but it's pretty low priority honestly.

matthchr:
We should consider changing/fixing this at least for service selectors to use a better label?

PR #4445, opened 2024-11-13
Enable high availability (HA) configuration for ASO

### LINKED ISSUE #4215: Change default pod count to 2
Even though ASO wouldn't use a second pod for reconciliation because it has a single primary, it could use it for serving webhooks, which would help avoid kubectl failures when interacting with ASO resources during a Kuberentes upgrade (as 1 of the 2 pods would be down due to node drain).

We should also consider setting a PDB of 1 if we do this.

bingikarthik:
## What this PR does

Closes #4215
As we adopt Azure Service Operator (ASO), it's essential to run multiple replicas for production workloads, especially during cluster upgrade operations. This PR enables high availability (HA) for ASO by increasing the replica count and implementing a Pod Disruption Budget (PDB) to maintain consistent uptime and resilience during potential disruptions.

gif

## Checklist

- [ ] this PR contains tests


config/manager/manager.yaml
```diff
@@ -17,7 +17,7 @@ spec:
   selector:
     matchLabels:
       control-plane: controller-manager
-  replicas: 1
+  replicas: 2
```
matthchr, line 20: This is ASOv1's config, you want to change ASOv2's, which is [here](https://github.com/Azure/azure-service-operator/tree/main/v2/config/manager).

Same goes from the `config/default/manager_pod_disruption_budget.yaml` file above - it should be in `v2/config/manager` instead I think.

config/default/manager_pod_disruption_budget.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: controller-manager
+  namespace: system
```
matthchr, line 5: Move to `v2/config/manager` folder, as mentioned below.

config/default/manager_pod_disruption_budget.yaml
```diff
@@ -0,0 +1,12 @@
...
+  labels:
+    control-plane: controller-manager
+spec:
+  selector:
+    matchLabels:
+      app: azure-service-operator-v1
```
matthchr, line 11: Should use the same `matchLabels` as the Helm variant.

matthchr: Which should also probably be:
```
    matchLabels:
      control-plane: controller-manager
```

since that's what the deployment and various other matchers use.

We have an item to change this #4277, but for now I think best to keep them all in sync.

bingikarthik: Thanks, this was updated now!

config/manager/manager.yaml
bingikarthik: Thanks, this was moved now.

v2/config/default/manager_pod_disruption_budget.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: policy/v1
```
matthchr, line 1: Rather than put this in `default` can you put it in `v2/config/manager/manager_pod_disruption_budget.yaml`, and add that file to the `kustomization.yaml`?

v2/config/default/manager_pod_disruption_budget.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: PodDisruptionBudget
+metadata:
+  name: controller-manager
+  namespace: system
+  labels:
+    control-plane: controller-manager
+spec:
+  selector:
+    matchLabels:
+      app: azure-service-operator-v1
```
matthchr, line 11: Fix this matchLabel to be `control-plane: controller-manager`

v2/charts/azure-service-operator/templates/policy_v1_ poddisruptionbudget_azureserviceoperator-pdb.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  labels:
+    app.kubernetes.io/name: azure-service-operator
+    app.kubernetes.io/version: v2.11.0
```
theunrepentantgeek, line 6: We'll need to update our release docs to include updating this.

theunrepentantgeek (commented):
How will things behave when upgrading ASO to the next version? 

I'm worried about the scenario where the version N+1 of ASO deploys new CRD versions which (still running) version N doesn't understand. 

In this scenario, since we use the newest resource version as the storage (hub) version of the resource, any resource that has been touched by the version N+1 will be unintelligible to version N, resulting in a panic/crash.

v2/charts/azure-service-operator/templates/policy_v1_ poddisruptionbudget_azureserviceoperator-pdb.yaml
matthchr: I think it's done automatically by tooling already

matthchr: it is, see most recent Helm udpate PR: https://github.com/Azure/azure-service-operator/pull/4442/files

v2/config/default/manager_pod_disruption_budget.yaml
bingikarthik: Done!

bingikarthik:
@matthchr I've made the requested changes. When you have a moment, could you kindly review the PR?

matthchr:
@bingikarthik - thanks! I think we need to make a deployment rollout change to make this safe (as @theunrepentantgeek called out). I'll send a separate PR for that and link it here, and then once that merges we can merge this too.

nishant221:
Does ASO support "leader" kind of approach? If multiple replicas are running (and reconciling independently), can the same request can be processed by multiple instances of ASO resulting in duplicate calls to Azure?

bingikarthik:
> Does ASO support "leader" kind of approach? If multiple replicas are running (and reconciling independently), can the same request can be processed by multiple instances of ASO resulting in duplicate calls to Azure?

Yes, it was indeed. Please check: https://github.com/Azure/azure-service-operator/blob/main/v2/charts/azure-service-operator/templates/apps_v1_deployment_azureserviceoperator-controller-manager.yaml#L59
https://github.com/Azure/azure-service-operator/blob/main/main.go#L63

matthchr (approved):

matthchr:
had a few comments - once they're fixed I'll kick off CI and we can merge this.

v2/config/manager/manager_pod_disruption_budget.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: controller-manager
+  namespace: system
+  labels:
```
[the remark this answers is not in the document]
bingikarthik: This was updated now, thanks!

v2/config/manager/manager_pod_disruption_budget.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: controller-manager
```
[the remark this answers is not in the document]
bingikarthik: This was updated now, thanks!

super-harsh (approved):

theunrepentantgeek:
Thanks @bingikarthik, much appreciated.

reespozzi:
Shouldn't this be an option via feature flagging instead of defaulted?

matthchr:
Are you having issues with having 2 pods?

Our thinking on making this on by default is that we run active/passive anyway, and so the marginal cost of an extra pod isn't much, and the extra speed during failovers and webhook resilience is worth that cost by default.

PR #4460, opened 2024-11-21
Allow recovery if resources have a missing AzureName

### LINKED ISSUE #4306: Allow setting AzureName if it is currently blank
# Describe the current behavior

Currently our validation blocks any change to `AzureName` because this can't be changed in Azure.

However, this also blocks the case of _setting_ `AzureName` when it's empty (which can occur if the webhooks fail).

# Describe the improvement

Allow `AzureName` to be changed if it's currently empty, allowing users to self-correct if something goes wrong (see #4285 for an example where this would have been useful).

## Additional context

We'll probably need to clear the annotation with the resource ID as well, as that will need to be regenerated.

### LINKED ISSUE #4307: Allow deletion of resources with `AzureName` empty
# Describe the current behavior

The `AzureName` of every resource should be set automatically by our webhook, but if that fails we can end up with resources where `AzureName` is empty. This results in an invalid ARM resource URL, resulting in an error when someone tries to delete the resource.

# Describe the improvement

When `AzureName` is empty (and therefore when any generated ARM resource ID will be invalid by definition), skip the attempt to delete the Azure resource and allow deletion of the cluster resource to proceed. (This would have enabled easier recovery for #4285.)

theunrepentantgeek:
## What this PR does

Allows `AzureName` to be updated if it somehow ends up blank due to a glitch with our webhooks.

Closes #4306 

Blocks construction of a fully qualified resource ID if `AzureName` is blank, preventing attempts to create the resource and ensuring an attempt to delete the resource can proceed.

Closes #4307 

### Special notes

Adds a guard clause to the generated `Owner()` function to avoid a nil panic if we end up with a resource lacking an owner.

## Checklist

- [x] this PR contains tests


v2/pkg/genruntime/admissions.go
```diff
@@ -38,8 +38,17 @@ func ValidateWriteOnceProperties(oldObj ARMMetaObject, newObj ARMMetaObject) (ad
 		return nil, nil
 	}
 
-	if oldObj.AzureName() != newObj.AzureName() {
-		errs = append(errs, errors.Errorf("updating 'spec.azureName' is not allowed for '%s : %s", oldObj.GetObjectKind().GroupVersionKind(), oldObj.GetName()))
+	// Prohibit changing the AzureName,
```
matthchr, line 41: Nice

v2/tools/generator/internal/interfaces/kubernetes_resource_interface.go
```diff
@@ -409,26 +409,33 @@ func getOwnerFunction(
...
 
 	owner := astbuilder.Selector(specSelector, astmodel.OwnerProperty)
+	nilOwnerCheck := astbuilder.IfNil(owner, astbuilder.Returns(astbuilder.Nil()))
+	nilOwnerCheck.Decs.After = dst.EmptyLine
 
 	switch r.Resource().Scope() {
 	case astmodel.ResourceScopeResourceGroup:
 		fn.AddComments("returns the ResourceReference of the owner")
 		fn.AddStatements(
+			nilOwnerCheck,
```
matthchr, line 419: Are these checks just being defensive? I'm not 100% following them (though I don't mind them if it's just being defensive either).

matthchr (approved):

v2/tools/generator/internal/interfaces/kubernetes_resource_interface.go
theunrepentantgeek: Defensive checks added as a result of new unit tests that discovered an issue in `admissions.go`.

PR #4461, opened 2024-11-21
Include golangci-lint fixes when formatting code

theunrepentantgeek:
## What this PR does

Now that we're enforcing import order using [**gci**](https://github.com/daixiang0/gci), the exacting rule means getting a pull request across the line might be challenging.

Fortunately, there's an automated fix that we can apply by running `golangci-lint` with the `--fix` parameter. Adding this to the `format-code` task in the `Taskfile.yml` makes this easy to access.

gif

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -1289,7 +1289,7 @@ tasks:
   verify-no-changes:
     desc: Checks that there are no uncommitted modifications to files
     cmds:
-      - cmd: 'if [ -n "{{.CHANGED}}" ]; then echo "Error: files were modified during the build (did you forget to commit generated files?):"; echo "{{.CHANGED}}"; git diff; exit 1; fi'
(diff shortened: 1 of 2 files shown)


matthchr (approved):

Taskfile.yml
```diff
@@ -103,6 +103,7 @@ tasks:
     desc: Ensure all code is formatted
     dir: v2
     cmds:
+      - golangci-lint run --fix ./...
```
matthchr, line 106: This is called both in the `ci` task and the local task `default`. Do we want it executed in both - is the experience w/ this changing code in CI good?

matthchr: Maybe we should change the `verify-no-changes` task to have an error that gives the cmd to run to fix things up?

matthchr (approved):

Taskfile.yml
theunrepentantgeek: Good idea. Done.

PR #4465, opened 2024-11-22
Adopt ASO-specific label to replace default control-plane: controller-manager

bingikarthik:
## What this PR does

Closes https://github.com/Azure/azure-service-operator/issues/4277

**Current Behavior**:
The kubebuilder default label `control-plane: controller-manager` is currently used. This label is common across many projects, which can lead to potential conflicts, especially when multiple deployments or webhooks exist in the same namespace. Such conflicts may cause misrouting of resources and selectors, impacting functionality.

**Improvement**:
This PR introduces a unique label specific to ASO, ensuring better namespace coexistence and reducing the risk of conflicts with other deployments. By aligning with best practices, this change improves ASO's compatibility and coexistence within shared Kubernetes environments.

**Impact**:

- Avoids label conflicts with other projects.
- Supports installation alongside other controllers in the same namespace.
- Enhances stability and maintainability of deployments using ASO.

This update addresses potential selector misconfigurations and promotes more robust operations in multi-controller setups.

gif

## Checklist

- [ ] this PR contains tests

### DIFF
--- a/scripts/add-openshift-cert-handling.sh
+++ b/scripts/add-openshift-cert-handling.sh
@@ -19,7 +19,7 @@ spec:
       protocol: TCP
       targetPort: 9443
   selector:
-    control-plane: controller-manager
+    app.kubernetes.io/name: azure-service-operator
   sessionAffinity: None
   type: ClusterIP
 EOF
--- a/v2/charts/azure-service-operator/templates/_helpers.tpl
+++ b/v2/charts/azure-service-operator/templates/_helpers.tpl
@@ -1,4 +1,39 @@
 {{/*
+Expand the name of the chart.
+*/}}
+{{- define "azure-service-operator.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Create chart name and version as used by the chart label.
+*/}}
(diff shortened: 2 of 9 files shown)


bingikarthik:
Yes, agree this won't apply on existing deployments! To apply probably deployment strategy to have: `Recreate`(i.e., need to rollout following PR: https://github.com/Azure/azure-service-operator/pull/4466)

matthchr:
I tested with strategy Recreate and it still doesn't work. The only way is to delete and recreate the deployment. See discussion on #4277.

I would like to do this, but we'll need to be careful about it, and I think the change is more involved than what you have done because the change as-proposed won't allow for Helm chart upgrades.

This means that (for now) we can't take this change as-is.

Can you expand more on where you ran into this issue? Even though I'd like to see this get fixed, in practice it shouldn't cause much issue if you install ASO into its own namespace (which is recommended anyway). Can you expand more on what you had it collide with?

bingikarthik:
> in practice it shouldn't cause much issue if you install ASO into its own namespace (which is recommended anyway)

Agreed on this point! We're already following a similar approach. This is more of a nice-to-have feature than a must-have. In case someone deploys a controller with control-plane: controller-manager labels in the same namespace where ASO is running, it could lead to a conflict. We can address this issue in the future. Closing the PR for now.

PR #4452, opened 2024-11-18
Separate success and failure flows in asoctl resource importing

theunrepentantgeek:
## What this PR does

Separates the handling of successful resource imports from failures, clearing the way for an upcoming change to the way resources are named. 

## How does this PR make you feel?

gif

PR #4475, opened 2024-11-28
Fix: asoctl deadlock

theunrepentantgeek:
## What this PR does

Fixes an ASO deadlock introduced in #4452.

The immediate cause of the deadlock was an omitted call to `Completed(1)` when collating import errors, leaving the progress bar incomplete.

However, this coupling between the progress bar and the importer isn't ideal, and leaves the door open for addition problems in the future. To avoid this, I've made two additional changes.

* Introduced a `watchdog` to keep track of inflight resources, to monitor the import, and to shut things down when everything is done
* Refactored construction of the data flow graph into separate sections that can be more easily reasoned about independently.

### DIFF
--- a/v2/cmd/asoctl/pkg/importresources/resource_importer.go
+++ b/v2/cmd/asoctl/pkg/importresources/resource_importer.go
@@ -11,6 +11,8 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/rotisserie/eris"
+	"github.com/sourcegraph/conc"
+	"golang.org/x/exp/maps"
 	"k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/Azure/azure-service-operator/v2/cmd/asoctl/pkg/importreporter"
@@ -81,35 +83,25 @@ func (ri *ResourceImporter) Import(
 	ctx context.Context,
 	done chan struct{},
 ) (*Result, error) {
-	workersRequired := ri.desiredWorkers()
+	resources := make(chan ImportableResource) // resources to import
+	allstages := conc.NewWaitGroup()
 
-	candidates := make(chan ImportableResource)  // candidates that need to be deduped
-	pending := make(chan ImportableResource)     // importers that are pending import
-	successes := make(chan ImportResourceResult) // importers that have been executed successfully
-	failures := make(chan ImportError)           // errors from importers that failed
-	completions := make(chan struct{})           // channel to signal completion
+	watchdog := ri.startWatchdog(resources)
 
-	// Dedupe candidates so we import each distinct resource only once
-	go ri.queueUniqueImporters(candidates, pending, ri.reporter)
+	pendingResources := ri.startDeduplicator(resources, watchdog, allstages)
+	successes, failures := ri.startWorkers(ctx, pendingResources, allstages)
 
-	// Create workers to run the import
-	for i := 0; i < workersRequired; i++ {
-		go ri.importWorker(ctx, pending, successes, failures, ri.reporter, completions)
-	}
-
-	// Collate the results
 	report := newResourceImportReport()
-	go ri.collateResults(successes, candidates, ri.reporter, report, completions)
-	go ri.collateErrors(failures, report, completions)
+	ri.startCollationOfResults(successes, resources, watchdog, allstages, report)
+	ri.startCollationOfErrors(failures, watchdog, allstages, report)
 
 	// Set up by adding our initial resources; these will be completed when we collate their results
 	for _, rsrc := range ri.resources {
-		candidates <- rsrc
+		resources <- rsrc
 		ri.reporter.AddPending(1)
 	}
 
-	// Wait while everything runs
-	<-done
+	allstages.Wait()
 
 	// Check for an abort, and return the error if so
 	if ctx.Err() != nil {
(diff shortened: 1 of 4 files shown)


babbageclunk (commented):
Partial review, I should keep packing rather than getting sidetracked by fun concurrency fixes! Looks good so far though!

v2/cmd/asoctl/pkg/importresources/resource_importer.go
```diff
@@ -81,116 +83,149 @@ func (ri *ResourceImporter) Import(
...
 	done chan struct{},
 ) (*Result, error) {
-	workersRequired := ri.desiredWorkers()
-
-	candidates := make(chan ImportableResource)  // candidates that need to be deduped
-	pending := make(chan ImportableResource)     // importers that are pending import
-	successes := make(chan ImportResourceResult) // importers that have been executed successfully
-	failures := make(chan ImportError)           // errors from importers that failed
-	completions := make(chan struct{})           // channel to signal completion
+	resources := make(chan ImportableResource) // resources to import
```
super-harsh, line 86: Thoughts on making this a buffered channel?

v2/cmd/asoctl/pkg/importresources/resource_importer.go
```diff
@@ -207,7 +242,6 @@ func (ri *ResourceImporter) importWorker(
 	completed chan<- ImportResourceResult,
 	failed chan<- ImportError,
 	progress importreporter.Interface,
-	done chan<- struct{},
```
super-harsh, line 210: Need to update fn doc to remove `done` channel

v2/cmd/asoctl/pkg/importresources/resource_importer.go
```diff
@@ -81,116 +83,149 @@ func (ri *ResourceImporter) Import(
...
+// resources is the channel of unique resources to import.
+// waitgroup is used to track the number of inflight goroutines.
+// returns two channels: one for successful imports, and one for errors.
+func (ri *ResourceImporter) startWorkers(
+	ctx context.Context,
+	resources <-chan ImportableResource,
+	waitgroup *conc.WaitGroup,
+) (chan ImportResourceResult, chan ImportError) {
+	successes := make(chan ImportResourceResult) // importers that have been executed successfully
+	failures := make(chan ImportError)           // errors from importers that failed
```
super-harsh, line 211: Should this be a buffered channel? 
Will need to double check what happens if multiple messages are sent on this channel and this gets closed in the end of the function (As we don't have an active listener at this point).

super-harsh (commented):
Minor comments. Just double checking on a few things, feel free to ignore if you think the comments are irrelevant.

v2/cmd/asoctl/pkg/importresources/resource_importer.go
```diff
@@ -81,116 +83,149 @@ func (ri *ResourceImporter) Import(
...
+			var upstream chan<- ImportableResource = nil
+			if current != nil {
+				upstream = uniqueResources
+			}
+
+			select {
+			case rsrc, ok := <-resources:
+				if !ok {
+					// Channel closed
+					running = false
```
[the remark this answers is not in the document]
theunrepentantgeek: Done.

v2/cmd/asoctl/pkg/importresources/resource_importer.go
theunrepentantgeek: I don't think it needs to be. 

The deduplicator created in `startDeduplicator` immediately consumes anything put on this channel and does all the buffering we need (and handles the size dynamically too).

v2/cmd/asoctl/pkg/importresources/resource_importer.go
theunrepentantgeek: Done.

v2/cmd/asoctl/pkg/importresources/resource_importer.go
theunrepentantgeek: It doesn't need to be, because the channels are closed in a controlled order (following advice from @babbageclunk to make sure things shut down well).

The internal WaitGroup `wg` is used to wait for all the workers to shut down cleanly before the channels `successes` and `failures` are closed. By using unbuffered channels, we know that the workers won't shut down until after all their errors and successes have been collated at the end.

super-harsh (approved):

v2/cmd/asoctl/pkg/importresources/resource_importer.go
```diff
@@ -81,116 +83,149 @@ func (ri *ResourceImporter) Import(
...
+				current = nil
 			}
-		case upstream <- current:
-			// We've sent the current importable, so clear it
-			ri.log.V(2).Info("Queued import", "resource", current.ID())
-			current = nil
 		}
+
+		// Close the channel when we're done, so that workers shut down too
+		close(uniqueResources)
```
babbageclunk, line 194: Defer this at the top of the goroutine func to ensure it's closed even in the case of panics?

v2/cmd/asoctl/pkg/importresources/resource_importer.go
```diff
@@ -225,68 +258,78 @@ func (ri *ResourceImporter) importWorker(
...
-		progress.Completed(1)
-	}
+			report.AddSuccessfulImport(gk)
+			ri.imported[rsrc.ID()] = rsrc
 
-	done <- struct{}{}
+			// Flag the main resource as complete
+			// We do this after everything else because it might indicate we're finished
+			ri.reporter.Completed(1)
+			watchdog.stopped()
```
babbageclunk, line 298: I'd put these into a method on ResourceImporter and then call it from both the success and error workers. Tying them together makes it harder to forget to do it in one place.

babbageclunk (approved):
Looks nice!

v2/cmd/asoctl/pkg/importresources/resource_importer.go
theunrepentantgeek: Done.

PR #4483, opened 2024-12-03
Minor testing improvements

theunrepentantgeek:
## What this PR does

Split the `format-code` task into separate targets, avoiding the problem where `golangci-lint` sometimes fails when applying fixes within a submodule. 

Improved logging in `create-kind-wi-storage.sh` for better traceability and debugging. 

Improved logging and increased timeouts `wait-for-operator-ready.sh` to provide more context.

gif

### DIFF
--- a/scripts/v2/wait-for-operator-ready.sh
+++ b/scripts/v2/wait-for-operator-ready.sh
@@ -28,27 +28,31 @@ function all_crds_have_cabundle() {
   for crd in $(kubectl get crd -l "app.kubernetes.io/name == azure-service-operator" -o name); do
     cabundle=$(kubectl get "$crd" -o jsonpath='{.spec.conversion.webhook.clientConfig.caBundle}')
     if [ -z "$cabundle" ]; then
-      echo "$crd has no CA bundle"
+      echo "[INF] $crd has no CA bundle"
       return 1
     fi
+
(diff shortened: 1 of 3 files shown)


scripts/v2/wait-for-operator-ready.sh
```diff
@@ -60,7 +64,7 @@ kubectl wait --for=condition=ready --timeout=3m pod -n "$OPERATOR_NAMESPACE" -l
 echo "Waiting for CRD cabundle..."
 export -f all_crds_have_cabundle
 export -f wait_for_crds_cabundle
-timeout 2m bash -c wait_for_crds_cabundle
+timeout 8m bash -c wait_for_crds_cabundle
```
super-harsh, line 67: Isn't 8 mins from 2 too high? 
Just thinking about if something else goes haywire then it waits for 4x longer than it used to.

super-harsh (approved):
LGTM! Left a minor comment on the wait time

scripts/v2/wait-for-operator-ready.sh
theunrepentantgeek: I incremented from 2m by 1m steps; things finally worked end to end at 7m, so I added one more for a little headroom.

Totally agree that this should be no higher than necessary.

PR #4487, opened 2024-12-05
Idiomatic resource naming for asoctl

### LINKED ISSUE #4349: Improve naming generated by asoctl

When importing resources using `asoctl`, it synthesizes resource names based on the hierarchy of ownership in order to generate unique names for every resource. 

This means that we generate different names based on which root is specified for import - importing an entire resource group will give different resource names compared to importing something more specific.

## Describe the improvement

Ideally, we'd generate the same names for resources regardless of the root used for the import. Even better, if we only synthesize unique names where required - if they're already unique, use the names directly from Azure.

The current approach is generating some long names with odd repetitions. 

> It seems like [asoctl importing] the resource group gets the correct owner, but pointing at the CosmosDB Database Account duplicates it wrongly
> `owner:`
> `    name: bulksend-cosmosdb-preview-nld2-bulksend-cosmosdb-preview-nld2 (edited) `

## Additional context

Reported on [Slack](https://kubernetes.slack.com/archives/C046DEVLAQM/p1729112594664079?thread_ts=1729103788.233489&cid=C046DEVLAQM).

theunrepentantgeek:
## What this PR does

Within a Kubernetes Cluster, resources need to have unique names only within resources of that type - the GroupKind of the resource is part of the uniqueness key.

This allows us to change `asoctl` to simply reuse the last part of the ARM ID for each resource as the name in most cases.

To disambiguate when multiple resources have the same "name" and collide, we add a unique random suffix as required. In testing, this happened rarely.

Closes #4349 

gif

## Checklist

- [x] this PR contains tests


v2/cmd/asoctl/pkg/importresources/importable_arm_resource.go
```diff
@@ -442,9 +445,10 @@ func (i *importableARMResource) getStatus(
 	ctx context.Context,
 	armID string,
 	importable genruntime.ImportableARMResource,
+	factory *importFactory,
```
super-harsh, line 448: We can pass _just the scheme_ here rather than the whole struct?

v2/cmd/asoctl/pkg/importresources/import_factory.go
```diff
@@ -0,0 +1,193 @@
...
+	"github.com/rotisserie/eris"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"github.com/Azure/azure-service-operator/v2/internal/set"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+)
+
+// importFactory is a helper struct that's used to create things.
+type importFactory struct {
```
super-harsh, line 25: `ImportHelper` sounds like a better name to me

super-harsh (approved):
LGTM! Just left a few minor comments.

v2/cmd/asoctl/pkg/importresources/import_factory.go
theunrepentantgeek: I try to avoid names like `Helper` and `Manager` as they become dumping grounds for too much.

I used factory as it's used for making things, but I'm open to a better suggestion.

v2/cmd/asoctl/pkg/importresources/importable_arm_resource.go
theunrepentantgeek: I have plans to tidy this further in the future, adding new methods to the factory that simplify this code further.

PR #4499, opened 2024-12-12
Add newer version of EventHub

### LINKED ISSUE #4221: Upgrade EventHub (and namespaces) to latest version
Can we get the event hubs and namespaces upgraded to the latest version as described here https://learn.microsoft.com/en-us/azure/templates/microsoft.eventhub/2024-01-01/namespaces?pivots=deployment-language-bicep, the current implementation does not support the ability to set TlsVersion which is default to 1.0 and will soon be deprecated.

I am still not certain how existing infrastructure will be affected in this change if they are on TLS1.0 but there is currently no way to set any other version of TLS through the ASO resources

theunrepentantgeek:
## What this PR does

Adds support for version 2024-01-01 of EventHub resources

Closes #4221

gif

## Checklist

- [x] this PR contains tests


super-harsh (approved):

.gitignore
```diff
@@ -81,3 +81,5 @@ hack/tools
 
 # goenv condiguration
 .go-version
+v2/controller
```
super-harsh, line 84: Are these binaries being generated by one of the targets?

theunrepentantgeek: No, they're what we get if you compile one of the projects directly with `go build`.

PR #4530, opened 2025-01-14
Fix "missing error information" error on 409s

matthchr:
This can happen if the Azure RP being called returns a JSON error response payload that doesn't match the Azure contract. Instead of overwriting the underlying error, just return it as-is to aid in debugging.

- [x] this PR contains tests

### DIFF
--- a/v2/internal/genericarmclient/policy_register_rp.go
+++ b/v2/internal/genericarmclient/policy_register_rp.go
@@ -64,11 +64,16 @@ func (r *rpRegistrationPolicy) Do(req *azpolicy.Request) (*http.Response, error)
 		return resp, err
 	}
 	var reqErr requestError
-	if err = runtime.UnmarshalAsJSON(resp, &reqErr); err != nil {
-		return resp, err
+	if unmarshalErr := runtime.UnmarshalAsJSON(resp, &reqErr); unmarshalErr != nil {
+		return resp, unmarshalErr
 	}
 	if reqErr.ServiceError == nil {
-		return resp, errors.New("missing error information")
+		// This is unexpected and likely due to Azure RPs failing to adhere to the expected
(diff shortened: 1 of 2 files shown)


v2/internal/genericarmclient/policy_register_rp.go
```diff
@@ -64,11 +64,16 @@ func (r *rpRegistrationPolicy) Do(req *azpolicy.Request) (*http.Response, error)
...
+	if unmarshalErr := runtime.UnmarshalAsJSON(resp, &reqErr); unmarshalErr != nil {
+		return resp, unmarshalErr
 	}
 	if reqErr.ServiceError == nil {
-		return resp, errors.New("missing error information")
+		// This is unexpected and likely due to Azure RPs failing to adhere to the expected
+		// error shape contract. The good news is that if this happens we know this isn't
+		// an unregistered RP because that API will have the right shape. We just return
+		// the resp/err directly in this case to let the caller deal with whatever the problem
+		// was as we know it wasn't unregistered RP.
```
theunrepentantgeek, line 75: I don't understand this - it appears to be saying that we know this is a registered RP because those have the right shape, but isn't the problem here due to the result having the wrong shape?

theunrepentantgeek (approved):
I think one comment needs cleanup, but otherwise this looks good.

v2/internal/genericarmclient/policy_register_rp.go
matthchr: What I'm trying to say is, the error for unregistered RP will have the correct shape. So if we end up in this if (incorrect shape), it had to come from the RP, which means its registered, which means we can just return the error as-is we don't need to try to parse/process it.

super-harsh (approved):

PR #4552, opened 2025-01-31
Fix linting in Taskfile

theunrepentantgeek:
## What this PR does

Including lint fixes in `format-code` has caused problems, as the linter requires the code to fully compile before a fix can be applied. 

We can easily end up in a situation where running `task` doesn't fix things because a linter failure blocks code generation.

This PR removes the linters from `format-code` and instead puts the `--fix` flag into the lint tasks.

gif

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -168,7 +168,6 @@ tasks:
     desc: Ensure all code for asoctl is formatted
     dir: v2/cmd/asoctl/
     cmds:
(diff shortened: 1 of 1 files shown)


matthchr:
Making sure I understand the ordering now:

1. We format the code
2. We run tests on the code (part of `quick-checks`)
3. We run linters on the code (including fixes)

That _seems_ reasonable to me - is that right?

PR #4555, opened 2025-02-02
Fix bugs in OneOf to Object conversion

theunrepentantgeek:
## What this PR does

Fixes a few small errors that I uncovered while troubleshooting issues with Kusto resources.

### Leaf Merging

Previously we've always merged object properties into the root of a one-of, then pushed them down to the leaves. Due to the way Kusto resources are structured, we now also need to support merging object properties directly into leaves. (Instead of embedding a OneOf directly into an AllOff, the Kusto resources instead refer to a separate type definition, triggering the change in behaviour.)

### Avoid lost updates

In two locations, the synthesizer was looking up an unmodified type definition when that type had already been modified, resulting in the previous updates being lost. All type lookup now goes through a helper method (`lookupType`) that avoids this propblem.

### OneOf Equality

We hadn't updated equality of one-of types, causing lost object updates when additional properties were added. (Two one-of objects, one containing three object-property types, the other four, were comparing as equal.)

### Special notes

Also updates CRD documentation for ContainerRegistry - now that documentation has been merged, the code generator has introduced linkes.

gif

## Checklist

- [x] this PR contains tests

### DIFF
--- a/v2/tools/generator/internal/astmodel/oneof_type_test.go
+++ b/v2/tools/generator/internal/astmodel/oneof_type_test.go
@@ -112,7 +112,7 @@ func TestOneOfType_WithoutAnyPropertyObjects_GivenProperties_ReturnsOneOfWithNon
 	g.Expect(result.PropertyObjects()).To(HaveLen(0))
 }
 
-func TestOneOfType_Equals_givenChange_recognisesDifference(t *testing.T) {
+func TestOneOfType_Equals_GivenChange_RecognisesDifference(t *testing.T) {
 	t.Parallel()
 
 	nameMixin := NewObjectType().WithProperties(
--- a/v2/tools/generator/internal/astmodel/oneof_type_test.go
+++ b/v2/tools/generator/internal/astmodel/oneof_type_test.go
@@ -15,33 +15,33 @@ func TestOneOfEqualityDoesNotCareAboutOrder(t *testing.T) {
 	t.Parallel()
 	g := NewGomegaWithT(t)
 
-	x := NewOneOfType("x", StringType, BoolType)
-	y := NewOneOfType("y", BoolType, StringType)
+	version1 := NewOneOfType("one", StringType, BoolType)
+	version2 := NewOneOfType("one", BoolType, StringType)
 
-	g.Expect(TypeEquals(x, y)).To(BeTrue())
-	g.Expect(TypeEquals(y, x)).To(BeTrue())
+	g.Expect(TypeEquals(version1, version2)).To(BeTrue())
+	g.Expect(TypeEquals(version2, version1)).To(BeTrue())
 }
 
 func TestOneOfMustHaveAllTypesToBeEqual(t *testing.T) {
 	t.Parallel()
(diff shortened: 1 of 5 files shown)


v2/tools/generator/internal/astmodel/oneof_type_test.go
```diff
@@ -15,33 +15,33 @@ func TestOneOfEqualityDoesNotCareAboutOrder(t *testing.T) {
 	t.Parallel()
 	g := NewGomegaWithT(t)
 
-	x := NewOneOfType("x", StringType, BoolType)
-	y := NewOneOfType("y", BoolType, StringType)
+	version1 := NewOneOfType("one", StringType, BoolType)
```
matthchr, line 18: This is technically a different assert than the one that was here before, because before the name of the type didn't matter and now it does (I see the name didn't matter before in the assert and now it does).

I assume that this is OK because we never have types that differ by names that are supposed to compare as equal (your change seems better than it was before...), but just want to confirm.

v2/tools/generator/internal/astmodel/oneof_type_test.go
```diff
@@ -111,3 +111,63 @@ func TestOneOfType_WithoutAnyPropertyObjects_GivenProperties_ReturnsOneOfWithNon
 	g.Expect(result).NotTo(BeNil())
 	g.Expect(result.PropertyObjects()).To(HaveLen(0))
 }
+
+func TestOneOfType_Equals_givenChange_recognisesDifference(t *testing.T) {
```
matthchr, line 115: minor:
```suggestion
func TestOneOfType_Equals_GivenChange_RecognisesDifference(t *testing.T) {
```

v2/tools/generator/internal/codegen/pipeline/convert_allof_and_oneof_to_objects.go
```diff
@@ -984,3 +1001,15 @@ func countTypeReferences(defs astmodel.TypeDefinitionSet) map[astmodel.TypeName]
 
 	return referenceCounts
 }
+
+func (s synthesizer) lookupType(name astmodel.InternalTypeName) (*astmodel.TypeDefinition, bool) {
```
matthchr, line 1005: minor: maybe a method comment?

matthchr (approved):

v2/tools/generator/internal/astmodel/oneof_type_test.go
theunrepentantgeek: Correct.

v2/tools/generator/internal/codegen/pipeline/convert_allof_and_oneof_to_objects.go
theunrepentantgeek: Added.

v2/tools/generator/internal/astmodel/oneof_type_test.go
theunrepentantgeek: Renamed.

PR #4572, opened 2025-02-11
Make reconcile-policy customizable for ASO operator (#4053)

### LINKED ISSUE #4053: Feature: Make reconcile-policy default value configurable
It would be ideal to be able to specify the default value for the annotation `reconcile-policy` to be changed. Currently its default is `manage`. 

The use case is, the teams are setting this through commonAnnotations in Kustomize for all azure resource manifests, which could easily be forgotten and as a cluster administrator it would be great to enforce this at the controller level. Then `reconcile-policy` can be explicitly set to `manage` to destroy those resources (via a extra deployment step for destruction).

### LINKED ISSUE #4562: Feature: Mutate/Annotate CRs to be read-only by default when explicitly required
Hello folks,
as of today, the default value of `serviceoperator.azure.com/reconcile-policy` annotation  is set to `manage` by default; when reconciliating ASO can perform any GET/PUT/DELETE actions on the remote resource.

However, it looks hard to me to support a multi-cluster scenario where CRs are supposed to be deployed to many clusters (ideally with a `gitops` approach) as part of the same chart (e.g. argocd application).

In such scenario electing a cluster as 'active' would require:
- wait for the deployment of the same application to all the clusters
- annotate all the CRs of interest with `serviceoperator.azure.com/reconcile-policy: skip`

This would require:
- Manually patch n-1 clusters (all the `passive`)
- Make sure that deployment mechanism (e.g. ArgoCD) ignores annotations CRs

This might suffer from a race condition where:
- the chart (or application) has been deployed on at least one cluster, eventually all
- while the annotation process is rolled out on all the `passive` clusters CRs are mistakenly deleted on a non-patched cluster
- ASO will reconciliate Azure resource deleting them permanently
- ASO on another cluster will create them back as part of the reconciliation process (CR changed -> reconcilie) or because of the sync period passed by

I believe it would be safer to think about a possibility to let ASO annotating CRs by default to `serviceoperator.azure.com/reconcile-policy: skip` and then patch the CRs on the elected `active` instance when a specific annotation (or label) is added to the namespace (or whatever)

This would require:
- Manually patch 1 cluster (only the `active`)
- Make sure that deployment mechanism (e.g. ArgoCD) ignores annotations on CRs and on namespace

Obviously, in case the remote resource does not exist on Azure, the CRs will not trigger any creation until an `active` instance is explicitly designated.

I'm opened to discuss about this need, basically here it would be helpful to have a mean to:
- Force annotation `serviceoperator.azure.com/reconcile-policy` to `skip` by default
- Identify a way to to switch it back to `manage` when explicitly required

What do you think about this?

antonioquintavalle1A:
## What this PR does

Extends ASO configuration to customize default reconcile policy for CRs when annotation `serviceoperator.azure.com/reconcile-policy` it not explicitly specified.
It keeps the same default behavior of today assuming that all CRs are in `manage` status, however this value can be overridden by setting DEFAULT_RECONCILE_POLICY environment variable.

closes #4562 
closes #4053 

I would appreciate to have your honest feedback on this code change; I'll add relevant unit tests and documentation if code looks correct on your side.

### DIFF
--- a/v2/internal/config/vars.go
+++ b/v2/internal/config/vars.go
@@ -13,6 +13,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
 	"github.com/rotisserie/eris"
 
+	"github.com/Azure/azure-service-operator/v2/pkg/common/annotations"
 	asocloud "github.com/Azure/azure-service-operator/v2/pkg/common/cloud"
 	"github.com/Azure/azure-service-operator/v2/pkg/common/config"
 )
@@ -100,6 +101,10 @@ type Values struct {
 	MaxConcurrentReconciles int
 
 	RateLimit RateLimit
+
+	// DefaultReconcilePolicy allows to override the default reconcile policy that should be used by ASO
+	// when the annotation serviceoperator.azure.com/reconcile-policy is omitted
+	DefaultReconcilePolicy annotations.ReconcilePolicyValue
 }
 
 type RateLimitMode string
@@ -180,6 +185,7 @@ func (v Values) String() string {
 	builder.WriteString(fmt.Sprintf("UserAgentSuffix:%s/", v.UserAgentSuffix))
 	builder.WriteString(fmt.Sprintf("MaxConcurrentReconciles:%d/", v.MaxConcurrentReconciles))
 	builder.WriteString(fmt.Sprintf("RateLimit:[%s]", v.RateLimit.String()))
+	builder.WriteString(fmt.Sprintf("DefaultReconcilePolicy:[%s]", v.DefaultReconcilePolicy))
 
 	return builder.String()
 }
@@ -246,6 +252,7 @@ func ReadFromEnvironment() (Values, error) {
 	if err != nil {
 		return result, err
 	}
+	result.DefaultReconcilePolicy = annotations.ReconcilePolicyValue(envOrDefault(config.DefaultReconcilePolicy, string(annotations.ReconcilePolicyManage)))
 
 	// Not calling validate here to support using from tests where we
 	// don't require consistent settings.
@@ -276,6 +283,9 @@ func (v Values) Validate() error {
 	}
 	if v.MaxConcurrentReconciles <= 0 {
 		return eris.Errorf("%s must be at least 1", config.MaxConcurrentReconciles)
+	}
+	if !(v.DefaultReconcilePolicy == annotations.ReconcilePolicyDetachOnDelete || v.DefaultReconcilePolicy == annotations.ReconcilePolicyManage || v.DefaultReconcilePolicy == annotations.ReconcilePolicySkip) {
+		return eris.Errorf("%s must be set to any of (%s, %s, %s)", config.DefaultReconcilePolicy, annotations.ReconcilePolicyDetachOnDelete, annotations.ReconcilePolicyManage, annotations.ReconcilePolicySkip)
 	}
 	return nil
 }
--- a/v2/internal/config/vars_test.go
+++ b/v2/internal/config/vars_test.go
@@ -80,3 +80,54 @@ func Test_Cloud_ResourceManagerAudienceSet(t *testing.T) {
 	g.Expect(cld.Services[cloud.ResourceManager].Endpoint).To(Equal(cloud.AzurePublic.Services[cloud.ResourceManager].Endpoint))
 	g.Expect(cld.Services[cloud.ResourceManager].Audience).To(Equal(cfg.ResourceManagerAudience))
 }
+
+func TestValidate(t *testing.T) {
+	t.Parallel()
+	g := NewGomegaWithT(t)
+
+	vOk := config.Values{
+		PodNamespace:            "test-namespace",
+		MaxConcurrentReconciles: 1,
+		OperatorMode:            3,
+		TargetNamespaces:        []string{},
+		DefaultReconcilePolicy:  "detach-on-delete",
+	}
+
+	g.Expect(vOk.Validate()).NotTo(HaveOccurred())
+
+	vKoNamespace := config.Values{
+		PodNamespace: "",
+	}
+
+	err := vKoNamespace.Validate()
+	g.Expect(err).To(MatchError(ContainSubstring("POD_NAMESPACE")))
+	g.Expect(err).To(MatchError(ContainSubstring("missing value")))
+
+	g.Expect(vOk.Validate()).NotTo(HaveOccurred())
+
+	vKoConcurrentReconciles := config.Values{
+		PodNamespace:            "test-namespace",
+		MaxConcurrentReconciles: 0,
+		DefaultReconcilePolicy:  "detach-on-delete",
+	}
+
+	g.Expect(vKoConcurrentReconciles.Validate().Error()).Error().Should(Equal("MAX_CONCURRENT_RECONCILES must be at least 1"))
+
+	vKoDefaultReconcilePolicy := config.Values{
+		PodNamespace:            "test-namespace",
+		MaxConcurrentReconciles: 1,
+		DefaultReconcilePolicy:  "detach",
+	}
+
+	g.Expect(vKoDefaultReconcilePolicy.Validate().Error()).Error().Should(Equal("DEFAULT_RECONCILE_POLICY must be set to any of (detach-on-delete, manage, skip)"))
+
+	vKoOperatorMode := config.Values{
+		PodNamespace:            "test-namespace",
+		MaxConcurrentReconciles: 1,
+		DefaultReconcilePolicy:  "detach-on-delete",
+		TargetNamespaces:        []string{"to-be-watched"},
+		OperatorMode:            2,
+	}
+
+	g.Expect(vKoOperatorMode.Validate().Error()).Error().Should(Equal("AZURE_TARGET_NAMESPACES must include watchers to specify target namespaces"))
+}
--- a/v2/internal/reconcilers/predicates.go
+++ b/v2/internal/reconcilers/predicates.go
@@ -35,9 +35,9 @@ func ARMPerResourceSecretAnnotationChangedPredicate() predicate.Predicate {
 }
 
 // GetReconcilePolicy gets the reconcile-policy from the ReconcilePolicy
-func GetReconcilePolicy(obj genruntime.MetaObject, log logr.Logger) annotations.ReconcilePolicyValue {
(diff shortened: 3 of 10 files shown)


v2/internal/config/vars.go
theunrepentantgeek, line 1: You need to also modify `Validate()` to ensure the configured default is one of the permissible values.

theunrepentantgeek (commented):
Looks good so far. I'd suggest adding a unit test in `reconcile_pollicy_test.go` to check that `ParseReconcilePolicy()` does the right thing in all cases.

v2/internal/config/vars.go
antonioquintavalle1A: Indeed, I forgot to amend accordingly the Validate() method here.
I took the liberty of adding the an unit test on the same module to assert validation issues we might have (existing + new one)

v2/internal/reconcilers/reconcile_policy.go
```diff
@@ -15,10 +15,12 @@ import (
...
 
 // ParseReconcilePolicy parses the provided reconcile policy.
-func ParseReconcilePolicy(policy string) (annotations.ReconcilePolicyValue, error) {
+// defaultPolicyValue is read from DEFAULT_RECONCILE_POLICY env variable, it set to "manage" when not specified
+func ParseReconcilePolicy(policy string, defaultReconcilePolicy string) (annotations.ReconcilePolicyValue, error) {
+	// policy is read from CR annotation, if it's empty it being read from defaultReconcilePolicy
 	switch policy {
 	case "":
-		return annotations.ReconcilePolicyManage, nil
+		return ParseReconcilePolicy(defaultReconcilePolicy, defaultReconcilePolicy)
```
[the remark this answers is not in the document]
antonioquintavalle1A: I didn't notice the ReconcilePolicyValue type, this makes the recusrive call totally useless.

v2/internal/config/vars_test.go
```diff
@@ -80,3 +80,52 @@ func Test_Cloud_ResourceManagerAudienceSet(t *testing.T) {
...
+		DefaultReconcilePolicy:  "detach-on-delete",
+	}
+
+	g.Expect(vOk.Validate()).NotTo(HaveOccurred())
+
+	vKoNamespace := config.Values{
+		PodNamespace: "",
+	}
+
+	g.Expect(vKoNamespace.Validate().Error()).Error().Should(Equal("missing value for POD_NAMESPACE"))
```
theunrepentantgeek, line 102: This assertion will blow up with a panic if `vKoNamespace.Validate()` fails to return an error. Plus, asserting on errors this way is brittle as any wording change to the error will break the test entirely. 

Suggested refactor:
```suggestion
	err := vKoNamespace.Validate()
	g.Expect(err).To(MatchError(ContainSubstring("POD_NAMESPACE")))
	g.Expect(err).To(MatchError(ContainSubstring("missing value"))
```
The `MatchError()` predicate only passes if an error is both present _and_ the error string matches the provided subpredicate.

theunrepentantgeek (commented):
This looks great. I'd like @mattchr to look at this too as he's more familiar with this part of ASO.

v2/internal/config/vars_test.go
antonioquintavalle1A: @theunrepentantgeek, at this stage I wonder if we should just check if an error is thrown when namespace is empty or if you feel comfortable in keeping the error message consistent with the substrings.
In the meanwhile, I've amended the PR with the latest suggestions.

v2/internal/config/vars.go
```diff
@@ -304,6 +312,9 @@ func (v Values) Validate() error {
 	if v.MaxConcurrentReconciles <= 0 {
 		return eris.Errorf("%s must be at least 1", config.MaxConcurrentReconciles)
 	}
+	if !(v.DefaultReconcilePolicy == string(annotations.ReconcilePolicyDetachOnDelete) || v.DefaultReconcilePolicy == string(annotations.ReconcilePolicyManage) || v.DefaultReconcilePolicy == string(annotations.ReconcilePolicySkip)) {
```
matthchr, line 315: minor, style: Other enum fields on `Values` we define the variable itself as the enum type, and then perform the validation in the parse function (see `OperatorMode`, or `RateLimitMode` for examples.

It's at best _marginally_ more clear to the user of the structure that the expected value is an enum, so there's not a ton of real advantage doing it that way over the way you've done it here, but I think from a consistency perspective it would be nice to match the other enums we parse.

v2/internal/config/vars_test.go
```diff
@@ -80,3 +80,55 @@ func Test_Cloud_ResourceManagerAudienceSet(t *testing.T) {
...
+		DefaultReconcilePolicy:  "detach-on-delete",
+	}
+
+	g.Expect(vOk.Validate()).NotTo(HaveOccurred())
+
+	vKoNamespace := config.Values{
+		PodNamespace: "",
+	}
+
+	// g.Expect(vKoNamespace.Validate().Error()).Error().Should(Equal("missing value for POD_NAMESPACE"))
```
matthchr, line 102: minor: Remove commented out assert?

v2/internal/config/vars_test.go
```diff
@@ -80,3 +80,55 @@ func Test_Cloud_ResourceManagerAudienceSet(t *testing.T) {
 	g.Expect(cld.Services[cloud.ResourceManager].Endpoint).To(Equal(cloud.AzurePublic.Services[cloud.ResourceManager].Endpoint))
 	g.Expect(cld.Services[cloud.ResourceManager].Audience).To(Equal(cfg.ResourceManagerAudience))
 }
+
+func TestValidate(t *testing.T) {
```
matthchr, line 84: Thanks for adding this!

v2/internal/reconcilers/predicates.go
```diff
@@ -35,9 +35,9 @@ func ARMPerResourceSecretAnnotationChangedPredicate() predicate.Predicate {
 }
 
 // GetReconcilePolicy gets the reconcile-policy from the ReconcilePolicy
-func GetReconcilePolicy(obj genruntime.MetaObject, log logr.Logger) annotations.ReconcilePolicyValue {
+func GetReconcilePolicy(obj genruntime.MetaObject, log logr.Logger, defaultReconcilePolicy string) annotations.ReconcilePolicyValue {
```
matthchr, line 38: If you do what I mentioned above too (make the config var the enum type), then you also save having to coerce it to an enum type here, which is nice.

matthchr (commented):
Some small feedback.

A larger question: do we want the default reconcile policy to "show up" on resources, so that users understand if they don't set it what they're getting?

We don't necessarily have to answer that in this PR, we could tack this change on later... the way to do it would probably be to create a webhook that automatically appends the default annotation on any resources that don't have an explicit reconcile-policy set.

v2/internal/config/vars.go
antonioquintavalle1A: Please let me know if now it looks a bit better

v2/internal/config/vars_test.go
theunrepentantgeek: This commented out assert is still here.

theunrepentantgeek (approved):

matthchr (approved):

theunrepentantgeek:
You're getting a CI failure due to a missing copyright header on a test file:
```
[header-check] ==> Checking copyright headers <==
[header-check] The following files are missing the required copyright header:
[header-check] 	./v2/internal/reconcilers/reconcile_policy_test.go
[header-check] Done checking copyright headers
task: Failed to run task "ci": exit status 1
```

Recommend you run `task` once you've fixed this to catch any other errors. Once it builds cleanly, we'll merge this.

I merged `main` into your branch and resolved some conflicts; if you intend on doing another force push, please fix those conflicts first.

theunrepentantgeek:
The file `reconcile_policy_test.go` is still missing a copyright header - this is enforced by our CI. Copy the header from `reconcile_policy.go`.

antonioquintavalle1A:
> The file `reconcile_policy_test.go` is still missing a copyright header - this is enforced by our CI. Copy the header from `reconcile_policy.go`.

Oh I'm sorry, I stashed that change and forgot to stage it later into the commit.
By the way, after the addition running locally the script `./scripts/validate-copyright-headers.sh` report such output:

`
==> Checking copyright headers <==
Copyright header check failed:
./v2/internal/ownerutil/ownerutil.go
`

is this expected?

theunrepentantgeek:
No, it's not expected. That file hasn't been changed in nearly 5 years, and PR #1643 fixed the header check so that style of header should pass. 

Let's see if the CI run (just triggered) flags any issues. If it does, I'll investigate and sort it out.

theunrepentantgeek:
Copyright header check passed fine; CI now failing on a lint error:

```
[controller:lint] internal/reconcilers/reconcile_policy_test.go:39:2: ineffectual assignment to returnedPolicy (ineffassign)
[controller:lint] 	returnedPolicy, err = ParseReconcilePolicy("whatever", annotations.ReconcilePolicySkip)
[controller:lint] 	^
```

You can reproduce this locally by running `task controller:lint`.

antonioquintavalle1A:
@theunrepentantgeek, I do not want to be picky here, but I haven't been able to run successfully the whole Taskfile.yml locally.
Since `task controller:lint` requires a lot of pre-requisites, including generation of dynamic files through controller-gen, wouldn't be easier to provide a container to ease local testing?

theunrepentantgeek:
We do have a container. Unless I'm misunderstanding your question.

The _devcontainer_ described in our [Developers Setup](https://azure.github.io/azure-service-operator/contributing/developer-setup/) guide is a container within which all the required versions of tooling are installed, ready to go.

This is the development environment I use every day when I work on ASO. I'm using it under WSL2 on Windows; @super_harsh uses it on his ARM based Mac. I don't think @matthchr uses it - he has all the tools installed directly on his Ubuntu VM, but he installs them using the same `install-dependencies.sh` script that the devcontainer uses.

Let me know how I can help you get unblocked. (My timeszone is NZDT, responses may be delayed during my weekend.)

matthchr:
> I don't think @matthchr uses it - he has all the tools installed directly on his Ubuntu VM, 

This isn't quite true. I use the container to run `task` and other commands, but I have Go dependencies/etc on a VM and use it with Goland. How to do this is covered in https://azure.github.io/azure-service-operator/contributing/developer-setup/#docker-on-linux

theunrepentantgeek:
This looks good, I'm hitting merge. Thanks for the contribution @antonioquintavalle1A

PR #4598, opened 2025-02-26
Improve FAQ documentation

### LINKED ISSUE #4576: Allow golang client generation with client-gen

The package does not currently seem to be compatible with [client-gen](https://github.com/kubernetes/code-generator/blob/master/cmd/client-gen/main.go). This is a shame because it means that I have to write ugly dynamic code in Golang to interact with the objects via the Kubernetes API.

## Describe the improvement

I am not 100% sure what needs to be done to add this compatibility. To start with, the CRD types need to be annotated with `// +genclient`. After generating from this package after patching that in, it seems to generate a client just fine, but now complains about missing a variable in the package called `SchemeGroupVersion`. There is a guide for how to use `client-gen` here, that might be enlightening:

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/generating-clientset.md

## Additional context

I might just be a noob, maybe there is some other way to get a golang client library for working with these CRDs that does not involve client-gen?

matthchr:
* Add documentation for workqueue overload performance issue previously seen by customers, and how to fix.
* Add documentation for importing ASO API types using controller-runtime client.

Fixes #4576.

- [ ] this PR contains tests


theunrepentantgeek (approved):

Peder2911:
Awesome, looks great!

PR #4658, opened 2025-03-28
Dependency updates

matthchr:
Note that this does NOT update CEL to the latest version, because kube-apiserver still gags if you do. I checked and kube-apiserver has a fix that allows CEL to go to 0.23.0+, but it's not in the latest apiserver tag yet. My guess is that it will be in 0.33.0+ APIServer

- [x] this PR contains tests


matthchr:
Note that this does NOT update CEL to the latest version, because kube-apiserver still gags if you do. I checked and kube-apiserver has a fix that allows CEL to go to 0.23.0+, but it's not in the latest apiserver tag yet. My guess is that it will be in 0.33.0+ APIServer

matthchr:
Finally think I've got this all working. Note that I had to switch some tests for KeyVault from Purge to Recover, due to policy.

v2/internal/controllers/compute_diskencryptionset_20220702_test.go
```diff
@@ -14,8 +14,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/keyvault/armkeyvault"
 
 	compute "github.com/Azure/azure-service-operator/v2/api/compute/v1api20220702"
-	"github.com/Azure/azure-service-operator/v2/api/keyvault/v1api20210401preview"
-	keyvault "github.com/Azure/azure-service-operator/v2/api/keyvault/v1api20210401preview"
+	keyvault "github.com/Azure/azure-service-operator/v2/api/keyvault/v1api20230701"
```
theunrepentantgeek, line 17: We had two imports of the same version? Yuck! Good to fix.

v2/internal/controllers/crd_keyvault_test_v1api20230701_test.go
```diff
@@ -120,7 +120,7 @@ func newVault20230701(name string, tc *testcommon.KubePerTestContext, rg *resour
 					Name:   to.Ptr(keyvault.Sku_Name_Standard),
 				},
 				TenantId:   to.Ptr(tc.AzureTenant),
-				CreateMode: to.Ptr(keyvault.VaultProperties_CreateMode_PurgeThenCreate),
+				CreateMode: to.Ptr(keyvault.VaultProperties_CreateMode_CreateOrRecover),
```
theunrepentantgeek, line 123: Reason for this change to the test?

theunrepentantgeek (approved):

v2/internal/controllers/crd_keyvault_test_v1api20230701_test.go
theunrepentantgeek: (Discussed offline.)

PR #4663, opened 2025-03-31
Use iterators to reduce memory heap debris

theunrepentantgeek:
## What this PR does

Uses Go iterators to reduce the amount of heap debris in hot paths. 

Doesn't have a great impact because we've already done a lot of performance work, but worth keeping.

gif


v2/tools/generator/internal/astmodel/package_import_set.go
```diff
@@ -55,6 +55,15 @@ func (set *PackageImportSet) AddImportsOfReferences(refs ...PackageReference) {
 	}
 }
 
+// AddImportsForPackageReferenceSet ensures this set includes an import of all the references in
+// the specified set.
+// Adding a reference already in the set is fine.
+func (set *PackageImportSet) AddImportsForPackageReferenceSet(refs *PackageReferenceSet) {
```
matthchr, line 61: minor: 
```suggestion
func (set *PackageImportSet) AddImportsOfPackageReferenceSet(refs *PackageReferenceSet) {
```

To match the naming of the above functions?

matthchr, line 61: We could make this take `refs iter.Seq[PackageReference]`, which then works with anything that can produce such an iterator, rather than the specific `PackageReferenceSet` type?

v2/tools/generator/internal/astmodel/package_reference_set.go
```diff
@@ -56,29 +56,15 @@ func (set *PackageReferenceSet) Contains(ref PackageReference) bool {
...
+			}
+		}
 	}
-
-	return result
-}
-
-// AsSortedSlice return a sorted slice containing all the references
-// less specifies how to order the imports
-func (set *PackageReferenceSet) AsSortedSlice(
```
matthchr, line 71: Was this function not actually used?

Interesting.

matthchr (approved):

v2/tools/generator/internal/astmodel/package_import_set.go
theunrepentantgeek: It would have to be `AddImportsOfPackageReferencesFromSet` which is getting too long, even for me. I think there's room to improve these names, but didn't come up with a viable approach for this PR.

v2/tools/generator/internal/astmodel/package_reference_set.go
theunrepentantgeek: I think we were using it at one point, but then we improved the way we generated imports and the order was no longer relevant.

v2/tools/generator/internal/astmodel/package_import_set.go
theunrepentantgeek: That would be a good followup. On my list.

Issue #2503, opened 2022-09-18
Handle additional property formats


theunrepentantgeek:
**Describe the current behavior**

During run of the code generator, we get warnings about unsupported property formats:

```
W0919 10:16:53.861756   37696 jsonast.go:339] unknown format "base64url"
W0919 10:16:53.908731   37696 jsonast.go:339] unknown format "uri"
W0919 10:17:24.732709   37696 jsonast.go:339] unknown format "byte"
```

**Describe the improvement**

We should add proper support for these formats so that our CRDs properly validate up front.

**Additional context**

This may technically be a breaking change, as we'll be rejecting values previously accepted - but given they would then be rejected by ARM, I don't think this will negatively impact any current users.

PawelHaracz:
@matthchr  can I take it and do it? should I know something more about these fields?

theunrepentantgeek:
You're welcome to take this issue and do it.

We already handle a bunch of formats - typically, a format gets translated into an underlying type (e.g. `IntType`) and some validation rules. For example, **byte** probably becomes `IntType` + Min: 0 + Max: 255

PawelHaracz:
@theunrepentantgeek can you help me and tell me where or how I can generate a config file for the code generator? or can you attach some an example?

theunrepentantgeek:
Apologies @PawelHaracz, somehow I missed your message. 😢 

If you're still interested (no problem if you're not), I'd suggest reusing the existing configuration file for the generator - `v2/azure-arm.yaml`. The commandline to run the generator would look like this:

```
$ generator gen-types azure-arm.yaml
```

When the data format is handled, you should see a few small changes in the generated code under `v2/api`.

PawelHaracz:
Thank you for your respons. Of course, I want to still try to help you. I gonna try to use this clip and resolve the issue 😃

matthchr:
We're still interested in doing this, although to date we haven't seen many issues with lack of this support.

matthchr:
No change from above.

theunrepentantgeek:
We still want this.

matthchr:
No change from above - not super critical but would be nice to have.

matthchr:
Seems related to https://github.com/Azure/azure-service-operator/pull/4676?

matthchr:
It looks like we might need to take another pass at this list and look at generator output to see what formats we have remaining.
edit: @theunrepentantgeek kindly did this and these are the results:

[...]

PR #4675, opened 2025-04-07
Reuse predefined value conversion functions

theunrepentantgeek:
## What this PR does

In `genruntime` we have a library of general purpose value conversion functions that weren't being as widely reused as they should have been. 

This stems from the use of exact type matches, which weren't matching for validated property types.

By matching validated variants as well, we broaden the use of these methods.

Discovered while addressing #2503 

The active code change is in `property_conversions.go`, most everything else is generated code.

gif

### DIFF
--- a/v2/api/apimanagement/v1api20220801/api_types_gen.go
+++ b/v2/api/apimanagement/v1api20220801/api_types_gen.go
@@ -872,28 +872,13 @@ func (api *Api_Spec) ConvertSpecTo(destination genruntime.ConvertibleSpec) error
 func (api *Api_Spec) AssignProperties_From_Api_Spec(source *storage.Api_Spec) error {
 
 	// APIVersion
-	if source.APIVersion != nil {
-		apiVersion := *source.APIVersion
-		api.APIVersion = &apiVersion
-	} else {
-		api.APIVersion = nil
-	}
+	api.APIVersion = genruntime.ClonePointerToString(source.APIVersion)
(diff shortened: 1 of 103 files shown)


v2/api/apimanagement/v1api20220801/api_types_gen.go
```diff
@@ -905,12 +890,7 @@ func (api *Api_Spec) AssignProperties_From_Api_Spec(source *storage.Api_Spec) er
 	}
 
 	// ApiVersionDescription
-	if source.ApiVersionDescription != nil {
-		apiVersionDescription := *source.ApiVersionDescription
-		api.ApiVersionDescription = &apiVersionDescription
-	} else {
-		api.ApiVersionDescription = nil
-	}
+	api.ApiVersionDescription = genruntime.ClonePointerToString(source.ApiVersionDescription)
```
matthchr, line 893: We could probably write a more generic `ClonePtr` function, using generics now.
```
func ClonePointer[T any](ptr *T) *T {
	if ptr == nil {
		return nil
	}

	result := *ptr
	return &result
}
```

matthchr (approved):

v2/api/apimanagement/v1api20220801/api_types_gen.go
theunrepentantgeek: Good idea for a followup PR.

PR #4683, opened 2025-04-10
Restructure new resource docs for clarity

theunrepentantgeek:
## What this PR does

Splits our existing documentation that describes how to add a new resource into a set of simpler pages to walk potential contributors through the process step by step.

gif

## Checklist


docs/hugo/content/contributing/add-a-new-code-generated-resource/before-you-begin.md
```diff
@@ -0,0 +1,98 @@
...
+To unambigously identify the resource you want to add, you need to know its _**group**_, _**version**_, and _**kind**_ (aka GVK).
+
+We work out these by reviewing the [Azure REST API specs](https://github.com/Azure/azure-rest-api-specs/tree/main/specification) repo. We will walk through an example of adding Azure Synapse Workspace.
+
+### Kind
+
+Also known as the _name_ of the resource, you usually know this going in. In our example above, the name of the resource is `workspace`. 
+
+If you're not sure, one approach is to look in the Swagger/OpenAPI specification file for the service and find the documented PUT for the resource you're interested in. The resource name will be the
+[second to last section of the URL](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/synapse/resource-manager/Microsoft.Synapse/stable/2021-06-01/workspace.json#L71).
```
davefellows, line 24: "second to last section" isn't very clear to me. Isn't it the json filename at the end of the URL?

docs/hugo/content/contributing/add-a-new-code-generated-resource/before-you-begin.md
```diff
@@ -0,0 +1,98 @@
...
+
+Another approach is to search for the ARM documentation of the resource. Including the search terms `Azure` `ARM` and `Bicep` will help you find the right documentation. For example, we can search for [`Azure ARM Bicep Synapse Workspaces`](https://www.google.com/search?q=azure+arm+bicep+synapse+workspaces&oq=azure+arm+bicep+synapse+workspaces) and (at least at the time of writing) the first result is the [Azure Synapse Workspace](https://learn.microsoft.com/en-us/azure/templates/microsoft.synapse/workspaces?pivots=deployment-language-arm-template) resource.
+
+**Note**: ARM almost exclusively uses plural names for resources. ASO uses singular forms, aligning with the usual practices in Kubernetes. For example, the Azure Synapse Workspace resource is called `workspaces` in ARM and `workspace` in ASO.
+
+### Group
+
+This is named after the Azure service, for example `resources` or `documentdb`.
```
davefellows, line 32: Is "resources" a good example given it's not a service?

docs/hugo/content/contributing/add-a-new-code-generated-resource/before-you-begin.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Development Environment
+
+With the resource identified, you should ensure your [development environment]({{< relref "developer-setup" >}}) is set up to add the resource.
+
+While it _**is**_ possible to do this using just `go`, you end up doing a lot of manual work that we've automated using `task` and other tools.
+
+The recommended approaches are 
+
+* Use our .devcontainer, either on [Linux]({{< relref "developer-setup" >}}#dev-container-with-vs-code-on-linux) or [using WSL on Windows]({{< relref "developer-setup" >}}#dev-container-with-vs-code-on-windows)
```
davefellows, line 55: What about Mac?

docs/hugo/content/contributing/add-a-new-code-generated-resource/run-the-code-generator.md
```diff
@@ -0,0 +1,224 @@
...
+The code generator is configured using the [v2/azure-arm.yaml](https://github.com/Azure/azure-service-operator/blob/main/v2/azure-arm.yaml) configuration. This is a _large_ YAML file and it can be daunting to work with.
+
+**Request:** Please don't apply any prettification tools to the file, as the flurry of changes will make reviewing your new resource much harder. There is wide variety in the preferences of YAML formatting - if you want to tidy up the file, please create an issue and we'll discuss it there.
+
+To add a new resource to this file, find the `objectModelConfiguration` section of the file.
+
+Then, find the configuration for the `group` you want; if it's not there, create a new one, inserting it into the existing list in alphabetical order. Within the group, find the `version` you want; again, create a new one if it's not already there, and keep things in numerical order as you do so.
+
+Add your new resource to the list for that version, including the directive `$exportAs: <name>` nested beneath.
+You must also include the `$supportedFrom:` annotation. This should be the _next release_ which will contain support
```
davefellows, line 21: What do you mean by "contain support"?

davefellows: Oh, you mean the next ASO release that will include the new resource?

docs/hugo/content/contributing/add-a-new-code-generated-resource/run-the-code-generator.md
```diff
@@ -0,0 +1,224 @@
...
+ 2021-06-01:
+   Workspace:
+     $exportAs: Workspace
+     $supportedFrom: v2.0.0
+```
+
+**Tip:** The `$exportAs` directive allows the name of the resource to be changed. We typically use this to simplify the name, relying on the **group** to disambiguate things. For example, if the original resource name was `SynapseWorkspace` we'd prefer to use `Workspace` to avoid stuttering.
+
+If ASO was already configured to generate resources from this group (or version), you will need to add your new
+configuration around the existing values.
```
davefellows, line 48: Not sure what this means

docs/hugo/content/contributing/add-a-new-code-generated-resource/run-the-code-generator.md
```diff
@@ -0,0 +1,224 @@
...
+Replace cross-resource references with genruntime.ResourceReference:  
+["github.com/Azure/azure-service-operator/v2/apis/containerservice/v1api20210501/PrivateLinkResource.Id" 
+looks like a resource reference but was not labelled as one.
+```
+
+To fix this, determine whether the property in question is an ARM ID or not, and then update the `objectModelConfiguration` section in the configuration file, as follows:
+
+* Find the section you added earlier
+* Add a new object if it's not already present
+* Adding your property with an `$referenceType:` declaration nested below.
```
davefellows, line 98:
```suggestion
* Adding your property with a `$referenceType:` declaration nested below.
```

docs/hugo/content/contributing/add-a-new-code-generated-resource/run-the-code-generator.md
```diff
@@ -0,0 +1,224 @@
...
+
+If the property is an ARM ID, use `$referenceType: arm` to flag that property as a reference:
+
+```yaml
+network:
+  2020-11-01:
+    NetworkSecurityGroup:
+      $export: true
```
davefellows, line 106: Is this the same configuration file referenced above? There are different properties listed here (no exportas, supportfrom and has export) which might be confusing.

docs/hugo/content/contributing/add-a-new-code-generated-resource/run-the-code-generator.md
```diff
@@ -0,0 +1,224 @@
...
+type DodgyResource: $exportAs: ReputableResource not consumed
+```
+
+It's possible the submodule `v2/specs/azure-rest-api-specs` is out of date. Try running `git submodule update --init --recursive` to update the submodule. If this fixes your issue, please submit a separate PR to update the submodule by itself. (Submodule updates often include documentation updates to existing resources, and it's worth keeping those updates separate from new reources).
+
+## Debugging
+
+Sometimes it is useful to see what each stage of the generator pipeline has produced and/or changed. 
+
+To write detailed debug logs detailing internal stage after each stage of the pipeline has run, use the `--debug` flag to specify which type definitions to include.
```
davefellows, line 190: State?
```suggestion
To write detailed debug logs detailing internal state after each stage of the pipeline has run, use the `--debug` flag to specify which type definitions to include.
```

docs/hugo/content/contributing/add-a-new-code-generated-resource/_index.md
```diff
@@ -0,0 +1,24 @@
+---
+title: Adding a new code generated resource to ASO v2
+linktitle: Add a resource
+layout: single
+---
+
```
davefellows, line 6: We should describe the overall process and outputs somewhere e.g. what does the code generator actually generate? Not sure whether that's best here or in one of the first to guides below...

docs/hugo/content/contributing/add-a-new-code-generated-resource/review-the-generated-resource.md
```diff
@@ -0,0 +1,127 @@
...
+
+There is a section in the config detailing a number of these errors which have been fixed. Look for "Deal with properties that should have been marked readOnly but weren't".
+
+Here's an example, where DocumentDB missed setting the `provisioningState` property to read-only. We override it in the config:
+
+```yaml
+  - group: documentdb
+    name: Location  # This type is subsequently flattened into NamespacesTopics_Spec
+    property: ProvisioningState
+    remove: true
```
davefellows, line 51: I think this needs a bit more explaination i.e. why are we removing it rather than marking it as readonly or something? I assume it's because readonly properties shouldn't be included in the "Spec" as that's what's used to create the new resource. Some more context eplaining the difference between Spec and Status and how they're used by ASO would probably help. I think this ties into my high-level comment in the overview page about providing more context on what the generator actually produces etc.

docs/hugo/content/contributing/add-a-new-code-generated-resource/review-the-generated-resource.md
```diff
@@ -0,0 +1,127 @@
...
+
+Since these properties are manually configured, they must also be retrieved from Azure manually. This is done using
+the extension framework supported by the operator. Use the [documentdb secrets extension](https://github.com/Azure/azure-service-operator/blob/main/v2/api/documentdb/customizations/database_account_extensions.go)
+as a template for authoring an extension for your resource. Most extensions go into the `customizations` package found at the group level.
+
+<!-- TODO: Do we have a link to give more details about extensions and how they work? -->
+
+## Check: consumption of secrets
+
+Some resources take secrets as input from the user, for example when creating a virtual machine the user may supply a `pasword`.
```
davefellows, line 89:
```suggestion
Some resources take secrets as input from the user, for example, when creating a virtual machine the user may supply a `pasword`.
```

docs/hugo/content/contributing/add-a-new-code-generated-resource/review-the-generated-resource.md
```diff
@@ -0,0 +1,127 @@
...
+This is a `string` field but because it represents a secret value we don't want to just stick it on the `Spec` as a string.
+It needs to be a `genruntime.SecretReference` (a reference to a secret) instead, so that users can safely supply their
+secret to the operator without it being accessible via plain-text.
+
+In most cases, the Swagger/OpenAPI document will annotate these properties with an extension `x-ms-secret`, which indicates
+that value is a secret. The ASO code-generator uses that annotation to automatically transform the annotated field to
+a `genruntime.SecretReference`.
+
+In some cases, the resources Swagger specification is missing this annotation. If that happens, use an override in
+our `azure-arm.yaml` to force a particular property to be treated as a secret. For example, under `synapse`:
```
davefellows, line 99: Provide link to this file?

docs/hugo/content/contributing/add-a-new-code-generated-resource/review-the-generated-resource.md
```diff
@@ -0,0 +1,127 @@
...
+
+```yaml
+WorkspaceProperties:
```
davefellows, line 102: The naming here, is this some convention where it's <resource>Properties? Or is this embedded within the workspace resource? Maybe clarify this.

docs/hugo/content/contributing/add-a-new-code-generated-resource/write-a-test.md
```diff
@@ -0,0 +1,49 @@
...
+title: Write a test for the resource
+linktitle: Write a test
+weight: 40
+---
+
+Every new resource should have a handwritten test as there is always the possibility of a resource behaving in a subtly different way than expected. This also applies to _new versions_ of existing resources.
+
+All these tests live in the [`v2/internal/controllers`](https://github.com/Azure/azure-service-operator/tree/main/v2/internal/controllers). There's a readme in that folder that describes the test structure and naming conventions.
+
+For each test there's a recording in the `recordings` folder that records the HTTP interactions between the test and Azure allowing for rapid replay for future runs.
```
davefellows, line 11: Maybe briefly describe the vcr/recording stuff (and/or link to another doc about it) as you mention "recording" here but no prior context?

docs/hugo/content/contributing/add-a-new-code-generated-resource/write-a-test.md
```diff
@@ -0,0 +1,49 @@
...
+
+One good approach is to start from an [existing test](https://github.com/Azure/azure-service-operator/blob/main/v2/internal/controllers/documentdb_mongodb_crud_v20231115_test.go) and modify it to work for your resource. It can also be helpful to refer to examples in the [ARM templates GitHub repo](https://github.com/Azure/azure-quickstart-templates).
+
+## Resource extensions
```
davefellows, line 15: Should this be in one of the other sections rather than testing? Or have I misunderstood (it sounds like this is modifying the resources behaviour, not how it should be test).

docs/hugo/content/contributing/add-a-new-code-generated-resource/write-a-test.md
```diff
@@ -0,0 +1,49 @@
...
+
+Other extension points allow blocking reconciliation attempts until a prerequisite is met, additional API calls after a successful reconciliation, or even modifying the resource before it is sent to Azure.
+
+For most resource extensions, any implementation will go in the `customizations` folder generated for each resource group.
+
+Reach out to us if you want some guidance on which (if any) extension points are appropriate for your resource.
+
+## Record the test passing
+
+When the test passes, you'll find a new recording file in the `recordings` folder. When you modify your test, you'll need to delete the recording file so that the test runs against the real Azure API.
```
davefellows, line 29: "When you modify your test, you'll need to delete the recording file"

Not related to these docs, but could this be automated/detected somehow in the future?

docs/hugo/content/contributing/add-a-new-code-generated-resource/write-a-test.md
```diff
@@ -0,0 +1,49 @@
...
+
+Once you're finished, include the recording file in your PR. This is important because it allows other developers to run the test without needing to make real API calls to Azure.
+
+See [the code generator test README](../testing/#running-envtest-integration-tests) for more information on how to run tests and record their HTTP interactions to allow replay.
+
+## Test volumes
+
+These integration tests can take a few minutes to run from recordings, and we're starting to see issues due to the large number of resources we're testing. 
+
+If you are adding a new version of an already supported resource, we may request that you remove tests for older versions of the resource as a part of your PR.
```
davefellows, line 39: Is that a good idea if the previous versions are still in use and someone in a service team mucks something up and introduces a breaking change?

docs/hugo/content/contributing/add-a-new-code-generated-resource/write-a-test.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Record the test passing
+
+When the test passes, you'll find a new recording file in the `recordings` folder. When you modify your test, you'll need to delete the recording file so that the test runs against the real Azure API. 
+
+Once you're finished, include the recording file in your PR. This is important because it allows other developers to run the test without needing to make real API calls to Azure.
+
+See [the code generator test README](../testing/#running-envtest-integration-tests) for more information on how to run tests and record their HTTP interactions to allow replay.
+
+## Test volumes
```
davefellows, line 35: I initially thought this was about testing volumes as in drives :)
Minor as it's obvious once you read the section but maybe:
```suggestion
## Test volumes and runtimes
```

docs/hugo/content/contributing/add-a-new-code-generated-resource/write-a-test.md
```diff
@@ -0,0 +1,49 @@
+---
+title: Write a test for the resource
+linktitle: Write a test
+weight: 40
+---
+
+Every new resource should have a handwritten test as there is always the possibility of a resource behaving in a subtly different way than expected. This also applies to _new versions_ of existing resources.
```
davefellows, line 7: Is only one main create test required?

docs/hugo/content/contributing/add-a-new-code-generated-resource/create-a-sample.md
```diff
@@ -0,0 +1,37 @@
...
+
+We mandate that every resource has a sample that demonstrates how to use it - and enforce this with checks in our CI pipeline. Good examples help other ASO users understand how to use each resource. 
+
+To ensure that each sample is correct, we run them - allowing us to catch any typos or errors before they trip up users in the wild.
+
+The samples are located in the [samples directory](https://github.com/Azure/azure-service-operator/blob/main/v2/samples). There should be at least one sample for each kind of supported resource. These currently need to be added manually. It's possible in the future we will automatically generate samples similar to how we automatically generate CRDs and types, but that doesn't happen today.
+
+The added new sample needs to be tested and recorded.
+
+If a recording for the test already exists, delete it. Look in the [recordings directory](https://github.com/Azure/azure-service-operator/blob/main/v2/internal/controllers/recordings/Test_Samples_CreationAndDeletion) for a file with the same name as your new test.
```
davefellows, line 15: How would this exist if it's a new sample? Or is the same test as in the previous step used to test the sample?

docs/hugo/content/contributing/add-a-new-code-generated-resource/create-a-sample.md
```diff
@@ -0,0 +1,37 @@
...
+To ensure that each sample is correct, we run them - allowing us to catch any typos or errors before they trip up users in the wild.
+
+The samples are located in the [samples directory](https://github.com/Azure/azure-service-operator/blob/main/v2/samples). There should be at least one sample for each kind of supported resource. These currently need to be added manually. It's possible in the future we will automatically generate samples similar to how we automatically generate CRDs and types, but that doesn't happen today.
+
+The added new sample needs to be tested and recorded.
+
+If a recording for the test already exists, delete it. Look in the [recordings directory](https://github.com/Azure/azure-service-operator/blob/main/v2/internal/controllers/recordings/Test_Samples_CreationAndDeletion) for a file with the same name as your new test.  
+
+Typically these are named `Test_<GROUP>_<VERSION_PREFIX>_CreationAndDeletion.yaml`.
+For example, if we're adding sample for NetworkSecurityGroup resource, check for `Test_Network_v1beta_CreationAndDeletion.yaml`
```
davefellows, line 18: Is this name correct: `Test_Network_v1beta_CreationAndDeletion.yaml`?
I wouldn't have thought "NetworkSecurityGroup" was called "Network"

docs/hugo/content/contributing/add-a-new-code-generated-resource/create-a-sample.md
```diff
@@ -0,0 +1,37 @@
...
+
+Run the test and record it using `task`:
+
+``` bash
+TEST_FILTER=Test_Samples_CreationAndDeletion task controller:test-integration-envtest
```
davefellows, line 23: Is there a way to run just the test for the sample that's been added rather than all of them?

docs/hugo/content/contributing/add-a-new-code-generated-resource/create-a-sample.md
```diff
@@ -0,0 +1,37 @@
...
+For example, to give your test a 60m timeout, use:
+
+``` bash
+TIMEOUT=60m TEST_FILTER=Test_Samples_CreationAndDeletion task controller:test-integration-envtest
+```
+
+Test replay runs at a faster rate than real Azure API calls, so the TIMEOUT is only needed for recording.
+
+Once you have a successful recording, _test it_ by running the same command again. It should run quickly and pass. If that doesn't happen, you may need to delete the recording file again and re-record it. If problems persist, contact us for help.
+
```
davefellows, line 37: Any final guidance on what to do now? Presumably it's just push the PR. Maybe just a quick comment saying you're done, you can now submit your PR.

docs/hugo/content/contributing/add-a-new-code-generated-resource/write-a-test.md
```diff
@@ -0,0 +1,49 @@
...
+
+All these tests live in the [`v2/internal/controllers`](https://github.com/Azure/azure-service-operator/tree/main/v2/internal/controllers). There's a readme in that folder that describes the test structure and naming conventions.
+
+For each test there's a recording in the `recordings` folder that records the HTTP interactions between the test and Azure allowing for rapid replay for future runs.
+
+One good approach is to start from an [existing test](https://github.com/Azure/azure-service-operator/blob/main/v2/internal/controllers/documentdb_mongodb_crud_v20231115_test.go) and modify it to work for your resource. It can also be helpful to refer to examples in the [ARM templates GitHub repo](https://github.com/Azure/azure-quickstart-templates).
```
matthchr, line 13: minor: Maybe worth calling out the philosophy of doing as much in parallel as possible in the test, for example in the test you link:
```
	// Create both resources in Azure
	tc.CreateResourcesAndWait(&acct, &db)
```

and the user of `RunParallelSubtests`, to simulate what happens in reality (user applies YAML with everything).

docs/hugo/content/contributing/add-a-new-code-generated-resource/write-a-test.md
matthchr: Maybe rephrase this to "removing old tests" or something?

I think the title is weird too, not sure "and runtimes" really clarifies it.

docs/hugo/content/contributing/add-a-new-code-generated-resource/create-a-sample.md
matthchr: Good point of clarification. There's a single go test for each subdirectory in the samples dir, so all resources for a given API version of say, network, are tested in one recording file.

This is because a lot of resources have dependencies between them, and trying to have each _resource_ have its own recording would result in a combinatoric set of tests/recordings.

docs/hugo/content/contributing/add-a-new-code-generated-resource/create-a-sample.md
matthchr: As above, its' for the whole group+version (group being network in this case)

docs/hugo/content/contributing/add-a-new-code-generated-resource/create-a-sample.md
matthchr: We use subtests so I think you can give the name of the subtest and it'll work?

It comes up a lot, we should document it - I often just comment the others out in the samples test code w/ a oneliner, but a cleaner way would be ideal to document.

docs/hugo/content/contributing/add-a-new-code-generated-resource/write-a-test.md
theunrepentantgeek: Broken out into separate step - and can be expanded in future with more details.

docs/hugo/content/contributing/add-a-new-code-generated-resource/write-a-test.md
theunrepentantgeek: I'm not sure how. From the perspective of the test runner, it can't tell the difference between "the test is failing because something is broken" and "the test is failing because the recording is stale".

docs/hugo/content/contributing/add-a-new-code-generated-resource/write-a-test.md
theunrepentantgeek: Rephrased.

docs/hugo/content/contributing/add-a-new-code-generated-resource/write-a-test.md
theunrepentantgeek: We're reluctant to remove older tests for exactly this reason, but our CI times are starting to climb unreasonably so we need to do something.

docs/hugo/content/contributing/add-a-new-code-generated-resource/before-you-begin.md
matthchr: I think the container itself should be fine, because they build it themselves on their arch so it should be ARM64.

The issue (and I do also recall Harsh having one) would have to do with any upstream binaries we directly get that don't support ARM. I'm not sure what all binaries don't support ARM at this point though - the Kubernetes ones via envtest all should AFAIK, and ASO when built locally will as well.

docs/hugo/content/contributing/add-a-new-code-generated-resource/write-a-test.md
matthchr: FWIW the replay tests when run in replay mode aren't the best for protecting against service teams making breaking changes anyway because we'll always replay the old (non-broken) payloads.

Removing old tests does suck too, but 1 test per API per resource for all of Azure is a lot of tests. As we support more and more resources and versions something has to give in the matrix.

docs/hugo/content/contributing/add-a-new-code-generated-resource/review-the-generated-resource.md
```diff
@@ -0,0 +1,127 @@
...
+          - PrimaryReadonlyMasterKey
+          - SecondaryReadonlyMasterKey
+          - DocumentEndpoint
+```
+
+Since these properties are manually configured, they must also be retrieved from Azure manually. This is done using
+the extension framework supported by the operator. Use the [documentdb secrets extension](https://github.com/Azure/azure-service-operator/blob/main/v2/api/documentdb/customizations/database_account_extensions.go)
+as a template for authoring an extension for your resource. Most extensions go into the `customizations` package found at the group level.
+
+<!-- TODO: Do we have a link to give more details about extensions and how they work? -->
```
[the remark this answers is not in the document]
matthchr: I do think we should remove this before merge.

docs/hugo/content/contributing/add-a-new-code-generated-resource/write-a-test.md
```diff
@@ -0,0 +1,56 @@
...
+* the latest `preview` version of the resource.
+
+In most cases you won't need to worry about this.
+
+### Prefer CreateResourcesAndWait
+
+If your rest needs to create multiple resources, prefer using `CreateResourcesAndWait` (note the plural `Resources`) to create all the resources at once, rather than calling `CreateResourceAndWait` (singular `Resource`) once for each resource in turn.
+
+When a user applies a YAML file containing multiple resources, ASO will be expected to handle creating all the resources in the correct order. Simulating this in the test by calling `CreateResourcesAndWait` to create all the resources at once is a good way to ensure that the test is realistic.
+### Prefer Parallel tests
```
matthchr, line 43:
```suggestion

### Prefer Parallel tests
```

docs/hugo/content/contributing/add-a-new-code-generated-resource/before-you-begin.md
```diff
@@ -8,45 +8,67 @@ weight: 10
 
 The first step in adding a new code-generated resource is to determine which resource you want to add. 
 
-Any ARM resource can be generated. If you're not sure if the resource you are interested in is an ARM resource, check if it is defined in a `resource-manager` folder in the [Azure REST API specs](https://github.com/Azure/azure-rest-api-specs/tree/main/specification) repo.
+Any ARM resource can be generated, provided you know the Azure type of the resource, and the version of the API you want to use.
 
-If it is, it's an ARM resource.
+To unambiguously identify the resource you want to add, you need to know the Azure type, and the version of the API you want to use. We will walk through an example of adding Azure Synapse Workspace.
+### Find the Azure type of the resource
```
matthchr, line 14:
```suggestion

### Find the Azure type of the resource
```

docs/hugo/content/contributing/add-a-new-code-generated-resource/implement-extensions.md
```diff
@@ -0,0 +1,22 @@
+---
+title: Customize the behaviour of generated resource
```
matthchr, line 2: Do we want some (if needed) suffix here?

Or maybe just a disclaimer at the start that many resources don't need any extensions defined?

docs/hugo/content/contributing/add-a-new-code-generated-resource/before-you-begin.md
```diff
@@ -0,0 +1,120 @@
...
+The Azure type of a resource consists of a resource provider and a name. For example, the Azure type of a Synapse Workspace is `Microsoft.Synapse/workspaces` - the resource provider is `Microsoft.Synapse` and the name is `workspaces`.
+
+If you're not sure of the Azure type for the resource you want to add, one approach is to search for an ARM or Bicep template that deploys the resource. These templates state the Azure type and version of the resource near the top.
+
+E.g. for an ARM template:
+``` json
+{
+  "type": "Microsoft.Synapse/workspaces",
+  "apiVersion": "2021-06-01",
+  ...
```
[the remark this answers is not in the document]
theunrepentantgeek: Added.

docs/hugo/content/contributing/add-a-new-code-generated-resource/before-you-begin.md
```diff
@@ -0,0 +1,120 @@
...
+For the resource you're adding, the GVK is derived from the Azure type and version, as follows:
+
+* **group**: Based on the resource provider, with the `Microsoft.` prefix removed, and lowercased. For example, `Microsoft.Synapse` becomes `synapse`.
+* **version**: The constant prefix `v1api` followed by the Azure `api-version`, with non-alphanumeric characters removed. For example, `2021-06-01` becomes `v1api20210601` and `2021-04-01-preview` becomes `v1api20210401preview`.
+* **kind**: The Azure type, with the resource provider prefix removed and the plural form converted to singular. For example, `Microsoft.Synapse/workspaces` becomes `workspace`.
+
+
+To illustrate, here are some examples of GVKs for common resources:
+
+| Azure Type | API Version | Group | Version | Kind |
```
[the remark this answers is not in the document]
theunrepentantgeek: Done.

docs/hugo/content/contributing/add-a-new-code-generated-resource/before-you-begin.md
```diff
@@ -0,0 +1,120 @@
...
+* Use our .devcontainer, either on [Linux]({{< relref "developer-setup" >}}#dev-container-with-vs-code-on-linux) or [using WSL on Windows]({{< relref "developer-setup" >}}#dev-container-with-vs-code-on-windows)
+* Install the tools locally on your machine, using the CLI under either [MacOS]({{< relref "developer-setup" >}}#cli-on-macos) or [Linux]({{< relref "developer-setup" >}}#cli-on-linux).
+
+The [Linux CLI]({{< relref "developer-setup" >}}#cli-on-linux) instructions also work for WSL on Windows.
+
+## Check for Repository Tags
+
+Once you've forked and cloned the ASO repository, make sure you have cloned the tags by running this command:
+
+```bash
```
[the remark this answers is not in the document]
theunrepentantgeek: I'll fix occurrences in these files, leaving the others for later.

docs/hugo/content/contributing/add-a-new-code-generated-resource/before-you-begin.md
```diff
@@ -0,0 +1,120 @@
...
+git submodule init
+```
+
+and then checkout the submodule with
+
+``` bash
+git submodule update
+```
+
+**Tip**: If you find you need to update the submodule to a later version as a part of creating your new resource, please create a separate PR for that. These updates often pull in documentation changes to existing resources and keeping those changes separate makes it easier to review and merge the updates.
```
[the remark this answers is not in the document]
theunrepentantgeek: Good idea.

docs/hugo/content/contributing/add-a-new-code-generated-resource/implement-extensions.md
theunrepentantgeek: Added disclaimer.

matthchr (approved):

docs/hugo/content/contributing/add-a-new-code-generated-resource/before-you-begin.md
```diff
@@ -0,0 +1,98 @@
...
+title: Before you begin adding a new code-generated resource
+linktitle: Before you begin
+weight: 10
+---
+
+## Determine a resource to add
+
+The first step in adding a new code-generated resource is to determine which resource you want to add. 
+
+Any ARM resource can be generated. If you're not sure if the resource you are interested in is an ARM resource, check if it is defined in a `resource-manager` folder in the [Azure REST API specs](https://github.com/Azure/azure-rest-api-specs/tree/main/specification) repo.
```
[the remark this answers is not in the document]
theunrepentantgeek: Clarified.

matthchr (approved):

Issue #4600, opened 2025-02-26
Feature: Let ASO exclude namespaces from its watchlist


antonioquintavalle1A:
_Describe the feature request._

Today we can specify the a of namespaces ASO actively watches through `AZURE_TARGET_NAMESPACES`.

In dynamic environments, it can be rather hard to retrieve an exhaustive list of namespaces but it might be easier to rely on an exclusion list instead.

The scope of this feature request is to discuss if there's a way to let ASO excluding namespaces to be watched, either with an environment variable either with any specific annotation on the namespace.

antonioquintavalle1A:
Gooday @matthchr,
the usage we're envisaging is the following:

1 - A platform team needs to use ASO to create some infrastructure on Azure to support lifecyle of cluster and all its internal components (mostly for storage class / keyvaults). This team is made of cluster admins and ASO will be deployed on all our  clusters at the same time.
2 - Application teams instead, want to deploy a set of ASO resources that are needed just for their business needs. They will deploy these CRs on a subset of all the existing clusters with their own credentials.

In order to better distinguish the two scenarios, we are thinking of providing an instance of ASO on ALL the clusters, which is going under the responsibility of Plaform team. 
Such instance of ASO can be configured with an exhaustive list of namespace, through `AZURE_TARGET_NAMESPACES` containing the list of all the namespaces in the interests of platform team.

On the other hand, for application teams we would like to have another dedicated instance of ASO, which is watching all the other namespaces but not the ones under control of platform team (here the need of the exclusion list).

This is required for different reasons:
1 - Isolation: number of CRs handled by the 2 different instances of ASO is different and we do not want that application can compromise the ASO running platform business.
2 - We would like to switch off the application related ASO to support active/passive scenarios.

For the active/passive scenario, please note that the suggested approach to use `manage` on CRs of active cluster and `skip` on the passive clusters has a double side effect:
1 - It violates a bit the gitops approach as we need to find a way to send 2 different versions of manifest to active (annotated with `manage`) and passive (annotated with `skip`) clusters.
2 - More importantly, we are not interested in making any read on the passive clusters, which could lead us to exceed quota limits on a given subscription (here we are talking about hundreds of applicative CRs; having all of them on an active cluster might be already challenging, considering the low API rate limits).

Of course this is a tentative to push forward an active/passive scenario, maybe there are other ways to achieve the same results and I'm interested in bringing forward such discussion if you don't mind.

matthchr:
[...]

PR #4692, opened 2025-04-15
Let ASO use reconcile policy defined at namespace level for all its r…

antonioquintavalle1A:
…esources (#4600)

## What this PR does

This addressed the discussion happening in here https://github.com/Azure/azure-service-operator/issues/4600.
With this change we can specify a reconcile policy at namespace level so that all the resources will be handled according to that one, even if another value is explicitly set at resource level.

Closes #[4600]

- [x] this PR contains tests

### DIFF
--- a/v2/internal/reconcilers/generic/generic_reconciler.go
+++ b/v2/internal/reconcilers/generic/generic_reconciler.go
@@ -18,6 +18,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -234,8 +235,7 @@ func (gr *GenericReconciler) createOrUpdate(ctx context.Context, log logr.Logger
 		return ctrl.Result{}, err
 	}
 
-	// Check the reconcile-policy to ensure we're allowed to issue a CreateOrUpdate
-	reconcilePolicy := reconcilers.GetReconcilePolicy(metaObj, log, gr.Config.DefaultReconcilePolicy)
+	reconcilePolicy := gr.mergeReconcilePolicy(ctx, log, metaObj)
 	if !reconcilePolicy.AllowsModify() {
 		return ctrl.Result{}, gr.handleSkipReconcile(ctx, log, metaObj)
 	}
@@ -247,7 +247,7 @@ func (gr *GenericReconciler) createOrUpdate(ctx context.Context, log logr.Logger
 
 func (gr *GenericReconciler) delete(ctx context.Context, log logr.Logger, metaObj genruntime.MetaObject) (ctrl.Result, error) {
 	// Check the reconcile policy to ensure we're allowed to issue a delete
-	reconcilePolicy := reconcilers.GetReconcilePolicy(metaObj, log, gr.Config.DefaultReconcilePolicy)
+	reconcilePolicy := gr.mergeReconcilePolicy(ctx, log, metaObj)
 	if !reconcilePolicy.AllowsDelete() {
 		log.V(Info).Info("Bypassing delete of resource due to policy", "policy", reconcilePolicy)
 		controllerutil.RemoveFinalizer(metaObj, genruntime.ReconcilerFinalizer)
@@ -351,7 +351,8 @@ func (gr *GenericReconciler) CommitUpdate(
 }
 
 func (gr *GenericReconciler) handleSkipReconcile(ctx context.Context, log logr.Logger, obj genruntime.MetaObject) error {
-	reconcilePolicy := reconcilers.GetReconcilePolicy(obj, log, gr.Config.DefaultReconcilePolicy) // TODO: Pull this whole method up here
+	reconcilePolicy := gr.mergeReconcilePolicy(ctx, log, obj)
+
 	log.V(Status).Info(
 		"Skipping creation/update of resource due to policy",
 		annotations.ReconcilePolicy, reconcilePolicy)
@@ -381,4 +382,37 @@ func (gr *GenericReconciler) writeReadyConditionErrorOrDefault(ctx context.Conte
 	log.Error(readyErr, "Encountered error impacting Ready condition")
 	err = gr.WriteReadyConditionError(ctx, log, metaObj, readyErr)
 	return err
+}
+
+func (gr *GenericReconciler) mergeReconcilePolicy(ctx context.Context, log logr.Logger, obj genruntime.MetaObject) annotations.ReconcilePolicyValue {
+	// We initially get the reconcile policy from the object itself
+	source := "default" // assume the source is the default policy for now - this source field is used only for logging purposes
+	policyStr := obj.GetAnnotations()[annotations.ReconcilePolicy]
+
+	// If the policy is not defined at object level, then we check if it's defined at namespace level
+	if policyStr == "" {
+		namespaceObject, err := gr.KubeClient.GetObject(ctx, types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetNamespace()}, schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Namespace"})
+		// if we cannot get the namespace, we return default reconcile policy
+		if err != nil {
+			log.V(Verbose).Info("Error while retrieving namespace object", "error", err)
+			return gr.Config.DefaultReconcilePolicy // return default in case of error
+		}
+		policyStr = namespaceObject.GetAnnotations()[annotations.ReconcilePolicy]
+		if policyStr != "" {
+			source = "namespace"
+		}
+	} else {
+		source = "object" // used to track where the policy was taken from for logging purposes
+	}
+
+	reconcilePolicy, err := reconcilers.ParseReconcilePolicy(policyStr, gr.Config.DefaultReconcilePolicy)
+	if err != nil {
+		log.Error(
+			err,
+			"failed to get reconcile policy. Applying default policy instead",
+			"chosenPolicy", reconcilePolicy,
+			"policyAnnotation", policyStr)
+	}
+	log.V(Verbose).Info("Retrieved reconcile policy", "policy", reconcilePolicy, "source", source)
+	return reconcilePolicy
 }
--- a/v2/internal/reconcilers/generic/register.go
+++ b/v2/internal/reconcilers/generic/register.go
@@ -10,17 +10,24 @@ import (
 	"fmt"
 
 	. "github.com/Azure/azure-service-operator/v2/internal/logging"
+	"github.com/Azure/azure-service-operator/v2/internal/reconcilers"
 
 	"github.com/go-logr/logr"
 	"github.com/rotisserie/eris"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/conversion"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
 	ctrlbuilder "sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/Azure/azure-service-operator/v2/internal/config"
 	"github.com/Azure/azure-service-operator/v2/internal/util/interval"
@@ -94,6 +101,7 @@ func RegisterAll(
 
 	var errs []error
 	for _, obj := range objs {
+		options.LogConstructor(nil).V(Info).Info("Registering", "objectName", obj.Name)
 		// TODO: Consider pulling some of the construction of things out of register (gvk, etc), so that we can pass in just
 		// TODO: the applicable extensions rather than a map of all of them
 		if err := register(mgr, kubeClient, positiveConditions, obj, options); err != nil {
@@ -148,6 +156,8 @@ func register(
 		WithOptions(options.Options)
 	builder.Named(info.Name)
 
+	builder = builder.Watches(&corev1.Namespace{}, handler.EnqueueRequestsFromMapFunc(registerNamespaceWatcher(options, kubeClient, info.Obj, gvk)), ctrlbuilder.WithPredicates(reconcilers.ARMReconcilerAnnotationChangedPredicate()))
+
 	for _, watch := range info.Watches {
 		builder = builder.Watches(watch.Type, watch.MakeEventHandler(kubeClient, options.LogConstructor(nil).WithName(info.Name)))
 	}
@@ -158,4 +168,33 @@ func register(
 	}
 
 	return nil
+}
+
+// This registers a watcher on the namespace for the watched resource
+func registerNamespaceWatcher(options Options, kubeclient kubeclient.Client, object client.Object, gvk schema.GroupVersionKind) func(ctx context.Context, obj client.Object) []reconcile.Request {
+	options.LogConstructor(nil).V(Status).Info("Handler for resource object", "name", object.GetName(), "object.namespace", object.GetNamespace(), "object.kind", gvk.Kind)
+	return func(ctx context.Context, obj client.Object) []reconcile.Request {
+		reconcileRequests := []reconcile.Request{}
+
+		aList := &unstructured.UnstructuredList{}
+		aList.SetGroupVersionKind(gvk)
+		if err := kubeclient.List(ctx, aList, &client.ListOptions{Namespace: obj.GetName()}); err != nil {
+			return []reconcile.Request{}
+		}
+		// list the objects for the current kind
+		options.LogConstructor(nil).V(Verbose).Info("Detected namespace reconcile-policy annotation")
+		for _, el := range aList.Items {
+			if _, ok := el.GetAnnotations()["serviceoperator.azure.com/reconcile-policy"]; ok {
+				// If the annotation is defined for the object, there's no need to reconcile it and we skip the object
+				continue
+			}
+			if _, ok := obj.GetAnnotations()["serviceoperator.azure.com/reconcile-policy"]; ok {
+				reconcileRequests = append(reconcileRequests, reconcile.Request{NamespacedName: types.NamespacedName{
+					Name:      el.GetName(),
+					Namespace: el.GetNamespace(),
+				}})
+			}
+		}
+		return reconcileRequests
+	}
 }
--- a/v2/internal/reconcilers/generic/register.go
+++ b/v2/internal/reconcilers/generic/register.go
@@ -10,7 +10,6 @@ import (
 	"fmt"
 
 	. "github.com/Azure/azure-service-operator/v2/internal/logging"
-	"github.com/Azure/azure-service-operator/v2/internal/reconcilers"
 
 	"github.com/go-logr/logr"
 	"github.com/rotisserie/eris"
@@ -30,6 +29,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/Azure/azure-service-operator/v2/internal/config"
+	"github.com/Azure/azure-service-operator/v2/internal/reconcilers"
 	"github.com/Azure/azure-service-operator/v2/internal/util/interval"
 	"github.com/Azure/azure-service-operator/v2/internal/util/kubeclient"
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
--- a/v2/internal/reconcilers/predicates.go
+++ b/v2/internal/reconcilers/predicates.go
@@ -8,12 +8,10 @@ package reconcilers
 import (
 	"reflect"
 
-	"github.com/go-logr/logr"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	"github.com/Azure/azure-service-operator/v2/internal/util/predicates"
 	"github.com/Azure/azure-service-operator/v2/pkg/common/annotations"
-	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
 )
 
 // ARMReconcilerAnnotationChangedPredicate creates a predicate that emits events when annotations
@@ -32,21 +30,6 @@ func ARMPerResourceSecretAnnotationChangedPredicate() predicate.Predicate {
 		map[string]predicates.HasAnnotationChanged{
 			annotations.PerResourceSecret: HasAnnotationChanged,
 		})
-}
-
-// GetReconcilePolicy gets the reconcile-policy from the ReconcilePolicy
-func GetReconcilePolicy(obj genruntime.MetaObject, log logr.Logger, defaultReconcilePolicy annotations.ReconcilePolicyValue) annotations.ReconcilePolicyValue {
-	policyStr := obj.GetAnnotations()[annotations.ReconcilePolicy]
-	policy, err := ParseReconcilePolicy(policyStr, defaultReconcilePolicy)
-	if err != nil {
-		log.Error(
-			err,
-			"failed to get reconcile policy. Applying default policy instead",
-			"chosenPolicy", policy,
-			"policyAnnotation", policyStr)
-	}
-
-	return policy
 }
 
 // HasAnnotationChanged returns true if the annotation has changed
--- a/v2/internal/reconcilers/reconcile_policy.go
+++ b/v2/internal/reconcilers/reconcile_policy.go
@@ -14,8 +14,8 @@ import (
 	"github.com/Azure/azure-service-operator/v2/pkg/common/annotations"
 )
 
-// ParseReconcilePolicy parses the provided reconcile policy.
-// defaultPolicyValue is read from DEFAULT_RECONCILE_POLICY env variable, it set to "manage" when not specified
+// ParseReconcilePolicy parses provided reconcile policy.
+// defaultPolicyValue is read from DEFAULT_RECONCILE_POLICY env variable or set to 'manage' when missing
 func ParseReconcilePolicy(policy string, defaultReconcilePolicy annotations.ReconcilePolicyValue) (annotations.ReconcilePolicyValue, error) {
 	// policy is read from CR annotation, if it's empty it being read from defaultReconcilePolicy
 	switch policy {
--- a/docs/hugo/content/guide/aso-controller-settings-options.md
+++ b/docs/hugo/content/guide/aso-controller-settings-options.md
@@ -309,7 +309,7 @@ RATE_LIMIT_BUCKET_SIZE is the size of the bucket. This value only has an effect 
 
 ### DEFAULT_RECONCILE_POLICY
 
-DEFAULT_RECONCILE_POLICY specify which reconcile strategy to be used by the operator. If not specified, it is set to 'manage'.
+DEFAULT_RECONCILE_POLICY specifies the reconcile strategy to be used by the operator. If not specified, it is set to 'manage'.
 
 **Format:** `string`
 
--- a/v2/api/resources/v1api20200601/permissions.go
+++ b/v2/api/resources/v1api20200601/permissions.go
@@ -13,3 +13,4 @@ package v1api20200601
 // +kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=coordination.k8s.io,resources=leases,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=core,resources=namespaces,verbs=get;list;watch
--- a/v2/internal/reconcilers/reconcile_policy_test.go
+++ b/v2/internal/reconcilers/reconcile_policy_test.go
@@ -40,3 +40,18 @@ func TestParseReconcilePolicy(t *testing.T) {
 	g.Expect(err).Should(HaveOccurred())
 	g.Expect(returnedPolicy).Should(Equal(annotations.ReconcilePolicySkip))
 }
+
+func TestHasReconcilePolicyAnnotationChanged(t *testing.T) {
+	old := "detach-on-delete"
+	new := "skip"
+
+	t.Parallel()
+	g := NewGomegaWithT(t)
+
+	result := HasReconcilePolicyAnnotationChanged(&old, &new)
+	g.Expect(result).Should(Equal(true))
+
+	old = "skip"
+	result = HasReconcilePolicyAnnotationChanged(&old, &new)
+	g.Expect(result).Should(Equal(false))
+}


v2/internal/reconcilers/generic/generic_reconciler.go
```diff
@@ -234,8 +235,32 @@ func (gr *GenericReconciler) createOrUpdate(ctx context.Context, log logr.Logger
...
 	}
 
-	// Check the reconcile-policy to ensure we're allowed to issue a CreateOrUpdate
-	reconcilePolicy := reconcilers.GetReconcilePolicy(metaObj, log, gr.Config.DefaultReconcilePolicy)
+	var reconcilePolicy annotations.ReconcilePolicyValue
+
+	if gr.Config.ReadReconciliationPolicyFromNamespace {
+		log.V(Debug).Info("Looking for namespace annotations")
+		// Claim resource namespace
+		namespaceObject, err := gr.KubeClient.GetObject(ctx, types.NamespacedName{Namespace: metaObj.GetNamespace(), Name: metaObj.GetNamespace()}, schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Namespace"})
```
matthchr, line 243: Doing this launches a new informer for Namespaces, because we don't set `cacheOpts.ReaderFailOnMissingInformer`. I think running a namespace informer here is probably correct... but wondering if, since we're anyway paying the cost of running the namespace informer, we might as well get reconciliation triggered if namespaces are updated.

Doing this correctly requires that we set up a Watch on each resource w/ type namespace, and using the already existing `reconcilers.ARMReconcilerAnnotationChangedPredicate()`, as well as a `handler.EnqueueRequestsFromMapFunc` handler that lists the corresponding resources (in memory from cache) and triggers a reconcile on all of them.

I don't think doing this is all that difficult with the infrastructure we have set up today.

v2/internal/reconcilers/generic/generic_reconciler.go
```diff
@@ -351,12 +401,39 @@ func (gr *GenericReconciler) CommitUpdate(
 }
 
 func (gr *GenericReconciler) handleSkipReconcile(ctx context.Context, log logr.Logger, obj genruntime.MetaObject) error {
-	reconcilePolicy := reconcilers.GetReconcilePolicy(obj, log, gr.Config.DefaultReconcilePolicy) // TODO: Pull this whole method up here
+
+	var reconcilePolicy annotations.ReconcilePolicyValue
```
matthchr, line 405: It feels like this whole section where we read the reconcilePolicy is somewhat duplicated - can we factor into a shared method on `gr` that does the hierarchical lookup, and then reuse that helper function in all the locations?

v2/internal/reconcilers/reconcile_policy.go
```diff
@@ -17,18 +17,27 @@ import (
...
+	// policy is read from CR annotation, if it's empty it being read from defaultReconcilePolicy
+	reconcilePolicy, err := ParseReconcilePolicyNoDefault(policy)
+	if err != nil {
+		return defaultReconcilePolicy, err
+	}
+	return reconcilePolicy, nil
+}
+
+// ParseReconcilePolicy parses the provided reconcile policy.
+// defaultPolicyValue is read from DEFAULT_RECONCILE_POLICY env variable, it set to "manage" when not specified
```
matthchr, line 29: > // defaultPolicyValue is read from DEFAULT_RECONCILE_POLICY env variable, it set to "manage" when not specified

This is not accurate/required on this method.

v2/internal/reconcilers/reconcile_policy.go
```diff
@@ -17,18 +17,27 @@ import (
 // ParseReconcilePolicy parses the provided reconcile policy.
 // defaultPolicyValue is read from DEFAULT_RECONCILE_POLICY env variable, it set to "manage" when not specified
 func ParseReconcilePolicy(policy string, defaultReconcilePolicy annotations.ReconcilePolicyValue) (annotations.ReconcilePolicyValue, error) {
+	// policy is read from CR annotation, if it's empty it being read from defaultReconcilePolicy
+	reconcilePolicy, err := ParseReconcilePolicyNoDefault(policy)
+	if err != nil {
+		return defaultReconcilePolicy, err
+	}
+	return reconcilePolicy, nil
```
matthchr, line 25: Isn't this missing the case where the default policy exists, and the provided policy is `""`?
That shouldn't result in an error it should result in `default, nil`.

v2/internal/reconcilers/generic/generic_reconciler.go
matthchr: Actually I think that's what `reconcilers.GetReconcilePolicy` was trying to do. Maybe delete that and make a method on `gr` that has the hierarchical lookup you want?

matthchr: I think if you do that, `ParseReconcilePolicy` (that honors the default) isn't needed anymore, because that logic is encoded in the `gr` helper instead.

v2/pkg/common/config/config.go
```diff
@@ -101,4 +101,6 @@ const (
 	// DefaultReconcilePolicy allows to change default reconciliation policy to use when serviceoperator.azure.com/reconcile-policy annotation
 	// is not explicitly defined. If omitted, it will be automatically set to "manage"
 	DefaultReconcilePolicy = "DEFAULT_RECONCILE_POLICY"
+	// ReadReconciliationPolicyFromNamespace boolean is used to determine if we need to read reconcile policy from namespace object
+	ReadReconciliationPolicyFromNamespace = "READ_RECONCILIATION_POLICY_FROM_NAMESPACE"
```
matthchr, line 105: I'm not sold we actually need an opt-in for this capability. Can you expand on your logic for why it's opt-in rather than always-on?

theunrepentantgeek (commented):
If `READ_RECONCILIATION_POLICY_FROM_NAMESPACE` is set, this implementation looks _only_ on the namespace for a reconciliation policy, ignoring anything on the individual resource. It's effectively a switch between "resource based annotations only" and "namespace based annotations only".

I'm not sure that's the best way to achieve the desired outcome - for one, I don't think other users would expect ASO to completely ignore annotations on individual resources. 

I'd suggest turning this inside out, so to speak:

If there is no annotation on a specific resource, check the namespace to see if there is one there.

* Achieves the same goal, of allowing control at the namespace level.
* Adds the ability to override for a specific resource if desired.
* Consistent with other things in ASO (such as credentials) where a setting on an individual resource takes precedence over a wider one.
* Doesn't require any configuration.

What do you think?

v2/pkg/common/config/config.go
antonioquintavalle1A: Opt-in has been provided because of the fact that it will change the expected behavior of the operator.
Today you can have 2 resources in the same namespace with a different reconcile policy.
Once activate this one, if an annotation is specified at namespace level, resources will suddenly be managed according to that policy.
This might have a side effect that might not be explictly wanted by ASO users.

antonioquintavalle1A:
> If `READ_RECONCILIATION_POLICY_FROM_NAMESPACE` is set, this implementation looks _only_ on the namespace for a reconciliation policy, ignoring anything on the individual resource. It's effectively a switch between "resource based annotations only" and "namespace based annotations only".
> 
> I'm not sure that's the best way to achieve the desired outcome - for one, I don't think other users would expect ASO to completely ignore annotations on individual resources.
> 
> I'd suggest turning this inside out, so to speak:
> 
> If there is no annotation on a specific resource, check the namespace to see if there is one there.
> 
>     * Achieves the same goal, of allowing control at the namespace level.
> 
>     * Adds the ability to override for a specific resource if desired.
> 
>     * Consistent with other things in ASO (such as credentials) where a setting on an individual resource takes precedence over a wider one.
> 
>     * Doesn't require any configuration.
> 
> 
> What do you think?

Rewieving this I guess that everybody agrees on the fact that resource override must be kept and namespace should not be the solely source of truth.
I'll re-adapt the code to support such use case and we'll check it again the overall change as soon as it gets ready.

theunrepentantgeek:
How are you going @antonioquintavalle1A? Do you need any assistance?

antonioquintavalle1A:
Hello @theunrepentantgeek @matthchr,
here's an attempt to push for the approach we discussed.

- annotations on objects take priority over these defined at namespace1 level
- ASO kinds get now a watcher on namespace; when a watch event gets fired we get the list of all the object kinds into it  and we process them only if it contains a reconcile annotation.

I'm far away to claim that this code is perfect but I prefer to submit such PR to evaluate the impact of this code change and to assess if I should change approach.

theunrepentantgeek (approved):

theunrepentantgeek:
The approach looks good - but currently the code does not compile.

The function `GetReconcilePolicy` returns a single result:

``` go
func GetReconcilePolicy(
        obj genruntime.MetaObject, 
        log logr.Logger, 
        defaultReconcilePolicy annotations.ReconcilePolicyValue) annotations.ReconcilePolicyValue {
```

but invocations are expecting an error as well, e.g. `generic_reconciler.go`:

``` go
reconcilePolicy, err := reconcilers.GetReconcilePolicy(obj, log, gr.Config.DefaultReconcilePolicy)
```

antonioquintavalle1A:
@theunrepentantgeek, effectively I missed to stage a file in the changeset.
This is why it compiled locally and not here. 
Sorry for that.

v2/internal/reconcilers/generic/generic_reconciler.go
```diff
@@ -76,6 +77,8 @@ func (gr *GenericReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 
 	log := gr.LoggerFactory(metaObj).WithValues("name", req.Name, "namespace", req.Namespace)
 
+	log.V(Info).Info("Object to reconcile", metaObj)
```
matthchr: This is already done below:
`reconcilers.LogObj(log, Verbose, "Reconcile invoked", metaObj)`

I think you should remove this.

v2/internal/reconcilers/generic/generic_reconciler.go
```diff
@@ -382,3 +385,31 @@ func (gr *GenericReconciler) writeReadyConditionErrorOrDefault(ctx context.Conte
 	err = gr.WriteReadyConditionError(ctx, log, metaObj, readyErr)
 	return err
 }
+
+func (gr *GenericReconciler) mergeReconcilePolicy(ctx context.Context, log logr.Logger, obj genruntime.MetaObject) annotations.ReconcilePolicyValue {
+	log.V(Info).Info("Looking for namespace annotations")
+	// Claim resource namespace
```
matthchr: minor: this comment about claiming namespace is a bit confusing here -- what is this meaning?

v2/internal/reconcilers/predicates.go
```diff
@@ -46,7 +46,7 @@ func GetReconcilePolicy(obj genruntime.MetaObject, log logr.Logger, defaultRecon
 			"policyAnnotation", policyStr)
 	}
 
-	return policy
+	return policy, err
```
matthchr: I don't think we originally wanted to return an error here, we just log above and use the default?

Though actually looking at that function it doesn't default to the right thing anymore it should default to `defaultReconcilePolicy`... I'll send a PR to fix that.

v2/internal/reconcilers/generic/register.go
```diff
@@ -94,6 +101,7 @@ func RegisterAll(
 
 	var errs []error
 	for _, obj := range objs {
+		options.LogConstructor(nil).V(Info).Info("Registering", "objectName", obj.Name)
```
matthchr, line 104: Already logged in the `register` function -- no need for a duplicate log here?

v2/internal/reconcilers/generic/register.go
```diff
@@ -143,19 +152,57 @@ func register(
 		PanicHandler:              options.PanicHandler,
 	}
 
+	// tHis registers the reconciler for the given resource
```
matthchr:
```suggestion
	// This registers the reconciler for the given resource
```

v2/internal/reconcilers/generic/generic_reconciler.go
```diff
@@ -382,3 +385,31 @@ func (gr *GenericReconciler) writeReadyConditionErrorOrDefault(ctx context.Conte
...
+	// We initially get the reconcile policy from the object itself
+	reconcilePolicy, err := reconcilers.GetReconcilePolicy(obj, log, gr.Config.DefaultReconcilePolicy)
+
+	// If the policy is not defined at object level, then we check if it's defined at namespace level
+	// in case this is not defined, reconcilePolicy will be set to defaultReconcilePolicy and we'll have an error here
+	if err != nil {
+		// here we check if any annotation is specified at namespace level
+		namespaceObject, err := gr.KubeClient.GetObject(ctx, types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetNamespace()}, schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Namespace"})
+		// if we cannot get the namespace, we return defaul reconcile policy
```
matthchr:
```suggestion
		// if we cannot get the namespace, we return default reconcile policy
```

v2/internal/reconcilers/generic/generic_reconciler.go
```diff
@@ -382,3 +385,31 @@ func (gr *GenericReconciler) writeReadyConditionErrorOrDefault(ctx context.Conte
...
+		if err != nil {
+			// reconcilePolicy here will be set to manage in case we get an error
+			log.V(Info).Info("Error while retrieving namespace object", "error", err)
```
matthchr: This should probably be `log.V(Info).Error` instead.

v2/internal/reconcilers/generic/generic_reconciler.go
```diff
@@ -382,3 +385,31 @@ func (gr *GenericReconciler) writeReadyConditionErrorOrDefault(ctx context.Conte
 	err = gr.WriteReadyConditionError(ctx, log, metaObj, readyErr)
 	return err
 }
+
+func (gr *GenericReconciler) mergeReconcilePolicy(ctx context.Context, log logr.Logger, obj genruntime.MetaObject) annotations.ReconcilePolicyValue {
+	log.V(Info).Info("Looking for namespace annotations")
```
matthchr: IMO remove this log and collapse with the one at the end. (See comment there).

v2/internal/reconcilers/generic/generic_reconciler.go
```diff
@@ -382,3 +385,31 @@ func (gr *GenericReconciler) writeReadyConditionErrorOrDefault(ctx context.Conte
...
+			log.V(Info).Info("Error while retrieving namespace object", "error", err)
+			return reconcilePolicy
+		}
+		log.V(Info).Info("Found Namespace annotations", namespaceObject.GetAnnotations()[annotations.ReconcilePolicy])
+		namespaceAnnotation := namespaceObject.GetAnnotations()[annotations.ReconcilePolicy]
+		// this will be either the reconcile policy defined at namespace level either the default one
+		reconcilePolicy, err = reconcilers.ParseReconcilePolicy(namespaceAnnotation, gr.Config.DefaultReconcilePolicy)
+	}
+
+	log.V(Info).Info("Retrieved reconcile policy", reconcilePolicy)
```
matthchr: minor:
```suggestion
	log.V(Verbose).Info("Retrieved reconcile policy", "reconcilePolicy", reconcilePolicy)
```

v2/internal/reconcilers/generic/generic_reconciler.go
```diff
@@ -382,3 +385,31 @@ func (gr *GenericReconciler) writeReadyConditionErrorOrDefault(ctx context.Conte
...
+	if err != nil {
+		// here we check if any annotation is specified at namespace level
+		namespaceObject, err := gr.KubeClient.GetObject(ctx, types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetNamespace()}, schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Namespace"})
+		// if we cannot get the namespace, we return defaul reconcile policy
+		if err != nil {
+			// reconcilePolicy here will be set to manage in case we get an error
+			log.V(Info).Info("Error while retrieving namespace object", "error", err)
+			return reconcilePolicy
+		}
+		log.V(Info).Info("Found Namespace annotations", namespaceObject.GetAnnotations()[annotations.ReconcilePolicy])
```
matthchr: Remove this log too?

v2/internal/reconcilers/generic/generic_reconciler.go
matthchr: and if we want to indicate where it came from we can possibly set some flags (a string `source`?) and use that to build a single log-line with `"source", source` or similar.

v2/internal/reconcilers/generic/register.go
```diff
@@ -143,19 +152,57 @@ func register(
...
+// This registers a watcher on the namespace for the watched resource
+func registerNamespaceWatcher(options Options, kubeclient kubeclient.Client, object client.Object, gvk schema.GroupVersionKind) func(ctx context.Context, obj client.Object) []reconcile.Request {
+	options.LogConstructor(nil).V(Status).Info("Handler for resource object", "name", object.GetName(), "object.namespace", object.GetNamespace(), "object.kind", gvk.Kind)
+	return func(ctx context.Context, obj client.Object) []reconcile.Request {
+		// options.LogConstructor(nil).V(Info).Info("Registering namespace watcher for resource", obj, "to reconcile", obj.GetName(), "in namespace", obj.GetNamespace())
+		// Trigger reconciliation for the BackupBusybox in the same namespace
+		// here we can be a bit more clever trying to reconcile the resource only if the namespace has the annotation to reconcile
+		aList := &unstructured.UnstructuredList{}
+		aList.SetGroupVersionKind(gvk)
+		// options.LogConstructor(nil).V(Status).Info("Registering watcher for", obj.GetName(), "in namespace", obj.GetNamespace())
```
matthchr: Commented code

v2/internal/reconcilers/generic/register.go
```diff
@@ -143,19 +152,57 @@ func register(
...
 	}
 
 	return nil
 }
+
+// This registers a watcher on the namespace for the watched resource
+func registerNamespaceWatcher(options Options, kubeclient kubeclient.Client, object client.Object, gvk schema.GroupVersionKind) func(ctx context.Context, obj client.Object) []reconcile.Request {
+	options.LogConstructor(nil).V(Status).Info("Handler for resource object", "name", object.GetName(), "object.namespace", object.GetNamespace(), "object.kind", gvk.Kind)
+	return func(ctx context.Context, obj client.Object) []reconcile.Request {
+		// options.LogConstructor(nil).V(Info).Info("Registering namespace watcher for resource", obj, "to reconcile", obj.GetName(), "in namespace", obj.GetNamespace())
```
matthchr: commented code.

v2/internal/reconcilers/generic/register.go
```diff
@@ -143,19 +152,57 @@ func register(
...
+		// Trigger reconciliation for the BackupBusybox in the same namespace
```
matthchr: Comments dont seem right?

v2/internal/reconcilers/generic/register.go
```diff
@@ -143,19 +152,57 @@ func register(
...
 	if err != nil {
 		return eris.Wrap(err, "unable to build controllers / reconciler")
 	}
 
 	return nil
 }
+
+// This registers a watcher on the namespace for the watched resource
+func registerNamespaceWatcher(options Options, kubeclient kubeclient.Client, object client.Object, gvk schema.GroupVersionKind) func(ctx context.Context, obj client.Object) []reconcile.Request {
+	options.LogConstructor(nil).V(Status).Info("Handler for resource object", "name", object.GetName(), "object.namespace", object.GetNamespace(), "object.kind", gvk.Kind)
```
matthchr, line 175: Only called on registration, which is fine but the log doesn't seem to match/reflect that.

theunrepentantgeek:
> @theunrepentantgeek, effectively I missed to stage a file in the changeset. This is why it compiled locally and not here. Sorry for that.

We've all done that from time to time.

theunrepentantgeek:
The build of your stuff looks good - unfortunately the CI run is failing because of some broken links in the documentation. :cry:

I have a PR to fix this (#4786) - once that's completed, merge `main` into your branch and I expect the CI build will pass and @matthchr will merge this for you (I'm on leave after today).

antonioquintavalle1A:
Thanks @theunrepentantgeek,
I removed useless comments and transitioned all relevant logs to Verbose level instead of Info.

v2/internal/reconcilers/generic/register.go
```diff
@@ -159,3 +169,30 @@ func register(
...
+	return func(ctx context.Context, obj client.Object) []reconcile.Request {
+		aList := &unstructured.UnstructuredList{}
+		aList.SetGroupVersionKind(gvk)
+		if err := kubeclient.List(ctx, aList, &client.ListOptions{Namespace: obj.GetName()}); err != nil {
+			return []reconcile.Request{}
+		}
+		// list the objects for the current kind
+		for _, el := range aList.Items {
+			if _, ok := obj.GetAnnotations()["serviceoperator.azure.com/reconcile-policy"]; ok {
+				options.LogConstructor(nil).V(Status).Info("Reconciling object", el.GetName())
```
matthchr: I think remove this log entirely? It will already be logged by the reconciler when it gets the event. A log outside of this loop that just says something like "detected namespace skip-reconcile annotation change" might be worth having though?

v2/internal/reconcilers/generic/generic_reconciler.go
```diff
@@ -382,3 +383,31 @@ func (gr *GenericReconciler) writeReadyConditionErrorOrDefault(ctx context.Conte
 	err = gr.WriteReadyConditionError(ctx, log, metaObj, readyErr)
 	return err
 }
+
+func (gr *GenericReconciler) mergeReconcilePolicy(ctx context.Context, log logr.Logger, obj genruntime.MetaObject) annotations.ReconcilePolicyValue {
```
matthchr, line 387: I don't love this `mergeReconcilePolicy` because it feels like it and `GetReconcilePolicy` are trying to do the same thing.

I think maybe we should just delete `reconcilers.GetReconcilePolicy` (it's only used here now) and move the logic it has to here. Also undo your changes to ParseReconcilePolicy (it should only error if it doesn't understand the policy).

Then this function becomes:
```
func (gr *GenericReconciler) mergeReconcilePolicy(ctx context.Context, log logr.Logger, obj genruntime.MetaObject) annotations.ReconcilePolicyValue {
	// We initially get the reconcile policy from the object itself
	source := "default" // assume the source is the default policy for now - this source field is used only for logging purposes
	policyStr := obj.GetAnnotations()[annotations.ReconcilePolicy]

	// If the policy is not defined at object level, then we check if it's defined at namespace level
	if policyStr == "" {
		namespaceObject, err := gr.KubeClient.GetObject(ctx, types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetNamespace()}, schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Namespace"})
		// if we cannot get the namespace, we return default reconcile policy
		if err != nil {
			log.V(Verbose).Info("Error while retrieving namespace object", "error", err)
			return gr.Config.DefaultReconcilePolicy // return default in case of error
		}
		policyStr = namespaceObject.GetAnnotations()[annotations.ReconcilePolicy]
		if policyStr != "" {
			source = "namespace"
		}
	} else {
		source = "object" // used to track where the policy was taken from for logging purposes
	}

	reconcilePolicy, err = reconcilers.ParseReconcilePolicy(policyStr, gr.Config.DefaultReconcilePolicy)
	if err != nil {
		log.Error(
			err,
			"failed to get reconcile policy. Applying default policy instead",
			"chosenPolicy", policy,
			"policyAnnotation", policyStr)
	}
	log.V(Verbose).Info("Retrieved reconcile policy", "policy", reconcilePolicy, "source", source)
	return reconcilePolicy
}
```

v2/internal/reconcilers/generic/register.go
matthchr: This was already resolved.

v2/internal/reconcilers/generic/generic_reconciler.go
matthchr: This still applies I think (even though this was changed to Verbose, I actually think Inf owas fine but you want the `Error` helper.

v2/internal/reconcilers/generic/generic_reconciler.go
matthchr: I took a crack at what this method might look like here: https://github.com/Azure/azure-service-operator/pull/4692#discussion_r2155166167

matthchr:
I think the docs link issue will be fixed by a rebase, see https://github.com/Azure/azure-service-operator/pull/4786

matthchr:
Test:
```
[produce-markdown-summary] - Test failed: TestParseReconcilePolicy
[produce-markdown-summary] === TEST OUTPUT ===
[produce-markdown-summary] === RUN   TestParseReconcilePolicy
[produce-markdown-summary] === PAUSE TestParseReconcilePolicy
[produce-markdown-summary] === CONT  TestParseReconcilePolicy
[produce-markdown-summary]     reconcile_policy_test.go:35: 
[produce-markdown-summary]         Unexpected error:
[produce-markdown-summary]             <*eris.rootError | 0xc0004cc570>: 
[produce-markdown-summary]             empty annotation
[produce-markdown-summary]             {
[produce-markdown-summary]                 global: false,
[produce-markdown-summary]                 msg: "empty annotation",
[produce-markdown-summary]                 ext: nil,
[produce-markdown-summary]                 stack: [0x162331f, 0x162395b],
[produce-markdown-summary]             }
[produce-markdown-summary]         occurred
[produce-markdown-summary] --- FAIL: TestParseReconcilePolicy (0.00s)
```

Failed in CI - PTAL.

Live tests (see errors here): https://github.com/Azure/azure-service-operator/actions/runs/15739972785/job/44362563255 - seems to have hit an issue with the webhooks:
> Message: "conversion webhook for resources.azure.com/v1api20200601storage, Kind=ResourceGroup failed: Post \"[https://azureserviceoperator-webhook-service.azureserviceoperator-system.svc:443/convert?timeout=30s\](https://azureserviceoperator-webhook-service.azureserviceoperator-system.svc/convert?timeout=30s\)": dial tcp 10.96.252.94:443: connect: connection refused",

This is probably due to the pod crashing. Best way to understand why is to run the operator in a kind cluster locally - you should be able to do this with `task controller:kind-create-with-workload-identity` - you may need to run `task provision-oidc-storage-acct` first (and I'm not sure how many people outside of MSFT have tried this so let us know if you have issues. It _should_ work but YMMV). If you struggle with this for long let me know and I can see if I can help figure out what the issue is.

v2/internal/reconcilers/generic/generic_reconciler.go
antonioquintavalle1A: Indeed, that has been removed while you were reviewing the PR.

v2/internal/reconcilers/generic/generic_reconciler.go
antonioquintavalle1A: This is a leftover, good catch

v2/internal/reconcilers/predicates.go
antonioquintavalle1A: To me it looks it still defaults to the right `defaultReconcilePolicy` in case a wrong policyString is provided as input.

However the reason why I'm returning an error is to be identified when a namespace has a reconcile policy but not the object itself (or even a not valid one).
Without returning an error here, I won't be able to determine if the lookup at resource level is correct (e.g. `reconcilePolicy` correclty speficied) or incorrect (e.g. `reconcilePolicy` not set/ set to empty "" / set to a random value).

I do not mind to create a dedicated wrapper function for this specific case if needed be.

v2/internal/reconcilers/generic/register.go
antonioquintavalle1A: What about if we let it print only in verbose / debug mode?

antonioquintavalle1A: On top, do we agree here that we watch a resource only if its namespace contains the wanted annotation? Shall we check only if it's present (as now) or shall we check also if it's valid or not?

antonioquintavalle1A:
> Test:
> 
> ```
> [produce-markdown-summary] - Test failed: TestParseReconcilePolicy
> [produce-markdown-summary] === TEST OUTPUT ===
> [produce-markdown-summary] === RUN   TestParseReconcilePolicy
> [produce-markdown-summary] === PAUSE TestParseReconcilePolicy
> [produce-markdown-summary] === CONT  TestParseReconcilePolicy
> [produce-markdown-summary]     reconcile_policy_test.go:35: 
> [produce-markdown-summary]         Unexpected error:
> [produce-markdown-summary]             <*eris.rootError | 0xc0004cc570>: 
> [produce-markdown-summary]             empty annotation
> [produce-markdown-summary]             {
> [produce-markdown-summary]                 global: false,
> [produce-markdown-summary]                 msg: "empty annotation",
> [produce-markdown-summary]                 ext: nil,
> [produce-markdown-summary]                 stack: [0x162331f, 0x162395b],
> [produce-markdown-summary]             }
> [produce-markdown-summary]         occurred
> [produce-markdown-summary] --- FAIL: TestParseReconcilePolicy (0.00s)
> ```
> 
> Failed in CI - PTAL.
> 
> Live tests (see errors here): https://github.com/Azure/azure-service-operator/actions/runs/15739972785/job/44362563255 - seems to have hit an issue with the webhooks:
> 
> > Message: "conversion webhook for resources.azure.com/v1api20200601storage, Kind=ResourceGroup failed: Post "[https://azureserviceoperator-webhook-service.azureserviceoperator-system.svc:443/convert?timeout=30s](https://azureserviceoperator-webhook-service.azureserviceoperator-system.svc/convert?timeout=30s)": dial tcp 10.96.252.94:443: connect: connection refused",
> 
> This is probably due to the pod crashing. Best way to understand why is to run the operator in a kind cluster locally - you should be able to do this with `task controller:kind-create-with-workload-identity` - you may need to run `task provision-oidc-storage-acct` first (and I'm not sure how many people outside of MSFT have tried this so let us know if you have issues. It _should_ work but YMMV). If you struggle with this for long let me know and I can see if I can help figure out what the issue is.

That's strange, I recompiled the operator and loaded it on an openshift cluster I have. Didn't observe any particular crash or instabilities.

Nevertheless, I'll try it on kind to see if I can replicate the issue spot by the CI.

v2/internal/reconcilers/generic/generic_reconciler.go
antonioquintavalle1A: I didn't want to remove legacy code here without having a discussion with you before.
I agree on the statement and on the proposed solution.

antonioquintavalle1A:
Rebased on main to include https://github.com/Azure/azure-service-operator/pull/4786

v2/internal/reconcilers/generic/register.go
matthchr: Doesn't seem changed?

v2/internal/reconcilers/generic/register.go
```diff
@@ -148,6 +156,8 @@ func register(
 		WithOptions(options.Options)
 	builder.Named(info.Name)
 
+	builder = builder.Watches(&corev1.Namespace{}, handler.EnqueueRequestsFromMapFunc(registerNamespaceWatcher(options, kubeClient, info.Obj, gvk)), ctrlbuilder.WithPredicates(predicate.AnnotationChangedPredicate{}))
```
matthchr: Please change this to use internal/reconcilers/predicates.go `ARMReconcilerAnnotationChangedPredicate`

That will filter the events to only the annotation we care about.

v2/internal/reconcilers/generic/register.go
matthchr: Great callout/question, I _think_ we want to run reconciliation only of the obj does NOT have this annotation -- right now you're doing it only if the obj DOES have the annotation which seems backwards to me (because then the obj's annotation wins, right?)

See also https://github.com/Azure/azure-service-operator/pull/4692/files#r2157801402

matthchr: and in terms of this log, I'm fine with it at `Verbose` or `Debug` level.

v2/internal/reconcilers/generic/generic_reconciler.go
```diff
@@ -382,3 +383,30 @@ func (gr *GenericReconciler) writeReadyConditionErrorOrDefault(ctx context.Conte
...
+	if err != nil {
+		// here we check if any annotation is specified at namespace level
+		namespaceObject, err := gr.KubeClient.GetObject(ctx, types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetNamespace()}, schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Namespace"})
+		// if we cannot get the namespace, we return default reconcile policy
+		if err != nil {
+			// reconcilePolicy here will be set to manage in case we get an error
+			log.V(Verbose).Info("Error while retrieving namespace object", "error", err)
+			return reconcilePolicy
+		}
+		log.V(Verbose).Info("Found Namespace annotations", namespaceObject.GetAnnotations()[annotations.ReconcilePolicy])
```
matthchr: This should probably be log.V(Verbose).Error instead.

v2/internal/reconcilers/generic/register.go
antonioquintavalle1A: > _nk_ we want to run reconciliation only of the obj does NOT have this annotation

Right now for each object we register, we watch the object (primary resource) and the namespace that contains it (secondary resource).
The event on the primary resource iwe have a dedicated reconcile loop, that I haven't changed with this PR.
On the other case, when an event on the namespace is fired we perform the reconciliation of the primary resource only if the namespace has the reconcile-policy is set.

Does it look right to you?

v2/internal/reconcilers/generic/generic_reconciler.go
matthchr: Were you planning on actually removing `reconcilers.GetReconcilePolicy`? As I mentioned above I dont think we need it anymore since everybody is calling this `mergeReconcilePolicy` helper instead?

v2/internal/reconcilers/generic/generic_reconciler.go
```diff
@@ -382,3 +383,30 @@ func (gr *GenericReconciler) writeReadyConditionErrorOrDefault(ctx context.Conte
...
+		namespaceAnnotation := namespaceObject.GetAnnotations()[annotations.ReconcilePolicy]
+		// this will be either the reconcile policy defined at namespace level either the default one
+		reconcilePolicy, err = reconcilers.ParseReconcilePolicy(namespaceAnnotation, gr.Config.DefaultReconcilePolicy)
+	}
+
+	log.V(Verbose).Info("Retrieved reconcile policy", reconcilePolicy)
+	return reconcilePolicy
+	// Here we have an annotation that comes from the namespace
```
matthchr: minor: This comment is in a weird spot now?

v2/internal/reconcilers/generic/register.go
matthchr: Ahh yes sorry I missed the namespace (called obj) versus object (called el) nuance here on first read.

I don't strictly speaking think the "check of obj has the annotation set" is required, but it doesn't hurt so yes I'm happy with that. What I was getting at above is, I don't think we want to re-reconcile the `el` if _it_ has the annotation, because if it does, its annotation takes precedence anyway so there's no point to re-reconciling even if the namespace annotation changed?

antonioquintavalle1A: Well yes, in that case we'll have an useless update.
In my opinion this is a bit limited by what kubebuilder framework allows to do with predicates and watch events.

In that case, we watch two distinct objects: `ASO CR` and `namespace`. A predicate will be aware of the change in one of these objects at the time but not for both at the same time.
As a result, a predicate filtering cannot be helpful here, not even if we leverage on some api calls to retrieve additional objects.
Even in that case, I do not see how can we get in any context  the old/new version of any object in the namespace to know if their annotation has been changed or not, implying again an useless reconcile loop.

Do you see maybe any other clever mean that we might use to avoid useless reconciliation for the use case you mentioned?

v2/internal/reconcilers/generic/generic_reconciler.go
antonioquintavalle1A: No, I was not planning to remove it.
As matter of facts, today `mergeReconcilePolicy` wraps a call to `GetReconcilePolicy`.

Shall I proceed with the method merging as you suggested before?
Don't you think we might need the same predicate for any other future scope?

antonioquintavalle1A:
@matthchr I haven't been able to create a local environment with Kind due to its complexity.
Is there maybe there a way to access the runner while it performs the tests?

antonioquintavalle1A:
Hello there @matthchr, I built locally the operator and I ran it on kind without any major issues.
I have the pods up and running since 20 hours and I didn't observe any reboot nor crash, but it was loaded with a limited set of CRDs.

I think the environment I was able to spin up is significatively different from what is built by the Taskfile.yaml, I cannot perform many of the steps reported in there easily in here.

Do you see any other way how can we move forward any troubleshooting activity?

matthchr:
Let me requeue the tests. If they fail, I can take a look on Monday (it's a holiday in the US today) to help move this forward.

Thanks for your perseverance!

matthchr:
Ok I checked this out and there are two issues:
**Failing unit test.** 
You still need to fix this test: https://github.com/Azure/azure-service-operator/pull/4692#issuecomment-2987335164. If you do as I suggested [here](https://github.com/Azure/azure-service-operator/pull/4692/files#r2155166167) and drop `GetReconcilePolicy` entirely, you don't need `ParseReconcilePolicy` to return an error and so you can change the impl of `ParseReconcilePolicy` back and the test will pass.

**Lack of namespaces permission**:
The pod is cycling in kind because we don't have permission to read namespaces.

```
E0708 00:22:52.713545       1 runtime.go:226] "Failed to watch" err="failed to list *v1.Namespace: namespaces is forbidden: User \"system:serviceaccount:azureserviceoperator-system:azureserviceoperator-default\" cannot list resource \"namespaces\" in API group \"\" at the cluster scope" logger="controller-runtime.cache.UnhandledError" reflector="pkg/mod/k8s.io/client-go@v0.33.1/tools/cache/reflector.go:285" type="*v1.Namespace"
E0708 00:22:55.475682       1 runtime.go:226] "Failed to watch" err="failed to list *v1.Namespace: namespaces is forbidden: User \"system:serviceaccount:azureserviceoperator-system:azureserviceoperator-default\" cannot list resource \"namespaces\" in API group \"\" at the cluster scope" logger="controller-runtime.cache.UnhandledError" reflector="pkg/mod/k8s.io/client-go@v0.33.1/tools/cache/reflector.go:285" type="*v1.Namespace"
E0708 00:23:00.855021       1 runtime.go:226] "Failed to watch" err="failed to list *v1.Namespace: namespaces is forbidden: User \"system:serviceaccount:azureserviceoperator-system:azureserviceoperator-default\" cannot list resource \"namespaces\" in API group \"\" at the cluster scope" logger="controller-runtime.cache.UnhandledError" reflector="pkg/mod/k8s.io/client-go@v0.33.1/tools/cache/reflector.go:285" type="*v1.Namespace"
E0708 00:23:10.846384       1 runtime.go:226] "Failed to watch" err="failed to list *v1.Namespace: namespaces is forbidden: User \"system:serviceaccount:azureserviceoperator-system:azureserviceoperator-default\" cannot list resource \"namespaces\" in API group \"\" at the cluster scope" logger="controller-runtime.cache.UnhandledError" reflector="pkg/mod/k8s.io/client-go@v0.33.1/tools/cache/reflector.go:285" type="*v1.Namespace"
E0708 00:23:30.293361       1 runtime.go:226] "Failed to watch" err="failed to list *v1.Namespace: namespaces is forbidden: User \"system:serviceaccount:azureserviceoperator-system:azureserviceoperator-default\" cannot list resource \"namespaces\" in API group \"\" at the cluster scope" logger="controller-runtime.cache.UnhandledError" reflector="pkg/mod/k8s.io/client-go@v0.33.1/tools/cache/reflector.go:285" type="*v1.Namespace"
E0708 00:24:00.946169       1 runtime.go:226] "Failed to watch" err="failed to list *v1.Namespace: namespaces is forbidden: User \"system:serviceaccount:azureserviceoperator-system:azureserviceoperator-default\" cannot list resource \"namespaces\" in API group \"\" at the cluster scope" logger="controller-runtime.cache.UnhandledError" reflector="pkg/mod/k8s.io/client-go@v0.33.1/tools/cache/reflector.go:285" type="*v1.Namespace"
E0708 00:24:33.750907       1 runtime.go:226] "Failed to watch" err="failed to list *v1.Namespace: namespaces is forbidden: User \"system:serviceaccount:azureserviceoperator-system:azureserviceoperator-default\" cannot list resource \"namespaces\" in API group \"\" at the cluster scope" logger="controller-runtime.cache.UnhandledError" reflector="pkg/mod/k8s.io/client-go@v0.33.1/tools/cache/reflector.go:285" type="*v1.Namespace"
```

We need to add the namespaces read permission to the operator.  You need to add it here. https://github.com/Azure/azure-service-operator/blob/15efd612c9d2c4a2c94634ba6b1b4dfe26034788/v2/api/resources/v1api20200601/permissions.go#L4

Should be just get/list/watch - no update/patch/delete for namespaces.

antonioquintavalle1A:
Hello there @matthchr, thanks for pointing out such issue, I think it went unspotted due to the fact that the service account has more permissions than the one defined in this repository.
In my opinion this brings in an important point related to RBAC.

As you pointed out, azure-service-operator serviceAccount must have a binding that allows to list/get/watch a namespace where the CR is created.

One simple way to address this, would be to give a clusterRoleBinding to allow reading of each and every namespace.

What do you think, instead if we fail safe in case the ServiceAccount does not have any permissions over the namespace so that we are not obliged to specify this as a ClusterRoleBinding and we read annotations on the namespace only, and only if, such permission is explicitly provided?

matthchr:
> One simple way to address this, would be to give a clusterRoleBinding to allow reading of each and every namespace.

I think we need to add the permission to the set of roles the operator has, similar to how it can read secrets and configmaps. I'm not too worried about failing safe if we don't have the perms, at least for now, because as far as I know, most/all users are using the default permissions we grant anyway, and if users were going to have problems with certain permissions I would imagine it would be the configmap/secrets read, rather than namespaces read

v2/internal/reconcilers/generic/register.go
matthchr: Still doesn't seem changed.

v2/internal/reconcilers/generic/register.go
matthchr: minor: ping on this item, I think we should improve this log message.

v2/internal/reconcilers/generic/register.go
matthchr: Reduce from `Status` to a more verbose log level.

matthchr: I'm not sure I follow your point here:
> In that case, we watch two distinct objects: ASO CR and namespace. A predicate will be aware of the change in one of these objects at the time but not for both at the same time.

You call:
```
		if err := kubeclient.List(ctx, aList, &client.ListOptions{Namespace: obj.GetName()}); err != nil {
			return []reconcile.Request{}
		}
```

If a given item in this list (say the `Foo` `ResourceGroup`), has the `serviceoperator.azure.com/reconcile-policy` set, then reconciling that resource is (AFAIK) useless. The only case I can think of where it wouldn't be useless is if actually the annotation on that object was removed, but our cache just wasn't updated yet. In this case though, we could still not emit an event here, and we will emit an event when we observe the cache update for the annotation change on the resource itself.

basically if you do:
```
for _, el := range aList.Items {
    if _, ok := el.GetAnnotations()["serviceoperator.azure.com/reconcile-policy"]; ok {
        continue // No point emitting event here as object already has its own annotation that takes precedence.
    }
```

That solves this? Am I missing the case where the above is actually not correct and we should emit an event to reconcile an object that looks to have its own `serviceoperator.azure.com/reconcile-policy` specified?

FWIW, if your argument is just "this is an optimization we don't need", that's fair enough - I'm just trying to make sure I understand if my mental model here is wrong and I'm missing something.

matthchr (commented):
This overall LGTM, I just have a few nits about logs.

If you can clean those up and response to the latest comment on this thread, I think this is good to merge: https://github.com/Azure/azure-service-operator/pull/4692/files#r2155134742

Thanks for the perseverance.

v2/internal/reconcilers/generic/register.go
antonioquintavalle1A: Hello @matthchr, indeed this will avoid triggering an useless reconcile request.

Re-looking at the code, I believe there's also another mistake though: in the loop we should build up the list of all the objects that might have been impacted by an addition / removal / change of the annotation at namespace level; at the time being it stops returning only the first one, which is clearly not exhaustive.

I'll amend the code and we can re-discuss if needed.

v2/internal/reconcilers/generic/generic_reconciler.go
matthchr: I think this is done now. Resolving this convo.

matthchr (approved):

matthchr:
Can you run `task controller:format-code`? Looks like some formatting not quite matching CIs expectations

antonioquintavalle1A:
Hello there @matthchr,
I ran `controller:format-code` locally and I've got no diffs in the code.
If I understand it correctly, the error comes from the fact that a file is getting modified when it should have not, as checked here:

`  verify-no-changes:
    desc: Checks that there are no uncommitted modifications to files
    cmds:
      - cmd: 'if [ -n "{{.CHANGED}}" ]; then echo "Error: files were modified during the build (ensure you have commited generated files, and run ''task format-code'' to tidy):"; echo "{{.CHANGED}}"; git diff; exit 1; fi'
        silent: true
    vars:
      CHANGED:
        sh: git status --porcelain`

However the change here alters solely the order of declarations not the file content:

` [verify-no-changes] --- a/v2/internal/reconcilers/generic/register.go
[verify-no-changes] +++ b/v2/internal/reconcilers/generic/register.go
[verify-no-changes] @@ -10,7 +10,6 @@ import (
[verify-no-changes]  	"fmt"
[verify-no-changes]  
[verify-no-changes]  	. "github.com/Azure/azure-service-operator/v2/internal/logging"
[verify-no-changes] -	"github.com/Azure/azure-service-operator/v2/internal/reconcilers"
[verify-no-changes]  
[verify-no-changes]  	"github.com/go-logr/logr"
[verify-no-changes]  	"github.com/rotisserie/eris"
[verify-no-changes] @@ -30,6 +29,7 @@ import (
[verify-no-changes]  	"sigs.k8s.io/controller-runtime/pkg/reconcile"
[verify-no-changes]  
[verify-no-changes]  	"github.com/Azure/azure-service-operator/v2/internal/config"
[verify-no-changes] +	"github.com/Azure/azure-service-operator/v2/internal/reconcilers"
[verify-no-changes]  	"github.com/Azure/azure-service-operator/v2/internal/util/interval"
[verify-no-changes]  	"github.com/Azure/azure-service-operator/v2/internal/util/kubeclient"
[verify-no-changes]  	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"`

Shall we relax a bit the way local files are checked?

matthchr:
We're intentionally picky about import ordering. I've pushed a fix to your branch. (I told you to run `controller:format-code` but that was wrong it should have been just `controller:lint`).

matthchr:
Thank you for your perseverance on this change! Very happy to see it merged.

theunrepentantgeek:
Ditto! Thanks for perservering.

PR #4720, opened 2025-04-29
Fix the Kusto Database extension to allow owners referenced by armID

### LINKED ISSUE #4712: Kusto Database resource can't be deployed with armId owner

The Kusto Database resource can only be deployed if the cluster is also managed by ASO. Using an armId as an owner is currently not supported.

The following fails to deploy:

```yaml
apiVersion: kusto.azure.com/v1api20230815
kind: Database
metadata:
  annotations:
    serviceoperator.azure.com/reconcile-policy: "detach-on-delete"
  name: {{ $kustoDatabaseName }}
spec:
  owner:
    armId: {{ $kustoClusterId }}
```

## Describe the improvement

Allow the cluster owner for a Database resource to be defined with an armId.

## Additional context

I get the following error message: `cannot run on unknown resource type <nil>, expected owner to be *kusto.Cluster`.

My guess is that it's coming from [database_extensions.go](https://github.com/Azure/azure-service-operator/blob/main/v2/api/kusto/customizations/database_extensions.go) on L42.

```go
	cluster, ok := owner.(*kusto.Cluster)
	if !ok {
		return extensions.PreReconcileCheckResult{},
			eris.Errorf("cannot run on unknown resource type %T, expected owner to be *kusto.Cluster", owner)
	}

```

theunrepentantgeek:
## What this PR does

Fixes the PreReconcileCheck extension for kusto.Database resources to allow their owner to be referenced by armID.

Note: We can only do the prerequisite check for Kusto Clusters that are managed by ASO, so users will end up with _conflict_ messages in their logs.

Closes #4712 

### Special notes

Also fixes up other implementations of `extensions.PreReconciliationChecker` to properly call the `next()` function, bringing them into alignment with the design of the extension.

gif

### DIFF
--- a/v2/api/kusto/customizations/database_extensions.go
+++ b/v2/api/kusto/customizations/database_extensions.go
@@ -34,7 +34,7 @@ func (ext *DatabaseExtension) PreReconcileCheck(
 ) (extensions.PreReconcileCheckResult, error) {
 	// Check to see if the owning cluster is in a state that will block us from reconciling
 	// Owner nil can happen if the owner of the database is referenced by armID
-	if owner == nil {
+	if owner != nil {
 		if cluster, ok := owner.(*kusto.Cluster); ok {
 			// If our owning *cluster* is in a state that will reject any PUT, then we should skip
 			// reconciliation of the database as there's no point in even trying.
--- a/v2/api/kusto/customizations/database_extensions.go
+++ b/v2/api/kusto/customizations/database_extensions.go
@@ -10,8 +10,6 @@ import (
 	"fmt"
 
 	"github.com/go-logr/logr"
-	"github.com/rotisserie/eris"
-	"sigs.k8s.io/controller-runtime/pkg/conversion"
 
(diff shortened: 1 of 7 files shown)


v2/api/kusto/customizations/database_extensions.go
```diff
@@ -26,39 +24,34 @@ var _ extensions.PreReconciliationChecker = &ClusterExtension{}
...
 ) (extensions.PreReconcileCheckResult, error) {
-	// This has to be the current hub storage version. It will need to be updated
-	// if the hub storage version changes.
-	cluster, ok := owner.(*kusto.Cluster)
-	if !ok {
-		return extensions.PreReconcileCheckResult{},
-			eris.Errorf("cannot run on unknown resource type %T, expected owner to be *kusto.Cluster", owner)
+	// Check to see if the owning cluster is in a state that will block us from reconciling
+	// Owner nil can happen if the owner of the database is referenced by armID
+	if owner == nil {
```
matthchr, line 37:
```suggestion
	if owner != nil {
```

matthchr (approved):

PR #4742, opened 2025-05-23
Add FAQ item for concurrent reconciles

theunrepentantgeek:
## What this PR does

Adds a FAQ item following up on a recent [Slack discussion](https://kubernetes.slack.com/archives/C046DEVLAQM/p1744859244823059).

gif

## Checklist

### DIFF
--- a/docs/hugo/content/guide/frequently-asked-questions.md
+++ b/docs/hugo/content/guide/frequently-asked-questions.md
@@ -100,6 +100,14 @@ We instead recommend an active-passive approach, where in 1 cluster the resource
 This can be accomplished with the [serviceoperator.azure.com/reconcile-policy: skip]( {{< relref "annotations#serviceoperatorazurecomreconcile-policy" >}} )
 annotation used in the second cluster. In the case of a DR event, automation or manual action can remove the `skip` annotation in the passive cluster, turning it into active mode.
 
+### What do I do if I suspect ASO is not keeping up with reconciles?
+
+To verify whether ASO is falling behind with reconciles or not, check the [metrics]({{< relref "metrics" >}}) for the operator to see if the workqueue length is getting long. 
+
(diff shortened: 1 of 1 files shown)


docs/hugo/content/guide/frequently-asked-questions.md
```diff
@@ -100,6 +100,14 @@ We instead recommend an active-passive approach, where in 1 cluster the resource
 This can be accomplished with the [serviceoperator.azure.com/reconcile-policy: skip]( {{< relref "annotations#serviceoperatorazurecomreconcile-policy" >}} )
 annotation used in the second cluster. In the case of a DR event, automation or manual action can remove the `skip` annotation in the passive cluster, turning it into active mode.
 
+### What do I do if I suspect ASO is not keeping up with reconciles?
```
matthchr, line 103: Kindof already covered in:
```
### How well does ASO scale?

ASO is designed to easily scale to thousands of resources, and we have customers who routinely reconcile thousands of resources with a single instance of ASO without issue.

A separate reconciler is run for each different kind of resource, and each reconciler is non-blocking. If resource creation (or update) triggers a long-running-operation (LRO), ASO doesn't sit there polling for completion of the operation. Instead, it stashes information about the operation on the resource as an annotation, schedules the resource for a retry of reconciliation, and moves on to reconcile the next resource due. Later, the LRO is picked up and checked to see if it has completed.

For the rare situation where users observe that ASO is failing to "keep up" with resource reconciliation (this can be monitored by looking at the queue length of the operator via [metrics]({{< relref metrics>}})), they can set MAX_CONCURRENT_RECONCILES to a larger value (the default is 1).
```

theunrepentantgeek:
After discussion with @matthchr we've decided that other recent updates to our FAQ (including restructuring) cover this off and we don't need the additional item.

Issue #4147, opened 2024-07-02
kubectl version priority results in get using older version of resources than it should


matthchr:
The versioning scheme we use for APIs doesn't comply with well with [version priority](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#version-priority), which states:

> Version priority
> Regardless of the order in which versions are defined in a CustomResourceDefinition, the version with the highest priority is used by kubectl as the default version to access objects. The priority is determined by parsing the name field to determine the version number, the stability (GA, Beta, or Alpha), and the sequence within that stability level.

> The algorithm used for sorting the versions is designed to sort versions in the same way that the Kubernetes project sorts Kubernetes versions. Versions start with a v followed by a number, an optional beta or alpha designation, and optional additional numeric versioning information. Broadly, a version string might look like v2 or v2beta1. Versions are sorted using the following algorithm:

> Entries that follow Kubernetes version patterns are sorted before those that do not.
For entries that follow Kubernetes version patterns, the numeric portions of the version string is sorted largest to smallest.
If the strings beta or alpha follow the first numeric portion, they sorted in that order, after the equivalent string without the beta or alpha suffix (which is presumed to be the GA version).

> If another number follows the beta, or alpha, those numbers are also sorted from largest to smallest.
Strings that don't fit the above format are sorted alphabetically and the numeric portions are not treated specially. Notice that in the example below, foo1 is sorted above foo10. This is different from the sorting of the numeric portion of entries that do follow the Kubernetes version patterns.

Our APIs are versioned like so:
`v1api20210201` - GA ASO, GA ARM API
`v1api20240201` - GA ASO, GA ARM API (newer)
`v1api20210201preview` - GA ASO, Preview ARM API
`v1beta20210201` - Beta ASO, GA ARM API
`v1beta20210201preview` - Beta ASO, Preview ARM API

According to the Kubernetes version priority ordering, these get ordered:
- `v1beta20210201` - Kubernetes treats this as iteration `20210201` of `v1beta`, so it complies the the Kubernetes versioning scheme and is ordered first.
- `v1api20210201` - Doesn't comply with Kuberentes versioning structure and so is ordered lexically.
- `v1api20210201preview` - Doesn't comply with Kuberentes versioning structure and so is ordered lexically.
- `v1api20240201` - Doesn't comply with Kuberentes versioning structure and so is ordered lexically.
[...]

PR #4750, opened 2025-05-26
Add ADR on how to address version priority

theunrepentantgeek:
## What this PR does

Captures offline discussion about how to address #4147 so that users don't get nasty surprises.

gif

## Checklist

### DIFF
--- a/docs/hugo/content/design/ADR-2025-05-Version-Priority/_index.md
+++ b/docs/hugo/content/design/ADR-2025-05-Version-Priority/_index.md
@@ -5,7 +5,7 @@ toc_hide: true
 
 ## Context
 
-We have a problem with the way we version resources.
+As initially identified in [#4147](https://github.com/Azure/azure-service-operator/issues/4147), we have a problem with the way we version resources.
 
 If a specific resource isn't specified, Kubernetes will automatically select one for use through a process called [_version priority_](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#version-priority). 
 
--- a/docs/hugo/content/design/ADR-2025-05-Version-Priority/_index.md
+++ b/docs/hugo/content/design/ADR-2025-05-Version-Priority/_index.md
@@ -78,7 +78,23 @@ Any resources introduced with older versions of ASO would continue to use the ex
 
 * Still allows the problem to occur (but prevents it from getting worse)
 
-## Version format
+## Option 4: Introduce a new version format for all resources with old version duplicates for existing resources
+
+Reuse the approach we took as we migrated from `v1alpha1` to `v1beta` and from `v1beta` to `v1api`, where we introduce a new version format for all resources, but also keep the old versions around for compatibility reasons.
+
+After several releases of ASO, we'd remove the old versions, as we did with `v1alpha1` and `v1beta`.
+
+### Pros
+
+* Migrates all resources to the new version format, so version priority will work as users expect for all resources.
+* Mitigates the breaking change of Option 2 by keeping the old versions around for a while.
+
+### Cons
+
+* Will eventually a breaking change for some users, but only for those who have not migrated to the new version format in the interim.
+* May not be possible for some resources, as the duplication may result in the aggregate CRD being too large. (This is definitely the case for `ManagedCluster` which is already near the maximum size for a CRD, and would not be able to accommodate the duplication of all the existing versions.)
+
+## Which new version format?
 
 A simple change to the format we use for resource versions would be to simplify the prefix used to just `v`, giving resource versions like this:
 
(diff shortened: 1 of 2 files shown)


docs/hugo/content/design/ADR-2025-05-Version-Priority/_index.md
```diff
@@ -0,0 +1,95 @@
...
+### Pros
+
+* Simple to achieve
+
+### Cons
+
+* We've already had a couple cases where version priority caused issues for users, so we know this problem isn't going away.
+* As we add more resource versions, the scope for this occurring can only increase.
+
+## Option 2: Cut over to versioning format
```
matthchr, line 54:
```suggestion
## Option 2: Cut over to a new versioning format
```

docs/hugo/content/design/ADR-2025-05-Version-Priority/_index.md
```diff
@@ -0,0 +1,95 @@
...
+
+Change all generated API versions to be compliant with Kubernetes version rules.
```
matthchr, line 56: minor: 
```suggestion
Change all generated API versions to be compliant with Kubernetes version rules.

```

docs/hugo/content/design/ADR-2025-05-Version-Priority/_index.md
matthchr: Or maybe you mean change all existing APIs here (seems yes based on cons below) so maybe I got the suggestion for what this should be wrong - but still feels like there would be a better way to phrase it

docs/hugo/content/design/ADR-2025-05-Version-Priority/_index.md
```diff
@@ -0,0 +1,95 @@
...
+
+* Version priority will work as users expect.
+
+### Cons
+
+* Would be breaking for all existing users.
+
+## Option 3: Migrate to new versioning format over time
+
+Introduce a new version format as from a particular release of ASO, leaving existing resources with the existing format.
```
matthchr, line 67: Phrasing: Clarify here that it would only impact _new versions_ of Azure resources

docs/hugo/content/design/ADR-2025-05-Version-Priority/_index.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Cons
+
+* Still allows the problem to occur (but prevents it from getting worse)
+
+## Version format
+
+A simple change to the format we use for resource versions would be to simplify the prefix used to just `v`, giving resource versions like this:
+
+* `v20210201`
```
matthchr, line 82: Another option is to do `v<date>1` or `v1<date>` (we'd probably want `v1<date>` so that updated ASO schemas result in those versions all "winning"), which results in a somewhat awkward looking number but maintains all of the ordering guarantees we want.

Though, I suppose the argument could be made that if we ever want to make a big change like that, we could still do it and just introduce `v1<date>` at that time, and since `1<date>` is larger than `<date>` it'll take precedence, and avoids us needing to commit to the awkward `1` now, only if we ever end up needing it?

I think I'm fine with that as the approach but it would be good to call out some of that logic in the ADR for future-us.

docs/hugo/content/design/ADR-2025-05-Version-Priority/_index.md
theunrepentantgeek: Rephrased to _Change all resources to use a new versioning format_.

docs/hugo/content/design/ADR-2025-05-Version-Priority/_index.md
theunrepentantgeek: Rephrased as "Use a new version format for new resources only" with narrative to clarify that new resource versions are included too.

docs/hugo/content/design/ADR-2025-05-Version-Priority/_index.md
theunrepentantgeek: Added, including discussion on why it's not a good idea.

docs/hugo/content/design/ADR-2025-05-Version-Priority/_index.md
theunrepentantgeek: Done.

docs/hugo/content/design/ADR-2025-05-Version-Priority/_index.md
```diff
@@ -75,21 +78,54 @@ Introduce a new version format as from a particular release of ASO, leaving exis
...
+After several releases of ASO, we'd remove the old versions, as we did with `v1alpha1` and `v1beta`.
+
+### Pros
+
+* Migrates all resources to the new version format, so version priority will work as users expect for all resources.
+* Mitigates the breaking change of Option 2 by keeping the old versions around for a while.
+
+### Cons
+
+* Will eventually a breaking change for some users, but only for those who have not migrated to the new version format in the interim.
```
matthchr:
```suggestion
* Will eventually be a breaking change for some users, but only for those who have not migrated to the new version format in the interim.
```

docs/hugo/content/design/ADR-2025-05-Version-Priority/_index.md
```diff
@@ -75,21 +78,54 @@ Introduce a new version format as from a particular release of ASO, leaving exis
...
+This would seem to be quite awkward, difficult for users to understand, and potentially for zero benefit. Maybe YAGNI applies.
 
-Recommendation: Option 3 with the proposed version change.
+## Status
 
+Recommendation: 
+_Either_
+Option 3 with the proposed version change to use just `v` as the prefix.
+_or_
+Option 4 with the proposed version change to use just `v` as the prefix, but with the understanding that some resources may not be able to accommodate the duplication of existing versions.
```
matthchr, line 129: This is my preference (as discussed).

matthchr (approved):

PR #4757, opened 2025-06-03
Improve error classification and add operationID to errors

### LINKED ISSUE #4738: Log operation id for requests

We do not log the operation ID (RequestID, ClientRequestID?)

## Describe the improvement

We should log at least one of the above.

## Additional context

This will help with debugging user-reported issues if we need to look at server-side logs.

matthchr:
Add operationID to errors from Azure in the logs (not in the Condition as we want the condition to be "stable" and not have a rotation UUID in it).

Fix a number of bugs around error formatting and classification
* Errors were not correctly classified by the DELETE operation.
* Conditions were including too much detail unintentionally.
* Error classification was lacking testing, which made the above issues harder to notice.

Fixes #4738.

## Checklist

- [x] this PR contains tests

### DIFF
--- a/v2/internal/errwrap/errwrap.go
+++ b/v2/internal/errwrap/errwrap.go
@@ -1,0 +1,31 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+package errwrap
+
+type hideError struct {
+	err error
+	msg string
+}
+
+var _ error = &hideError{}
+
+// Hide wraps an error with a message, hiding the original error.
+// This is useful when you want to provide a user-friendly message without exposing the underlying error details.
+// The original error can still be accessed via the Unwrap method (so errors.Is and errors.As work as expected).
+func Hide(err error, msg string) error {
+	return &hideError{
+		err: err,
+		msg: msg,
+	}
+}
+
+func (w *hideError) Error() string {
+	return w.msg
+}
+
+func (w *hideError) Unwrap() error {
+	return w.err
+}
--- a/v2/internal/genericarmclient/cloud_error.go
+++ b/v2/internal/genericarmclient/cloud_error.go
@@ -7,7 +7,10 @@ package genericarmclient
 
 import (
 	"encoding/json"
+	"fmt"
+	"net/http"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/rotisserie/eris"
 
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/core"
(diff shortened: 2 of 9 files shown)


matthchr:
I think I broke a test, need to understand why.

v2/internal/errwrap/errwrap.go
```diff
@@ -0,0 +1,31 @@
...
+	err error
+	msg string
+}
+
+var _ error = &wrapHider{}
+
+// HiddenWrapped wraps an error with a message, hiding the original error.
+// This is useful when you want to provide a user-friendly message without exposing the underlying error details.
+// The original error can still be accessed via the Unwrap method (so errors.Is and errors.As work as expected).
+func HiddenWrapped(err error, msg string) error {
```
theunrepentantgeek: Given the invocations will use the package name `errwrap` maybe rename this to `Hide` - usage would then be

``` go
errwrap.Hide(err, "Happy message")
```

which is kinda nice.

v2/internal/genericarmclient/cloud_error.go
```diff
@@ -38,7 +41,15 @@ func NewCloudError(err error) *CloudError {
...
-func NewTestCloudError(code string, message string) *CloudError {
+type TestOption func(e *CloudError)
+
+func WithTestInnerError(err error) TestOption {
+	return func(e *CloudError) {
+		e.error = err
+	}
+}
+
+func NewTestCloudError(code string, message string, options ...TestOption) *CloudError {
```
theunrepentantgeek, line 52: You're not using `options` here - is this why your test broke?

v2/internal/reconcilers/arm/error/error_classifier.go
```diff
@@ -3,7 +3,7 @@
  * Licensed under the MIT license.
  */
 
-package arm
+package error
```
theunrepentantgeek: Calling this package `error` might be unwise, given it conflits with the (always in scope) standard interface `error`. I even think one of our linters will flag this.

v2/internal/reconcilers/arm/error/ready_condition_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	// Check that a new ReadyConditionImpactingError is created with appropriate severity
+	readyErr, ok := conditions.AsReadyConditionImpactingError(result)
+	g.Expect(ok).To(BeTrue())
+	g.Expect(readyErr.Severity).To(Equal(conditions.ConditionSeverityWarning))
+	g.Expect(readyErr.Reason).To(ContainSubstring("Conflict"))
+
+	// Verify the cause contains our original error
+	g.Expect(eris.Unwrap(readyErr.Cause())).To(Equal(retryableErr))
+	g.Expect(readyErr.Error()).ToNot(ContainSubstring("This is an inner error"))
```
theunrepentantgeek: Suggest using `MatchError` - it will fail the assertion if the error is missing, and gives a better failure message if the substring isn't found. (Probably worth updating for every `err` assertion.)

```suggestion
	g.Expect(readyErr).To(MatchError(Not(ContainSubstring("This is an inner error"))))
```

v2/internal/genericarmclient/cloud_error.go
matthchr: DOH

v2/internal/errwrap/errwrap.go
matthchr: Yeah I also didn't love this name but didn't have a better one in the moment. Agree that just `hide` is a lot better.

v2/internal/genericarmclient/cloud_error.go
matthchr: Fixed

v2/internal/reconcilers/arm/error/ready_condition_test.go
matthchr: Fixed this here and elsewhere in this file.

v2/internal/reconcilers/arm/error/error_classifier.go
matthchr: Changed to `arm/errorclassification`

theunrepentantgeek (approved):

PR #4765, opened 2025-06-09
Improve type extraction from resource hierarchy for ownerDetails

fabian-heib:
## What this PR does
Improve type extraction from resource hierarchy for ownerDetails

Replaced fixed index-based type assertions with a dynamic type check loop over the hierarchy slice. This enhances robustness by allowing flexible ordering and avoiding panics on unexpected slice lengths.

Updated error messages to clearly indicate missing types instead of reporting incorrect type assertions.

## Why?

We have a case where the resource group is created within terraform so aso doesn't know it exists with objects. The armId exists and works but this breaks check which check's the amount of objects found. Change so it will check for the needed objects instead of checking a number. 

## Checklist

- [ ] this PR contains tests
- [ ] this PR has been run

### DIFF
--- a/v2/internal/reconcilers/azuresql/connector.go
+++ b/v2/internal/reconcilers/azuresql/connector.go
@@ -46,21 +46,24 @@ func getOwnerDetails(ctx context.Context, resourceResolver *resolver.Resolver, u
 		return ownerDetails{}, err
 	}
 
-	if len(hierarchy) != 3 {
-		return ownerDetails{}, eris.Errorf("failed to look up ownerDetails: expected resource hierarchy len=3 but was %d", len(hierarchy))
+	var server *sql.Server
+	var database *sql.ServersDatabase
+	for _, object := range hierarchy {
+		switch object.(type) {
+		case *sql.Server:
+			server = object.(*sql.Server)
+		case *sql.ServersDatabase:
+			database = object.(*sql.ServersDatabase)
+		default:
+		}
 	}
 
-	genericServer := hierarchy[1]
-	genericDatabase := hierarchy[2]
-
-	server, ok := genericServer.(*sql.Server)
-	if !ok {
-		return ownerDetails{}, eris.Errorf("owner's owner was not type Server, instead: %T", genericServer)
+	if server == nil {
+		return ownerDetails{}, eris.New("could not find a value of type Server in the hierarchy")
 	}
 
-	database, ok := genericDatabase.(*sql.ServersDatabase)
-	if !ok {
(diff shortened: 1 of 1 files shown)


fabian-heib:
@microsoft-github-policy-service agree company="Stakater"

matthchr:
This change looks reasonable to me - have you tested it? Can you talk about how?

I don't think any of our existing tests cover this case, but the updated code looks correct to me.

Thanks for the PR!

fabian-heib:
> This change looks reasonable to me - have you tested it? Can you talk about how?
> 
> I don't think any of our existing tests cover this case, but the updated code looks correct to me.
> 
> Thanks for the PR!

I have not tested it, Im thinking for me it would be easier to add a specific test for this function instead of trying to run a build version of ASO, might have some time during this week or early next week

matthchr:
Would love to see a test for this but absent that, we're releasing soon and this code looks right to me. I'm going to run the tests on this and merge it, it should get published as an experimental image and you can try that out with your scenario and confirm it works end to end. Assuming it does. great -- if not we're no worse than before.

We _do_ have a test for the "standard" owner hierarchy so if that test fails we won't merge - but I don't expect it to fail.

fabian-heib:
> Would love to see a test for this but absent that, we're releasing soon and this code looks right to me. I'm going to run the tests on this and merge it, it should get published as an experimental image and you can try that out with your scenario and confirm it works end to end. Assuming it does. great -- if not we're no worse than before.
> 
> We _do_ have a test for the "standard" owner hierarchy so if that test fails we won't merge - but I don't expect it to fail.

I still would like to add a test but can do it in a seperate PR. 
Trying to setup a resolver seems impossible somehow. I tried writing a small test around the `getOwnerDetails` but can't seem to create a resolver at all, tried with the `testcommon` package but that creates a import cycle. Any tips on how to create test resolver? Or should I maybe move the test to other testing package?

matthchr:
You can see an example of constructing a resolver here: [v2/internal/resolver/resolver_test.go](https://github.com/Azure/azure-service-operator/blob/main/v2/internal/resolver/resolver_test.go#L44), as well as the one in testcommon that I believe you found already.

If you aren't able to write a test for this, that's OK - fully admit that it's not that easy to test this scenario currently. If you don't plan on adding one, can you open an issue for adding a test for this scenario and link to this PR, and I can add one sometime next week or the week after.

In the meantime going to merge this to make sure that it gets into the next release, as I believe the code is correct.

matthchr (approved):

PR #4766, opened 2025-06-10
Add README files for many packages

theunrepentantgeek:
## What this PR does

I've long wanted to write an overview README for most of our packages, as a way to give new contributors a quick _lay of the land_ when they start reading code and working out where to make their changes.

I've used GitHub CoPilot to write these, with a manual review pass for accuracy and clarity.

### Special notes

Docs are driven by the `package-documentation.instructions`

gif

## Checklist


v2/pkg/genruntime/PACKAGE.md
```diff
@@ -0,0 +1,49 @@
+# Genruntime Package
```
matthchr, line 1: There's a standard with doc.go (see https://go.dev/blog/godoc) -- maybe we should be using that rather than `PACKAGE.md`?

theunrepentantgeek: Agreed. I'll abandon this PR and create a new one.

PR #4768, opened 2025-06-10
Add support for Entra SecurityGroup

theunrepentantgeek:
## What this PR does

Support for the first Entra based resource, SecurityGroup, including related infrastructure and tests.

gif

## Checklist

- [x] this PR contains tests
- [ ] Exporting EntraID to a config map for consumption elsewhere


v2/api/entra/v1/securitygroup_types.go
```diff
@@ -0,0 +1,186 @@
...
+const (
+	SecurityGroupMembershipTypeAssigned SecurityGroupMembershipType = "assigned"
+	SecurityGroupMembershipTypeDynamic  SecurityGroupMembershipType = "dynamic"
+)
+
+type SecurityGroupOperatorSpec struct {
+	// CreationMode: Specifies how ASO will try to create the resource.
+	// If not specified, defaults to "AdoptOrCreate".
+	// +kubebuilder:default=AdoptOrCreate
+	// +kubebuilder:validation:Enum=AdoptOrCreate;AlwaysCreate
```
matthchr: minor: I think you can put this on the enum itself rather than the property?
See for example [v2/api/app/v1api20240301/arm/container_app_spec_types_gen.go](https://github.com/Azure/azure-service-operator/blob/main/v2/api/app/v1api20240301/arm/container_app_spec_types_gen.go#L189)

matthchr: Oh and actually you do this above on MembershipType

v2/api/entra/v1/creation_mode.go
```diff
@@ -0,0 +1,30 @@
...
+// Licensed under the MIT license.
+
+package v1
+
+// CreationMode specifies how ASO will try to create the Entra resource.
+type CreationMode string
+
+const (
+	// AlwaysCreate means that ASO will always attempt to create the resource,
+	// regardless of whether it already exists, never trying to adopt an existing resource.
```
matthchr: "regardless of whether it already exists" is somewhat confusing.

How do I create a resource that already exists? I think this probably has something to do with displayNames and the fact that the server generates the "real" name - but it might be useful to describe that a bit more here? Or else maybe on the field where this is used... or both?

v2/api/entra/v1/securitygroup_types.go
```diff
@@ -0,0 +1,186 @@
...
+	DisplayName *string `json:"displayName,omitempty"`
+
+	// MailNickname: The email address of the group.
+	// +kubebuilder:validation:Required
+	MailNickname *string `json:"mailNickname,omitempty"`
+
+	// Description: The description of the group.
+	Description *string `json:"description,omitempty"`
+
+	// MembershipType: The membership type of the group.
```
matthchr, line 70: What does assigned vs dynamic do? Maybe clarify here would be useful?

v2/api/entra/v1/securitygroup_types.go
```diff
@@ -0,0 +1,186 @@
...
+		// Default is AdoptOrCreate
+		return true
+	}
+
+	return spec.CreationMode.AllowsAdoption()
+}
+
+type SecurityGroupOperatorSecrets struct {
+	// EntraID: The Entra ID of the group.
+	EntraID *genruntime.SecretDestination `json:"entraID,omitempty"`
```
matthchr: Why as a secret versus as a configmap? This doesn't actually seem "secret-y" to me?

v2/api/entra/v1/securitygroup_types.go
```diff
@@ -0,0 +1,186 @@
...
+	// +kubebuilder:default=AdoptOrCreate
+	// +kubebuilder:validation:Enum=AdoptOrCreate;AlwaysCreate
+	CreationMode *CreationMode `json:"creationMode,omitempty"`
+
+	Secrets *SecurityGroupOperatorSecrets `json:"secrets,omitempty"`
+}
+
+// CreationAllowed checks if the creation mode allows ASO to create a new security group.
+func (spec *SecurityGroupOperatorSpec) CreationAllowed() bool {
+	if spec.CreationMode == nil {
```
matthchr, line 191: Could set this with a mutating webhook on the way in, same as we do for AzureName, which might not be a bad idea as it helps users "see" it more obviously when they view the resource later?

Was there a reason you avoided that?

v2/api/entra/v1/webhook/securitygroup_webhook_types.go
```diff
@@ -0,0 +1,139 @@
...
+
+	return typedObj, nil
+}
+
+// defaultImpl applies defaults to the SecurityGroup resource
+func (webhook *SecurityGroup_Webhook) defaultImpl(
+	_ context.Context,
+	securityGroup *v1.SecurityGroup,
+) error {
+	return nil
```
matthchr: The logic for defaulting I mentioned above could go here.

v2/cmd/asoctl/go.mod
```diff
@@ -106,37 +97,36 @@ require (
 	github.com/mattn/go-isatty v0.0.20 // indirect
 	github.com/mattn/go-runewidth v0.0.16 // indirect
 	github.com/microsoft/go-mssqldb v1.8.2 // indirect
-	github.com/moby/spdystream v0.5.0 // indirect
+	github.com/microsoft/kiota-abstractions-go v1.9.1 // indirect
```
matthchr, line 100: Huh, TIL
https://github.com/microsoft/kiota

There's a whole separate SDK team and generator for Graph =P

v2/cmd/controller/app/setup.go
```diff
@@ -44,15 +44,17 @@ import (
 	"github.com/Azure/azure-service-operator/v2/internal/genericarmclient"
 	"github.com/Azure/azure-service-operator/v2/internal/identity"
 	asometrics "github.com/Azure/azure-service-operator/v2/internal/metrics"
+	"github.com/Azure/azure-service-operator/v2/internal/reconcilers/arm"
```
matthchr: duplicate import ,see `armreconciler` below.

v2/cmd/controller/app/setup.go
```diff
@@ -44,15 +44,17 @@ import (
...
 	armreconciler "github.com/Azure/azure-service-operator/v2/internal/reconcilers/arm"
+	"github.com/Azure/azure-service-operator/v2/internal/reconcilers/entra"
```
matthchr: Duplicate import, see `entrareconciler` below

v2/cmd/controller/app/setup.go
```diff
@@ -403,9 +418,11 @@ func initializeClients(cfg config.Values, mgr ctrl.Manager) (*clients, error) {
 
 	genericarmclient.AddToUserAgent(cfg.UserAgentSuffix)
 
-	var connectionFactory armreconciler.ARMConnectionFactory = func(ctx context.Context, obj genruntime.ARMMetaObject) (armreconciler.Connection, error) {
```
matthchr, line 406: huh, I have no idea why we bothered with this - good catch.

v2/internal/controllers/controller_resources.go
```diff
@@ -32,12 +32,16 @@ import (
...
 	postgresqlv1webhook "github.com/Azure/azure-service-operator/v2/api/dbforpostgresql/v1/webhook"
+	entrav1 "github.com/Azure/azure-service-operator/v2/api/entra/v1"
+	entrav1webhook "github.com/Azure/azure-service-operator/v2/api/entra/v1/webhook"
 	azuresqlv1 "github.com/Azure/azure-service-operator/v2/api/sql/v1"
 	azuresqlv1webhook "github.com/Azure/azure-service-operator/v2/api/sql/v1/webhook"
 	"github.com/Azure/azure-service-operator/v2/internal/identity"
 	"github.com/Azure/azure-service-operator/v2/internal/reconcilers"
 	"github.com/Azure/azure-service-operator/v2/internal/reconcilers/arm"
 	azuresqlreconciler "github.com/Azure/azure-service-operator/v2/internal/reconcilers/azuresql"
+	"github.com/Azure/azure-service-operator/v2/internal/reconcilers/entra"
```
matthchr: duplicate entra reconcile imports.

v2/internal/controllers/controller_resources.go
```diff
@@ -54,20 +58,25 @@ type Schemer interface {
 	GetScheme() *runtime.Scheme
 }
 
+type ClientsProvider struct {
```
matthchr, line 60: minor: You could change this to an interface definition that required the following methods:
```
func (c *clients) KubeClient() kubeclient.Client {
	return c.kubeClient
}

func (c *clients) ARMConnectionFactory() arm.ARMConnectionFactory {
	return c.armConnectionFactory
}

func (c *clients) EntraConnectionFactory() entra.EntraConnectionFactory {
	return c.entraConnectionFactory
}
```

(without the impls, obviously) - that would enable us to just pass `clients` into the helper rather than move the fields to a different type, and might be cleaner?

v2/internal/controllers/controller_resources.go
```diff
@@ -275,27 +305,42 @@ func getControllerName(obj client.Object) (string, error) {
...
+	// MySQL
 	knownTypes = append(
 		knownTypes,
 		&registration.KnownType{
 			Obj:       &mysqlv1.User{},
 			Defaulter: &mysqlv1webhook.User_Webhook{},
 			Validator: &mysqlv1webhook.User_Webhook{},
 		})
+
+	// Entra
```
matthchr: Extremely minor: Consider making this order and the order above the same (entra at the bottom)? Or we can refactor it to be alphabetical and make them match -- I'm fine either way but I think consistency between the two would be good.

v2/go.sum
```diff
@@ -433,6 +454,8 @@ github.com/spf13/pflag v1.0.5/go.mod h1:McXfInJRrz4CZXVZOBLb0bTZqETkiAhM9Iw0y3An
 github.com/spf13/pflag v1.0.6 h1:jFzHGLGAlb3ruxLB8MhbI6A8+AQX/2eW4qeyNZXNp2o=
 github.com/spf13/pflag v1.0.6/go.mod h1:McXfInJRrz4CZXVZOBLb0bTZqETkiAhM9Iw0y3An2Bg=
 github.com/spf13/viper v1.8.1/go.mod h1:o0Pch8wJ9BVSWGQMbra6iw0oQ5oktSIBaujf1rJH9Ns=
+github.com/std-uritemplate/std-uritemplate/go/v2 v2.0.3 h1:7hth9376EoQEd1hH4lAp3vnaLP2UMyxuMMghLKzDHyU=
+github.com/std-uritemplate/std-uritemplate/go/v2 v2.0.3/go.mod h1:Z5KcoM0YLC7INlNhEezeIZ0TZNYf7WSNO0Lvah4DSeQ=
 github.com/stoewer/go-strcase v1.3.0 h1:g0eASXYtp+yvN9fK8sH94oCIk0fau9uV1/ZdJ0AVEzs=
 github.com/stoewer/go-strcase v1.3.0/go.mod h1:fAH5hQ5pehh+j3nZfvwdk2RgEgQjAoM8wodgtPmh1xo=
 github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
```
matthchr, line 461: Can't actually comment on the file that was renamed but, was it intentional to rename this file as part of this PR? v2/internal/controllers/authorization_roleassignment_20220401_test.go

It seems somewhat unrelated to the rest of the stuff here?

v2/go.mod
```diff
@@ -66,6 +66,11 @@ require (
 	sigs.k8s.io/yaml v1.4.0
 )
 
+require (
```
matthchr: huh: Why does this get its own `require` section? Shouldn't these just be in the above section?

v2/internal/controllers/entra_securitygroup_v1_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+	entra "github.com/Azure/azure-service-operator/v2/api/entra/v1"
+	"github.com/Azure/azure-service-operator/v2/internal/util/to"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+)
+
+func Test_Entra_SecurityGroup_v1_CRUD(t *testing.T) {
+	t.Parallel()
+
+	// g := NewGomegaWithT(t)
```
matthchr: minor: Commented code.

v2/internal/controllers/entra_securitygroup_v1_test.go
```diff
@@ -0,0 +1,60 @@
...
+	// ctx := context.Background()
+	tc := globalTestContext.ForTest(t)
+
+	securityGroup := &entra.SecurityGroup{
+		ObjectMeta: tc.MakeObjectMeta("securitygroup"),
+		Spec: entra.SecurityGroupSpec{
+			DisplayName:    to.Ptr("ASO Test Security Group"),
+			MailNickname:   to.Ptr("asotest-security-group"),
```
matthchr, line 30: Hmm, this isn't actually a valid email address. Is that OK? I mean I assume it must be.

v2/internal/controllers/entra_securitygroup_v1_test.go
```diff
@@ -0,0 +1,60 @@
...
+	tc.ListResources(secretList, client.InNamespace(tc.Namespace))
+	tc.Expect(secretList.Items).To(HaveLen(1))
+
+	// Save an update
+	old := securityGroup.DeepCopy()
+	securityGroup.Spec.DisplayName = to.Ptr("ASO Test Security Group Updated")
+	tc.PatchResourceAndWait(old, securityGroup)
+
+	// Delete the resource and wait for it to be gone
+	tc.DeleteResourceAndWait(securityGroup)
```
matthchr, line 61: Do we need an assertion that the resource is actually deleted in entra too?

For Azure we often do this:
```
	// Ensure that the resource was really deleted in Azure
	exists, _, err := tc.AzureClient.CheckExistenceWithGetByID(
		tc.Ctx,
		armId,
		string(dataprotection.APIVersion_Value))
	tc.Expect(err).ToNot(HaveOccurred())
```

It might be a good idea to do something like this with entra too, just to make sure it really was removed?

v2/internal/controllers/recordings/Test_Entra_SecurityGroup_v1_CRUD.yaml
```diff
@@ -0,0 +1,242 @@
...
+                - application/json;odata.metadata=minimal;odata.streaming=true;IEEE754Compatible=false;charset=utf-8
+            Odata-Version:
+                - "4.0"
+            Request-Id:
+                - 7167885b-a311-44de-b9dc-8d764b66b883
+            Strict-Transport-Security:
+                - max-age=31536000
+            Vary:
+                - Accept-Encoding
+            X-Ms-Ags-Diagnostic:
```
matthchr: minor: wonder if we should prune these two `X-Ms` headers in this case?

v2/api/entra/v1/securitygroup_types.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+type SecurityGroupSpec struct {
+	// DisplayName: The display name of the group.
+	// +kubebuilder:validation:Required
+	DisplayName *string `json:"displayName,omitempty"`
+
+	// MailNickname: The email address of the group.
+	// +kubebuilder:validation:Required
+	MailNickname *string `json:"mailNickname,omitempty"`
```
matthchr, line 65: In the JSON for the payload, I see that there is a `mailEnabled` field -- is that something that we can/should be exposing to users, to allow them to create non-mail enabled SGs?

matthchr: (and would doing so allow this MailNickname to be optional, or maybe not still?)

v2/internal/controllers/recordings/Test_Entra_SecurityGroup_v1_CRUD.yaml
```diff
@@ -0,0 +1,242 @@
...
+        body: '{"@odata.context":"https://graph.microsoft.com/v1.0/$metadata#groups/$entity","id":"6071e091-80dc-49e9-aea7-21f67858ea40","deletedDateTime":null,"classification":null,"createdDateTime":"2001-02-03T04:05:06Z","creationOptions":[],"description":null,"displayName":"ASO Test Security Group","expirationDateTime":null,"groupTypes":["Unified"],"isAssignableToRole":null,"mail":"asotest-security-group@microsoft.onmicrosoft.com","mailEnabled":true,"mailNickname":"asotest-security-group","membershipRule":null,"membershipRuleProcessingState":null,"onPremisesDomainName":null,"onPremisesLastSyncDateTime":null,"onPremisesNetBiosName":null,"onPremisesSamAccountName":null,"onPremisesSecurityIdentifier":null,"onPremisesSyncEnabled":null,"preferredDataLocation":"APC","preferredLanguage":null,"proxyAddresses":["SMTP:asotest-security-group@microsoft.onmicrosoft.com"],"renewedDateTime":"2001-02-03T04:05:06Z","resourceBehaviorOptions":[],"resourceProvisioningOptions":[],"securityEnabled":true,"securityIdentifier":"S-1-12-1-1618075793-1240039644-4129400750-1089099896","theme":null,"uniqueName":null,"visibility":"Public","onPremisesProvisioningErrors":[],"serviceProvisioningErrors":[]}'
```
matthchr: Is the `securityIdentifier` field OK to log here?

matthchr: preferredDataLocation is also interesting, is that something users might want to be able to set?

v2/api/entra/v1/securitygroup_types.go
matthchr: Yeah Terraform has MailNickname as optional?
https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/resources/group - but notes that it is required for mail-enabled groups.

matthchr: Might be worth going through the Terraform options and determining which ones we want to expose in the initial configuration, if you haven't already. If some were purposefully omitted here, maybe call out why in either the PR description or a code comment?

v2/internal/controllers/recordings/Test_Entra_SecurityGroup_v1_CRUD.yaml
```diff
@@ -0,0 +1,242 @@
...
+        headers:
+            Cache-Control:
+                - no-cache
+            Content-Type:
+                - application/json;odata.metadata=minimal;odata.streaming=true;IEEE754Compatible=false;charset=utf-8
+            Location:
+                - https://graph.microsoft.com/v2/00000000-0000-0000-0000-000000000000/directoryObjects/6071e091-80dc-49e9-aea7-21f67858ea40/Microsoft.DirectoryServices.Group
+            Odata-Version:
+                - "4.0"
+            Request-Id:
```
matthchr: Interestingly we redact/remove:
```
	"X-Ms-Client-Request-Id",
	"Client-Request-Id",
```

but not just plain old `Request-Id`.

matthchr: We have a few (4 that I can see) Azure recordings with Request-Id in them too. Wondering if we should redact it as well?

v2/internal/controllers/entra_securitygroup_v1_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+	// Make sure the secret was created
+	secretList := &v1.SecretList{}
+	tc.ListResources(secretList, client.InNamespace(tc.Namespace))
+	tc.Expect(secretList.Items).To(HaveLen(1))
+
+	// Save an update
+	old := securityGroup.DeepCopy()
+	securityGroup.Spec.DisplayName = to.Ptr("ASO Test Security Group Updated")
+	tc.PatchResourceAndWait(old, securityGroup)
```
matthchr, line 55: Missing an assertion test that the new status reflects the new displayname?

v2/internal/reconcilers/entra/annotations.go
```diff
@@ -0,0 +1,64 @@
...
+
+// setEntraID stored the Entra ID in the annotations of the object.
+func setEntraID(obj v1.Object, id string) {
+	addAnnotation(obj, genruntime.ResourceIDAnnotation, id)
+}
+
+// addAnnotation adds the specified annotation to the object.
+// Empty string annotations are not allowed. Attempting to add an annotation with a value
+// of empty string will result in the removal of that annotation.
+func addAnnotation(obj v1.Object, key string, value string) {
```
matthchr: Why not use `genruntime.AddAnnotation`? I think it has these semantics

v2/pkg/genruntime/base_types.go
```diff
@@ -79,6 +79,11 @@ type ARMOwnedMetaObject interface {
 	ARMOwned
 }
 
+// EntraMetaObject represents an arbitrary ASO resource that is an Entra resource
```
matthchr, line 82: Is this to enable future expansion, as opposed to just using MetaObject directly?

v2/internal/reconcilers/entra/connection.go
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+package entra
+
+import (
+	msgraphsdk "github.com/microsoftgraph/msgraph-sdk-go"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+type Connection interface {
```
matthchr, line 13: minor, style, feel free to ignore: You could define this alongside the entra_client rather than giving it its own file. IIRC that's a pretty common pattern in Go.

v2/internal/reconcilers/entra/entra_client_cache.go
```diff
@@ -0,0 +1,117 @@
...
+
+// EntraClientCache is a cache for entraClients to hold multiple credential clients and global credential client.
+type EntraClientCache struct {
+	lock sync.Mutex
+
+	clients            map[string]*entraClient // clients allows quick lookup of an entraClient for each namespace
+	cloudConfig        cloud.Configuration     //!! What's the entra equivalent?
+	credentialProvider identity.CredentialProvider
+	httpClient         *http.Client
+	//!! TODO entraMetrics         *metrics.EntraClientMetrics
```
matthchr: Is this actually remaining? Needs a GH issue filed, or PR update?

v2/internal/reconcilers/entra/entra_client_cache.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/Azure/azure-service-operator/v2/internal/identity"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+)
+
+// EntraClientCache is a cache for entraClients to hold multiple credential clients and global credential client.
+type EntraClientCache struct {
+	lock sync.Mutex
+
+	clients            map[string]*entraClient // clients allows quick lookup of an entraClient for each namespace
+	cloudConfig        cloud.Configuration     //!! What's the entra equivalent?
```
matthchr: I think you're right that this doesn't contain graph endpoint details. I am not sure what does but I imagine there is a different graph URL for different clouds, which is what should go here.

v2/internal/controllers/controller_resources.go
matthchr: Ah, but I do see that in other contexts like test, they'd still need to define their owns structs - so the approach that you have may be better.

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
```diff
@@ -0,0 +1,499 @@
...
+	log logr.Logger,
+	eventRecorder record.EventRecorder,
+	obj genruntime.MetaObject,
+) (ctrl.Result, error) {
+	group, err := r.asSecurityGroup(obj)
+	if err != nil {
+		return ctrl.Result{}, eris.Wrapf(err, "creating or updating security group %s", group.Name)
+	}
+
+	// TODO: Respect annotation serviceoperator.azure.com/reconcile-policy
```
matthchr: Should probably be done as part of this PR?

matthchr: Oh wait, you shouldn't actually need to do anything here - handling of this is taken care of in the `reconcilers/generic/generic_reconciler` -- see `handleSkipReconcile`. I was thinking it needed to be handled in each and every impl, but now that I am thinking about it I believe we refactored that so it was taken care of at the upper level.

v2/api/entra/v1/securitygroup_types.go
```diff
@@ -0,0 +1,186 @@
...
+
+	groupTypes := []string{"Unified"}
+	if spec.MembershipType != nil && *spec.MembershipType == SecurityGroupMembershipTypeDynamic {
+		groupTypes = append(groupTypes, "DynamicMembership")
+	}
+
+	model.SetGroupTypes(groupTypes)
+
+	// This is a security group, not a mail distribution group
+	model.SetMailEnabled(to.Ptr(false))
```
matthchr, line 123: wait... we `SetMailEnabled` here to false, but in the HTTP recording, it shows as `"mailEnabled":true` for the POST in the JSON - that doesn't seem right?

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
```diff
@@ -0,0 +1,499 @@
...
+			}, nil
+		}
+
+		return ctrl.Result{}, eris.Wrapf(err, "getting group by ID %s", id)
+	}
+
+	// Update - PATCH
+	group.Spec.AssignToGroup(g)
+
+	_, err = client.Client().Groups().ByGroupId(id).Patch(ctx, g, nil)
```
matthchr, line 201: Do you know what the semantics of this PATCH are? What happens if there's an optional field that the user had previously specified, and then removes from the specification. Now the PATCH will (presumably) not include that field in its payload - will it get cleared or will it be left?

We have some resources in Azure that behave like that too (cough: AKS), but it's not ideal for the IaC style work users will be doing here.

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
```diff
@@ -0,0 +1,499 @@
...
+
+	// Create our Entra Client
+	client, err := r.EntraClientFactory(ctx, group)
+	if err != nil {
+		return nil, eris.Wrap(err, "creating entra client prior to adoption search")
+	}
+
+	// Try to find the group by DisplayName
+	displayName := group.Spec.DisplayName
+	if displayName == nil || *displayName == "" {
```
matthchr: very minor, style: could be `to.FromPtr(displayName) == ""` I think

matthchr: Or above: `to.FromPtr(group.Spec.DisplayName)`

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
```diff
@@ -0,0 +1,499 @@
...
+
+	groups := result.GetValue()
+	return groups, nil
+}
+
+// isNotFound returns true if the error is a 404 error.
+func (r *EntraSecurityGroupReconciler) isNotFound(err error) bool {
+	var odataError *odataerrors.ODataError
+	if eris.As(err, &odataError) {
+		if odataError.ResponseStatusCode == 404 {
```
matthchr: minor: consider `== http.NotFound` here?

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
```diff
@@ -0,0 +1,499 @@
...
+	group.Spec.AssignToGroup(g)
+
+	status, err := client.Client().Groups().Post(ctx, g, nil)
+	if err != nil {
+		// Failed to create
+		return ctrl.Result{}, eris.Wrapf(err, "failed to create group %s", group.Name)
+	}
+
+	group.Status.AssignFromGroup(status)
+	setEntraID(group, *status.GetId())
```
matthchr: Is this ptr deref safe? Consider guarding it?

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
```diff
@@ -0,0 +1,499 @@
...
+// eventRecorder is used to record events for the group.
+// log is the logger to use for logging.
+// Returns the Entra ID of the group to adopt, and nil if found;
+// nil, nil if no group was found to adopt;
+// or nil, error if there was a problem finding the group.
+func (r *EntraSecurityGroupReconciler) tryAdopt(
+	ctx context.Context,
+	group *asoentra.SecurityGroup,
+	log logr.Logger,
+) (*string, error) {
```
matthchr: minor, style: could avoid str-ptr return here probably if you wanted and just return empty string if the ID wasn't found?

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
```diff
@@ -0,0 +1,499 @@
...
+) (ctrl.Result, error) {
+	group, err := r.asSecurityGroup(obj)
+	if err != nil {
+		return ctrl.Result{}, eris.Wrapf(err, "creating or updating security group %s", group.Name)
+	}
+
+	// TODO: Respect annotation serviceoperator.azure.com/reconcile-policy
+
+	// If we already know the Entra ID of the group (captured in an annotation), we can update it directly
+	if id, ok := getEntraID(obj); ok {
```
matthchr, line 80: minor, style: Could you refactor this to have one section which is:
// get the ID (either from `getEntraID` or `tryAdopt`), and then a single `r.update` call with that id?

Ignore this if that refactoring sucks for some other reason, the reason I ask is it just means that there's one path to update and one path to create, which is nice from a readability perspective.

matthchr (commented):
Looks great overall. A couple of questions and possibly one or two bugs?

v2/api/entra/v1/creation_mode.go
theunrepentantgeek: I'll update the description on both, let me know if you're happy with the changes.

v2/api/entra/v1/securitygroup_types.go
theunrepentantgeek: According to [the docs](https://learn.microsoft.com/en-us/graph/api/group-post-groups?view=graph-rest-1.0&tabs=http#request-body), there are four mandatory properties: `displayName`, `mailEnabled`, `mailNickname` and `securityEnabled`.

I don't know how Terraform can leave `MailNickname` as optional.

Most other properties are of interest to `DistributionGroups` - which are handled by the same resource in Entra - but I'm specifically creating a `SecurityGroup` resource.

v2/api/entra/v1/securitygroup_types.go
theunrepentantgeek: I don't know what it does, the API page doesn't specify it's behaviour.

v2/api/entra/v1/securitygroup_types.go
theunrepentantgeek: The request has it set to false:
``` json
{
	"@odata.type":"#microsoft.graph.group",
	"displayName":"ASO Test Security Group",
	"groupTypes":["Unified"],
	"mailEnabled":false,
	"mailNickname":"asotest-security-group",
	"securityEnabled":true
}
```

but it comes back as
``` json
	"mailEnabled":true,
```

Maybe the docs are wrong and `mailNickname` is only needed for mail enabled groups?

theunrepentantgeek: Given that security groups aren't supposed to be used as distribution lists, this would make sense. I'll remove `mailNickname`.

theunrepentantgeek: The plot thickens:
```
failed to create group asotest-securitygroup-bsehla: 
A value is required for property 'mailNickname' of resource 'Group'
```

It really is mandatory - if you're creating an M365 group. 
Optional if you're creating an old style group - but I don't have sufficient rights to test that.

v2/api/entra/v1/securitygroup_types.go
theunrepentantgeek: Moved to the type.

v2/api/entra/v1/securitygroup_types.go
theunrepentantgeek: Honestly? Didn't think of it.

v2/api/entra/v1/securitygroup_types.go
theunrepentantgeek: No particular reason, will change to configmap.

v2/api/entra/v1/webhook/securitygroup_webhook_types.go
theunrepentantgeek: Done.

v2/cmd/controller/app/setup.go
theunrepentantgeek: Fixed.

v2/cmd/controller/app/setup.go
theunrepentantgeek: Fixed.

v2/cmd/controller/app/setup.go
theunrepentantgeek: I bet it started more complicated and got simplified to this stage.

v2/go.mod
theunrepentantgeek: No idea. Will fix up.

v2/internal/controllers/controller_resources.go
theunrepentantgeek: Removed.

v2/internal/controllers/controller_resources.go
theunrepentantgeek: Moved into the same order. 

Aside: These methods are getting too big, going to tidy them up later.

v2/internal/controllers/entra_securitygroup_v1_test.go
theunrepentantgeek: Removed.

v2/internal/controllers/entra_securitygroup_v1_test.go
theunrepentantgeek: Can be specified either as the `nickname` (no domain) or `nickname@domain.com` (with domain).

v2/internal/controllers/entra_securitygroup_v1_test.go
theunrepentantgeek: Done.

v2/internal/controllers/entra_securitygroup_v1_test.go
theunrepentantgeek: Good idea, but I might leave that for an update PR - running out of time today and want you to be able to review over my weekend.

v2/internal/reconcilers/entra/annotations.go
theunrepentantgeek: Done.

v2/internal/reconcilers/entra/connection.go
theunrepentantgeek: I tend to favour putting interfaces into separate files so they're more discoverable by new readers trying to become familiar with the code.

v2/internal/reconcilers/entra/entra_client_cache.go
theunrepentantgeek: #4775

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
theunrepentantgeek: I thought that was how things work, but left this TODO to ensure we followed up. Thanks for confirming.

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
theunrepentantgeek: I originally had something along those lines but it was a bit convoluted handling all the edge cases correctly (especially around optional adoption); this refactoring was my attempt to make the flow easy to understand.

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
theunrepentantgeek: If omitted from the payload, the value isn't changed. 

I tried modifying `SecurityGroupSpec.AssignToGroup` to explicitly set every field, but that fails validation - it seems many properties can't be cleared, once set. 

Not sure how we should handle those. Terraform seems to delete/re-create, which seems drastic.

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
theunrepentantgeek: Good idea.

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
theunrepentantgeek: Package `"github.com/Azure/azure-service-operator/v2/internal/util/to"` doesn't have a FromPtr method. Leaving the way it is.

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
theunrepentantgeek: Added guard clause.

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
theunrepentantgeek: Done.

v2/pkg/genruntime/base_types.go
theunrepentantgeek: Yes; following the precedent of `ARMMetaObject`.

v2/go.sum
theunrepentantgeek: Must have been a lurking rename that got picked up accidentally.

v2/internal/controllers/recordings/Test_Entra_SecurityGroup_v1_CRUD.yaml
theunrepentantgeek: Good idea. Done.

v2/internal/controllers/recordings/Test_Entra_SecurityGroup_v1_CRUD.yaml
theunrepentantgeek: I think it's ok - aren't they equivalent to `clientID` etc - a randomly assigned GUID that's useless as soon as the resource is deleted?

theunrepentantgeek: It's not directly settable, as far as I can tell. From the docs:

>The value of the preferredDataLocation property is inherited from the group creator's preferred data location.

v2/internal/controllers/recordings/Test_Entra_SecurityGroup_v1_CRUD.yaml
theunrepentantgeek: Yes, I think we should. Changed.

v2/internal/controllers/recordings/Test_Entra_SecurityGroup_v1_CRUD.yaml
matthchr: That seems believable to me, just wanted to check because I wasn't sure. I don't really see why logging it would be an issue I just wanted to make sure =)

v2/internal/reconcilers/entra/entra_client_cache.go
matthchr: Resolving this, we can track it via the issue.

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
matthchr: Yeah delete+recreate is very awkward for sure, not something we should be doing IMO. Leaving this comment unresolved as it may be worth leaving a comment in the code w.r.t, but it wouldn't be the only resource that has this problem in theory (as I mentioned, AKS does in certain cases too, though usually there there is a way to force the thing off, rather than unsetting it)

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
matthchr: Ah sorry I got it mixed up, it's `to.Value`, `FromPtr` is from sambler/lo

v2/api/entra/v1/securitygroup_types.go
matthchr: It's also possible the Terraform docs are wrong and you actually have to set it for the API to accept it.

v2/internal/controllers/entra_securitygroup_v1_test.go
matthchr: Worth documenting on the Spec field something to that effect?

v2/api/entra/v1/creation_mode.go
matthchr: I like the rephrasing more, but still wonder if we want to describe more about how we treat the display name as the key for "already existing"

matthchr: Oh wait you say this more on the property (not the type). Carry on, happy with this then.

v2/api/entra/v1/securitygroup_types.go
matthchr: Ahh you're right I missed that that was the response not the request.

Weird that it seems to be ignored even if you set it to false, but nothing you can do about that.

matthchr (approved):

matthchr:
CI error though? Not totally clear to me what the CI error actually is though - looks like maybe a build issue in the entra folder but I can't tell what the actual build error is.

v2/internal/controllers/entra_securitygroup_v1_test.go
theunrepentantgeek: Done.

v2/internal/reconcilers/entra/entra_client_cache.go
theunrepentantgeek: As far as I can tell, it looks that while you _can_ specify a scope, the default behaviour is to discern the cloud from the Azure credential being passed in, so I don't think any specific action is needed. 

I've added a comment to this effect - and also documented what we'll need to do if that understanding is wrong.

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go

matthchr (approved):

PR #4789, opened 2025-06-19
Add support for 2025-06-01 CognitiveServices

### LINKED ISSUE #4618: Bug: Missing OpenAI Resource & Deployment Model Support in Azure Service Operator
**## Describe the bug**

Azure Service Operator (ASO) does not currently support provisioning Azure OpenAI resources (Microsoft.CognitiveServices/accounts) and Deployment Models (Microsoft.CognitiveServices/accounts/deployments). 
This limitation prevents managing OpenAI services via Kubernetes and blocks the adoption of a GitOps approach for deploying and managing these resources efficiently. Adding a Custom Resource Definition (CRD) for OpenAI would enable seamless deployment and lifecycle management through ASO.

**Azure Service Operator Version**: ASOv2

🔹 Description
Azure Service Operator (ASO) currently does not support provisioning Azure OpenAI resources (Microsoft.CognitiveServices/accounts) and Deployment Models (Microsoft.CognitiveServices/accounts/deployments). This feature is essential for managing OpenAI workloads via Kubernetes.

🔹 Expected Behavior
ASO should provide a CRD (Custom Resource Definition) for "Microsoft.CognitiveServices/accounts" and "Microsoft.CognitiveServices/accounts/deployments".
Users should be able to deploy and manage OpenAI resources through Kubernetes manifests.

🔹 Current Behavior
Attempting to apply a Kubernetes manifest for OpenAI resources results in "Unknown resource type" errors.
No ASO documentation exists for Microsoft.CognitiveServices/accounts and Microsoft.CognitiveServices/accounts/deployments.

## Screenshots

_If applicable, add screenshots to help explain your problem._

## Additional context

_Add any other context about the problem here._

JamesJuddAVEVA:
## What this PR does
Adds support for Azure's AI suite. Allows for different deployment models.
Closes #4618 

gif

## Checklist
- [x] this PR contains tests


JamesJuddAVEVA:
@microsoft-github-policy-service agree company="AVEVA"

v2/azure-arm.yaml
```diff
@@ -1530,6 +1530,64 @@ objectModelConfiguration:
       AFDOriginProperties:
         HostName:
           $importConfigMapMode: optional
+  cognitiveservices:
+    2024-10-01:
```
matthchr: There's a newer API version for this: 2025-06-01, as far as I can see here: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/resource-manager/Microsoft.CognitiveServices/stable/2025-06-01/cognitiveservices.json

Can you retarget this to use the latest API version? You'll need to update the sample names and re-record the tests using that newer API version - the rest of the code (the samples themselves and the test itself) you should just be able to swap the imports/swap the API version and they'll work.

v2/internal/controllers/cognitiveservices_account_20241001_test.go
```diff
@@ -0,0 +1,67 @@
...
+	rg := tc.CreateTestResourceGroupAndWait()
+
+	account := &cognitiveservices.Account{
+		ObjectMeta: tc.MakeObjectMetaWithName(tc.NoSpaceNamer.GenerateName("cogsvc")),
+		Spec: cognitiveservices.Account_Spec{
+			Identity: &cognitiveservices.Identity{
+				Type: to.Ptr(cognitiveservices.Identity_Type("SystemAssigned")),
+			},
+			Kind:     to.Ptr("AIServices"),
+			Location: to.Ptr("eastus"),
```
matthchr: Can you use `tc.AzureRegion` here?

If not, can you do:
```
	// Moving test to a different region:
	// "Sorry, we are currently experiencing high demand in West US 2 region, and cannot fulfill your request at this time"
	tc.AzureRegion = to.Ptr("eastus")
```

at the top of the test file (before rg create) rather than hardcode region here, and include a reason why it needs to be in eastus?

v2/internal/controllers/cognitiveservices_account_20241001_test.go
```diff
@@ -0,0 +1,67 @@
...
+func Test_CognitiveServices_Account_CRUD(t *testing.T) {
+	t.Parallel()
+	tc := globalTestContext.ForTest(t)
+	rg := tc.CreateTestResourceGroupAndWait()
+
+	account := &cognitiveservices.Account{
+		ObjectMeta: tc.MakeObjectMetaWithName(tc.NoSpaceNamer.GenerateName("cogsvc")),
+		Spec: cognitiveservices.Account_Spec{
+			Identity: &cognitiveservices.Identity{
+				Type: to.Ptr(cognitiveservices.Identity_Type("SystemAssigned")),
```
matthchr: very minor: I think there should be an autogenerated enum for this, like `cognitiveservices.Identity_Type_SystemAssigned` (or something like that).

v2/internal/controllers/cognitiveservices_account_20241001_test.go
```diff
@@ -0,0 +1,67 @@
...
+					Key2:     &genruntime.SecretDestination{Name: "cogsecrets", Key: "key2"},
+					Endpoint: &genruntime.SecretDestination{Name: "cogsecrets", Key: "endpoint"},
+					Endpoints: &genruntime.SecretDestination{
+						Name: "cogsecrets", Key: "endpoints",
+					},
+				},
+			},
+			Owner: testcommon.AsOwner(rg),
+			Properties: &cognitiveservices.AccountProperties{
+				PublicNetworkAccess: to.Ptr(cognitiveservices.AccountProperties_PublicNetworkAccess("Enabled")),
```
matthchr: Same minor comment, there should be an enum defined for this, something like `cognitiveservices.AccountProperties_PublicNetworkAccess_Enabled`

v2/internal/controllers/cognitiveservices_account_20241001_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+	account := &cognitiveservices.Account{
+		ObjectMeta: tc.MakeObjectMetaWithName(tc.NoSpaceNamer.GenerateName("cogsvc")),
+		Spec: cognitiveservices.Account_Spec{
+			Identity: &cognitiveservices.Identity{
+				Type: to.Ptr(cognitiveservices.Identity_Type("SystemAssigned")),
+			},
+			Kind:     to.Ptr("AIServices"),
+			Location: to.Ptr("eastus"),
+			OperatorSpec: &cognitiveservices.AccountOperatorSpec{
```
matthchr: You should have an assert like `tc.ExpectSecretHasKeys` that checks the `cogsecrets` secret has these keys, run after the `CreateResourcesAndWait`

v2/internal/controllers/cognitiveservices_deployment_20241001_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	. "github.com/onsi/gomega"
+
+	cognitiveservices "github.com/Azure/azure-service-operator/v2/api/cognitiveservices/v1api20241001"
+	"github.com/Azure/azure-service-operator/v2/internal/testcommon"
+	"github.com/Azure/azure-service-operator/v2/internal/util/to"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+)
+
+func Test_CognitiveServices_Deployment_CRUD(t *testing.T) {
```
matthchr: We usually do the parent + child resources in the same test, just to save on needing duplicate parents created.

I'm normally 100% in favor of each thing having its own test (clearer what causes the failure then), but due to the nature of parent-child resource relationships in Azure, this test is always basically a superset of the parent test so we combine them.

See for example v2/internal/controllers/containerservice_managedcluster_crud_v1api20240901_test.go, which uses:
```
	tc.RunParallelSubtests(
		testcommon.Subtest{
```

to run the child resource CRUD tests as a subtest of the parent.

v2/internal/controllers/cognitiveservices_deployment_20241001_test.go
```diff
@@ -0,0 +1,92 @@
...
+	tc.PatchResourceAndWait(old, deployment)
+	tc.Expect(deployment.Status.Properties.Model.Version).To(Equal(to.Ptr("2024-11-20")))
+
+	tc.DeleteResourceAndWait(deployment)
+	exists, _, err := tc.AzureClient.CheckExistenceWithGetByID(tc.Ctx, *deployment.Status.Id, string(cognitiveservices.APIVersion_Value))
+	tc.Expect(err).ToNot(HaveOccurred())
+	tc.Expect(exists).To(BeFalse())
+	tc.Expect(account.Status.Id).ToNot(BeNil())
+
+	tc.DeleteResourceAndWait(account)
```
matthchr: as mentioned above, if you move to the subtest style testing, this assert already happens in the parent test, so you don't need it in the child test.

v2/api/cognitiveservices/v1api20241001/structure.txt
```diff
@@ -0,0 +1,431 @@
...
+│   │   ├── ConfigMapExpressions: *core.DestinationExpression[]
+│   │   └── SecretExpressions: *core.DestinationExpression[]
+│   ├── Owner: *genruntime.KnownResourceReference
+│   ├── Properties: *Object (7 properties)
+│   │   ├── CapacitySettings: *Object (2 properties)
+│   │   │   ├── DesignatedCapacity: Validated<*int> (1 rule)
+│   │   │   │   └── Rule 0: Minimum: 0
+│   │   │   └── Priority: Validated<*int> (1 rule)
+│   │   │       └── Rule 0: Minimum: 0
+│   │   ├── CurrentCapacity: *int
```
matthchr: Is this field actually writable? Its name makes me think it's readonly? I'm wondering if we should prune it from the spec?

v2/azure-arm.yaml
```diff
@@ -1530,6 +1530,64 @@ objectModelConfiguration:
...
+      Account:
+        $exportAs: Account
+        $supportedFrom: v2.15.0
+        $azureGeneratedSecrets:
+          - Key1
```
matthchr, line 1545: These fields unfortunately don't magically get handled. This AzureGeneratedSecrets requires that you implement a hook/handler like [Azure Storage does](https://github.com/Azure/azure-service-operator/blob/main/v2/api/storage/customizations/storage_account_extensions.go#L34), where we call the API that serves the secrets and map those secrets into the underlying secret.

I'm not seeing where you've done that here.

matthchr (commented):
Thanks for this PR! It looks great.

I left a few comments. I couldn't comment on the file itself but you've committed the aso-gen tool in v2/tools/generator - can you please remove that?

v2/api/cognitiveservices/v1api20241001/structure.txt
JamesJuddAVEVA: Played around with some PUTs, Azure ignores it so definitely should be readonly. Pruned and resolved :)

v2/api/cognitiveservices/customizations/account_extension.go
```diff
@@ -0,0 +1,149 @@
...
+		resp, err := acctClient.ListKeys(ctx, id.ResourceGroupName, typedObj.AzureName(), nil)
+		if err != nil {
+			return nil, eris.Wrapf(err, "Failed listing keys")
+		}
+
+		keys[accountKey1] = to.Value(resp.Key1)
+		keys[accountKey2] = to.Value(resp.Key2)
+	}
+
+	keys[accountEp] = getSafeString(typedObj.Status.Properties.Endpoint)
```
matthchr: minor: We have a helper for this: v2/internal/util/to/to.go
`Value[T any](v *T) T` -- use `to.Value(...)`

```suggestion
	keys[accountEp] = to.Value(typedObj.Status.Properties.Endpoint)
```

v2/api/cognitiveservices/customizations/account_extension.go
```diff
@@ -0,0 +1,149 @@
...
+	storage "github.com/Azure/azure-service-operator/v2/api/cognitiveservices/v1api20250601/storage"
+	"github.com/Azure/azure-service-operator/v2/internal/genericarmclient"
+	"github.com/Azure/azure-service-operator/v2/internal/set"
+	"github.com/Azure/azure-service-operator/v2/internal/util/to"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/secrets"
+)
+
+const (
+	accountKey1 = "Key1"
```
matthchr: should be lowercase characters to start with.

So `key`, `key2`, etc

v2/api/cognitiveservices/customizations/account_extension.go
matthchr: Actually, I think better to not export endpoint or endpoints in the secret at all?

If you want them in the generated secret (which is totally reasonable), you can get them with the following CEL expression, which I think is a bit clearer about where the value is actually coming from (and saves us the work of having to manage the endpoint as a secret). This is most notably a better experience for the `map[string]string` endpoints field, which can avoid the need for JSON serialization and is a much cleaner user experience when using CEL expressions via `secretExpressions`

```
spec:
  operatorSpec:
    secret:
      key1:
        name: my-secret
        key: key1
      key2:
        name: my-secret
        key: key2
    secretExpressions:
    - name: my-secret
      key: endpoint
      value: self.status.properties.endpoint
    - name: my-secret
      value: self.status.properties.endpoints  # No key here, means dump the whole thing with the keys/values of this map becoming keys/values of the secret
```

OR (using only secretExpressions)
```
spec:
  operatorSpec:
    secretExpressions:
    - name: my-secret
      key: key1
      value: secret.key1
    - name: my-secret
      key: key2
      value: secret.key2
    - name: my-secret
      key: endpoint
      value: self.status.properties.endpoint
    - name: my-secret
      value: self.status.properties.endpoints  # No key here, means dump the whole thing with the keys/values of this map becoming keys/values of the secret
```

v2/internal/controllers/cognitiveservices_account_20250601_test.go
```diff
@@ -0,0 +1,110 @@
...
+			Identity: &cognitiveservices.Identity{
+				Type: to.Ptr(cognitiveservices.Identity_Type_SystemAssigned),
+			},
+			Kind:     to.Ptr("OpenAI"),
+			Location: tc.AzureRegion,
+			OperatorSpec: &cognitiveservices.AccountOperatorSpec{
+				Secrets: &cognitiveservices.AccountOperatorSecrets{
+					Key1:      &genruntime.SecretDestination{Name: "cogsecrets", Key: "key1"},
+					Key2:      &genruntime.SecretDestination{Name: "cogsecrets", Key: "key2"},
+					Endpoint:  &genruntime.SecretDestination{Name: "cogsecrets", Key: "endpoint"},
```
matthchr: minor: as mentioned, would need to change this slightly if you take my suggestion for secret changes above.

v2/internal/controllers/cognitiveservices_account_20250601_test.go
```diff
@@ -0,0 +1,110 @@
...
+		oldAcc.Spec.Tags = map[string]string{}
+	}
+	account.Spec.Tags = map[string]string{"env": "test"}
+	tc.PatchResourceAndWait(oldAcc, account)
+	tc.Expect(account.Status.Tags).To(Equal(map[string]string{"env": "test"}))
+
+	tc.RunParallelSubtests(
+		testcommon.Subtest{
+			Name: "CognitiveServices Deployment CRUD",
+			Test: func(tc *testcommon.KubePerTestContext) {
```
matthchr: minor: Can you put this subtest into its own named function?

You still call it here just do: `Test: funcName` rather than the whole function def.

matthchr (commented):
Looks great - thanks for updating to the latest API version.

I had a few minor comments on the secret export implementation. Rest looks great. Kicked off tests to make sure things are passing and surface any issues that need to be fixed.

JamesJuddAVEVA:
Thanks for the comments so far Matt, you've been very helpful!

Regarding using secretExpressions for Endpoints, the keys contain spaces so the default CEL mapping fails the kubernetes regex. As far as I can tell ASO doesn't support transformation in CELs so I implemented a mapping in the extension. Let me know if you'd prefer me to revert it back to an azureGeneratedSecret or remove it altogether. Cheers :)

v2/api/cognitiveservices/customizations/account_extension.go
```diff
@@ -135,15 +120,25 @@ func secretsToWrite(obj *storage.Account, keys map[string]string) ([]*v1.Secret,
...
 	collector.AddValue(operatorSpecSecrets.Key2, keys[accountKey2])
-	collector.AddValue(operatorSpecSecrets.Endpoint, keys[accountEp])
-	collector.AddValue(operatorSpecSecrets.Endpoints, keys[accountEps])
+
+	for k, v := range keys {
+		if k == accountKey1 || k == accountKey2 {
+			continue
+		}
+		collector.AddValue(&genruntime.SecretDestination{
+			Name: operatorSpecSecrets.Key1.Name,
```
matthchr: This isn't correct: If we support endpoints as a value output, users should opt-in to it and have the ability to specify its destination (like we do for CEL), rather than assuming that it's always the location of key1.

Take a look at my comment further down though, I think I have a way to use CEL to do this (curious as to your thoughts on usage), which would allow you to remove this bit entirely from here.

v2/internal/controllers/cognitiveservices_account_20250601_test.go
```diff
@@ -36,10 +37,18 @@ func Test_CognitiveServices_Account_20250601_CRUD(t *testing.T) {
...
+					Key1: &genruntime.SecretDestination{Name: "cogsecrets", Key: "key1"},
+					Key2: &genruntime.SecretDestination{Name: "cogsecrets", Key: "key2"},
+				},
+				SecretExpressions: []*core.DestinationExpression{
+					{
+						Name:  "cogsecrets",
+						Key:   "endpoint",
+						Value: "self.status.properties.endpoint",
+					},
+					// Keys in 'Endpoints' contain spaces so default CEL mapping fails kubernetes regex.
```
matthchr: Huh, this hasn't come up before.

CEL has weak support for map transformations, unfortunately. We _do_ support writing more complex expressions, but it needs to evaluate to a `map[string]string` in CEL at the end of the day and there aren't many macros in CEL that do that (in fact I'm pretty sure there aren't any).

I see a couple options for fixes:
1. ASO just performs this transformation automatically.
2. ASO provides our own `mapKeys` macro, which allows you to write this yourself, with something like: `value: "self.status.properties.endpoints.mapKeys(k, k.join('-'))"`
3. We augment the parsing/validation rules to allow lists of key-value pairs to be rendered into k8s map-like objects (secrets/configmaps), so you could then do something like `value: "self.status.properties.map(k, {k.join(' '): self.status.properties.endpoints[k]})`

I don't love 1 because it's a bit implicit, and maybe there are other cases where users want their spaces replaced by something else (or trimmed, if at the end?), I'd sorta rather give users control I _think_, even though it does end up making more work for users when rendering the keys (it doesn't just work)

Do you have thoughts on the above?

I'd rather do one of those, than do the trick you did with endpoints here. WYDT? I'm open to fixing/improving this in the next few days to unblock you.

matthchr: @theunrepentantgeek do you have opinions here?

I think I prefer option 2, as long as we can come up with a macro name we like. It gets the job done and might be useful elsewhere (Azure tags?)

theunrepentantgeek: The key problem with Option 1 would be the surprise factor - the behaviour wouldn't be obvious to anyone looking at the resource definition, they would either have to _just know_ what happens, or manage to find it in our docs.

Option 2 would be workable, but I'm worried about potential edge cases that would be difficult to handle in CEL (e.g. there's nothing in `join('-')` to prevent a result that starts or ends with `-`, which is still an illegal identifier.

What about a variation of Option 2 where the parameter specifies one of a fixed set of transformations?

``` cel
self.status.properties.endpoints.mapKeys("kebab-case")
```

| Style | Description | Sample |
| -- | -- | -- |
| concat | Remove all invalid characters and concatenate the rest | `quickbrownfox` |
| kebab-case | Separate words with a hyphen (`-`) | `quick-brown-fox` |
| snake_case | Separate words with an underscore (`_`) | `quick_brown_fox` |
| dot.case | Separate words with dots (`.`) | `quick.brown.fox` |

Not saying this is the right set of transformations, just illustrating the idea. Using names that match the style helps with recollection and recognition.

JamesJuddAVEVA: I like theunrepentantgeek's idea of have a fixed set of transformations. If a user needs a new transform they can open a ticket or PR the transform themselves using the ones we create as a template.

I'm happy to give this a try - before looking into exactly how I would implement this I estimate it'll take me at least a day or two. I'm going away next Wednesday so if you're keen to implement this before the next release I'd be happy to hand over this little bit to you guys instead.

Matt I'm on UK time hence the lack of overlap!

matthchr: I looked a bit more and actually CEL has the functions I was thinking of defined in an extension library we're just not including right now: https://github.com/google/cel-go/tree/master/ext#twovarcomprehensions.

I'll see if I can prototype a PR that adds that (safely) to our CEL env

matthchr: Usage is `self.metadata.annotations.transformMapEntry(k, v, {k.replace(" ", "-"): v})`

matthchr: PR is here: https://github.com/Azure/azure-service-operator/pull/4812

matthchr: This has merged. I think you should be able to update this to do the same thing for endpoint/endpoints, which is to say you don't need to do anything with endpoints in the secret exporter code. Just handle the secrets there and use CEL for the rest.

matthchr:
Did you intend to close this?

It was basically ready to merge except that one remaining (small) fix in the endpoints handling.

JamesJuddAVEVA:
That was an accident while rebasing! PR should now be ready to go :)

v2/api/cognitiveservices/customizations/account_extension.go
```diff
@@ -0,0 +1,127 @@
...
+	operatorSpecSecrets := obj.Spec.OperatorSpec.Secrets
+	if operatorSpecSecrets == nil {
+		return nil, nil
+	}
+
+	collector := secrets.NewCollector(obj.Namespace)
+	collector.AddValue(operatorSpecSecrets.Key1, keys[accountKey1])
+	collector.AddValue(operatorSpecSecrets.Key2, keys[accountKey2])
+
+	for k, v := range keys {
```
matthchr: I think this code is now dead, because there are no keys but key1/key2, and they're ignored? So we can just delete it?

v2/internal/controllers/recordings/Test_Samples_CreationAndDeletion/Test_Compute_v1api_CreationAndDeletion.yaml
```diff
@@ -1,10250 +0,0 @@
----
-version: 2
```
matthchr: I don't think that deleting this file was intentional?

v2/samples/cognitiveservices/v1api20250601/v1api20250601_account.yaml
```diff
@@ -0,0 +1,29 @@
...
+        key: key1
+      key2:
+        name: sample-secrets
+        key: key2
+    secretExpressions:
+      - name: sample-secrets
+        key: endpoint
+        value: self.status.properties.endpoint
+      - name: sample-secrets
+        value: 'self.status.properties.endpoints.transformMapEntry(k, v, {k.replace(" ", "-").lowerAscii(): v})'
```
matthchr, line 23: Nice, I think having this here will help users avoid having to figure this out for themselves.

matthchr (commented):
This LGTM. I'm ready to approve, I just think you need to revert the delete of that one recording file for the tests to pass.

v2/internal/controllers/recordings/Test_Samples_CreationAndDeletion/Test_Compute_v1api_CreationAndDeletion.yaml
JamesJuddAVEVA: You're correct!

matthchr:
Thanks for this contribution!

PR #4791, opened 2025-06-20
Add startup probe to controller manager

nojnhuh:
## What this PR does

I spend quite a bit of time spinning up CAPZ management clusters which include ASO, and I noticed that the `initialDelaySeconds` on the readiness probe is bottlenecking that process where I usually see ASO finish starting up after about 10s, then idle for an extra ~45s waiting for the first readiness probe.

I see the `initialDelaySeconds` was increased in #2844, but a startup probe should be able to account for longer startup times without holding things up if things are actually ready more quickly.

gif

### DIFF
--- a/v2/charts/azure-service-operator/templates/apps_v1_deployment_azureserviceoperator-controller-manager.yaml
+++ b/v2/charts/azure-service-operator/templates/apps_v1_deployment_azureserviceoperator-controller-manager.yaml
@@ -208,7 +208,6 @@ spec:
           httpGet:
             path: /healthz
             port: 8081
-          initialDelaySeconds: 60
         name: manager
         ports:
         - containerPort: {{ .Values.webhook.port }}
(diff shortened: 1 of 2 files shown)


matthchr:
Can you make this change in the [Helm chart](https://github.com/Azure/azure-service-operator/blob/main/v2/charts/azure-service-operator/templates/apps_v1_deployment_azureserviceoperator-controller-manager.yaml#L223) too?

We have a test that asserts that the raw YAML and Helm chart render to the same results, so make sure to make it in the same spot and also remove the initialDelay there as well.

edit: Thanks for the fix btw!

nojnhuh:
> Can you make this change in the [Helm chart](https://github.com/Azure/azure-service-operator/blob/main/v2/charts/azure-service-operator/templates/apps_v1_deployment_azureserviceoperator-controller-manager.yaml#L223) too?

Done, thanks!

matthchr (approved):

PR #4817, opened 2025-07-15
Improve ownership assignment to

theunrepentantgeek:
## What this PR does

Improves the way we handle resource ownership assignment to handle the case where the parent resource doesn't exist in the same version as the child resource - as encountered with some of the `Microsoft.Insights` resources.

### Special notes

I think I've retained the same behaviour for `Microsoft.SimpleStor` as we had before, just without special casing. Worth giving that a closer look than usual.

gif

### DIFF
--- a/v2/tools/generator/internal/codegen/pipeline/determine_resource_ownership.go
+++ b/v2/tools/generator/internal/codegen/pipeline/determine_resource_ownership.go
@@ -156,50 +156,74 @@ func (o *ownershipStage) updateChildResourceDefinitionsWithOwner(
 	owningResourceName astmodel.InternalTypeName,
 	updatedDefs astmodel.TypeDefinitionSet,
 ) error {
+	applyOwnerToChild := func(
+		owner astmodel.InternalTypeName,
+		def astmodel.TypeDefinition,
+		rsrc *astmodel.ResourceType,
+	) {
+		// Update the child resource to have the requested owner
+		rsrc = rsrc.WithOwner(owner)
+		def = def.WithType(rsrc)
+
+		// We update the definition set with overwrite semantics, as we may have already added this child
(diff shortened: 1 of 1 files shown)


v2/tools/generator/internal/codegen/pipeline/determine_resource_ownership.go
```diff
@@ -156,50 +156,74 @@ func (o *ownershipStage) updateChildResourceDefinitionsWithOwner(
...
+			applyOwnerToChild(owningResourceName, childResourceDef, childResource)
+			continue
 		}
 
-		// Update the definition of the child resource type to point to its owner
-		childResource, ok := childResourceDef.Type().(*astmodel.ResourceType)
-		if !ok {
-			return eris.Errorf("child resource %s not of type *astmodel.ResourceType, instead %T", typeName, childResourceDef.Type())
+		// if the child already has an owner in the same package as the child,
+		// keep that owner unchanged
```
matthchr, line 215: Maybe worth expanding this comment here to talk about the weird storsimple case in addition to the "standard" case where the owner is from a differently versioned package?

Maybe something like:
```suggestion
		// keep that owner unchanged. In addition to the standard case of a child having possible owners from multiple package versions
		// where we want to prefer the same package version, providers like StorSimple have the same URIs on multiple 
		// "different" types, so we need to match by API version.
```

Not sure if here is best or up above is best, but basically keeping the comment about StorSimple being weird and the fact that we really need to make sure those line up by version (even if it matters a bit less for others).

theunrepentantgeek: Clarified doc comment.

matthchr (approved):

PR #4819, opened 2025-07-15
Add support for network firewall resources

### LINKED ISSUE #4729: Feature: Support azureFirewalls & firewallPolicy
_Describe the feature request._

Support azureFirewalls & firewallPolicy
https://learn.microsoft.com/en-us/rest/api/firewall/azure-firewalls/create-or-update?view=rest-firewall-2024-05-01&tabs=HTTP

matthchr:
* Added support for AzureFirewall, FirewallPolicy, and FirewallPoliciesRuleCollectionGroup resources.
 * Added samples + tests.

Fixes #4729.

- [x] this PR contains tests


v2/azure-arm.yaml
```diff
@@ -3683,6 +3683,30 @@ objectModelConfiguration:
...
+      AzureFirewallIPConfiguration:
+        $resourceEmbeddedInParent: AzureFirewall
+      AzureFirewallApplicationRuleCollection:
+        $resourceEmbeddedInParent: AzureFirewall
+      AzureFirewallNatRuleCollection:
+        $resourceEmbeddedInParent: AzureFirewall
+      AzureFirewallNetworkRuleCollection:
+        $resourceEmbeddedInParent: AzureFirewall
+      FirewallPolicies_RuleCollectionGroup:
+        $exportAs: FirewallPoliciesRuleCollectionGroup
```
matthchr, line 3701: Confirm this is the naming pattern we want to follow.

I think that it is, because calling this `RuleCollectionGroup` (not disambiguated by parent) here risks collision with other resources with different parents but same child name.

theunrepentantgeek: Generally speaking, we should be avoiding any common prefixes when naming resources as the naming adds no value. Many of our older resource names are poor because of this.

In this specific case, I think the name you've selected is a good one, precisely because there's a risk of a name clash.

v2/azure-arm.yaml
```diff
@@ -3683,6 +3683,30 @@ objectModelConfiguration:
         $export: true
         $supportedFrom: v2.10.0
     2024-03-01:
+      AzureFirewall:
+        $export: true
```
theunrepentantgeek: Prefer avoiding `$export`, I want to remove it.

```suggestion
        $exportAs: AzureFirewall
```

v2/azure-arm.yaml
```diff
@@ -3683,6 +3683,30 @@ objectModelConfiguration:
...
+        $supportedFrom: v2.15.0
+      FirewallPolicy:
+        $export: true
```
theunrepentantgeek:
```suggestion
        $exportAs: FirewallPolicy
```

v2/internal/controllers/networking_firewall_20240301_test.go
```diff
@@ -0,0 +1,124 @@
...
+	t.Parallel()
+
+	tc := globalTestContext.ForTest(t)
+
+	rg := tc.CreateTestResourceGroupAndWait()
+
+	vnet := newVNet20240301(tc, testcommon.AsOwner(rg), []string{"10.0.0.0/16"})
+	publicIP := newPublicIP20240301(tc, testcommon.AsOwner(rg))
+	subnet := newSubnet20240301(tc, vnet, "10.0.0.0/24")
+	subnet.Name = "azurefirewallsubnet" // Name must be AzureFirewallSubnet for Azure Firewall
```
theunrepentantgeek: Slight (casing) inconsistency between the name and the comment - is this maybe case sensitive?

theunrepentantgeek (approved):

v2/internal/controllers/networking_firewall_20240301_test.go
matthchr: It's not case sensitive (because it's an Azure/ARM name and they're all case-insensitive + case-preserving), but I'll fix the comment.

PR #4822, opened 2025-07-17
Update default max parallelism to 4 and add perf tests

matthchr:
- [ ] this PR contains tests


matthchr:
Opening this but want to do a bit more profiling before merging

v2/charts/azure-service-operator/values.yaml
```diff
@@ -241,16 +241,16 @@ tolerations: []
...
 
 # MaxConcurrentReconciles is the number of threads/goroutines dedicated to reconciling each resource type.
-# If not specified, the default is 1.
+# If not specified, the default is 4.
 # IMPORTANT: Having MaxConcurrentReconciles set to N does not mean that ASO is limited to N interactions with
 # Azure at any given time, because the control loop yields to another resource while it is not actively issuing HTTP
 # calls to Azure. Any single resource only blocks the control-loop for its resource-type for as long as it takes to issue
 # an HTTP call to Azure, view the result, and make a decision. In most cases the time taken to perform these actions
 # (and thus how long the loop is blocked and preventing other resources from being acted upon) is a few hundred
 # milliseconds to at most a second or two. In a typical 60s period, many hundreds or even thousands of resources
```
theunrepentantgeek: Do we need to scale back this comment? For some resources it's certainly true, but (as we've recently learned), not for all. 

```suggestion
# milliseconds to at most a second or two. In a typical 60s period, many hundreds of resources
```

matthchr: Fixed (and elsewhere)

matthchr:
Updated with perf tests (yay). These graphs are from 1RG 1VNET created every 1s for 5m w/ a cap at 200 after which we start deleting. The spike in CPU you see in both at the end is all the resources being deleted.

Here's w/ it set to 4:
image
image

And here's with it set to 1:
image
image

Interestingly mem usage is a bit lower w/ 4 - not sure if that's just a fluke of small number of runs (N=1) or if it's a real finding. CPU is a bit higher on average but not by much and the CPU numbers overall are pretty low. It seems safe to me to merge this.

v2/internal/testcommon/podmetrics/collector.go
```diff
@@ -0,0 +1,194 @@
...
+		<-mc.done
+	}
+}
+
+// Samples returns a copy of all collected samples.
+func (mc *MetricsCollector) Samples() []Sample {
+	mc.mu.Lock()
+	defer mc.mu.Unlock()
+	result := make([]Sample, len(mc.samples))
+	copy(result, mc.samples)
```
theunrepentantgeek: Use `slices.Clone()` #nit

v2/tools/collect-metrics/main.go
theunrepentantgeek, line 1: Why is this not under `v2/tools`?

v2/test/perf/cmd/collect-metrics/main.go
```diff
@@ -0,0 +1,134 @@
...
+//	# Collect for 10 minutes with defaults, writing to metrics.csv
+//	go run ./v2/test/perf/cmd/collect-metrics/ --duration 10m --output metrics.csv
+//
+//	# Custom namespace, prefix, and interval
+//	go run ./v2/test/perf/cmd/collect-metrics/ --namespace my-ns --pod-prefix my-pod --interval 2s --duration 5m
+//
+//	# Run in background alongside tests
+//	go run ./v2/test/perf/cmd/collect-metrics/ --duration 30m --output metrics.csv &
+//	go test -v -run Test_Perf_Dynamic_ResourceGroups ./v2/test/perf/...
+//	kill %1
```
theunrepentantgeek: Will it still be `%1` if there's another background app already running?

v2/test/perf/cmd/collect-metrics/main.go
```diff
@@ -0,0 +1,134 @@
...
+
+	"github.com/Azure/azure-service-operator/v2/internal/testcommon/podmetrics"
+)
+
+func main() {
+	namespace := flag.String("namespace", podmetrics.DefaultNamespace, "Kubernetes namespace to monitor")
+	podPrefix := flag.String("pod-prefix", podmetrics.DefaultPodPrefix, "Pod name prefix to filter on")
+	interval := flag.Duration("interval", podmetrics.DefaultInterval, "Polling interval (e.g. 5s, 10s) - default is 5s")
+	duration := flag.Duration("duration", 0, "Collection duration (e.g. 5m, 1h). 0 means run until interrupted.")
+	output := flag.String("output", "", "Output CSV file path (default: metrics_<timestamp>.csv)")
```
theunrepentantgeek: Aside: I've been trying out [kong](https://github.com/alecthomas/kong) for command line parsing and it's pretty nice - lightweight enough to little tools and powerful enough for complex ones.

Not saying you should change this - but you should check it out.

v2/test/perf/cmd/collect-metrics/main.go
```diff
@@ -0,0 +1,134 @@
...
+	flag.Parse()
+
+	log := textlogger.NewLogger(textlogger.NewConfig())
+	ctrl.SetLogger(log)
+
+	if *output == "" {
+		outDir := filepath.Join("..", "..", "reports")
```
theunrepentantgeek: This is very dependent on the current directory when starting - I'd suggest one of the following:
* make `--output` mandatory, _or_
* write into the current folder whatever it is, _or_ 
* walk up from the current directory until you find a `.git` folder, and use that destination

v2/test/perf/reports/metrics_StartUp_20260305_1821.csv
theunrepentantgeek, line 1: Do we need these metrics files?

v2/tools/collect-metrics/main.go
matthchr: Ahh good idea. I'll move it.

v2/test/perf/reports/metrics_StartUp_20260305_1821.csv
matthchr: I wanted to commit them so we had a record of perf from previous runs. Maybe I need to add a README.md here to explain the cluster topology I used as well.

Basically that way we can always see if we're improving/regressing.

v2/test/perf/cmd/collect-metrics/main.go
matthchr: Fixed

v2/test/perf/cmd/collect-metrics/main.go
matthchr: Yeah I'll make it the current directory and then give a cmdline example of using it to output to the perf folder, that seems better.

v2/test/perf/reports/metrics_StartUp_20260305_1821.csv
matthchr: Added a readme.

v2/test/perf/cmd/collect-metrics/main.go
matthchr: I think for this relatively simple tool it's OK to just use the standard `flag` -- but yeah something I'll keep in mind to try. (I think I have used it for a few other tools in AKS RP but it's been a bit)

PR #4846, opened 2025-08-01
Add azure and {group} categories to CRDs

matthchr:
## Checklist

- [ ] this PR contains tests

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -267,7 +267,9 @@ tasks:
     desc: Update {{.GENERATOR_APP}} golden test outputs.
     dir: '{{.GENERATOR_ROOT}}'
     cmds:
-      - go test ./... -run ^TestGolden -update
+      - go test ./...
+    env:
+      GOLDIE_UPDATE: 1
 
   generator:format-code:
     desc: Ensure all code for the code generator is formatted
--- a/v2/internal/crdmanagement/manager.go
+++ b/v2/internal/crdmanagement/manager.go
@@ -378,8 +378,6 @@ func (m *Manager) applyCRDs(
 
 	i := 0
 	for _, instruction := range instructionsToApply {
-		instruction := instruction
-
 		i += 1
 		toApply := &apiextensions.CustomResourceDefinition{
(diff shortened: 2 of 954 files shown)


Taskfile.yml
```diff
@@ -267,7 +267,9 @@ tasks:
     desc: Update {{.GENERATOR_APP}} golden test outputs.
     dir: '{{.GENERATOR_ROOT}}'
     cmds:
-      - go test ./... -run ^TestGolden -update
+      - go test ./...
+    env:
```
matthchr, line 271: This fixes the error from some tests that don't support `-update`.

v2/internal/crdmanagement/manager.go
```diff
@@ -399,6 +397,9 @@ func (m *Manager) applyCRDs(
 			return nil
 		})
 		if err != nil {
+			// Logging here before returning because we may cancel the lease (due to defer cancel()) above, and we want to ensure
```
matthchr, line 400: This change would be easy to miss. It's a simple logging fix because I had the operator crashing at this point and I had a hard time debugging it. This makes it better.

v2/tools/generator/internal/astmodel/resource_type.go
```diff
@@ -654,6 +654,8 @@ func (resource *ResourceType) AsDeclarations(
 	}
 
 	astbuilder.AddComment(&comments, "// +kubebuilder:object:root=true")
+	categoryName := strings.Replace(strings.ToLower(declContext.Name.InternalPackageReference().Group()), ".", "", -1)
+	astbuilder.AddComment(&comments, fmt.Sprintf("// +kubebuilder:resource:categories={azure,%s}", categoryName))
```
matthchr, line 658: My main question/struggle here is if we want to include the group or not. The `azure` category I think makes a ton of sense and we should have it. I _think_ I like the `group` category but open to other opinions.

theunrepentantgeek: Makes sense to me. :+1:

theunrepentantgeek (approved):

PR #4869, opened 2025-08-11
Add some more tests for ReconcilePolicy at namespace scope

matthchr:
- [x] this PR contains tests


v2/internal/controllers/reconcile_policy_test.go
```diff
@@ -198,7 +250,7 @@ func Test_ReconcilePolicySkip_CreatesSecretAndConfigMap(t *testing.T) {
 	tc.Patch(old, acct)
 }
 
-func testDeleteSkipped(t *testing.T, policy string) {
+func testDeleteSkipped(t *testing.T, policy string, onNamespace bool) {
```
theunrepentantgeek: Should this actually two methods that possibly share some helper methods?

If not, consider using a constant - `testDeleteSkipped(t, "skip, false)` is pretty opaque as to what the **false** (or **true**) means.

v2/internal/controllers/reconcile_policy_test.go
```diff
@@ -77,12 +78,63 @@ func Test_ReconcilePolicy_UnknownPolicyIsIgnored(t *testing.T) {
...
+
+func Test_ReconcilePolicyOnNamespace_DetachOnDelete_SkipsDelete(t *testing.T) {
+	t.Parallel()
+	testDeleteSkipped(t, "detach-on-delete", true)
+}
+
+func Test_ReconcilePolicyOnNamespace_Skip_SkipsDelete(t *testing.T) {
+	t.Parallel()
+	testDeleteSkipped(t, "skip", true)
+}
```
theunrepentantgeek: Should these three (four?) be a table test?

matthchr: Changed to a table test.

v2/internal/controllers/reconcile_policy_test.go
matthchr: I think the same method is OK here. There's so much overlap and factoring the common bits out into helpers would require at least a couple helpers (because the common bits are split between some uncommon bits - see the if statements).

I refactored to a table-style test and also swapped the arguments to be a struct to make the call pattern clearer. I think this + the table style test is enough for clarity here.

theunrepentantgeek (approved):

PR #4874, opened 2025-08-12
Add crd storedVersions deprecation controller

### LINKED ISSUE #4376: ASO should help the user manage CRD storedVersions
If we deprecate a CRD version that was at one point the storage version (in `storedVersions` field of CRD status), the user/ASO itself will be unable to update their CRDs. An example of this happened in 2.9.0 with `v1api20230202preview` of `containerservice`.

**Describe the improvement**
We should've supported cleanup up `storedVersions` via `asoctl` in the above case, but in the mode where ASO installs its own CRDs we also should just manage this oureslves so that users don't have to fiddle with `storedVersions`. 

We could take a page from Karpenter and do something like this: https://github.com/kubernetes-sigs/karpenter/pull/1689 - or if not a controller (which may not be fully necessary for our case) we could do this checking during or before CRD installation.

An implementation note is that it should probably share logic with asoctl.

**Additional context**

matthchr:
## What this PR does

* Add crd storedVersions deprecation controller
* Add controllutils for singleton controller

Fixes #4376.

## Checklist

- [x] this PR contains tests

### DIFF
--- a/v2/cmd/controller/app/setup.go
+++ b/v2/cmd/controller/app/setup.go
@@ -47,6 +47,7 @@ import (
 	armreconciler "github.com/Azure/azure-service-operator/v2/internal/reconcilers/arm"
 	entrareconciler "github.com/Azure/azure-service-operator/v2/internal/reconcilers/entra"
 	"github.com/Azure/azure-service-operator/v2/internal/reconcilers/generic"
+	"github.com/Azure/azure-service-operator/v2/internal/reconcilers/migration/crd"
 	asocel "github.com/Azure/azure-service-operator/v2/internal/util/cel"
 	"github.com/Azure/azure-service-operator/v2/internal/util/interval"
 	"github.com/Azure/azure-service-operator/v2/internal/util/kubeclient"
@@ -94,8 +95,15 @@ func SetupControllerManager(ctx context.Context, setupLog logr.Logger, flgs *Fla
 
 	k8sConfig := ctrl.GetConfigOrDie()
 	ctrlOptions := ctrl.Options{
-		Scheme:           scheme,
-		Cache:            cacheOpts,
+		Scheme: scheme,
+		Cache:  cacheOpts,
+		Client: client.Options{
+			Cache: &client.CacheOptions{
+				DisableFor: []client.Object{
+					&apiextensions.CustomResourceDefinition{},
+				},
+			},
+		},
 		LeaderElection:   flgs.EnableLeaderElection,
 		LeaderElectionID: "controllers-leader-election-azinfra-generated",
 		// Manually set lease duration (to default) so that we can use it for our leader elector too.
@@ -190,7 +198,7 @@ func SetupControllerManager(ctx context.Context, setupLog logr.Logger, flgs *Fla
 
 	if cfg.OperatorMode.IncludesWatchers() {
 		//nolint:contextcheck
-		err = initializeWatchers(readyResources, cfg, mgr, clients)
+		err = initializeWatchers(readyResources, cfg, mgr, clients, flgs)
 		if err != nil {
 			setupLog.Error(err, "failed to initialize watchers")
 			os.Exit(1)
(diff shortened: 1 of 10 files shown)


matthchr:
I should add documentation as well for the new label.

edit: did this

v2/cmd/controller/app/setup.go
```diff
@@ -482,6 +496,19 @@ func initializeWatchers(readyResources map[string]apiextensions.CustomResourceDe
 		return eris.Wrap(err, "failed to register gvks")
 	}
 
+	// Register other controllers
+	if flgs.CRDManagementMode == "auto" {
+		deprecatedVersions := crdmanagement.GetAllDeprecatedStorageVersions(readyResources)
```
theunrepentantgeek, line 501: How long does this take to run, and is that a suitable timeframe in the midst of initialization?

v2/cmd/controller/app/setup.go
```diff
@@ -482,6 +496,19 @@ func initializeWatchers(readyResources map[string]apiextensions.CustomResourceDe
...
+		for crdName, versions := range deprecatedVersions {
+			clients.log.V(Status).Info("Will migrate storedVersions", "crdName", crdName, "versions", versions)
+		}
+		// Register the CRD migration reconciler
+		err = crd.NewReconciler(clients.kubeClient, mgr.GetCache(), deprecatedVersions, crd.Options{}).SetupWithManager(mgr)
```
theunrepentantgeek, line 506: Should you skip starting the reconciler if there's nothing to upgrade?

v2/internal/reconcilers/migration/crd/reconciler_test.go
```diff
@@ -0,0 +1,193 @@
...
+	"github.com/Azure/azure-service-operator/v2/internal/config"
+	"github.com/Azure/azure-service-operator/v2/internal/crdmanagement"
+	"github.com/Azure/azure-service-operator/v2/internal/reconcilers/migration/crd"
+	"github.com/Azure/azure-service-operator/v2/internal/testcommon"
+	"github.com/Azure/azure-service-operator/v2/internal/util/kubeclient"
+)
+
+// This test requires that the task target `bundle-crds` has been run
+func TestReconcileCRDs(t *testing.T) {
+	tests := []struct {
```
theunrepentantgeek: Consider using a `map[string]struct{}` with the test name as the key, I find this makes it easier to scan the test names when reading the code.

theunrepentantgeek (approved):

v2/cmd/controller/app/setup.go
matthchr: The expensive part of this is listing all of the CRDs from the server -- this is what `readyResources` is derived from (note the type there is `map[string]apiextensions.CustomResourceDefinition`). We already get this list today to determine what is installed.

This then just goes through them in-memory and filters for the ones that have deprecated versions as defined in the static (hardcoded in code in `crdmanagement` package) list of deprecated versions.

tl;dr: It's an in-memory filter over at most N items where N == number of CRDs from crdPattern. It should be very fast.

v2/cmd/controller/app/setup.go
matthchr: We could, yea.

The reason I didn't is because this way the "we've migrated all the CRDs" and the "we never had any CRDs to migrate" paths end up the same - both will be running this reconciler, but it won't do anything except wake up once an hour and try to clean (nothing) up, and then sleep for an hour again.

Open to changing that behavior though, wydt?

theunrepentantgeek: I'm comfortable either way. Was mostly thinking about minimizing resource use, but the cost here is very low.

v2/internal/reconcilers/migration/crd/reconciler_test.go
matthchr: Fixed.

theunrepentantgeek (approved):

PR #4883, opened 2025-08-16
Add 2024-04-13 support for current Kusto resources, plus principal assignments

JamesJuddAVEVA:
Add 2024-04-13 support for current Kusto resources, plus principal assignments

gif

## Checklist

- [x] this PR contains tests


v2/internal/controllers/kusto_cluster_20240413_test.go
```diff
@@ -0,0 +1,134 @@
...
+				SoftDeletePeriod: to.Ptr("P1D"),
+			},
+		},
+	}
+
+	tc.CreateResourcesAndWait(cluster, db)
+	tc.Expect(cluster.Status.Identity).ToNot(BeNil())
+	tc.Expect(cluster.Status.Identity.PrincipalId).ToNot(BeNil())
+
+	principalId := *cluster.Status.Identity.PrincipalId
```
theunrepentantgeek, line 57: Here in the coded test, you have the option to copy out the `PrincipalId` to reuse for the assignment in a sub-test, but a user of the resource can't do that via YAML unless we generate the required support.

Doing this requires two minor change to `azure_arm.yaml`, followed by regenerating the code.

On `Cluster`, we want to generate support for exporting the clusters `PrincipalId` and `TenantId` to config maps. This will look something like this:

```yaml
$generatedConfigs:
	clusterPrincipalId: $.Status.Identity.PrincipalId
	clusterTenantId: $.Status.Identity.TenantId
```

On `PrincipleAssignment` we want these two properities to be configurable:

```yaml
      Clusters_Databases_PrincipalAssignment:
        $exportAs: PrincipalAssignment
        $supportedFrom: v2.15.0
        PrincipalId:
        TenantId:
```

All of the code required to handle these is then handle by the code generator.

v2/samples/kusto/v1api20240413/v1api20240413_principalassignment.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: aso-sample-principalassignment
+  namespace: default
+spec:
+  owner:
+    name: aso-sample-database
+  principalId: "james.judd@aveva.com"
+  principalType: "User"
+  role: "Viewer"
+  tenantId: "d1fe8e44-4d35-4dad-8ea9-d9d9cb9a988c"
```
theunrepentantgeek: With the use of `$importedConfigMapMode` as detailed in a separate comment, you can avoid putting _your_ `tenantId` in this sample.

v2/samples/kusto/v1api20240413/v1api20240413_principalassignment.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: kusto.azure.com/v1api20240413
+kind: PrincipalAssignment
+metadata:
+  name: aso-sample-principalassignment
+  namespace: default
+spec:
+  owner:
+    name: aso-sample-database
+  principalId: "james.judd@aveva.com"
```
theunrepentantgeek: I had to go look up the docs for this.

Usually, a `principalId` is strictly a GUID, but in this case it's much more flexible:

>The principal ID assigned to the cluster principal. It can be a user email, application ID, or security group name.

Nice.

theunrepentantgeek (commented):
This is really really good, thank you for the PR.

I have one minor comment about making the `PrincipalId` and `TenantId` fields exportable from `Cluster` and importable by `PrincipalAssignment`, but that's more of a usability improvement than anything else.

theunrepentantgeek:
Minor failure running the continuous integration tests. After manually undoing all the `\\\"\\n\\t` encoding, we end up with this:

```
Shortest body diff:   strings.Join({
  	`{"name":"asotestwieyui","properties":{"principalId":"james.judd@`,
  	`aveva.com","principalType":"User","role":"Viewer","tenantId":"`,
- 	"00000000-0000-0000-0000-000000000000",
+ 	"d1fe8e44-4d35-4dad-8ea9-d9d9cb9a988c",
  	`"}}`,
  }, "")
```

The recording for your sample was automatically redacted, removing the tenant ID `d1fe8e44-4d35-4dad-8ea9-d9d9cb9a988c` and replacing it with the nil GUID `00000000-0000-0000-0000-000000000000`.  (This automatic redaction keeps sensitive things like authentication tokens, API keys, and subscription Ids out of the repo.)

When the test replayed, the discrepancy caused a test failure - because it's supposed to be the same every time.

Easy fix: Once you've made the changes I suggested above to make `principalId` and `tenantId` exportable to configmaps, update your sample to use the configmap and re-record.

matthchr (approved):

PR #4904, opened 2025-08-29
Avoid redundant manual dereferences 

theunrepentantgeek:
## What this PR does

We've been generating a pointer dereference that's not strictly required.

Instead of this:
``` go
	if rule.Detector != nil {
		detector_ARM, err := (*rule.Detector).ConvertToARM(resolved)
		if err != nil {
			return nil, err
		}
```

we will now generate
``` go
	if rule.Detector != nil {
		detector_ARM, err := rule.Detector.ConvertToARM(resolved)
		if err != nil {
			return nil, err
		}
```

The only manual change is to `builder.go` - everything else follows from that.


theunrepentantgeek:
Bypassed failed CLA check

image

PR #4921, opened 2025-09-04
Upgrade devcontainer support for docker-in-docker 

theunrepentantgeek:
## What this PR does

Upgrades our support for running containers inside our devcontainer by upgrading to the latest version of docker-in-docker.

This was triggered by the package `software-properties-common` being removed from the upstream apt repository, apparently because the maintainer has been slow at updating the package recently. 

The `software-properties-common` package was a prerequisite for the `docker-debian.sh` script, which comes from a now-archived repository.

I've pulled in the latest version of the feature, and tested by running `task controller:test-integration-envtest` locally (as that requires container support to pass).

gif


matthchr (approved):

v2/internal/testcommon/kube_test_context_envtest.go
```diff
@@ -54,7 +54,7 @@ import (
 )
 
 func getRoot() (string, error) {
-	cmd := exec.Command("git", "rev-parse", "--show-toplevel")
+	cmd := exec.CommandContext(context.Background(), "git", "rev-parse", "--show-toplevel")
```
matthchr, line 57: Why make this change? I assume `Command` is calling `CommandContext` w/ background context already, so this is no different than we had before?

theunrepentantgeek: Required to fix a linting error - I think it was `contextcheck`, flagging that you should always call the overload with a context parameter if one is available.

PR #4922, opened 2025-09-04
Extend ResourceReferences to support WellKnown names and backward compatibility

theunrepentantgeek:
## What this PR does

Implements our [design](https://azure.github.io/azure-service-operator/design/adr-2025-01-arm-references/) for improvements to ResourceReference

* `WellknownResourceReference` is enabled by specifying `$resourceType: arm+wellknown` and enables situations where a resource may be identified by a well known name instead of a full ARM ID.

* Compatibility mode is enabled by specifying `$resourceType: arm+compat` and allows us to introduce `ResourceReference` properties alongside existing `string` references, allowing us to correct oversights not spotted prior to release of certain resources.

## How does this PR make you feel?

gif

PR #4923, opened 2025-09-04
Allow specifying Role Assignment by using the well-known name of a RoleDefinition

### LINKED ISSUE #3642: Improvement: support for the friendly names of the builtin roles
Currently, RoleAssignment resource requires full `armId` value for the role which is big and unfriendly and requires `subscriptionId` to be present.

**Describe the improvement**
Please add in addition support for a friendly name of the builtin role so that instead of this
```yaml
apiVersion: authorization.azure.com/v1api20200801preview
kind: RoleAssignment
metadata:
  name: aso-sample-contributor
  namespace: default
spec:
  ...
  roleDefinitionReference:
    # This ARM ID represents "Contributor" - you can read about other built in roles here: https://docs.microsoft.com/en-us/azure/role-based-access-control/built-in-roles
    armId: /subscriptions/00000000-0000-0000-0000-000000000000/providers/Microsoft.Authorization/roleDefinitions/b24988ac-6180-42a0-ab88-20f7382dd24c
```
user could use name of the role like this
```yaml
apiVersion: authorization.azure.com/v1api20200801preview
kind: RoleAssignment
metadata:
  name: aso-sample-contributor
  namespace: default
spec:
  ...
  roleDefinitionReference:
    armName: Contributor
```

theunrepentantgeek:
## What this PR does

Uses the `WellknownResourceReference` support from #4922 to enable specification of RoleAssignments using the well known name of a built-in RoleDefinition instead of its full ARM ID.

Closes #3642 

### Prerequisites

- [x] PR #4922 

When the prerequisites merge, GitHub will automatically change the target branch to `main` and run CI.

gif


v2/api/authorization/customizations/role_assignment_extensions.go
```diff
@@ -43,3 +55,112 @@ func (extension *RoleAssignmentExtension) Import(
...
+			}
+
+			return armObj, nil
+		}
+	}
+
+	return armObj, nil
+}
+
+var builtInRoleDefinitions map[string]string = make(map[string]string, 800)
```
matthchr: minor: Missing comment about why ~800?

matthchr: Parallelism risk here. We can have many callers of this. Map is constructed once and written to many times. Go maps are not concurrent-safe.

Additionally, _technically speaking_ it's possible for new role definitions to be added. Depending on how long the ASO pod runs for, that could happen during its lifetime, but if we only ever write it once they won't be able to use the new ones without a pod restart.

We could have a threadsafe cache (there are some existing OSS implementations, Karpenter uses patrickmn/go-cache, though it doesn't have generics support and I think there are others that do) with a really long expiry like 1d?

matthchr (commented):
Main issue is thread-safety of map-cache. Rest LGTM.

v2/api/authorization/customizations/role_assignment_extensions.go
matthchr: Cache is maybe not required (I'm open to either opinion). Thread-safety is required I think.

theunrepentantgeek: Updated to protect map access with locks.

I don't think we need worry about cache expiry - restarting the pod isn't difficult and we're talking about an extremely rare event.

theunrepentantgeek: Added comment. There are currently ~700 such roles; initializing the map to 800 gives room for growth.

matthchr: very minor: Possibly worth a note in a comment here just acknowledging this case?

matthchr:
Looks like an error?

```
[controller:test-integration-envtest] fatal error: sync: Unlock of unlocked RWMutex
[controller:test-integration-envtest] 
[controller:test-integration-envtest] goroutine 14128 [running]:
[controller:test-integration-envtest] sync.fatal({0xbd7a15b?, 0x8a?})
[controller:test-integration-envtest] 	/usr/local/go/src/runtime/panic.go:1038 +0x18
[controller:test-integration-envtest] sync.(*RWMutex).Unlock(0x12a5e980)
[controller:test-integration-envtest] 	/usr/local/go/src/sync/rwmutex.go:212 +0x45
[controller:test-integration-envtest] github.com/Azure/azure-service-operator/v2/api/authorization/customizations.resolveBuiltInRoleDefinition({0xc03b94d0f0, 0xb}, {0xc969f70, 0xc010ba46f0})
[controller:test-integration-envtest] 	/workspace/v2/api/authorization/customizations/role_assignment_extensions.go:180 +0x278
[controller:test-integration-envtest] github.com/Azure/azure-service-operator/v2/api/authorization/customizations.(*RoleAssignmentExtension).ModifyARMResource(0x0?, {0xc97dfd0?, 0xc022c5f110?}, 0xa1fb000?, {0xc969f70, 0xc010ba46f0}, {0xca124d0?, 0xc0747b1208?}, {0xc9b0f00, 0xc001836080}, ...)
[controller:test-integration-envtest] 	/workspace/v2/api/authorization/customizations/role_assignment_extensions.go:90 +0xe6
[controller:test-integration-envtest] github.com/Azure/azure-service-operator/v2/internal/reconcilers/arm.(*azureDeploymentReconcilerInstance).ConvertResourceToARMResource.CreateARMResourceModifier.func2({0xc97dfd0, 0xc022c5f110}, {0xca124d0, 0xc0747b1208}, {0xc969f70, 0xc010ba46f0})
[controller:test-integration-envtest] 	/workspace/v2/pkg/genruntime/extensions/arm_resource_modifier.go:58 +0x164
[controller:test-integration-envtest] github.com/Azure/azure-service-operator/v2/internal/reconcilers/arm.(*azureDeployment
```

v2/api/authorization/customizations/role_assignment_extensions.go
```diff
@@ -43,3 +56,126 @@ func (extension *RoleAssignmentExtension) Import(
...
+
+	return nil
+}
+
+func resolveBuiltInRoleDefinition(
+	roleDefinitionName string,
+	armObj genruntime.ARMResource,
+) (string, error) {
+	builtInRoleDefinitionsLock.RLock()
+	defer builtInRoleDefinitionsLock.Unlock()
```
matthchr: This needs to be `RUnlock` - it's what is causing the test failure I think.

v2/api/authorization/customizations/role_assignment_extensions.go
```diff
@@ -43,3 +56,126 @@ func (extension *RoleAssignmentExtension) Import(
...
+)
+
+func ensureBuiltInRoleDefinitionsLoaded(
+	ctx context.Context,
+	armClient *genericarmclient.GenericClient,
+) error {
+	builtInRoleDefinitionsLock.Lock()
+	defer builtInRoleDefinitionsLock.Unlock()
+
+	// Short circuit if already loaded
```
matthchr, line 131: minor: you could double-check this (before/after lock). Advantage is that it avoids taking the lock at all after the first time which is nice.

It's a small optimization though so feel free to ignore if you don't think it's worth it. If you do do this, I'd leave a comment explaining the 2x check.

theunrepentantgeek: I don't think it's worth the additional complexity in this case.

PR #4924, opened 2025-09-04
Allow specifying `system` for Identity references in `app` resources

theunrepentantgeek:
## What this PR does

A number of the Identity references used in the `app` group permit specifying `system` to use a system assigned identity instead of an explicit one. 

Previously ASO had no support for these, but with the completion of #4922 we'll have full support via the `wellKnownName` property.

### Prerequisites

- [ ] #4922 

gif

### DIFF
--- a/v2/azure-arm.yaml
+++ b/v2/azure-arm.yaml
@@ -1241,13 +1241,13 @@ objectModelConfiguration:
         passwordSecretRef:
           $isSecret: false
         Identity:
-          $referenceType: arm
+          $referenceType: arm+wellknown
       ServiceBind:
         ServiceId:
           $referenceType: arm
       Secret:
         Identity:
-          $referenceType: arm
+          $referenceType: arm+wellknown
       ContainerApp_Spec:
         ManagedBy:
           $referenceType: arm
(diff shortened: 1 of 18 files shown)


jakjang (dismissed):

v2/pkg/genruntime/well_known_resource_reference.go
```diff
@@ -10,7 +10,7 @@ package genruntime
 type WellKnownResourceReference struct {
```
matthchr, line 10: This file is not correctly named still (has both underscores and dashes).

v2/azure-arm.yaml
```diff
@@ -1247,13 +1247,13 @@ objectModelConfiguration:
         passwordSecretRef:
           $isSecret: false
         Identity:
-          $referenceType: arm
+          $referenceType: arm+wellknown
```
matthchr: Maybe worth a comment on these with a link to either the ASO issue and/or the documentation from the service that explains what the well known reference actually is?

matthchr: Don't necessarily need to repeat it on every one either, could just put it once and then say it applies to all of the identity references in `app`.

theunrepentantgeek: Added a comment on each one for clarity.

v2/pkg/genruntime/well_known_resource_reference.go
theunrepentantgeek: Renamed.

v2/azure-arm.yaml
```diff
@@ -1247,13 +1247,13 @@ objectModelConfiguration:
         passwordSecretRef:
           $isSecret: false
         Identity:
-          $referenceType: arm
+          $referenceType: arm+wellknown # permits `system` to use a system-assigned identity
```
theunrepentantgeek, line 1250: Annotated every case individually to be explicit.

v2/pkg/genruntime/well_known_resource_reference.go
matthchr: You fixed this the wrong way I think. It should be all underscores not all dashes =)

PR #4925, opened 2025-09-04
Fix up broken ResourceReferenceProperties

theunrepentantgeek:
## What this PR does

Fixes up the 4 known cases we have where ARM ID properties were released without being properly tagged-as/converted-to KnownResourceReference structs.

Thanks to #4922 we can now add the correct `ResourceReference` property alongside the legacy one, avoiding any breakage for existing users.

I had to tweak our test generation to exclude the legacy properties from our round-trip tests, because these properties break the previous rule that there's a 1-1 correspondence between properties in different versions of a resource. Now we have two properties both updating one in the later version:

```mermaid
flowchart LR
v1r[SourceServerResourceReference ] --> v2r[SourceServerResourceReference]
v1i[SourceServerResourceId] --> v2r
```

Forward conversions are fine, but for backward conversions we don't know whether the value came from `SourceServerResourceReference` or `SourceServerResourceId`.

### Prerequisites

- [ ] #4922 

gif


matthchr:
> Forward conversions are fine, but for backward conversions we don't know whether the value came from SourceServerResourceReference or SourceServerResourceId.

What do we do in that case then? and is it going to cause issues with behavior since we convert a lot? (since everything goes to/from the `xyzstorage` types?

v2/azure-arm.yaml
```diff
@@ -2379,7 +2379,7 @@ objectModelConfiguration:
         $nameInNextVersion: MySQLServerSku_STATUS
       ServerProperties:
         SourceServerResourceId:
-          $referenceType: simple # Actually, this *IS* a resource id, but we want to avoid breaking changes so we're fibbing here
+          $referenceType: arm+compat
```
matthchr: Do we actually want to change this?

This is an old version (2021-05-01). It was already supported in the newer versions. I guess this is safe to do, but I thought that the objective for arm+compat was especially targeted at the versions we couldn't fix (because there was no new version to fix it in), which is not the case here.

v2/api/dbformysql/v1api20210501/storage/flexible_server_types_overrides.go
```diff
@@ -32,7 +33,11 @@ func (server *FlexibleServer_Spec) AssignPropertiesTo(dst *v20230630s.FlexibleSe
 	// Get the current property bag
 	pb := genruntime.NewPropertyBag(dst.PropertyBag)
 
-	if server.SourceServerResourceId != nil {
+	// If we only have the legacy property, populate the new one from it
```
matthchr: Is this right? Does this RoundTrip correctly?

If old has `ResourceId` set, so it flows to new `ServerResourceReference` and back to `ResourceId` (based on the `AssignPropertiesFrom` above), which si good.
but if old gets `ResourceReference` set, it will flow to `ResourceReference` as well, and then back to `ResourceId` which isn't correct, isn't that right?

matthchr: See my comment on azure-arm.yaml about if we even want to change this old API version at all.

v2/pkg/genruntime/wellknown-resource-reference.go
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+package genruntime
+
+// WellknownResourceReference is a variation of ResourceReference that also permits the referenced resource to be
+// identified by a well known name.
```
matthchr: I don't think this file should be here -- I thought we already had it?

Also, file name is wrong again =)

theunrepentantgeek: Problems due to rebasing - now fixed.

v2/api/dbformysql/v1api20210501/storage/flexible_server_types_overrides.go
theunrepentantgeek: Agreed. Since that resource has a newer version, we don't need the special handling; removed.

v2/azure-arm.yaml
theunrepentantgeek: Agreed. Removed.

matthchr (approved):
Nice!

PR #3967, opened 2024-04-23
Update available version of DocumentDB

theunrepentantgeek:
**What this PR does / why we need it**:

Adds the `2023-11-15` version of resources from the `documentdb` group.

Closes #3768 
**How does this PR make you feel**:
gif

**If applicable**:
- [x] this PR contains documentation
- [x] this PR contains tests

PR #4937, opened 2025-09-14
Improve diagnostics from TypeTransformers

### LINKED ISSUE #4867: Modify TypeTransformers to give better diagnostics

I've observed CoPilot getting stuck in a cycle trying to work out which type name to specify so that a TypeTransformer does the expected thing. 

The root of the problem is that we rename types after the TypeTransformers run, so the final runtime type name isn't the right one. 

## Describe the improvement

We should detect when a TypeTransformer isn't used, and generate a useful error that can be used to correct the configuration. Our existing TypoAdvisor will be helpful here.

theunrepentantgeek:
## What this PR does

Tightens our validation of TypeTransformer configuration to ensure that all parts of the TypeMatcher are used, and to leverage our `TypoAdvisor` to generate good error messages.

Also fixes ineffective removal of the property `CustomerManagedKeyStatus`, dating back to #3967

Closes #4867 

Sample error (line-breaks added):

```
ERR failed to execute root command error="
error generating code: 
failed to execute stage 9: 
Modify types using configured type transforms: 
type transformer: 
type matcher [Group: 'dataprotection'; Name: 'BackupVaults_BackupPolicySpec'; Because: 'This property should be marked readonly in Swagger'], incomplete match for name: 
no match for \"BackupVaults_BackupPolicySpec\" (did you mean BackupVaults_BackupPolicy_Spec?)"
```

### Special notes

Much of the infrastructure needed was already present, so this involved less work than expected.

gif

## Checklist

- [x] this PR contains tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,3 +84,4 @@ hack/tools
 v2/controller
 v2/cmd/asoctl/asoctl
 v2/tools/generator/aso-gen
+v2/tools/generator/generator
--- a/v2/api/documentdb/v1api20210515/storage/database_account_types_gen.go
+++ b/v2/api/documentdb/v1api20210515/storage/database_account_types_gen.go
@@ -481,13 +481,6 @@ func (account *DatabaseAccount_Spec) AssignProperties_From_DatabaseAccount_Spec(
 		propertyBag.Remove("CreateMode")
 	}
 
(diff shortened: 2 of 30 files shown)


jakjang (approved):
Better error messages are always a plus 😃

PR #4941, opened 2025-09-16
Update our Azure rest api specs submodule

theunrepentantgeek:
## What this PR does

Updates our submodule at `v2/specs/azure-rest-api-specs` and our generated code.

### Breaking

Upstream work has resulted in some breaking changes that we'll need to call out in our next release.

#### CognitiveServices
* Upstream cardinality of Account.networkInjections changed (See [PR#36395](https://github.com/Azure/azure-rest-api-specs/pull/36395))

#### Compute
* Upstream definition PurchasePlan renamed to DiskPurchasePlan to avoid name collision (See [PR#32748](https://github.com/Azure/azure-rest-api-specs/pull/32748))
* Status property Snapshot.DiskState now correctly marked as readonly upstream, and removed from our Spec

In addition, there are many types that are no longer anonymous, but which have official names - our code generator defers to those names when present so some types have been renamed in our code. This will be transparent to CRD users, but anyone consuming our generated code directly may be impacted.

gif

## Checklist

- [x] this PR contains tests


v2/tools/generator/internal/config/configuration_group_shard_test.gox
```diff
@@ -0,0 +1,6 @@
+/*
+ * Copyright (c) Microsoft Corporation.
```
matthchr: What are these .gox files down here? Bunch are empty?

theunrepentantgeek: Debris from WIP; removed.

matthchr (approved):

Issue #2811, opened 2023-03-27
Reconcile should perform a diff with Azure


theunrepentantgeek:
At the moment we are doing a unilateral PUT of each resource when we reconcile; this works but has some drawbacks.

* Users with many resources have run up against throttling issues
* A couple of users have raised bugs because they thought the operator should only PUT when there's a change required

We should diff the current state of the resource with the spec and only do a PUT where required.

Follow-up to #1491 as much of that has been implemented already.

theunrepentantgeek:
Copied from [earlier comment](https://github.com/Azure/azure-service-operator/issues/1491#issuecomment-1302492095):

Some examples of where this is done in other operators:

[CAPZ](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/7eb7c4281575049e06c17a0effb95c774f234289/azure/services/agentpools/spec.go#L175)
[Terraform](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/internal/services/containers/kubernetes_cluster_node_pool_resource.go#L572)

matthchr:
Another idea that just occurred to me, documenting it here in case it's actually a good one (no guarantees): We could have a heuristic that reduced PUT load while at the same time not requiring the full investment and problems that come with client-side diffing.

A good example of such a heuristic would be:
1. After the PUT, take the shape of the returned Azure resource (the Status), hash it and store it in an annotation
2. Before any subsequent PUT, issue a GET and compare the hash vs the hash we have.

We could have manually crafted customizations per resource, or possibly automatically generate customizations from the Swagger to ignore fields like readonly times that are constantly changing and would always cause the payloads to mismatch.

This approach has a lot of advantages:
1. It works with Azure policy.
2. It's (relatively) easy to implement.
3. It doesn't require perfect upstream Swagger specifications to correctly notate every readOnly field

matthchr:
I believe that the above is actually [what Terraform does](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/internal/services/containers/kubernetes_cluster_node_pool_resource.go#L599) - from what I can tell they issue a GET after a PUT and then use the result of that GET as the statefile.

They only (manually) extract certain fields from the GET response: presumably fields that they know are writable, thus ignoring all of the readonly fields.

theunrepentantgeek:
We're still keen on doing this.

petrk-75:
[...]

PR #4962, opened 2025-09-29
Update our ADR on Azure differencing with latest discussions

theunrepentantgeek:
## What this PR does

Updates the existing ADR with information from several recent discussions and makes a recommendation on how to proceed.

Recommend using the markdown view to review.

gif

## Checklist


docs/hugo/content/design/ADR-2022-11-Change-Detection/_index.md
```diff
@@ -4,79 +4,115 @@ title: '2022-11: Change Detection'
 
 ## Context
 
-In ASO v2 up to at least the `beta.4` release, we reconcile each resource by doing a PUT, relying on the Azure Resource Manager (ARM) to do the goal state comparison and only update the resource if it has changed. 
+In ASO v2 up to at least the v2.15 release, we reconcile each resource by performing a PUT operation, relying on the Azure Resource Manager (ARM) to do goal state comparison and to only update the resource if it has changed.
 
-While this works, customers are already running up against severe ARM throttling with moderate numbers of resources. Typically, ARM throttles PUT requests handled by a given ARM endpoint to just 1200 per hour per subscription. With our current use of a single active HTTP connection and a reconcile period of 15m, ASO users are hitting this limit with just 300 active resources.
+While this approach works, some customers are running into ARM throttling limits with moderate numbers of resources. This occurs despite ARM throttling limits being lifted substantially since ASO went GA.
```
matthchr, line 9: Customers also sometimes run into issues where resources themselves are not idempotent (AKS AgentPool is an example of this although I am not sure we've had many complaints).

docs/hugo/content/design/ADR-2022-11-Change-Detection.md
```diff
@@ -4,79 +4,115 @@ title: '2022-11: Change Detection'
...
 
 * Some services normalize identifier fields (converting them to all lowercase)
 * Some fields accept a wider range of values than they return.
 
-For example the CosmosDB API will return the region `West US` when the resource specified `westus`.
+For example, the CosmosDB API will return the region `West US` when the resource specified `westus`.
+
+**Potential mitigations**:
+
+* Generally speaking, Azure identifiers are expected to be "case-preserving, case-insensitive", so we should compare those string fields in a case-insensitive manner.
```
matthchr: Only required for IDs (names, etc). Enum values and/or regions as we call out (also VM SKUs, which are somewhat enum-y) are all supposed to not be normalized but sometimes are.

tl;dr: I think we'll be playing whack-a-mole with fields if our approach doesn't handle them in a holistic fashion.

docs/hugo/content/design/ADR-2022-11-Change-Detection.md
```diff
@@ -4,79 +4,115 @@ title: '2022-11: Change Detection'
...
-* Generally speaking, Azure identifiers are expected to be "case-preserving, case-insensitive", so we should compare those string fields in a case insensitive manner. 
-* Region names and virtual machines SKUS are two known special cases where we may need to do some special case handling. 
+Some resources have arrays of items. These arrays may not be guaranteed to be returned in the same order as they were specified.
 
-### Array ordering
+**Potential mitigations**:
 
-Some resources have arrays of items. These arrays may not be guaranteed to be returned in the same order as they were specified. 
+* In the common case where the items in an array have a known identifier (e.g., when they are sub-resources with an `id` field), use that identifier to match them up. Where the items in an array do not have a known identifier, compare them by index.
+* Add specific configuration for array properties to specify how they should be compared — whether by an id field or by index. Consistent with our similar configuration used elsewhere, we'd require every array property to be annotated to ensure we don't do the wrong thing by default.
```
matthchr: FWIW Swagger actually has this field. It's called [x-ms-identifiers](https://github.com/Azure/autorest/tree/main/docs/extensions#x-ms-identifiers) and it can be used to call out which field to use as the key.

I think it has pretty wide adoption because there's a linter that checks for it. With that said, you can say `[]` in which case we'd have to assume "by index"

docs/hugo/content/design/ADR-2022-11-Change-Detection.md
```diff
@@ -4,79 +4,115 @@ title: '2022-11: Change Detection'
...
+What's the CPU overhead of doing a GET and diff on every reconciliation compared to just doing a PUT?
+
+We'll be adding another network round trip (with the associated bandwidth and latency costs) for every reconciliation, plus the CPU cost of doing the diff itself. We need to ensure that this doesn't cause us to exceed our resource limits in the controller pod, especially when reconciling large numbers of resources.
+
 ## Decision
 
-TBC
+Recommendation: Implement differencing as follows:
+
+* Update our code generator to add a `DiffWith()` method to every spec type that accumulates a list of the differences found
```
matthchr: Can we expand on the actual algorithm here some more? We cover above for example the fact that we'll cover the status type back to spec and discard the readonly fields, but that's not really covered here.

docs/hugo/content/design/ADR-2022-11-Change-Detection.md
```diff
@@ -4,79 +4,115 @@ title: '2022-11: Change Detection'
...
+* Add helper methods to a package `genruntime/diff` to help with common comparison tasks
+* Introduce `operatorStatus` on each top level status type, with a `knownDifferences map[string]string` field to capture the differences found.
```
matthchr: I'm not sure that I follow the need for `knownDifferences` (or maybe I am misunderstanding what it is doing?)

My thinking was the algo was basically:
1. PUT, get result of PUT and store in status (we already do this today so no change from today). Store the last observed generation in condition (We also already do this today). OR write a hash of spec to an annotation (I don't think this is needed I think generation is enough but maybe I'm missing some case where it is not).
2. Next time around reconcile loop, there are two cases: case A: generation has changed from observedGeneration of condition, and case B: generation has _not_ changed.
3. Case A: generation has changed from observedGeneration of condition. Means user has changed the spec so we issue a PUT (effectively same codepath as today).
4. Case B: generation has not changed from observedGeneration. Convert the current status to the spec shape (so discarding known readonly fields, etc) and hold that in-memory. This is `expected` . Issue a GET and update status. Now convert the newly updated status to the spec shape (discarding known readonly fields, etc). This is `actual`. Diff `expected` and `actual`. If there are any differences, log them and issue a `PUT`, otherwise no-op.

I _think_ this is basically what Terraform does and what I outlined in [this comment](https://github.com/Azure/azure-service-operator/issues/2811#issuecomment-1489081906), although I think I've explained it better here than I did there.

IMO this has a lot of advantages:
1. No extra storage required.
2. Only diffing Azure with itself, so we don't need to worry about normalization or policy because they should be being applied universally on all requests. We're effectively just diffing "last GET" with "current GET" rather than true "goal -> actual", but the result is the same if we use generation to get us the rest of the way there, which Kubernetes already computes for us.

There are some _small_ issues like for example using generation doesn't account for the fact that there are fields in spec that don't contribute to the upstream Azure shape, such as the CEL expression stuff, so if users change that we'll re-PUT. This is where using a hash + excluding those fields from the hash for case A would maybe be a slight improvement, but we could tackle that later if we really wanted to.

docs/hugo/content/design/ADR-2022-11-Change-Detection/_index.md
theunrepentantgeek: Added this as a motivating factor.

docs/hugo/content/design/ADR-2022-11-Change-Detection.md
theunrepentantgeek: Expanded with considerably more detail.

docs/hugo/content/design/ADR-2022-11-Change-Detection.md
theunrepentantgeek: When everything works properly, I think this approach will work - but I think it falls down if/when something goes wrong, as follows:

Scenario: Resource in Azure has recently drifted from desired state

spec<sub>0</sub> is our desired configuration
status<sub>0</sub> was observed before the resource drifted

Time passes. 
Time to reconcile.

Work out what we expect the resource to look like
spec<sub>expected</sub> := convertToSpec( status<sub>0</sub> )

status<sub>1</sub> := convertToStatus( GET<sub>1</sub> )
spec<sub>actual</sub> := convertToSpec( status<sub>1</sub> )

Diff( spec<sub>expected</sub>, spec<sub>actual</sub>) indicates there are changes, so Azure needs updating.

*But then updating Azure fails.*

* Maybe we never issue the PUT, having already saved the resource state
* Maybe the PUT is rejected by ARM
* Maybe the PUT succeeds, but the RP crashes
* Regardless, the resource has NOT been updated

Time passes.
Time to reconcile.

Work out what we expect the resource to look like
spec<sub>expected</sub> := convertToSpec( status<sub>1</sub> )

status<sub>2</sub> := convertToStatus( GET<sub>2</sub> )
spec<sub>actual</sub> := convertToSpec( status<sub>2</sub> )

Diff( spec<sub>expected</sub>, spec<sub>actual</sub>) indicates there are NO changes because the resource didn't change between GET<sub>1</sub>  and GET<sub>2</sub>.

**busted**.

We can try to finesse the failure modes to make sure we only commit status changes after we've updated Azure - but that has two problems:
* We don't update status until after we've finished a potentially long running operation (which users would notice)
* We run the risk of missing a failure mode.

If we are always comparing with the original status of the resource - status<sub>0</sub> - then we can't be caught out by odd failure modes. 

Introducing `OperatorStatus.knownDifferences` is an attempt to mitigate the large size of status<sub>0</sub> by only keeping the bits that matter.

matthchr: > We don't update status until after we've finished a potentially long running operation (which users would notice)

I don't think this is an issue. This is already the behavior we have today and it's also the behavior that Azure "wants" us to have. See [handleCreateorUpdateSuccess](https://github.com/Azure/azure-service-operator/blob/main/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go#L404) for example where we only write the status after the create/update LRO completes.

What I mean by this is the behavior Azure "wants" us to have: it's the way Azure's LRO/APIs are built - you issue a PUT, you get a polling URL, you poll that URL to completion and at the end of that polling the current resource state is returned to you. Obviously that doesn't stop you from breaking out of that flow and issuing separate GETs while the resource is in a transitioning state, but if you look at what all of the SDKs do _automatically_ it's always "update, wait, get status" - get status happens at the end..

The point you've raised about Azure interactions failing is a good one, we can't know if the changes we made landed or not in the case we get some connection error or internal server error, but I think as you say that is taken care of by not committing the status until the end. Though it's worth noting that it seems very likely that unless the service behavior is very weird, the update we're making to drive back to goal state is going to result in a status whose _spec component_ is the same as before, so for the purposes of the diffing algorithm there may not be a difference between the "old status" and the "new status". There are probably some resources where this isn't true for but it definitely won't be the majority of them.

> We run the risk of missing a failure mode.

What failure mode though? The above pattern seems safe to me? In fact I think it's actually "the same" as using known differences, isn't it? The two patterns are either:

1. Get Status after PUT and store someplace (in status)
2. Get Status after PUT, compute knownDifferences and store someplace.

In both cases, we rely on getting the status after the PUT, and rely on making an update via kube-apiserver to our object (either an annotation or status), so I am not clear on how there's any real difference between the two from a reliability perspective? The only case that I can see where there would be a difference would be if somebody went and updated the status in k8s to something that wasn't derived from Azure, which they could also do to knownDifferences (either in status or as an annotation).

matthchr: Maybe we should have a meeting where we talk this over more? The more I look at the two proposals the more they actually look like variants of the same idea (which is good), but I'm still not really seeing what paying the storage (and documentation) cost of adding a field for KnownDifferences actually buys us - we already have the data that was derived from (the whole status), so I don't think there's any reason to actually use it?

spec<sub>0</sub> + knownDifferences<sub>0</sub> = status<sub>0</sub>
or (moving spec to the other side of the equation)
knownDifferences<sub>0</sub> = status<sub>0</sub> - spec<sub>0</sub>

We already have status<sub>0</sub> and spec<sub>0</sub> (and we have to have them), so we can compute knownDifferences anytime we like on the fly, can't we?

In the case where the spec hasn't changed the two are (I think) identical. Where they maybe differ is when the spec has changed, with `knownDifferences` you would add those differences to the new spec<sub>1</sub> and compare w/ the GET and see a difference (at least in theory, though there are maybe some edge-cases where the new spec just sets one of the KnownDifferences so there is no diff - it's not clear to me if we want to send to Azure in this case or not. On the one hand it feels safe to avoid a PUT, OTOH that assumes that there's no difference between a field being defaulted versus a field being set by the user, and I suspect there are cases where those two things _are_ different and we would need to send a PUT to realize it), whereas if we don't keep knownDifferences we would just check generation of the status versus generation of the spec and blindly PUT regardless of diff.

TBH I see flaws with both of those approaches so it's not clear to me that one is objectively better than the other. I _think_ PUT-ing whenever user changes generation is _maybe_ a bit safer but obviously does eat a PUT request quota. Still, I'm not convinced the difference here is worth the cost of storing knownDifferences (especially because I think actually doing generation-based is safer)

docs/hugo/content/design/ADR-2022-11-Change-Detection/_index.md
```diff
@@ -0,0 +1,163 @@
...
+When creating or updating a resource:
+
+* After performing a PUT (or completing the LRO polling), capture the returned Status on the resource (this already happens).
+* Convert that Status to a Spec using our existing generated conversion code.
+* Compare the original Spec (the desired state) with the converted Spec (the actual state) to identify _known differences_ for the resource, capturing any normalizations or default values applied by Azure.
+* Store those _known differences_ in the operatorStatus on the resource.
+
+To support this functionality:
+
+* Update our code generator to add a `DiffWith()` method to every spec type that accumulates a list of the differences found.
```
matthchr, line 135: As discussed over chat, can you update this to clarify the slightly modified proposal where we think we can compute knownDifferences dynamically? (I'm also happy to contribute a paragraph on why I think that's possible/guaranteed correct, including the case/change we could make which would make it not guaranteed correct)

matthchr: Actually as discussed... thinking on this a bit more...

matthchr (approved):
Approved as-is. I think we can/should pivot if, when we implement this, it turns out that knownDifferences is huge for some resource types, as my concern is really large knownDiffs.

Let's see what we actually see in practice though, as I think having it makes the algorithm for diffing simpler.

docs/hugo/content/design/ADR-2022-11-Change-Detection.md
matthchr: Did some more thinking on this and tried writing up the two algo's and determined:

The downside of my approach of using the status we have is that it relies on that status as the source of truth, which means that when the spec changes we need to check if the generation has changed as well - this ends up making the algorithm more complicated because there's a few extra checks/steps.
 
So maybe we should err on the side of simpler algorithm for initial impl, w/ as proposed a field on status for knownDifferences (or we could use an annotation if we preferred, I'm open to either), and then fall back to my approach if we find that there are some resources with huge knownDifferences that are actually going to be problematic

PR #4966, opened 2025-10-02
Fix bug preventing resource reconciliation due to "Secret not cached"

### LINKED ISSUE #4926: Bug: Regression in v2.15.0: “/v1, Kind=Secret is not cached” during resource claim (fixed by downgrading to v2.14.0)
## Describe the bug

ASO version(s):

- Broken: v2.15.0
- Works: v2.14.0

**Kubernetes:** AKS (private), Kubernetes server v1.33.x
**Install method:** Helm (controller in azureserviceoperator-system)
**Controller SA:** azureserviceoperator-default
**Namespaces involved:** azureserviceoperator-system, tenant-blue

**What happened**

After upgrading to ASO v2.15.0, reconciles for ARM resources in tenant-blue fail early with:

`generic_reconciler.go:206] "Error claiming resource" err="/v1, Kind=Secret is not cached" logger="resources_resourcegroup" name="rg-tenant-blue" namespace="tenant-blue"
`
Downgrading back to v2.14.0 (same cluster and manifests) makes the error disappear and resources reconcile normally.

## Expected behavior

ASO should be able to claim/reconcile resources without failing on “Secret is not cached”. If Secrets are intentionally not cached, reads should use the uncached API reader, or the controller should register the Secrets informer for the watched namespaces.

## To Reproduce

Deploy ASO v2.15.0 in azureserviceoperator-system (SA azureserviceoperator-default).

Apply a simple ARM resource (e.g., resources.azure.com/v1api20200601, ResourceGroup) in namespace tenant-blue.

Observe controller logs:

`Error claiming resource ... err="/v1, Kind=Secret is not cached"
`

## Additional context

What I’ve already checked

### RBAC:

Granted get,list,watch (and also tried with create,update,patch) on core secrets to the controller SA in both namespaces:

- azureserviceoperator-system
- tenant-blue

**Verified with:**

```
kubectl auth can-i --as=system:serviceaccount:azureserviceoperator-system:azureserviceoperator-default \
  -n azureserviceoperator-system watch secrets   # yes
kubectl auth can-i --as=system:serviceaccount:azureserviceoperator-system:azureserviceoperator-default \
  -n tenant-blue watch secrets                   # yes
```

### Watch scope:

**Tried default (no scope var set) and also explicitly set:**

AZURE_TARGET_NAMESPACES=tenant-blue in aso-controller-settings Secret.

Rolled the deployment after each change.

### Controller restart:

`kubectl -n azureserviceoperator-system rollout restart deploy/azureserviceoperator-controller-manager
`
### ClusterRole test (to rule out namespace scoping issues):

Temporarily bound a ClusterRole with get,list,watch on Secrets across the cluster → still hit the same error on v2.15.0.

### AKS features/policies:

Azure Policy / Defender / Workload Identity do not affect list/watch on K8s Secrets; nothing else in AKS configuration changes the controller-runtime cache behavior.

All of the above did not resolve the issue on v2.15.0. Switching only the ASO version back to v2.14.0 resolved it immediately.

### Hypothesis

This looks like a behavior change in v2.15.0 where the controller now fails reads against types without a registered informer (e.g., stricter cache reader behavior such as ReaderFailOnMissingInformer). If any code path in the reconcile/claim phase reads a Secret via the cached client without registering a Secret informer (or without using the API reader), it will surface exactly as ErrResourceNotCached (“Kind=Secret is not cached”).

### Ask

- Is this a known regression/change in v2.15.0?
- Should ASO be registering an informer for corev1.Secret, or should the relevant read be using the API reader instead?
- Guidance/workaround for v2.15.0 users? (env flag/setting to restore previous behavior, or patch to register Secret informer)

Happy to provide more logs. Here’s the representative log line again:

`generic_reconciler.go:206] "Error claiming resource" err="/v1, Kind=Secret is not cached" logger="resources_resourcegroup" name="rg-tenant-blue" namespace="tenant-blue"
`

matthchr:
This was caused by PR #4857 setting ReaderFailOnMissingInformer. It happens only if the CRDs installed via crdPattern do not include a resource which requires a Secret informer.

This issue was missed in our testing because we generally run with a full (or close to full) set of CRDs installed, but the problem only appears with a minimal subset installed.

Fixes #4926.

## Checklist

- [ ] this PR contains tests

### DIFF
--- a/v2/cmd/controller/app/setup.go
+++ b/v2/cmd/controller/app/setup.go
@@ -416,7 +416,6 @@ func initializeClients(cfg config.Values, mgr ctrl.Manager) (*clients, error) {
 
 	armClientCache := armreconciler.NewARMClientCache(
 		credentialProvider,
-		kubeClient,
 		cfg.Cloud(),
 		nil,
 		armMetrics)
--- a/v2/internal/reconcilers/arm/arm_client_cache.go
+++ b/v2/internal/reconcilers/arm/arm_client_cache.go
@@ -15,7 +15,6 @@ import (
 	"github.com/Azure/azure-service-operator/v2/internal/genericarmclient"
 	"github.com/Azure/azure-service-operator/v2/internal/identity"
 	"github.com/Azure/azure-service-operator/v2/internal/metrics"
-	"github.com/Azure/azure-service-operator/v2/internal/util/kubeclient"
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
 )
 
@@ -26,14 +25,12 @@ type ARMClientCache struct {
 	clients            map[string]*armClient
 	cloudConfig        cloud.Configuration
 	credentialProvider identity.CredentialProvider
-	kubeClient         kubeclient.Client
 	httpClient         *http.Client
 	armMetrics         *metrics.ARMClientMetrics
 }
 
 func NewARMClientCache(
 	credentialProvider identity.CredentialProvider,
-	kubeClient kubeclient.Client,
 	configuration cloud.Configuration,
 	httpClient *http.Client,
 	armMetrics *metrics.ARMClientMetrics,
@@ -42,7 +39,6 @@ func NewARMClientCache(
 		lock:               sync.Mutex{},
 		clients:            make(map[string]*armClient),
 		cloudConfig:        configuration,
-		kubeClient:         kubeClient,
 		credentialProvider: credentialProvider,
 		httpClient:         httpClient,
 		armMetrics:         armMetrics,
--- a/v2/internal/reconcilers/arm/arm_client_cache_test.go
+++ b/v2/internal/reconcilers/arm/arm_client_cache_test.go
@@ -69,7 +69,7 @@ func NewTestARMClientCache(client kubeclient.Client) (*ARMClientCache, error) {
 		return nil, err
 	}
 
-	return NewARMClientCache(credentialProvider, client, cfg.Cloud(), nil, metrics.NewARMClientMetrics()), nil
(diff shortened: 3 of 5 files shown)


matthchr:
One open question here is if we think it's worth having a test that covers this case. Normally I would say we need to, but doing so would probably require us to run a separate `kind` environment that only installs a few CRDs. The question then becomes... which CRDs and what tests do we actually run there? I was able to reproduce this with just installing ResourceGroup, but other issues could at least theoretically show up with a different set of resources installed. We probably practically can't coven all combinations of resources... but it _does_ probably make sense to test a "complete" installation and a "minimal" installation?

Assuming there's agreement on that, I can file an item to myself to do this as a followup.

matthchr:
I also think we should consider picking this change to a v2.15.1 branch and releasing v2.15.1 with this fixed, even though v2.16 is coming soon, as it's a bad enough bug that it feels like it probably warrants that.

LMK what you think @theunrepentantgeek and @jakjang

jakjang:
> The question then becomes... which CRDs and what tests do we actually run there?

Maybe we can choose from some of the more popular CRDs/resources that are deployed w/ ASO; e.g. a subset of the top X number of them?

> Cherrypicking

I agree on the cherrypick; rereading #4926, it's an actively blocking issue right now. Argument can be made that since reverting version is a fix that should be the path until v2.16, but it'd stop v2.15.1 whenever `Error claiming resource` pops up.

jakjang (approved):

matthchr:
Conclusion is that I'll ship 2.15.1 once this merges (probably tomorrow).

PR #4967, opened 2025-10-02
Fix select annotation changed predicate

matthchr:
It was mistakenly classifying creates as containing an annotation change. This was previously harmless because we always reconciled the resource (say, ResourceGroup) if the obj was created anyway, but now that we also use this predicate for watching namespaces, we do not want events on namespace create, only on namespace updates that contain an annotation change.

## Checklist

- [x] this PR contains tests


v2/internal/reconcilers/generic/register.go
```diff
@@ -188,7 +188,7 @@ func registerNamespaceWatcher(kubeclient kubeclient.Client, gvk schema.GroupVers
 		}
 		// list the objects for the current kind
 
-		log.V(Verbose).Info("Detected namespace reconcile-policy annotation", "namespace", obj.GetName())
+		log.V(Verbose).Info("Detected namespace reconcile-policy annotation change", "namespace", obj.GetName(), "reconcile-policy", obj.GetAnnotations()["serviceoperator.azure.com/reconcile-policy"])
```
theunrepentantgeek: I like to make the name/value pairs more visible in our logging code:

```suggestion
		log.V(Verbose).Info(
			"Detected namespace reconcile-policy annotation change", 
			"namespace", obj.GetName(), 
			"reconcile-policy", obj.GetAnnotations()["serviceoperator.azure.com/reconcile-policy"])
```

matthchr: Done

jakjang (approved):

PR #4974, opened 2025-10-06
Fix taskfile ordering bug with zz_generated.go files

matthchr:
We run the full header-check script at v2/ directory for many tasks running in parallel. This can cause issues where the controller task deletes the zz_generated.go files after the check_headers script has discovered them via file-listing, which causes the script to error because the file was deleted out from under it.

Fix to only run check_headers on the directory in quesiton. This also has the added benefit of reducing duplication of checks slightly (we still have some because controller checks all of v2 which covers generator and asoctl anyway).

## Checklist

- [ ] this PR contains tests

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -122,7 +122,6 @@ tasks:
     deps: 
       - header-check
       - specifier-check
-      - controller:check-crd-size
 
   make-release-artifacts:
     desc: Produce files required for an ASOv2 release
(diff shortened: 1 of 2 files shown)


Taskfile.yml
```diff
@@ -1249,7 +1259,9 @@ tasks:
 
   crossplane:ci:
     deps: 
-      - basic-checks
+      - task: basic-checks
+        vars:
+          DIR: '{{.CROSSPLANE_ROOT}}'
```
theunrepentantgeek, line 1264: I didn't know you could do this. Nice.

theunrepentantgeek (approved):
I'd tried to fix this but never worked out where the problem came from.

PR #4976, opened 2025-10-07
Add PreReconcileOwnerCheck and implement for Kusto Databases

### LINKED ISSUE #4933: Bug: Kusto database creation attempts while cluster is stopped
## Describe the bug

We are creating some ADX clusters with auto stop enabled with some databases attached. When the cluster is stoped the `kusto_database` controller still attempts PUT requests.

```
Reason: Failed, Severity: Warning, RetryClassification: RetrySlow, Cause: error getting status for resource ID "<arm_id>": 
getting resource with ID: "<arm_id>": GET https://management.azure.com/<arm_id>
 RESPONSE 400: 400 Bad Request ERROR CODE: BadRequest 
{ "error": { "code": "BadRequest", "message": "Cannot fetch databases while resource is in state 'Stopped'." } }
```

Azure Service Operator Version: 2.15.0

## Expected behavior

Reconciliations should be retried without performing the PUT request that it's always failing.

## To Reproduce

Create a kusto cluster & database and then stop the cluster.

## Additional context

Doesn't seem to be exactly a terminal state, but seems like adding "stopped" to the [set](https://github.com/Azure/azure-service-operator/blob/b6e426001b6d457725abe987b449a647450263ce/v2/api/kusto/customizations/cluster_extensions.go#L32) would do the trick

theunrepentantgeek:
## What this PR does

It turns out existing `PreReconcileCheck` extension point has a fatal flaw - it relies on the ability to GET the child resource regardless of the state of the owner resource.

For Kusto Clusters, this is not the case - if the cluster is stopped, you can't do a GET of the child resource at all.

This PR adds a second extension point, similar to the first but without the requirement that the child resource be updated before the extension is invoked.

Resolves #4933 

gif


v2/api/kusto/customizations/database_extensions.go
```diff
@@ -19,23 +19,22 @@ import (
...
 
-var _ extensions.PreReconciliationChecker = &ClusterExtension{}
+var _ extensions.PreReconciliationOwnerChecker = &DatabaseExtension{}
 
 // Ensure we're dealing with the hub version of Cluster
 // If this no longer compiles (due to a newer version being imported), change the import above to that version.
 var _ conversion.Hub = &kusto.Cluster{}
 
-// PreReconcileCheck is called before the reconciliation of the resource to see if the cluster
+// PreReconcileOwnerCheck is called before the reconciliation of the resource to see if the cluster
```
matthchr, line 28: minor, maybe clarify docs here a bit to explain owner vs not?

theunrepentantgeek: Done.

Issue #4957, opened 2025-09-26
Feature: Add Support for Azure Quota Groups in ASO


vronnG-dw:
_Describe the feature request._

Currently, Azure Service Operator (ASO) does not support the `Microsoft.Quota/groupQuotas` resource type. Azure Quota Groups are critical for governance and capacity management, especially in multi-subscription setup

https://learn.microsoft.com/en-us/rest/api/quota/group-quotas?view=rest-quota-2025-09-01

Extend ASO to support the Azure Quota Group resource (Microsoft.Quota/groupQuotas).

matthchr:
Seems like a reasonable request.

PR #4979, opened 2025-10-11
feat: Add Azure Quota resource support to ASO v2

bingikarthik:
## What this PR does

This PR adds support for Azure Quota resources to Azure Service Operator (ASO) v2, enabling Kubernetes-native management of Azure quotas.

**Key additions:**
- New `GroupQuota` Kubernetes resource (`quota.azure.com/v1api20250901`)
- Complete code generation from Azure Quota REST API (2025-09-01)
- Integration test for CRUD operations
- YAML samples and documentation

Closes https://github.com/Azure/azure-service-operator/issues/4957

gif

## Checklist

- [x] this PR contains tests

### DIFF
--- a/v2/internal/controllers/quota_quota_crud_v20250901_test.go
+++ b/v2/internal/controllers/quota_quota_crud_v20250901_test.go
@@ -10,6 +10,7 @@ import (
 	"testing"
 
 	. "github.com/onsi/gomega"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	quota "github.com/Azure/azure-service-operator/v2/api/quota/v1api20250901"
@@ -24,7 +25,7 @@ func Test_Quota_Quota_v20250901_CRUD(t *testing.T) {
 	tc := globalTestContext.ForTest(t)
 
 	// Compute quotas must be scoped to provider-location, not ResourceGroup
-	providerLocationScope := fmt.Sprintf("/subscriptions/%s/providers/Microsoft.Compute/locations/%s", 
+	providerLocationScope := fmt.Sprintf("/subscriptions/%s/providers/Microsoft.Compute/locations/%s",
 		tc.AzureSubscription, *tc.AzureRegion)
 
 	quotaResource := &quota.Quota{
@@ -69,7 +70,7 @@ func Test_Quota_Quota_v20250901_CRUD(t *testing.T) {
 	tc.PatchResourceAndWait(old, quotaResource)
 
 	if quotaResource.Status.Properties != nil &&
-		quotaResource.Status.Properties.Limit != nil && 
+		quotaResource.Status.Properties.Limit != nil &&
 		quotaResource.Status.Properties.Limit.LimitValue != nil {
 		g.Expect(quotaResource.Status.Properties.Limit.LimitValue.Value).To(Equal(to.Ptr(200)))
 	}
@@ -83,7 +84,6 @@ func Test_Quota_Quota_v20250901_CRUD(t *testing.T) {
 	// We simply verify the quota exists and has the correct updated values
 	g.Expect(quotaResource.Status.Id).ToNot(BeNil())
 	armId := *quotaResource.Status.Id
-	
 	// Verify the quota still exists in Azure with updated values
 	exists, retryAfter, err := tc.AzureClient.CheckExistenceWithGetByID(tc.Ctx, armId, string(quota.APIVersion_Value))
 	tc.Expect(err).ToNot(HaveOccurred())
@@ -94,8 +94,6 @@ func Test_Quota_Quota_v20250901_CRUD(t *testing.T) {
 	// since Azure Quota API doesn't support deletion
 	old = quotaResource.DeepCopy()
 	quotaResource.SetOwnerReferences([]metav1.OwnerReference{}) // Remove owner reference
-	quotaResource.SetFinalizers([]string{})                    // Remove finalizers
+	quotaResource.SetFinalizers([]string{})                     // Remove finalizers
 	tc.Patch(old, quotaResource)
 }
-
-


v2/azure-arm.yaml
```diff
@@ -4481,6 +4481,11 @@ objectModelConfiguration:
       WorkspaceFeatures:
         ClusterResourceId:
           $referenceType: arm
+  quota:
+    2025-09-01:
+      Quota:
+        $exportAs: GroupQuota
```
theunrepentantgeek: Is this the right import/export combo?

Looking at the ARM/Bicep docs, it seems we have both [Quota](https://learn.microsoft.com/en-us/azure/templates/microsoft.quota/quotas?pivots=deployment-language-bicep) and [GroupQuota](https://learn.microsoft.com/en-us/azure/templates/microsoft.quota/groupquotas?pivots=deployment-language-bicep), so the cross-over where the resource `Quota` is exported as `GroupQuota` seems surprising.

v2/samples/quota/v1api20250901/v1api20250901_groupquota.yaml
theunrepentantgeek, line 1: Missing a recording file to demonstrate the sample works - see [Create a sample to demonstrate the resource](https://azure.github.io/azure-service-operator/contributing/add-a-new-code-generated-resource/create-a-sample/).

v2/internal/controllers/quota_groupquota_crud_v20250901_test.go
```diff
@@ -0,0 +1,66 @@
...
+						Value:           to.Ptr(100),
+						LimitObjectType: to.Ptr(quota.LimitType_LimitValue),
+					},
+				},
+			},
+		},
+	}
+
+	tc.CreateResourceAndWait(groupQuota)
+	defer tc.DeleteResourceAndWait(groupQuota)
```
theunrepentantgeek: Rather than scheduling the deletion like this, we prefer to explicitly test that deletions work properly at the end of the test - here's an example using [BatchAccount](https://github.com/Azure/azure-service-operator/blob/main/v2/internal/controllers/crd_batch_account_test.go).

v2/internal/controllers/quota_groupquota_crud_v20250901_test.go
theunrepentantgeek, line 1: Missing a recording to demonstrate the test passes - see [Write a test for the resource](https://azure.github.io/azure-service-operator/contributing/add-a-new-code-generated-resource/write-a-test/).

v2/azure-arm.yaml
bingikarthik: Ya! That was confusing naming on my part.
I originally intended to implement `Microsoft.Quota/groupQuotas` but ASO's generator couldn't handle that resource type properly due to its non-standard ARM path structure. So I ended up generating the general `Microsoft.Quota/quotas` resource instead, but mistakenly kept calling it "GroupQuota" in the export name.

I've now fixed this to be consistent - the `Quota resource` is exported as `Quota`. This provides general quota management functionality, which is still valuable even though it's not the specific GroupQuotas feature I initially aimed for.
Thanks for catching that - the naming should definitely match what we're actually generating!

v2/internal/controllers/quota_groupquota_crud_v20250901_test.go
bingikarthik: This was fixed now..

v2/samples/quota/v1api20250901/v1api20250901_groupquota.yaml
bingikarthik: This was added now, thanks for doc!

v2/internal/controllers/quota_groupquota_crud_v20250901_test.go
bingikarthik: This was added now, thanks for doc!

v2/api/quota/v1api20250901/structure.txt
```diff
@@ -0,0 +1,64 @@
...
+│       │   └── LimitValue: *Object (3 properties)
+│       │       ├── LimitObjectType: *Enum (1 value)
+│       │       │   └── "LimitValue"
+│       │       ├── LimitType: *Enum (2 values)
+│       │       │   ├── "Independent"
+│       │       │   └── "Shared"
+│       │       └── Value: *int
+│       ├── Name: *Object (1 property)
+│       │   └── Value: *string
+│       ├── Properties: *v1.JSON
```
matthchr, line 25: Thought this v1.JSON was weird, but confirmed that's the actual spec for the API

matthchr:
It looks like the test is failing because of the changes that @theunrepentantgeek asked you to do here:
https://github.com/Azure/azure-service-operator/pull/4979#discussion_r2425061361

Can you re-record this test and push?

The error is:
> [controller:test-integration-envtest]     logr.go:280: I2025-10-13T19:20:59Z] resources_resourcegroup "msg"="Resource deletion failure" name="asotest-rg-mpunbm" namespace="aso-test-quota-quota-crud" azureName="asotest-rg-mpunbm" resourceID="/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/asotest-rg-mpunbm" error="Delete \"https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/asotest-rg-mpunbm?api-version=2020-06-01\": Reason: ReconciliationFailedPermanently, Severity: Error, RetryClassification: None, Cause: cannot find go-vcr recording for request from test \"Test_Quota_Quota_CRUD\" (cassette: \"recordings/Test_Quota_Quota_CRUD\") (no responses recorded for this method/URL): DELETE https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/asotest-rg-mpunbm?api-version=2020-06-01 (attempt: 1750)\n\n"

bingikarthik:
> Can you re-record this test and push?

Done, it was pushed now..!

v2/samples/quota/v1api20250901/refs/v1api20200601_resourcegroup.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: resources.azure.com/v1api20200601
+kind: ResourceGroup
```
matthchr: You don't need this file here (see other samples don't have it). It's somewhat implied atm...

@theunrepentantgeek for thoughts on this, but for now you should remove this file.

bingikarthik: This file was removed now, thanks..!

v2/internal/testcommon/samples_tester.go
```diff
@@ -239,19 +239,25 @@ func (t *SamplesTester) setSamplesOwnershipAndReferences(samples map[string]clie
 		if asoType.Owner().Kind == resolver.ResourceGroupKind {
 			ownersName = t.rgName
 		} else if t.useRandomName {
-			// Check if the owner exists in refs, then continue. We don't use random names for refs so its correct anyway.
-			_, found := refs[asoType.Owner().Kind]
-			if found {
-				continue
-			}
+			// Skip direct ARM references as they don't need owner lookup by Kind
```
matthchr: hmm, it would probably be easier to just follow the existing sample pattern of referring to the RG by ref:
```
  owner:
    name: aso-sample-rg
    group: resources.azure.com
    kind: ResourceGroup
```

Like so - you dont need to create that RG in the refs folder, it will just work.

bingikarthik: got it, let me check the same!

bingikarthik: Could you please check now @mattchr?
Main change would be here: https://github.com/Azure/azure-service-operator/pull/4979/files#diff-129601e30b4f3f665d7dc7d3d4a7de8f3a2c2ed03d75f6613d26fab6c16aee66R80-R81
As cannot delete the quota's, Ref: https://learn.microsoft.com/en-us/rest/api/reserved-vm-instances/quotaapi#using-quota-api

theunrepentantgeek:
The bad news - CI failed with the following error:

```
[controller:test-integration-envtest] --- FAIL: Test_Samples_CreationAndDeletion (0.01s)
[controller:test-integration-envtest]     --- FAIL: Test_Samples_CreationAndDeletion/Test_Quota_v1api20250901_CreationAndDeletion (0.00s)
[controller:test-integration-envtest]         kube_per_test_context.go:167: creating recorder: required environment variable "AZURE_SUBSCRIPTION_ID" was not supplied
```

which means the recording is missing for sample validation.

The good news is that this is one of the last checks, so things should work once the recording is present.

matthchr:
I see you disabled the sample test for the quota sample... but wondering, it seemed like it was passing previously - was it not? The fact that the test in Go code passes makes me think the sample test actually would pass fine? Is that not the case?

bingikarthik:
> I see you disabled the sample test for the quota sample... but wondering, it seemed like it was passing previously - was it not? The fact that the test in Go code passes makes me think the sample test actually would pass fine? Is that not the case?

Ya, I have purposely removed it
Since we can't delete the quota's, please check: https://github.com/Azure/azure-service-operator/pull/4979/files#diff-129601e30b4f3f665d7dc7d3d4a7de8f3a2c2ed03d75f6613d26fab6c16aee66R80-R81
As cannot delete the quota's, Ref: https://learn.microsoft.com/en-us/rest/api/reserved-vm-instances/quotaapi#using-quota-api

theunrepentantgeek:
Quota can't be deleted from Azure - but I'm not yet convinced that means the sample should be skipped. 

ASO [knows](https://github.com/Azure/azure-service-operator/pull/4979/files#diff-d096d3aea9ca3619a2b05a90c0f8edacc6cb18f8c584127d53864052f66698fcR131-R137) that the Azure doesn't support Deletion of the resource:

``` go
// GetSupportedOperations returns the operations supported by the resource
func (quota *Quota) GetSupportedOperations() []genruntime.ResourceOperation {
	return []genruntime.ResourceOperation{
		genruntime.ResourceOperationGet,
		genruntime.ResourceOperationPut,
	}
}
```

This should mean that deleting the resource _from the cluster_ works just fine, as ASO won't try to delete it from Azure. 

Let me pull your branch and test that.

theunrepentantgeek:
It definitely doesn't work. 😲 

What do you think, @matthchr? Is this a bug or is there some reason for this behaviour I've forgotten?

v2/internal/controllers/quota_quota_crud_v20250901_test.go
```diff
@@ -0,0 +1,101 @@
...
+	// Create a Resource Group to own the Quota (since it's an extension resource)
+	// Use fixed names to match the recording
+	rg := &resources.ResourceGroup{
+		ObjectMeta: tc.MakeObjectMetaWithName("asotest-rg-mpunbm"),
+		Spec: resources.ResourceGroup_Spec{
+			Location: tc.AzureRegion,
+			Tags: map[string]string{"CreatedAt": "2001-02-03T04:05:06Z"},
+		},
+	}
+	tc.CreateResourceAndWait(rg)
```
theunrepentantgeek: You don't need to overcontrol like this - just use

``` go
```

And that will do the right thing for you.

v2/internal/controllers/quota_quota_crud_v20250901_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	quotaResource := &quota.Quota{
+		ObjectMeta: tc.MakeObjectMetaWithName("asotest-quota-gicqir"),
```
theunrepentantgeek: The use of suffixes to isolate tests (allowing them to be run in parallel against actual Azure) is a detail you don't normally need to worry about here.

Why not use

``` go
ObjectMeta := tc.MakeObjectMeta("asotest-quota")
```

The `tc.MakeObjectMetaWithName` variant is mostly used for things that must have specific exact names like `default`.

bingikarthik:
> as ASO won't try to delete it from Azure.

Actually, ASO trying to delete the quota from Azure while running the sample recording test:
image

theunrepentantgeek:
> Actually, ASO trying to delete the quota from Azure while running the sample recording test

Yes, you're right. I discovered that while testing your branch yesterday.

For now, I think the suppression of `quota` in `samples_test.go` is actually correct. 

To test this, I've triggered our workflow runs, but I expect they'll fail on the sample and we'll need to put the suppression back.

bingikarthik:
@matthchr Could you please re-trigger the CI? thanks..

matthchr:
Looks like you missed running gofmt? Can you fix and push and I'll re-run tests and we can merge?

bingikarthik:
> Looks like you missed running gofmt? Can you fix and push and I'll re-run tests and we can merge?

Sure, it's done now!

theunrepentantgeek:
Thanks @bingikarthik, much appreciated. We're intending on releasing v2.16 this week.

PR #4987, opened 2025-10-16
Improve resource condition/error when deletion is not possible in Azure

### LINKED ISSUE #3975: Bug: ASO should not try to delete undeletable resources

v2.7.0 (and earlier)

**Describe the bug**

Discovered while testing API version 2023-11-15 of `documentdb` - the type `SqlDatabaseContainerThroughputSetting` may not be deleted directly via ARM.

ASO knows this already, as we generate the correct list of supported operations in `sql_database_container_throughput_setting_types_gen.go`:

``` go
// GetSupportedOperations returns the operations supported by the resource
func (setting *SqlDatabaseContainerThroughputSetting) GetSupportedOperations() []genruntime.ResourceOperation {
	return []genruntime.ResourceOperation{
		genruntime.ResourceOperationGet,
		genruntime.ResourceOperationPut,
	}
}
```

However, we don't respect that when running, resulting in noisy errors:

```
test_logger.go:160: I2024-04-28T22:19:01Z] SqlDatabaseThroughputSettingController 
    "msg"="Error during Delete" 
    "err"="deleting resource "/subscriptions/82acd5bb-4206-47d4-9c12-a65db028483d/resourceGroups/asotest-rg-aokpof/providers/Microsoft.DocumentDB/databaseAccounts/sample-sqldb-account/sqlDatabases/sample-sql-db/throughputSettings/default": DELETE https://management.azure.com/subscriptions/82acd5bb-4206-47d4-9c12-a65db028483d/resourceGroups/asotest-rg-aokpof/providers/Microsoft.DocumentDB/databaseAccounts/sample-sqldb-account/sqlDatabases/sample-sql-db/throughputSettings/default
    ERROR CODE: MethodNotAllowed
    {
      "code": "MethodNotAllowed",
      "message": "Message: The requested verb is not supported."
    }
    " name="sample-sql-throughput" 
    namespace="aso-test-samples-creationanddeletion-test-sqldatabase-v1a-7e1fc" 
    azureName="default" 
    action="BeginDelete"
```

**Expected behavior**

We should respect the available verbs for the resource, and skip deletion from Azure where it's not possible, even though this may produce results surprising for some users.

theunrepentantgeek:
## What this PR does

Some Azure resources can't be directly deleted - you have to delete their parent.

Currently if one of those resources is represented by an ASO cluster resource, any attempt to delete the resource from the cluster fails with an indefinite sequence of repeated failures.

But we can enhance our workflow to return a more informative error, including details on how to unblock the deletion from the cluster.

Closes #3975 

gif

## Additional information

Manually tested the scenario where the required annotation is not present on the resource - as long as the resource is eventually deleted by other means (e.g. by deleting the parent or the containing resource group), the delete retry loop will eventually terminate.

### DIFF
--- a/v2/internal/testcommon/samples_tester.go
+++ b/v2/internal/testcommon/samples_tester.go
@@ -43,6 +43,7 @@ var wholeSampleExclusions = []*regexp.Regexp{
 	regexp.MustCompile(`/subscription/`),                        // Can't easily be run/recorded in our standard subscription
 	regexp.MustCompile(`/redhatopenshift/`),                     // This requires SP creation
 	regexp.MustCompile(`/documentdb/sqldatabase/v1api20210515`), // This is blocked by corp policy (can't set DisableLocalAuth)
+	regexp.MustCompile(`dbforpostgresql/v20250801`),             // Fails due to a race condition saving owner updates that can result in children getting stalled. See #5062
 }
 
 var exclusions = []*regexp.Regexp{


v2/samples/sql/v1api20211101/refs/v1api20210401_storageaccountsblobservice.yaml
```diff
@@ -3,6 +3,8 @@ kind: StorageAccountsBlobService
 metadata:
   name: sqlstorageservice
   namespace: default
+  annotations:
```
matthchr, line 6: Should we leave a comment here explaining why this annotation is here (given it's a sample?)

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -129,6 +130,11 @@ func (r *azureDeploymentReconcilerInstance) DetermineDeleteAction() (DeleteActio
 		return DeleteActionMonitorDelete, r.MonitorDelete, nil
 	}
 
+	if !genruntime.ResourceOperationDelete.IsSupportedBy(r.Obj) {
+		// Resource doesn't support delete, so we just remove the finalizer and stop managing it
```
matthchr: Uh, we don't remove the finalizer, right?

If we did, it would actually get deleted in Kubernetes. I don't see where we remove it though so I think we do not.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler.go
```diff
@@ -38,8 +38,9 @@ const (
 type DeleteAction string
 
 const (
-	DeleteActionBeginDelete   = DeleteAction("BeginDelete")
-	DeleteActionMonitorDelete = DeleteAction("MonitorDelete")
+	DeleteActionBeginDelete        = DeleteAction("BeginDelete")
+	DeleteActionMonitorDelete      = DeleteAction("MonitorDelete")
+	DeleteActionNotPossibleInAzure = DeleteAction("NotPossibleInAzure")
```
matthchr, line 43: Why add another "possible action" here, rather than just doing the check at the start if `StartDeleteOfResource` (or possibly `deleteResource` helper since then we could write an extension to override that behavior if we wanted?)

v2/internal/controllers/crd_dataprotection_backupinstance_20231101_test.go
```diff
@@ -308,3 +311,17 @@ func newBackupInstanceManagedCluster(tc *testcommon.KubePerTestContext, rg *reso
 	}
 	return cluster
 }
+
+func addAnnotation(
```
matthchr: Maybe add to test helpers/testcontext instead so others can use it?

v2/samples/sql/v1api20211101/refs/v1api20210401_storageaccountsblobservice.yaml
theunrepentantgeek: Done.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler.go
theunrepentantgeek: Because it's ARM specific, and the earliest point in the process where we know we have an ARM instance - and we have all the information needed to make the decision - is in `azureDeploymentReconcilerInstance.Delete()`. 

The first thing that method does is call `DetermineDeleteAction()`, which returns an enumeration value and a func to call, so it seemed a natural fit.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
theunrepentantgeek: Comment was obsolete, have updated.

v2/internal/controllers/crd_dataprotection_backupinstance_20231101_test.go
theunrepentantgeek: Good idea.

v2/api/dbforpostgresql/customizations/flexible_servers_firewall_rule_extensions.go
```diff
@@ -18,16 +18,15 @@ import (
 	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/extensions"
 )
 
-var _ extensions.PreReconciliationChecker = &FlexibleServersFirewallRuleExtension{}
+var _ extensions.PreReconciliationOwnerChecker = &FlexibleServersFirewallRuleExtension{}
 
-func (ext *FlexibleServersFirewallRuleExtension) PreReconcileCheck(
+func (ext *FlexibleServersFirewallRuleExtension) PreReconcileOwnerCheck(
```
matthchr, line 23: Was this change just a drive-by? (I assume yes?)

v2/internal/controllers/eventgrid_domain_v20200601_test.go
```diff
@@ -60,6 +60,7 @@ func Test_EventGrid_Domain(t *testing.T) {
 			Owner: testcommon.AsOwner(acct),
 		},
 	}
+	tc.AddAnnotation(&queueServices.ObjectMeta, "serviceoperator.azure.com/reconcile-policy", "detach-on-delete")
```
matthchr, line 66: Missing friendly comment?
> // Don't try to delete directly, this is not a real resource - to delete it you must delete its parent

matthchr: And below on some others too

v2/internal/controllers/eventgrid_topic_v20200601_test.go
```diff
@@ -116,6 +116,9 @@ func Topic_Subscription_CRUD(tc *testcommon.KubePerTestContext, rg *resources.Re
 		},
 	}
 
+	// Don't try to delete directly, this is not a real resource - to delete it you must delete its parent
+	tc.AddAnnotation(&queueService.ObjectMeta, "serviceoperator.azure.com/reconcile-policy", "detach-on-delete")
```
matthchr, line 121: Should we either add an explicit test, or just purposefully leave this out of one of these tests, to ensure that the test still passes?

Possibly an explicit test is better?

Create rg, resource, create child that can't be deleted, delete child. Then delete parent and assert that both parent/child end up deleted in Kuberentes (RG is still there). Then delete RG and clean up?

v2/internal/controllers/sql_server_crud_v20211101_test.go
matthchr: Remove this?

matthchr: Remove?

v2/internal/controllers/storage_storageaccount_crud_20210401_test.go
```diff
@@ -131,6 +134,9 @@ func StorageAccount_QueueServices_Queue_20210401_CRUD(tc *testcommon.KubePerTest
 		},
 	}
 
+	// Don't try to delete directly, this is not a real resource - to delete it you must delete its parent
+	tc.AddAnnotation(&queue.ObjectMeta, "serviceoperator.azure.com/reconcile-policy", "detach-on-delete")
```
matthchr: This doesn't seem right to me? We can delete queues for sure. We do just below in fact (see the defer)

v2/internal/controllers/documentdb_mongodb_crud_v20240815_test.go
matthchr: minor: clean up comment?

v2/internal/controllers/documentdb_mongodb_crud_v20231115_test.go
matthchr: minor: clean up comment?

v2/internal/controllers/documentdb_mongodb_crud_v20210515_test.go
matthchr: minor: Clean up comment?

v2/internal/controllers/dbforpostgresql_flexibleserver_crud_v1api20240801_test.go
```diff
@@ -225,6 +225,7 @@ func FlexibleServer_Configuration_20240801_CRUD(tc *testcommon.KubePerTestContex
 			Value:     to.Ptr("READ"),
 		},
 	}
+	tc.AddAnnotation(&configuration.ObjectMeta, "serviceoperator.azure.com/reconcile-policy", "detach-on-delete")
 
 	tc.CreateResourceAndWait(configuration)
 	// This isn't a "real" resource so it cannot be deleted directly
```
matthchr: minor: Clean up comment?

v2/internal/controllers/dbforpostgresql_flexibleserver_crud_v1api20230601preview_test.go
```diff
@@ -203,6 +203,7 @@ func FlexibleServer_Configuration_20230601Preview_CRUD(tc *testcommon.KubePerTes
 			Value:     to.Ptr("READ"),
 		},
 	}
+	tc.AddAnnotation(&configuration.ObjectMeta, "serviceoperator.azure.com/reconcile-policy", "detach-on-delete")
 
 	tc.CreateResourceAndWait(configuration)
 	// This isn't a "real" resource so it cannot be deleted directly
```
matthchr: Clean up comment?

v2/internal/controllers/dbforpostgresql_flexibleserver_crud_v1api20221201_test.go
```diff
@@ -166,6 +166,7 @@ func FlexibleServer_Configuration_20221201_CRUD(tc *testcommon.KubePerTestContex
 			Value:     to.Ptr("READ"),
 		},
 	}
+	tc.AddAnnotation(&configuration.ObjectMeta, "serviceoperator.azure.com/reconcile-policy", "detach-on-delete")
 
 	tc.CreateResourceAndWait(configuration)
 	// This isn't a "real" resource so it cannot be deleted directly
```
matthchr: Clean up comment? Or did you wanna leave these? (If so ignore the comments below asking about it on other files).

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -205,6 +211,37 @@ func (r *azureDeploymentReconcilerInstance) MonitorDelete(ctx context.Context) (
...
+	if !hasResourceID {
+		// No resource ID means nothing to delete
+		return ctrl.Result{}, nil
+	}
+
+	if _, _, err := r.getStatus(ctx, resourceID); err != nil {
+		if genericarmclient.IsNotFoundError(err) {
+			// Resource no longer exists
+			return ctrl.Result{}, nil
+		}
```
matthchr: What about if the err is not nil but the error was not `IsNotFound`?

Probably we shouldn't ignore it then?

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -205,6 +211,37 @@ func (r *azureDeploymentReconcilerInstance) MonitorDelete(ctx context.Context) (
...
+	}
+
+	msg := fmt.Sprintf(
+		"Resource does not support deletion in Azure; set annotation %s: %s to permit deletion in Kubernetes",
```
matthchr:
```suggestion
		"Resource does not support deletion in Azure; set annotation '%s: %s' to permit deletion in Kubernetes",
```

maybe?

v2/samples/documentdb/mongodb/v1api20240815/v1api20240815_mongodbdatabasecollectionthroughputsetting.yaml
```diff
@@ -3,6 +3,9 @@ kind: MongodbDatabaseCollectionThroughputSetting
 metadata:
   name: sample-mongodb-throughput
   namespace: default
+  annotations:
```
matthchr, line 6: One thing to consider with this, versus just not having the annotation here, which I am somewhat concerned about:

If we set detach-on-delete here for all the samples and people blindly copy+paste that, they may think that they're "un-setting" whatever they had configured on this ThroughputSetting, when in reality they are not.

The advantage of letting it fail is that they'll (at least in theory) see the resource stuck in deleting and thus conclude that the spec still applies (i.e. their changes were not reverted).

If we're going to do this in the samples (which I think I am OK with) we should be making it clear in the comment here that deleting this resource will not "undo" or otherwise edit the settings in Azure and that if they want them set back to the default values they must do that themselves on the resource by changing the spec (not 100% sure how to phrase that but you get the idea).

v2/api/dbforpostgresql/customizations/flexible_servers_firewall_rule_extensions.go
theunrepentantgeek: No - I spotted errors rerecording the test. A FirewallRule can't be accessed if the server isn't in a steady state.

v2/internal/controllers/dbforpostgresql_flexibleserver_crud_v1api20221201_test.go
theunrepentantgeek: Cleaned up.

v2/internal/controllers/eventgrid_domain_v20200601_test.go
theunrepentantgeek: Added.

v2/internal/controllers/eventgrid_topic_v20200601_test.go
theunrepentantgeek: There is a test where it's left out - the most recent PostgreSQL flexible server, if memory serves. It's not an explicit test targeting this scenario though.

v2/internal/controllers/sql_server_crud_v20211101_test.go
theunrepentantgeek: Revised the comment.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
theunrepentantgeek: The original intent was to minimize the change in behaviour, but exposing the error makes sense.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
theunrepentantgeek: Done.

v2/samples/documentdb/mongodb/v1api20240815/v1api20240815_mongodbdatabasecollectionthroughputsetting.yaml
theunrepentantgeek: I got 80% of the way there by adding this:

``` yaml
# We can delete it from the cluster by applying this annotation, but this won't change anything in Azure.
```

Not sure how to capture the last bit though.

v2/internal/controllers/eventgrid_topic_v20200601_test.go
matthchr: Maybe just add a comment in that test flagging that we don't set it there on purpose?

v2/internal/controllers/dbforpostgresql_flexibleserver_crud_v1api20221201_test.go
matthchr: Was this actually cleaned up/changed? I see the comment above was but not this one?

v2/samples/documentdb/mongodb/v1api20240815/v1api20240815_mongodbdatabasecollectionthroughputsetting.yaml
matthchr: I think what you've said is good enough. "won't change anything in Azure" should hopefully give them pause, if that's what they're trying to do.

matthchr (approved):

v2/internal/controllers/dbforpostgresql_flexibleserver_crud_v1api20221201_test.go
theunrepentantgeek: Missed that one. Did a different search, found three dregs - now fixed.

PR #5040, opened 2025-11-18
Allow skipping delete if a resource (or its parent) has already been removed

### LINKED ISSUE #5009: Bug: RoleAssignments stuck in deletion because of authorization (scope) issues
## Describe the bug

When a managed cluster (AKS) is deleted through the cluster-api it also triggers the deletion of the corresponding ASO resources.
ASO seems to clean up most of the azure resources without any issues. However, it struggles with `RoleAssignments`.

The `RoleAssignment` is attached to another ASO resource (owner: `StorageAccount`) & when a deletion is initiated (by deleting the resource group or storage account object) ASO deletes the resources in azure. The RoleAssignment remains in a `Failed` state with the following message:
```
RESPONSE 403: 403 Forbidden
ERROR CODE: AuthorizationFailed
{
  "error": {
    "code": "AuthorizationFailed",
    "message": "The client '<CLIENT_ID>' with object id '<OBJECT_ID>' does not have authorization to perform action 'Microsoft.Authorization/roleAssignments/delete' over scope '/subscriptions/<SUBSCRIPTION_ID>/resourceGroups/<CLUSTER_RG>/providers/Microsoft.Storage/storageAccounts/<STORAGE_ACCOUNT>/providers/Microsoft.Authorization/roleAssignments/<ROLE_ASSIGNMENT_UUID>' or the scope is invalid. If access was recently granted, please refresh your credentials."
  }
}
```

When I check the Azure portal, the storage account is indeed deleted so that the scope message above makes absolutely sense (scope is invalid because it doesn´t exist anymore) but shouldn´t the RoleAssignment be deleted first before the StorageAccount is even touched by ASO?

Azure Service Operator Version: v2.11.0 (via cluster-api-azure)

## Expected behavior

RoleAssignments are deleted first if they exist, before moving on with their owner.
Or: If the owner resource (e.g. `StorageAccount`) doesn´t exist anymore -> delete the `RoleAssignment` object since the referenced azure resource doesn´t exist anymore anyway.

## To Reproduce

1. create a `ResourceGroup`object
2. create a `StorageAccount` within that resource group
3. create a `UserAssignedIdentity` within that resource group
4. assign a role to that managed identity via `RoleAssignment` -> owner / scope is the `StorageAccount`
5. initiate a deletion either by...
5.1. ...deleting the `ResourceGroup`object or....
5.2. ...deleting the `StorageAccount` object
6. wait until the resources have been deleted
7. check out the `RoleAssignment` object which still exists, though in a `Failed` state complaining about the scope of the deletion request

## Additional context

We are using the cluster-api (CAPZ) to manage our kubernetes clusters in Azure. CAPZ includes the ASO deployment.
With that, we can not only deploy Kubernetes clusters in Azure but also configure all the other necessary components (e.g. Private Endpoints, Storage Accounts etc.).

I am aware, that the ASO version (v2.11.0) is already a bit older but I couldn´t find any github issue or feature / bugfix introducing any change on that matter (or maybe my searching skills are lacking). So I am unsure wether an update of the ASO controller would help or not.

theunrepentantgeek:
## What this PR does

Before issuing a DELETE to Azure, check to see if the resource has already been removed.

May resolve #5009 once all groups are migrated.

### Special notes

Most groups are currently in a deny list (see `azure_generic_arm_reconciler_instance.go`) so that we don't have to re-record all the groups at once. Tracking issue #5108 has been created for the follow-up tasks of removing groups from the deny-list.

gif

### DIFF
--- a/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
+++ b/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
@@ -922,6 +922,7 @@ func (r *azureDeploymentReconcilerInstance) deleteResource(
 		if _, _, err := r.getStatus(ctx, resourceID); err != nil {
 			if genericarmclient.IsNotFoundError(err) {
 				// Resource no longer exists
+				log.V(Info).Info("Resource is already gone, skipping issue of DELETE to Azure")
 				return ctrl.Result{}, nil
 			}
 		}
--- a/v2/internal/testcommon/kube_test_context_envtest.go
+++ b/v2/internal/testcommon/kube_test_context_envtest.go
@@ -183,6 +183,12 @@ func createSharedEnvTest(
 	if cfg.Replaying {
 		requeueDelay = 10 * time.Millisecond
 		minBackoff = 10 * time.Millisecond
+
+		// maxBackoff is the maximum length of time we wait to retry.
+		// This used to be 10ms during testing but this caused an extremely large number of rapid retries and
+		// effectively starved other resources of CPU, resulting in tests taking much longer to complete because they
+		// progressed slowly. Increasing this to 1 second seems to strike a better balance between fast retries and
+		// allowing other resources to make progress.
 		maxBackoff = 1000 * time.Millisecond
 	}
 


theunrepentantgeek:
@juice928 stop.

We welcome human contributors to ASO, and request that any AI contributions be human mediated/reviewed.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -865,6 +914,19 @@ func (r *azureDeploymentReconcilerInstance) deleteResource(
...
 	}
 
+	// Check to see if the resource has already been deleted from Azure - if so, we're done.
+	// But, first check to see if this resource is in a deny group, and skip the check if so.
+	// This is to allow us to fix up remaining issues one by one instead of all at once.
+	group := obj.GetObjectKind().GroupVersionKind().Group
+	if !skipDeletionPrecheck.Has(group) {
+		if _, _, err := r.getStatus(ctx, resourceID); err != nil {
+			if genericarmclient.IsNotFoundError(err) {
+				// Resource no longer exists
```
matthchr, line 924: We used to always have this log:
```
log.V(Info).Info("Successfully issued DELETE to Azure")
```

now, it's possible we won't get that far. Should we have a log here that says something like 
```
log.V(Info).Info("Determined resource was already deleted in Azure - not issuing DELETE as it is not required")
```

or something like that?

v2/internal/testcommon/vcr/redact_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+package vcr
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+)
+
+func Test_hideAppConfigurationKeySecrets(t *testing.T) {
```
matthchr, line 14: minor, style: Would it be better to test via the exported `func (r *Redactor) HideRecordingData(s string) string {` interface?

Feel like such a test scales better to testing the other redactions we have as well (good idea adding tests here btw).

v2/internal/testcommon/kube_test_context_envtest.go
```diff
@@ -183,7 +183,7 @@ func createSharedEnvTest(
 	if cfg.Replaying {
 		requeueDelay = 10 * time.Millisecond
 		minBackoff = 10 * time.Millisecond
-		maxBackoff = 10 * time.Millisecond
+		maxBackoff = 1000 * time.Millisecond
```
matthchr, line 192: Can you expand on why this needs to be here? It's not mentioned in the PR description, or maybe a code-comment here would be useful?

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
theunrepentantgeek: Good idea. Done.

v2/internal/testcommon/vcr/redact_test.go
theunrepentantgeek: Definitely need to improve our test coverage here. That said, I specifically wanted to test the new redaction in isolation, to avoid the problems of other redactions getting in the way. I'll take a TODO to improve coverage.

v2/internal/testcommon/kube_test_context_envtest.go
theunrepentantgeek: Added a comment.

PR #5050, opened 2025-11-24
Auto-detect CRD versions to deprecate

### LINKED ISSUE #5045: CRD deprecation controller could auto-detect CRDs to migrate

We migrate only a hardcoded list

## Describe the improvement

We could auto-detect and migrate anything that has a newer storage version. This would remove the manual toil of keeping the list of deprecated versions up to date.

matthchr:
Fixes #5045.

## Checklist

- [x] this PR contains tests

### DIFF
--- a/v2/cmd/asoctl/internal/crd/cleaner.go
+++ b/v2/cmd/asoctl/internal/crd/cleaner.go
@@ -97,9 +97,8 @@ func (c *Cleaner) Run(ctx context.Context) error {
 		asoCRDsSeen++
 		newStoredVersions, deprecatedVersions := crdmanagement.GetDeprecatedStorageVersions(crd)
 
-		// If there is no new version found other than the matched version, we short circuit here, as there is no updated version found in the CRDs
-		if len(newStoredVersions) <= 0 {
-			// TODO: test?
+		// If there is no new version found other than alpha/beta, it means there wasn't a newer version installed/reconciled yet and we short circuit here, as there is no updated version found in the CRDs
+		if len(newStoredVersions) == 1 && (strings.HasPrefix(newStoredVersions[0], "v1alpha") || strings.HasPrefix(newStoredVersions[0], "v1beta")) {
 			return eris.Errorf("it doesn't look like your version of ASO is one that supports deprecating CRD %q, versions %q. Have you upgraded ASO yet?", crd.Name, deprecatedVersions)
 		}
 
--- a/v2/internal/crdmanagement/deprecation.go
+++ b/v2/internal/crdmanagement/deprecation.go
@@ -12,7 +12,7 @@ import (
 	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 )
 
-var versionRegexp = regexp.MustCompile(`^v(1api)?(\d{8})(preview)?(storage)?$`)
+var versionRegexp = regexp.MustCompile(`^v(1api|1alpha1api|1beta)?(\d{8}|\d)(preview)?(storage)?$`)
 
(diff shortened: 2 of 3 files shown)


v2/internal/crdmanagement/deprecation.go
```diff
@@ -66,3 +50,54 @@ func GetAllDeprecatedStorageVersions(crds map[string]apiextensions.CustomResourc
...
+// CompareVersions compares two group version strings, for example: v1api20230101preview, v1api20220101, v20220101, v1api20220101storage, and v20220101storage
+// and returns -1 if a < b, 0 if a == b, and 1 if a > b. The storage suffix is ignored in the comparison.
+// Preview versions are considered less than non-preview versions.
+func CompareVersions(a string, b string) int {
+	matchA := versionRegexp.FindStringSubmatch(a)
+	matchB := versionRegexp.FindStringSubmatch(b)
+
+	if matchA == nil || matchB == nil {
+		// If either version doesn't match the expected pattern, fall back to string comparison
+		// TODO: Do we actually want to do this?
```
matthchr, line 63: Maybe I should look at how Kuberentes orders versions and see if they export a function for that and use that...?

matthchr: Ah wait we can't do that because we don't (yet) comply fully with that versioning scheme so we'll need our own.

theunrepentantgeek: We do already have a proper version comparison routine that might be useful here - search for `ComparePathAndVersion()`. I'm pretty sure we've exposed this outside the generator.

matthchr: You're right. We would need to move where it is (from the generator to `v2/internal` so that both could import it).

Are you thinking that we would just it just for this fallback case where the patterns don't match, or are you thinking we'd replace the whole compare here with it? I looked at replacing the whole compare but it doesn't really give the ordering we need and I don't think it necessarily makes sense for it to do so since it seems to be attempting to be generic (and work for paths such as import paths too) whereas this is specific and really works best for kuberentes versions (the versions we emit).

wdyt?

theunrepentantgeek: On reflection, using that might be overkill - and it doesn't quite give the ordering we want here.

theunrepentantgeek (approved):

v2/cmd/asoctl/internal/crd/cleaner.go
```diff
@@ -97,9 +97,8 @@ func (c *Cleaner) Run(ctx context.Context) error {
 		asoCRDsSeen++
 		newStoredVersions, deprecatedVersions := crdmanagement.GetDeprecatedStorageVersions(crd)
 
-		// If there is no new version found other than the matched version, we short circuit here, as there is no updated version found in the CRDs
-		if len(newStoredVersions) <= 0 {
-			// TODO: test?
+		// If there is no new version found other than alpha/beta, it means there wasn't a newer version installed/reconciled yet and we short circuit here, as there is no updated version found in the CRDs
+		if len(newStoredVersions) == 1 && (strings.HasPrefix(newStoredVersions[0], "v1alpha") || strings.HasPrefix(newStoredVersions[0], "v1beta")) {
```
matthchr, line 101: This preserves the behavior of the `asoctl clean` command when the behavior of the underlying `GetDeprecatedStorageVersions` changed _slightly_.

Previously, `GetDeprecatedStorageVersions` would count a version as deprecated even if it was the last version in the list of stored versions. Since this is not actually valid, we've updated the logic to where it doesn't suggest to deprecate something if it's the latest/last version.

So if we want to error on this case in asoctl we needed a slightly more explicit check.

theunrepentantgeek (approved):

PR #5051, opened 2025-11-25
Allow opting out of property conversions if an error occurs

theunrepentantgeek:
## What this PR does

Allows a specific property to opt-out of property conversions, leaving room for manual intervention.

New property config `$conversionStrategry` is only consulted if an attempted direct property conversion generates an error; if set to `manual`, the proposed conversion is dropped, leaving the properties without conversion.

### Special notes

Manual implementation of the appropriate `augment*` interface(s) will be necessary to make the generated conversion tests pass.

gif


v2/azure-arm.yaml
```diff
@@ -2373,6 +2385,12 @@ objectModelConfiguration:
       ManagedClusters_MaintenanceConfiguration:
         $exportAs: MaintenanceConfiguration
         $supportedFrom: v2.11.0
+      ScaleProfile:
```
matthchr, line 2388: There's a newer version of ManagedCluster in `2025-08-01`, does this not go there?

matthchr (approved):
Had one small question

v2/azure-arm.yaml
theunrepentantgeek: No - the preview version (I think it was `2025-07-02preview` ?) got inserted in between the two.

PR #5067, opened 2025-12-04
Abort tests in case of go-vcr failure

theunrepentantgeek:
## What this PR does

We're seeing an increasing number of test failures where a particular resource gets stuck in a non-terminal condition, leaving the test to iterate in place until it times out. This failure mode means that it takes a long time to iterate on fixes - CI can take over an hour to run, most of that time spent in a failed condition.

This PR makes changes to improve reliability of tests, and to ensure failed tests terminate quickly - these should make it easier to iterate on fixes for remaining underlying issues.

* Limit the number of replays of a GET response to 10
* Only replay GET responses for terminal states (e.g. don't replay for `Creating`, `Deleting`, `InProgress`, etc)
* Don't use the `TEST-REQUEST-ATTEMPT` header when matching requests (this was needed for go-vcr v1 but not for newer versions)
* Immediately abort the `Eventually()` wait if `go-vcr` doesn't find a response to playback

gif

### DIFF
--- a/v2/internal/testcommon/vcr/v3/replay_roundtripper_test.go
+++ b/v2/internal/testcommon/vcr/v3/replay_roundtripper_test.go
@@ -84,6 +84,7 @@ func TestReplayRoundTripperRoundTrip_GivenSingleGETReturningNonterminalState_Ret
 	assertExpectedResponse(t, replayer, req, 200, `{"provisioningState": "Deleting"`)
 
 	// Assert - second request fails
+	//nolint:bodyclose // there's no actual body in this response to close
 	_, err := fake.RoundTrip(req)
 	g.Expect(err).To(MatchError(ContainSubstring("requested interaction not found")))
 }


theunrepentantgeek:
The current CI test failures appear to be the "stale owner status" race condition that @matthchr has already identified and plans to fix early in the new year.

v2/internal/testcommon/vcr/v3/replay_roundtripper.go
```diff
@@ -30,13 +32,23 @@ import (
...
 }
 
+type replayResponse struct {
+	response         *http.Response
+	remainingReplays int
+}
+
+const (
+	putReplays = 1  // Number of times to replay a PUT request
+	getReplays = 10 // Number of times to replay a GET request
```
matthchr:
```suggestion
	getReplays = 10 // Maximum number of times to replay a GET request
```

Is that right?

v2/internal/testcommon/vcr/v3/replay_roundtripper.go
```diff
@@ -30,13 +32,23 @@ import (
...
 	redactor *vcr.Redactor
 }
 
+type replayResponse struct {
+	response         *http.Response
+	remainingReplays int
+}
+
+const (
+	putReplays = 1  // Number of times to replay a PUT request
```
matthchr:
```suggestion
	putReplays = 1  // Maximum number of times to replay a PUT request
```

matthchr: You could probably change the variable names here too to include `max`?

v2/internal/testcommon/vcr/v3/replay_roundtripper.go
```diff
@@ -30,13 +32,23 @@ import (
...
 	padlock  sync.Mutex
 	redactor *vcr.Redactor
 }
 
+type replayResponse struct {
+	response         *http.Response
+	remainingReplays int
+}
+
+const (
```
matthchr, line 47: minor: Some exposition about why we set these maxes (what is bad if we don't) would be good.

I think the reasoning is, because of how we have our test framework set up, we don't expect to need to replay a GET more than 1-2x or a PUT ever. Anything more than that is "suspicious" and we'd rather fail at that time than wait for the timeout after replaying a zillion times?

v2/internal/testcommon/vcr/v3/replay_roundtripper.go
```diff
@@ -69,6 +81,8 @@ func (replayer *replayRoundTripper) RoundTrip(request *http.Request) (*http.Resp
 }
 
 func (replayer *replayRoundTripper) roundTripGet(request *http.Request) (*http.Response, error) {
+	requestURL := request.URL.RequestURI()
```
matthchr, line 95: Was `request.URL.String()`, now `request.URL.RequestURI()` - is there a difference?

v2/internal/testcommon/vcr/v3/replay_roundtripper.go
```diff
@@ -81,20 +95,34 @@ func (replayer *replayRoundTripper) roundTripGet(request *http.Request) (*http.R
...
+			delete(replayer.gets, requestURL)
 		}
 
 		// No cached response, return the original response and error
 		return response, err
 	}
 
-	replayer.padlock.Lock()
-	defer replayer.padlock.Unlock()
+	// We have a response, skip caching if it represents a transient state (ends in 'ing')
```
matthchr: It might be cleaner to just define which states are stable, since IIRC ARM only has 3?

v2/internal/testcommon/vcr/v3/replay_roundtripper.go
```diff
@@ -155,3 +188,56 @@ func (replayer *replayRoundTripper) hashOfBody(request *http.Request) string {
...
+	if err := json.Unmarshal([]byte(body), &res); err == nil {
+		if res.Properties.ProvisioningState != "" {
+			return res.Properties.ProvisioningState, true
+		}
+	}
+
+	return "", false
+}
+
+func (replayer *replayRoundTripper) bodyOfResponse(response *http.Response) string {
```
matthchr, line 259: Are we correctly gc-ing/disposing of the body here? (I think we are because it happens at a lower level?)

v2/internal/testcommon/vcr/v3/replay_roundtripper.go
theunrepentantgeek: That's right. I'll add some commentary.

v2/internal/testcommon/vcr/v3/replay_roundtripper.go
theunrepentantgeek: Good idea.

v2/internal/testcommon/vcr/v3/replay_roundtripper.go
theunrepentantgeek: `String()` includes the scheme and host, where `RequestURI` is just the path?query. 

I don't remember why I made the change, but I believe this will still cache correctly.

v2/internal/testcommon/vcr/v3/replay_roundtripper.go
theunrepentantgeek: I've refactored this code for clarity, and to make this change.

v2/internal/testcommon/vcr/v3/replay_roundtripper.go
theunrepentantgeek: Yes; while we have an `http.Response` and body is an `io.ReadCloser`, `go-vcr` has already read and replaced the original stream and replaced it with a string, so there's nothing to GC.

v2/internal/testcommon/vcr/v3/replay_roundtripper.go
theunrepentantgeek: Yes.

v2/internal/testcommon/vcr/v3/replay_roundtripper.go
```diff
@@ -53,7 +53,11 @@ const (
 	// GET requests may be replayed multiple times to allow multiple reconciles to observe the same stable final state.
 	// We set this to accommodate timing variations during test replay, while avoiding unbounded replays as they might
 	// result in a test getting stuck and continuing to run until the entire test suite times out.
-	maxGetReplays = 10 // Maximum number of times to replay a GET request
+	//
+	// Currently we need to set this to an apparently comical limit because some of our tests requiring many many
+	// repetitions to work due to changes in timing used during test replay. Once we've addressed other issues causing
+	// test instability, we should be able to reduce this limit significantly.
```
matthchr, line 59: minor: Do we have GH issues tracking these? Can we link here? Do we know the specific issues we think are making this number large?

theunrepentantgeek: I've added the one I know about to #5109

PR #5076, opened 2025-12-10
Improve testing with a kind cluster

theunrepentantgeek:
## What this PR does

When running `controller:kind-create-wi` within a devcontainer, the environment variable `HOSTROOT` is required in order for docker (which runs _outside_) the devcontainer to find the right PKI files. 

I've added a precondition to the target to ensure this variable is defined before running `kind`.

Also tidies the docs a little.

gif

## Checklist

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -930,9 +930,9 @@ tasks:
     status:
       - "kind get clusters | grep \"^asov2-wi$\""
     preconditions:
-      # If our working directory starts with /workspaces then we're running in a container and need HOSTROOT to be defined
-      - sh: '[ "${PWD#/workspaces}" = "$PWD" ] || [ -n "${HOSTROOT-}" ]'
-        msg: "When running in a container (detected via PWD starting with /workspaces) you must set HOSTROOT to the host's root path so that files can be mounted into kind clusters correctly."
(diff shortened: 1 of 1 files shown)


theunrepentantgeek:
@copilot, please `merge main` into this PR.

Taskfile.yml
```diff
@@ -929,6 +929,10 @@ tasks:
             {{.SCRIPTS_ROOT}}/kind-with-registry.sh"
     status:
       - "kind get clusters | grep \"^asov2-wi$\""
+    preconditions:
+      # If our working directory starts with /workspaces then we're running in a container and need HOSTROOT to be defined
```
matthchr: Is this assumption actually correct? I don't think it necessarily is? My container at least I often don't run at `/workspaces`, and I think the same is true of the CI containers (it clones into `/workspace` not `/workspaces`).

At least for docker containers you can check:
`[ -f /.dockerenv ]`

which is probably more reliable?

theunrepentantgeek: My devcontainer environment always uses `/workspaces`, good to know that's not always the case in other contexts.

I've changed to use `[ -f /.dockerenv ]` as that will be more reliable.

PR #5092, opened 2026-01-09
Do not give up retrying on most common errors

### LINKED ISSUE #4546: Bug: Creation of storage account fails permanently if KeyVault inaccessible
## Describe the bug

As described in #4459 

> When attempting to create a storage account with customer managed key encryption, if the user-assigned managed identity of the storage account does not have RBAC to the Key Vault, the operator/Azure correctly errors out for that resource saying that there is a key vault authentication failure.

## Expected behavior

This shouldn't be a permanent error - ASO should continue reconciling the storage account, with that reconciliation succeeding after the RBAC for the storage accounts user assigned managed identity gets updated.

## Additional context

See the full discussion in #4459

### LINKED ISSUE #4902: Bug: Resource is never reconciled if Azure API has a temporary issue
## Describe the bug

We had a resource created via Azure Service Operator (`FederatedIdentityCredential`). This resource was created in March. 

On a recent reconciliation we got an error of `MethodNotAllowed`, see below:

```
Status:
  ...
  Conditions:
    Last Transition Time:  2025-08-22T19:32:05Z
    Message:               The request format was unexpected : Support for federated identity credentials not enabled: PUT https://management.azure.com/subscriptions/redacted....
ERROR CODE: MethodNotAllowed
{
  "error": {
    "code": "MethodNotAllowed",
    "message": "The request format was unexpected : Support for federated identity credentials not enabled"
  }
}

    Observed Generation:  1
    Reason:               MethodNotAllowed
    Severity:             Error
    Status:               False
    Type:                 Ready
```

According to Azure documentation [here](https://learn.microsoft.com/en-us/entra/workload-id/workload-identity-federation-considerations#errors), this error occurs if you are trying to create this resource in an unsupported region (of which [these](https://learn.microsoft.com/en-us/entra/workload-id/workload-identity-federation-considerations#unsupported-regions-user-assigned-managed-identities) are the only two that are unsupported). This FederatedIdentityCredential already existed since March, and is in the `eastus` region. So this leads to the suspicion that the Azure API had a temporary blip/outage/something went wrong.

If the Azure Service Operator was to reconcile this resource again afterwards, then this wouldn't be an issue. However, [MethodNotAllowed](https://github.com/Azure/azure-service-operator/blob/main/v2/internal/reconcilers/arm/errorclassification/error_classifier.go#L88) is classified as a fatal error, which gives it an [Error Severity](https://github.com/Azure/azure-service-operator/blob/aa89d188ef482440864274b313201376e9cc0bb6/v2/internal/reconcilers/arm/errorclassification/ready_condition.go#L51-L62). Then the logic [here](https://github.com/Azure/azure-service-operator/blob/main/v2/internal/util/interval/interval_calculator.go#L136-L140) means that it fast return and stops tracking the resources in those cases. This tracks with our logs for the Azure Service Operator, where we see the error and then no more reconciliation attempts against this resource.

For us, this means that a temporary API mishap leaves this resource broken in the cluster until we either restart the controllers or delete/re-apply the resource.

Azure Service Operator Version: 2.12.0

## Expected behavior

Ideally, Azure wouldn't have API issues that could put our resource in this state, since I understand the reasoning behind not wanting to retry if the error suggests it won't ever be recoverable. However, if the API could do this unintentionally, then maybe it's worth reconciling regardless with a backoff?

## To Reproduce

Tough to reproduce, but I guess for the same behaviour you could cause another error like `PasswordTooLong` and validate that reconciliations against the resource don't occur.

### LINKED ISSUE #5074: Bug: RequestDisallowedByPolicy is not Retryable
## Describe the bug
When Resources like `ResourceGroupPolicyExemption` and `StorageAccount` are applied simultaneously, `StorageAccount` gets stuck in `RequestDisallowedByPolicy` and never recovers.

```yaml
status:
  conditions:
    - lastTransitionTime: '2025-12-09T20:16:43Z'
      message: >-
        Resource '<name>' was disallowed by policy. Policy identifiers:
        '[{"policyAssignment":{"name":"Storage accounts should disable public
        network
        access","id":"/providers/Microsoft.Management/managementGroups/<id>/providers/Microsoft.Authorization/policyAssignments/<id>"},"policyDefinition":{"name":"Storage
        accounts should disable public network
        access","id":"/providers/Microsoft.Authorization/policyDefinitions/<id>","version":"1.0.1"}}]'.
      observedGeneration: 1
      reason: RequestDisallowedByPolicy
      severity: Error
      status: 'False'
      type: Ready

```

Azure Service Operator Version: v2.16.0

## Expected behavior
`StorageAccount` is eventually reconciled.

## To Reproduce
1. Create a Resource block by Azure Policy
2. Exampt Resource from the Policy

matthchr:
We still give up retrying on some well-known non-retryable errors, although we're open to changing even this behavior if it makes sense.

Fixes #4902.
Fixes #5074.
Fixes #4546.

## Checklist

- [x] this PR contains tests


v2/internal/reconcilers/arm/errorclassification/error_classifier.go
```diff
@@ -66,54 +66,59 @@ func classifyCloudError(err *genericarmclient.CloudError) core.CloudErrorDetails
...
 	return result
 }
 
-func classifyHTTPError(err *genericarmclient.CloudError) core.ErrorClassification {
+func classifyHTTPError(err *genericarmclient.CloudError) (core.ErrorClassification, retry.Classification) {
 	var httpError *azcore.ResponseError
 	if !eris.As(err.Unwrap(), &httpError) {
-		return core.ErrorRetryable
+		// If we can't determine the error type, assume we can retry
+		return core.ErrorRetryable, retry.None
```
theunrepentantgeek: Inconsistent return values - the error is retryable, but the retry classification is None. 

Probably want to use either `RetrySlow` or `RetryVerySlow`.

v2/internal/reconcilers/arm/errorclassification/error_classifier.go
```diff
@@ -66,54 +66,59 @@ func classifyCloudError(err *genericarmclient.CloudError) core.CloudErrorDetails
...
 
 	if httpError.StatusCode == 400 {
-		return core.ErrorFatal
+		// HTTP 400 errors are generally fatal, but we don't know that for sure and Azure has a lot of services which return
+		// 400 even when the error is transient, so we treat them all as retryable (at very slow speed) for now.
+		return core.ErrorRetryable, retry.VerySlow
 	}
 
-	return core.ErrorRetryable
+	return core.ErrorRetryable, retry.None
```
theunrepentantgeek: Ditto here.

v2/internal/reconcilers/arm/errorclassification/error_classifier.go
theunrepentantgeek: Should we have comments on these to indicate the circumstances when a retry would work so that future maintainers know what's in our heads?

e.g.


```suggestion
	case "AnotherOperationInProgress",  // Can retry once that operation has completed
```

v2/internal/reconcilers/arm/errorclassification/error_classifier.go
```diff
@@ -66,54 +66,59 @@ func classifyCloudError(err *genericarmclient.CloudError) core.CloudErrorDetails
...
 		"PrivateIPAddressNotInSubnet",
-		"PropertyChangeNotAllowed",
-		"RegionIsOfferRestricted",
-		"RequestDisallowedByPolicy", // TODO: Technically could probably retry through this?
-		"ReservedResourceName",
-		"SkuNotAvailable",
-		"SubscriptionNotFound":
+		"PropertyChangeNotAllowed":
 		result.Classification = core.ErrorFatal
 	default:
```
theunrepentantgeek, line 105: Given the size of these blocks, would it make sense to separate them with a whitespace line so it's easier to visually scan to find the actual action?

theunrepentantgeek (approved):
Overall looks good, just a couple minor questions/suggestions.

v2/cmd/controller/app/setup.go
```diff
@@ -522,7 +522,7 @@ func makeControllerOptions(cfg config.Values) generic.Options {
 	}
 
 	// If sync period isn't set, set verySlow delay at 24h, otherwise set it
-	// to the sync period.
+	// to the sync period (whose default is 1h)
```
matthchr: Do we want a separate retry interval configurable for this in the settings secret?

v2/internal/reconcilers/arm/errorclassification/error_classifier.go
matthchr: Added

v2/internal/reconcilers/arm/errorclassification/error_classifier.go
matthchr: You mean the case statements basically? I don't personally find that making much of a difference in scan-ability, but happy to add it if it helps you/others.

matthchr: Added

v2/internal/reconcilers/arm/errorclassification/error_classifier.go
matthchr: Yeah we do this elsewhere too. Since this is "nonsensical" we actually default to `Slow` if `retry.None` is set and it's retryable. See `v2/pkg/genruntime/conditions/ready_condition_builder.go` `MakeReason`.

I've fixed this just to be clearer.

v2/internal/reconcilers/arm/errorclassification/error_classifier.go
matthchr: Fixed

v2/cmd/controller/app/setup.go
matthchr: Discussed with @theunrepentantgeek and we can add one later if users ask for it, but for now figure just using the default sync interval makes sense.

theunrepentantgeek (approved):

PR #5099, opened 2026-01-11
Fix subtle startup bug

theunrepentantgeek:
## What this PR does

An update to golangci-lint caught a subtle bug in our startup that would have resulted in a panic.

```
[controller:lint] v2/cmd/controller/app/setup.go:218:18: call function with a nil value error after check error (nilnesserr)
[controller:lint] 			setupLog.Error(err, "failed to register webhook for gvks")
```

gif


theunrepentantgeek:
Cyclic dependency with PR #5088, going to merge this oneline fix into that.

PR #5115, opened 2026-01-19
Improve our test summary report

theunrepentantgeek:
## What this PR does

Reduces the memory consumption, as it appears we've begun OOMing near the end of test runs. This is likely triggered by the increasing number of tests.

Memory overhead has been reduced by:

* Read the input file line by line instead of all at once
* Process lines (instances of `JSONFormat`) as they are parsed, instead of accumulating a full list
* Use `[]byte` instead of `string` when parsing JSON

Crude benchmarking indicates our allocations have reduced from 640MiB to 227MiB (65% saving). Additionally, less memory will remain "live", which should permit running in more constrained environments.

Also changes to the format of the `Package Summary` section to use a table, and to remove the common prefix shared by all packages in the test run. Both increase readability and decreases the size of the file.

### Special notes

There's a 1MB maximum size for the _test summary_ that GitHub will display on the results page for a workflow action. 
We sometimes breech that limit when things don't go well, though in the successful case the file is only ~50KB in size.

Groot

### DIFF
--- a/v2/tools/mangle-test-json/test_run_factory.go
+++ b/v2/tools/mangle-test-json/test_run_factory.go
@@ -5,6 +5,8 @@
 
 package main
 
+import "unique"
+
 type testRunFactory struct {
 	// testRuns contains all tests, keyed by package then test name
 	testRuns map[string]map[string]*TestRun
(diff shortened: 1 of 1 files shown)


v2/tools/mangle-test-json/test_run.go
```diff
@@ -11,9 +11,42 @@ import (
 	"time"
 )
 
+type TestAction byte
+
+const (
+	Running    TestAction = 1 << iota // Test is running
```
matthchr: minor: The comments here are probably not necessary

v2/tools/mangle-test-json/test_run_factory.go
```diff
@@ -0,0 +1,46 @@
...
+
+func newTestRunFactory() *testRunFactory {
+	return &testRunFactory{
+		runs: make(map[string]*TestRun),
+	}
+}
+
+func (f *testRunFactory) apply(action JSONFormat) {
+	// Find (or create) the test run for this item of data
+	testrun, found := f.runs[action.key()]
```
matthchr:
```suggestion
	testRun, found := f.runs[action.key()]
```

v2/tools/mangle-test-json/test_run.go
theunrepentantgeek: Ended up removing the enumeration entirely - was nervous about the **panic** in case of an unexpected value, really don't want to end up without a test summary after a CI run.

Instead, using the `unique` package to ensure we don't have duplicate values on the heap.

v2/tools/mangle-test-json/test_run_factory.go
theunrepentantgeek: Fixed.

PR #5133, opened 2026-01-26
Disable RoleDefinition cache during tests

theunrepentantgeek:
## What this PR does

The cached used for built-in RoleDefinitions is used to resolve well-known names like `Contributor` and `Owner` to the required well-known GUIDs.

By design, this cache is populated once and then reused for the lifetime of the ASO pod, as this collection is very slowly changing. 

However, this can make our testing non-deterministic as whichever test first references a `RoleDefinition` by its well known name incurs the overhead of populating the cache - and if the recording for that test doesn't include the required HTTP request/response, the test will fail.

This PR adds the ability to disable caching, and uses that where required to ensure our tests are deterministic.

gif

## Checklist

- [x] this PR contains tests

### DIFF
--- a/v2/internal/controllers/samples_test.go
+++ b/v2/internal/controllers/samples_test.go
@@ -20,7 +20,6 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	"github.com/Azure/azure-service-operator/v2/api/authorization/customizations"
 	"github.com/Azure/azure-service-operator/v2/internal/reflecthelpers"
 	"github.com/Azure/azure-service-operator/v2/internal/set"
 	"github.com/Azure/azure-service-operator/v2/internal/testcommon"
@@ -53,9 +52,6 @@ func Test_Samples_CreationAndDeletion(t *testing.T) {
 	}
 
 	g := NewGomegaWithT(t)
-
-	// Disable caching of built-in role definitions to ensure test isolation
-	customizations.DisableBuiltInRoleDefinitionsCaching()
 
 	regex, err := regexp.Compile("^v(1api)?[a-z0-9]*$")
 	g.Expect(err).To(BeNil())
--- a/v2/internal/controllers/suite_test.go
+++ b/v2/internal/controllers/suite_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/onsi/gomega"
 	"github.com/onsi/gomega/format"
 
+	authcustomizations "github.com/Azure/azure-service-operator/v2/api/authorization/customizations"
 	"github.com/Azure/azure-service-operator/v2/internal/testcommon"
 )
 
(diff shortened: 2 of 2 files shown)


v2/api/authorization/customizations/role_assignment_extensions.go
```diff
@@ -108,77 +103,112 @@ func (extension *RoleAssignmentExtension) ModifyARMResource(
...
+) (map[string]string, error) {
 	builtInRoleDefinitionsLock.Lock()
 	defer builtInRoleDefinitionsLock.Unlock()
 
 	// Short circuit if already loaded
-	if len(builtInRoleDefinitions) > 0 {
-		return nil
+	if len(builtInRoleDefinitionsCache) > 0 {
+		return builtInRoleDefinitionsCache, nil
 	}
```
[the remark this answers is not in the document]
theunrepentantgeek: Forcing definitions to load every time is the point of the change. #wontfix

v2/api/authorization/customizations/role_assignment_extensions.go
```diff
@@ -108,77 +103,112 @@ func (extension *RoleAssignmentExtension) ModifyARMResource(
...
-// We load them once, and then keep them for the lifetime of the pod.
-// Given the list of built-in role definitions changes very slowly, this seems reasonable.
-func ensureBuiltInRoleDefinitionsLoaded(
+// DisableBuiltInRoleDefinitionsCaching disables caching of built-in role definitions.
+// This is intended for use in tests only.
+func DisableBuiltInRoleDefinitionsCaching() {
+	builtInRoleDefinitionsLock.Lock()
+	defer builtInRoleDefinitionsLock.Unlock()
+
+	builtInRoleDefinitionsCachingDisabled = true
```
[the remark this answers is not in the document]
theunrepentantgeek: Agreed.

v2/internal/controllers/authorization_roleassignment_20220401_test.go
```diff
@@ -25,6 +26,9 @@ import (
 func Test_Authorization_RoleAssignment_OnResourceGroup_CRUD(t *testing.T) {
 	t.Parallel()
 
+	// Disable caching of built-in role definitions to ensure test isolation
+	customizations.DisableBuiltInRoleDefinitionsCaching()
```
matthchr, line 30: We have common/shared test infrastructure which runs via `tc := globalTestContext.ForTest(t)` - this sets up envtest and that sort of stuff.

While it's a bit awkward to have an `authorization/customizations` dependency there, wondering if it'd be better to just universally disable the cache across all tests so that new users don't miss adding this magical line?

Otherwise we're still racing between tests that remember to set this and tests that don't.

theunrepentantgeek: I've done this, and cleaned up `samples_test.go`, but left the disable in place here specifically for visibility.

matthchr: Happy with that.

PR #5140, opened 2026-01-29
Improve PreReconcileOwner check

matthchr:
## What this PR does

Solves issue where the owner has stale state from Azure.

* Removes owner from PreReconcileCheck. Owner checks must now always be
  done via PreReconcileOwnerCheck. This ensures that we only issue HTTP
  calls to Azure to refresh the owner state when we actually need the
  owner for something.
* Resources that implement PreReconcileOwnerCheck now issue a GET on the
  owner ARM resource before calling the check, meaning that the check
  will always be acting on an up-to-date owner status + spec.
* Remove possibility of owner being nil in PreReconcileOwnerCheck. We
  won't call that API at all if the owner is nil, which saves the
  individual check implementations from needing to always include a nil
  check.

Test updates:

* Updated Kuberentes version in tests and samples as the Kuberentes
  version we were using is now out of support.
* Updated containerservice samples to use a different KeyVault name for
  each sample so that the KeyVault names don't collide when being
  recovered/purged.
* Removed Test_Samples_CreationAndDeletion/Test_Dbforpostgresql_v20250801_CreationAndDeletion
  from the list of skipped samples.
* Re-recorded required samples (containerservice and dbforpostgresql)

Fixes #5040.

## Checklist

- [x] this PR contains tests

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -10,12 +10,9 @@ ARG TARGETARCH
 # APT dependencies
 # including docker-cli for docker-from-docker pattern
 # and az-cli for interacting with azure resources
-# NOTE: The yarn gpg key expired and was rolled to a new key, see https://github.com/yarnpkg/yarn/issues/9218 and https://github.com/yarnpkg/yarn/issues/9216.
-# We don't actually install or use yarn but the fact that the key has issues causes problems when apt-get update is run so we just removed yarn from apt sources.
 ENV DEBIAN_FRONTEND=noninteractive
 RUN curl -fsSL https://download.docker.com/linux/debian/gpg | gpg --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg \
     && echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/docker-archive-keyring.gpg] https://download.docker.com/linux/debian $(lsb_release -cs) stable" > /etc/apt/sources.list.d/docker.list \
-    && rm /etc/apt/sources.list.d/yarn.list \
     && apt-get update \
     && apt-get -y install --no-install-recommends bash-completion lsb-release graphviz zip nodejs npm python3-pip docker-ce-cli docker-compose-plugin\
     # install az-cli


v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -650,38 +681,76 @@ func (r *azureDeploymentReconcilerInstance) getStatus(ctx context.Context, id st
...
+	// Cast to ARMMetaObject
+	metaObj, ok := obj.(genruntime.ARMMetaObject)
+	if !ok {
+		return nil, eris.Errorf("object of type %T does not implement genruntime.ARMMetaObject", obj)
+	}
+
+	// Set the resource ID annotation so updateStatus can find it
+	genruntime.SetResourceID(metaObj, armID)
+	// Set the spec.OriginalVersion to the latest version
+	reflecthelpers.SetProperty(metaObj.GetSpec(), "OriginalVersion", strings.TrimSuffix(gvk.Version, "storage")) // This is real hacky
```
matthchr: I do not love this. Doing this is required to be able to re-use the `updateStatus` method, which is very nice... but possibly we could refactor it to make this unneeded?

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -358,30 +360,59 @@ func (r *azureDeploymentReconcilerInstance) preReconciliationCheck(
...
+			// Note that this update is not committed back to api-server currently, so if we come back around here
+			// and run through this extension again, we will re-fetch the owner status from Azure again.
+			err := r.updateStatus(ctx, ownerDetails.Owner)
+			if err != nil {
+				return extensions.PreReconcileCheckResult{}, err
+			}
+			ownerObj = ownerDetails.Owner
+		} else if ownerDetails.Result == resolver.OwnerFoundARM {
+			// Owner is an ARM ID - create a temporary object and fetch status
+			ownerGroup, ownerKind := genruntime.LookupOwnerGroupKind(r.Obj.GetSpec())
```
theunrepentantgeek, line 380: Might be worth adding a comment here to the effect that this will always be the _hub_ (aka storage) version of this resource, not necessarily the same version as the child resource.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -576,19 +607,19 @@ func (r *azureDeploymentReconcilerInstance) resultBasedOnGenerationCount() ctrl.
 
 var zeroDuration time.Duration = 0
 
-func (r *azureDeploymentReconcilerInstance) getStatus(ctx context.Context, id string) (genruntime.ConvertibleStatus, time.Duration, error) { //nolint:unparam
-	armStatus, err := genruntime.NewEmptyARMStatus(r.Obj, r.ResourceResolver.Scheme())
+func (r *azureDeploymentReconcilerInstance) getStatus(ctx context.Context, obj genruntime.ARMMetaObject, id string) (genruntime.ConvertibleStatus, time.Duration, error) { //nolint:unparam
```
theunrepentantgeek: nit: for long argument lists like this, I like to format one per line for clarity


```suggestion
func (r *azureDeploymentReconcilerInstance) getStatus(
    ctx context.Context, 
    obj genruntime.ARMMetaObject, 
    id string,
) (genruntime.ConvertibleStatus, time.Duration, error) { //nolint:unparam
```

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -650,38 +681,76 @@ func (r *azureDeploymentReconcilerInstance) getStatus(ctx context.Context, id st
 	return status, zeroDuration, nil
 }
 
-func (r *azureDeploymentReconcilerInstance) setStatus(status genruntime.ConvertibleStatus) error {
+// TODO: Move
```
theunrepentantgeek: Now, or later?

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
theunrepentantgeek: What are you thinking? 

It occurs to me that updateStatus could look up the version if `OriginalVersion` is blank, so it always does the right thing.

Something for a followup fix though, don't want to delay this.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
matthchr: Done

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
matthchr: Yep agreed - fixed.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
matthchr: oops - this was from a previous take on the refactor. I've removed it.

jakjang (approved):

PR #5155, opened 2026-02-04
Remove usage of NewDefaultAzureCredential

matthchr:
Replaced with NewChainedTokenCredential and a more limited set of credentials.

For the ASO controller, this shouldn't have any impact on actual users, as we're still supporting the same authentication options that we supported before. The options that were removed were:
* NewWorkloadIdentityCredential - but we already take care of this up above via cfg.UseWorkloadIdentityAuth, and the workload identity option via DefaultCredential never worked because we don't set the AZURE_FEDERATED_TOKEN_FILE env variable.
* NewAzureCLICredential - which is obviously not recommended for running in production due to the need to constantly refresh the credentials, and we never supported as there is no way to mount the CLI credentials file into the pod.
* NewAzureDeveloperCLICredential - Same as above.
* NewAzurePowerShellCredential - Same as above.

For asoctl, there is a slight change in that we now do not support:
* NewAzureDeveloperCLICredential
* NewAzurePowerShellCredential

The correct way is to use Azure CLI credentials.

## Checklist

- [ ] this PR contains tests

### DIFF
--- a/v2/cmd/asoctl/cmd/import_azure_resource.go
+++ b/v2/cmd/asoctl/cmd/import_azure_resource.go
@@ -13,6 +13,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/go-logr/logr"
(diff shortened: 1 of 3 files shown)


v2/cmd/asoctl/cmd/import_azure_resource.go
```diff
@@ -228,19 +229,75 @@ func importAzureResource(
...
+	}
+
+	cliCred, err := azidentity.NewAzureCLICredential(nil)
+	if err != nil {
+		credErrors = append(credErrors, fmt.Sprintf("AzureCLICredential: %s", err.Error()))
+	} else {
+		creds = append(creds, cliCred)
+	}
+
+	if len(creds) == 0 {
```
theunrepentantgeek, line 297: Perhaps add a comment here to make it clear that it's intentional that errors are only returned if NO credentials are available - in the success case, we suppress errors for unused credential types.

v2/cmd/controller/app/setup.go
```diff
@@ -357,18 +357,52 @@ func getDefaultAzureTokenCredential(cfg config.Values, setupLog logr.Logger) (az
...
+		miCredOptions.ID = azidentity.ClientID(cfg.ClientID)
+	}
+	miCred, err := azidentity.NewManagedIdentityCredential(miCredOptions)
+	if err != nil {
+		setupLog.Error(err, "ManagedIdentityCredential not available")
+	} else {
+		creds = append(creds, miCred)
+	}
+
+	if len(creds) == 0 {
```
theunrepentantgeek, line 399: Ditto here, same motivation.

jakjang (approved):
LGTM!

PR #5162, opened 2026-02-09
Move testing of samples to separate workflow step

theunrepentantgeek:
## What this PR does

Moves testing of samples to a separate step in our CI workflow so it runs in parallel with other time-consuming tests.

This may also reduce the memory footprint of our controller tests, which have been occasionally OOMing recently.

gif

## Checklist

- [x] this PR contains tests


v2/internal/testsamples/samples_test.go
```diff
@@ -3,7 +3,7 @@ Copyright (c) Microsoft Corporation.
 Licensed under the MIT license.
 */
```
matthchr, line 5: Since these are all tests we could call the path `v2/internal/samplestest` or something?

v2/internal/samples/samples_test.go
```diff
@@ -3,7 +3,7 @@ Copyright (c) Microsoft Corporation.
 Licensed under the MIT license.
 */
 
-package controllers_test
+package samples
```
matthchr:
```suggestion
package samples_test
```
maybe?

v2/internal/testsamples/samples_test.go
```diff
@@ -47,10 +47,6 @@ var randomNameExclusions = []string{
 func Test_Samples_CreationAndDeletion(t *testing.T) {
 	t.Parallel()
 
-	if *isLive {
```
matthchr, line 50: Reason it's OK to remove this is that the live tests only run the main e2e tests?

Taskfile.yml
```diff
@@ -578,6 +578,26 @@ tasks:
...
+          task: produce-markdown-summary
+          vars:
+            INPUT_FILE: '{{.TEST_OUT}}/samples-test.json'
+            OUTPUT_FILE: '{{.TEST_OUT}}/samples-test.md'
+      - go test -count=1 -timeout {{.TIMEOUT}} -run '{{default ".*" .TEST_FILTER}}' {{.VERBOSE}} -json ./internal/samples > '{{.TEST_OUT}}/samples-test.json'
+    vars:
+      VERBOSE:
+        sh: if [ $TEST_FILTER ];  then echo "-v"; fi
+      TIMEOUT: '{{default "25m" .TIMEOUT}}'  # Should be kept in sync with the tests below
+#    sources:
```
matthchr: Commented out?

Remove or uncomment?

theunrepentantgeek: Restored - will make thing run faster locally when you're _not_ working on samples.

v2/internal/testsamples/samples_test.go
theunrepentantgeek: Correct, this was never used for testing samples.

v2/internal/testsamples/samples_test.go
theunrepentantgeek: Renamed the package to `testsamples`

v2/internal/samples/samples_test.go
theunrepentantgeek: Since there's nothing but tests in the package, using the special suffix `_test` to isolate the tests seems unnecessary; I've renamed the package simply to `testsamples`.

matthchr (approved):

PR #5165, opened 2026-02-10
Fix devcontainer image caching

theunrepentantgeek:
## What this PR does

The devcontainer image wasn't being cached properly in PR validation workflows, causing [install-dependencies.sh](vscode-file://vscode-app/c:/Users/bearps/AppData/Local/Programs/Microsoft%20VS%20Code/bdd88df003/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to run every time (~3+ minutes).

Problem: The "Build Devcontainer image" workflow uses docker/build-push-action@v6, which defaults to Buildx. It was configured with build-args: BUILDKIT_INLINE_CACHE=1, which is the legacy Docker approach for embedding cache metadata. Buildx ignores this build-arg and requires cache-to: type=inline instead. Without proper cache metadata, the PR validation workflows couldn't match layers when using --cache-from.

Fix: Replace build-args: BUILDKIT_INLINE_CACHE=1 with cache-to: type=inline so Buildx correctly embeds the cache metadata into the pushed image.

gif

### DIFF
--- a/.github/workflows/build-devcontainer-image.yml
+++ b/.github/workflows/build-devcontainer-image.yml
@@ -43,8 +43,8 @@ jobs:
         with:
           context: .devcontainer
           push: true
-          # Build an image usable as cache-from, per: https://docs.docker.com/engine/reference/commandline/build/#specifying-external-cache-sources
-          build-args: BUILDKIT_INLINE_CACHE=1
+          # Embed inline cache metadata so consumers can use --cache-from
+          cache-to: type=inline
           tags: |
             docker.pkg.github.com/azure/azure-service-operator/aso-devcontainer:latest
             docker.pkg.github.com/azure/azure-service-operator/aso-devcontainer:${{ github.sha }}


.github/workflows/build-devcontainer-image.yml
```diff
@@ -43,8 +43,8 @@ jobs:
         with:
           context: .devcontainer
           push: true
-          # Build an image usable as cache-from, per: https://docs.docker.com/engine/reference/commandline/build/#specifying-external-cache-sources
-          build-args: BUILDKIT_INLINE_CACHE=1
+          # Embed inline cache metadata so consumers can use --cache-from
+          cache-to: type=inline
```
matthchr, line 47: Are we expecting to see improvement in this PR, or only in subsequent PRs? I chceked this PR and it still seems to have taken 4m

theunrepentantgeek: Only in later ones - the change is to the _build devcontainer_ workflow that's only run periodically, not for every PR.

PR #5181, opened 2026-02-16
Move testing of controllers to separate workflow step

theunrepentantgeek:
## What this PR does

Moves testing of `controllers` to a separate step in our CI workflow so it runs in parallel with other time-consuming tests.

Followup to #5162 

### Special notes

I've rewired tasks so that the `controllers` tests run when using `controllers:test-integration-envtest` locally, but not from CI.,

gif

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -618,6 +618,30 @@ tasks:
     sources:
       - '{{joinPath .CONTROLLER_ROOT "samples"}}'
 
+  controller:test-genruntime:
+    desc: Run integration tests with envtest using record/replay.
+    dir: "{{.CONTROLLER_ROOT}}"
+    deps: 
+      - controller:run-kustomize-for-envtest
+    cmds:
+      - go test -count=1 -timeout {{.TIMEOUT}} -run '{{default ".*" .TEST_FILTER}}' {{.VERBOSE}} ./pkg/genruntime/test
+    vars:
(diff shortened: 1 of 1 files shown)


Taskfile.yml
```diff
@@ -598,6 +618,30 @@ tasks:
...
+    deps: 
+      - controller:run-kustomize-for-envtest
+    cmds:
+      - go test -count=1 -timeout {{.TIMEOUT}} -run '{{default ".*" .TEST_FILTER}}' {{.VERBOSE}} ./internal/genericarmclient
+    vars:
+      VERBOSE:
+        sh: if [ $TEST_FILTER ];  then echo "-v"; fi
+      TIMEOUT: '{{default "25m" .TIMEOUT}}'  # Should be kept in sync with the -cover variant below
+
   controller:test-integration-envtest:
```
matthchr, line 645: Maybe update the description here to make sure that it's clear this is for local testing of everything together, and that the more focused targets are used for CI to allow splitting things up?

Taskfile.yml
matthchr, line 767: This target is probably poorly named now, given that many other integration tests are also for `ci` but aren't named as such?

Or maybe the name is OK and we should just fix the `desc`? Ideas for names are `controller:test-integration-dependencies` or `controller:test-integration-core`, or similar...?

PR #5214, opened 2026-03-03
Add tasks to tidy recordings

theunrepentantgeek:
## What this PR does

Adds new tasks to `Taskfile.yaml` allowing us to "tidy" test recordings, optimizing them so that tests run faster.

Mostly serve as an interim step until we integrate go-vcr-tidy into our recording process.

gif

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -618,11 +618,15 @@ tasks:
     sources:
       - '{{joinPath .CONTROLLER_ROOT "samples"}}'
 
+  # Once we've finished integrating go-vcr-tidy into our test framework, we won't need these tasks anymore as the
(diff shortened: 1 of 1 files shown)


Taskfile.yml
```diff
@@ -618,6 +618,17 @@ tasks:
     sources:
       - '{{joinPath .CONTROLLER_ROOT "samples"}}'
 
+  controller:tidy-recordings:
```
matthchr, line 623: minor, maybe leave a TODO comment here that these can be removed when we update the recordings to use the library version of tidy?

theunrepentantgeek: Good idea. Done.

PR #5218, opened 2026-03-07
split webhooks into multiple to avoid hitting max size

### LINKED ISSUE #5208: Bug: Generated webhook configuration files too large to store in ETCD
## Describe the bug

When installing the latest [v2.18.0](https://github.com/Azure/azure-service-operator/releases/tag/v2.18.0) version of the helm chart we encountered an issue where we could not apply changes to the generated webhook configuration yaml files that are included in the helm-chart, due to them being too large to store in ETCD.

Error message: `etcdserver: request is too large`.

When investigating we found that this happened in all of our persistent clusters where cert-manager had performed a certificate rotation at some point so that the `clientConfig.caBundle` field of a webhook configuration contained the two latest versions of the ca, which increases the total size of the object. 
In clusters where no certificate rotation had been performed yet the upgrade went fine due to the smaller size.

Files:
- `admissionregistration.k8s.io_v1_validatingwebhookconfiguration_azureserviceoperator-validating-webhook-configuration.yaml`
- `admissionregistration.k8s.io_v1_mutatingwebhookconfiguration_azureserviceoperator-mutating-webhook-configuration.yaml`

We use ArgoCD to deploy changes using Kubernetes server side apply already to try and bring the size of the object down.

Azure Service Operator Version:  mcr.microsoft.com/k8s/azureserviceoperator:v2.18.0/k8s/azureserviceoperator:v2.18.0
Kubernetes Version: 1.33.6

**Temporary Remediation**: 

Deleting the webhook configuration files from the cluster and applying them again(letting Argo CD sync them) works because then only the latest ca.crt is injected as part of the caBundle.

## Expected behavior

I would expect the webhook configurations to respect the `crdPattern` field of the helm chart and only create webhook configurations for the crd resources actually created and used in the cluster. With the current setup where webhook configurations are created for every resource regardless I would still assume this to work seamlessly, perhaps by splitting the objects into several.

## To Reproduce

I have not tried to reproduce this in a fresh cluster but this is roughly the chain of events that we have identified as likely.

1. Install Azure Service Operator v2.17.0 using the helm chart.
2. Wait for the caBundle to be injected into the webhook configurations by cert-manager.
3. Perform a certificate rotation so that the two latest ca.crt are injected as part of the caBundle by cert-manager.
4. Attempt to upgrade to v2.18.0.

## Additional context

We have a way of getting around the issue for now and we could probably solve it ourselves using the Kustomize layer or similar but an official fix or recommendation for workaround would be appreciated.

matthchr:
Fixes #5208

## Checklist

- [ ] this PR contains tests


scripts/v2/generate-webhooks-per-group.py
```diff
@@ -0,0 +1,157 @@
...
+            group_tmp = tmp_path / group
+            group_tmp.mkdir()
+            group_tmp_dirs[group] = group_tmp
+
+        # Phase 2: run controller-gen concurrently for all groups
+        print(f"  Running controller-gen for {len(groups)} groups concurrently...")
+        with ThreadPoolExecutor() as executor:
+            futures = {
+                executor.submit(run_controller_gen, api_dir, group, group_tmp_dirs[group]): group
+                for group in groups
```
theunrepentantgeek, line 135: How can the **for** _follow_ the body? I figure this is valid because it runs ... but #perplexed

matthchr: This is a Python dict comprehension: 
The general syntax for them is:
`<result_key>: <result_value> for <var> in <thing>`

So for example `{x: x**2 for x in (2, 4, 6)}` would iterate over the dictionary and return a resulting dict: `{2:4, 4:16, 6:36}`

In this case, the keys are the future and the value is the group it is processing.

PR #5220, opened 2026-03-09
Address gosec alerts in mangle-test-json

theunrepentantgeek:
## What this PR does

The `gosec` linter started flagging code in `mangle-test-json` as unstafe.

In one case the alert was real, so I applied `os.OpenRoot()` to prevent abuse; in the other cases, the output is already escaped, and the alert could be suppressed.

gif


theunrepentantgeek:
@copilot Please merge `main` into this branch.

PR #5224, opened 2026-03-09
Fix reconciliation blocking

theunrepentantgeek:
## What this PR does

I noticed during work on Cassandra resources that an implementation of PreReconcileOwnerCheck which blocked reconciliation was having no effect.

This fixes the bug.

gif


matthchr (approved):

matthchr:
Looks like tests are still failing after re-record - do you know if it's a real issue or a flake?

Also curious why only the one API version needed to be re-recorded, since there are other API versions of that same resource that seemingly work?

theunrepentantgeek:
> Looks like tests are still failing after re-record - do you know if it's a real issue or a flake?

As far as I can tell, it's a flake.

>Also curious why only the one API version needed to be re-recorded, since there are other API versions of that same resource that seemingly work?

I'm not sure on this. There are multiple resources implementing `PreReconcileOwnerCheck` and I think that's the source of the problem - all the extra GETs for the parent resource causing issues. But I'm not 100% on that. 

I've got things to pass - but only by refactoring the test. Many of the subtests were just creating child resources, so I pulled those into the main test so all the resources get created together.

v2/internal/controllers/dbforpostgresql_flexibleserver_crud_v1api20240801_test.go
```diff
@@ -79,57 +79,103 @@ func Test_DBForPostgreSQL_FlexibleServer_20240801_CRUD(t *testing.T) {
...
-			Test: func(tc *testcommon.KubePerTestContext) {
-				FlexibleServer_Configuration_20240801_CRUD(tc, flexibleServer)
-			},
-		},
-		testcommon.Subtest{
-			Name: "Flexible servers backup CRUD",
-			Test: func(tc *testcommon.KubePerTestContext) {
-				FlexibleServer_Backup_20240801_CRUD(tc, flexibleServer)
-			},
-		},
```
theunrepentantgeek, line 132: Theses subtests were just testing whether resource creation should work; inlining these better matches what will happen when a YAML file is applied to the cluster, plus it appears to make the test more reliable.

v2/internal/testcommon/vcr/v3/replay_roundtripper.go
```diff
@@ -249,6 +249,15 @@ func (replayer *replayRoundTripper) resourceStateFromBody(response *http.Respons
 	return "", false
 }
 
+// isTerminalHTTPStatus returns true if the specified HTTP status code represents a terminal state for a resource.
+// "Not existing" is pretty terminal.
+func (*replayRoundTripper) isTerminalHTTPStatus(status int) bool {
+	return status == http.StatusNotFound ||
+		status == http.StatusGone ||
+		status == http.StatusCreated ||
+		status == http.StatusOK
```
matthchr, line 258: minor: Document why status OK is terminal? (I _think_ you mean that the "operation" is done at that point in a way that something like Accepted is not, but not 100% clear.

PR #5231, opened 2026-03-12
Improve performance of CI

theunrepentantgeek:
## What this PR does

Eliminates some duplication of work by running `test-samples` and `test-controllers` directly, without regenerating any code. 

### Special notes

I've split the `ci` functions out into a nested Taskfile so they're not "on the beaten path". 

gif

### DIFF
--- a/ci/Taskfile.yml
+++ b/ci/Taskfile.yml
@@ -15,7 +15,7 @@ tasks:
     dir: "{{.CONTROLLER_ROOT}}"
     deps: 
       - setup
-      - generate-crds
+      - run-kustomize-for-envtest
     cmds:
       - defer: # want to run even on failure
           task: produce-markdown-summary
@@ -38,7 +38,7 @@ tasks:
     dir: "{{.CONTROLLER_ROOT}}"
     deps: 
       - setup
-      - generate-crds
+      - run-kustomize-for-envtest
     cmds:
       - defer: # want to run even on failure
           task: produce-markdown-summary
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-validation.yml
```diff
@@ -200,18 +200,17 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
         if: steps.check-changes.outputs.code-changed == 'true'
 
-      # Note: Changes to this step must also be mirrored into pr-validation-with-secrets.yaml
-      - name: Build devcontainer image
+      - name: Pull devcontainer image
         run: |
-          docker build --cache-from ghcr.io/azure/azure-service-operator/aso-devcontainer:latest --tag devcontainer:latest .devcontainer
+          docker pull ghcr.io/azure/azure-service-operator/aso-devcontainer:latest
```
matthchr, line 205: Doesn't this assume that the container image exists?

Oh wait, it always should because we have another periodic job that builds + caches it. Ok fair enough.

ci/Taskfile.yml
```diff
@@ -0,0 +1,74 @@
...
+      TEST_OUTPUT: '{{joinPath .TEST_OUT "test-samples.json"}}'
+      TEST_REPORT: '{{joinPath .TEST_OUT "test-samples.md"}}'
+
+  setup:
+    internal: true
+    cmds:
+      - echo "Test output will be stored in {{.TEST_OUT}}"
+      - mkdir -p {{.TEST_OUT}}
+
+  build-mangle-test:
```
matthchr, line 127: Should we delete the targets from the other Taskfile that are duplicates of these?

Taskfile.yml
```diff
@@ -27,8 +30,7 @@ vars:
   HEADER_FILE:
     sh: 'realpath v2/boilerplate.go.txt'
```
matthchr, line 32: minor: Maybe we should update `controller:test-integration-envtest` (and friends?) to point out that they are for local runs only now?

Taskfile.yml
```diff
@@ -3,6 +3,9 @@ version: '3.7'
 run: when_changed
 output: prefixed
 
+includes:
+  ci: ./ci/Taskfile.yml
+
```
matthchr, line 8: Maybe for a separate PR but:
`controller:test-integration-kind-ci` probably could move too? Though it uses a bunch of shared kind infrastructure so maybe not easily...

matthchr (approved):
Overall LGTM but wondering if we need to clean up the existing Taskfile some?

.github/workflows/pr-validation.yml
theunrepentantgeek: It's possible we might hit a problem if the image is stale, but that will be easy to address - and should be rare.

ci/Taskfile.yml
theunrepentantgeek: They're still needed from both locations, so for the moment duplication is necessary. I think we can factor things out - but I want to make small cautious steps.

Taskfile.yml
theunrepentantgeek: Yeah, that's why it didn't move here.

theunrepentantgeek:
> wondering if we need to clean up the existing Taskfile some?

Most certainly! I want to take small steps though.

Taskfile.yml
theunrepentantgeek: Good idea. I'll do that in my next Taskfile PR.

ci/Taskfile.yml
matthchr: Ok, as long as you're tracking that work =)

PR #5232, opened 2026-03-12
feat: Add support for Entra Application

### LINKED ISSUE #5217: Feature: Add support for Entra Applications
# Describe the feature request.

As requested in #2474, add support for Entra (Azure AD) Applications to ASO.

Since this is not an ARM resource, will require a hand coded implementation, but #4768 added most of the foundation required to make this straightforward.

theunrepentantgeek:
## What this PR does

Adds support for Entra Application as a new resource type. 

gif

Fixes #5217


v2/internal/reconcilers/entra/entra_application_reconciler.go
```diff
@@ -0,0 +1,494 @@
...
+			return ctrl.Result{
+				Requeue: true,
+			}, nil
+		}
+
+		return ctrl.Result{}, eris.Wrapf(err, "getting application by ID %s", id)
+	}
+
+	// Update - PATCH
+	app.Spec.AssignToApplication(a)
```
theunrepentantgeek, line 192: 🔴 **Critical: Nil pointer panic when application is deleted externally**

`loadApplicationByID` (line 350-354) returns `(nil, nil)` for 404 errors — it swallows the OData 404 and returns `nil, nil`. But here in `update()`, the `isNotFound(err)` check on line 177 is **dead code** because `err` is always `nil` for a 404.

The nil `Applicationable` reaches this line and panics:
```go
app.Spec.AssignToApplication(a)  // a is nil → panic
```

**How to trigger**: Delete the application in Azure portal while the K8s resource exists. Next reconcile crashes the controller.

Same bug exists in `UpdateStatus()` at line 332, and the identical pre-existing bug exists in the SecurityGroup reconciler (lines 183, 349).

**Suggested fix**: Either:
1. Change `loadApplicationByID` to return a sentinel error for 404 so the `isNotFound` check works, OR
2. Add a nil guard: `if a == nil { setEntraID(app, ""); return ctrl.Result{Requeue: true}, nil }`

_All three reviewers (Advocate, Skeptic, Architect) independently identified this bug._

v2/internal/reconcilers/entra/entra_application_reconciler.go
```diff
@@ -0,0 +1,494 @@
...
+}
+
+// loadApplicationsByDisplayName loads applications from Entra by display name.
+func (r *EntraApplicationReconciler) loadApplicationsByDisplayName(
+	ctx context.Context,
+	displayName string,
+	client *msgraphsdkgo.GraphServiceClient,
+) ([]msgraphmodels.Applicationable, error) {
+	// Try to get the application by display name
+	filterStr := fmt.Sprintf("displayName eq '%s'", displayName)
```
theunrepentantgeek: 🟠 **High: OData filter injection via unescaped display names**

```go
filterStr := fmt.Sprintf("displayName eq '%s'", displayName)
```

A display name containing single quotes (e.g., `O'Brien's App`) produces malformed OData: `displayName eq 'O'Brien's App'`. This causes the Graph API to return a 400 error, surfaced to the user as a cryptic reconciliation failure. Applications with apostrophes in names cannot be adopted.

Same issue exists in the SecurityGroup reconciler (line 387).

**Suggested fix**:
```go
escaped := strings.ReplaceAll(displayName, "'", "''")
filterStr := fmt.Sprintf("displayName eq '%s'", escaped)
```

_All three reviewers flagged this._

v2/api/entra/v1/application_types.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	// SignInAudience
+	if spec.SignInAudience != nil {
+		signInAudience := string(*spec.SignInAudience)
+		model.SetSignInAudience(&signInAudience)
+	}
+
+	// IdentifierUris
+	if len(spec.IdentifierUris) > 0 {
```
theunrepentantgeek, line 116: 🟠 **High: `AssignToApplication` cannot clear optional fields**

All fields use `if != nil` / `if len > 0` guards (lines 99-152). If a user *removes* a previously-set field from the spec (e.g., clears `identifierUris`, `tags`, `web`, `spa`, or `publicClient`), the empty/nil value is never sent to the API — the field persists unchanged in Entra.

For example, this check means setting `identifierUris: []` in the K8s spec has no effect. The previous URIs remain in Entra.

**Suggested fix**: Always send the field value (including empty), or track which fields were previously set and explicitly clear them. The simplest approach is to remove the nil/empty guards and always call the setter — Graph API PATCH semantics will handle absent vs empty correctly.

FEATURE.md
```diff
@@ -0,0 +1,238 @@
+# Entra Application Support
```
theunrepentantgeek: 🟠 **High: Implementation plan document should not be committed to repo root**

This file is an implementation plan that belongs in the PR description or issue #5217, not alongside `README.md`, `CONTRIBUTING.md`, etc. It will become stale and confusing to future contributors.

**Suggested fix**: Remove this file from the commit and move its content to the PR description or the linked issue.

v2/internal/reconcilers/entra/entra_application_reconciler.go
```diff
@@ -0,0 +1,494 @@
...
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/conditions"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/configmaps"
+)
+
+// EntraApplicationReconciler reconciles an Entra application.
+type EntraApplicationReconciler struct {
+	reconcilers.ReconcilerCommon
+	ResourceResolver   *resolver.Resolver
+	CredentialProvider identity.CredentialProvider
```
theunrepentantgeek: 🟠 **High: Dead `CredentialProvider` field**

`CredentialProvider` is declared on the struct but never populated by the constructor (lines 49-65) and never referenced anywhere in the reconciler. The credential flow goes through `EntraClientFactory` → `EntraClientCache.GetConnection()` → the provider, so this field is unnecessary.

Same issue exists in `EntraSecurityGroupReconciler` (line 43). This appears to be a copy-paste artifact.

**Suggested fix**: Remove the field from both reconcilers.

v2/internal/reconcilers/entra/entra_application_reconciler.go
```diff
@@ -0,0 +1,494 @@
...
+			}, nil
+		}
+
+		return ctrl.Result{}, eris.Wrapf(err, "getting application by ID %s", id)
+	}
+
+	// Update - PATCH
+	app.Spec.AssignToApplication(a)
+
+	_, err = client.Client().Applications().ByApplicationId(id).Patch(ctx, a, nil)
```
theunrepentantgeek: 🟡 **Medium: PATCH response discarded; status assigned from pre-PATCH local model**

```go
_, err = client.Client().Applications().ByApplicationId(id).Patch(ctx, a, nil)
...
app.Status.AssignFromApplication(a)  // 'a' is the local model, not the server response
```

The PATCH return value is discarded. Status is populated from the locally-mutated model `a`, not the server's authoritative response. If the Graph API normalizes, rejects, or augments any field, status diverges from reality until the next `UpdateStatus()` cycle corrects it.

Same pattern as SecurityGroup (line 201-207).

**Suggested fix**: Use the PATCH response:
```go
patched, err := client.Client().Applications().ByApplicationId(id).Patch(ctx, a, nil)
...
app.Status.AssignFromApplication(patched)
```

v2/internal/reconcilers/entra/entra_application_reconciler.go
```diff
@@ -0,0 +1,494 @@
...
+	if err != nil {
+		// Failed to update
+		return ctrl.Result{}, eris.Wrapf(err, "failed to update application %s", id)
+	}
+
+	app.Status.AssignFromApplication(a)
+
+	return ctrl.Result{}, nil
```
theunrepentantgeek, line 224: 🟡 **Medium: `update()` does not call `saveAssociatedKubernetesResources`**

`create()` saves ConfigMaps (line 290) and `UpdateStatus()` saves them (line 334), but `update()` does not. After an update, ConfigMaps won't be refreshed until the next `UpdateStatus()` cycle.

The practical impact is small since `UpdateStatus()` runs in the same reconciliation loop shortly after. Same pattern as SecurityGroup. However, this asymmetry between create/update paths increases maintenance risk.

v2/api/entra/v1/application_types.go
```diff
@@ -0,0 +1,306 @@
...
+type ApplicationList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []Application `json:"items"`
+}
+
+type ApplicationSpec struct {
+	// DisplayName: The display name of the application.
+	// +kubebuilder:validation:Required
+	DisplayName *string `json:"displayName,omitempty"`
```
theunrepentantgeek, line 60: 🟡 **Medium: `DisplayName` validation may not enforce as expected**

```go
DisplayName *string `json:"displayName,omitempty"`
```

The `+kubebuilder:validation:Required` marker and the `*string` with `omitempty` JSON tag send mixed signals. While the kubebuilder marker should generate an OpenAPI `required` constraint enforced at admission, the `omitempty` tag means a roundtrip through JSON marshal/unmarshal could drop a nil value.

Consider adding a webhook validation for `DisplayName != nil` as defense-in-depth (the `createValidations()` and `updateValidations()` methods in the webhook currently return nil).

v2/internal/reconcilers/entra/entra_application_reconciler.go
```diff
@@ -0,0 +1,494 @@
...
+// EntraApplicationReconciler reconciles an Entra application.
+type EntraApplicationReconciler struct {
+	reconcilers.ReconcilerCommon
+	ResourceResolver   *resolver.Resolver
+	CredentialProvider identity.CredentialProvider
+	Config             config.Values
+	EntraClientFactory EntraConnectionFactory
+}
+
+var _ genruntime.Reconciler = &EntraApplicationReconciler{}
```
theunrepentantgeek, line 42: 🔵 **Low: Consider extracting shared Entra reconciler pattern**

The SecurityGroup reconciler has a TODO at its line 39:
> `// TODO: Factor out common code shared with other Entra resources into entraGenericReconciler`

This Application reconciler is a near-identical copy (~90% structural overlap). With two consumers now, the extraction is justified. At minimum, `isNotFound` (duplicated verbatim at lines 390-399 and SecurityGroup:407-416) should be extracted to a package-level function.

This doesn't need to block the PR, but the debt is growing.

v2/api/entra/v1/application_types.go
```diff
@@ -0,0 +1,306 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+package v1
+
+import (
+	"github.com/microsoftgraph/msgraph-sdk-go/models"
```
theunrepentantgeek, line 6: 🔵 **Low: API types package imports Microsoft Graph SDK**

```go
import "github.com/microsoftgraph/msgraph-sdk-go/models"
```

The `AssignToApplication`/`AssignFromApplication` methods couple the CRD types package to the Graph SDK. Any consumer of these types (tools, tests, external controllers) must transitively depend on the SDK.

Consider moving the conversion logic to the reconciler package (e.g., `entra/convert.go`), keeping API types as pure data structs. Same issue exists in SecurityGroup types.

v2/internal/reconcilers/entra/entra_application_reconciler.go
```diff
@@ -0,0 +1,494 @@
...
+	ctx context.Context,
+	app *asoentra.Application,
+	log logr.Logger,
+) (ctrl.Result, error) {
+	log.V(Status).Info("Creating Entra application")
+
+	// Create our Entra Client
+	client, err := r.EntraClientFactory(ctx, app)
+	if err != nil {
+		return reconcile.Result{}, eris.Wrap(err, "creating entra client prior to creation")
```
theunrepentantgeek: 🔵 **Low: Inconsistent `reconcile.Result{}` vs `ctrl.Result{}`**

This line uses `reconcile.Result{}` while all other return sites use `ctrl.Result{}`. They're the same underlying type (the `ctrl` import aliases `controller-runtime`), so there's no runtime impact — purely cosmetic.

Same inconsistency exists in SecurityGroup reconciler (line 289).

theunrepentantgeek (commented):
## Deep Adversarial Code Review

Three independent review perspectives (Advocate, Skeptic, Architect) analyzed this PR. Findings synthesized below, posted as inline comments at relevant locations.

| Priority | Count | Summary |
|----------|-------|---------|
| 🔴 Critical | 1 | Nil pointer panic on external deletion |
| 🟠 High | 4 | OData injection, can't clear fields, FEATURE.md, dead field |
| 🟡 Medium | 3 | Status from pre-PATCH, missing configmap save, DisplayName validation |
| 🔵 Low | 3 | Extract shared reconciler, decouple SDK, cosmetic consistency |

**Verified robust**: Annotation persistence on partial create failure ✅, Delete path ✅, Webhook defaulting ✅

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: Agreed, we should fix this.

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: Agreed, we should fix this.

v2/api/entra/v1/application_types.go
theunrepentantgeek: Agreed, we should fix this.

FEATURE.md
theunrepentantgeek: Agreed, we'll remove this file.

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: Agreed, we should fix this.

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: Agreed, we should fix this.

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: Agreed, we should fix this.

v2/api/entra/v1/application_types.go
theunrepentantgeek: Disagree. #wontfix

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: Let's take small steps, and just extract `isNotFound()` as a helper for now.

v2/api/entra/v1/application_types.go
theunrepentantgeek: Disagree. This coupling is deliberate, treating our model as an anti-corruption layer.

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: Agree, we should always use `ctrl.Result{}`

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: Fixed in 12f171d0cd. The `update()` method now checks for `a == nil` after `loadApplicationByID` returns, since that function returns `(nil, nil)` for 404s rather than an error. The nil check triggers the requeue logic. Same fix applied to the SecurityGroup reconciler.

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: Fixed in 36e02fc908. Added an `escapeODataString()` helper that doubles single quotes per OData conventions. Applied to both the Application and SecurityGroup reconcilers' display name filter queries.

v2/api/entra/v1/application_types.go
theunrepentantgeek: Fixed in 733508e307. Removed all `if != nil` / `if len > 0` guards from `AssignToApplication` and its sub-type methods. All fields are now always set on the model, so clearing a value in the K8s spec (e.g., setting `identifierUris: []` or removing `web`) correctly propagates to the Graph API.

FEATURE.md
theunrepentantgeek: Fixed in 87b6c33449. Removed FEATURE.md from the repository.

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: Fixed in b6842c42dd. Removed the `CredentialProvider` field from both `EntraApplicationReconciler` and `EntraSecurityGroupReconciler`, along with the unused `identity` import.

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: Fixed in d6c216a5c9. The `update()` method now uses the return value from PATCH to populate status, ensuring it reflects the server's authoritative response. Applied to both Application and SecurityGroup reconcilers.

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: Fixed in 809e4162be. The `update()` method now calls `saveAssociatedKubernetesResources()` after updating status, matching the `create()` path. Applied to both Application and SecurityGroup reconcilers.

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: Fixed in 676b816d8c. Extracted `isNotFound()` as a package-level function in `errors.go`, removing the duplicate method from both reconcilers. Keeping the larger refactoring to a generic reconciler for a future PR.

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: Fixed in 695c57c434. Replaced `reconcile.Result{}` with `ctrl.Result{}` in both reconcilers and removed the now-unused `reconcile` import.

v2/azure-arm/entra.yaml
```diff
@@ -1,4 +1,6 @@
 groupModelConfiguration:
   v1: # Handcrafted
+    Application:
+      $supportedFrom: v2.20.0
     SecurityGroup:
-      $supportedFrom: v2.14.0
+      $supportedFrom: v2.14.0
```
theunrepentantgeek: Ensure EoLn at EoF.

v2/api/entra/v1/application_types.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+
+	// Tags
+	model.SetTags(spec.Tags)
+
+	model.SetIsFallbackPublicClient(spec.IsFallbackPublicClient)
+	model.SetGroupMembershipClaims(spec.GroupMembershipClaims)
+}
+
+type ApplicationStatus struct {
```
matthchr, line 164: How did you decide what to put in status, versus what not to?

v2/api/entra/v1/application_types.go
```diff
@@ -0,0 +1,298 @@
...
+
+	// DisplayName: The display name of the application.
+	DisplayName *string `json:"displayName,omitempty"`
+
+	// Conditions: The observed state of the resource
+	Conditions []conditions.Condition `json:"conditions,omitempty"`
+}
+
+func (status *ApplicationStatus) AssignFromApplication(model models.Applicationable) {
+	if model != nil {
```
matthchr: minor, style:

you could avoid nesting by bailing out early if model != nil
```
if model != nil {
    return
}
```

v2/internal/reconcilers/entra/entra_application_reconciler.go
```diff
@@ -0,0 +1,518 @@
...
+func (r *EntraApplicationReconciler) CreateOrUpdate(
+	ctx context.Context,
+	log logr.Logger,
+	eventRecorder record.EventRecorder,
+	obj genruntime.MetaObject,
+) (ctrl.Result, error) {
+	app, err := r.asApplication(obj)
+	if err != nil {
+		return ctrl.Result{}, eris.Wrapf(err, "creating or updating application %s", app.Name)
+	}
```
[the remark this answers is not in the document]
matthchr: This should probably be fixed

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
matthchr, line 35: Is this something that makes sense to do as part of this PR?

v2/internal/reconcilers/entra/entra_application_reconciler.go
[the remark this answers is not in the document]
matthchr: Applies throughout

v2/internal/reconcilers/entra/entra_application_reconciler.go
```diff
@@ -0,0 +1,518 @@
...
+		ReconcilerCommon: reconcilers.ReconcilerCommon{
+			KubeClient:         kubeClient,
+			PositiveConditions: positiveConditions,
+		},
+	}
+}
+
+func (r *EntraApplicationReconciler) CreateOrUpdate(
+	ctx context.Context,
+	log logr.Logger,
```
matthchr, line 64: minor, style/consistency: You dont have `logr.Logger` as a consistent parameter 2. Sometimes its 2, sometimes 3, sometimes 4 -- may be worth unifying on it always being second (if it's passed?)

v2/internal/reconcilers/entra/entra_application_reconciler.go
```diff
@@ -0,0 +1,518 @@
...
+		if isNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+
+		var displayName string
+		if app.Spec.DisplayName != nil {
+			displayName = *app.Spec.DisplayName
+		}
+
+		return ctrl.Result{}, eris.Wrapf(err, "failed to delete application %q (%s)", displayName, id)
```
matthchr: Why not just log the ASO resource name here? Seems more k8s-native?

v2/internal/reconcilers/entra/entra_application_reconciler.go
```diff
@@ -0,0 +1,518 @@
...
+
+	err = client.Client().Applications().ByApplicationId(id).Delete(ctx, nil)
+	if err != nil {
+		// If the application doesn't exist, return nil as we've successfully ensured that it doesn't exist
+		if isNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+
+		var displayName string
+		if app.Spec.DisplayName != nil {
```
matthchr: `displayName = to.FromPtr()` would work here, though see comment below I dont even think you need/want this naem at all?

v2/internal/reconcilers/entra/entra_application_reconciler.go
```diff
@@ -0,0 +1,518 @@
...
+	app.Spec.AssignToApplication(a)
+
+	result, err := client.Client().Applications().ByApplicationId(id).Patch(ctx, a, nil)
+	if err != nil {
+		// Failed to update
+		return ctrl.Result{}, eris.Wrapf(err, "failed to update application %s", id)
+	}
+
+	if result == nil {
+		// Didn't get a result back from the patch, load the application again to get the latest state
```
matthchr, line 201: Can you expand... is this expected? I don't fully understand what "no response from patch" means in terms of if it was expected or what ends up happening (which is that we load it and then return?)

v2/internal/reconcilers/entra/entra_application_reconciler.go
```diff
@@ -0,0 +1,518 @@
...
+	// Create our Entra Client
+	client, err := r.EntraClientFactory(ctx, app)
+	if err != nil {
+		return "", eris.Wrap(err, "creating entra client prior to adoption search")
+	}
+
+	// Try to find the application by DisplayName
+	displayName := app.Spec.DisplayName
+	if to.Value(displayName) == "" {
+		// Can't adopt without a display name
```
matthchr, line 250: Is this an error condition? Or success?

matthchr: Seems unexpected, should this be an error?

v2/internal/reconcilers/entra/entra_application_reconciler.go
```diff
@@ -0,0 +1,518 @@
...
+		// Default is AdoptOrCreate
+		return true
+	}
+
+	return app.Spec.OperatorSpec.CreationAllowed()
+}
+
+// saveAssociatedKubernetesResources retrieves Kubernetes resources to create and saves them to Kubernetes.
+// If there are no resources to save this method is a no-op.
+func (r *EntraApplicationReconciler) saveAssociatedKubernetesResources(
```
matthchr, line 468: Duplicate from EntraSecurityGroupReconciler? Wondering if we can combine some of this w/ like an interface that exposes `operatorSpec.ConfigMaps`? Similar to what we do with the generic reconciler?

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
matthchr: Related to https://github.com/Azure/azure-service-operator/pull/5232/changes#r3439149410

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
```diff
@@ -182,29 +177,45 @@ func (r *EntraSecurityGroupReconciler) update(
...
-			setEntraID(group, "")
-			return ctrl.Result{
-				Requeue: true,
-			}, nil
-		}
-
 		return ctrl.Result{}, eris.Wrapf(err, "getting group by ID %s", id)
 	}
 
+	if g == nil {
```
matthchr, line 183: minor: why move this?

v2/internal/testsamples/recordings/Test_Samples_CreationAndDeletion/Test_Entra_v1_CreationAndDeletion.yaml
```diff
@@ -153,45 +175,86 @@ interactions:
...
...vanarpsoutlook.onmicrosoft.com","serviceManagementReference":null,"signInAudience":"AzureADMyOrg","tags":[],"tokenEncryptionKeyId":null,"uniqueName":null,"samlMetadataUrl":null,"defaultRedirectUri":null,"certification":null,"optionalClaims":null,"servicePrincipalLockConfiguration":null,"requestSignatureVerification":null,"addIns":[],"api":{"acceptMappedClaims":null,"knownClientApplications":[],"requestedAccessTokenVersion":null,"oauth2PermissionScopes":[],"preAuthorizedApplications":[]},"appRoles":[],"info":{"logoUrl":null,"marketingUrl":null,"privacyStatementUrl":null,"supportUrl":null,"termsOfServiceUrl":null},"keyCredentials":[],"parentalControlSettings":{"countriesBlockedForMinors":[],"legalAgeGroupRule":"Allow"},"passwordCredentials":[],"publicClient":{"redirectUris":[]},"requiredResourceAccess":[],"verifiedPublisher":{"displayName":null,"verifiedPublisherId":null,"addedDateTime":null},"web":{"homePageUrl":null,"logoutUrl":null,"redirectUris":["https://localhost:5001/signin-oidc"],"implicitGrantSettings":{"enableAccessTokenIssuance":false,"enableIdTokenIssuance":true},"redirectUriSettings":[{"uri":"https://localhost:5001/signin-oidc","index":null}]},"spa":{"redirectUris":[]}}'
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

v2/internal/controllers/recordings/Test_Entra_SecurityGroup_v1_CRUD.yaml
```diff
@@ -57,19 +59,21 @@ interactions:
...
...ontext":"https://graph.microsoft.com/v1.0/$metadata#groups/$entity","id":"ffffdfde-468b-48be-b11d-4c1303f10cee","deletedDateTime":null,"classification":null,"createdDateTime":"2001-02-03T04:05:06Z","creationOptions":[],"description":null,"displayName":"ASO Test Security Group","expirationDateTime":null,"groupTypes":["Unified"],"infoCatalogs":[],"isAssignableToRole":null,"mail":"asotest-security-group@bevanarpsoutlook.onmicrosoft.com","mailEnabled":true,"mailNickname":"asotest-security-group","membershipRule":null,"membershipRuleProcessingState":null,"onPremisesDomainName":null,"onPremisesLastSyncDateTime":null,"onPremisesNetBiosName":null,"onPremisesSamAccountName":null,"onPremisesSecurityIdentifier":null,"onPremisesSyncEnabled":null,"preferredDataLocation":null,"preferredLanguage":null,"proxyAddresses":["SMTP:asotest-security-group@bevanarpsoutlook.onmicrosoft.com"],"renewedDateTime":"2001-02-03T04:05:06Z","resourceBehaviorOptions":[],"resourceProvisioningOptions":[],"securityEnabled":true,"securityIdentifier":"S-1-12-1-4294959070-1220429451-323755441-3993825539","theme":null,"uniqueName":null,"visibility":"Public","onPremisesProvisioningErrors":[],"serviceProvisioningErrors":[]}'
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

v2/internal/controllers/recordings/Test_Entra_Application_v1_CRUD.yaml
```diff
@@ -0,0 +1,214 @@
...
...PublicClient":null,"nativeAuthenticationApisEnabled":null,"notes":null,"publisherDomain":"bevanarpsoutlook.onmicrosoft.com","serviceManagementReference":null,"signInAudience":"AzureADMyOrg","tags":[],"tokenEncryptionKeyId":null,"uniqueName":null,"samlMetadataUrl":null,"defaultRedirectUri":null,"certification":null,"optionalClaims":null,"servicePrincipalLockConfiguration":null,"requestSignatureVerification":null,"addIns":[],"api":{"acceptMappedClaims":null,"knownClientApplications":[],"requestedAccessTokenVersion":null,"oauth2PermissionScopes":[],"preAuthorizedApplications":[]},"appRoles":[],"info":{"logoUrl":null,"marketingUrl":null,"privacyStatementUrl":null,"supportUrl":null,"termsOfServiceUrl":null},"keyCredentials":[],"parentalControlSettings":{"countriesBlockedForMinors":[],"legalAgeGroupRule":"Allow"},"passwordCredentials":[],"publicClient":{"redirectUris":[]},"requiredResourceAccess":[],"verifiedPublisher":{"displayName":null,"verifiedPublisherId":null,"addedDateTime":null},"web":{"homePageUrl":null,"logoutUrl":null,"redirectUris":[],"implicitGrantSettings":{"enableAccessTokenIssuance":false,"enableIdTokenIssuance":false},"redirectUriSettings":[]},"spa":{"redirectUris":[]}}'
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed with improved redaction.

v2/internal/controllers/recordings/Test_Entra_Application_v1_CRUD.yaml
```diff
@@ -0,0 +1,214 @@
...
+        headers:
+            Cache-Control:
+                - no-cache
+            Content-Type:
+                - application/json;odata.metadata=minimal;odata.streaming=true;IEEE754Compatible=false;charset=utf-8
+            Location:
+                - https://graph.microsoft.com/v1.0/9b0c84c7-21c4-4c92-ae62-d9c1eb53333f/directoryObjects/12bffd98-e5b5-4a9e-a8c1-25ac9b532f27/Microsoft.DirectoryServices.Application
```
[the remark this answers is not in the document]
theunrepentantgeek: Recorded with improved redaction.

v2/internal/reconcilers/entra/entra_application_reconciler.go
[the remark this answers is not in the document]
theunrepentantgeek: Fixed all three uses of `asApplication()`.

v2/internal/reconcilers/entra/entra_application_reconciler.go
```diff
@@ -0,0 +1,518 @@
...
+	log logr.Logger,
+	eventRecorder record.EventRecorder,
+	obj genruntime.MetaObject,
+) (ctrl.Result, error) {
+	log.V(Status).Info("Deleting Entra application")
+
+	app, err := r.asApplication(obj)
+	if err != nil {
+		return ctrl.Result{}, eris.Wrapf(err, "deleting application %s", app.Name)
+	}
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

v2/internal/reconcilers/entra/entra_application_reconciler.go
```diff
@@ -0,0 +1,518 @@
...
+func (r *EntraApplicationReconciler) UpdateStatus(
+	ctx context.Context,
+	log logr.Logger,
+	eventRecorder record.EventRecorder,
+	obj genruntime.MetaObject,
+) error {
+	app, err := r.asApplication(obj)
+	if err != nil {
+		return eris.Wrapf(err, "updating status of application %s", app.Name)
+	}
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
```diff
@@ -182,29 +177,45 @@ func (r *EntraSecurityGroupReconciler) update(
...
-	group.Status.AssignFromGroup(g)
+	if result == nil {
+		// Didn't get a result back from the patch, load the group again to get the latest state
+		log.V(Status).Info("No result returned from update, reloading group to get latest state")
+
+		result, err = r.loadGroupByID(ctx, id, client.Client())
+		if err != nil {
+			return ctrl.Result{}, eris.Wrapf(err, "getting group by ID %s after update returned no result", id)
+		}
+	}
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

v2/api/entra/v1/application_types.go
theunrepentantgeek: Honestly, I don't remember. Looks like it's just the stuff that has to be there, nothing more.

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
```diff
@@ -338,7 +349,7 @@ func (r *EntraSecurityGroupReconciler) UpdateStatus(
 	groupable, err := r.loadGroupByID(ctx, id, client.Client())
 	if err != nil {
 		// If the group doesn't exist, nothing to do as we're probably in the midst of deleting it
-		if r.isNotFound(err) {
+		if isNotFound(err) {
 			return nil
```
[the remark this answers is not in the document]

v2/api/entra/v1/application_types.go
theunrepentantgeek: Good idea, done.

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
theunrepentantgeek: Given how long this has sat around, I'm going to say _No_, but confirm should do it for the next Entra resource.

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: Generally, I keep `log` as the last parameter (as it's the least important), unless something overrides that.

Here, it's only the methods required by `genruntime.Reconciler` that differ from that.

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: Good catch. Fixed.

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: Yes, it's weird but [expected](https://learn.microsoft.com/en-us/graph/api/application-upsert?view=graph-rest-1.0&tabs=http#response):

> If an application or agentIdentityBlueprint object with uniqueName already exists, this method updates the [application](https://learn.microsoft.com/en-us/graph/api/resources/application?view=graph-rest-1.0) or [agentIdentityBlueprint](https://learn.microsoft.com/en-us/graph/api/resources/agentidentityblueprint?view=graph-rest-1.0) object and returns a 204 No Content response code.

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: Not an error - if we know the unique ID for an existing application, we could use that to "adopt" it and then read the DisplayName from the status. Adoption by DisplayName is a alternative for when we don't know that ID.

Though, having said that, I just saw that `DisplayName` is tagged as mandatory, so we'll never encounter this.

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: Yes, but not in this PR.

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
theunrepentantgeek: Because `loadGroupByID()` already uses `isNotFound()` and returns nil for that case.

So, a missing group won't show up as an error return, but as a nil result.

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: Removed as we're no longer using displayName in the error.

That said, we do need this name - it's mandatory when creating a new Entra Application, and it's what we use to do adoption (to hook up to existing Applications).

v2/api/entra/v1/application_types.go
matthchr: Are any of the other fields something that might make sense in status? For example if the user doesn't specify `IdentifierUris` are there any defaults? That sort of thing? If yes, would make sense to have in status too?

v2/api/entra/v1/application_types.go
```diff
@@ -0,0 +1,301 @@
...
+	SignInAudienceAzureADandPersonalMicrosoftAccount SignInAudience = "AzureADandPersonalMicrosoftAccount"
+	// SignInAudiencePersonalMicrosoftAccount indicates only users with personal Microsoft accounts can sign in.
+	SignInAudiencePersonalMicrosoftAccount SignInAudience = "PersonalMicrosoftAccount"
+)
+
+type ApplicationOperatorSpec struct {
+	// CreationMode: Specifies how ASO will try to create the resource.
+	// Specify "AlwaysCreate" to always create a new application when first reconciled.
+	// Or specify "AdoptOrCreate" to first try to adopt an existing application with the same display name.
+	// If multiple applications with the same display name are found, the resource condition will show an error.
```
matthchr:
```suggestion
	// If multiple applications with the same display name are found, the resource ready condition will show an error.
```

maybe?

Since there can technically be multiple conditions, it's the "ready" condition we're concerned with here.

v2/internal/reconcilers/entra/entra_application_reconciler.go
```diff
@@ -0,0 +1,514 @@
...
+	if err != nil {
+		// Failed to create
+		return ctrl.Result{}, eris.Wrapf(err, "failed to create application %s", app.Name)
+	}
+
+	if status != nil {
+		app.Status.AssignFromApplication(status)
+
+		if id := status.GetId(); id != nil {
+			setEntraID(app, *id)
```
matthchr: entraID only set after create (I assume UUID generated from server?), means that unlike with standard ASO resources where we know the name beforehand (so pod crash after create but before save updated state to etcd can still correctly DELETE the resource (see AddInitialResourceState and its call in `Claim`).

Is there any way to handle that case for Entra?

If not, maybe worth a callout in a comment in the code somewhere?
If yes, how do we tackle that? Delete by Display Name as a fallback?

matthchr: btw, assuming this is somewhat hard to fix, I am fine with an issue + followup later.

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
matthchr: Could also turn this TODO into a GH issue, if you wanted.

v2/internal/reconcilers/entra/entra_application_reconciler.go
```diff
@@ -0,0 +1,514 @@
...
+		// Didn't get a result back from the patch, load the application again to get the latest state
+		log.V(Status).Info("No result returned from update, reloading application to get latest state")
+
+		result, err = r.loadApplicationByID(ctx, id, client.Client())
+		if err != nil {
+			return ctrl.Result{}, eris.Wrapf(err, "getting application by ID %s after update returned no result", id)
+		}
+	}
+
+	app.Status.AssignFromApplication(result)
```
matthchr, line 217: 🤖 **BOT:** Could we handle `result == nil` after the reload here? Graph normally returns an empty PATCH response, and if the application is deleted between PATCH and the follow-up GET, `loadApplicationByID()` returns `(nil, nil)`. We then retain stale status/ConfigMap IDs and report the resource Ready. The updated SecurityGroup reconciler in this PR already clears the Entra ID and requeues for this case.

v2/internal/reconcilers/entra/entra_application_reconciler.go
```diff
@@ -0,0 +1,514 @@
...
+		// If the application doesn't exist, nothing to do as we're probably in the midst of deleting it
+		if isNotFound(err) {
+			return nil
+		}
+
+		// If the error is not a 404, return the error
+		return eris.Wrapf(err, "failed to update status of application %s", id)
+	}
+
+	app.Status.AssignFromApplication(applicationable)
```
matthchr, line 380: 🤖 **BOT:** `loadApplicationByID()` converts a 404 into `(nil, nil)`, so the `isNotFound(err)` branch above cannot handle a deleted application. Could we check `applicationable == nil` here before updating status and ConfigMaps? Otherwise reconcile-policy `skip` can publish stale IDs and report success after the Entra application has been deleted.

v2/api/entra/v1/application_types.go
```diff
@@ -0,0 +1,301 @@
...
+	// SignInAudience
+	if spec.SignInAudience != nil {
+		signInAudience := string(*spec.SignInAudience)
+		model.SetSignInAudience(&signInAudience)
+	} else {
+		model.SetSignInAudience(nil)
+	}
+
+	// IdentifierUris
+	model.SetIdentifierUris(spec.IdentifierUris)
```
matthchr: 🤖 **BOT:** Can we verify the clearing semantics for collection fields? The recorded requests show nil slices are omitted by Kiota rather than serialized as `null`, so removing `identifierUris`, `tags`, or redirect URIs from the CR appears to leave the old values in Entra while reporting Ready. We may need to normalize removed collections to non-nil empty slices and add a set-then-remove convergence test.

v2/api/entra/v1/application_types.go
```diff
@@ -0,0 +1,301 @@
...
+	// EnableIdTokenIssuance: Whether to enable ID token issuance in the implicit flow.
+	EnableIdTokenIssuance *bool `json:"enableIdTokenIssuance,omitempty"`
+
+	// EnableAccessTokenIssuance: Whether to enable access token issuance in the implicit flow.
+	EnableAccessTokenIssuance *bool `json:"enableAccessTokenIssuance,omitempty"`
+}
+
+// AssignToImplicitGrantSettings configures the provided instance with the details of implicit grant settings
+func (settings *ImplicitGrantSettings) AssignToImplicitGrantSettings(model models.ImplicitGrantSettingsable) {
+	if settings.EnableIdTokenIssuance != nil {
```
matthchr, line 320: 🤖 **BOT:** This differs from the parent assignment contract that fields are always set so prior values can be cleared. If token issuance was previously enabled and the field is later omitted, skipping the setter leaves it enabled in Entra. Could these setters be called unconditionally, or could the API explicitly document merge-only behavior here?

v2/internal/reconcilers/entra/entra_application_reconciler.go
```diff
@@ -0,0 +1,514 @@
...
+	a := msgraphmodels.NewApplication()
+	app.Spec.AssignToApplication(a)
+
+	status, err := client.Client().Applications().Post(ctx, a, nil)
+	if err != nil {
+		// Failed to create
+		return ctrl.Result{}, eris.Wrapf(err, "failed to create application %s", app.Name)
+	}
+
+	if status != nil {
```
matthchr: 🤖 **BOT:** Could a successful POST with a nil or incomplete response be treated as an error/reload condition instead of success? In this branch we record no resource ID and export no ConfigMap values, but still return success/Ready. With `AlwaysCreate`, a subsequent reconciliation could also create another orphaned application. At minimum, I think we should require a non-nil response and ID before completing creation.

matthchr (approved):
Approved, but please merge with caution as there are some comments from myself and review bot that give me pause.

v2/api/entra/v1/application_types.go
theunrepentantgeek: Given the usual ASO approach is to include everything in the Status unless there's a reason not to (e.g. secrets), I'll bring everything across.

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
theunrepentantgeek: Created #5632 to track.

v2/api/entra/v1/application_types.go
theunrepentantgeek: Done.

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: I don't see a quick fix - have lodged at #5633

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: Done.

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: Done.

v2/api/entra/v1/application_types.go
theunrepentantgeek: Now explicitly setting collections to empty when there are no members.

v2/api/entra/v1/application_types.go
theunrepentantgeek: Changed to always assign.

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: I don't think this can happen - the API is documented as
> If successful, this method returns 201 Created response code and an [application](https://learn.microsoft.com/en-us/graph/api/resources/application?view=graph-rest-1.0) ... object in the response body.

That said, it's easy enough to invert the condition and return an error if `status` is nil, ensuring the orphans don't happen.

v2/api/entra/v1/application_types.go
```diff
@@ -110,7 +113,11 @@ func (spec *ApplicationSpec) AssignToApplication(model models.Applicationable) {
 	}
 
 	// IdentifierUris
-	model.SetIdentifierUris(spec.IdentifierUris)
+	if len(spec.IdentifierUris) > 0 {
+		model.SetIdentifierUris(spec.IdentifierUris)
+	} else {
+		model.SetIdentifierUris([]string{})
```
matthchr, line 121: minor: May be worth a comment here explaining why we set empty slice? (and below)

v2/internal/reconcilers/entra/entra_application_reconciler.go
```diff
@@ -113,6 +113,12 @@ func (r *EntraApplicationReconciler) Delete(
 	}
 
 	// If we don't know the Entra ID of the application (captured in an annotation), there's nothing to do.
+	//
+	// Note that there's a hard to handle race condition here:
+	// If we attempt creation of the application in Entra, crash before we update the resource state in etcd,
```
matthchr, line 118: minor: Could link the issue that refers to this too

matthchr (approved):

v2/internal/reconcilers/entra/entra_application_reconciler.go
theunrepentantgeek: Done.

v2/api/entra/v1/application_types.go
theunrepentantgeek: Added comments to all 5 locations where this is done.

matthchr (approved):

PR #5267, opened 2026-03-22
Cache Go modules as a part of our devcontainer

theunrepentantgeek:
## What this PR does

Ensures our devcontainer has an up to date Go module cache, which should make our builds a little faster, and a lot more consistent. (Download times can vary from a few 10s of seconds to several minutes, depending on external factors.)

gif


.devcontainer/Dockerfile
```diff
@@ -22,9 +22,29 @@ RUN curl -fsSL https://download.docker.com/linux/debian/gpg | gpg --dearmor -o /
...
-COPY install-dependencies.sh .
+COPY .devcontainer/install-dependencies.sh .
 RUN ./install-dependencies.sh devcontainer && rm install-dependencies.sh
 
+# Pre-download Go modules for faster builds
+# v2 first (largest, most stable, needed by replace directives)
+COPY v2/go.mod v2/go.sum /tmp/mod-download/v2/
+RUN cd /tmp/mod-download/v2 && go mod download
+
+# asoctl (has replace => ../../ which resolves to v2/)
```
matthchr, line 33: I'm confused as to what this `(has replace => ../../ which resolves to v2/)` is saying?

matthchr: and below

.devcontainer/Dockerfile
```diff
@@ -22,9 +22,29 @@ RUN curl -fsSL https://download.docker.com/linux/debian/gpg | gpg --dearmor -o /
     && apt-get -y install azure-cli=2.76.0-1~bookworm \
     && rm -rf /var/lib/apt/lists/*
 
-COPY install-dependencies.sh .
+COPY .devcontainer/install-dependencies.sh .
 RUN ./install-dependencies.sh devcontainer && rm install-dependencies.sh
 
+# Pre-download Go modules for faster builds
+# v2 first (largest, most stable, needed by replace directives)
+COPY v2/go.mod v2/go.sum /tmp/mod-download/v2/
```
matthchr, line 32: Can this whole set of copy + runs be replaced with a single:
```
COPY . /tmp/mod-download/
RUN find /tmp/mod-download -name go.mod -execdir go mod download \;
```

?

I think the find command will find v2/ first, and we've ignored everything except the go.mod/go.sum files so I _think_ everything else will be avoided and it saves having to maintain the list of files anywhere except in the dockerignore?

matthchr: You'll need to remove the v2/ directory after the above though so we have a clean slate (not doing it in the temp dir)

.devcontainer/Dockerfile
theunrepentantgeek: The asoctl go.mod file has a replace like this:

```
replace github.com/Azure/azure-service-operator/v2 => ../../
```

so for the download to work, the `go.mod` for `v2` needs to be in the same relative position.

I'll add comments to make this clear.

.devcontainer/Dockerfile
theunrepentantgeek: Using `COPY . /tmp/mod-download` includes all the *.go files and other stuff (verified by including `RUN find /tmp/mod-download -name "*.go" -execdir pwd` as a temporary stage) so I've kept the individual copy statements.

.devcontainer/devcontainer.json
```diff
@@ -2,7 +2,10 @@
 // https://github.com/microsoft/vscode-dev-containers/tree/v0.137.0/containers/go
 {
 	"name": "Azure Service Operator",
-	"build": { "dockerfile": "Dockerfile" },
+	"build": {
+		"dockerfile": "Dockerfile",
+		"context": ".."
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

.devcontainer/Dockerfile
```diff
@@ -22,9 +22,29 @@ RUN curl -fsSL https://download.docker.com/linux/debian/gpg | gpg --dearmor -o /
...
+COPY v2/tools/generator/go.mod v2/tools/generator/go.sum /tmp/mod-download/v2/tools/generator/
+RUN cd /tmp/mod-download/v2/tools/generator && go mod download
+
+# mangle-test-json (standalone, no replace directives)
+COPY v2/tools/mangle-test-json/go.mod v2/tools/mangle-test-json/go.sum /tmp/mod-download/v2/tools/mangle-test-json/
+RUN cd /tmp/mod-download/v2/tools/mangle-test-json && go mod download
+
+# Clean up temporary module download directory
+RUN rm -rf /tmp/mod-download
+
```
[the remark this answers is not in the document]
theunrepentantgeek: Won't change, this is reversing the PR.

PR #5271, opened 2026-03-24
Add skill for hybrid versioning

matthchr:
## Checklist

- [ ] this PR contains tests


.agents/skills/migrate-group-versioning/SKILL.md
theunrepentantgeek, line 1: I think this is in the wrong location - should be `.agents/skills/migrate-group-versioning` with the name `migrate-group-versioning` specified in the header.

.github/skills/hybrid-versioning/SKILL.md
```diff
@@ -0,0 +1,222 @@
...
+AppConfiguration_KeyValue_v1api20240601_CRUD(tc, cs)
+// After
+AppConfiguration_KeyValue_v20240601_CRUD(tc, cs)
+```
+
+> **Note:** Some tests may already use the `v20*` naming in the test name even though they have `v1api` references in the code. In that case, no test function rename is needed. You can also skip step 4d as well.
+
+#### 4d. Re-record CRUD tests
+
+If the test function was renamed, the old recording files no longer match. Delete the old recording files and re-record the tests by running them against Azure:
```
theunrepentantgeek: Include the environment variables as prerequisites here. I used this:

> **Environment Variables:** You MUST have the following environment variables set to record the tests: `AZURE_SUBSCRIPTION_ID`, `AZURE_TENANT_ID`, and `ENTRA_APP_ID`. If any are missing, do not proceed because the tests will fail. Tell the user what is missing and stop the process.

.agents/skills/migrate-group-versioning/SKILL.md
matthchr: Are you sure that's correct?

https://docs.github.com/en/copilot/how-tos/use-copilot-agents/coding-agent/create-skills

Says:
> For a project skill, this file would be located in a .github/skills/github-actions-failure-debugging directory of your repository.

matthchr: Moved, looks like either location works at least for vscode.

PR #5273, opened 2026-03-25
Add testing skills for controllers and samples

theunrepentantgeek:
## What this PR does

Adds two agent skills, one for updating tests run by `controller:test-samples` and one for updating tests run by `controller:test-controllers`.

gif

### DIFF
--- a/.agents/skills/diagnosing-vcr-failures/SKILL.md
+++ b/.agents/skills/diagnosing-vcr-failures/SKILL.md
@@ -11,10 +11,10 @@ For running tests and understanding the test suites, see the **testing-aso-recor
 
 ## Quick Reference
 
-| | Controllers | Samples |
-|---|---|---|
-| **Task command** | `controller:test-controllers` | `controller:test-samples` |
-| **Log file** | `reports/test-controllers.log` | `reports/test-samples.log` |
+|                    | Controllers                           | Samples                                                                |
+| ------------------ | ------------------------------------- | ---------------------------------------------------------------------- |
+| **Task command**   | `controller:test-controllers`         | `controller:test-samples`                                              |
+| **Log file**       | `reports/test-controllers.log`        | `reports/test-samples.log`                                             |
 | **Recordings dir** | `v2/internal/controllers/recordings/` | `v2/internal/testsamples/recordings/Test_Samples_CreationAndDeletion/` |
 
 ## Identifying Failures
@@ -45,14 +45,14 @@ The test made an HTTP request that doesn't exist in the recording.
 The test found the right URL in the recording but the request body differs.
 
 - Recording **unmodified** → stale, re-record.
-- Recording **just created** (you literally just recorded it) → non-deterministic serialization. The request body is being serialized differently across runs. This is a **systemic issue**. Do NOT keep re-recording — investigate the VCR matching/hashing code and report to user.
+- Recording **just created** (you literally just recorded it) → non-deterministic serialization. The request body is being serialized differently across runs. This is a **systemic issue**. Do NOT keep re-recording — report to user.
 
 ### Test assertion failure
 
 **Signature:** assertion error from test code (e.g. "Expected true to be false", "Expected ... to have length"), NOT a go-vcr error.
 
 - **During recording** (live Azure) → likely a real bug in the controller or test code. Report to user.
-- **During playback** → typically means the replay roundtripper is returning stale cached responses that don't reflect the current resource state. The recording may need re-creating, but first check if it's a known issue with the replay infrastructure. Report the specific assertion and context to user.
+- **During playback** → typically means the replay roundtripper is returning stale cached responses that don't reflect the current resource state. The recording may need re-creating, but if not, report the specific assertion and context to user.
 
 ### Timeout while recording
 
@@ -84,18 +84,21 @@ Delete the recording file for the failing test from the appropriate recordings d
 Run as a **background terminal**:
 
 **Controllers:**
+
 ```bash
-source test.env && TIMEOUT=60m TEST_FILTER="<your-test-here>" ./hack/tools/task controller:test-controllers
+source test.env && TIMEOUT=60m TEST_FILTER="<your-test-here>" task controller:test-controllers
 ```
 
 **Samples** (note the `Test_Samples_CreationAndDeletion/` prefix is required):
+
 ```bash
-source test.env && TIMEOUT=60m TEST_FILTER="Test_Samples_CreationAndDeletion/<your-test-here>" ./hack/tools/task controller:test-samples
+source test.env && TIMEOUT=60m TEST_FILTER="Test_Samples_CreationAndDeletion/<your-test-here>" task controller:test-samples
 ```
 
 Example: to re-record `Test_Redis_v1api20230801_CreationAndDeletion`:
+
 ```bash
-source test.env && TIMEOUT=60m TEST_FILTER="Test_Samples_CreationAndDeletion/Test_Redis_v1api20230801_CreationAndDeletion" ./hack/tools/task controller:test-samples
+source test.env && TIMEOUT=60m TEST_FILTER="Test_Samples_CreationAndDeletion/Test_Redis_v1api20230801_CreationAndDeletion" task controller:test-samples
 ```
 
 For slow-provisioning resources, use `TIMEOUT=90m`. Record one test at a time to isolate problems.
@@ -105,7 +108,7 @@ For slow-provisioning resources, use `TIMEOUT=90m`. Record one test at a time to
 Run the **same command without `source test.env`**:
 
 ```bash
-TIMEOUT=60m TEST_FILTER="<your-test-here>" ./hack/tools/task controller:<SUITE>
+TIMEOUT=60m TEST_FILTER="<your-test-here>" task controller:<SUITE>
 ```
 
 Playback takes 1–3 min. If playback fails but recording succeeded, this is a systemic VCR issue (e.g. non-deterministic request serialization) — do NOT keep re-recording. Investigate and report.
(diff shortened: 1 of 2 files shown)


.agents/skills/test-controllers/SKILL.md
```diff
@@ -0,0 +1,148 @@
...
+Before running tests:
+
+1. **Check environment variables.** Verify that `AZURE_SUBSCRIPTION_ID`, `AZURE_TENANT_ID`, and `ENTRA_APP_ID` are set (or that the user has provided a `test.env` file to source). These are required for recording new tests against live Azure. If any are missing and recordings may need to be created, do not proceed — tell the user what is missing and stop. Note: environment variables are only needed for **recording** (creating new recordings against live Azure). Playback runs do not need them.
+
+### Run tests
+
+Start by running all the tests as a background terminal command:
+
+```bash
+source test.env && ./hack/tools/task controller:test-controllers
```
matthchr: I think hack is added to the path if you're in `dev.sh` so probably unneeded to specify the full path?

matthchr: May need to use the `-record` variant, though now that I look I see that `controller:test-integration-envtest-record` exists but we dont have record variants for the individual ones... should we?

The main difference is that some tests will not pass with the default timeout, they need a longer timeout, and `-record` variants pass LIVE_TEST_TIMEOUT which is longer.

.agents/skills/test-controllers/SKILL.md
```diff
@@ -0,0 +1,148 @@
...
+### Verify success
+
+After the test command finishes with exit code 0, always confirm there are no failures in the log:
+
+```bash
+grep "FAIL:" reports/test-controllers.log || echo "No failures found"
+```
+
+* If no failures are found, you're done and can stop working.
+* If failures are found despite a zero exit code, proceed to the "Identify test failures" section below.
```
matthchr: Can this actually happen?

I dont think I've seen it, and if it can, we should fix it?

.agents/skills/test-controllers/SKILL.md
matthchr: We should share how to focus tests, as it's quite useful. Ahh wait it's below... maybe should be called out more clearly/explicitly?

.agents/skills/test-controllers/SKILL.md
```diff
@@ -0,0 +1,148 @@
...
+3. Update the sample YAML to use a different Azure region (e.g. change `eastus` to `australiaeast` or `westus2`).
+4. Delete the failed/partial recording if one exists, and re-run the test to create a new recording.
+5. If the new region also fails with a capacity error, try another region and repeat.
+
+### Update a recording
+
+Once you've identified a test that needs to be rerecorded, delete the recording file and run this command as a **background terminal** to create a new recording:
+
+```bash
+source test.env && TIMEOUT=60m TEST_FILTER="<your-test-here>" ./hack/tools/task controller:test-controllers
```
matthchr: same comment about full path probably unneeded here

.agents/skills/test-controllers/SKILL.md
```diff
@@ -0,0 +1,148 @@
...
+```
+
+For tests involving slow-provisioning resources (e.g. ApiManagement, Kusto), use `TIMEOUT=90m`.
+
+Recordings should be updated one at a time to make it easier to identify problems.
```
matthchr: I think this is somewhat incorrect, or at least not what I do. If I have a lot of recordings to update (for example in the networking PR), I run a baseline run first to see what passes first go (usually most things), and then I go 1 at a time after that. If you're doing 20+ recordings, 1 at a time is crazy slow. When I did networking only 2 tests didnt pass first go, so I only had to focus in on those.

.agents/skills/test-controllers/SKILL.md
```diff
@@ -0,0 +1,148 @@
...
+
+After you have successfully updated a recording, run the test again (same command, **without** `source test.env`) to verify that playback works correctly. Playback is much faster than recording — typically 1-3 minutes. If playback fails but recording succeeded, this indicates a systemic issue with the VCR infrastructure (e.g. non-deterministic request serialization), NOT a stale recording. Do not keep re-recording — investigate and report to the user.
```
matthchr: source test.env is not the key. If the file exists, you can source test.env as much as you want it wont re-record.

.agents/skills/test-controllers/SKILL.md
```diff
@@ -0,0 +1,148 @@
...
+
+### Confirmation
+
+Once you've updated one or more recordings and the tests pass, start over from the top of this workflow and run all the tests again to confirm that everything is working correctly. If you find new failures, investigate and address them as described above until you have a clean test run with no failures.
+
+Why do this? A tests run winds up as soon as one test fails, leaving later tests unexecuted. By starting over and running all the tests, you can confirm that all problems are fully resolved and that there are no other hidden issues.
+
+## Key information
+
+Recordings of samples tests are found in `v2/internal/controllers/recordings/`, named for the test being executed.
```
matthchr: Do we need 2 skills? Wondering if just one would do, given we're talking about samples here too and there's not much difference in guidance between the two?

Feels like a lot of duplication between the two skills atm.

.agents/skills/test-controllers/SKILL.md
matthchr: Oh wait I see you call this out below. Don't you have to be in dev.sh anyway, or else envtest won't work? I'm confused how `task` could be the wrong version but something like kube-apiserver wouldn't have the same issues?

I think we should just require/expect that you're in the context of dev.sh or the devcontainer here, and that those envs work, rather than working around them not being set up correctly (which will likely cause other issues with the envtest tools, etc).

.agents/skills/test-controllers/SKILL.md
```diff
@@ -0,0 +1,130 @@
+---
+name: test-controllers
+description: Use when you need to verify that ASO CRUD tests of controllers are correct and functional.
+---
+
+# Verify ASO controller tests
+
+When creating new resources, importing new resource versions, implementing or modifying resource extensions, or making other changes that change the sequence or content of HTTP requests made to Azure, we can need to replay and/or rerecord tests of our controllers.
```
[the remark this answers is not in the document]
theunrepentantgeek: I've refactored the skills, so this is stale.

.agents/skills/test-controllers/SKILL.md
theunrepentantgeek: Yeah, this is weird. I've updated the skills to expect `task` on the path and to stop if it's not.

theunrepentantgeek: I don't think we actually need the `-record` variants; at least, I never use them. I always prefix `TIMEOUT` if needed, and the skills do that.

theunrepentantgeek: With the refactored tests, it's up near the front.

.agents/skills/test-controllers/SKILL.md
```diff
@@ -0,0 +1,130 @@
...
+
+```bash
+./hack/tools/task controller:test-controllers
+```
+
+* Do NOT add `| tee` or other output redirection — the taskfile already handles piping output to `reports/test-controllers.log`.
+
+* If recordings for all tests exist, this typically completes in about 6-10 minutes. Do not interrupt the test, just wait for it to complete.
+
+* If any recordings are missing, each will result in a live test automatically run against Azure to create the missing recording. For most resources, this will take under 20m, but for some resources a recording may take 60m or more.
```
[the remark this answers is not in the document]
theunrepentantgeek: I've refactored the skills, so this is stale.

.agents/skills/test-controllers/SKILL.md
theunrepentantgeek: We don't need this. Removed.

.agents/skills/test-controllers/SKILL.md
theunrepentantgeek: Fixed up.

.agents/skills/test-controllers/SKILL.md
theunrepentantgeek: The skills only go one at a time once a problem test has been identified - they start by running the entire test suite at once, which will update most (all) of the recordings. 

So, I think they're already doing the right thing.

.agents/skills/test-controllers/SKILL.md
theunrepentantgeek: Updated, but I think it was just saying "you don't need to source test.env for playback", not that this controlled the choice between record/playback.

.agents/skills/test-controllers/SKILL.md
theunrepentantgeek: I refactored to two different skills. One for the overall process, one for troubleshooting failures.

.agents/skills/test-samples/SKILL.md
```diff
@@ -0,0 +1,132 @@
+---
+name: test-samples
+description: Use when you need to verify that ASO samples are correct and functional.
+---
+
+# Verify ASO samples
+
+When creating new samples, modifying existing samples, implementing resource extensions, or making other changes that change the sequence or content of HTTP requests made to Azure, we can need to replay and/or rerecord tests of our samples.
```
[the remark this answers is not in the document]
theunrepentantgeek: Things have changed, this comment is stale.

.agents/skills/test-samples/SKILL.md
```diff
@@ -0,0 +1,132 @@
...
+
+* If all the tests pass, then you're done and can stop working.
+
+Variation: if you've been asked to run tests only for a specific group of resources, add the TEST_FILTER environment variable on the command line:
+
+```bash
+TEST_FILTER="Test_Samples_CreationAndDeletion/Test_<test-group-here>.*" ./hack/tools/task controller:test-samples
+```
+
+The test name pattern is constructed from the sample directory structure: for a sample at `v2/samples/<group>/<subgroup>/<version>/`, the test name is `Test_<TitleCasedGroup><TitleCasedSubgroup>_<version>_CreationAndDeletion`. For example, `v2/samples/documentdb/mongodb/v1api20231115/` produces `Test_Mongodb_v1api20231115_CreationAndDeletion`. (Note: only the last path segment before the version is used — `documentdb` is the parent directory and `mongodb` becomes `Mongodb`.)
```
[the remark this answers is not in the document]
theunrepentantgeek: Things have changed, this comment is stale.

.agents/skills/test-samples/SKILL.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Identify test failures
+
+If the test run fails, look through the logs to identify which tests failed, and why. 
+
+* Logs are likely to overflow the terminal buffer, so they get written to 'reports/test-samples.log' as well. 
+* Test failures can be identified by searching for "FAIL:" in the logs (the trailing colon is important to avoid false positives).
+* For each tests, work out why the test failed. Note that initial failures tend to result in a cascade of following failures, so it's important to identify the first problem.
```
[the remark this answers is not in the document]
theunrepentantgeek: This change is on another branch. It's correct.

PR #5274, opened 2026-03-26
Fix subscription check guardrail

theunrepentantgeek:
## What this PR does

The `checkSubscription()` method in `azure_generic_arm_reconciler_instance.go` can incorrectly allow reconciliation to proceed for subscription scoped resources outside the subscription for which the credential is intended.

With a properly scoped credential (one limited to just the configured subscription), this will result in a reconciliation failure and an error message that may be hard to parse (depending on the exact error returned).

With an overly broad credential (e.g. one with rights across an entire tenant) this might result in users creating resources in unexpected places.

gif

### DIFF
--- a/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance_test.go
+++ b/v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance_test.go
@@ -27,13 +27,13 @@ func TestAzureDeploymentReconcilerInstance_CheckSubscription_GivenResourceID_Ret
 		resourceID             string
 		expectedErrorSubstring string
 	}{
-		"valid resource group ID": {
+		"resource group in expected subscription": {
 			resourceID: "/subscriptions/{expectedSub}/resourceGroups/asotest-rg",
 		},
-		"valid resource group ID (lowercase)": {
+		"resource group in expected subscription (lowercase)": {
 			resourceID: "/subscriptions/{expectedSubLower}/resourceGroups/asotest-rg",
 		},
-		"invalid resource group ID": {
+		"resource group in different subscription": {
 			resourceID:             "/subscriptions/{unexpectedSub}/resourceGroups/asotest-rg",
 			expectedErrorSubstring: "resource does not match with Client Credential",
 		},


timmy-wright:
## 🔍 Deep Review: PR #5274 — Fix subscription check guardrail

**Panel**: Advocate · Skeptic · Architect | **Verdict**: ✅ Approve with required fixes

---

### What the PR Does

The `checkSubscription()` guard previously used `parsedRID.ResourceGroupName != ""` as its gate — meaning it only validated subscription IDs for *resource-group-scoped* resources. Subscription-scoped resources (e.g. subscription-level role assignments) had **no subscription check at all**. The fix is a single-word change: `ResourceGroupName` → `SubscriptionID`. A test file is added.

---

### 🚨 Required Before Merge

#### 1. Two "matching subscription" test cases are inert — they test nothing

**Severity**: High | **Panel consensus**: All three agents flagged this independently.

The test cases `"Role assignment to resource group with matching subscription"` and `"Role assignment to subscription with matching subscription"` use malformed resource IDs missing the `/subscriptions/` prefix:

```go
// BROKEN - missing /subscriptions/ prefix:
"/{expectedSub}/resourceGroups/asotest-rg-mnlufu/providers/Microsoft.Authorization/roleAssignments/..."
"/{expectedSub}/providers/Microsoft.Authorization/roleAssignments/..."
```

`arm.ParseResourceID` fails on both. Since `checkSubscription` is wrapped in `if err == nil`, the guard is never entered. The tests pass trivially via the silent-skip path — **not** because the subscription check worked. If the old buggy condition (`ResourceGroupName != ""`) were restored, these tests would still pass.

**Fix**: Use valid ARM IDs with the `/subscriptions/` prefix. Also remove `?api-version=...` query strings which can also cause parse failures.

```go
"Role assignment to resource group with matching subscription": {
    resourceID: "/subscriptions/{expectedSub}/resourceGroups/asotest-rg/providers/Microsoft.Authorization/roleAssignments/ed3c9f84-ae20-5aea-8602-58ed2eeffef1",
},
"Role assignment to subscription with matching subscription": {
    resourceID: "/subscriptions/{expectedSub}/providers/Microsoft.Authorization/roleAssignments/ed3c9f84-ae20-5aea-8602-58ed2eeffef1",
},
```

---

#### 2. Case-sensitive `!=` comparison should use the existing `EqualFold` helper

**Severity**: High | **Panel consensus**: All three agents flagged this independently.

The new condition:
```go
parsedRID.SubscriptionID != r.ARMConnection.SubscriptionID()
```
is case-sensitive. Subscription IDs are GUIDs (case-insensitive by RFC 4122). A credential secret storing `"00000000-0000-0000-0000-00000000ABCD"` against an ARM resource ID containing `"00000000-0000-0000-0000-00000000abcd"` would trigger a **false-positive mismatch**, permanently blocking reconciliation.

The codebase already has the correct implementation: `genruntime.CheckARMIDMatchesSubscription` (`v2/pkg/genruntime/arm_id.go`) uses `strings.EqualFold` and handles the empty-subscription case. It is already used in `resource_hierarchy.go`. The `genruntime` package is already imported in this file.

**Fix**: Replace the inline logic with the existing helper:

```go
if err == nil {
    if !genruntime.CheckARMIDMatchesSubscription(r.ARMConnection.SubscriptionID(), parsedRID) {
        err = eris.Errorf(
            "SubscriptionID %q for %q resource does not match with Client Credential: %q",
            parsedRID.SubscriptionID,
            resourceID,
            r.ARMConnection.SubscriptionID())
        return conditions.NewReadyConditionImpactingError(
            err, conditions.ConditionSeverityError, conditions.ReasonSubscriptionMismatch)
    }
}
```

This also resolves the `TODO: we need a better way?` comment — the helper *is* the better way.

---

### 💡 Recommended Improvements

#### 3. Add a case-insensitive test case

Once fix #2 is applied, add a test to cover it:
```go
"Resource with uppercase subscription ID in credential": {
    resourceID: "/subscriptions/{expectedSubUppercase}/resourceGroups/asotest-rg",
    // Should succeed — same GUID, different casing
},
```

#### 4. Consider logging on parse failure (pre-existing, low priority)

When `arm.ParseResourceID` fails, the check is silently skipped with no log entry. A `Debug`-level log on this path would make the bypass visible in production diagnostics. Not introduced by this PR, but worth a follow-up.

---

### ✅ What's Correct

- **The core fix is right.** `parsedRID.SubscriptionID != ""` correctly gates both resource-group-scoped *and* subscription-scoped resources. The old condition was simply wrong.
- **Both call sites are covered** (~line 301 and ~line 990) since the fix is at the method level.
- **Subscription aliases bypass is correct.** `/providers/Microsoft.Subscription/aliases/...` parses to an empty `SubscriptionID`, so the guard is correctly skipped.
- **No `c := c` needed.** The codebase uses Go 1.24+, where loop variable semantics changed in 1.22 — each iteration already has its own binding. The existing Copilot reviewer comment on this is a false positive.
- **`fakeConnection` is well-designed** — compile-time interface satisfaction via `var _ Connection = &fakeConnection{}`, scoped to the test package.

---

### Summary

| Issue | Severity | Introduced? | Fix |
|---|---|---|---|
| Two "matching" test cases are hollow | High | ✅ This PR | Fix resource ID format |
| Case-sensitive GUID comparison | High | ✅ This PR | Use `genruntime.CheckARMIDMatchesSubscription` |
| `c := c` rebinding comment | — | Pre-existing reviewer | False positive for Go 1.22+ |
| Silent bypass on parse failure | Low | Pre-existing | Add debug log (follow-up) |

The core change is sound and the direction is right. Fix the two test bugs and adopt the existing `EqualFold` helper before merging.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance_test.go
```diff
@@ -0,0 +1,105 @@
...
+		"Role assignment to subscription with non-matching subscription": {
+			resourceID:             "/subscriptions/{unexpectedSub}/providers/Microsoft.Authorization/roleAssignments/asotest-role-assignment",
+			expectedErrorSubstring: "resource does not match with Client Credential",
+		},
+	}
+
+	expectedSub := "00000000-0000-0000-0000-000000000000"
+	unexpectedSub := "11111111-1111-1111-1111-111111111111"
+
+	for name, c := range cases {
```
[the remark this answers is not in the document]
theunrepentantgeek: Range variables are not reused, since Go 1.22.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance_test.go
```diff
@@ -0,0 +1,105 @@
...
+		},
+		"Role assignment to resource group with matching subscription": {
+			resourceID: "/{expectedSub}/resourceGroups/asotest-rg-mnlufu/providers/Microsoft.Authorization/roleAssignments/ed3c9f84-ae20-5aea-8602-58ed2eeffef1?api-version=2022-04-01",
+		},
+		"Role assignment to resource group with non-matching subscription": {
+			resourceID:             "/subscriptions/{unexpectedSub}/resourceGroups/asotest-rg/providers/Microsoft.Authorization/roleAssignments/asotest-role-assignment",
+			expectedErrorSubstring: "resource does not match with Client Credential",
+		},
+		"Role assignment to subscription with matching subscription": {
+			resourceID: "/{expectedSub}/providers/Microsoft.Authorization/roleAssignments/ed3c9f84-ae20-5aea-8602-58ed2eeffef1?api-version=2022-04-01",
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed resource IDs, but note that role assignments can go outside subscriptions.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -430,14 +430,22 @@ func (r *azureDeploymentReconcilerInstance) preReconciliationCheck(
...
-		if parsedRID.ResourceGroupName != "" && parsedRID.SubscriptionID != r.ARMConnection.SubscriptionID() {
-			err = eris.Errorf("SubscriptionID %q for %q resource does not match with Client Credential: %q", parsedRID.SubscriptionID, resourceID, r.ARMConnection.SubscriptionID())
-			return conditions.NewReadyConditionImpactingError(err, conditions.ConditionSeverityError, conditions.ReasonSubscriptionMismatch)
+		if parsedRID.SubscriptionID != "" && parsedRID.SubscriptionID != r.ARMConnection.SubscriptionID() {
+			err = eris.Errorf(
+				"SubscriptionID %q for %q resource does not match with Client Credential: %q",
+				parsedRID.SubscriptionID,
+				resourceID,
+				r.ARMConnection.SubscriptionID())
+
```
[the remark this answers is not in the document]
theunrepentantgeek: Modified to use case insensitive check.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -429,15 +429,29 @@ func (r *azureDeploymentReconcilerInstance) preReconciliationCheck(
...
-		if parsedRID.ResourceGroupName != "" && parsedRID.SubscriptionID != r.ARMConnection.SubscriptionID() {
-			err = eris.Errorf("SubscriptionID %q for %q resource does not match with Client Credential: %q", parsedRID.SubscriptionID, resourceID, r.ARMConnection.SubscriptionID())
-			return conditions.NewReadyConditionImpactingError(err, conditions.ConditionSeverityError, conditions.ReasonSubscriptionMismatch)
-		}
+	if err != nil {
+		// We never expect the resource ID to be invalid, so return an error here to avoid someone
+		// mangling a resource annotation and bypassing the check
+		err = eris.Wrapf(err, "parsing resource ID %q", resourceID)
+
+		return conditions.NewReadyConditionImpactingError(
```
matthchr, line 440: I think this is causing `Test_SubscriptionAndAlias_CRUD` to fail - check CI

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance_test.go
```diff
@@ -0,0 +1,110 @@
...
+ * Licensed under the MIT license.
+ */
+
+package arm
+
+import (
+	"strings"
+	"testing"
+
+	. "github.com/onsi/gomega"
```
matthchr, line 12: CI gagging on formatting up here

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance_test.go
```diff
@@ -0,0 +1,110 @@
...
+		resourceID             string
+		expectedErrorSubstring string
+	}{
+		"valid resource group ID": {
+			resourceID: "/subscriptions/{expectedSub}/resourceGroups/asotest-rg",
+		},
+		"valid resource group ID (lowercase)": {
+			resourceID: "/subscriptions/{expectedSubLower}/resourceGroups/asotest-rg",
+		},
+		"invalid resource group ID": {
```
matthchr: Are these test names accurate? Says invalid resource group ID but the only thing that looks invalid about it is that it has a non-matching sub ID?

So either it's the sub-ID that's invalid (invalid sub ID)? or maybe just in general need some better names here?

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
theunrepentantgeek: That test was failing, but not due to this. We were failing to redact the subscription ID properly, so when the test ran we were comparing `00000000-0000-0000-0000-000000000000` with a non-nil GUID and failing.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance_test.go
theunrepentantgeek: Fixed.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance_test.go
theunrepentantgeek: Updated test names.

PR #5285, opened 2026-03-31
Experiment: Use GitHub runners

theunrepentantgeek:
## What this PR does

Given we have recently split our builds into several steps, testing the hypothesis that they are now small enough to succeed on standard runners.

gif


theunrepentantgeek:
A successful experiment - showed that using the GH runners isn't viable.

PR #5295, opened 2026-04-06
Update Networking API version 2025-03-01

### LINKED ISSUE #5251: Support forAzure NAT Gateway StandardV2

As of today ASO cannot create NAT Gateway with the AKS SKU StandardV2 

## Describe the improvement

Needs to be supported in special after GA annoncenment from Jan 2026
https://azure.microsoft.com/en-us/updates/?id=547772
https://learn.microsoft.com/en-us/azure/nat-gateway/nat-overview#standardv2-nat-gateway 

Requeired StandardV2 Public IP https://learn.microsoft.com/en-us/azure/virtual-network/ip-services/public-ip-addresses 

## Additional context

Azure API https://learn.microsoft.com/en-us/rest/api/virtualnetwork/nat-gateways/create-or-update?view=rest-virtualnetwork-2025-05-01&tabs=HTTP#natgatewayskuname

matthchr:
Fixes #5251

Note that there is a newer networking API version, 2025-05-01, but it has migrated to typespec and doesn't seem to have a Swagger available for it anymore (though it _used_ to). I am asking about this through other channels, but in the meantime bumping to the newest version that will work for us.

gif

## Checklist

- [x] this PR contains tests


v2/azure-arm.yaml
```diff
@@ -4236,6 +4236,12 @@ objectModelConfiguration:
       VirtualNetworks_VirtualNetworkPeering:
         $exportAs: VirtualNetworksVirtualNetworkPeering
         $supportedFrom: v2.0.0-alpha.1
+      VpnClientConfiguration:
+        RadiusServerSecret:
+          $isSecret: true
+      RadiusServer:
+        RadiusServerSecret:
+          $isSecret: true
```
theunrepentantgeek, line 4244: Breaking change, right? We need to call this out in the release notes.

matthchr: Tracking via https://github.com/Azure/azure-service-operator/issues/5298

PR #5303, opened 2026-04-09
Create ADR for properties that are sometimes secret and sometimes not

theunrepentantgeek:
## What this PR does

Adds a design document discussing the options for handling properties that might sometimes be sensitive, but mostly are not.

gif

## Checklist


docs/hugo/content/design/ADR-2026-04-Secrets/_index.html
```diff
@@ -0,0 +1,158 @@
+---
+title: '2026-04: Optional Secrets'
+toc_hide: true
```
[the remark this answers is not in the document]
theunrepentantgeek: This is correct for Hugo.

docs/hugo/content/design/ADR-2026-04-Secrets/_index.html
```diff
@@ -0,0 +1,158 @@
...
+
+To support this, we need more flexibility than a simple true/false answer to the question "is this property a secret?" -
+we need the ability to handle a property which is an optional secret, i.e. it can be treated as a secret if the user
+chooses to do so, but it doesn't have to be.
+
+The problem falls into two natural parts:
+
+* How do we represent this in our configuration?
+* How do we transform the property in our code generator?
+
```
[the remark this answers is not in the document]
theunrepentantgeek: Updated.

docs/hugo/content/design/ADR-2026-04-Secrets/_index.html
```diff
@@ -0,0 +1,158 @@
...
+As we've done for properties that can be optionally populated from ConfigMaps, we could have two properties in the
+generated code - one for the plain text value and one for the secret reference. They'd be mutually exclusive, and the
+user would populate one or the other.
+
+For consistency with the way we've handled properties that can be optionally populated from ConfigMaps, we could use the
+suffix `Secret` for the secret reference property
+
+### Pros
+
+Consistent with existing patterns for optional ConfigMap properties.
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

docs/hugo/content/design/ADR-2026-04-Secrets/_index.html
```diff
@@ -0,0 +1,175 @@
...
+as either `true` or `false`.
+
+Specifying `$isSecret: true` has been required because the Azure REST API specs don't always call out which properties
+are secrets; our config allows us to address omissions.
+
+We've also used some heuristics to detect properties which might be secrets, requiring an explicit decision to be made.
+Supporting `$isSecret: false` allows us to configure false positives in these cases.
+
+To date, this has covered all of our needs - but we now have an interesting case (identified in
+[#5095](https://github.com/Azure/azure-service-operator/issues/5095)) where a property is _sometimes_ a secret but not
```
matthchr, line 18: I just found another hit of this case too, in APIM

docs/hugo/content/design/ADR-2026-04-Secrets/_index.html
```diff
@@ -0,0 +1,175 @@
...
+
+### Cons
+
+* Less consistent with existing patterns for optional ConfigMap properties.
+* Contrary to Kubernetes conventions, which explicitly recommend avoiding union/polymorphic properties.
+
+
+## Decision
+
+Configuration Option 4: Change to new Enum, with backward compatibility.
```
matthchr, line 157: I'm not sold that we really need the back-compat complexity. It seems like it might be easier to just make a one-shot change? The number of outstanding PRs that would be impacted is not huge, and _probably_ it's mostly us. I'd honestly rather do a bit more work now to ensure that we have a limited/more understandable number of options in azure-arm.yaml at the end of the day.

docs/hugo/content/design/ADR-2026-04-Secrets/_index.html
```diff
@@ -0,0 +1,175 @@
...
+### Cons
+
+* Most significant change to the configuration and code, as we need to replace an existing property with a new one and
+update all references to it.
+* Requires users to update their configuration to use the new property, which is a breaking change to any work currently
+in progress.
+
+## Configuration Option 4: Change to new Enum, with backward compatibility
+
+Replace the existing `$isSecret` boolean with a new enum property, say `$secret` with values `always`, `never`, and
```
matthchr: Should we just match `importConfigMapMode`?

(From our azure-arm.yaml):
```
# $importConfigMapMode: <optional|required>
#     Specifies that the property can be imported from a config map.
#     Optional: The property may be specified as string or imported from a config map.
#               To achieve this in a non-breaking way, a new property is added to the object living alongside
#               the existing property. The new property is called <propName>FromConfig.
#     Required: The property must be specified from a config map, it cannot be given as a raw string.
```

Maybe with an added `never` if you want the 3 states explicitly (I _think_ w/ `importConfigMapMode` we treat the absence of the field as an implied 3rd state == never. I don't love that but am fine with it if we wanted to do it for this secret field too)

Mostly I think the names and values should match as much as possible for consistency.

matthchr (approved):
Approved, though I have notes about specifics (see comments)

docs/hugo/content/design/ADR-2026-04-Secrets/_index.html
theunrepentantgeek: We have a bunch of examples of `$isSecret: false` - used to turn off our heuristics, so we will definitely need `never` as a possible value. ConfigMaps don't have heuristics popping up, so they don't need the third value.

theunrepentantgeek: I don't love the name `$importConfigMapMode` but going with `$importSecretMode` for consistency would be good. Changed.

docs/hugo/content/design/ADR-2026-04-Secrets/_index.html
theunrepentantgeek: I'm only thinking of having back-compat for the very short term - removing it before v2.20 is released, and mostly to avoid us having to rework PRs that are otherwise completed. It also provides a window for external PRs to pop up without problems.

matthchr: Fine with that too as long as the end result is the simpler variant (not both in perpetuity)

docs/hugo/content/design/ADR-2026-04-Secrets/_index.html
matthchr: TBH I'd be fine changing them both to something else too. Just feels like they should be consistent between them is all.

matthchr (approved):

PR #5304, opened 2026-04-10
Improve CRD management performance

### LINKED ISSUE #4574: Improve management of CRDs

CRDs are managed on pod start. There's interplay with the lease, which can result in the pod calling the List CRDs API multiple times. We currently compare CRDs by doing a `reflect.DeepEqual` on the whole CRD.

## Describe the improvement

Listing the CRDs multiple times (or applying them multiple times) can have memory usage and CPU impact on both the ASO pods as well as APIServer. We really want to limit the lists and applies that we do.

See for example https://github.com/Azure/azure-service-operator/pull/4573/files, which is related to this.

We should consider:
- [ ] Breaking CRD management out into an init container that runs before the main container launches (though managing leases across these might be annoying).
- [ ] Using the [PartialObjectMetadata](https://pkg.go.dev/k8s.io/apimachinery@v0.32.1/pkg/apis/meta/v1#PartialObjectMetadata) APIs and relying on an annotation or label for diffing.
- [ ] Using the [unique](https://pkg.go.dev/unique) package to somehow reduce memory usage of the lists we do by reducing duplication of data.

## Additional context

_Add any other context about the suggested improvement._

matthchr:
Fixes #4574.

- [x] this PR contains tests


v2/internal/crdmanagement/manager.go
```diff
@@ -178,18 +180,26 @@ func (m *Manager) ListCRDs(ctx context.Context, list *apiextensions.CustomResour
 
 	for _, crd := range list.Items {
 		m.logger.V(Verbose).Info("Found an existing CRD", "CRD", crd.Name)
+		// manually clear the spec.versions[].schema.openAPIV3Schema fields because they're large and we don't use them
+		for i := range crd.Spec.Versions {
+			crd.Spec.Versions[i].Schema.OpenAPIV3Schema = nil
+		}
```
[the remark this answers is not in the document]
matthchr: Good catch, fixed.

matthchr:
**Without changes:**
metrics_StartUp_20260305_1821_memory
metrics_StartUp_20260305_1821_cpu

**With changes:**
metrics_StartUp_20260410_1848_cpu
metrics_StartUp_20260410_1848_memory


Actual startup time for full set of CRDs is about the same if they're already installed (30s -> 30s). But memory is down ~120mb (26% decrease) and CPU usage is down a ton from pegged at basically 500mcpu (pod limit) to 140mcpu (72% decrease)

v2/internal/crdmanagement/helpers_test.go
```diff
@@ -88,6 +89,52 @@ func Test_FiveCRDsReady_AllOthersAreFiltered(t *testing.T) {
 	g.Expect(objs).To(HaveLen(5))
 }
 
+// This test requires that the task target `bundle-crds` has been run
```
theunrepentantgeek, line 94: Can you add a guard clause at the start of the test to fail it if `bundle-crds` hasn't been run?

That would give a clear failure to the user rather than something they may try to troubleshoot.

v2/internal/crdmanagement/helpers_test.go
```diff
@@ -88,6 +89,52 @@ func Test_FiveCRDsReady_AllOthersAreFiltered(t *testing.T) {
...
+	knownTypes, err := testData.getKnownStorageTypes()
+	g.Expect(err).ToNot(HaveOccurred())
+
+	// Filter the types to register
+	objs, err := crdmanagement.FilterStorageTypesByReadyCRDs(testData.logger, testData.s.GetScheme(), readyResources, knownTypes)
+	g.Expect(err).ToNot(HaveOccurred())
+	g.Expect(objs).To(HaveLen(1))
+}
+
+// This test requires that the task target `bundle-crds` has been run
```
theunrepentantgeek, line 118: Similar guard clause here.

theunrepentantgeek (commented):
A few suggestions, but otherwise looks good.

v2/internal/crdmanagement/helpers_test.go
matthchr: Yes we can probably count the files in out/crds and fail early if there are less than 50 or something

matthchr: Done

v2/internal/crdmanagement/helpers_test.go
matthchr: Done (and elsewhere)

theunrepentantgeek (approved):

PR #5337, opened 2026-04-27
Rename PDB helm chart template file name

boomer41:
## What this PR does

Other templates do not contain any whitespace in their file names. It seems that this whitespace was appended by mistake. Remove it to align with the rest of the files.

## Checklist

Nothing applied.


matthchr (approved):

matthchr:
Good catch! No idea where that came from

PR #5413, opened 2026-05-28
Add support for secret + configmap annotations/labels

### LINKED ISSUE #1398: Feature: Secrets created by ASO should have configurable annotations
As originally requested in #1396, @EdYa was looking for a way to allow another namespace access to the secret created by a `PostgreSQLUser`.
>  I could imagine that annotation support for secrets could be quite helpful. If we could add some annotations for kubernetes-reflector then we could mirror the secrets to the correct namespace.

An example of what this might look like:
```
apiVersion: azure.microsoft.com/v1alpha1
kind: PostgreSQLUser
metadata:
  name: psqluser
spec:
  server: psqlserver
  dbName: psqldb
  resourceGroup: myrg
  secretConfig:
    metadata:
      annotations:
        my-annotation: "foo"
```

We have other fields which are often applicable for secrets as well, such as `KeyVaultToStoreSecrets`. It's possible that it would make sense to move this value into this `secretConfig` at some point in the future as well.

matthchr:
* Webhooks reject conflicting label/annotation writes.
* Added tests.

Fixes #1398

## Checklist

- [x] this PR contains tests


plans/issue-1398-configurable-secret-annotations.md
```diff
@@ -0,0 +1,156 @@
+# Plan: Configurable Annotations/Labels on Secrets (Issue #1398)
```
matthchr: Will remove this but left it in case it was useful for review.

v2/pkg/genruntime/merger/merger.go
```diff
@@ -63,16 +63,20 @@ func mergeSecrets(namespace string, s []*v1.Secret) ([]*v1.Secret, error) {
 		for key, value := range secret.StringData {
 			collector.AddValue(
 				&genruntime.SecretDestination{
-					Name: secret.Name,
-					Key:  key,
+					Name:        secret.Name,
+					Key:         key,
+					Annotations: secret.Annotations,
+					Labels:      secret.Labels,
```
theunrepentantgeek, line 71: Is it safe to reuse these maps, or should we clone them to avoid alias effects?

v2/pkg/genruntime/merger/merger.go
```diff
@@ -63,16 +63,20 @@ func mergeSecrets(namespace string, s []*v1.Secret) ([]*v1.Secret, error) {
...
 		}
 		for key, value := range secret.Data {
 			collector.AddBinaryValue(
 				&genruntime.SecretDestination{
-					Name: secret.Name,
-					Key:  key,
+					Name:        secret.Name,
+					Key:         key,
+					Annotations: secret.Annotations,
+					Labels:      secret.Labels,
```
theunrepentantgeek, line 83: Ditto here, and elsewhere.

theunrepentantgeek (approved):
A couple minor comments, but looks good to me.

v2/pkg/genruntime/merger/merger.go
matthchr: Good question.

So, the objects being merged in this context are generated from the various exporters such as `autoGeneratedConfigExporter`, `manualConfigExporter`, `configMapExpressionExporter`, and `secretExpressionExporter`.

Each of these exporters returns a set of additional secrets that it has fabricated, and those secrets and then joined here in the merge step. That means nothing else is using those secrets or those maps, or should be writing to them.

It means, it should be safe to just copy them as-is. I will add some comments talking about this here to make that clear.

matthchr: Also the maps aren't actually modified they are just read in the collector (it constructs a new merged map)

theunrepentantgeek (approved):

theunrepentantgeek (approved):

PR #5432, opened 2026-06-03
[asoctl] Fix import failure due to problems looking for Quota resources

theunrepentantgeek:
## What this PR does

Fix for failure of `ascoctl` to handle errors returned when attempting to GET Quota resources.

### Special notes

Instead of just adding the new error codes to the silence list, I've also added a specific check to silence all errors encountered when trying to load extension resources, which should preclude this bug from happening again.

gif


matthchr (approved):
LGTM - the extension check sounds reasonable especially in the recursive context.

PR #5436, opened 2026-06-04
[asoctl] Only include ManagedIdentityCredential if IMDS is available

theunrepentantgeek:
## What this PR does

Opts out of including `ManagedIdentityCredential` in the credential chain for `asoctl` if IMDS is unavailable.

### Special notes

When calling `azidentity.NewDefaultAzureCredential()`, any instance of `ManagedIdentityCredential` that's created gets an internal flag set if there are other credential types in the chain:

``` go
		o := &ManagedIdentityCredentialOptions{
			ClientOptions: options.ClientOptions,
			// enable special DefaultAzureCredential behavior (IMDS probing) only when the chain contains another credential
			dac: selected^managedIdentity != 0,
		}
```
(See [`default_azure_credential.go#171`](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/default_azure_credential.go#L171))

With this flag true, the `ManagedIdentityCredential` probes IMDS for availability before requesting a token, ensuring it doesn't block the chain if another credential should be used.

Unfortunately, the flag is internal, there's no API to set it - and we need to avoid calling `NewDefaultAzureCredential()`.

So, we do the probe ourselves, and only include the MI Cred in the chain if IMDS is available.

gif

### DIFF
--- a/v2/cmd/asoctl/cmd/import_azure_resource.go
+++ b/v2/cmd/asoctl/cmd/import_azure_resource.go
@@ -328,7 +328,8 @@ func isIMDSAvailable(ctx context.Context) bool {
 		probeCtx,
 		http.MethodGet,
 		"http://169.254.169.254/metadata/identity/oauth2/token",
-		nil)
+		nil,
+	)
 	if err != nil {
 		return false
 	}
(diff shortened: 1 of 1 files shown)


matthchr (approved):
Copilot has a comment that seems reasonable.

Rest LGTM

matthchr (approved):

Issue #81, opened 2019-08-14
Split helper function unit tests into 2 files


WilliamMortlMicrosoft:
#24 

Split apart the API endpoint and API Mgmt Svc unit tests

Take resource group & API Mgmt Svc instance names from environment variables for API endpoint tests

PR #5448, opened 2026-06-05
fix: pin container images by digest in stolostron Dockerfile

RadekCap:
## Summary
- Pin `ubi9/go-toolset` and `ubi9/ubi` base images to sha256 manifest list digests in `stolostron/Dockerfile.stolostron`
- Resolves Scorecard Pinned-Dependencies alerts #80 and #81

## Context
OpenSSF Scorecard flagged unpinned container images as a supply-chain risk. Pinning to digest ensures builds are reproducible and immune to tag mutation.

Alert #70 (`FROM toolchain AS builder`) was dismissed as a false positive — it's a multi-stage build alias.

## Test plan
- [ ] Verify Konflux builds succeed with the pinned images
- [ ] Confirm Scorecard alerts #80 and #81 close after merge


RadekCap:
Opened against wrong repo — should target stolostron/azure-service-operator.

Issue #11, opened 2019-07-25
As a user, I have documentation that helps me deploy the operator and use it


Azadehkhojandi:
This user story covers the following:
Creation of README for deploying the operator, using it to deploy Azure services and running tests.

**Acceptance Criteria**
[Update this once the owner starts looking at the user story]

- [ ] Documentation covers instructions to deploy the operator based on the public docker image released.

- [ ] Documentation should cover instructions to setup their environment to build the code

- [ ] Documentation should cover how to deploy/delete//update Azure resources once the CRDs are installed (on Operator install) with example YAMLs and information on how to store the environment variables

- [ ] Documentation should cover how to run unit tests/integration tests

**Tasks**
Reference this issue when you create tasks for this user story

Issue #10, opened 2019-07-25
code clean up


Azadehkhojandi:
clean up the current repo, removing customer-specific details (namespaces,…) and send pr

PR #5462, opened 2026-06-11
feat: Add Owners and Members fields to entra SecurityGroup spec

### LINKED ISSUE #5669: Bug: Asymmetric Microsoft.Graph API blocking owner support for Entra SecurityGroup
## Describe the bug

The v1.0 API for Microsoft.Graph is asymmetric:

* POST `/v1.0/groups/{id}/owners/$ref` supports both users and service principals.
* GET `/v1.0/groups/{id}/owners` currently does _**not**_ return service principals.

This is causing problems with the goal-seeking behaviour desired for `SecurityGroup` as there's no good way to list all the ServicePrincipals on a group.

Blocking #5462 

The beta SDK does apparently work but is not supported for production use.

theunrepentantgeek:
This pull request introduces major improvements to how owners and members are specified and validated for Entra Security Groups, with a focus on supporting both direct object IDs and references to ConfigMap values. It adds robust validation logic, new helper methods for resolving and deduplicating references, and comprehensive unit tests to ensure correctness. Additionally, webhook validation is enhanced to enforce correct usage of optional ConfigMap references.

**Security Group Owners/Members Enhancements:**

* Added `Owners` and `Members` fields to `SecurityGroupSpec`, allowing directory objects to be specified either by direct object ID or via a reference to a ConfigMap value. Includes validation for uniqueness, maximum size, and mutual exclusivity of reference fields. [[1]](diffhunk://#diff-b527c962b18f5c288632a3be3205e4990ab2e9b981ec7da6d350997398daed37R61) [[2]](diffhunk://#diff-b527c962b18f5c288632a3be3205e4990ab2e9b981ec7da6d350997398daed37R83-R228)
* Introduced the `SecurityGroupMemberReference` type with logic to resolve object IDs from either inline values or ConfigMaps, including validation of GUID format and error handling.
* Added methods to `SecurityGroupSpec` for resolving owner/member object IDs, ensuring deduplication even across mixed reference types, and for constructing Microsoft Graph `@odata.bind` URIs during group creation. [[1]](diffhunk://#diff-b527c962b18f5c288632a3be3205e4990ab2e9b981ec7da6d350997398daed37R83-R228) [[2]](diffhunk://#diff-b527c962b18f5c288632a3be3205e4990ab2e9b981ec7da6d350997398daed37R357-R376)
* Added a helper function `DirectoryObjectRefURI` in `entra/id.go` to generate directory object URIs for Microsoft Graph.

**Validation and Webhook Improvements:**

* Added webhook validations to ensure that only one of `objectID` or `objectIDFromConfig` is set per reference, and that optional ConfigMap reference pairs are used correctly on create and update. [[1]](diffhunk://#diff-73a4460e989631d7655567869c213c87a2b76e075fdf015be7f339887717ac01L134-R138) [[2]](diffhunk://#diff-73a4460e989631d7655567869c213c87a2b76e075fdf015be7f339887717ac01L144-R162)

**Testing:**

* Introduced comprehensive tests in `securitygroup_types_test.go` to cover all new logic, including correct population of `@odata.bind` fields, error handling for missing/duplicate references, and order preservation.

**Other:**

* Updated the allowed values for `SecurityGroupMembershipType` to include "dynamic" types.

These changes significantly improve the flexibility, safety, and test coverage of Security Group owner/member assignment and validation logic.

Closes #5669


v2/api/entra/v1/webhook/securitygroup_webhook_types_test.go
```diff
@@ -0,0 +1,276 @@
...
+				},
+			}},
+		},
+	}
+
+	_, err := (&SecurityGroup_Webhook{}).validateOptionalConfigMapReferences(context.Background(), obj)
+	g.Expect(err).To(HaveOccurred())
+	g.Expect(err.Error()).To(ContainSubstring("cannot specify both"))
+	g.Expect(err.Error()).To(ContainSubstring("ObjectID"))
+	g.Expect(err.Error()).To(ContainSubstring("ObjectIDFromConfig"))
```
theunrepentantgeek: Should use `MatchError()` here.

v2/api/entra/v1/webhook/securitygroup_webhook_types_test.go
```diff
@@ -0,0 +1,276 @@
...
+			},
+		},
+	}
+
+	warnings, err := (&SecurityGroup_Webhook{}).validateOptionalConfigMapReferences(context.Background(), obj)
+	g.Expect(err).ToNot(HaveOccurred())
+	g.Expect(warnings).To(BeNil())
+}
+
+func TestSecurityGroup_Validation_AdmissionSchemaUniqueness(t *testing.T) {
```
theunrepentantgeek: This one test is checking three independent conditions - and should be separated into three separate tests.

v2/api/entra/v1/securitygroup_types.go
```diff
@@ -76,13 +79,134 @@ type SecurityGroupSpec struct {
 
 	// IsAssignableToRole: Indicates whether the group can be assigned to a role.
 	IsAssignableToRole *bool `json:"isAssignableToRole,omitempty"`
+
+	// Owners: Directory objects (users, service principals, groups) to assign as owners of the security
+	// group at creation time. Applied only during the initial POST to Microsoft Graph via `owners@odata.bind`
+	// and used as desired owner state for reconciliation. Required when ASO authenticates with an app-only token and the calling
```
theunrepentantgeek: This comment needs to be updated, as we are now updating this list during follow-up reconciliation passes too.

v2/api/entra/v1/securitygroup_types.go
```diff
@@ -76,13 +79,134 @@ type SecurityGroupSpec struct {
...
+	// principal lacks Group.ReadWrite.All — otherwise the created group has no owners and is unmanageable.
+	// +kubebuilder:validation:MaxItems=20
+	// +kubebuilder:validation:XValidation:rule="size(self.filter(x, has(x.objectID)).map(x, x.objectID).distinct()) == size(self.filter(x, has(x.objectID)))",message="owners must be unique"
+	// +kubebuilder:validation:XValidation:rule="size(self.filter(x, has(x.objectIDFromConfig)).map(x, x.objectIDFromConfig).distinct()) == size(self.filter(x, has(x.objectIDFromConfig)))",message="owners must be unique"
+	Owners []SecurityGroupMemberReference `json:"owners,omitempty"`
+
+	// Members: Directory objects (users, service principals, groups) to assign as members of the security
+	// group at creation time. Applied only during the initial POST to Microsoft Graph via `members@odata.bind`
+	// and used as desired member state for reconciliation.
```
theunrepentantgeek: Another comment that needs updating.

v2/api/entra/v1/securitygroup_types_test.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	group := msgraphmodels.NewGroup()
+	err := spec.AssignODataBindOnCreate(group, genruntime.MakeResolved[genruntime.ConfigMapReference, string](nil))
+	g.Expect(err).ToNot(HaveOccurred())
+
+	additional := group.GetAdditionalData()
+	g.Expect(additional).To(HaveKey("owners@odata.bind"))
+	g.Expect(additional).To(HaveKey("members@odata.bind"))
+	g.Expect(additional["owners@odata.bind"]).To(Equal([]string{"https://graph.microsoft.com/v1.0/directoryObjects/11111111-1111-1111-1111-111111111111"}))
```
theunrepentantgeek: Recommend formatting the nested array across multiple lines:

```suggestion
	g.Expect(additional["owners@odata.bind"]).To(Equal(
		[]string{
			"https://graph.microsoft.com/v1.0/directoryObjects/11111111-1111-1111-1111-111111111111",
		}))
```

v2/api/entra/v1/securitygroup_types_test.go
```diff
@@ -0,0 +1,138 @@
...
+	g.Expect(additional["members@odata.bind"]).To(Equal([]string{"https://graph.microsoft.com/v1.0/directoryObjects/22222222-2222-2222-2222-222222222222"}))
```
theunrepentantgeek: Ditto here.

v2/api/entra/v1/securitygroup_types_test.go
```diff
@@ -0,0 +1,138 @@
...
+		}},
+	}
+
+	group := msgraphmodels.NewGroup()
+	err := spec.AssignODataBindOnCreate(group, resolved)
+	g.Expect(err).ToNot(HaveOccurred())
+
+	additional := group.GetAdditionalData()
+	g.Expect(additional["owners@odata.bind"]).To(Equal([]string{"https://graph.microsoft.com/v1.0/directoryObjects/aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa"}))
+	g.Expect(additional["members@odata.bind"]).To(Equal([]string{"https://graph.microsoft.com/v1.0/directoryObjects/bbbbbbbb-bbbb-bbbb-bbbb-bbbbbbbbbbbb"}))
```
theunrepentantgeek: Format these inline arrays across multiple lines too.

v2/internal/reconcilers/entra/entra_securitygroup_retry.go
```diff
@@ -0,0 +1,131 @@
...
+		result := time.Until(retryAfterTime)
+		if result > 0 {
+			return result
+		}
+	}
+
+	return 0
+}
+
+func parseHTTPDate(s string) (time.Time, error) {
```
theunrepentantgeek: This seems like a function that should already exist elsewhere - handling a `Retry-After` header isn't unusual.

v2/api/entra/v1/securitygroup_types.go
```diff
@@ -76,13 +79,134 @@ type SecurityGroupSpec struct {
 
 	// IsAssignableToRole: Indicates whether the group can be assigned to a role.
 	IsAssignableToRole *bool `json:"isAssignableToRole,omitempty"`
+
+	// Owners: Directory objects (users, service principals, groups) to assign as owners of the security
```
theunrepentantgeek: **P0 · Finding #1 · Immutability contract inconsistent with code and comments**

The PR body states *"Immutable after creation: CEL XValidation rule (`self == oldSelf`) enforced by webhook"* and this field comment says *"Applied only during the initial POST"* — but grep for `self == oldSelf` in the diff returns zero, and `reconcileOwnersAndMembers` runs on every reconcile pass (see `entra_securitygroup_reconciler.go:208`), continuously mutating owners/members after creation.

**Why it matters:** users trusting the field docs will assume day-2 spec edits are ignored. They aren't — every reconcile pass will add/remove owners/members to converge on the current spec. Combined with the CRD tolerating `spec.owners: []`, a stray spec edit can silently strip every owner of an already-adopted group (see finding #3).

**Response — design call.** Either:
- (a) add the missing CEL `self == oldSelf` rule and remove `reconcileOwnersAndMembers` from the update path (matches PR body), OR
- (b) rewrite PR body + field comments to accurately describe continuous reconciliation, and add a guard that treats present-but-empty as no-op when the group is already adopted.

_Reviewers: api-contract, security, adversarial, fast-pass · confidence 100_

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
```diff
@@ -204,11 +205,193 @@ func (r *EntraSecurityGroupReconciler) update(
...
+				log,
+			)
+			if err != nil {
+				sideErrors = append(sideErrors, eris.Wrapf(err, "reconciling members for group %s", id))
+			}
+		}
+	}
+
+	if len(sideErrors) > 0 {
+		return ctrl.Result{}, errors.Join(sideErrors...)
```
theunrepentantgeek: **P1 · Finding #2 · Joined Graph errors bypass throttle classification** — strongest-signal finding

`reconcileOwnersAndMembers` returns `errors.Join(sideErrors...)`; `classifyRelationshipError` calls `asODataError`, which uses `eris.As` on the joined error. `eris.As` walks `Unwrap() error` but **not** `Unwrap() []error`, so both nested OData 429 (Retry-After) and 403 (permission) errors are lost.

**Why it matters:** under Graph throttling, the controller ignores the server's Retry-After header and re-hammers Graph on the default exponential backoff — the exact regression the new retry classifier was written to prevent. A permission drop on one side also fails to surface as `GraphPermissionDenied`.

**Fix:** rewrite `asODataError` in `entra_securitygroup_retry.go` to walk both `Unwrap() error` and `Unwrap() []error` recursively. Then have `tryThrottleRequeue` take the **max** positive Retry-After across all nested 429s, and `isPermissionError` return true if **any** nested error is 403. Add a regression test that feeds `errors.Join(eris.Wrap(makeODataError(429, "10")))` into `classifyRelationshipError`.

Same helper fix also resolves finding #10 (partial-success semantics).

_Reviewers: correctness, adversarial (Claude), adversarial-gpt5.6-sol (independent cross-model), reliability, fast-pass · confidence 100_

theunrepentantgeek (commented):
## Code review — 11-reviewer parallel pass with cross-model adversarial peer

Ran the full ce-code-review roster (correctness, testing, maintainability, project-standards, agent-native, learnings) plus conditional security, reliability, api-contract, adversarial — and an **independent cross-model adversarial pass on `gpt-5.6-sol`** (different provider than the Claude host, for genuine independence on the highest-risk lens).

**Verdict: Not ready.** The blocker is finding #1 — the PR body's immutability contract is a promise the code doesn't keep, and #3 shows the current continuous-reconcile behavior is genuinely unsafe on adopted groups. Pick the semantics, then #2 (errors.Join throttle bypass — corroborated by 4 independent reviewers including the GPT peer) needs to land in the same or an immediate follow-up commit; the retry classifier introduced here doesn't actually work as designed today.

Findings posted as individual line comments below. **Finding #6 (OData injection via unescaped displayName)** targets a pre-existing line outside this PR's diff — posted separately as a file-scoped comment.

### Actionable order

1. **Decide immutability (#1)** — the answer resolves #3.
2. **Fix `asODataError` to walk joined error trees** — #2 + #10 in one helper.
3. **Close reference-validation gaps** — #4 (combined ≤20 cap), #7 (GUID validation on ConfigMap values), #8 (exactly-one-of CEL rule).
4. **Harden pagination** — #5 (iteration bound) + #9 (ctx cancellation), same 25-line loop.
5. **Escape displayName (#6)** and **normalize Retry-After timezone (#11)** — small, independent.
6. **Style/dedup cleanup (#13)** — after everything else.

### Housekeeping notes

- **project-standards P1** flagged the new `SecurityGroupMemberReference` entries in `zz_generated.deepcopy.go` (lines 78-100, 182-192) as possibly hand-written. Direct-verified the pattern matches idiomatic controller-gen output for pointer/slice fields — most likely a false positive, but **please confirm by running `task controller:generate-types` before merge** and diffing the result.
- **Two fast-pass items withdrawn** for lack of corroboration: (a) duplicate-URI dedup gap on the create path — a ConfigMap resolving to the same GUID as a sibling `objectID` still POSTs duplicate URIs today; worth checking whether #7's fix incidentally covers it; (b) `directoryObjectIDFromRef` trailing-slash edge.
- **One trivial fix already applied** to my working tree (not pushed): `fmt.Errorf` → `eris.Wrapf` in `securitygroup_relationships.go:83, 89` to match the eris convention used everywhere else in the entra reconciler. Build + `go test ./internal/reconcilers/entra/...` clean.

### Testing gaps (union across reviewers)

- No e2e reconcile covering `create-with-inline-@odata.bind` → PATCH → converged reconcile.
- No test for `errors.Join(eris.Wrap(makeODataError(...)))` through `classifyRelationshipError` — the bug in #2 is invisible to current tests.
- No test for 403 on one side proving the other side still reconciles.
- No Retry-After edge inputs: negative integer, past HTTP-date, empty string, non-GMT timezone (#11).
- No pagination cycle test / duplicate `nextLink` test (#5, #9).
- No admission test for cross-field-same-GUID (`objectID:X` + `objectIDFromConfig` resolving to `X`) — the CEL uniqueness rules check each field in isolation.
- No test for a non-GUID value in a resolved ConfigMap (#7).
- No test for `spec.owners: []` on an already-adopted group (#3).

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
```diff
@@ -204,11 +205,193 @@ func (r *EntraSecurityGroupReconciler) update(
...
+	}
+
 	return ctrl.Result{}, nil
 }
 
+func relationshipSidesToManage(spec asoentra.SecurityGroupSpec) (bool, bool) {
+	// Nil means omitted (unmanaged); explicit empty means managed-to-empty.
+	return spec.Owners != nil, spec.Members != nil
```
theunrepentantgeek: **P1 · Finding #3 · Present-but-empty owners/members strips Entra state** — dual-adversarial corroborated

`relationshipSidesToManage` deliberately distinguishes nil (omitted, unmanaged) from `[]` (managed to empty). Combined with the continuous-reconcile behavior called out in finding #1, editing an adopted group's spec to `owners: []` will remove every current owner. Graph also rejects removing the last owner, which will then loop as retryable errors.

**Why it matters:** for an adopted group with real owners, a spec edit to `owners: []` is either silent data loss or a permanent reconciler stall — depending on Graph's response. Both are surprising given the field docs.

**Response depends on finding #1.** If immutability wins, moot. If continuous reconcile stays, reject present-but-empty via CEL when the group is already adopted, or surface a terminal condition instead of retrying forever on Graph's last-owner rule.

_Reviewers: adversarial (Claude), adversarial-gpt5.6-sol · confidence 100 (dual-adversarial promoted)_

v2/api/entra/v1/securitygroup_types.go
```diff
@@ -76,13 +79,134 @@ type SecurityGroupSpec struct {
...
+	// group at creation time. Applied only during the initial POST to Microsoft Graph via `owners@odata.bind`
+	// and used as desired owner state for reconciliation. Required when ASO authenticates with an app-only token and the calling
+	// principal lacks Group.ReadWrite.All — otherwise the created group has no owners and is unmanageable.
+	// +kubebuilder:validation:MaxItems=20
```
theunrepentantgeek, line 88: **P1 · Finding #4 · Combined owners+members can exceed Graph's 20-item POST cap**

`MaxItems=20` is enforced per field, but Graph caps `owners@odata.bind + members@odata.bind` **combined** on a single POST at 20 items. A spec with 15 owners and 15 members passes admission and then fails at create time.

**Fix:** add a spec-level CEL rule `size(self.owners) + size(self.members) <= 20`, or split overflow references into post-create Graph add-calls.

_Reviewers: adversarial-gpt5.6-sol · confidence 75_

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
```diff
@@ -204,11 +205,193 @@ func (r *EntraSecurityGroupReconciler) update(
...
+	firstPage func(context.Context) (msgraphmodels.StringCollectionResponseable, error),
+	nextPage func(string) (msgraphmodels.StringCollectionResponseable, error),
+) ([]string, error) {
+	response, err := firstPage(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	result := make([]string, 0)
+	for response != nil {
```
theunrepentantgeek: **P1 · Finding #5 · Pagination loop unbounded on cyclic/malformed nextLink**

`for response != nil` follows `nextLink` without a max-iteration bound or cycle detection. A malformed/malicious `@odata.nextLink` that cycles or is missing termination will loop until OOM. The `nextLink` field is controlled by the remote service.

**Fix:** add a max-iteration counter (~100) and return a clear error when exceeded. Fix together with finding #9 (context cancellation) — same 25-line loop.

_Reviewers: reliability · confidence 75_

v2/api/entra/v1/securitygroup_types.go
```diff
@@ -76,13 +79,134 @@ type SecurityGroupSpec struct {
...
+	case ref.ObjectID != nil:
+		return *ref.ObjectID, nil
+
+	case ref.ObjectIDFromConfig != nil:
+		val, err := resolved.Lookup(*ref.ObjectIDFromConfig)
+		if err != nil {
+			return "", fmt.Errorf("failed resolving objectIDFromConfig: %w", err)
+		}
+
+		return val, nil
```
theunrepentantgeek: **P2 · Finding #7 · ConfigMap-resolved object IDs bypass GUID validation**

The GUID pattern is enforced on `spec.owners[].objectID` via kubebuilder, but bypassed when the value comes from `objectIDFromConfig` — `ResolveObjectID` returns whatever text is in the ConfigMap.

**Why it matters:** `graphDirectoryObjectURI` then builds a URI from arbitrary text; at best you get a 400 from Graph, at worst an unintended target ID.

**Fix:** apply the same GUID regex to the resolved value before returning; reject on mismatch with a Ready condition message.

_Reviewers: security, adversarial-gpt5.6-sol · confidence 100 (cross-reviewer promoted)_

v2/api/entra/v1/securitygroup_types.go
```diff
@@ -76,13 +79,134 @@ type SecurityGroupSpec struct {
...
+
+	default:
+		return "", fmt.Errorf("missing objectID or objectIDFromConfig")
```
theunrepentantgeek: **P2 · Finding #8 · Empty reference entries pass admission then retry forever**

The runtime error `"missing objectID or objectIDFromConfig"` fires when a reference entry has neither field set, but the CRD's CEL rules don't cover this — admission accepts `[{}]`. The reconciler then loops on the runtime error indefinitely.

**Fix:** add a CEL rule requiring exactly one of `objectID` / `objectIDFromConfig` per entry, on both `owners` and `members`.

_Reviewers: adversarial-gpt5.6-sol · confidence 100_

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
theunrepentantgeek: **P2 · Finding #9 · No context cancellation check in pagination loop**

Same loop as finding #5. `ctx.Err()` is never checked between pages — operator shutdown / user delete / context timeout is silently ignored until the loop finishes or errors.

**Fix:** add `if err := ctx.Err(); err != nil { return nil, err }` at the top of each iteration. Fix together with #5.

_Reviewers: reliability · confidence 100_

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
theunrepentantgeek: **P2 · Finding #10 · errors.Join partial-success hides successful side**

Same `errors.Join(sideErrors...)` as finding #2, different concern: when one side fails and the other succeeds, the whole reconcile returns error and the succeeded side isn't reflected in `Status` — so the next pass re-does successful work, wasting Graph quota and slowing convergence.

**Fix:** track partial success in Status (separate conditions for owners and members), or update `Status` for the succeeded side before returning the aggregated error. Same root cause as #2 — fix together.

_Reviewers: reliability · confidence 75_

v2/internal/reconcilers/entra/entra_securitygroup_retry.go
theunrepentantgeek: **P2 · Finding #11 · Retry-After HTTP-date parsing ignores timezone requirement**

The three `time.Parse` layouts use `MST`. `time.Parse` treats unknown zone abbreviations by returning a `Time` with a fabricated zero offset — so a header like `Sun, 06 Nov 1994 08:49:37 EST` parses to an ambiguous timestamp, and `time.Until` can silently return the wrong duration, dropping the throttle signal. RFC 7231 mandates GMT.

**Fix:** after `parseHTTPDate` succeeds, verify the returned location is UTC/GMT (or use `time.ParseInLocation(..., time.UTC)`). Log and return zero on ambiguous input.

_Reviewers: reliability · confidence 75_

v2/internal/reconcilers/entra/securitygroup_relationships.go
```diff
@@ -0,0 +1,129 @@
...
+		"Reconciled relationship side",
+		"side", side,
+		"added", len(delta.ToAdd),
+		"removed", len(delta.ToRemove),
+	)
+
+	return nil
+}
+
+func directoryObjectRefURI(objectID string) string {
```
theunrepentantgeek: **P2 · Finding #13 · Duplicate directoryObject URI helpers** (advisory)

`directoryObjectRefURI` here and `graphDirectoryObjectURI` at `v2/api/entra/v1/securitygroup_types.go:340` have byte-identical bodies in two packages.

**Fix:** promote a single shared helper into `v2/internal/reconcilers/entra` and drop the v1 copy — the reverse direction (api importing internal reconciler) isn't acceptable. Or leave as-is and accept the drift risk. Advisory.

_Reviewers: maintainability, fast-pass · confidence 75_

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
theunrepentantgeek, line 1: **P1 · Finding #6 · OData injection via unescaped displayName** — `entra_securitygroup_reconciler.go:581`

```go
filterStr := fmt.Sprintf("displayName eq '%s'", displayName)
```

The line predates this PR (last changed 2025-06-18), so it's posted here as file-scoped rather than as a review line comment. It's on the adoption path this PR exercises, which is why the security reviewer flagged it now.

`displayName` is user-controlled via the CR spec; a value like `x' or 1 eq 1--` alters the OData filter, a bare quote breaks it. In the **adoption** path this means a crafted displayName could match another group and cause ASO to adopt an object it doesn't own. Kubernetes RBAC on the CR does not stop this — the Graph query runs with the operator's app credentials.

**Fix:** escape single quotes (`'` → `''`) before interpolation, or use an SDK helper that constructs OData string literals safely. Add a regression test for a displayName containing a quote.

_Reviewer: security · confidence 75 · pre-existing but tightly coupled to this PR's adoption path_

v2/api/entra/v1/securitygroup_types.go
theunrepentantgeek: Done.

v2/api/entra/v1/webhook/securitygroup_webhook_types_test.go
theunrepentantgeek: Split.

v2/api/entra/v1/webhook/securitygroup_webhook_types_test.go
theunrepentantgeek: Changed.

v2/api/entra/v1/securitygroup_types.go
theunrepentantgeek: Changed.

v2/api/entra/v1/securitygroup_types_test.go
theunrepentantgeek: Done.

v2/api/entra/v1/securitygroup_types.go
theunrepentantgeek: Not immutable; updated comments.

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
theunrepentantgeek: Done.

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
theunrepentantgeek: ASO resource is the one-source-of-truth, if the user sets the group to empty, that's on them.

v2/api/entra/v1/securitygroup_types.go
theunrepentantgeek: Done.

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
theunrepentantgeek: Done.

v2/api/entra/v1/securitygroup_types.go
theunrepentantgeek: Done.

v2/api/entra/v1/securitygroup_types.go
theunrepentantgeek: Done.

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
theunrepentantgeek: Eventual consistency is fine.

v2/internal/reconcilers/entra/securitygroup_relationships.go
theunrepentantgeek: Done.

v2/internal/reconcilers/entra/entra_securitygroup_retry.go
theunrepentantgeek: Won't fix.

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
theunrepentantgeek: Fixed.

v2/internal/util/interval/interval_calculator.go
```diff
@@ -109,7 +109,12 @@ func (i *calculator) NextInterval(req ctrl.Request, result ctrl.Result, err erro
 	return result, nil
 }
 
-func (i *calculator) failureResult(req ctrl.Request, err error) (ctrl.Result, error) {
+// failureResult computes the requeue interval for an error. If the caller supplies a non-zero
+// result.RequeueAfter (e.g. parsed from an HTTP Retry-After header), the returned interval is the
+// larger of the caller's value and the classification-based exponential backoff — throttling
+// signals from upstream services may slow us down but must never speed us up beyond our own
+// backoff schedule.
+func (i *calculator) failureResult(req ctrl.Request, result ctrl.Result, err error) (ctrl.Result, error) {
```
theunrepentantgeek: What do you think of this @matthchr - on the one hand, guaranteeing that we never work faster than `Retry-After` has merit, but I'm not sure about the signature here.

v2/api/entra/v1/securitygroup_types.go
```diff
@@ -212,6 +356,26 @@ type SecurityGroupOperatorConfigMaps struct {
...
+func graphDirectoryURIs(
+	references []SecurityGroupMemberReference,
+	fieldName string,
+	resolved genruntime.Resolved[genruntime.ConfigMapReference, string],
+) ([]string, error) {
+	uris := make([]string, 0, len(references))
+	for i, ref := range references {
+		id, err := ref.ResolveObjectID(resolved)
+		if err != nil {
+			return nil, fmt.Errorf("%s[%d]: %w", fieldName, i, err)
```
theunrepentantgeek: Should be using `eris.Wrapf()` here.

v2/api/entra/id.go
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+package entra
+
+func DirectoryObjectRefURI(objectID string) string {
+	return "https://graph.microsoft.com/v1.0/directoryObjects/" + objectID
```
theunrepentantgeek: `objectID` is externally controlled, so we shouldn't assume the value is URL safe - we should encode it before constructing the URI (such encoding will be a NOP for anything valid, but will protect us if something odd comes through).

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
```diff
@@ -23,7 +24,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
-	asoentra "github.com/Azure/azure-service-operator/v2/api/entra/v1"
+	asoentrav1 "github.com/Azure/azure-service-operator/v2/api/entra/v1"
```
theunrepentantgeek: Adding `v1` here adds no value, just use `asoentra` as the import alias.

v2/internal/reconcilers/entra/entra_securitygroup_reconciler_pagination_test.go
```diff
@@ -0,0 +1,155 @@
...
+				Owners:  []asoentra.SecurityGroupMemberReference{ownerRef},
+				Members: []asoentra.SecurityGroupMemberReference{memberRef},
+			},
+			wantOwners:  true,
+			wantMembers: true,
+		},
+	}
+
+	for name, tc := range cases {
+		tc := tc
```
theunrepentantgeek: Since Go 1.24, this capture line is not required - and will in fact fail linting.

v2/internal/reconcilers/entra/securitygroup_relationships.go
```diff
@@ -0,0 +1,320 @@
...
+}
+
+// planRelationshipDelta returns which ids need to be added or removed to move
+// current to desired. Both inputs are expected to be pre-deduplicated by their
+// caller (collectDirectoryObjectIDs for current, ResolveOwnerObjectIDs /
+// ResolveMemberObjectIDs for desired); duplicates in the inputs will appear
+// duplicated in the output.
+func planRelationshipDelta(current []string, desired []string) relationshipDelta {
+	currentSet := make(map[string]struct{}, len(current))
+	desiredSet := make(map[string]struct{}, len(desired))
```
theunrepentantgeek: We should be using the existing set datatype - see `internal/set`, which includes `.Except()` for differencing, and `.Values()` to return a slice at the end.

```suggestion
	currentSet := set.Make[string]()
	desiredSet := set.Make[string]()
```

theunrepentantgeek (commented):
@copilot a few minor pieces of feedback to address please.

theunrepentantgeek:
@copilot please update the PR description too.

v2/internal/reconcilers/entra/entra_securitygroup_retry.go
```diff
@@ -0,0 +1,131 @@
...
+// multiple independent sources should take the max of the resulting durations.
+func retryAfterResult(err error) ctrl.Result {
+	if err == nil {
+		return ctrl.Result{}
+	}
+
+	odataError, ok := errors.AsType[*odataerrors.ODataError](err)
+	if !ok {
+		return ctrl.Result{}
+	}
```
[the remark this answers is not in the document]
theunrepentantgeek: Incrrect; `AsType[]` is the preferred new API.

v2/api/entra/v1/securitygroup_types.go
```diff
@@ -55,6 +60,7 @@ type SecurityGroupList struct {
 	Items           []SecurityGroup `json:"items"`
 }
 
+// +kubebuilder:validation:XValidation:rule="(has(self.owners) ? size(self.owners) : 0) + (has(self.members) ? size(self.members) : 0) <= 20",message="the combined number of owners and members must not exceed 20"
```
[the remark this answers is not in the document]
theunrepentantgeek: The limit of 20 comes from Entra. #wontfix

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
```diff
@@ -204,6 +205,10 @@ func (r *EntraSecurityGroupReconciler) update(
 		return ctrl.Result{}, eris.Wrapf(err, "failed to update group %s", id)
 	}
 
+	if result, err := r.reconcileOwnersAndMembers(ctx, group, client.Client(), log); err != nil {
+		return classifyRelationshipError(result, err)
+	}
```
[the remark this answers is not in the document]
theunrepentantgeek: Updating the PR description.

v2/api/entra/v1/securitygroup_types.go
```diff
@@ -76,13 +80,152 @@ type SecurityGroupSpec struct {
...
+
+	// Owners: Directory objects (users, service principals, groups) to assign as owners of the security
+	// group at creation time. Applied during the initial POST to Microsoft Graph via `owners@odata.bind`
+	// and then used as desired owner state for later reconciliation.
+	// Required when ASO authenticates with an app-only token and the calling principal lacks Group.ReadWrite.All —
+	// otherwise the created group has no owners and is unmanageable.
+	// +kubebuilder:validation:MaxItems=20
+	// +kubebuilder:validation:XValidation:rule="size(self.filter(x, has(x.objectID)).map(x, x.objectID).distinct()) == size(self.filter(x, has(x.objectID)))",message="owners must be unique"
+	// +kubebuilder:validation:XValidation:rule="size(self.filter(x, has(x.objectIDFromConfig)).map(x, x.objectIDFromConfig).distinct()) == size(self.filter(x, has(x.objectIDFromConfig)))",message="owners must be unique"
+	Owners []SecurityGroupMemberReference `json:"owners,omitempty"`
```
matthchr, line 89: Both owners + members being `SecurityGroupMemberReference`. Is this convenience or reflected in the API as well? Wondering if we want different types to allow for separate evolution?

matthchr: The type does look pretty simple (just ObjectID + ObjectIDFromConfig)... so likely the reuse here is safe, and anyway if we wanted to change it we could since our guarantee is YAML compatability not object-name compatibility.

v2/api/entra/v1/securitygroup_types.go
```diff
@@ -76,13 +80,152 @@ type SecurityGroupSpec struct {
...
 
 // OriginalVersion returns the original API version used to create the resource.
 func (spec *SecurityGroupSpec) OriginalVersion() string {
 	return GroupVersion.Version
 }
 
+// AssignODataBindOnCreate sets the `owners@odata.bind` and `members@odata.bind` additional data
+// on the group model. These annotations are only valid during the initial POST to Microsoft Graph;
+// they must NOT be included in PATCH requests.
+// The typed setters (SetOwners/SetMembers) serialize as nested objects which Graph rejects on create —
```
matthchr, line 158: Ok so making sure I understand here: additional data on the group model for create (POST).

PATCH, use the SetOwners/SetMembers section, which performs updates.

v2/api/entra/v1/securitygroup_types.go
```diff
@@ -76,13 +80,152 @@ type SecurityGroupSpec struct {
...
+		additionalData["members@odata.bind"] = members
+	}
+
+	model.SetAdditionalData(additionalData)
+	return nil
+}
+
+// ResolveOwnerObjectIDs resolves owner object IDs, returning an error if two entries
+// resolve to the same identifier.
+func (spec *SecurityGroupSpec) ResolveOwnerObjectIDs(resolved genruntime.Resolved[genruntime.ConfigMapReference, string]) ([]string, error) {
```
matthchr: minor: not sold these two functions (ResolveOwnerObjectIDs and ResolveMemberObjectIDs) are adding a lot.

v2/api/entra/v1/securitygroup_types.go
```diff
@@ -160,8 +303,7 @@ func (status *SecurityGroupStatus) AssignFromGroup(model models.Groupable) {
 	}
 }
 
-// +kubebuilder:validation:Enum={"assigned","enabled","assignedm365","enabledm365"}
-// +kubebuilder:default=AdoptOrCreate
```
matthchr, line 164: Default here was just a bug?

v2/internal/reconcilers/entra/securitygroup_relationships.go
```diff
@@ -0,0 +1,299 @@
...
+
+	if len(sideErrors) > 0 {
+		return throttle, errors.Join(sideErrors...)
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// ownersSide adapts the msgraph SDK's owners endpoint into a relationshipSide.
+func ownersSide(
```
matthchr: minor, style:

```suggestion
func ownersRelationship(
```

Maybe?

and you could possibly just call `relationshipSide` `relationship`?

v2/internal/reconcilers/entra/securitygroup_relationships.go
```diff
@@ -0,0 +1,299 @@
...
+
+	for _, id := range delta.ToRemove {
+		if err := side.remove(ctx, id); err != nil {
+			return eris.Wrapf(err, "%s remove %s", side.name, id)
+		}
+	}
+
+	log.V(1).Info(
+		"Reconciled relationship side",
+		"side", side.name,
```
matthchr: minor: generally could possibly just call all of these relationships rather than relationship sides (in the docs, etc too)

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
matthchr, line 1: 🤖 **High — 404 is converted to `(nil, nil)`, but both callers dereference the nil group.** `loadGroupByID` returns `(nil, nil)` when the Entra group was deleted externally, so the `isNotFound(err)` recovery branch in `update()` is unreachable and `AssignToGroup(g)` panics. `UpdateStatus()` has the same issue when it calls `AssignFromGroup(groupable)`. Please either preserve the not-found error/sentinel or explicitly handle a nil group at both callers, and add a regression test for external deletion.

v2/api/entra/v1/securitygroup_types.go
```diff
@@ -212,6 +354,26 @@ type SecurityGroupOperatorConfigMaps struct {
...
+	resolved genruntime.Resolved[genruntime.ConfigMapReference, string],
+) ([]string, error) {
+	uris := make([]string, 0, len(references))
+	for i, ref := range references {
+		id, err := ref.ResolveObjectID(resolved)
+		if err != nil {
+			return nil, eris.Wrapf(err, "%s[%d]", fieldName, i)
+		}
+
+		uris = append(uris, asoentra.DirectoryObjectRefURI(id))
```
matthchr: 🤖 **Medium — create bypasses the mixed inline/ConfigMap duplicate check.** This helper resolves and appends every reference directly, while the update path uses `resolveMemberObjectIDs()` and rejects IDs that collide after ConfigMap resolution. The same desired state can therefore fail differently on create and update, with create sending duplicate `@odata.bind` values to Graph. Please resolve/deduplicate IDs through the shared path before mapping them to URIs, and add a create-path regression test.

v2/internal/reconcilers/entra/entra_securitygroup_retry.go
```diff
@@ -0,0 +1,131 @@
...
+	if b.RequeueAfter > a.RequeueAfter {
+		return b
+	}
+	return a
+}
+
+func isPermissionError(err error) bool {
+	// errors.AsType walks both `Unwrap() error` and `Unwrap() []error`, so it
+	// finds an ODataError nested inside any wrapping (including errors.Join).
+	odataError, ok := errors.AsType[*odataerrors.ODataError](err)
```
matthchr, line 72: 🤖 **Medium — this only classifies the first `ODataError` found in a joined error.** `errors.AsType` traverses `errors.Join`, but it returns one matching error; it does not answer whether any branch is a 403. For example, owners returning 429 followed by members returning 403 is classified as a generic fast retry rather than permission-denied/slow. Please walk all joined branches and return true when any nested OData error is 403, with a regression test where the 403 is not first.

v2/internal/reconcilers/entra/entra_securitygroup_retry.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	values := odataError.ResponseHeaders.Get("Retry-After")
+	if len(values) == 0 {
+		return 0, false
+	}
+
+	retryAfterStr := values[0]
+	if retryAfterVal, parseErr := strconv.ParseInt(retryAfterStr, 10, 64); parseErr == nil {
+		return time.Duration(retryAfterVal) * time.Second, true
```
matthchr, line 104: 🤖 **Low — `Retry-After` is unbounded and this multiplication can overflow.** The value flows through `reconcileOwnersAndMembers` into `interval.Calculator`, which takes `max(backoff, RequeueAfter)` and returns a nil error; `GenericReconciler` then returns it to controller-runtime, whose workqueue calls `AddAfter` with the raw duration and applies no maximum. A large valid value or far-future HTTP date can park the resource for years, while multiplication overflow can wrap unpredictably. Please reject negative values and clamp both numeric and date-based Retry-After values to a documented maximum.

v2/internal/reconcilers/entra/securitygroup_relationships.go
```diff
@@ -0,0 +1,299 @@
...
+				QueryParameters: &groups.ItemMembersRefRequestBuilderDeleteQueryParameters{
+					Id: &deleteID,
+				},
+			})
+		},
+	}
+}
+
+func relationshipSidesToManage(spec asoentrav1.SecurityGroupSpec) (bool, bool) {
+	// Nil means omitted (unmanaged); explicit empty means managed-to-empty.
```
matthchr: Is it actually possible to get explicit empty though, since we have `omitempty` on these fields? Certainly if we serialize the Go type it won't retain `[]`, though I guess APIServer isn't using our Go types and so will probably preserve `[]` to etcd... so maybe it is possible to get?

Do you have a test showing that works?

v2/internal/util/interval/interval_calculator.go
matthchr: We could encode the Retry-After in the `err`?

That seems to also have some advantages because we might process the error in a separate caller (as you do in `classifyRelationshipError`) might change the error and want the opportunity to manage the retry as part of that change?

Accomplishing this could possbily be as simple as just adding an optional `DelayHint` or similar field to `ReadyConditionImpactingError`.

v2/internal/util/interval/interval_calculator.go
```diff
@@ -197,3 +202,11 @@ func (i *calculator) calculateExponentialDelay(base time.Duration, exp int, max
 
 	return calculated
 }
+
+// maxDuration returns the larger of two durations.
+func maxDuration(a, b time.Duration) time.Duration {
```
matthchr: `max(delay, result.RequeueAfter)`

Should work I think? Not sure we need this guy.

v2/internal/reconcilers/entra/entra_securitygroup_retry.go
```diff
@@ -0,0 +1,131 @@
...
+		return ctrl.Result{}
+	}
+
+	return ctrl.Result{RequeueAfter: retryAfter}
+}
+
+// maxThrottleResult returns whichever of a and b has the larger RequeueAfter. Used
+// to collapse per-side throttle signals into a single result while reconciling
+// owners and members in parallel.
+func maxThrottleResult(a, b ctrl.Result) ctrl.Result {
```
matthchr: minor, naming: More like `maxResult` or `maxResultDuration` or something (there also might already be something in controller-runtime doing this), but see comment about maybe encoding backoff in the error in which case you don't need this at all.

v2/api/entra/v1/securitygroup_types.go
theunrepentantgeek: Both collections are defined as a [`directoryObject`](https://learn.microsoft.com/en-us/graph/api/resources/directoryobject?view=graph-rest-1.0) collection, so it comes from the API.

v2/api/entra/v1/securitygroup_types.go
theunrepentantgeek: They're serving two functions - giving the operation a meaningful name, and exposing the operations outside the current package. I'd be uncomfortable exposing the helper method as public.

v2/api/entra/v1/securitygroup_types.go
theunrepentantgeek: Yes.

v2/api/entra/v1/securitygroup_types.go
theunrepentantgeek: Correct. We have to use the annotations when creating the security group to avoid being cut off without access to update a security group we just created. 

Put another way, if we use a simple POST to create the security group, the group ends up with no Owners. For some kinds of groups, that's fine; but for others, we'd end up without the right permissions to update the group at all.

v2/internal/reconcilers/entra/securitygroup_relationships.go
theunrepentantgeek: I've improved the naming: `relationshipSide` --> `relationshipDefinition` and `reconcileRelationshipSide` -> `reconcileRelationship`

v2/internal/reconcilers/entra/securitygroup_relationships.go
theunrepentantgeek: Agree.

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
theunrepentantgeek: Fixed in #5232

v2/api/entra/v1/securitygroup_types.go
theunrepentantgeek: Consolidated to use the same helper methods as the Update path.

v2/internal/reconcilers/entra/entra_securitygroup_retry.go
theunrepentantgeek: Doing an error tree walk is hard to get right, gets flagged aggressively by linters, and gets rewritten by `go fix` to use `AsType[]` instead of type casts.

Took a different approach and simplified reconciliation of owners and members to work sequentially, avoiding any possibility of multiple returns.

v2/internal/reconcilers/entra/entra_securitygroup_retry.go
theunrepentantgeek: Clamped to the range 0-3600 seconds.

v2/internal/reconcilers/entra/securitygroup_relationships.go
theunrepentantgeek: You're right - we never have a _don't care_ scenario, it's always defined (even if empty).

v2/internal/util/interval/interval_calculator.go
theunrepentantgeek: Good idea.

v2/internal/util/interval/interval_calculator.go
theunrepentantgeek: Correct, I've simplified this area a lot.

v2/internal/reconcilers/entra/entra_securitygroup_retry.go
theunrepentantgeek: Yeah, don't need this any longer.

v2/api/entra/v1/securitygroup_types.go
matthchr: Hmm fair enough

v2/api/entra/v1/securitygroup_types.go
```diff
@@ -87,16 +86,12 @@ type SecurityGroupSpec struct {
 	// Required when ASO authenticates with an app-only token and the calling principal lacks Group.ReadWrite.All —
 	// otherwise the created group has no owners and is unmanageable.
 	// +kubebuilder:validation:MaxItems=20
-	// +kubebuilder:validation:XValidation:rule="size(self.filter(x, has(x.objectID)).map(x, x.objectID).distinct()) == size(self.filter(x, has(x.objectID)))",message="owners must be unique"
```
matthchr: Intentional that we lost these validations? They seemed reasonable to me

v2/internal/reconcilers/entra/entra_securitygroup_retry.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	return conditions.NewReadyConditionImpactingError(
+		eris.Wrap(err, "error reconciling SecurityGroup owners/members"),
+		conditions.ConditionSeverityWarning,
+		reasonRelationshipFailed,
+	)
+}
+
+// maxThrottleResult returns whichever of a and b has the larger RequeueAfter. Used
```
matthchr: Where is this used now?

matthchr (approved):
Approved but failing CI and few minor questions.

v2/api/entra/v1/securitygroup_types.go
theunrepentantgeek: Yes, intentional. They look like valid CEL but don't run in our context because of missing functions, including `distinct`. We validate these conditions via our webhook anyway so removing them seemed reasonable.

v2/internal/reconcilers/entra/entra_securitygroup_retry.go
theunrepentantgeek: It's not - will remove.

theunrepentantgeek:
I'm experimenting with an AI skill to provide review guidance - not to do the review, but to provide a starting point for a human reviewer. 

🤖 TheUnrepentantRobot writes:

# Guide to declarative Entra SecurityGroup relationships

Owners and members become validated desired state, applied atomically at creation and continuously reconciled thereafter.

**Start here:** `securitygroup_types.go:83` — this defines the new API contract and its creation semantics.

## 1. Relationship references become API state
**What changed:** Owners and members accept either inline GUIDs or ConfigMap references. Resolution catches duplicates across both forms, while creation uses Graph’s `@odata.bind` representation to avoid creating an inaccessible ownerless group.

**Why it matters:** Relationship identity is now declarative, indirect where needed, and established during the initial Graph transaction.

**Grok first:** `securitygroup_types.go:83`, `securitygroup_types.go:113`, `securitygroup_types.go:155`

## 2. Relationships are continuously converged
**What changed:** Status records observed owners and members, and updates diff that state against the resolved specification. Adds precede removals; dynamic groups reject and skip explicit members; Entra’s last-owner restriction is tolerated.

**Why it matters:** The fields are mutable desired state, not merely creation-time inputs. This behavior is broader than the PR description’s emphasis on fields and validation.

**Grok first:** `securitygroup_relationships.go:40`, `securitygroup_relationships.go:81`, `entra_securitygroup_reconciler.go:227`

## 3. Relationship failures acquire retry semantics
**What changed:** Graph permission failures and ordinary relationship failures produce distinct Ready-condition reasons and slow/fast retry classifications. Graph `Retry-After` values can lengthen, but never shorten, that backoff.

**Why it matters:** A relationship failure now affects resource readiness and reconciliation timing without being treated like an ordinary group PATCH failure.

**Grok first:** `entra_securitygroup_retry.go:20`, `interval_calculator.go:114`

**How the rest follows:** Webhook/CEL rules enforce the reference and dynamic-membership contract; paginated Graph reads populate status; the beta SDK supplies typed relationship endpoints; generated APIs, dependencies, recordings, and unit tests cascade from those decisions.

theunrepentantgeek:
@copilot resolve the merge conflicts in this pull request

v2/api/entra/v1/securitygroup_types.go
```diff
@@ -112,34 +112,46 @@ type SecurityGroupMemberReference struct {
...
+
+	return id, nil
 }
 
 // OriginalVersion returns the original API version used to create the resource.
 func (spec *SecurityGroupSpec) OriginalVersion() string {
 	return GroupVersion.Version
 }
 
+// HasDynamicMembership returns true when membership is managed by an Entra dynamic membership rule.
```
matthchr, line 145: minor: Can you expand on what exactly an entra dyanic membership rule is (is there a doc link somewhere?)

v2/cmd/asoctl/go.sum
```diff
@@ -205,8 +227,8 @@ github.com/microsoft/kiota-serialization-multipart-go v1.1.2 h1:1pUyA1QgIeKslQwb
 github.com/microsoft/kiota-serialization-multipart-go v1.1.2/go.mod h1:j2K7ZyYErloDu7Kuuk993DsvfoP7LPWvAo7rfDpdPio=
 github.com/microsoft/kiota-serialization-text-go v1.1.3 h1:8z7Cebn0YAAr++xswVgfdxZjnAZ4GOB9O7XP4+r5r/M=
 github.com/microsoft/kiota-serialization-text-go v1.1.3/go.mod h1:NDSvz4A3QalGMjNboKKQI9wR+8k+ih8UuagNmzIRgTQ=
-github.com/microsoftgraph/msgraph-sdk-go v1.101.0 h1:9Ox6mlDTm9BroNpj9i4B241OIcnyv0kjwJHHNaeXkoY=
-github.com/microsoftgraph/msgraph-sdk-go v1.101.0/go.mod h1:qxzY5SaoPigY6/Dpyfg4uigQjNDvL+sZl6fzD6EpWeQ=
+github.com/microsoftgraph/msgraph-beta-sdk-go v0.166.0 h1:BNbXKLwfooenreoEFj2dZrbmmzPb63YGa8/Me2tnnLI=
```
matthchr, line 230: Is there a note somewhere about why we're using the beta SDK?

v2/internal/reconcilers/entra/entra_securitygroup_reconciler.go
```diff
@@ -408,9 +432,106 @@ func (r *EntraSecurityGroupReconciler) loadGroupByID(
...
+	if err != nil {
+		return nil, err
+	}
+
+	return makeDirectoryObjects(ids), nil
+}
+
+// ownerIDRequestConfiguration limits the fields returned to the id
+// This avoids us pulling back unnecessary fields that may be sensitive.
+func ownerIDRequestConfiguration() *groups.ItemOwnersRequestBuilderGetRequestConfiguration {
```
matthchr, line 480: Nice

v2/internal/reconcilers/entra/entra_securitygroup_retry.go
```diff
@@ -76,3 +76,40 @@ func isPermissionError(err error) bool {
...
+
+	values := odataError.ResponseHeaders.Get("Retry-After")
+	if len(values) == 0 {
+		return 0, false
+	}
+
+	return parseRetryAfter(values[0], time.Now())
+}
+
+func parseRetryAfter(value string, now time.Time) (time.Duration, bool) {
```
matthchr, line 101: minor, style: parse doesn't (to me) imply enforcing max. Should parse just parse and then we have a separate step that does the max enforcement?

v2/go.mod
```diff
@@ -129,40 +126,52 @@ require (
...
+	k8s.io/kube-openapi v0.0.0-20260721132016-d427ff9ee9ad // indirect
+	k8s.io/utils v0.0.0-20260626114624-be93311217bd // indirect
+	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.36.0 // indirect
 	sigs.k8s.io/json v0.0.0-20250730193827-2d320260d730 // indirect
 	sigs.k8s.io/randfill v1.0.0 // indirect
-	sigs.k8s.io/structured-merge-diff/v6 v6.3.3 // indirect
+	sigs.k8s.io/structured-merge-diff/v6 v6.4.2 // indirect
 )
 
+require github.com/microsoftgraph/msgraph-beta-sdk-go v0.166.0
```
matthchr, line 161: minor: why in its own section?

matthchr (approved):
Basically lgtm, few minor questions. Approved feel free to ping me to re-approve if you fxi the minor nits

v2/go.mod
theunrepentantgeek: Quirk of the way Go added. Manually relocated.

v2/api/entra/v1/securitygroup_types.go
theunrepentantgeek: I've added a doc link here - https://learn.microsoft.com/en-us/entra/identity/users/groups-dynamic-membership

v2/cmd/asoctl/go.sum
theunrepentantgeek: I've added one in `entra_securitygroup_reconciler.go`.

v2/internal/reconcilers/entra/entra_securitygroup_retry.go
theunrepentantgeek: Agreed.

matthchr (approved):

PR #5474, opened 2026-06-16
chore: Update the azure-rest-api-specs submodule

theunrepentantgeek:
## What this PR does

Updates the submodule used to drive our code generator.

This is the first update since January 2026 and there's been a LOT of movement as product teams migrate to typespec internally. 

### Special notes

We should probably plan to do this monthly in the medium term, so that we don't get further large changes like this one.

gif

## Checklist

- [x] this PR contains tests


v2/api/apimanagement/v1api20240501/structure.txt
```diff
@@ -366,7 +366,8 @@ Backend: Resource
 │   │   ├── ConfigMapExpressions: *core.DestinationExpression[]
 │   │   └── SecretExpressions: *core.DestinationExpression[]
 │   ├── Owner: *genruntime.KnownResourceReference
-│   ├── Pool: *Object (1 property)
+│   ├── Pool: *Object (2 properties)
+│   │   ├── AdditionalProperties: map[string]v1.JSON
```
matthchr: Would like to understand where this came from and if it's intended/supposed to be here.

matthchr: (Same applies to the previous API version)

v2/api/cache/v1api20230401/arm/structure.txt
```diff
@@ -125,7 +125,8 @@ Redis_STATUS: Object (8 properties)
 │   ├── PublicNetworkAccess: *Enum (2 values)
 │   │   ├── "Disabled"
 │   │   └── "Enabled"
-│   ├── RedisConfiguration: *Object (17 properties)
+│   ├── RedisConfiguration: *Object (18 properties)
+│   │   ├── AdditionalProperties: map[string]v1.JSON
```
matthchr, line 129: and here

v2/api/network/v1api20220401/structure.txt
```diff
@@ -72,40 +72,8 @@ TrafficManagerProfile: Resource
...
-    │   ├── Priority: *int
-    │   ├── Subnets: Object (3 properties)[]
-    │   │   ├── First: *string
-    │   │   ├── Last: *string
-    │   │   └── Scope: *int
-    │   ├── Target: *string
-    │   ├── TargetResourceId: *string
-    │   ├── Type: *string
-    │   └── Weight: *int
+    ├── Endpoints: Object (1 property)[]
```
matthchr: Did we determine the cause here?

v2/api/operationalinsights/v20250701/structure.txt
```diff
@@ -13,7 +13,8 @@ Workspace: Resource
 │   │   └── Rule 2: Pattern: "^[A-Za-z0-9][A-Za-z0-9-]+[A-Za-z0-9]$"
 │   ├── DefaultDataCollectionRuleResourceReference: *genruntime.ResourceReference
 │   ├── Etag: *string
-│   ├── Features: *Object (5 properties)
+│   ├── Features: *Object (6 properties)
+│   │   ├── AdditionalProperties: map[string]v1.JSON
```
matthchr, line 17: Similar pattern again -- I need to look into what caused it.

v2/azure-arm/compute.yaml
```diff
@@ -202,21 +203,35 @@ groupModelConfiguration:
...
+    EncryptionSetProperties:
+      FederatedClientId:
+        $importConfigMapMode: optional
     SourceVault:
       Id:
         $referenceType: arm
     KeyForDiskEncryptionSet:
       KeyUrl:
         $importConfigMapMode: optional
-    EncryptionSetProperties:
```
matthchr, line 211: Curious: why move?

Alpha order? But if so `KeyForDisk...` is in the wrong spot too, no?

v2/azure-arm/cognitiveservices.yaml
```diff
@@ -2,7 +2,8 @@ typeTransformers:
   - name: DeploymentProperties
     property: CurrentCapacity
     remove: true
-    because: This property should have been marked readonly but wasn't.
+    because: This property is read-only and cannot be set by the user.
```
matthchr: Weird changing only this comment?

v2/azure-arm/network.yaml
```diff
@@ -166,6 +167,8 @@ groupModelConfiguration:
       # TODO[breaking]: This is actually a real resource that can be created as a child of a loadbalancer,
       # but we didn't model it that way initially, so marking it as not a resource for now to maintain backcompat
       $resourceEmbeddedInParent: LoadBalancer
+    # BackendAddressPool_STATUS_LoadBalancer_SubResourceEmbedded:
```
matthchr: Commented out? Intentional? Delete?

v2/azure-arm/operationalinsights.yaml
```diff
@@ -19,6 +19,12 @@ groupModelConfiguration:
...
+    WorkspaceProperties_PublicNetworkAccessForQuery:
+      $renameTo: PublicNetworkAccessForQuery
+    WorkspaceProperties_PublicNetworkAccessForIngestion:
+      $renameTo: PublicNetworkAccessForIngestion
+    WorkspaceSkuNameEnum:
+      $renameTo: WorkspaceSkuName
     WorkspaceFeatures:
       ClusterResourceId:
-        $referenceType: arm
+        $referenceType: arm
```
matthchr, line 30: minor: missing newline

v2/tools/generator/internal/jsonast/jsonast.go
```diff
@@ -102,7 +103,11 @@ func (scanner *SchemaScanner) AddTypeHandler(schemaType SchemaType, handler Type
 }
 
 // RunHandler triggers the appropriate handler for the specified schemaType
-func (scanner *SchemaScanner) RunHandler(ctx context.Context, schemaType SchemaType, schema Schema) (astmodel.Type, error) {
+func (scanner *SchemaScanner) RunHandler(
```
matthchr, line 106: These updates are nice but I am curious why they were needed?

matthchr (commented):
Few questions, overall looks great to me

v2/api/apimanagement/v1api20240501/structure.txt
theunrepentantgeek: In commit [6389cc78](https://github.com/Azure/azure-service-operator/pull/5474/changes/6389cc78297918cdb696c4fa7bbd1365649325f4) we improved our handling for arrays of unknown values, opting to treat them as arrays of `any` - this flows through into `AdditionalProperties` showing up in a few new places.

(We needed this because the typespec migrations would otherwise have resulted in a few of our CRDs losing properties they'd already had.)

v2/api/network/v1api20220401/structure.txt
theunrepentantgeek: It was to do with the `$resourceEmbeddedInParent` config - did I miss fixing one up? I'll check this.

v2/azure-arm/compute.yaml
theunrepentantgeek: Alphabetical ordering, yes - but I didn't want to create _too_ much chaos by moving all the things, so I just made it closer.

v2/tools/generator/internal/jsonast/jsonast.go
theunrepentantgeek: Not strictly needed, I made them as I was messing around trying to solve the `v1.JSON` problem, and then decided to keep them when I committed that fix.

v2/api/apimanagement/v1api20240501/structure.txt
matthchr: This guy: 
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/apimanagement/resource-manager/Microsoft.ApiManagement/ApiManagement/stable/2024-05-01/definitions.json#L1983

matthchr: Ok so I had a bunch of questions here and I dug into them some:

[AutoRest docs](https://github.com/Azure/autorest/blob/9510695785c91d411695458c6c9c8d2a27dab58d/docs/generate/how-autorest-generates-code-from-openapi.md?plain=1#L184) are probably relevant.

So in JSON schema _by default_, and also in Swagger (since it embeds JSON schema), `additionalProperties` is assumed to be `true` unless explicitly `false`. Obviously AutoRest and we do not make that same assumption (and I think it's common for code generators to not make that assumption when generating concrete types), as otherwise everything allows additional properties.

As discussed [here](https://json-schema.org/understanding-json-schema/reference/object#extending), extending a closed schema (`additionalProperties = false`) is also problematic and doesn't behave "as expected". So JSON schema added a keyword [unevaluatedproperties](https://json-schema.org/understanding-json-schema/reference/object#unevaluatedproperties).

So I am thinking we effectively want to (and do, in most places?) assume `unenaluatedProperties: false`.

Here's what AI says: (take with grain of salt)
> This is a well-known design tension in code generators. The pragmatic industry consensus:
> 
> For code generation: assume closed (i.e., unevaluatedProperties: false).
> 
> Rationale:
> 
> If a user explicitly defines a schema with named properties, their intent is to describe a concrete shape. If they wanted free-form, they'd just use "type": "object" with no properties.
> Generating a strongly-typed struct where "anything goes" provides zero value — you might as well use map[string]any.
> This is what AutoRest, OpenAPI Generator, TypeSpec, and most Azure SDK generators already do — they treat schemas as closed for type generation even when the spec technically allows extra fields.
> For validation: follow the spec literally — additional properties are allowed unless explicitly restricted.
> 
> So the answer is: two different behaviors for two different concerns:
> 
> Concern	Behavior
> Code generation (types/structs)	Assume closed — only generate fields for declared properties
> Wire-level validation	Permissive — don't reject unknown fields unless schema says to
> Round-tripping	Optionally include an AdditionalProperties map[string]any field to preserve unknown fields without losing data
> For the description-only sub-schema ({"description": "Backend Pool Properties"}), it adds no properties — it's purely annotation. The effective shape for codegen is just what BackendPool defines (services).

So IMO we want to treat this:
```
        "pool": {
          "allOf": [
            {
              "$ref": "#/definitions/BackendPool"
            },
            {
              "description": "Backend Pool Properties"
            }
          ]
```

as identical to `"$ref": "#/definitions/BackendPool"`, and if the user wanted to express `AdditionalProperties`, they would need to do:

```
        "pool": {
          "allOf": [
            {
              "$ref": "#/definitions/BackendPool"
            },
            {
              "description": "Backend Pool Properties",
              "additionalProperties: {}
            }
          ]
```

v2/api/cache/v1api20230401/structure.txt
```diff
@@ -35,7 +35,8 @@ Redis: Resource
 │   ├── PublicNetworkAccess: *Enum (2 values)
 │   │   ├── "Disabled"
 │   │   └── "Enabled"
-│   ├── RedisConfiguration: *Object (14 properties)
+│   ├── RedisConfiguration: *Object (15 properties)
+│   │   ├── AdditionalProperties: map[string]v1.JSON
```
matthchr, line 39: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/redis/resource-manager/Microsoft.Cache/Redis/stable/2023-04-01/redis.json#L1973C12-L1973C38

This one looks like they wanted additionalProperties and we were missing it before.

v2/api/network/v1api20220401/structure.txt
matthchr: Did you figure this out?

v2/api/operationalinsights/v20250701/structure.txt
matthchr: Also intended?

https://github.com/Azure/azure-rest-api-specs/blob/main/specification/operationalinsights/resource-manager/Microsoft.OperationalInsights/OperationalInsights/stable/2025-07-01/openapi.json#L8313

matthchr: and they had this before too actually but we weren't doing anything with it?

["additionalProperties": true,](https://github.com/Azure/azure-rest-api-specs/pull/36056/changes#diff-45f60a8ea95e6fd5109271feab0ac78f52f27abf41247eedb5ad6f81ec874b8aR637) was set even before the move to TypeSpec. I guess we just weren't honoring it before.

v2/api/quota/v1api20250901/structure.txt
```diff
@@ -22,7 +22,7 @@ Quota: Resource
 │       │       └── Value: *int
 │       ├── Name: *Object (1 property)
 │       │   └── Value: *string
-│       ├── Properties: *v1.JSON
+│       ├── Properties: map[string]v1.JSON
```
matthchr: This is here:
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/quota/resource-manager/Microsoft.Quota/Quota/stable/2025-09-01/openapi.json#L3367C1-L3369C10

and was https://github.com/Azure/azure-rest-api-specs/pull/37145/changes#diff-b830567827bbf68b75cf98654bdb95e1599483d7f409c41b9aab84f9244362afL702 before.

Which seem the same to me (one just omitted object).

This change seems correct to me.

v2/api/synapse/v20210601/structure.txt
```diff
@@ -20,7 +20,7 @@ Workspace: Resource
 │   ├── Encryption: *Object (1 property)
 │   │   └── Cmk: *Object (2 properties)
 │   │       ├── KekIdentity: *Object (2 properties)
-│   │       │   ├── UseSystemAssignedIdentity: *v1.JSON
+│   │       │   ├── UseSystemAssignedIdentity: map[string]v1.JSON
```
matthchr: Also changed but not clear why:
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/synapse/resource-manager/Microsoft.Synapse/stable/2021-06-01/workspace.json#L1015

This one had no TypeSpec change, it's always been like this. (No explicit Object)

v2/tools/generator/internal/jsonast/jsonast.go
```diff
@@ -863,11 +866,24 @@ func arrayHandler(ctx context.Context, scanner *SchemaScanner, schema Schema, lo
 	}
 
 	// get the only child type and wrap it up as an array type:
-
 	onlyChild := items[0]
 
 	astType, err := scanner.RunHandlerForSchema(ctx, onlyChild)
 	if err != nil {
+
+		// If we can't determine the type of the items, assume any
```
matthchr, line 882: As is this code cannot be hit, I think:

AI investigation:
> I can see the commented-out code in the arrayHandler (around line 870):
> 

> This code is dead because it can never be triggered. Here's why:
> 
> The commented-out code would catch UnknownSchemaError from the array item. But look at getSubSchemaType at the bottom of the file (line ~940):
> 
> 
> An UnknownSchemaError is only returned if the schema has no type, no ref, no allOf/oneOf/anyOf, no enum, no properties, and additionalPropertiesAllowed() returns false.
> 
> But as we discussed earlier, additionalPropertiesAllowed() returns true when AdditionalProperties is nil (the default). So any description-only schema with no explicit additionalProperties: false will be classified as Object by that fallback check — not Unknown.
> 
> In practice, array items that only have a description (the scenario the comment mentions) will always be handled as objects via that fallback, meaning RunHandlerForSchema never returns an UnknownSchemaError for them

matthchr: Though, given the behavior of other UnknownSchema handling, what yo uhave here seems correct.

v2/api/apimanagement/v1api20240501/structure.txt
matthchr: Actually above is a nice idea but the problem is people use this stuff in the OpenAPI very inconsistently... so we can't just assume `unenaluatedProperties: false` because there are some usages where users use `type: object` and _clearly_ mean `unenaluatedProperties: true`.

v2/api/synapse/v20210601/structure.txt
matthchr: I think this is wrong though, this should be a boolean as-per the doc and the name of the field.

Maybe let's type-transform this to boolean...? I think treating this as object will not work whereas technically v1.JSON worked before (because you could at least pass a raw bool to that).

matthchr (commented):
Ok, I went through all of the structural changes and here is my summary of them and what I think about them:

**apimanagement**

Added additionalProperties due to weird allOf w/ description only object

I don't think this is intended, **I think we should type-transform these away, or see my PR to your branch below which resolves this**

**cache**

Added additionalProperties and they want it there. **Keep.**

**operationalInsights**

Added additionalProperties and they want it there. **Keep.**

**quota**

Type changed from `*v1.JSON` to `map[string]v1.JSON`, **I don't think this is actually correct?**
Actually looking into it more, they lost the `object` here when it was migrated to TypeSpec -- seems a duplicate of this bug: https://github.com/Azure/azure-rest-api-specs/issues/44513

**synapse**

Let's type transform to a boolean, since I think that's what the field is actually supposed to be based on name + docs. Moving from `*v1.JSON` to `map[string]v1.JSON` is not right because it will block passing a boolean there, which is what the field is supposed to be.

I made these and a few other changes in this PR: https://github.com/Azure/azure-service-operator/pull/5526

My justification for going back to the old `if len(schema.properties()) > 0 || schema.additionalPropertiesSchema() != nil` check on line 956 is that if we use the more broad `schema.supportsAdditionalProperties` there, then basically nothing ever falls through to `UnknownSchemaError` because by default all objects support additional properties in JSON schema and nobody in Azure turns it off explicitly (they only turn it on explicitly).

We have logic that is predicated on returning `UnknownSchemaError` though, if you look for example at ln 136 of jsonast.go, it has this comment:
```
			if unknownSchema, ok := errors.AsType[*UnknownSchemaError](err); ok {
				if unknownSchema.Schema.description() != nil {
					// some Swagger types (e.g. ServiceFabric Cluster) use allOf with a description-only schema
					scanner.log.V(2).Info(
						"skipping description-only schema type",
						"schema", unknownSchema.Schema.url(),
						"description", *unknownSchema.Schema.description(),
					)
					continue
				}
			}
```

Which is precisely the `apimanagement` comment above as well. We have specific handling for that case but we can't hit it anymore if we assume every field is Object. (I was wrong about Object being the default btw. In JSON schema if no type is specified the default is Any)

v2/api/containerservice/v1api20240901/structure.txt
```diff
@@ -164,7 +164,7 @@ ManagedCluster: Resource
 │   │   ├── EnableAzureRBAC: *bool
 │   │   ├── Managed: *bool
 │   │   ├── ServerAppID: *string
-│   │   ├── ServerAppSecret: *string
+│   │   ├── ServerAppSecret: *genruntime.SecretReference
```
matthchr, line 167: Do we need to document this as breaking?

theunrepentantgeek: Yea, will capture this in a separate tracking issue.

v2/api/synapse/v1api20210601/structure.txt
```diff
@@ -20,7 +20,7 @@ Workspace: Resource
 │   ├── Encryption: *Object (1 property)
 │   │   └── Cmk: *Object (2 properties)
 │   │       ├── KekIdentity: *Object (2 properties)
-│   │       │   ├── UseSystemAssignedIdentity: *v1.JSON
+│   │       │   ├── UseSystemAssignedIdentity: *bool
```
matthchr, line 23: Worth putting this in the release notes/breaking issue too?

I don't actually think this is breaking (it was bool before just mis-labelled in the upstream Swagger), but still worth calling out explicitly probably?

v2/azure-arm/network.yaml
matthchr: Ping on this one

matthchr (approved):

v2/api/synapse/v1api20210601/structure.txt
theunrepentantgeek: Done.

v2/azure-arm/network.yaml
theunrepentantgeek: I've cleaned up this config file.

v2/azure-arm/cognitiveservices.yaml
theunrepentantgeek: AI glitch, I think. Fixed.

matthchr (approved):

PR #5476, opened 2026-06-19
test: Disable some tests during quick-checks

theunrepentantgeek:
## What this PR does

Disables our property based tests when running in `-short` mode.

AI agents are keen on running `task controller:quick-checks` even though it can take as long as 20 minutes (or more) to run; this PR neuters that somewhat so it runs in reasonable time.

We still run the full test suite in CI, so this seems a good compromise - all the tests are still compiled, we just skip the long running ones when working locally.

## Special Notes

This makes small changes to a LOT of files, you probably want to review the first commit in detail before looking at the rest.

gif


matthchr (approved):

matthchr (approved):

matthchr:
Looks like some conflicting changes, maybe due to resources that were added after you made this change?

matthchr (approved):

PR #5479, opened 2026-06-22
fix(generator): Stop deep-copying definitions on every pipeline state transition

theunrepentantgeek:
## What

The pipeline runs ~85 stages and `State.copy()` is invoked twice per stage (once during the action's transformation, once via `WithSeenStage`). Each call previously cloned the entire `TypeDefinitionSet` — a map of 100k+ entries — even though the map is treated as immutable by convention and the pipeline runner immediately discards the previous state.

This change shares the definitions map by reference (mirroring how `stagesSeen` and `stagesExpected` are already shared) and expands the doc-comment on `State` to explain the invariants callers must preserve.

## Why

`State.copy()` produces ~170 unnecessary deep-copies of the full definitions map per generator run. The biggest single pipeline hotspot — *Remove properties that point to embedded resources* (Stage 19) — drops from ~92s to ~48–65s in repeated measurements on `azure-arm.yaml`, with allocation traffic correspondingly reduced. Total wall time on this dataset varies with disk cache state but is consistently no worse than baseline.

## Why this is safe

The defensive deep-copy was already inconsistent: `stagesSeen` and `stagesExpected` are already shared by reference, and the pipeline runner (`CodeGenerator.Generate`) replaces `state = newState` immediately and never revisits the prior state. All mutations to `definitions` flow through `WithDefinitions` (replaces the map outright) or `WithOverlaidDefinitions` (allocates a new map via `OverlayWith`), so no in-place mutation of an earlier state's definitions occurs anywhere in the pipeline.

Verified by searching for direct mutations of `state.Definitions()` — none exist.

## Verification

- `task generator:quick-checks` passes (build, unit tests, lint, golden files)
- Full `go test ./...` in `v2/tools/generator` passes
- End-to-end `gen-types azure-arm.yaml` produces identical output

Identified as one of ten generator-performance improvements in #5479 / sibling PRs.

### DIFF
--- a/v2/tools/generator/internal/codegen/pipeline/state.go
+++ b/v2/tools/generator/internal/codegen/pipeline/state.go
@@ -21,10 +21,10 @@ import (
 // independent from those of the receiver. In practice, several of those structures are shared by
 // reference rather than deep-copied — this is safe because:
 //
-//   - The pipeline runner replaces the previous state with the new one immediately and never
-//     reads the previous state again (see CodeGenerator.Generate).
-//   - All mutations are performed only via the "With…" helpers, which either replace the entire
-//     map (definitions) or perform a single targeted mutation just after copy() returns.
+//   - The pipeline runner may still read the prior state for logging/debugging (for example,
+//     to diff definitions in CodeGenerator.Generate/logStateChanges) but then replaces it.
+//   - Callers must not mutate the shared definitions map in-place; when definitions change,
+//     "With…" helpers must replace it with a newly allocated map (for example via OverlayWith).
 //
 // Sharing the definitions map (rather than cloning the full 100k+ entry TypeDefinitionSet on
 // every stage transition) is a significant performance win.
@@ -114,7 +114,7 @@ func (s *State) CheckFinalState() error {
 	return kerrors.NewAggregate(errs)
 }
 
-// copy creates a new independent copy of the state.
+// copy creates a new State instance based on the receiver.
 //
 // definitions, stagesSeen, and stagesExpected are intentionally shared by reference with the
 // receiver: see the documentation on State for the reasoning. The caller is responsible for
--- a/v2/tools/generator/internal/codegen/pipeline/state.go
+++ b/v2/tools/generator/internal/codegen/pipeline/state.go
@@ -14,7 +14,20 @@ import (
 	"github.com/Azure/azure-service-operator/v2/tools/generator/internal/astmodel"
 )
 
-// State is an immutable instance that captures the information being passed along the pipeline
+// State is an immutable instance that captures the information being passed along the pipeline.
+//
+// Conceptually the State is immutable: each "With…" method returns a new *State whose internal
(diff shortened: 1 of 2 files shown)


v2/tools/generator/internal/codegen/pipeline/state.go
```diff
@@ -14,7 +14,20 @@ import (
...
 
-// State is an immutable instance that captures the information being passed along the pipeline
+// State is an immutable instance that captures the information being passed along the pipeline.
+//
+// Conceptually the State is immutable: each "With…" method returns a new *State whose internal
+// data structures (definitions, stagesSeen, stagesExpected, stateInfo) are treated as logically
+// independent from those of the receiver. In practice, several of those structures are shared by
+// reference rather than deep-copied — this is safe because:
+//
+//   - The pipeline runner may still read the prior state for logging/debugging (for example,
```
matthchr: I am not sure how this matters? How does the pipeline runner reading the old state have bearing here?

v2/tools/generator/internal/codegen/pipeline/state.go
```diff
@@ -14,7 +14,20 @@ import (
...
+//     to diff definitions in CodeGenerator.Generate/logStateChanges) but then replaces it.
+//   - Callers must not mutate the shared definitions map in-place; when definitions change,
```
matthchr: I think this is the key point but could we highlight it a bit more? The contract is _callers must produce a new map rather than modify the existing map when there are changes_ to preserve correctness.

v2/tools/generator/internal/codegen/pipeline/state.go
```diff
@@ -14,7 +14,20 @@ import (
 	"github.com/Azure/azure-service-operator/v2/tools/generator/internal/astmodel"
 )
 
-// State is an immutable instance that captures the information being passed along the pipeline
+// State is an immutable instance that captures the information being passed along the pipeline.
+//
+// Conceptually the State is immutable: each "With…" method returns a new *State whose internal
```
matthchr, line 19: 2. Are there stages that update the definitions map directly?
 Yes — exactly one:  create_resource_extension_types.go:55 

 state.definitions.AddTypes(extendedResourceDefs)   // in-place mutation of the (now shared) map
 return state, nil

theunrepentantgeek: I've replaced that with a call to `state.WithOverlaidDefinitions()`.

v2/tools/generator/internal/codegen/pipeline/state.go
theunrepentantgeek: Yeah, that's more of a transitory thing. Will update the comment.

v2/tools/generator/internal/codegen/pipeline/state.go
theunrepentantgeek: I've made this more explicit.

matthchr (approved):

PR #5480, opened 2026-06-22
fix(generator): Avoid allocating diff sets just to log added/removed counts per stage

theunrepentantgeek:
## What

`logStateChanges` previously called `TypeDefinitionSet.Except` twice on every pipeline stage to build two diff `TypeDefinitionSet` maps that were used only for their `len()`. Replace with `countDefinitionChanges`, a small helper that counts adds/removes by iterating both maps with key-only lookups (no allocations). When the two maps share the same underlying pointer — the common case when a stage doesn't touch the definitions set — the work is O(1).

## Why

With ~85 stages and 100k+ definitions in flight, the old implementation allocated several million map entries per generator run purely for logging. This delivered no information beyond the two integers we end up logging.

## Verification

- `task generator:quick-checks` passes
- Full `go test ./...` in `v2/tools/generator` passes
- Log output is byte-identical for the `added`/`removed`/`totalDefs` fields on representative pipelines

Identified as one of ten generator-performance improvements in sibling PRs (#5479, #5481).

### DIFF
--- a/v2/tools/generator/internal/codegen/code_generator.go
+++ b/v2/tools/generator/internal/codegen/code_generator.go
@@ -8,6 +8,7 @@ package codegen
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"runtime/debug"
 	"time"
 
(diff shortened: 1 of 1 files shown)


v2/tools/generator/internal/codegen/code_generator.go
```diff
@@ -327,18 +328,41 @@ func (generator *CodeGenerator) logStateChanges(
...
+		"removed", removed,
 		"totalDefs", len(newState.Definitions()),
 	)
 }
 
+// countDefinitionChanges returns the number of definitions added and removed when going
+// from the "before" set to the "after" set, without allocating any intermediate sets.
+// If both sets share the same underlying map (the common no-change case), the work is O(1).
+func countDefinitionChanges(after, before astmodel.TypeDefinitionSet) (added, removed int) {
+	if reflect.ValueOf(after).Pointer() == reflect.ValueOf(before).Pointer() {
```
matthchr, line 349: Is this possible without https://github.com/Azure/azure-service-operator/pull/5479? No right, since the maps are always copied then?

matthchr (approved):

v2/tools/generator/internal/codegen/code_generator.go
theunrepentantgeek: The shortcut wouldn't be triggered, but doing counts instead of set differencing is still a win.

PR #5502, opened 2026-06-30
fix(network): model ApplicationGateway private-link IP configuration

andreidorin-oprea:
## Problem

`ApplicationGateway` private-link gateways can't be fully expressed: the generated
`ApplicationGatewayPrivateLinkIpConfiguration` (in `api/network/v1api20220701`) collapses to a single
bare reference —

```go
type ApplicationGatewayPrivateLinkIpConfiguration struct {
    Reference *genruntime.ResourceReference `armReference:"Id"`
}
```

— so there is no way to set the inline `Name`, `Primary`, `PrivateIPAllocationMethod`, or `Subnet`
that a Private-Link Application Gateway PUT requires. The parent `...PrivateLinkConfiguration_STATUS`
is likewise collapsed to `{Id}`, so the status path to the nested IP configs doesn't exist either.

## Root cause

In `azure-arm/network.yaml`, `ApplicationGatewayPrivateLinkConfiguration` is marked
`$resourceEmbeddedInParent: ApplicationGateway`, but the **nested**
`ApplicationGatewayPrivateLinkIpConfiguration` (and both `_STATUS` types) are not — so the generator's
default heuristic treats the property-bag sub-resource as a bare ARM reference. The
`PrivateLinkService` block in the same file is the correct precedent: it marks **both** the service
and its `PrivateLinkServiceIpConfiguration` (spec + `_STATUS`) as embedded, and they generate with the
full field set.

## Fix

Add the missing embedded entries to the `ApplicationGateway` block (2022-07-01):

```yaml
ApplicationGatewayPrivateLinkConfiguration_STATUS:   { $resourceEmbeddedInParent: ApplicationGateway }
ApplicationGatewayPrivateLinkIpConfiguration:        { $resourceEmbeddedInParent: ApplicationGateway }
ApplicationGatewayPrivateLinkIpConfiguration_STATUS: { $resourceEmbeddedInParent: ApplicationGateway }
```

After regeneration, `ApplicationGatewayPrivateLinkIpConfiguration` carries
`Name` / `Primary` / `PrivateIPAddress` / `PrivateIPAllocationMethod` / `Subnet`, and the parent
`_STATUS` carries `IpConfigurations[]`. Subnet is the generic `SubResource` and there is no
`privateIPAddressVersion` — correct per the App Gateway swagger (distinct from PrivateLinkService).

## Tests (local only)

- `go build ./api/network/...` → ok; `go vet ./api/network/v1api20220701/...` → ok
- `go test ./api/network/v1api20220701/...` (api + arm + storage round-trip conversion) → ok

## Notes for maintainers

- The generated files were produced by a local generator run; please **regenerate via the canonical
  `task controller:generate-crds`** and treat the `azure-arm/network.yaml` change as the source of
  intent.
- No resource sample added yet — happy to add one. Opening as draft.
- Out of scope but related: the other AppGateway embedded sub-resources (BackendAddressPool,
  FrontendIPConfiguration, …) are still spec-only with the same collapsed-`_STATUS` shape — the same
  one-line-each fix would give them full status parity.


andreidorin-oprea:

theunrepentantgeek:
Good catch - and this looks like a clean fix. :+1:

I don't see anything here that would make this a breaking change - @matthchr  do you agree?

I've set our PR Checks running.

theunrepentantgeek (approved):

PR #5503, opened 2026-06-30
feat: add AppService 2025-05-01 for FlexConsumption function apps

andreidorin-oprea:
## What

Add AppService **2025-05-01** to the `web` group so FlexConsumption function apps can be expressed.
The `web` group is currently pinned at 2022-03-01, which predates FlexConsumption — there is no way
to set `Site.Properties.functionAppConfig` (deployment storage + runtime + scale/concurrency).
2025-05-01 is the latest stable AppService API version in the pinned specs and models
`functionAppConfig`.

> **Depends on #5504** (generator receiver / import-alias fix). That fix is required for the
> generated web code to compile: the new `FunctionsDeploymentStorage` type would otherwise yield a
> method receiver named `storage`, shadowing the `storage` conversion-package import when the
> conversion body constructs `storage.FunctionsDeploymentStorageAuthentication`. #5504 appears as the
> first two commits in this branch and drops out of the diff once it merges. **Please do not merge
> before #5504.**

## Change

`azure-arm/web.yaml`: add a 2025-05-01 block (coexisting with 2022-03-01), exporting `Site` /
`ServerFarm` / `SitesSourcecontrol` at `$supportedFrom: v2.21.0`, and labelling the ARM-ID
properties the generator flags for this version:

- `SiteProperties.ServerFarmId`, `SiteProperties.VirtualNetworkSubnetId`,
  `SiteProperties.ManagedEnvironmentId`
- `ServerFarmNetworkSettings.VirtualNetworkSubnetId`
- `DefaultIdentity.UserAssignedIdentityResourceId`
- `FunctionsDeploymentStorageAuthentication.UserAssignedIdentityResourceId`

2025-05-01 uses the consolidated TypeSpec-generated `openapi.json`, so its model names differ from
the older split-swagger versions (e.g. `SiteProperties` rather than `Site_Properties_Spec`).

The regenerated `Site_Spec.FunctionAppConfig` exposes `Deployment.Storage{Authentication,Type,Value}`
/ `Runtime{Name,Version}` / `ScaleAndConcurrency{AlwaysReady,InstanceMemoryMB,MaximumInstanceCount,
Triggers}`.

## How generated

Regenerated with the canonical `task controller:generate-crds` pipeline against the pinned
`azure-rest-api-specs` submodule (controller-gen v0.19.0, gofumpt v0.10.0). The run is idempotent —
re-running produces no diff, and `verify-no-changes` is green. Adding the newest version repoints the
existing 2022-03-01 storage conversions to the new 2025-05-01 hub, so those files change (as expected
for a version addition — they are not, and should not be, byte-identical).

## Tests (local)

- `go build ./api/web/...`, `go vet ./internal/controllers/` → ok
- `go test ./api/web/...` (incl. v20250501 + arm + storage conversion round-trips) → ok
- `check_samples.py` → ok (samples present for the new version)
- No intended 2022-03-01 API-surface change; its generated storage conversions are regenerated as
  expected because the conversion hub moves to 2025-05-01.

## Samples

Added `samples/web/v20250501/{serverfarm,site,sitessourcecontrol}.yaml`, mirroring the existing
2022-03-01 samples.

## Remaining work (needs Azure)

I don't have an Azure subscription to record against, so the following still need to be captured by
someone with cloud access:

- The **sample recording** `Test_Web_v20250501_CreationAndDeletion.yaml` (the samples suite
  auto-discovers the new folder and needs it in replay).
- A **CRUD test + recording** in `internal/controllers`. I've included compiling skeletons
  (`web_serverfarm_crud_v20250501_test.go`, `web_site_crud_v20250501_test.go`, copied from the
  2022-03-01 tests with versions bumped) as a starting point; they need their recordings captured
  against Azure.

Happy for a maintainer to record these, or to guide me if there's a way to run them against a test
subscription.


andreidorin-oprea:

theunrepentantgeek:
Looks like a little work is needed before we can approve the workflows and do a proper review. 

> Generated files were produced by a local, web-scoped generator run (verified to reproduce the committed web types byte-identical). 

I'm not sure what this means. Skimming the generated types under `v2/api/web`, the changes to existing files look correct for the addition of the new version - those files are not byte-identical, and shouldn't be.

> No resource sample / envtest recording added yet — happy to add.

Our CI both enforces the presence of samples, and runs a test cycle (using a recording) to ensure they work. We've got some docs on this (see [Create a sample](https://azure.github.io/azure-service-operator/contributing/add-a-new-code-generated-resource/create-a-sample/)) but will be happy to answer any questions you have. Oh, and the [/testing-aso-recordings](https://github.com/Azure/azure-service-operator/tree/main/.agents/skills/testing-aso-recordings) skill may be useful too.

We'll also need to add a CRUD test to the `internal/controllers` test suite and record its use. Copying the existing one and bumping the imported resource versions would be a good place to start. See [Write a test for the resource](https://azure.github.io/azure-service-operator/contributing/add-a-new-code-generated-resource/write-a-test/).

> Maintainers may prefer a newer AppService version (2024-11-01 / 2025-05-01 are also published) —
2024-04-01 is the minimum that has functionAppConfig.

Pulling a later stable release would be preferable, yes.

One final note - given we just released ASO v2.20, you'll need to tweak `$supportedFrom` to `v2.21`.

Thanks for the contribution, very much appreciated.

theunrepentantgeek (approved):

theunrepentantgeek:
CI was failing because the generated index files for the documentation were out of date - as soon as you added samples, our code generator started linking to them from the index files.

I took the liberty of updating those files and pushing the change to your branch, hope that's ok.

theunrepentantgeek:
This is looking really good. Your new tests and new samples are missing the required recordings, so our CI is currently failing.

Some background - we don't run all the tests under `internal/controllers` live all the time - not only would that be expensive, but it would be slow and potentially unreliable too (with the breadth of ASO resource support, the likelihood of a test run being impacted by an outage, upgrade, or capacity issue would be non-trivial). 

Instead, we run them once (when they're created) and we capture a recording of the HTTP interactions with the Azure Resource Manager (ARM), allowing the tests to be replayed during every CI run. This can (and has!) catch regressions where ASO's behaviour has changed.

We do something similar for all the samples - ensuring the samples are (a) correctly formed, and (b) work as advertised. This helps when users choose a sample as the basis for their own resources.

We've got good documentation on how to do this - see [Write a test for the resource](https://azure.github.io/azure-service-operator/contributing/add-a-new-code-generated-resource/write-a-test/) and [Create a sample](https://azure.github.io/azure-service-operator/contributing/add-a-new-code-generated-resource/create-a-sample/) for full details.

andreidorin-oprea:
Thanks for sorting out the doc index files and pushing that — much appreciated, and good to understand how the sample linking hangs together.

On the recordings: I'd genuinely love to add them, but I'm blocked by the subscription I have access to, and I don't think I can get there from my side. Two separate walls:

- it's Contributor, not Owner, so I can't create the RoleAssignment the identity-based deployment storage needs (the UAMI → Storage Blob Data Owner grant);
- and resource-group creation is denied by policy unless the RG carries an approved `Project` tag your test framework doesn't set — and that policy lives at a management-group scope I can't get an exemption for.

So producing a clean recording just isn't something I can do from here. I did validate the `functionAppConfig` shape end-to-end live earlier — both CRUD tests passed with a real create + delete — using a throwaway connection-string variant, so the resource itself works; it's purely the recording capture I'm stuck on.

Is there any chance your team could capture these on your test subscription? I completely understand if that's not how contributions normally flow — if so I'll keep trying to sort out access my end. And if it makes recording easier, I'm happy to switch the deployment auth back to a connection string (I went identity-based to keep secrets out of the cassette, but I can add an `AccountKey=` redaction to the redactor if you'd prefer that route).

The sample honesty flags from before still stand: the CRUD test is the shape-validated one, but the sample I couldn't run — so `AzureWebJobsStorage` isn't wired (plain-string app settings can't reference the UAMI clientId), the storage account uses a fixed name, and the `sitessourcecontrol` sample now points at a Flex site and is unverified. Glad to drop the sample and keep just the CRUD test if that's cleaner for you.

Thanks again for the help here!

theunrepentantgeek:
@andreidorin-oprea - I'm working on adding the requisite test recordings (and addressing any bugs discovered as a side effect); you'll see my changes as additional commits. Once it's working, we'll pass it to @matthchr for a review.

v2/samples/web/v20250501/v20250501_serverfarm_webapp.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: ServerFarm
+metadata:
+  name: aso-sample-webapp-plan
+  namespace: default
+spec:
+  location: westus3
+  owner:
+    name: aso-sample-rg
+  sku:
+    name: F1
```
matthchr, line 11: Do we actually really need samples for both free/flex?

Maybe just a single one with a comment on the sku saying both are supported would be better? The rest of the sample seems the same?

matthchr: Oh wait I see, the issue is that the site is very differently configured.

matthchr (approved):
LGTM

theunrepentantgeek:
And it's merged! Thanks for the contribution @andreidorin-oprea

andreidorin-oprea:
Thank you for all the support @theunrepentantgeek @matthchr 🙏

matthchr:
> Thank you for all the support @theunrepentantgeek @matthchr 🙏

Thank you for the contribution!

PR #5504, opened 2026-06-30
fix(generator): qualify method receivers that collide with arm/storage import aliases

andreidorin-oprea:
## Problem

`CreateReceiver()` derives a method-receiver name from the last word of a type name. For a type whose
name ends in `Storage` (e.g. `FunctionsDeployment_Storage` in the web group), the receiver becomes
`storage` — which **shadows the `storage` conversion-package import alias** used in generated
`AssignProperties_To_*` / `AssignProperties_From_*` methods. The generated code then fails to compile,
e.g. `storage.FunctionsDeployment_Storage_Authentication is not a type`.

`CreateReceiver` already qualifies Go reserved words and very short (≤3-char) identifiers — so `arm`
(len 3) was already safe, but `storage` was not. This extends the same guard to the fixed
conversion-package import aliases (`arm`, `storage`) so a type-derived receiver can never shadow them.
The receiver is renamed (e.g. `storage` → `deploymentStorage`); type names and JSON wire tags are
unchanged.

General fix: any current or future type whose name ends in `Storage` (or otherwise collides with these
aliases) benefits.

## Tests

- `go test ./internal/astmodel/...` (incl. a new case covering the receiver/alias collision) → ok
- `go test ./internal/codegen/...` (golden tests) → ok, no regressions

### DIFF
--- a/v2/api/cognitiveservices/v1api20250601/account_types_gen.go
+++ b/v2/api/cognitiveservices/v1api20250601/account_types_gen.go
@@ -7448,21 +7448,21 @@ type UserOwnedStorage struct {
 var _ genruntime.ARMTransformer = &UserOwnedStorage{}
 
 // ConvertToARM converts from a Kubernetes CRD object to an ARM object
-func (storage *UserOwnedStorage) ConvertToARM(resolved genruntime.ConvertToARMResolvedDetails) (interface{}, error) {
-	if storage == nil {
+func (ownedStorage *UserOwnedStorage) ConvertToARM(resolved genruntime.ConvertToARMResolvedDetails) (interface{}, error) {
+	if ownedStorage == nil {
 		return nil, nil
 	}
 	result := &arm.UserOwnedStorage{}
 
 	// Set property "IdentityClientId":
-	if storage.IdentityClientId != nil {
-		identityClientId := *storage.IdentityClientId
+	if ownedStorage.IdentityClientId != nil {
+		identityClientId := *ownedStorage.IdentityClientId
 		result.IdentityClientId = &identityClientId
 	}
 
 	// Set property "ResourceId":
-	if storage.ResourceReference != nil {
-		resourceReferenceARMID, err := resolved.ResolvedReferences.Lookup(*storage.ResourceReference)
+	if ownedStorage.ResourceReference != nil {
+		resourceReferenceARMID, err := resolved.ResolvedReferences.Lookup(*ownedStorage.ResourceReference)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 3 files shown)


andreidorin-oprea:

theunrepentantgeek:
Thanks for the contribution, much appreciated.

> The generated code then fails to compile, e.g. `storage.FunctionsDeployment_Storage_Authentication` is not a type.

When did this error occur? I'd like to reproduce the problem in order to verify your fix works without breaking any of the existing edge cases - getting `IdentifierFactory` to work originally took a _lot_ of iteration, so I'm sure you'll understand my caution.

theunrepentantgeek:
Our build checks [have failed](https://github.com/Azure/azure-service-operator/actions/runs/28445327743/job/84389992640?pr=5504) - some of the existing code is getting a new receiver name. 

This looks to be behaving as intended, and we most likely should merge this.

What I need to understand is why you are seeing compilation errors in your situation, but we haven't seen similar errors previously, even when the receiver name selected has been `storage`.

andreidorin-oprea:
Sure — I hit this adding the web 2025-05-01 version (#5503). The generated conversion for the new `FunctionsDeployment_Storage` type wouldn't build:

```
storage.FunctionsDeployment_Storage_Authentication is not a type
```

It's down to where the receiver shadows the `storage` import — it only shadows it *inside the method body*, not in the signature. That's why the existing `*Storage` types have always been fine: their bodies only use `storage.` as field access on the receiver (e.g. `storage.ResourceReference` on `UserOwnedStorage`), never as a package qualifier.

`FunctionsDeployment_Storage` is the first one that trips it, because it has a nested `Authentication` sub-object and the conversion builds it in the body. With the receiver name the old logic picks (`storage`):

```go
func (storage *FunctionsDeployment_Storage) AssignProperties_To_FunctionsDeployment_Storage(destination *storage.FunctionsDeployment_Storage) error {
    // ...
    if storage.Authentication != nil {
        var authentication storage.FunctionsDeployment_Storage_Authentication  // 'storage' is the receiver here, not the package
        err := storage.Authentication.AssignProperties_To_FunctionsDeployment_Storage_Authentication(&authentication)
        // ...
    }
}
```

In the body `storage` resolves to the receiver, so `storage.FunctionsDeployment_Storage_Authentication` reads as a selector on it → "is not a type". I threw together a small 4-case repro to be sure: a `storage` receiver builds fine when `storage.` is only in the signature or is field access, and fails only when the body names a type from the `storage` package.

So the fix just stops `CreateReceiver` handing out `storage`/`arm` as receiver names — same idea as the existing reserved-word/short-name guard. The only pre-existing fallout across the repo is the `UserOwnedStorage` receiver rename, which I've regenerated and pushed so `verify-no-changes` is clean.

theunrepentantgeek (approved):

theunrepentantgeek:
> So the fix just stops `CreateReceiver` handing out `storage`/`arm` as receiver names — same idea as the existing reserved-word/short-name guard. The only pre-existing fallout across the repo is the `UserOwnedStorage` receiver rename, which I've regenerated and pushed so `verify-no-changes` is clean.

I understand now, thanks for the explanation. The change looks solid to me, I've set the PR checks going.

theunrepentantgeek:
Checks passed, I've queued this to merge. Thanks once again!

PR #5507, opened 2026-07-03
feat: Migrate cdn group to hybrid versioning

theunrepentantgeek:
Migrates the `cdn` group from `VersionMigrationModeLegacy` to `VersionMigrationModeHybrid`. Both `v20210601` and `v20230501` are now available as new-style packages alongside the existing `v1api20210601` and `v1api20230501` for backward compatibility.

## Changes (in 8 focused commits)

1. **feat(generator): migrate cdn group to hybrid versioning mode** — flipped `cdn` in `legacy.go`.
2. **chore(generator): regenerate cdn types** — created `v20210601` and `v20230501` packages alongside `v1api*`.
3. **test(cdn): add samples for new v20210601 and v20230501 packages** — copied YAML files with new `apiVersion`.
4. **chore(cdn): generate deepcopy** — ran `controller:generate-crds`.
5. **test(cdn): update CRUD test imports** — imports flipped in `cdn_profile_afd_test.go` and `cdn_profile_crud_v20210601_test.go`. Test function names did not contain `v1api`, so no rename or re-record was required.
6. **docs(cdn): update reference docs** — regenerated to link to the new sample paths.
7. **test(cdn): make v20210601/v20230501 samples record-able** — Azure API changes forced three sample adjustments:
   - `v20210601` Profile switched to `Standard_AzureFrontDoor` SKU (Azure no longer accepts new classic Microsoft CDN profiles).
   - `v20210601` ProfilesEndpoint excluded from the samples test (inherently classic-only). The file is retained so `scripts/v2/check_samples.py` continues to find a sample for every supported resource version.
   - `v20230501` AfdCustomDomain now sets `tlsSettings.certificateType: ManagedCertificate` (Azure recently made this property required rather than server-defaulting).
8. **test(cdn): add sample recordings** — new recordings for `Test_Cdn_v20210601_CreationAndDeletion` and `Test_Cdn_v20230501_CreationAndDeletion`.

## Verification

All local checks pass:

- `task format-code` ✅
- `task generator:quick-checks` ✅
- `task controller:quick-checks` ✅
- `task asoctl:quick-checks` ✅
- All 4 CDN samples tests (v1api + v20*) pass in playback ✅
- `Test_CDN_Profile_CRUD` and `Test_CDN_Profile_AFD_CRUD` pass in playback ✅


v2/internal/testcommon/samples_tester.go
```diff
@@ -52,6 +52,11 @@ var exclusions = []*regexp.Regexp{
 	// ------------------------------
 	regexp.MustCompile(`insights/.*_webtest.yaml`), // Excluding webtest as it contains hidden link reference
 
+	// Classic CDN endpoint (Microsoft.Cdn/profiles/endpoints) can only be created under a classic Microsoft
+	// CDN profile, and Azure no longer allows creating new classic Microsoft CDN profiles. Keep the sample
+	// file present so scripts/v2/check_samples.py is satisfied, but skip it in the samples test.
+	regexp.MustCompile(`cdn/v20210601/.*_profilesendpoint.yaml`),
```
matthchr: Should we exclude from the v1api samples too, since apperantly it won't re-record?

Though... does that mean we also need to change those samples to use the new (non-classic?) profile?

v2/internal/controllers/cdn_profile_afd_v20230501_test.go
```diff
@@ -10,7 +10,7 @@ import (
```
matthchr, line 10: Consider renaming this file to match our standard scheme (I think it usually has the version in it at least?)

theunrepentantgeek: Renamed

v2/internal/testcommon/samples_tester.go
theunrepentantgeek: Excluded. 

I think it's a whole resource thing - the `ProfilesEndpoint` resource is effectively deprecated, in favour of `AfdEndpoint`.

So, there's no possible change to a make a `ProfilesEndpoint` work, we need to drop them.

matthchr (approved):

PR #5519, opened 2026-07-07
chore: migrate SQL to new style versioning

matthchr:
* Move SQL tests and samples to Azure Entra authentication. This is required for the tests to pass in our test subscriptions (enforced by policy), but is also a best-practice anyway, so better to highlight it rather than obfuscate it.
* Add *FromConfig for AAD Admin fields on the server. They were already on the admin object but were not on the Server field of the same name.
* Re-record all tests to use Entra rather than uname/password.

## Checklist

- [x] this PR contains tests


theunrepentantgeek (changes requested):
Looks good, but CI failing with modified crossplane files.

theunrepentantgeek (approved):

PR #5527, opened 2026-07-08
perf(config): speed up ObjectModelConfiguration lookups

theunrepentantgeek:
Improve the performance of the code generator's configuration lookups.

The generator consults `ObjectModelConfiguration` a very large number of times during code generation — every pipeline stage that walks resources or properties asks the same handful of type/property attributes for each item. All of those lookups fan out through the visitor hierarchy, and each one paid a cost for work that was already known.

Two focused changes here, both preserving existing semantics; a benchmark added first captures the baseline and makes the improvement measurable.

## Changes

**1. `test(config): add lookup benchmarks for ObjectModelConfiguration`**

Adds `go test -bench` coverage for the hot paths: `TypeAccess.Lookup` hit/miss, `PropertyAccess.Lookup` hit/miss, `IsTypeConfigured`, and a "repeated same name" variant that highlights redundant work.

**2. `perf(typo): fast-path AddTerm when the term is already known`**

`typo.Advisor.AddTerm` was called on every `findGroup`/`findVersion`/`findType`/`findProperty` and always took the exclusive lock, even though the same terms are added many thousands of times over during a run. Take the read lock and check for membership first; only escalate to the write lock the first time each unique term is seen.

**3. `perf(config): cache findGroup and findType resolutions`**

Add per-level resolution caches (`ObjectModelConfiguration.groupCache`, `VersionConfiguration.typeCache`), keyed on the raw name from the caller so that a hit also skips the `strings.ToLower` allocation. Nil values record known-absent lookups. Populated lazily, protected by an `sync.RWMutex`, and invalidated by `addGroup`/`addType`/`addTypeAlias` so that late additions from tests remain observable. The visitor hierarchy is unchanged, so `MetaType` wrapping semantics are preserved.

## Benchmark results

Measured on this machine (Intel i9-13900H, Go 1.26); all numbers are ns/op, B/op, allocs/op.

| Benchmark | Baseline | After (both changes) | Δ |
| --- | --- | --- | --- |
| `TypeAccess_Lookup_Hit` | 255.8 ns · 48 B · 3 allocs | 147.6 ns · 40 B · 2 allocs | **-42%** |
| `TypeAccess_Lookup_Miss` | 244.6 ns · 64 B · 3 allocs | 147.8 ns · 40 B · 2 allocs | **-40%** |
| `PropertyAccess_Lookup_Hit` | 540.5 ns · 80 B · 5 allocs | 285.0 ns · 72 B · 4 allocs | **-47%** |
| `PropertyAccess_Lookup_Miss` | 386.7 ns · 88 B · 5 allocs | 283.4 ns · 80 B · 4 allocs | **-27%** |
| `IsTypeConfigured` | 223.3 ns · 32 B · 3 allocs | 137.3 ns · 17 B · 2 allocs | **-38%** |
| `TypeAccess_Lookup_RepeatedSameName` | 235.6 ns · 48 B · 3 allocs | 136.8 ns · 40 B · 2 allocs | **-42%** |

## Testing

* `task generator:quick-checks` passes (unit tests, header check, specifier check, and lint all green).
* `typo` package tests pass.

## Notes

* No behavioural change is intended — the typo advisor still records the same set of terms and still generates the same suggestions; the cache always returns the same `*Configuration` pointers that the un-cached walk would have.
* Bypassing the visitor pattern for single-target lookups was considered but not pursued in this PR, per feedback that it would interact badly with `MetaType` wrappers (`ValidatedType`, `FlaggedType`) around `ObjectType`.

### DIFF
--- a/v2/internal/util/typo/typo_advisor.go
+++ b/v2/internal/util/typo/typo_advisor.go
@@ -29,11 +29,11 @@ func NewAdvisor() *Advisor {
 }
 
 // AddTerm records that we saw the specified item.
-// The advisor is consulted on every lookup made against the ObjectModelConfiguration, so the
-// same terms are added many times over. We take a read lock first and only escalate to the
-// exclusive write lock when the term is genuinely new; this keeps the hot lookup path free of
-// serialization once each unique term has been observed at least once.
 func (advisor *Advisor) AddTerm(item string) {
+	// The advisor is consulted on every lookup made against the ObjectModelConfiguration, so the
+	// same terms are added many times over. We take a read lock first and only escalate to the
+	// exclusive write lock when the term is genuinely new; this keeps the hot lookup path free of
+	// serialization once each unique term has been observed at least once.
 	advisor.lock.RLock()
 	if advisor.terms.Contains(item) {
 		advisor.lock.RUnlock()
--- a/v2/internal/util/typo/typo_advisor.go
+++ b/v2/internal/util/typo/typo_advisor.go
@@ -28,8 +28,19 @@ func NewAdvisor() *Advisor {
 	}
 }
 
-// AddTerm records that we saw the specified item
+// AddTerm records that we saw the specified item.
+// The advisor is consulted on every lookup made against the ObjectModelConfiguration, so the
+// same terms are added many times over. We take a read lock first and only escalate to the
+// exclusive write lock when the term is genuinely new; this keeps the hot lookup path free of
+// serialization once each unique term has been observed at least once.
 func (advisor *Advisor) AddTerm(item string) {
+	advisor.lock.RLock()
+	if advisor.terms.Contains(item) {
+		advisor.lock.RUnlock()
+		return
+	}
+	advisor.lock.RUnlock()
+
 	advisor.lock.Lock()
 	defer advisor.lock.Unlock()
 	advisor.terms.Add(item)
--- a/v2/tools/generator/internal/config/object_model_configuration.go
+++ b/v2/tools/generator/internal/config/object_model_configuration.go
@@ -326,34 +326,46 @@ func (omc *ObjectModelConfiguration) visitGroups(visitor *configurationVisitor) 
 }
 
 // findGroup uses the provided TypeName to work out which nested GroupConfiguration should be used
-func (omc *ObjectModelConfiguration) findGroup(ref astmodel.InternalPackageReference) *GroupConfiguration {
-	group := ref.Group()
-
+func (omc *ObjectModelConfiguration) findGroup(
+	ref astmodel.InternalPackageReference,
+) *GroupConfiguration {
 	if omc == nil || omc.groups == nil {
 		return nil
 	}
 
 	// Fast path: consult the resolution cache under a read lock. Cached entries may be nil to
 	// record a known-absent group so that repeated defensive lookups do not re-do the work.
-	omc.groupCacheLock.RLock()
-	if cached, ok := omc.groupCache[group]; ok {
-		omc.groupCacheLock.RUnlock()
-		omc.typoAdvisor.AddTerm(group)
+	if cached, ok := omc.findGroupFromCache(ref); ok {
 		return cached
 	}
-	omc.groupCacheLock.RUnlock()
 
+	group := ref.Group()
 	omc.typoAdvisor.AddTerm(group)
 	g := omc.groups[group]
 
 	omc.groupCacheLock.Lock()
+	defer omc.groupCacheLock.Unlock()
+
 	if omc.groupCache == nil {
 		omc.groupCache = make(map[string]*GroupConfiguration)
 	}
+
 	omc.groupCache[group] = g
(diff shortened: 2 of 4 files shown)


v2/internal/util/typo/typo_advisor.go
```diff
@@ -28,8 +28,19 @@ func NewAdvisor() *Advisor {
 	}
 }
 
-// AddTerm records that we saw the specified item
+// AddTerm records that we saw the specified item.
+// The advisor is consulted on every lookup made against the ObjectModelConfiguration, so the
```
matthchr: minor, style: Maybe put this impl note comment in the method body rather than in the public/external comments?

v2/tools/generator/internal/config/object_model_configuration.go
```diff
@@ -29,6 +30,13 @@ type ObjectModelConfiguration struct {
 	groups      map[string]*GroupConfiguration // nested configuration for individual groups
 	typoAdvisor *typo.Advisor
 
+	// groupCache memoises findGroup results so that repeated lookups against the same group
+	// name (which is the norm during code generation) skip the small amount of work involved.
+	// The key is the raw name from ref.Group(); a nil value means "known absent". Cache
+	// entries are cleared by addGroup so that late additions from tests remain observable.
+	groupCacheLock sync.RWMutex
```
matthchr, line 37: I am curious: why locking here, when other aspects of this struct are not threadsafe atm? Seems somewhat weird to have a lock on this but nothing else? Is this struct actually intended to be used in a threaded context? Do we really need threadsafety, or could we just drop the lock?

matthchr: I guess typoadvisor is threadsafe and the other aspects of this struct are not edited after construction, so that's why it didn't have a lock before?

v2/tools/generator/internal/config/object_model_configuration_bench_test.go
```diff
@@ -0,0 +1,177 @@
...
+		if _, ok := model.SupportedFrom.Lookup(name); !ok {
+			b.Fatalf("expected hit for %s", name)
+		}
+	}
+}
+
+// BenchmarkTypeAccess_Lookup_Miss measures the cost of a Lookup that does not resolve, so we walk
+// the whole hierarchy for nothing. Misses are common because pipelines defensively check for
+// configuration on many types that have none.
+func BenchmarkTypeAccess_Lookup_Miss(b *testing.B) {
```
matthchr, line 120: minor: Worth leaving in a comment at the top of this file the results of running this benchmark on your machine at least?

Otherwise if we don't know what we had in terms of perf before, how do we compare if we regressed or not? It would at least show the relative perfs on one machine?

v2/tools/generator/internal/config/version_configuration.go
```diff
@@ -87,13 +106,28 @@ func (vc *VersionConfiguration) visitTypes(visitor *configurationVisitor) error
 
 // findType uses the provided name to work out which nested TypeConfiguration should be used
 func (vc *VersionConfiguration) findType(name string) *TypeConfiguration {
+	// Fast path: consult the resolution cache under a read lock. The cache is keyed on the
+	// raw name (not the lowercased form) so a hit avoids allocating a lowercase copy.
+	vc.typeCacheLock.RLock()
```
matthchr: minor, style, not sure if better: Both here and in the object_model file, you could technically make this a `findTypeCached() *TypeConfiguration, bool` and then the impl can use `defer` to unlock (which is nice, I guess?) and usage would be:
```
if cachedResult, ok vc.findTypeCached(); ok {
    return cachedResult
}
```

Up to you if that's better.

matthchr (approved):
Generally LGTM, few nits

v2/internal/util/typo/typo_advisor.go
theunrepentantgeek: Agreed. Done.

v2/tools/generator/internal/config/object_model_configuration.go
theunrepentantgeek: Configuration is _largely_ read only and we've gotten away with things to this point. 

The only changes made after loading the configuration are to do with supporting aliases - for example when we rename a type with `$exportAs` then we add the new name as an alias for the old, so that later configuration lookups work.

Fortunately, those stages are single threaded, and we haven't had any issues. But we _do_ have concurrent configuration access in the Loader, stage 1 of the pipeline, and possibly elsewhere.

v2/tools/generator/internal/config/object_model_configuration_bench_test.go
theunrepentantgeek: Done.

v2/tools/generator/internal/config/version_configuration.go
theunrepentantgeek: I like that approach. Done.

matthchr (approved):

Issue #3756, opened 2024-02-02
Feature: Add Microsoft.Authorization/locks; Lock resources


mehighlow:
Add implementation to lock resources to prevent them from being destroyed in the Azure Portal while ASO keeps managing them.

[Lock your resources to protect your infrastructure](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/lock-resources?tabs=json)

Let's say I want to minimize any chance of my resources being deleted while I manage them with ASO. From the ASO side, I have an option to set `serviceoperator.azure.com/reconcile-policy: detach-on-delete` to prevent accidental deletes of K8s objects from being reconciled with Azure. However, those resources can still be deleted from the Azure Portal. It would be great to use ASO to have an option to set a lock to prevent them from being deleted from the Azure Portal.

There are 2 types of locks:

1. **Read-only**, which can be projected to `serviceoperator.azure.com/reconcile-policy: skip`
2. **Delete** - this feature request.

The way I propose implementing delete lock is through an annotation like `serviceoperator.azure.com/reconcile-policy: lock-delete`, which can be used only in combination with `serviceoperator.azure.com/reconcile-policy: detach-on-delete`, unless you consider combining them together into a single annotation like `serviceoperator.azure.com/reconcile-policy: locked-detach-on-delete`

matthchr:
This seems like a reasonable ask.

matthchr:
Put in to 2.10 for design and discussion.

theunrepentantgeek:
At least part of the solution will involve importing [Microsoft.Authorization/locks](https://learn.microsoft.com/en-us/azure/templates/microsoft.authorization/locks?pivots=deployment-language-arm-template).

When the locks are in place, even ASO will be blocked from updates - which means natural drift might come into play. At the very least, ASO needs to know not to PUT a resource that's been locked, though it might be better if we couple this with #2811 (Resource should do a Diff with Azure)

matthchr:
Even if we are issuing a PUT on resources that are locked, we may want to back off much more aggressively in that case - see https://github.com/Azure/azure-service-operator/issues/4583.

What we should do at a very minimum is just classify this error as "expected" and allow AZURE_SYNC_INTERVAL to be the only retries, for locked resources.

matthchr:
Still definitely something we're interested in but needs a design which we haven't had time to do.

s4heid:
[...]

PR #5547, opened 2026-07-20
Classify ScopeLocked ARM errors as fatal rather than retryable

marcosaranda-appspace:
## What this PR does

When a `CanNotDelete` or `ReadOnly` management lock blocks an operation, ARM returns HTTP `409` with the code `ScopeLocked`. Right now that code isn't in the classifier's `switch` in `error_classifier.go`, so it falls through to `classifyHTTPError`; since 409 isn't 400, it ends up classified as retryable. The effect is that ASO keeps retrying an operation that can't succeed until someone removes the lock, and the Ready condition shows up as a `Warning` rather than an `Error`.

This moves `ScopeLocked` into the fatal bucket next to `Conflict`, so the Ready condition is surfaced with `Error` severity, which is more accurate for a state that won't resolve on its own. The ARM error message (which already names the lock) is still shown either way.

I confirmed against a live subscription that the code returned is `ScopeLocked` with HTTP 409, and added a unit test that mirrors the existing classifier tests.

refs #3756

gif

## Checklist

- [x] Added unit test(s)
- [x] `gofmt`, `go vet`, and `golangci-lint` clean on the changed package
- [x] Full `./internal/reconcilers/arm/...` test suite passes

### DIFF
--- a/v2/internal/reconcilers/arm/error_classifier.go
+++ b/v2/internal/reconcilers/arm/error_classifier.go
@@ -66,6 +66,7 @@ func classifyCloudError(err *genericarmclient.CloudError) core.ErrorClassificati
 		return core.ErrorRetryable
 	case "BadRequestFormat",
 		"Conflict",
+		"ScopeLocked", // Returned when a CanNotDelete/ReadOnly management lock blocks the operation. Needs a human to remove the lock; won't resolve by itself, so this belongs with Conflict rather than the retryable bucket. See issue #3756.
 		"BadRequest",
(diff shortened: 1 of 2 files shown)


marcosaranda-appspace:
Closing this — I was working from an old local checkout and didn't realize `ScopeLocked` is already handled on `main`. It was added in #4757 (merged June 2025), and the approach there is better than what I proposed here: it keeps the error retryable but on a new `VerySlow` retry tier, so ASO backs off hard instead of hammering the API, and still recovers on its own once the lock is removed. Classifying it as fatal (what this PR did) would have lost that self-recovery.

Sorry for the noise, and thanks for the existing work here.

Issue #4810, opened 2025-07-07
Improve single-operator multitenancy credential management for Workload Identity


matthchr:
## Describe the current behavior

For workload identity in particular, if the user was able to discover the credential details (clientID/etc) via the Azure API and create their own copy of the secret in their namespace, they can use that secret because the FederatedIdentityCredential is for `system:serviceaccount:azure-service-operator:azureserviceoperator-default`.

## Describe the improvement

I think we may be able to harden the single-operator multitenancy against this case by optionally allowing serviceaccount impersonation and allowing you to configure ASO such that it doesn't use its own SA directly to token-exchange with Azure, instead it impersonates a configured SA in the target namespace and uses that SA token -- would need to do more investigation into how exactly this would all work though.

matthchr:
See where this came up: #4807

oysteinje:
What is the ETA on this? The current implementation of how the workload identity authentication scope works is not acceptable to use from a security perspective where the cluster is shared between different teams.

boomer41:
Ping from my side, too.
This is really a security concern and an relatively easy privilege escalation.

matthchr:
We need a detailed design for how we can improve this (probably using service accounts + impersonation, but we can maybe talk with some kube-upstream folks if they have other ideas/recommendations as well)

matthchr:
Still working on the above design.

boomer41:
> Still working on the above design.

The Design used by External Secrets Operator may be suitable for this, too.

jlhuilier-1a:
For me same requirement using the workload identity in multitenancy use case requires this SA impersonation feature. Current implementation is clearly not usable and only way is to fallback using clientSecret which is not great.

matthchr:
I've created a tentative design plan here: https://github.com/Azure/azure-service-operator/pull/5554

I will gather feedback on that design for a bit. Assuming there is no objections to the approach defined there I'll work on implementing this approach once that PR merges.

wasabii:
Any movement on this? It is the remaining thing keeping me from using ASO.

matthchr:
Design is merged, I am planning to have implementation (probably behind a flag) in 2.22

wasabii:
So, reading throught he design, it looks like the name of the secret is just hard coded? Not specified by the resource? Bummer. That won't work for me. Will have to wait for some additional thing, so resources can specify service account specifically.

[...]

PR #5554, opened 2026-07-21
feat: secure workload identity management design

matthchr:
Related to #4810.

## Checklist

- [ ] this PR contains tests

### DIFF
--- a/docs/hugo/content/design/ADR-2026-07-Secure-Workload-Identity-Multitenancy.md
+++ b/docs/hugo/content/design/ADR-2026-07-Secure-Workload-Identity-Multitenancy.md
@@ -1,0 +1,486 @@
+---
+title: "2026-07: Secure Workload Identity for Single-Operator Multitenancy"
+---
+
+## Context
+
+ASO supports [single-operator multitenancy]( {{< relref "credential-scope" >}} ) via a three-level credential hierarchy:
+
+- **Global credentials** (via the `aso-controller-settings` secret in the operator namespace)
+- **Per-namespace credentials** (via a fixed `aso-credential` secret in the resource's namespace)
+- **Per-resource credentials** (via the `serviceoperator.azure.com/credential-from` annotation)
+
+When using [Workload Identity](https://learn.microsoft.com/en-us/azure/aks/workload-identity-overview) for
+authentication, each credential secret contains only `AZURE_SUBSCRIPTION_ID`, `AZURE_TENANT_ID`, and `AZURE_CLIENT_ID`.
+The operator then uses its own projected ServiceAccount token
+(`system:serviceaccount:azureserviceoperator-system:azureserviceoperator-default`) to perform an OAuth2 token exchange
+with Azure AD, presenting the `AZURE_CLIENT_ID` from the secret. This works because the Managed Identity has a
+FederatedIdentityCredential registered for the ASO controller's ServiceAccount subject.
+
+### The Security Problem
+
+Unlike service principal and certificate authentication modes, none of the values in a Workload Identity credential 
+secret (`AZURE_SUBSCRIPTION_ID`, `AZURE_TENANT_ID`, `AZURE_CLIENT_ID`) are truly secret. They are GUIDs that can be discovered through:
+
+- The Azure portal or CLI (e.g. `az identity show`)
+- Azure Resource Graph queries
+- Reading secrets in other namespaces if the user has RBAC access to do so
+- Social engineering or documentation leaks
+
+Because all namespaces share the same ServiceAccount token for the Azure token exchange, a user who can create ASO
+resources in _any_ namespace can forge a credential secret referencing _any_ Managed Identity that has a
+FederatedIdentityCredential configured for the ASO ServiceAccount. This is a **privilege escalation vulnerability**:
+a tenant in namespace `team-a` can create a secret with `team-b`'s `AZURE_CLIENT_ID` and use `team-b`'s Azure identity
+to manage Azure resources.
+
+This has been raised as a security concern by multiple users
+([#4810](https://github.com/Azure/azure-service-operator/issues/4810),
+[#4807](https://github.com/Azure/azure-service-operator/issues/4807)):
+
+The only current mitigation is deploying ASO in
+[multi-operator multitenancy]( {{< relref "multitenant-deployment" >}} ) mode, where each namespace has its own
+ASO pod and ServiceAccount. This provides true isolation because each pod's SA token can only be exchanged for the
+Managed Identity it is federated with. However, multi-operator multitenancy is significantly harder to administer
+(multiple CRD installs, webhook routing, independent upgrades) and does not scale well.
+
+## Requirements
+
+1. **Namespace-level credential isolation**: It must not be possible for a tenant in one namespace to use a Managed
+   Identity that has been authorized for a different namespace, even if they know the Client ID.
+2. **Privileged authorization**: The binding between a namespace and an Azure identity must be authorized by a user
+   with appropriate privileges (cluster admin or similar), not merely by possession of credential values.
+3. **Self-service bootstrapping**: Teams that have already created a Managed Identity with appropriate
+   FederatedIdentityCredentials should be able to bootstrap their namespace without requiring write access to the
+   `azureserviceoperator-system` namespace.
+4. **Backward compatibility**: The existing credential hierarchy and secret format should continue to work for users
+   who don't need the hardened mode. This should be opt-in.
+5. **Workload Identity focus**: While the solution should be designed with extensibility in mind, the primary target
+   is Workload Identity, as client secret and client certificate authentication modes have genuinely secret credential
+   values that provide their own authorization factor.
+6. **Single operator**: The solution must work within the single-operator multitenancy model (one ASO pod for the
+   whole cluster).
+7. **Manageable complexity**: The solution should be simpler to administer than multi-operator multitenancy — that is
+   the entire motivation for single-operator mode.
+
+## Options
+
+### Option 1: ServiceAccount Impersonation (Per-Namespace ServiceAccounts for Azure Auth Only)
+
+The ASO controller impersonates a namespace-scoped ServiceAccount to obtain a token for Azure AD token exchange,
+instead of using its own ServiceAccount token directly.
+
+**Setup (admin actions):**
+
+1. Cluster admin creates a `ServiceAccount` in each tenant namespace (e.g., `aso-team-a` in namespace `team-a`).
+2. Cluster admin creates a `FederatedIdentityCredential` in Azure for `team-a`'s Managed Identity with subject
+   `system:serviceaccount:team-a:aso-team-a`.
+3. Cluster admin creates an RBAC `Role`/`RoleBinding` or `ClusterRole`/`ClusterRoleBinding` granting ASO's controller
+   ServiceAccount permission to create tokens for (`serviceaccounts/token`) the per-namespace SA.
+4. The credential secret in `team-a` references the Client ID of `team-a`'s Managed Identity as before.
+
+**Auth flow:**
+
+1. User creates an ASO resource in namespace `team-a`.
+2. ASO reads the credential secret from `team-a`, which includes the Client ID and the SA name (`aso-team-a`).
+3. ASO uses the TokenRequest API to mint a short-lived token for `ServiceAccount team-a/aso-team-a` with audience
+   `api://AzureADTokenExchange`. The token's subject is `system:serviceaccount:team-a:aso-team-a`.
+4. ASO exchanges this token with Azure AD for an access token, presenting the Client ID from the secret.
+5. Azure AD validates that the FederatedIdentityCredential on the Managed Identity matches the token's subject —
+   since the FIC is scoped to `team-a`'s SA, other namespaces cannot use this identity even if they know the Client ID.
+
+**Implementation details:**
+
+- ASO would use the [Kubernetes TokenRequest API](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/)
+  to create a short-lived token for the target namespace's SA with the `api://AzureADTokenExchange` audience.
+- The controller needs `serviceaccounts/token: create` permission for the well-known SA name. With the fixed-name
+  convention, this is a single `ClusterRole`/`ClusterRoleBinding` scoped via `resourceNames: ["aso-workload"]`,
+  shipped as part of the ASO Helm chart.
+- In the configurable variant (not chosen), a new field in the credential secret would indicate the SA name.
+  With the fixed-name convention chosen in this design, no credential secret changes are needed.
+
+**Variant: Fixed vs. configurable SA name**
+
+There are two sub-variants for how the SA name is determined:
+
+- **Configurable (per-secret):** The credential secret includes a field (`AZURE_WORKLOAD_IDENTITY_SERVICE_ACCOUNT`) specifying
+  which SA to use. This is the most flexible — different credentials in the same namespace could use different SAs.
+- **Fixed convention:** ASO uses a well-known SA name (e.g., `aso-workload`) in every namespace. No configuration
+  is needed — if the SA exists and the RBAC grant is in place, ASO uses it automatically.
+
+The fixed-name approach is simpler to administer and requires no changes to the credential secret format. The
+configurable variant could be offered as a future extension if there's demand.
+
+- When this mode is enabled, if ASO encounters a namespace without the designated SA or without the RBAC grant, it
+  refuses to authenticate and sets a clear error condition on the resource.
+- Note that the per-namespace SA only needs to exist and be tokenizable; it doesn't need any Kubernetes RBAC permissions itself.
+- Both namespace-scoped credentials and per-resource credentials within that namespace would use the same SA for
+  token exchange. Since Kubernetes RBAC is also namespace-scoped, there is no hard security boundary between these
+  two levels within the same namespace — the isolation boundary is the namespace itself.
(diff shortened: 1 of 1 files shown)


docs/hugo/content/design/ADR-2026-07-Secure-Workload-Identity-Multitenancy.md
```diff
@@ -0,0 +1,486 @@
...
+
+Because all namespaces share the same ServiceAccount token for the Azure token exchange, a user who can create ASO
+resources in _any_ namespace can forge a credential secret referencing _any_ Managed Identity that has a
+FederatedIdentityCredential configured for the ASO ServiceAccount. This is a **privilege escalation vulnerability**:
+a tenant in namespace `team-a` can create a secret with `team-b`'s `AZURE_CLIENT_ID` and use `team-b`'s Azure identity
+to manage Azure resources.
+
+This has been raised as a security concern by multiple users
+([#4810](https://github.com/Azure/azure-service-operator/issues/4810),
+[#4807](https://github.com/Azure/azure-service-operator/issues/4807)):
```
theunrepentantgeek:
```suggestion
[#4807](https://github.com/Azure/azure-service-operator/issues/4807)).
```

docs/hugo/content/design/ADR-2026-07-Secure-Workload-Identity-Multitenancy.md
```diff
@@ -0,0 +1,486 @@
...
+
+1. **Namespace-level credential isolation**: It must not be possible for a tenant in one namespace to use a Managed
+   Identity that has been authorized for a different namespace, even if they know the Client ID.
+2. **Privileged authorization**: The binding between a namespace and an Azure identity must be authorized by a user
+   with appropriate privileges (cluster admin or similar), not merely by possession of credential values.
+3. **Self-service bootstrapping**: Teams that have already created a Managed Identity with appropriate
+   FederatedIdentityCredentials should be able to bootstrap their namespace without requiring write access to the
+   `azureserviceoperator-system` namespace.
+4. **Backward compatibility**: The existing credential hierarchy and secret format should continue to work for users
+   who don't need the hardened mode. This should be opt-in.
```
theunrepentantgeek:
```suggestion
   who don't need the hardened mode. This should be opt-in (at least, for now).
```

docs/hugo/content/design/ADR-2026-07-Secure-Workload-Identity-Multitenancy.md
```diff
@@ -0,0 +1,486 @@
...
+The ASO controller impersonates a namespace-scoped ServiceAccount to obtain a token for Azure AD token exchange,
+instead of using its own ServiceAccount token directly.
+
+**Setup (admin actions):**
+
+1. Cluster admin creates a `ServiceAccount` in each tenant namespace (e.g., `aso-team-a` in namespace `team-a`).
+2. Cluster admin creates a `FederatedIdentityCredential` in Azure for `team-a`'s Managed Identity with subject
+   `system:serviceaccount:team-a:aso-team-a`.
+3. Cluster admin creates an RBAC `Role`/`RoleBinding` or `ClusterRole`/`ClusterRoleBinding` granting ASO's controller
+   ServiceAccount permission to create tokens for (`serviceaccounts/token`) the per-namespace SA.
```
theunrepentantgeek: SA = ServiceAccount?

```suggestion
   ServiceAccount permission to create tokens for (`serviceaccounts/token`) the per-namespace ServiceAccount.
```

docs/hugo/content/design/ADR-2026-07-Secure-Workload-Identity-Multitenancy.md
```diff
@@ -0,0 +1,486 @@
...
+3. ASO uses the TokenRequest API to mint a short-lived token for `ServiceAccount team-a/aso-team-a` with audience
+   `api://AzureADTokenExchange`. The token's subject is `system:serviceaccount:team-a:aso-team-a`.
+4. ASO exchanges this token with Azure AD for an access token, presenting the Client ID from the secret.
+5. Azure AD validates that the FederatedIdentityCredential on the Managed Identity matches the token's subject —
+   since the FIC is scoped to `team-a`'s SA, other namespaces cannot use this identity even if they know the Client ID.
+
+**Implementation details:**
+
+- ASO would use the [Kubernetes TokenRequest API](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/)
+  to create a short-lived token for the target namespace's SA with the `api://AzureADTokenExchange` audience.
```
theunrepentantgeek: When would this be created? 
When it expires, how does ASO know how to create a new one? 
Are there any issues with a short-lived token being used for a long-running operation (that is, an operation where it would expire part way through)?

docs/hugo/content/design/ADR-2026-07-Secure-Workload-Identity-Multitenancy.md
```diff
@@ -0,0 +1,486 @@
...
+- The controller needs `serviceaccounts/token: create` permission for the well-known SA name. With the fixed-name
+  convention, this is a single `ClusterRole`/`ClusterRoleBinding` scoped via `resourceNames: ["aso-workload"]`,
+  shipped as part of the ASO Helm chart.
+- In the configurable variant (not chosen), a new field in the credential secret would indicate the SA name.
+  With the fixed-name convention chosen in this design, no credential secret changes are needed.
+
+**Variant: Fixed vs. configurable SA name**
```
theunrepentantgeek:
```suggestion
**Variant: Fixed vs. configurable ServiceAccount name**
```

docs/hugo/content/design/ADR-2026-07-Secure-Workload-Identity-Multitenancy.md
```diff
@@ -0,0 +1,486 @@
...
+
+**Pros:**
+
+- Uses native Kubernetes primitives (ServiceAccount, TokenRequest API, RBAC) — no new CRDs.
+- RBAC for `serviceAccounts/token` on the main ASO service account can be scoped to specific SA 
+  names using `resourceNames`, if desired.
+- The authorization gate is the FederatedIdentityCredential in Azure (which requires Azure-level permissions to
+  create) and the RBAC grant allowing ASO to mint tokens for the per-namespace SA (which requires cluster-admin).
+  Without both, ASO can't authenticate in that namespace. Since each serviceAccount must have a FederatedIdentityCredential
+  bound to a unique subject (including that SA name + namespace), stealing a Client ID is useless since each Client
```
theunrepentantgeek: What if the customer has several clusters set up, say _dev_, _test_, _staging_, and _prod_, does the subject `SA name + namespace` still block the attack?

### Scenario 

_Alice_ is an evil software engineer with access to development cluster `phoenix-dev` who wants to compromise the namespace `payroll-system` in the production system `phoenix-prod`. Systems administrator _Bob_ has (unwisely) configured all the clusters with the same ServiceAccount.

Can _Alice_ gain access to the Azure resources in production by creating a namespace `payroll-system` in the dev cluster `phoenix-dev` and messing around creatively?

Is this blocked if _Bob_ uses a different ServiceAccount for each cluster?

docs/hugo/content/design/ADR-2026-07-Secure-Workload-Identity-Multitenancy.md
```diff
@@ -0,0 +1,486 @@
...
+  ID only works for the SA it is bound to.
+- Credential secrets retain their existing format — no changes needed.
+- Minimal change to ASO's Kubernetes RBAC model and controller architecture.
+- Supports self-service bootstrapping: once the ClusterRole is installed with ASO, teams can create the SA in their
+  own namespace and the FIC in Azure without needing write access to `azureserviceoperator-system`.
+
+**Cons:**
+
+- Requires cluster admin to create a ServiceAccount and RBAC binding per namespace — this is the "cost" of security,
+  but it's the same cost ESO imposes and is significantly less than multi-operator multitenancy.
```
theunrepentantgeek: ESO? Best to define initialisms and acronyms at first use.

docs/hugo/content/design/ADR-2026-07-Secure-Workload-Identity-Multitenancy.md
```diff
@@ -0,0 +1,486 @@
...
+- Kubernetes audit logs clearly attribute operations to the per-namespace SA.
+
+**Cons:**
+
+- Significantly more complex to implement. The ASO controller currently uses a single informer/cache for all
+  namespaces. Impersonating different SAs for different namespaces would require per-namespace clients or
+  impersonation headers on every API call.
+- Each per-namespace SA would need its own RBAC grants for the ASO resource types, secrets, configmaps, and events
+  in its namespace. This is a large amount of RBAC to administer.
+- Performance implications: per-namespace informer caches or impersonation on every call adds overhead.
```
theunrepentantgeek: I think we need to be explicit

```suggestion
- Performance implications: per-namespace informer caches or impersonation on every call adds both CPU and Memory overhead.
```

Can we quantify those yet?

docs/hugo/content/design/ADR-2026-07-Secure-Workload-Identity-Multitenancy.md
```diff
@@ -0,0 +1,486 @@
...
+- Centralized configuration — admin manages one allowlist.
+- Simple conceptual model.
+
+**Cons:**
+
+- Centralized configuration does not scale: every namespace/identity change requires updating the operator config
+  and potentially restarting the operator.
+- Awkward to manage in GitOps workflows where namespace and identity provisioning are decentralized.
+- If the allowlist is not enforced (i.e., made optional or left unconfigured), the security benefit is lost — same
+  as Option 3 without `AZURE_DISALLOW_WORKLOAD_IDENTITY_SECRETS`.
```
theunrepentantgeek: This seems significant. If someone gained access to the `azure-service-operator` namespace they could turn off the enforcement. 

Image

docs/hugo/content/design/ADR-2026-07-Secure-Workload-Identity-Multitenancy.md
```diff
@@ -0,0 +1,486 @@
...
+   and a credential Secret in their namespace. These are common namespace-admin permissions. The ClusterRole shipped
+   with ASO pre-authorizes token creation for the well-known SA name, so no per-namespace cluster-admin approval is
+   needed.
+2. **Azure side** (the real security boundary): The team must have Azure-level permissions to create a
+   FederatedIdentityCredential on the Managed Identity, binding it to the namespace's SA subject. Without this,
+   the token exchange will fail — even if the Kubernetes resources exist. This is the authorization gate that
+   prevents unauthorized use of an identity.
+
+Together these mean the Kubernetes setup is a prerequisite, but the actual trust decision is made in Azure: only
+users who can configure a FIC for a given Managed Identity can authorize a namespace to use it.
```
theunrepentantgeek: Can this creation be automated in some fashion by ASO to lower the overhead?

docs/hugo/content/design/ADR-2026-07-Secure-Workload-Identity-Multitenancy.md
```diff
@@ -0,0 +1,486 @@
...
+non-secret.
+
+Q: What happens if the ServiceAccount doesn't exist or ASO doesn't have permission to create tokens for it?
+
+A: ASO will fail the reconciliation with a clear error condition on the resource, indicating that the ServiceAccount
+is missing or the RBAC is not configured. This is the same behavior as other credential resolution failures today.
+
+Q: In `strict` mode, can a tenant's resources still fall back to the global credential?
+
+A: Yes. If a namespace has no namespace-scoped or per-resource credential, ASO will still fall back to the global
```
theunrepentantgeek, line 416: So, someone could trigger use of the global credential by deleting the namespace one from their namespace?

theunrepentantgeek (commented):
First pass review, with lots of questions. I need to digest this further, will do at least one more pass.

docs/hugo/content/design/ADR-2026-07-Secure-Workload-Identity-Multitenancy.md
matthchr: Fixed this and generally removed "SA" terminology in favor of the full "Service Account"

docs/hugo/content/design/ADR-2026-07-Secure-Workload-Identity-Multitenancy.md
matthchr: > When would this be created?

Some time during the reconcile loop, possibly/probably in a custom `WorkloadIdentityCredential` impmenentation very similar to [azidentity workload_identity.go](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/workload_identity.go), except instead of using a token file it uses namespace + secret details to dynamically request the token from APIServer.

> When it expires, how does ASO know how to create a new one?

We would have to manage token expiry - but this is included in the k8s API request. If we somehow failed to refresh a token and tried to use it, I believe the error would bubble-up to controller-runtime via our error mechanism and we would retry (but that would be the exceptional case, the expected case is we refresh via GetToken).

> Are there any issues with a short-lived token being used for a long-running operation (that is, an operation where it would expire part way through)?

I don't think so, because we break LROs up into multiple reconcile loops, we can refresh the token over the course of the LRO. As far as Azure is concerned, a separate token is still exchanged for the same Azure Managed Identity, so it doesn't really matter "which" token we present to Azure as long as it's valid and has a FIC registered to allow token exchange to the Azure identity. We _basically_ already have quite a similar flow, through [getCredentialFromSecret](https://github.com/Azure/azure-service-operator/blob/a8a9df7a5561769d98e84930e850578b3605ff44/v2/internal/identity/credential_provider.go#L220). That returns a token credential but even that token can expire - the tokenCredential implementation takes care of renewing the actual token, IIRC.

docs/hugo/content/design/ADR-2026-07-Secure-Workload-Identity-Multitenancy.md
matthchr: External Secrets Operator. Fixed this and a few other places with this issue

docs/hugo/content/design/ADR-2026-07-Secure-Workload-Identity-Multitenancy.md
matthchr: Updated to clarify CPU + Memory.

In terms of quantification of the overhead, there's not a great way to quantify it without doing some testing, but I am not currently planning to go this route (there don't seem to be a lot of advantages), so not sure that we need specific accurate numbers.

I added a bit more text here explaining what the overhead comes from (running informer cache infrastructure 1x per namespace means we eventually have the same resources held in memory but have many more active watches, plus whatever overhead the actual cache infrastructure has.

docs/hugo/content/design/ADR-2026-07-Secure-Workload-Identity-Multitenancy.md
matthchr: It's actually true for all 4 options. I have added appropriate con's to the other options as well.

The reason for this is that none of these options would be required by default, because that would be breaking, so all of them have some operator-level configuration enabling them.

FWIW though, it doesn't _really_ matter because having access to the ASO namespace at an admin level means you could make a pod doing anything ASO could do (since you have the k8s RBAC + Azure RBAC to do it), so ASO namespace is always going to be a juicy target. I don't think this design really changes that, but it does change the types of attacks you could execute _without_ ASO namespace access.

docs/hugo/content/design/ADR-2026-07-Secure-Workload-Identity-Multitenancy.md
matthchr: They can use ASO to create the FICs for other namespaces, provided that one namespace has the permission to create the UMI + RoleAssignments + FIC. See https://azure.github.io/azure-service-operator/guide/authentication/#using-aso-to-create-its-own-credentials (which we already support).

I've added a note to this effect in the document too

docs/hugo/content/design/ADR-2026-07-Secure-Workload-Identity-Multitenancy.md
matthchr: Yes.

Note though that this is the same as today and we do call this out explicitly in a few places:
https://azure.github.io/azure-service-operator/guide/authentication/credential-scope/
https://azure.github.io/azure-service-operator/guide/best-practices/security/

docs/hugo/content/design/ADR-2026-07-Secure-Workload-Identity-Multitenancy.md
matthchr: This is already safe even today as long as the user is using different Azure identities for the different clusters. This is because each FIC for a given identity has a UUID in it which is basically the clusters UUID.

Assume the following layout:
`phoenix-dev/payroll-system` UMI `dev`, FIC for the namespace
`phoenix-prod/payroll-system` UMI `prod`, FIC for the namespace

So in cluster `phoenix-dev` you can only token exchange for the `dev` UMI, which would presuambly have access to whatever the `dev` UMI is supposed to have access to. You can't token exchange for `prod` because the UUID for that cluster doesn't match.

Now, this all goes out the window if you share the same UMI for dev+prod and have 2 FICs (one for each cluster). Then you _can_ escalate from `dev` to `prod`. This doesn't doesn't change any of that though, it only changes the `intra-cluster` case where everybody has the same UUID so the only thing differentiating different token exchange requests is the namespace + sa-name, which today ASO uses the same NS+SA for everything which means Azure will accept the token exchange if the user guesses the right clientID/etc.

theunrepentantgeek (approved):

Issue #5108, opened 2026-01-16
Enable groups to skip deletion if the resource or parent no longer exists


theunrepentantgeek:
## Describe the current behavior

In PR #5040 we added support to improve ASO so that it won't issue DELETE to Azure for resources that don't exist.

However, most groups were deny-listed in that PR to avoid the need to re-record all the controller tests and samples in one go. 

## Describe the improvement

Progressively remove groups from the deny-list in `azure_generic_arm_reconciler_instance.go`, re-recording the controller and sample tests as we go. 

Once the deny-list is empty, it can be removed and this issue closed.

## Current Status

| Group                             | Enabled | Disabled |
| --------------------------------- | :-----: | :------: |
| alertsmanagement.azure.com        |         |    🚫     |
| apimanagement.azure.com           |         |    🚫     |
| app.azure.com                     |         |    🚫     |
| appconfiguration.azure.com        |         |    🚫     |
| authorization.azure.com           |    ✅     |          |
| cache.azure.com                   |         |    🚫     |
| cdn.azure.com                     |         |    🚫     |
| cognitiveservices.azure.com       |         |    🚫     |
| compute.azure.com                 |         |    🚫     |
| containerinstance.azure.com       |         |    🚫     |
| containerregistry.azure.com       |         |    🚫     |
| containerservice.azure.com        |         |    🚫     |
| datafactory.azure.com             |         |    🚫     |
| dataprotection.azure.com          |         |    🚫     |
| dbformariadb.azure.com            |         |    🚫     |
| dbformysql.azure.com              |    ✅     |         |
| dbforpostgresql.azure.com         |         |    🚫     |
| devices.azure.com                 |         |    🚫     |
| documentdb.azure.com              |         |    🚫     |
| eventgrid.azure.com               |         |    🚫     |
| eventhub.azure.com                |         |    🚫     |
| insights.azure.com                |         |    🚫     |
| keyvault.azure.com                |         |    🚫     |
| kubernetesconfiguration.azure.com |         |    🚫     |
| kusto.azure.com                   |         |    🚫     |
| machinelearningservices.azure.com |         |    🚫     |
| managedidentity.azure.com         |         |    🚫     |
| monitor.azure.com                 |         |    🚫     |
| network.azure.com                 |         |    🚫     |
| network.frontdoor.azure.com       |         |    🚫     |
| notificationhubs.azure.com        |         |    🚫     |
| operationalinsights.azure.com     |         |    🚫     |
[...]

PR #5575, opened 2026-07-23
feat: Enable smart deletion precheck for dbforpostgresql

theunrepentantgeek:
## Summary

Removes `dbforpostgresql.azure.com` from the ARM deletion-precheck deny list.

Makes the `v20250801` FlexibleServer sample deterministic for recording by using the test tenant placeholder directly. The AAD administrator sample is excluded from sample tests, consistent with existing administrator samples that require external AAD setup.

Refreshes the `v20250801` sample recording.

Part of #5108.

## Validation

- `TIMEOUT=60m TEST_FILTER='Test_Samples_CreationAndDeletion/Test_Dbforpostgresql_v20250801_CreationAndDeletion' task controller:test-samples`
- `TIMEOUT=60m TEST_FILTER='Test_DBForPostgreSQL_.*' task controller:test-controllers`


v2/samples/dbforpostgresql/v20250801/v20250801_flexibleserver.yaml
```diff
@@ -17,10 +17,7 @@ spec:
     key: password
   storage:
     storageSizeGB: 128
-  # Needed if you want to use FlexibleServers Administrator with AAD
```
matthchr, line 20: Why these changes? This resource was able to be successfully created before AFAIK

v2/internal/testcommon/samples_tester.go
```diff
@@ -67,6 +67,7 @@ var exclusions = []*regexp.Regexp{
 	// which the samples recordings aren't using.
 	regexp.MustCompile(`sql/.*_serversadministrator.yaml`),
 	regexp.MustCompile(`sql/.*_serversazureadonlyauthentication.yaml`),
+	regexp.MustCompile(`dbforpostgresql/.*_flexibleserversadministrator.yaml`),
```
matthchr: This resource should be possible to create, AFAIK it just requires a managed identity which we can make in our samples? Why does it need to be excluded?

theunrepentantgeek: I think I've got this fixed up.

v2/samples/dbforpostgresql/v20250801/v20250801_flexibleserver.yaml
theunrepentantgeek: No longer excluded.

matthchr (approved):

PR #5588, opened 2026-07-28
feat: support additional labels on managed CRDs

mboersma:
## What this PR does

Adds a `--crd-labels` flag (`crdLabels` in the Helm chart) so additional labels can be applied to all CRDs ASO manages. Labels not managed by ASO are preserved across CRD upgrades, and labels reserved by ASO (`app.kubernetes.io/name`, `app.kubernetes.io/version`, and the `serviceoperator.azure.com/` prefix) are rejected at startup.

This lets CAPZ label ASO's CRDs with `cluster.x-k8s.io/provider=infrastructure-azure` so they're included in the provider's cached informers, working around the CRD watch timeouts described in kubernetes-sigs/cluster-api-provider-azure#6492.

gif

## Checklist

- [x] this PR contains tests

### DIFF
--- a/v2/charts/azure-service-operator/values.yaml
+++ b/v2/charts/azure-service-operator/values.yaml
@@ -149,6 +149,23 @@ installCRDs: true
 # See https://azure.github.io/azure-service-operator/guide/crd-management for more details.
 crdPattern: ""
 
+# crdLabels is a comma-delimited (or semicolon-delimited) string, or a list of key=value strings, containing
+# additional labels to apply to all CRDs managed by ASO.
+# Setting this has no effect if installCRDs is false.
+# Existing labels not managed by ASO are preserved when CRDs are upgraded. Note that removing a label from this
+# list does not remove it from CRDs already in the cluster.
+# Labels reserved by ASO itself (app.kubernetes.io/name, app.kubernetes.io/version and any label prefixed with
+# serviceoperator.azure.com/) cannot be set here and will be rejected at startup.
+# Example: "cluster.x-k8s.io/provider=infrastructure-azure"
+# Example: "cluster.x-k8s.io/provider=infrastructure-azure,environment=production" would apply both labels.
+# Example: "cluster.x-k8s.io/provider=infrastructure-azure;environment=production" is equivalent, using the
+# semicolon delimiter also accepted by crdPattern.
+# Example as a list:
+#   crdLabels:
+#     - cluster.x-k8s.io/provider=infrastructure-azure
+#     - environment=production
+crdLabels: ""
+
 useJSONLogging: false
 
 # deploymentAnnotations contain the deployment annotations for Azure Service Operator
--- a/v2/cmd/asoctl/internal/crd/cleaner.go
+++ b/v2/cmd/asoctl/internal/crd/cleaner.go
@@ -26,6 +26,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/Azure/azure-service-operator/v2/internal/crdmanagement"
+	asolabels "github.com/Azure/azure-service-operator/v2/pkg/common/labels"
 )
 
 type Cleaner struct {
@@ -64,7 +65,11 @@ func (c *Cleaner) Run(ctx context.Context) error {
 		c.log.Info("Starting update")
 	}
 
-	appLabelRequirement, err := labels.NewRequirement(crdmanagement.ServiceOperatorAppLabel, selection.Equals, []string{crdmanagement.ServiceOperatorAppValue})
+	appLabelRequirement, err := labels.NewRequirement(
+		asolabels.ServiceOperatorAppLabel,
+		selection.Equals,
+		[]string{asolabels.ServiceOperatorAppValue},
+	)
 	if err != nil {
 		return err
 	}
@@ -75,7 +80,11 @@ func (c *Cleaner) Run(ctx context.Context) error {
 		return eris.Wrap(err, "failed to list CRDs")
 	}
 
-	versionLabelRequirement, err := labels.NewRequirement(crdmanagement.ServiceOperatorVersionLabelOld, selection.Exists, []string{})
+	versionLabelRequirement, err := labels.NewRequirement(
+		asolabels.ServiceOperatorVersionLabelOld,
+		selection.Exists,
+		[]string{},
+	)
 	if err != nil {
 		return err
 	}
--- a/v2/cmd/controller/app/flags.go
+++ b/v2/cmd/controller/app/flags.go
@@ -8,6 +8,11 @@ package app
 import (
 	"flag"
 	"fmt"
+
+	"github.com/rotisserie/eris"
+
+	"github.com/Azure/azure-service-operator/v2/internal/crdmanagement"
+	"github.com/Azure/azure-service-operator/v2/internal/labels"
 )
 
 type Flags struct {
@@ -21,11 +26,29 @@ type Flags struct {
 	EnableLeaderElection bool
 	CRDManagementMode    string
 	CRDPatterns          string // This is a ';' delimited string containing a collection of patterns
+	CRDLabels            string // This is a ',' or ';' delimited string containing labels to apply to managed CRDs
+}
+
+// parseCRDLabels parses the --crd-labels flag, rejecting any label reserved for ASO's own use.
+func parseCRDLabels(value string) (map[string]string, error) {
+	result, err := labels.ParseMap(value)
+	if err != nil {
+		return nil, err
+	}
+
+	// Done as a separate pass so that the general purpose label parsing stays free of CRD specific rules.
+	for key := range result {
+		if crdmanagement.IsReservedLabel(key) {
+			return nil, eris.Errorf("label %q is reserved for use by Azure Service Operator and cannot be overridden", key)
+		}
+	}
+
+	return result, nil
 }
 
 func (f Flags) String() string {
 	return fmt.Sprintf(
-		"MetricsAddr: %s, SecureMetrics: %t, ProfilingMetrics: %t, MetricsCertDir: %s, HealthAddr: %s, WebhookPort: %d, WebhookCertDir: %s, EnableLeaderElection: %t, CRDManagementMode: %s, CRDPatterns: %s",
+		"MetricsAddr: %s, SecureMetrics: %t, ProfilingMetrics: %t, MetricsCertDir: %s, HealthAddr: %s, WebhookPort: %d, WebhookCertDir: %s, EnableLeaderElection: %t, CRDManagementMode: %s, CRDPatterns: %s, CRDLabels: %s",
 		f.MetricsAddr,
 		f.SecureMetrics,
 		f.ProfilingMetrics,
@@ -36,6 +59,7 @@ func (f Flags) String() string {
 		f.EnableLeaderElection,
 		f.CRDManagementMode,
 		f.CRDPatterns,
+		f.CRDLabels,
 	)
 }
 
@@ -55,6 +79,7 @@ func InitFlags(flagSet *flag.FlagSet) *Flags {
 	flagSet.StringVar(&result.CRDManagementMode, "crd-management", "auto",
 		"Instructs the operator on how it should manage the Custom Resource Definitions. One of 'auto', 'none'")
 	flagSet.StringVar(&result.CRDPatterns, "crd-pattern", "", "Install these CRDs. CRDs already in the cluster will also always be upgraded.")
+	flagSet.StringVar(&result.CRDLabels, "crd-labels", "", "Comma-separated (or semicolon-separated) labels to apply to all managed CRDs (for example, example.com/owner=aso,environment=production). Labels reserved by the operator (app.kubernetes.io/name, app.kubernetes.io/version and the serviceoperator.azure.com/ prefix) cannot be set.")
 
 	return result
 }
--- a/v2/cmd/controller/app/flags_test.go
+++ b/v2/cmd/controller/app/flags_test.go
@@ -1,0 +1,100 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package app
+
+import (
+	"flag"
+	"testing"
+
+	. "github.com/onsi/gomega"
+)
+
+func TestCRDLabelsFlag(t *testing.T) {
+	t.Parallel()
+	g := NewWithT(t)
+	flagSet := flag.NewFlagSet("test", flag.ContinueOnError)
+	flags := InitFlags(flagSet)
+
+	g.Expect(flagSet.Parse([]string{"--crd-labels=cluster.x-k8s.io/provider=infrastructure-azure,example.com/owner=aso"})).To(Succeed())
+	g.Expect(flags.CRDLabels).To(Equal("cluster.x-k8s.io/provider=infrastructure-azure,example.com/owner=aso"))
+
+	parsed, err := parseCRDLabels(flags.CRDLabels)
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Expect(parsed).To(HaveLen(2))
+	g.Expect(parsed).To(HaveKeyWithValue("cluster.x-k8s.io/provider", "infrastructure-azure"))
+	g.Expect(parsed).To(HaveKeyWithValue("example.com/owner", "aso"))
+}
+
+// TestParseCRDLabels covers the CRD specific behaviour layered on top of labels.ParseMap.
+// General label parsing and validation is covered by TestParseMap in the labels package.
+func TestParseCRDLabels(t *testing.T) {
+	t.Parallel()
+
+	cases := map[string]struct {
+		value         string
+		expected      map[string]string
+		expectedError string
+	}{
+		"empty": {
+			value:    "",
+			expected: map[string]string{},
+		},
+		"reserved app label": {
+			value:         "app.kubernetes.io/name=mine",
+			expectedError: `label "app.kubernetes.io/name" is reserved`,
+		},
+		"reserved version label": {
+			value:         "app.kubernetes.io/version=v1.0.0",
+			expectedError: `label "app.kubernetes.io/version" is reserved`,
+		},
+		"reserved old version label": {
+			value:         "serviceoperator.azure.com/version=v1.0.0",
+			expectedError: `label "serviceoperator.azure.com/version" is reserved`,
+		},
+		"reserved prefix": {
+			value:         "serviceoperator.azure.com/anything=value",
+			expectedError: `label "serviceoperator.azure.com/anything" is reserved`,
+		},
+		"reserved label alongside a valid one": {
+			value:         "example.com/owner=aso,app.kubernetes.io/name=mine",
+			expectedError: `label "app.kubernetes.io/name" is reserved`,
+		},
+		"reserved label with surrounding whitespace": {
+			value:         " app.kubernetes.io/version = v1.0.0 ",
+			expectedError: `label "app.kubernetes.io/version" is reserved`,
+		},
+		"unreserved app.kubernetes.io label": {
+			value:    "app.kubernetes.io/part-of=platform",
+			expected: map[string]string{"app.kubernetes.io/part-of": "platform"},
+		},
+		"valid labels are passed through": {
+			value:    "environment=production;example.com/owner=aso",
+			expected: map[string]string{"environment": "production", "example.com/owner": "aso"},
+		},
+		"parse errors are propagated": {
+			value:         "not a label=value",
+			expectedError: `invalid label key "not a label"`,
+		},
+	}
+
+	for name, c := range cases {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			g := NewWithT(t)
+
+			actual, err := parseCRDLabels(c.value)
+
+			if c.expectedError != "" {
+				g.Expect(err).To(MatchError(ContainSubstring(c.expectedError)))
+				return
+			}
+
+			g.Expect(err).NotTo(HaveOccurred())
+			g.Expect(actual).To(HaveLen(len(c.expected)))
+			for key, value := range c.expected {
+				g.Expect(actual).To(HaveKeyWithValue(key, value))
+			}
+		})
+	}
+}
--- a/v2/internal/crdmanagement/manager.go
+++ b/v2/internal/crdmanagement/manager.go
@@ -30,16 +30,23 @@ import (
 
 	"github.com/Azure/azure-service-operator/v2/internal/util/kubeclient"
 	"github.com/Azure/azure-service-operator/v2/internal/util/match"
+	asolabels "github.com/Azure/azure-service-operator/v2/pkg/common/labels"
 )
 
-// ServiceOperatorVersionLabelOld is the label the CRDs have on them containing the ASO version. This value must match the value
-// injected by config/crd/labels.yaml
-const (
-	ServiceOperatorVersionLabelOld = "serviceoperator.azure.com/version"
-	ServiceOperatorVersionLabel    = "app.kubernetes.io/version"
-	ServiceOperatorAppLabel        = "app.kubernetes.io/name"
-	ServiceOperatorAppValue        = "azure-service-operator"
-)
+// IsReservedLabel returns true if the given label key is reserved for ASO's own use and so may not be
+// set by the user. Overwriting these labels would break CRD discovery (ListCRDs matches on
+// ServiceOperatorAppLabel) or CRD upgrade detection (VersionEqual compares ServiceOperatorVersionLabel).
+// Surrounding whitespace is ignored so that the result doesn't depend on the caller having trimmed the key.
+func IsReservedLabel(key string) bool {
+	key = strings.TrimSpace(key)
+
+	switch key {
+	case asolabels.ServiceOperatorAppLabel, asolabels.ServiceOperatorVersionLabel, asolabels.ServiceOperatorVersionLabelOld:
+		return true
+	}
+
+	return strings.HasPrefix(key, asolabels.ServiceOperatorLabelPrefix)
+}
 
 const CRDLocation = "crds"
 
@@ -163,7 +170,11 @@ func (m *Manager) ListCRDs(ctx context.Context, list *apiextensions.CustomResour
 	list.ResourceVersion = ""
 
 	selector := labels.NewSelector()
-	requirement, err := labels.NewRequirement(ServiceOperatorAppLabel, selection.Equals, []string{ServiceOperatorAppValue})
+	requirement, err := labels.NewRequirement(
+		asolabels.ServiceOperatorAppLabel,
+		selection.Equals,
+		[]string{asolabels.ServiceOperatorAppValue},
+	)
 	if err != nil {
 		return err
 	}
@@ -323,6 +334,18 @@ func (m *Manager) DetermineCRDsToInstallOrUpgrade(
 		result.DiffResult = VersionDifferent
 	}
 
+	goalCRDsWithDifferentMetadata := m.FindNonMatchingCRDs(existingCRDs, filteredGoalCRDs, DesiredMetadataEqual)
+	for name := range goalCRDsWithDifferentMetadata {
+		result, ok := resultMap[name]
+		if !ok {
+			return nil, eris.Errorf("Couldn't find goal CRD %q. This is unexpected!", name)
+		}
+
+		if result.DiffResult == NoDifference {
+			result.DiffResult = MetadataDifferent
+		}
+	}
+
 	// Collapse result to a slice
 	results := maps.Values(resultMap)
 	return results, nil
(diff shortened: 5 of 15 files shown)


v2/charts/azure-service-operator/values.yaml
```diff
@@ -149,6 +149,14 @@ installCRDs: true
 # See https://azure.github.io/azure-service-operator/guide/crd-management for more details.
 crdPattern: ""
 
+# crdLabels is a comma-delimited string or a list of key=value strings containing additional labels to apply to all
+# CRDs managed by ASO. Existing labels not managed by ASO are preserved when CRDs are upgraded.
+# Labels reserved by ASO itself (app.kubernetes.io/name, app.kubernetes.io/version and any label prefixed with
```
matthchr, line 157: minor: include `Setting this has no effect if installCRDs is false.` (since I assume that's true?) similar to crdPattern above.

Also may be worth clarifying that this supports either list or `;`-delimited string (crdpattern shows the `;` delimited string at least though it doesn't do a great job explaining that this can be a list too).

v2/cmd/controller/app/flags.go
```diff
@@ -21,11 +26,41 @@ type Flags struct {
 	EnableLeaderElection bool
 	CRDManagementMode    string
 	CRDPatterns          string // This is a ';' delimited string containing a collection of patterns
+	CRDLabels            string // This is a ',' delimited string containing labels to apply to managed CRDs
```
matthchr: Why `,` when `patterns` is `;`? Should we match?

matthchr: oh except the issue is maybe `,` is more common for labels lists and so more canonical upstream?

v2/cmd/controller/app/flags.go
```diff
@@ -21,11 +26,41 @@ type Flags struct {
...
+}
+
+func parseCRDLabels(value string) (map[string]string, error) {
```
matthchr, line 33: See v2/internal/labels/parse.go

Currently only used in asoctl so probably safe to combine this logic with that (with maybe a variant that accepts additional restrictions or something to pass the reserved-crd-label bit? Or could just do that in a second iteration pass since perf not too much of a concern?)

matthchr: Could then move TestParseCRDLabels there too.

v2/internal/crdmanagement/manager.go
```diff
@@ -41,6 +41,21 @@ const (
 	ServiceOperatorAppValue        = "azure-service-operator"
 )
 
+// ServiceOperatorLabelPrefix is the label domain reserved for ASO's own use.
+const ServiceOperatorLabelPrefix = "serviceoperator.azure.com/"
```
matthchr: minor: maybe define in v2/pkg/common/labels/labels.go (yeah I know too many possible places to put this stuff and it's a bit spread out).

Can probably leave `IsReservedLabel` here though since it's CRD specific.

v2/internal/crdmanagement/manager.go
```diff
@@ -323,6 +338,18 @@ func (m *Manager) DetermineCRDsToInstallOrUpgrade(
 		result.DiffResult = VersionDifferent
 	}
 
+	goalCRDsWithDifferentMetadata := m.FindNonMatchingCRDs(existingCRDs, filteredGoalCRDs, DesiredMetadataEqual)
```
matthchr, line 337: This function is actually _really_ expensive atm because it DeepCopy's the CRDs. This is actually not needed anymore (I just realized after re-looking at this code), because we got rid of the comparator where we were modifying the CRDs and doing a DeepEqual.

That means we should probably:
1. Drop the DeepCopy
2. Allow for multiple comparators _and_ have the result indicate which comparator succeeded/failed.

That way we just iterate the CRD collection once and don't deepcopy it 2x like we would here if merged as-is.

I can take a crack at fixing the perf issues with this function and adding support for what I am talking about above and then maybe you can build on top of that PR?

v2/internal/crdmanagement/metadata_test.go
```diff
@@ -0,0 +1,207 @@
...
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+
+	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestMergeCRDMetadata(t *testing.T) {
```
matthchr, line 17: minor, style: Can you either put these tests into `manager_test.go`, since the functions under test are also defined in manager.go, OR move the actual function definitions to `metadata.go`, that way the _test.go matches the file under test?

v2/internal/crdmanagement/manager.go
```diff
@@ -711,3 +743,66 @@ func VersionEqual(a apiextensions.CustomResourceDefinition, b apiextensions.Cust
...
+		if !found || existingValue != value {
+			return false
+		}
+	}
+
+	return true
+}
+
+// mergeMetadata preserves existing metadata while allowing desired metadata to take precedence.
+func mergeMetadata(existing map[string]string, desired map[string]string) map[string]string {
```
matthchr: minor: this is `samber/lo` `lo.Assign`, I believe.

Maybe just use that? IIRC we have `lo` imported here already.

v2/internal/crdmanagement/manager.go
```diff
@@ -711,3 +743,66 @@ func VersionEqual(a apiextensions.CustomResourceDefinition, b apiextensions.Cust
...
+		}
+		safeLabels[key] = value
+	}
+
+	for i := range crds {
+		crds[i].Labels = mergeMetadata(crds[i].Labels, safeLabels)
+	}
+}
+
+func copyCRDLabels(crds []apiextensions.CustomResourceDefinition) []apiextensions.CustomResourceDefinition {
```
matthchr: I am not sure I follow the point of this.

We mutate the goal CRD anyway, so we're just avoiding mutating the underlying labels map? Is there a particular reason it's OK to mutate the goal CRD itself but not that map?

v2/internal/crdmanagement/manager.go
```diff
@@ -711,3 +743,66 @@ func VersionEqual(a apiextensions.CustomResourceDefinition, b apiextensions.Cust
 
 	return aVersion == bVersion
 }
+
+// DesiredMetadataEqual returns true when all labels and annotations specified by the desired CRD are present on the existing CRD.
+// Additional metadata on the existing CRD is intentionally ignored and preserved during updates.
+func DesiredMetadataEqual(existing apiextensions.CustomResourceDefinition, desired apiextensions.CustomResourceDefinition) bool {
+	return metadataContains(existing.Labels, desired.Labels) && metadataContains(existing.Annotations, desired.Annotations)
```
matthchr: Annotations included here for completeness but not actually used anywhere atm?

v2/internal/crdmanagement/metadata_test.go
```diff
@@ -0,0 +1,207 @@
...
+	g.Expect(firstInstall[0].Labels).To(Equal(map[string]string{"bundled": "preserved", "first": "configured"}))
+	g.Expect(secondInstall[0].Labels).To(Equal(map[string]string{"bundled": "preserved", "second": "configured"}))
+	g.Expect(cached[0].Labels).To(Equal(map[string]string{"bundled": "preserved"}))
+}
+
+func TestApplyCRDLabelsIgnoresReservedLabels(t *testing.T) {
+	t.Parallel()
+	g := NewWithT(t)
+	crds := []apiextensions.CustomResourceDefinition{
+		{ObjectMeta: metav1.ObjectMeta{Labels: map[string]string{
```
matthchr: minor: object formatting a bit wonky here can you fix?

v2/internal/crdmanagement/manager_test.go
```diff
@@ -253,6 +253,22 @@ func Test_DetermineCRDsToInstallOrUpgrade(t *testing.T) {
 				g.Expect(apply).To(BeFalse())
 			},
 		},
+		{
```
matthchr, line 257: Do we also want tests for label exists but value wrong?
and a negative case like label exists but already there == no apply? Or label exists on existing but no label in desired == no apply?

Probably don't need to cover all cases as you have unit tests elsewhere for a bunch of them, but at least a few of the above seem to make sense at this level too?

v2/cmd/controller/app/flags.go
theunrepentantgeek: Probably best to support both, even if one is preferred over the other.

v2/internal/crdmanagement/manager.go
mboersma: They're load-bearing today, because we ship `cert-manager.io/inject-ca-from` on every conversion-webhook CRD, and `fixCRDNamespace` rewrites its namespace at load time.

So the annotation check catches a case `VersionEqual` misses: redeploying into a different namespace without a version bump. That now surfaces as `MetadataDifferent` and the CRDs get reapplied.

You're right there's no *config* surface for annotations, and I'm not proposing `--crd-annotations`. I added two tests pinning the above so it doesn't get dropped later.

v2/internal/crdmanagement/manager.go
mboersma: You're right, I've removed it. `applyCRDLabels` assigns a fresh map rather than mutating in place, so the `maps.Clone` was guarding against something that can't happen. The slice copy did stop `m.crds` being re-pointed at the merged map, but nothing reads the cache after `Install`, so there was no consumer for that isolation either. It now mutates the goal CRDs directly, consistent with the rest of the function.

v2/cmd/controller/app/flags.go
mboersma: Good call. I've moved the parsing into `internal/labels` as `ParseMap`, and `parseCRDLabels` is now just that call plus a second pass rejecting ASO-reserved keys, as you suggested. `TestParseCRDLabels` moved its general cases over to `TestParseMap`.

I left the existing `Parse`/`ParseAll` alone rather than routing them through the same validation, because they're currently annotation-rules-based and tightening them would change `asoctl import --label` behavior.

It seems there's a latent bug: `annotations.Parse` checks `len(key) > 63`, but the 63-char limit applies to the name segment only, so a valid key with a long DNS prefix gets rejected, and label *values* aren't validated at all. I can file an issue and unify them in a follow-up if you'd like.

v2/internal/crdmanagement/manager.go
mboersma: Done, the four CRD label constants and `ServiceOperatorLabelPrefix` now live in `v2/pkg/common/labels/labels.go`. I left `IsReservedLabel` in `crdmanagement` as you suggested, since it's CRD-specific.

v2/internal/crdmanagement/manager.go
mboersma: > I can take a crack at fixing the perf issues with this function

Sounds good. Let me know if you don't have time for that, I could give that a try too if need be.

v2/charts/azure-service-operator/values.yaml
```diff
@@ -149,6 +149,21 @@ installCRDs: true
...
 
+# crdLabels is a comma-delimited (or semicolon-delimited) string, or a list of key=value strings, containing
+# additional labels to apply to all CRDs managed by ASO.
+# Setting this has no effect if installCRDs is false.
+# Existing labels not managed by ASO are preserved when CRDs are upgraded. Note that removing a label from this
+# list does not remove it from CRDs already in the cluster.
+# Labels reserved by ASO itself (app.kubernetes.io/name, app.kubernetes.io/version and any label prefixed with
+# serviceoperator.azure.com/) cannot be set here and will be rejected at startup.
+# Example: "cluster.x-k8s.io/provider=infrastructure-azure"
+# Example: "cluster.x-k8s.io/provider=infrastructure-azure,environment=production" would apply both labels.
```
theunrepentantgeek, line 160: Maybe add a semicolon separated example too?

v2/cmd/asoctl/internal/crd/cleaner.go
```diff
@@ -26,6 +26,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/Azure/azure-service-operator/v2/internal/crdmanagement"
+	asolabels "github.com/Azure/azure-service-operator/v2/pkg/common/labels"
```
theunrepentantgeek, line 29: The package name `common` isn't very useful - it's about as bad as `util` or `misc`.

I suggest promoting the `labels` package up one level so that the import path is just `github.com/Azure/azure-service-operator/v2/pkg/labels`.

v2/cmd/asoctl/internal/crd/cleaner.go
```diff
@@ -64,7 +65,7 @@ func (c *Cleaner) Run(ctx context.Context) error {
 		c.log.Info("Starting update")
 	}
 
-	appLabelRequirement, err := labels.NewRequirement(crdmanagement.ServiceOperatorAppLabel, selection.Equals, []string{crdmanagement.ServiceOperatorAppValue})
+	appLabelRequirement, err := labels.NewRequirement(asolabels.ServiceOperatorAppLabel, selection.Equals, []string{asolabels.ServiceOperatorAppValue})
```
theunrepentantgeek: From a readability perspective, I'd suggest formatting this across multiple lines:

```suggestion
	appLabelRequirement, err := labels.NewRequirement(
		asolabels.ServiceOperatorAppLabel, 
		selection.Equals, 
		[]string{
			asolabels.ServiceOperatorAppValue,
		})
```

v2/cmd/asoctl/internal/crd/cleaner.go
```diff
@@ -75,7 +76,7 @@ func (c *Cleaner) Run(ctx context.Context) error {
 		return eris.Wrap(err, "failed to list CRDs")
 	}
 
-	versionLabelRequirement, err := labels.NewRequirement(crdmanagement.ServiceOperatorVersionLabelOld, selection.Exists, []string{})
+	versionLabelRequirement, err := labels.NewRequirement(asolabels.ServiceOperatorVersionLabelOld, selection.Exists, []string{})
```
theunrepentantgeek: Here too:

```suggestion
	versionLabelRequirement, err := labels.NewRequirement(
		asolabels.ServiceOperatorVersionLabelOld, 
		selection.Exists, 
		[]string{})
```

v2/cmd/controller/app/flags.go
```diff
@@ -21,11 +24,29 @@ type Flags struct {
...
+	result, err := labels.ParseMap(value)
+	if err != nil {
+		return nil, err
+	}
+
+	// Done as a separate pass so that the general purpose label parsing stays free of CRD specific rules.
+	for key := range result {
+		if crdmanagement.IsReservedLabel(key) {
+			return nil, fmt.Errorf("label %q is reserved for use by Azure Service Operator and cannot be overridden", key)
```
theunrepentantgeek: We prefer `eris.Errorf()`.

```suggestion
			return nil, eris.Errorf("label %q is reserved for use by Azure Service Operator and cannot be overridden", key)
```

v2/cmd/controller/app/flags_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	g.Expect(flagSet.Parse([]string{"--crd-labels=cluster.x-k8s.io/provider=infrastructure-azure,example.com/owner=aso"})).To(Succeed())
+	g.Expect(flags.CRDLabels).To(Equal("cluster.x-k8s.io/provider=infrastructure-azure,example.com/owner=aso"))
+
+	parsed, err := parseCRDLabels(flags.CRDLabels)
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Expect(parsed).To(Equal(map[string]string{
+		"cluster.x-k8s.io/provider": "infrastructure-azure",
+		"example.com/owner":         "aso",
+	}))
```
theunrepentantgeek: Lean into the assertion framework - we get better diagnostics for failures if we do.

```suggestion
	g.Expect(parsed).To(HaveLen(2))
	g.Expect(parsed).To(HaveKeyWithValue("cluster.x-k8s.io/provider", "infrastructure-azure"))
	g.Expect(parsed).To(HaveKeyWithValue("example.com/owner", "aso"))
```

v2/cmd/controller/app/flags_test.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// TestParseCRDLabels covers the CRD specific behaviour layered on top of labels.ParseMap.
+// General label parsing and validation is covered by TestParseMap in the labels package.
+func TestParseCRDLabels(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
```
theunrepentantgeek: I'm encouraging people to use `map[string]struct{...}` for table tests - it pushes the test name to a more prominent position. I'll demo, below.

v2/cmd/controller/app/flags_test.go
```diff
@@ -0,0 +1,69 @@
...
+		name  string
+		value string
+		want  map[string]string
+		ok    bool
+	}{
+		{name: "empty", value: "", want: map[string]string{}, ok: true},
+		{name: "reserved app label", value: "app.kubernetes.io/name=mine", ok: false},
```
theunrepentantgeek:
```suggestion
		"empty": {value: "", want: map[string]string{}, ok: true},
		"reserved app label": {value: "app.kubernetes.io/name=mine", ok: false},
```

v2/cmd/controller/app/flags_test.go
```diff
@@ -0,0 +1,69 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			g := NewWithT(t)
+			actual, err := parseCRDLabels(tt.value)
+			if tt.ok {
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(actual).To(Equal(tt.want))
+			} else {
+				g.Expect(err).To(HaveOccurred())
```
theunrepentantgeek: This just asserts that _any_ error occurs, without constraint. I like to always do some assertions on the content of the error too.

Assuming you replace `ok: bool` with `expectedError: string`, this could be:

```suggestion
				g.Expect(err).To(MatchError(ContainSubstring(tt.expectedError)))
```

BTW - `MatchError()` implies `HaveOccurred()` - if `err` is nil, you'll get a useful failure message.

v2/internal/crdmanagement/manager.go
```diff
@@ -30,16 +30,20 @@ import (
...
-const (
-	ServiceOperatorVersionLabelOld = "serviceoperator.azure.com/version"
-	ServiceOperatorVersionLabel    = "app.kubernetes.io/version"
-	ServiceOperatorAppLabel        = "app.kubernetes.io/name"
-	ServiceOperatorAppValue        = "azure-service-operator"
-)
+// IsReservedLabel returns true if the given label key is reserved for ASO's own use and so may not be
+// set by the user. Overwriting these labels would break CRD discovery (ListCRDs matches on
+// ServiceOperatorAppLabel) or CRD upgrade detection (VersionEqual compares ServiceOperatorVersionLabel).
+func IsReservedLabel(key string) bool {
```
theunrepentantgeek, line 40: Does this predicate need to be case-insensitive or whitespace-insensitive?

If someone manages to tunnel a label of `" app.kubernetes.io/Version"` to this point, should it be rejected? I suspect it should.

v2/internal/crdmanagement/manager.go
```diff
@@ -163,7 +167,7 @@ func (m *Manager) ListCRDs(ctx context.Context, list *apiextensions.CustomResour
 	list.ResourceVersion = ""
 
 	selector := labels.NewSelector()
-	requirement, err := labels.NewRequirement(ServiceOperatorAppLabel, selection.Equals, []string{ServiceOperatorAppValue})
+	requirement, err := labels.NewRequirement(asolabels.ServiceOperatorAppLabel, selection.Equals, []string{asolabels.ServiceOperatorAppValue})
```
theunrepentantgeek:
```suggestion
	requirement, err := labels.NewRequirement(
		asolabels.ServiceOperatorAppLabel, 
		selection.Equals, 
		[]string{
			asolabels.ServiceOperatorAppValue,
		})
```

v2/internal/crdmanagement/manager.go
theunrepentantgeek: @matthchr, how much of an impact will this have on startup resource requirements? I seem to remember we've had a couple reports of high memory usage as ASO starts up, maybe this has been a contributing factor?

v2/internal/crdmanagement/manager.go
```diff
@@ -323,6 +327,18 @@ func (m *Manager) DetermineCRDsToInstallOrUpgrade(
...
+	goalCRDsWithDifferentMetadata := m.FindNonMatchingCRDs(existingCRDs, filteredGoalCRDs, DesiredMetadataEqual)
+	for name := range goalCRDsWithDifferentMetadata {
+		result, ok := resultMap[name]
+		if !ok {
+			return nil, eris.Errorf("Couldn't find goal CRD %q. This is unexpected!", name)
+		}
+
+		if result.DiffResult == NoDifference {
+			result.DiffResult = MetadataDifferent
+		}
```
theunrepentantgeek, line 346: Is this test code that's left over? If not, what's it for?

v2/internal/crdmanagement/metadata.go
```diff
@@ -0,0 +1,54 @@
...
+			return false
+		}
+	}
+
+	return true
+}
+
+// mergeMetadata preserves existing metadata while allowing desired metadata to take precedence.
+func mergeMetadata(existing map[string]string, desired map[string]string) map[string]string {
+	return lo.Assign(existing, desired)
```
theunrepentantgeek, line 35: This should now be done with the standard library - `maps` contains everything you need.


```suggestion
    maps.Insert(existing, maps.All(desired))
    return existing
```

I don't know whether `lo.Assign()` creates a new map up front or not, so you might want `maps.Clone(existing)` as the first parameter if you want an independent map result.

v2/internal/crdmanagement/metadata.go
```diff
@@ -0,0 +1,54 @@
...
+	desired.Annotations = mergeMetadata(existing.Annotations, desired.Annotations)
+}
+
+// applyCRDLabels merges the user configured labels into the labels of the CRDs loaded from disk.
+// Labels reserved for ASO's own use are never overwritten, as doing so would break CRD discovery and
+// upgrade detection.
+func applyCRDLabels(crds []apiextensions.CustomResourceDefinition, configuredLabels map[string]string) {
+	safeLabels := lo.OmitBy(configuredLabels, func(key string, _ string) bool {
+		return IsReservedLabel(key)
+	})
```
theunrepentantgeek, line 49: Consider [`maps.DeleteFunc()`](https://pkg.go.dev/maps#All).

v2/internal/crdmanagement/metadata_test.go
```diff
@@ -0,0 +1,196 @@
...
+	mergeCRDMetadata(existing, desired)
+
+	g.Expect(desired.Labels).To(Equal(map[string]string{
+		"external-label": "preserved",
+		"managed-label":  "desired",
+	}))
+	g.Expect(desired.Annotations).To(Equal(map[string]string{
+		"external-annotation": "preserved",
+		"managed-annotation":  "desired",
+	}))
```
theunrepentantgeek:
```suggestion
	g.Expect(desired.Labels).To(HaveLen(2))
	g.Expect(desired.Labels).To(HaveKeyWithValue("external-label", "preserved"))
	g.Expect(desired.Labels).To(HaveKeyWithValue("managed-label", "desired"))
	g.Expect(desired.Annotations).To(HaveLen(2))
	g.Expect(desired.Annotations).To(HaveKeyWithValue("external-annotation", "preserved"))
	g.Expect(desired.Annotations).To(HaveKeyWithValue("managed-annotation", "desired"))
```

I won't make this suggestion in other places, but it still applies.

v2/internal/labels/parse.go
```diff
@@ -6,18 +6,64 @@
...
+	if value == "" {
+		return result, nil
+	}
+
+	assignments := strings.FieldsFunc(value, func(r rune) bool {
+		return strings.ContainsRune(Separators, r)
+	})
+
+	for _, assignment := range assignments {
+		key, labelValue, found := strings.Cut(assignment, "=")
```
theunrepentantgeek, line 49: Here would be a good spot to trim leading/trailing whitespace.

v2/internal/labels/parse_test.go
```diff
@@ -50,3 +50,60 @@ func TestParse(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			g := NewWithT(t)
+			actual, err := labels.ParseMap(tt.value)
+			if tt.ok {
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(actual).To(Equal(tt.want))
+			} else {
+				g.Expect(err).To(HaveOccurred())
```
theunrepentantgeek: As mentioned above, using `expectedError: string` instead of `ok: bool` as a trigger would let you verify the errors are the expected ones:

```suggestion
				g.Expect(err).To(MatchError(ContainsSubstring(tt.expectedError)))
```

theunrepentantgeek (commented):
Looks good. Most of my comments are around testing.

v2/internal/crdmanagement/manager.go
mboersma: Not leftover, it's the metadata drift check, and it mirrors the `VersionEqual` block just above it.

Until now `VersionEqual` was the only comparator, so a CRD whose labels or annotations changed didn't get updated unless the version changed too. That's a problem for `--crd-labels`, since adding a label doesn't bump the version. It also fixes an existing gap with the `cert-manager.io/inject-ca-from` annotation, which goes stale if you redeploy ASO into a different namespace.

The `NoDifference` guard just stops this pass from downgrading a CRD that the version check already flagged.

v2/internal/crdmanagement/metadata.go
mboersma: `maps.Insert` panics on a nil map, and `crds[i].Labels` is nil for any CRD that doesn't already carry labels, so this would panic in practice. It also mutates `existing` rather than returning a new map.

`lo.Assign` already allocates a fresh map and handles nil inputs, so I'd prefer to keep it unless you feel strongly. (Also `lo.Assign` here was @matthchr's suggestion, so I want to make sure you two agree before I flip it back.)

v2/internal/crdmanagement/metadata.go
mboersma: Same as above: `maps.DeleteFunc` edits in place, so this would mutate the caller's map. `lo.OmitBy` returns a new map and leaves the input alone, which I think is what we want here.

v2/cmd/asoctl/internal/crd/cleaner.go
mboersma: Yeah `common` is pretty weak. But `pkg/` is public API, there are 5 non-test code importers, and `pkg/common/` also has `annotations`, `cloud`, `config`, and `controllerUtils`. Moving just `labels` leaves things kind of inconsistent.

I'm happy to do that if it's still preferable, although these constants landed in `pkg/common/labels` at @matthchr's suggestion, so we should agree.

v2/internal/crdmanagement/manager.go
mboersma: Whitespace is handled now: `ParseMap` trims around keys and values, so  `" app.kubernetes.io/Version"`  becomes `app.kubernetes.io/Version`. There's a test pinning that a padded reserved label is still rejected.

For case, Kubernetes treats `app.kubernetes.io/Version` as a distinct label from `app.kubernetes.io/version`, so there's no real collision, and the selector in `ListCRDs` won't match it. But nobody does that deliberately, and it's two lines to fold the comparison. I'll add it if you think it's preferable.

v2/cmd/asoctl/internal/crd/cleaner.go
theunrepentantgeek: Have caught up with @matthchr - leave these where they are in this PR. We'll promote the packages (removing the `common` intermediate step) in a later cleanup PR.

v2/internal/crdmanagement/metadata.go
theunrepentantgeek: We have a mild preference to stick with the standard libraries, but it's not a big deal. Leave this the way it is.

v2/internal/crdmanagement/metadata.go
theunrepentantgeek: Agreed. No change needed.

v2/internal/crdmanagement/manager.go
theunrepentantgeek: After discussing with @matthchr offline, we think this would have a positive effect, reducing the memory surge on startup significantly. But, not something to worry about in this PR - Matt has taken a TODO to tackle that separately, after this PR merges.

v2/internal/crdmanagement/manager.go
theunrepentantgeek: Mostly I was thinking about how useful predicates like `IsReservedLabel` tend to get reused in other code paths - potentially paths that don't include checks that current precede its use.

This doesn't need to be case-insensitive, but I'd be happier if it trimmed whitespace.

mboersma: It does now. I updated it to use `TrimSpace` and added some whitespace test cases.

theunrepentantgeek (approved):

theunrepentantgeek:
Looks really good, I have queued this for merging. Thanks @mboersma

PR #5593, opened 2026-07-30
feat: add Managed Cassandra 2026-03-15 for cluster extensions, autoReplicate and scheduledEventStrategy

andreidorin-oprea:
## What this PR does

Adds Managed Instance for Apache Cassandra at API version **2026-03-15**, alongside the existing 2025-10-15.

2026-03-15 is the latest stable `Microsoft.DocumentDB/cassandraClusters` version, and it is the first **stable** version to model three cluster properties that 2025-10-15 does not surface at all:

| Property | 2025-10-15 | 2026-03-15 |
|---|---|---|
| `extensions` | absent | ✅ |
| `autoReplicate` | absent | ✅ |
| `scheduledEventStrategy` | absent | ✅ |

Today an ASO-managed `CassandraCluster` cannot install a Cassandra extension (e.g. `cassandra-lucene-index`), cannot choose which keyspaces are auto-replicated into data centers joining the cluster, and cannot pick how nodes react to Azure scheduled events. Since the cluster PUT is a full property replace, that also means ASO cannot adopt an existing cluster that has any of them set without clearing them.

`CassandraDataCenter` is unchanged between the two versions — the data-center payload is byte-identical, so this PR adds the new version for it purely to keep the pair in step.

`clusterType` is deliberately **not** picked up: it is still `@added(Versions.v2026_04_01_preview)` in the TypeSpec, so no stable version carries it.

### Special notes

cosmos-db has moved to TypeSpec, and from 2026-03-15 it emits a single `openapi.json` with **named** properties bags. That renames the generated types and made this more than a version-block copy:

- `CassandraCluster_Properties_*` → `ClusterResourceProperties*`, `CassandraClusters_DataCenter_Properties_*` → `DataCenterResourceProperties*`. The arm-id / secret / config-map labels are re-declared under the new names.
- The 2025-10-15 types therefore need `$nameInNextVersion` entries — without them, stage 63 fails with `no configuration to rename CassandraCluster_Properties_STATUS to ClusterResourceProperties_STATUS` while building the storage conversions to the new hub.
- The existing `typeTransformers` that drop the readonly-but-not-marked `Deallocated` / `ProvisionError` / `ProvisioningState` from the spec key on the old type names, so equivalents were added for the new ones. Verified the generated `v20260315` specs no longer carry those three.

Down-converting 2026-03-15 → 2025-10-15 round-trips `Extensions`, `AutoReplicate` and `ScheduledEventStrategy` through the property bag, so nothing is lost across the hub.

One incidental generated change: `documentdb/v1api20240815/storage/database_account_types_gen.go` gains `SystemData_STATUS` conversions, because that type is shared across the group and now has a copy in the new version's package.

**The go-vcr recordings are not in this PR.** `Test_DocumentDB_CassandraCluster_v20260315_CRUD` and the sample suite need capturing against a live subscription — the test creates a managed-Cassandra cluster plus the two `Network Contributor` role assignments the Cosmos DB service principal needs on the delegated subnets, which I can't record on the subscription I have. Happy to hand that off, or to point it at a subscription if you can suggest one.

### Verified locally

- `task controller:generate-types` and `task controller:generate-crds` — clean
- `go build ./...`, `go vet ./api/documentdb/... ./internal/controllers/...` — clean
- `go test ./api/documentdb/...` — all pass, including the new `v20260315` packages and the 2025-10-15 conversions to the new hub
- `scripts/v2/check-crd-size.sh` — no CRD over the limit
- `upgradable_resources.md` no longer lists `CassandraCluster` as having a newer stable version available

`golangci-lint` was not run locally (not installed on this machine); the only hand-written Go in the PR is the CRUD test, which is `gofumpt`-clean.

gif

## Checklist

- [x] this PR contains tests


theunrepentantgeek:
I've added recordings and triggered our PR checks.

theunrepentantgeek (approved):

matthchr (approved):

PR #5597, opened 2026-08-03
feat: Allow recording tests in tenants that tag resource groups by policy

andreidorin-oprea:
## What this PR does

Lets recording-mode tests run in a tenant whose policy requires tags on every resource group, and keeps the
recording free of both those tags and the identity that made it.

Recording a test today fails in such a tenant: `CreateTestResourceGroupDefaultTags` sets only `CreatedAt`, so
ARM rejects the resource group with `RequestDisallowedByPolicy` before the test under test gets a chance to
run. Adding the tags unconditionally isn't an option either — they name a tenant, and often a person, and they
would be matched on when the recording is replayed.

So:

- `TEST_RESOURCE_GROUP_TAGS` (a JSON object, e.g. `{"Owner":"someone","Project":"something"}`) is added to the
  tags of the resource groups tests create. It is read as JSON rather than as a list of pairs so a value can
  contain a comma or significant whitespace and still survive the round trip — a value that doesn't would be
  sent to Azure in one form and looked for in the recording in another, leaving it unredacted.
- Those tags are then removed from the recording, matching on key *and* value, and only within a `"tags"`
  object. Matching the value leaves alone a tag a test sets for its own purposes even when the tenant's policy
  mandates a tag of the same name; confining it to `"tags"` objects stops a tag key that happens to name
  another field from taking that field with it. The result is that a recording made in a tagged tenant replays
  in one without the policy, and vice versa.
- Two things that name whoever recorded the test are redacted: the `createdBy` / `lastModifiedBy` values ARM
  stores on a resource, and the caller's Entra object ID that ARM echoes back in a header. Both name a person
  rather than a service principal when the recording wasn't made by CI.

Nothing changes for a tenant without such a policy, or for replay: with the variable unset the redactor is a
no-op.

## Checklist

- [x] this PR contains tests


andreidorin-oprea:
@theunrepentantgeek I had eyes on a couple of api version bumps or new resources. This should allow injecting random tags and redacting the info unblocking at least on my side the recording and stop asking for a maintainer to do it

theunrepentantgeek:
> @theunrepentantgeek I had eyes on a couple of api version bumps or new resources. This should allow injecting random tags and redacting the info unblocking at least on my side the recording and stop asking for a maintainer to do it

Really appreciate your contributions here. I'm a little behind on code reviews but expect to get through your reviews tomorrow (Friday 7 Aug). 🤞

theunrepentantgeek (approved):

matthchr (approved):

PR #5606, opened 2026-08-07
feat: Avoid receiver collision in generated webhook functions

andreidorin-oprea:
## What

The generator names a webhook function's resource parameter `resource`, and names the receiver
after the kind. For a kind whose name ends in `Resource` — `AlertRuleResource`, say — the
receiver is also `resource`, and the generated code redeclares it.

This routes the parameter name through `KnownLocalsSet`, so it becomes `resourceObj` when the
receiver has already taken `resource`.

## Why now

`Microsoft.DatabaseWatcher` has a `SharedPrivateLinkResource` kind, which is the first one in
tree to hit this. Without the fix that group doesn't generate.

## Testing

- New unit test covering both the free case and the colliding one.
- Three golden files updated; the diff is the renamed identifier only.

---

📚 **PR 1 of 4** — adding `Microsoft.DatabaseWatcher`.

| | |
|---|---|
| **1** | **this PR** — webhook receiver collision |
| 2 | #5607 — the resource group |
| 3 | #5608 — reconcile policy on the context |
| 4 | #5609 — starting watchers |

This is the base of the series and depends on nothing. It stands on its own: the collision is a
generator bug independent of the group that exposed it.


v2/tools/generator/internal/functions/common.go
```diff
@@ -27,6 +27,16 @@ func createBodyReturningLiteralString(
 	)
 }
 
+// resourceParameterIdent returns the identifier a webhook function uses for the resource, whether as a
+// parameter (the validators) or as the local the defaulter type-asserts obj into. It renames the
+// identifier when the receiver has already taken it, as happens for a kind called AlertRuleResource.
+func resourceParameterIdent(receiverIdent string, idFactory astmodel.IdentifierFactory) string {
```
theunrepentantgeek: This helper method creates a trip-hazard for future developers.

The whole point of the `KnownLocalsSet` data abstraction is to keep track of the identifiers that are in-scope and ensure that conflicts don't occur. Creating one and immediately discarding it again goes directly that intent - even if (in the current context) there's only one local identifier required.

Delete this helper, inlining it's functionality at the call points.

v2/tools/generator/internal/functions/kubernetes_admissions_defaulter.go
```diff
@@ -223,7 +223,8 @@ func (d *DefaulterBuilder) defaultFunction(
 	receiver astmodel.TypeName,
 	methodName string,
 ) (*dst.FuncDecl, error) {
-	resourceIdent := "resource"
+	receiverIdent := k.IDFactory().CreateReceiver(receiver.Name())
+	resourceIdent := resourceParameterIdent(receiverIdent, k.IDFactory())
```
theunrepentantgeek: Inlining the helper to avoid future maintainers creating their own `KnownLocalsSet`.

```suggestion
	receiverIdent := k.IDFactory().CreateReceiver(receiver.Name())
	knownLocals := astmodel.NewKnownLocalsSet(k.IDFactory())
	knownLocals.Add(receiverIdent)

	resourceIdent := knownLocals.CreateLocal("resource", "", "Obj")
```

theunrepentantgeek (changes requested):
This is a good addition to handle the edge case where things conflict - though the helper method is problematic and needs to be removed.

I suspect the reason we haven't run into this before is that the suffix `Resource` on a resource name is redundant and has been removed.

I'll merge this once the helper has been inlined.

andreidorin-oprea:
Thanks — helper removed and inlined at all three call sites (the defaulter plus both validators), so each one builds and owns its own `KnownLocalsSet` rather than borrowing a discarded one.

The golden files are unchanged, which confirms the generated output is identical. I dropped the helper's unit test along with it; the goldens cover the behaviour.

Agreed on the root cause — the `Resource` suffix being redundant is why this stayed hidden. `Microsoft.DatabaseWatcher` has a `SharedPrivateLinkResource` where the suffix is part of the ARM type name, which is how it surfaced.

theunrepentantgeek (approved):

matthchr (approved):

PR #5607, opened 2026-08-07
Add Microsoft.DatabaseWatcher (v20241001preview)

andreidorin-oprea:
## What

Adds the `databasewatcher` group (`v20241001preview`): `Watcher`, `Target`, and
`SharedPrivateLinkResource`.

## Testing

- Recorded CRUD tests for `Watcher` and for `Target`. The `Target` test also exercises
  `SharedPrivateLinkResource`, since the two have to be torn down in order — the shared private
  link's managed private endpoint must go before the server it points at.
- Samples for all three kinds, with a recorded samples test.
- The `Target` and `SharedPrivateLinkResource` samples are excluded from the samples test: a
  target needs the FQDN of a server that doesn't exist until it's created, and the shared
  private link has the teardown ordering above.

All recordings were made fresh against a live subscription.

## Notes for review

Most of the diff is generated. The hand-written files are the two CRUD tests, the samples, and
the two exclusions in `samples_tester.go`.

---

📚 **PR 2 of 4** — adding `Microsoft.DatabaseWatcher`.

| | |
|---|---|
| 1 | #5606 — webhook receiver collision *(required for this group to generate)* |
| **2** | **this PR** — the resource group itself |
| 3 | reconcile policy on the context |
| 4 | starting watchers |

Stacked on #5606. Until that merges this PR's diff includes its commit; I'll rebase as each
lands. Review the commits after `Avoid receiver collision in generated webhook functions`.


v2/internal/testcommon/samples_tester.go
```diff
@@ -127,6 +127,13 @@ var exclusions = []*regexp.Regexp{
 
 	// Excluding quota as Azure Quota API does not support deletion - quotas are read-only system resources
 	regexp.MustCompile(`quota/.*_quota.yaml`),
+
+	// Excluding databasewatcher target as it needs the FQDN of a server that doesn't exist until created
+	regexp.MustCompile(`databasewatcher/.*_target.yaml`),
```
theunrepentantgeek: We have many samples that create dependencies (usually underneath a `refs` subfolder) in order to test that resources work properly - can this be applied for this resource?

I think you could essentially lift a SQL Server, put it into the `refs` folder of your sample and include a Target resource pointing to it.
* Export the FQDN of the server into a configMap using a CEL expression pointing to `.status.FullyQualifiedDomainName`
* Configure the Target to import the same configMap

v2/api/databasewatcher/v20241001preview/structure.txt
```diff
@@ -0,0 +1,310 @@
...
+
+Watcher: Resource
+├── Owner: resources/v1apiv20191001.ResourceGroup
+├── Spec: Object (8 properties)
+│   ├── AzureName: Validated<string> (1 rule)
+│   │   └── Rule 0: Pattern: "^[a-zA-Z0-9]{1}[a-zA-Z0-9-_]{3,60}$"
+│   ├── Datastore: *Object (7 properties)
+│   │   ├── AdxClusterResourceReference: *genruntime.ResourceReference
+│   │   ├── KustoClusterDisplayName: *string
+│   │   ├── KustoClusterUri: *string
```
theunrepentantgeek, line 236: This should probably be tagged with `$importConfigMapMode: optional` to allow `KustoClusterUri` to be populated with a value pulled from a configMap at runtime.

theunrepentantgeek: Ditto for `KustoDatIngestionUri` and `KustoManagemetnUrl`, I think.

v2/internal/testcommon/samples_tester.go
theunrepentantgeek: Though, I don't see which property of `target` needs to receive the FQDN - this property will need to be configured with `$importConfigMapMode: optional` in the azure-arm YAML file and the generator rerun.

v2/azure-arm/databasewatcher.yaml
```diff
@@ -0,0 +1,18 @@
...
+groupModelConfiguration:
+  2024-10-01-preview:
+    Watcher:
+      $exportAs: Watcher
+      $supportedFrom: v2.21.0
+    Watchers_Target:
+      $exportAs: Target
+      $supportedFrom: v2.21.0
+    Watchers_SharedPrivateLinkResource:
+      $exportAs: SharedPrivateLinkResource
```
theunrepentantgeek: Consider exporting this as simply

```suggestion
      $exportAs: SharedPrivateLink
```

We usually (but not always) drop `Resource` as a suffix, simply because everything is a resource to ASO.

theunrepentantgeek (commented):
Looks really good - a few quick changes to make things dynamic (most notably to introduce selected *FromConfig properties), and then we can run the CI checks.

v2/api/databasewatcher/v20241001preview/structure.txt
andreidorin-oprea: Done — `KustoClusterUri` now carries `$importConfigMapMode: optional`, so a `kustoClusterUriFromConfig` companion is generated alongside it.

andreidorin-oprea: Both done, the same way — `kustoDataIngestionUriFromConfig` and `kustoManagementUrlFromConfig`.

v2/azure-arm/databasewatcher.yaml
andreidorin-oprea: Renamed to `SharedPrivateLink`.

Small knock-on worth noting: with the `Resource` suffix gone, this type's generated webhook receiver is no longer `resource`, so it stops being an instance of the collision #5606 fixed. That does not undermine #5606 — the generator bug is real whether or not this particular resource happens to trip it.

v2/internal/testcommon/samples_tester.go
andreidorin-oprea: `ConnectionServerName` is the one. It now carries `$importConfigMapMode: optional` on all four target flavours — `SqlDbSingleDatabaseTargetProperties`, `SqlDbElasticPoolTargetProperties`, `SqlMiTargetProperties` and `SqlVmTargetProperties` — since every one of them names a server, and wiring up only the flavour the sample happens to use would be arbitrary.

andreidorin-oprea: It can go, and it has. The sample now follows your sketch: a SQL server and its database live under `refs/`, the server publishes its FQDN to a config map, and the target reads it back through `connectionServerNameFromConfig`. Recorded end to end — the target is created against a server that actually exists, and the recording shows the resolved value reaching ARM:

```
"connectionServerName":"asotestdbwserver.database.windows.net"
```

Two deviations worth flagging, both easy to reverse if you'd rather I didn't:

1. `sql.Server` already exposes `fullyQualifiedDomainName` as a first-class `operatorSpec.configMaps` destination, so the sample uses that rather than a CEL expression. Same config map, one less moving part. Happy to switch to `configMapExpressions` if you'd prefer the sample to demonstrate the CEL form.

2. The server has a fixed, distinctive name (`asotestdbwserver`) rather than `aso-sample-…`. Names under `refs/` aren't randomised by the samples tester and a SQL logical server name is globally unique DNS, so this follows the `asotestsqlstorageref` precedent in `v2/samples/sql/v20250101/refs/`.

The `sharedprivatelink` exclusion just below stays: it creates a managed private endpoint on that server, and the endpoint has to be torn down before the server is — an ordering the samples harness doesn't offer, since it only deletes the resource group. The CRUD test covers that ordering explicitly.

andreidorin-oprea:
Thanks — all five addressed, and rebased onto main now that #5606 has merged, so this PR is down to just the databasewatcher work.

The feedback went in as four new commits on top of the original four rather than being squashed in, so the delta is reviewable on its own:

| Commit | What |
| --- | --- |
| `Let watchers and targets take runtime values from config maps` | the `$importConfigMapMode` directives and the `$exportAs` rename |
| `Regenerate databasewatcher types` | generated fallout |
| `Regenerate the databasewatcher reference index` | docs fallout |
| `Build the target sample on a real SQL server` | the sample, the exclusion removal, and the re-recording |

Verified locally: `go build ./...` and `go vet` clean, `golangci-lint` clean on the touched packages, `check_samples.py` green, both databasewatcher CRUD recordings still replay, and the samples suite was re-recorded live and then replays green with no credentials set.

theunrepentantgeek (approved):

PR #5608, opened 2026-08-07
Make the reconcile policy in effect available to resource extensions

andreidorin-oprea:
## What

Records the resolved reconcile policies on the context, and resolves against them in one place
rather than leaving each extension to combine the pieces.

- `reconcilers.ReconcilePolicyFromContext(ctx)` — the policy for the resource being reconciled.
- `reconcilers.ReconcilePolicyForAnnotation(ctx, annotation)` — the policy for some *other*
  resource, given that resource's own annotation.

## Why

A post-reconciliation check still runs when the policy forbids modification, because the skip
path updates status. An extension that acts on Azure — invokes an ARM action, say — currently
has no way to know that: it sees only the resource, while the namespace and the operator supply
the rest of the answer.

Resolution happens here rather than at the call site because it's easy to get backwards. An
annotation that can't be parsed falls back to the **operator's** configured policy, not the
namespace's; an absent annotation takes the **namespace's**. Getting that wrong either strands a
resource nobody acts on, or acts on one that was meant to be left alone.

## Behaviour changes

Both are in `mergeReconcilePolicy`, so worth a careful look:

1. **`Inherited` fails closed.** A namespace that can't be read may be the one saying to skip,
   and acting on a resource we were told to leave alone is worse than not acting on one we could
   have touched.
2. **A namespace read failure no longer returns a raw config value.** It previously returned
   `Config.DefaultReconcilePolicy` directly, which is `""` when unset. It now goes through the
   same fallback as every other path, so the answer is `manage` rather than an empty policy.

## Testing

First unit tests in `internal/reconcilers/generic` — five cases over `mergeReconcilePolicy`
covering object/namespace/default precedence and both unreadable-namespace paths, plus unit
tests for each accessor. The existing recorded `Test_ReconcilePolicy*` suite passes unchanged.

## Review notes

Addressed from the automated review:
- The namespace read used a namespaced key for a cluster-scoped `Namespace`. Corrected, and the
  unit test now exercises a real fake client instead of intercepting the call — reverting the fix
  makes that test fail.
- `handleSkipReconcile` no longer recomputes the policies; it reads the ones already on the
  context, dropping a redundant namespace read.

---

📚 **Standalone.** This does not depend on the `Microsoft.DatabaseWatcher` PRs and touches no
files belonging to them; it was initially opened stacked on #5607 by mistake, which made its diff
look far larger than it is. It is now based directly on `main`: 5 files, ~375 lines.

Its consumer is #5609, which needs to know whether it may invoke an ARM action, but this stands on
its own — the second behaviour change above is a latent correctness issue independent of that.

### DIFF
--- a/docs/hugo/content/contributing/extension-points/post-reconciliation-checker.md
+++ b/docs/hugo/content/contributing/extension-points/post-reconciliation-checker.md
@@ -231,6 +231,11 @@ When testing `PostReconciliationChecker` extensions:
 
 - **Call `next()` if appropriate**: Allows for check chaining (rarely needed)
 - **Don't modify the resource**: This is for validation only
+- **Respect the reconcile policy**: This check runs even when the policy forbids modification, since the
+  skip path still updates status, so an extension that acts on Azure must first call
+  `reconcilers.ReconcilePolicyFromContext`. To act on *another* resource, check that one too with
+  `reconcilers.ReconcilePolicyForAnnotation(ctx, itsAnnotation)`. Resolve it there rather than by hand: an
+  unusable annotation falls back to the operator's policy, not the namespace's
 - **Be patient**: Checks may run many times before succeeding
 - **Use factory methods**: Always uses the factory methods for `PostReconcileCheckResult` to ensure consistency
 - **Provide clear reasons**: Failure messages shown to users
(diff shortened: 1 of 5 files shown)


theunrepentantgeek (changes requested):
Allowing implementations of `PostReconciliationChecker` to access the active reconciliation policy is a sound idea.

But, tunneling the information through via the active Context is an issue - it leads to silent coupling between different parts of the application and can lead to bugs best described as _spooky-action-at-a-distance_.

Given that the reconcile policy is needed only by implementations of `PostReconciliationChecker` (including your new one in PR #5609), I suggest we add ReconcilePolicies to the signature of the extension point, making the dependency explicit.

I _think_ it should be straightforward to plumb the value through the existing call chain(s) to the extension.

andreidorin-oprea:
Superseded by #5625, which landed this work with the reconcile policies passed explicitly rather than tunnelled through the context — the better design. Closing.

Thanks @theunrepentantgeek for taking it the rest of the way.

PR #5609, opened 2026-08-07
Start database watchers once they have a target

andreidorin-oprea:
## What

ARM creates every database watcher stopped, and a stopped watcher collects no data. There is no
spec property to change that — starting one is an ARM action. This adds a
`PostReconciliationChecker` on `Target` that invokes it, plus generic-client support for
long-running POST actions.

## Why the target and not the watcher

ARM refuses to start a watcher until it has at least one target
(`WatcherStartFailedDueToNoTargets`), and a watcher can't tell whether it has one without going
to look. A watcher that started itself speculatively would also hold a POST open on itself, so
the start that *would* have worked then fails as a conflicting operation.

## Notable pieces

- **The action's response body is dropped.** ARM answers an action with the resource it acted
  on, whose `provisioningState` describes the resource and not the action — so an action only
  just accepted reads as already finished and the poller stops before it starts. The body is
  drained (so the connection is reused) and replaced, leaving the poller to follow the
  operation's own status URL.
- **The resume token lives on the target.** Only the resource being reconciled is written back,
  never its owner. An action reports success through its operation and nowhere else, since a
  watcher has no failure state of its own.
- **The operator check comes first.** Operators sharing a cluster have their own policies and
  credentials, none of it visible from here, so a watcher another operator has claimed is left
  alone — settled before anything resolves that watcher's policy against *this* operator's
  configuration.
- **Token redaction is by suffix.** Every resume token carries the signed URL of the operation it
  follows. Matching `-resume-token` rather than naming each annotation also fixes the existing
  `dataprotection` token, which was being logged in full.

## Known limitation

Two targets of the same watcher reconciling at once can both submit a start; the loser gets a
conflict from ARM and retries. Left alone deliberately — the retry converges, and the
alternative is coordination between targets that ASO has no mechanism for.

## Testing

Recorded end-to-end test: watcher created stopped, target created, the operator starts the
watcher, and the watcher is observed `Running` **read back from Azure** rather than from its
possibly-stale status. The recording contains exactly one `POST .../start` and no conflicts.

Unit tests against a fake ARM state machine cover:

- a start submitted exactly once and followed across reconciles;
- a failed operation surfacing as an error;
- a watcher already `Starting` not being started again;
- a foreign watcher refused with zero ARM requests;
- a target whose own policy says skip doing nothing at all;
- a credential mismatch blocking the start but not readiness.

---

📚 **PR 4 of 4** — adding `Microsoft.DatabaseWatcher`.

| | |
|---|---|
| 1 | #5606 — webhook receiver collision |
| 2 | #5607 — the resource group |
| 3 | #5608 — reconcile policy on the context |
| **4** | **this PR** — starting watchers |

Stacked on #5608. Until that merges this PR's diff includes the commits below it; I'll rebase as
each lands. This PR's own change is the last commit.


andreidorin-oprea:
Rebased onto @theunrepentantgeek's #5625, which replaces the context-based policy plumbing from #5608 with an explicit `ReconcilePolicies` parameter on the extension point.

The extension now reads `reconcilePolicies.Effective.AllowsModify()` for the target being reconciled and `policies.ForAnnotation(...)` for the watcher's own policy, and passes the policies through to `next()`.

Note that #5625 is still a draft based on a pre-#5607 `main`, so I replayed its commits onto current `main` to rebase on top — which means the first eight commits here are its, and will disappear from this diff once it merges. Everything from `Start database watchers once they have a target` onward is this PR.

Green locally: build, vet, unit tests, and all three databasewatcher CRUD recordings replay.

andreidorin-oprea:
Rebased onto the latest #5625 (`3c4afe8f`) and ported to the new API — ready for review whenever #5625 lands.

The consumer side now reads:

```go
policy, err := policies.ForResource(watcher)
if err != nil {
    return false, eris.Wrapf(err, "resolving the reconcile policy of watcher %q", watcher.Name)
}

return policy.AllowsModify(), nil
```

An owner always shares its target's namespace, so a mismatch is a question the policies can't answer rather than a refusal to start — the error surfaces instead of being flattened into "not allowed". Added `Test_StartAllowed_givenWatcherInAnotherNamespace_reportsTheMismatch` to cover it.

Green locally: build, vet, `golangci-lint` (0 issues), `task format-code` clean, the unit tests, and all three databasewatcher recordings replay.

Since #5625 now merges `main`, this branch sits directly on top of it — the first commits in the diff are its, and will drop out once it merges.

v2/internal/controllers/databasewatcher_watcher_start_v20241001preview_test.go
```diff
@@ -0,0 +1,135 @@
...
+					SqlDbResourceReference:   tc.MakeReferenceFromResource(sqlDatabase),
+					TargetAuthenticationType: to.Ptr(databasewatcher.TargetAuthenticationType_Aad),
+					TargetType:               to.Ptr(databasewatcher.SqlDbSingleDatabaseTargetProperties_TargetType_SqlDb),
+				},
+			},
+		},
+	}
+
+	// The target isn't ready until the watcher it belongs to is running, so this waits for the start
+	tc.CreateResourceAndWait(target)
```
theunrepentantgeek: When ASO resources are applied to a cluster (e.g. via `kubectl apply`) they all get created at the same time and it's up to ASO to handle the sequencing to ensure they all get created properly.

Because of this, we normally prefer to write these CRUD tests with a single `tc.CreateResourcesAndWait()` call that takes all of the resources and forces ASO to handle the sequencing itself. 

I don't see anything in this PR that would preclude doing this here.

theunrepentantgeek (commented):
Looks good to me, bar one minor nitpick on testing. Given that we are overdue for the v2.21 release and would like to include this, I lean towards merging this as is. 

@matthchr, what do you think?

theunrepentantgeek:
@andreidorin-oprea, I've pushed two commits to your branch to fix broken PR checks. 

* The first updates the sample to use Managed Identity - security policies on our subscriptions block use of username/password authentication for Administrators on SQL servers.
* The second updates the recording file.

With these in place, I expect the PR checks to pass.

v2/internal/controllers/databasewatcher_watcher_start_v20241001preview_test.go
andreidorin-oprea: Done in `84c071f3` — all four resources now go in through a single `CreateResourcesAndWait`.

The one thing standing in the way was the target taking the server's FQDN from `server.Status`, which isn't populated yet when everything is created at once. The server now publishes it to a config map and the target reads it back through `connectionServerNameFromConfig` — the same wiring the sample uses, so the operator does the sequencing rather than the test.

Dropped the intermediate `Status == Stopped` assertion since it can't hold once creation is concurrent; the `"Running"` check at the end is unchanged. Re-recorded, and all three databasewatcher recordings replay green with no credentials.

andreidorin-oprea:
Thanks for pushing the Managed Identity fix and the re-recording — I'd recorded against a subscription that still allows SQL username/password, so that constraint wasn't visible from here.

Addressed the sequencing comment in `84c071f3`.

v2/api/databasewatcher/customizations/target_extensions.go
```diff
@@ -0,0 +1,292 @@
...
+	if err != nil {
+		return extensions.PostReconcileCheckResult{}, err
+	}
+
+	if !allowed {
+		return next(ctx, obj, owner, resourceResolver, armClient, log, reconcilePolicies)
+	}
+
+	// ARM rejects the start (WatcherStartFailedDueToNoDataStore), so say what's missing
+	if watcher.Spec.Datastore == nil {
```
matthchr: minor: Do we actually need this check here? Or would it be better to just issue the request to ARM and propagate the error directly?

v2/api/databasewatcher/customizations/target_extensions.go
```diff
@@ -0,0 +1,292 @@
...
+}
+
+func operatorNamespace(obj genruntime.MetaObject) string {
+	return obj.GetAnnotations()[reconcilers.OperatorNamespaceAnnotation]
+}
+
+func describeOperator(obj genruntime.MetaObject) string {
+	namespace := operatorNamespace(obj)
+	if namespace == "" {
+		return "no namespace it has recorded"
```
matthchr: minor: 
```suggestion
		return "<unknown namespace>"
```

or something similar? I think would read better in the error message

v2/api/databasewatcher/customizations/target_extensions.go
```diff
@@ -0,0 +1,292 @@
...
+	// the hub type has been changed but this extension has not
+	var _ conversion.Hub = target
+
+	// This check still runs when the policy forbids modification, and a target ARM was never given is no
+	// reason to start anything
+	if !reconcilePolicies.Effective.AllowsModify() {
+		return next(ctx, obj, owner, resourceResolver, armClient, log, reconcilePolicies)
+	}
+
+	// A target owned by an ARM ID has no watcher of ours to start
```
matthchr: question: Is this actually true? It seems to me that we could actually make the below code work for arm-id based owners as well. Ahh but the issue is ASO doesn't own that watcher, now I am understanding what this is saying.

I guess actually maybe this would be answered by a minor update to the comment:
```suggestion
	// A target owned by an ARM ID has no ASO-managed watcher to start
```

v2/api/databasewatcher/customizations/target_extensions.go
```diff
@@ -0,0 +1,292 @@
...
+		done, err := resumeStart(ctx, armClient, token)
+		if err != nil {
+			clearStartResumeToken(target)
+
+			return extensions.PostReconcileCheckResult{},
+				eris.Wrapf(err, "cannot start watcher %q, which owns this target", watcher.Name)
+		}
+
+		if !done {
+			return extensions.PostReconcileCheckResultFailure("waiting for the watcher to start"), nil
```
matthchr: minor: Should we have watcher.Name in this error somewhere?

v2/api/databasewatcher/customizations/target_extensions.go
```diff
@@ -0,0 +1,292 @@
...
+		return next(ctx, obj, owner, resourceResolver, armClient, log, reconcilePolicies)
+	}
+
+	// Nothing below holds for a watcher another operator has claimed
+	if reason := foreignWatcher(target, watcher); reason != "" {
+		return extensions.PostReconcileCheckResultFailure(reason), nil
+	}
+
+	// A policy that forbids modifying the watcher forbids starting it
+	allowed, err := startAllowed(reconcilePolicies, watcher)
```
matthchr, line 124: minor: Somewhat confused by check order here.

We check:
- Allowed
- `watcher.Spec.Datastore == nil`

before we check if we even have to issue a start, and error in both cases.

Then, we resume the poller if need be.
Then we check if we need to issue a start
Then we check if the credential is different (error if so)
Then we actually issue the start

Assuming we want to keep the `Spec.Datastore == nil` check, it also seems reasonable to only check that before we actually issue the start, near the credential diff check, as opposed to up here?

I'm somewhat torn on the Allowed check -- it could stay here or it could move to before the actual attempt to issue the start. I _think_ you want to move it because otherwise a scenario like create resources, later annotate with skip (so you can delete or migrate the resources) will be blocked as we'll fail early here blocking start even though start may have happened ages ago.

v2/api/databasewatcher/customizations/target_extensions.go
```diff
@@ -0,0 +1,292 @@
...
+
+// StartPollerResumeTokenAnnotation holds a watcher's start on the target, the resource written back.
+const StartPollerResumeTokenAnnotation = "serviceoperator.azure.com/watcher-start-resume-token"
+
+var _ extensions.PostReconciliationChecker = &TargetExtension{}
+
+// PostReconcileCheck starts the watcher that owns this target. ARM creates every watcher stopped, with no
+// property in the spec to change that, and refuses to start one until it has a target - so the target
+// invokes the action rather than the watcher.
+func (extension *TargetExtension) PostReconcileCheck(
```
matthchr, line 38: I am curious: is there any reason to stop database watchers once they have started?

v2/api/databasewatcher/customizations/target_extensions_test.go
```diff
@@ -0,0 +1,412 @@
...
+const ourOperator = "azureserviceoperator-system"
+
+// managed is the usual case: nothing anywhere says to leave these resources alone
+var managed = annotations.ResolvedReconcilePolicies{
+	Effective:       annotations.ReconcilePolicyManage,
+	NamespacePolicy: annotations.ReconcilePolicyManage,
+	Global:          annotations.ReconcilePolicyManage,
+}
+
+func Test_StartAllowed_givenPolicies_returnsExpectedResult(t *testing.T) {
```
matthchr: minor, style: 
```suggestion
func Test_StartAllowed_GivenPolicies_ReturnsExpectedResult(t *testing.T) {
```

v2/api/databasewatcher/customizations/target_extensions.go
```diff
@@ -0,0 +1,292 @@
...
+		return extensions.PostReconcileCheckResultFailure(
+			fmt.Sprintf("watcher %q has no datastore, so it cannot be started", watcher.Name),
+		), nil
+	}
+
+	// The watcher has no failure state of its own, so a start already submitted is followed to its end
+	if token, submitted := startResumeToken(target); submitted {
+		done, err := resumeStart(ctx, armClient, token)
+		if err != nil {
+			clearStartResumeToken(target)
```
matthchr: 🤖 Could we retain the resume token for retryable polling errors? `resumeStart` can return a 429, 5xx, or transport error while the operation is still in flight. Clearing the token here loses the operation handle, and the next reconcile may submit another start while Azure still reports `Stopped`. Please clear it only after a terminal operation failure and add a transient poll-error test that asserts exactly one start POST.

v2/api/databasewatcher/customizations/target_extensions.go
```diff
@@ -0,0 +1,292 @@
...
+		}
+
+		if token != "" {
+			setStartResumeToken(target, token)
+		}
+	}
+
+	// Stay short of ready so we're asked again, which is how the start is seen to have worked. Nothing is
+	// owned by a target, so this can't withhold anything the start itself needs.
+	return extensions.PostReconcileCheckResultFailure("waiting for the watcher to run"), nil
```
matthchr: 🤖 What happens if the preview API omits `properties.status` or introduces a terminal/non-transitional value? This branch waits forever with a generic message. Please distinguish known transitional states, include the observed status in the condition, and treat empty or unknown states as errors unless the RP contract guarantees they are transient.

v2/api/databasewatcher/customizations/target_extensions.go
```diff
@@ -0,0 +1,292 @@
...
+			return extensions.PostReconcileCheckResultFailure(fmt.Sprintf(
+				"cannot start watcher %q, which asks for %s while this target asks for %s",
+				watcher.Name,
+				describeCredential(watcher),
+				describeCredential(target),
+			)), nil
+		}
+
+		token, err := submitStart(ctx, watcher, armClient, log)
+		if err != nil {
```
matthchr: 🤖 Could an expected `409`/`AnotherOperationInProgress` be handled as a non-error wait? With concurrent targets, a losing target can retry while the watcher still reads `Stopped`, repeatedly POSTing, conflicting, and setting an error condition. A multi-target contention test would help pin the intended convergence behavior.

docs/hugo/content/contributing/extension-points/post-reconciliation-checker.md
```diff
@@ -233,18 +233,27 @@ When testing `PostReconciliationChecker` extensions:
...
   skip path still updates status. Before acting on Azure, inspect the explicit `reconcilePolicies`
-    parameter, which contains policies already resolved by the reconciler. Use
-        `reconcilePolicies.Effective` for the resource being reconciled. To act on another resource in the same
-        namespace, use `reconcilePolicies.ForResource(resource)` and handle the returned error. The method checks
-        that the resource belongs to the namespace for which the policies were resolved. `reconcilePolicies.Namespace`
-        is the policy resolved from that namespace's annotation or the global policy, while
-        `reconcilePolicies.Global` is resolved from operator configuration or the built-in default
+  parameter, which contains policies already resolved by the reconciler. Use
+  `reconcilePolicies.Effective` for the resource being reconciled. To act on another resource in the same
+  namespace, use `reconcilePolicies.ForResource(resource)` and handle the returned error. The method checks
```
matthchr, line 245: 🤖 `ResolvedReconcilePolicies` exposes `NamespacePolicy`, not `Namespace`. Please update this identifier so extension authors do not get a compile error when following the documentation.

v2/api/databasewatcher/customizations/target_extensions_test.go
matthchr: Same stylistic comment applies to all these tests.

matthchr: You could make the arugment that the test could most correctly be called 
```suggestion
func Test_startAllowed_GivenPolicies_ReturnsExpectedResult(t *testing.T) {
```

(if you want to capture the fact that the `startAllowed` function is not exported in the test name), but the way it's cased atm seems strange

v2/internal/controllers/databasewatcher_watcher_start_v20241001preview_test.go
```diff
@@ -0,0 +1,146 @@
...
+	databasewatcher "github.com/Azure/azure-service-operator/v2/api/databasewatcher/v20241001preview"
+	sql "github.com/Azure/azure-service-operator/v2/api/sql/v20211101"
+	"github.com/Azure/azure-service-operator/v2/internal/testcommon"
+	"github.com/Azure/azure-service-operator/v2/internal/util/to"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+)
+
+// ARM creates every watcher stopped and refuses to start one without a data store or a target, so this
+// supplies both and expects the operator to start it
+func Test_DatabaseWatcher_WatcherStart_v20241001preview(t *testing.T) {
```
matthchr: I'm confused how this test is fundamentally different than `Test_DatabaseWatcher_Target_v20241001preview_CRUD`

Doesn't that test do basically this? Could we just combine them and avoid having 2x tests executing the start path?

v2/internal/genericarmclient/post_action_test.go
```diff
@@ -0,0 +1,192 @@
...
+	asometrics "github.com/Azure/azure-service-operator/v2/internal/metrics"
+	"github.com/Azure/azure-service-operator/v2/internal/testcommon/creds"
+)
+
+const (
+	postActionResourceID = "/subscriptions/12345/resourceGroups/myrg/providers/Microsoft.Fake/fakeResource/fake"
+	postActionAPIVersion = "2019-01-01"
+)
+
+func Test_BeginPostActionByID_givenAction_postsToTheActionURL(t *testing.T) {
```
matthchr: minor: 
```suggestion
func Test_BeginPostActionByID_GivenAction_PostsToTheActionURL(t *testing.T) {
```

matthchr: Applies throughout

v2/internal/reconcilers/common_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"github.com/go-logr/logr/funcr"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	resources "github.com/Azure/azure-service-operator/v2/api/resources/v1api20200601"
+)
+
+// An extension's token is logged with the rest of the annotations, so redaction has to hold for one this
+// package never named
+func Test_LogObj_givenResumeTokenAnnotation_doesNotLogIt(t *testing.T) {
```
matthchr: minor, style:
```suggestion
func Test_LogObj_GivenResumeTokenAnnotation_DoesNotLogIt(t *testing.T) {
```

appleis below as well

v2/internal/genericarmclient/post_action_test.go
theunrepentantgeek: Addressed.

v2/api/databasewatcher/customizations/target_extensions.go
theunrepentantgeek: Comment updated.

v2/api/databasewatcher/customizations/target_extensions.go
theunrepentantgeek: After a discussion with @matthchr, I've rearranged the blocks in `PostReconcileCheck()` to better reflect lessons we've learned from other reconcilers (both in ASO and other operators).

v2/api/databasewatcher/customizations/target_extensions.go
theunrepentantgeek: Including this check helps us avoid burning ARM Request quota on requests that can't possibly succeed. I've added a comment to this effect.

v2/api/databasewatcher/customizations/target_extensions.go
theunrepentantgeek: I don't see a reliable way to decide whether a 429 5xx (or whatever) is a _problem_ with the operation or the _result_ of the operation.

Clearing the token is fine, as we'll pick up the watcher on the next reconcile and hit one of the following conditions:

* status is `Started` - job don
* status is `Stopped` - we'll try to start it again, creating a new operation
* status is something else (`Running` `Cancelling` etc), we'll wait for the transient condition to terminate

v2/api/databasewatcher/customizations/target_extensions.go
theunrepentantgeek: Yes, done.

v2/api/databasewatcher/customizations/target_extensions.go
theunrepentantgeek: With the test above for foreign ownership, we should never have concurrent targets conflicting.

v2/api/databasewatcher/customizations/target_extensions.go
theunrepentantgeek: Changed, but without the chevrons.

v2/api/databasewatcher/customizations/target_extensions_test.go
theunrepentantgeek: Osherove test naming (which is what we're using) normally uses leading Capitals for each section. 

I don't normally worry about matching letter case for exported/unexported because the test should be unambiguous anyway ... if we ever have code with both `StartAllowed()` and `startAllowed()` in the same scope, there are bigger issues to solve.

v2/internal/controllers/databasewatcher_watcher_start_v20241001preview_test.go
theunrepentantgeek: Agree, we don't need an isolated test just to check watchers start; I'll combine them.

v2/internal/reconcilers/common_test.go
theunrepentantgeek: Updated.

docs/hugo/content/contributing/extension-points/post-reconciliation-checker.md
theunrepentantgeek: Updated.

theunrepentantgeek (approved):
I've gone through and addressed all of the feedback from @matthchr, simply to expedite things so we can merge this and do the (late) v2.21 release. 

Thanks @andreidorin-oprea for all your contributions to this release, very much appreciated.

andreidorin-oprea:
Thank you @theunrepentantgeek — and no complaints at all about you taking the last round yourself to get the release out.

Genuinely grateful for how much you put into this one: the `ReconcilePolicies` refactor in #5625, catching the cross-namespace hole in `ForAnnotation`, the Managed Identity fix when my recordings turned out to depend on a subscription that still allows SQL password auth, and now folding in @matthchr's feedback. The end result is a good deal better than what I opened, and I picked up a fair amount along the way.

Thanks to @matthchr too for the thorough reviews across the stack.

Happy to pick up any follow-ups after v2.21 ships.

matthchr (approved):

matthchr (approved):

PR #5612, opened 2026-08-10
feat: Enhance the Deleter resource extension interface

### LINKED ISSUE #3471: Extend delete extension point to more easily allow users to hook the actual delete entirely
#3467 merged support for deleting subscriptions owned by APIM products, but it did so in a way that results in a double DELETE HTTP request being sent.

It did this because it's easier to call `next` and delegate the tracking of the various delete states to ASO. But it would be ideal if there were a way to allow the extension to totally own the delete (thereby avoiding double-delete).

**Describe the improvement**
@theunrepentantgeek suggested some sort of return-result paradigm where the extension could signal to the caller if it did or did not issue the delete (and possibly return a future to poll if it did, for LRO cases?), and then the ASO infrastructure could do the bookkeeping.

If we do the above, we should update the changes made in #3467 to move to a single delete. This has advantages for error reporting as well as API throttling.

theunrepentantgeek:
## What this PR does

Modifies the existing `extensions.Deleter` extension point by introducing a dedicated `DeleteResponse` struct that allows extensions to integrate with greater nuance.

As a part of this, handling for long-running-operations (those returning a PollerResponse) is now accessible by extensions, instead of being supported only for internal controller actions.

Closes #3471 

gif

## Checklist

- [x] this PR contains tests

### DIFF
--- a/docs/hugo/content/contributing/extension-points/deleter.md
+++ b/docs/hugo/content/contributing/extension-points/deleter.md
@@ -6,13 +6,39 @@ weight: 30
 
 ## Description
 
-`Deleter` allows resources to customize how the reconciler deletes them from Azure. This extension is invoked when a resource has a deletion timestamp in Kubernetes (indicating the user wants to delete it) and gives the resource control over the deletion process.
+`Deleter` allows a resource to customize how ASO deletes it from Azure. ASO invokes the extension after Kubernetes sets the resource's deletion timestamp and before issuing the standard ARM DELETE request.
 
-The interface is called after Kubernetes marks the resource for deletion but before the standard ARM DELETE operation. This allows resources to perform cleanup, handle special deletion scenarios, or coordinate multiple deletion operations.
+The extension returns an `extensions.DeleteResult` describing what ASO should do next. This allows the extension to complete or block deletion, delegate to the standard ARM deletion path, or start a long-running operation that ASO monitors.
 
 ## Interface Definition
 
-See the [Deleter interface definition](https://github.com/Azure/azure-service-operator/blob/main/v2/pkg/genruntime/extensions/deleter.go) in the source code.
+See the [Deleter interface definition](https://github.com/Azure/azure-service-operator/blob/main/v2/pkg/genruntime/extensions/deleter.go) and [DeleteResult definition](https://github.com/Azure/azure-service-operator/blob/main/v2/pkg/genruntime/extensions/delete_result.go) in the source code.
+
+`Delete()` has the following signature:
+
+```go
+Delete(
+    ctx context.Context,
+    log logr.Logger,
+    resolver *resolver.Resolver,
+    armClient *genericarmclient.GenericClient,
+    obj genruntime.ARMMetaObject,
+    next extensions.DeleteFunc,
+) (extensions.DeleteResult, error)
+```
+
+## Deletion Results
+
+Return one of the following results to describe the deletion state:
+
+| Result | Meaning |
+| --- | --- |
+| `extensions.ProceedWithDelete()` | Continue a composed deletion flow. Most extensions should call `next()` and return its result instead. |
+| `extensions.BlockDelete(message)` | Keep the finalizer and retry later. ASO exposes `message` in the resource's Ready condition. |
+| `extensions.DeleteCompleted()` | Deletion is complete. ASO can remove the finalizer without issuing another ARM DELETE request. |
+| `extensions.MonitorDelete(pollerResponse)` | A long-running deletion operation has started. ASO stores its resume token and requeues until it completes. |
+
+Return `extensions.DeleteResult{}` with an error. ASO ignores the result when the error is non-nil and retries reconciliation according to the error classification.
 
 ## Motivation
 
@@ -29,19 +55,19 @@ The `Deleter` extension exists to handle cases where:
 
 Implement `Deleter` when:
 
-- ✅ Pre-deletion operations must be performed (e.g., canceling, disabling)
-- ✅ Multiple Azure API calls are needed for complete deletion
-- ✅ Deletion order matters across related resources
-- ✅ Custom error handling is needed during deletion
-- ✅ Soft-delete or purge operations require special logic
-- ✅ The resource should be preserved in Azure in some scenarios
+- Pre-deletion operations must be performed, such as canceling or disabling a resource.
+- Multiple Azure API calls are needed for complete deletion.
+- Deletion order matters across related resources.
+- Custom error handling is needed during deletion.
+- Soft-delete or purge operations require special logic.
+- The resource should be preserved in Azure in some scenarios.
 
 Do **not** use `Deleter` when:
 
-- ❌ The standard DELETE operation works correctly
-- ❌ You only need to clean up Kubernetes resources (use finalizers)
-- ❌ The logic should apply to all resources (modify the controller)
-- ❌ You're working around an Azure API bug (fix/report the bug)
+- The standard DELETE operation works correctly.
+- You only need to clean up Kubernetes resources; use finalizers instead.
+- The logic should apply to all resources; modify the controller instead.
+- You're working around an Azure API bug; fix or report the bug instead.
 
 ## Example: Subscription Alias Deletion
 
@@ -68,13 +94,13 @@ func (ex *ResourceExtension) Delete(
     armClient *genericarmclient.GenericClient,
     obj genruntime.ARMMetaObject,
     next extensions.DeleteFunc,
-) (ctrl.Result, error) {
+) (extensions.DeleteResult, error) {
     resource := obj.(*myservice.MyResource)
 
     // Perform cleanup operation
     log.V(Status).Info("Performing pre-deletion cleanup")
     if err := ex.performCleanup(ctx, resource, armClient); err != nil {
-        return ctrl.Result{}, eris.Wrap(err, "cleanup failed")
+        return extensions.DeleteResult{}, eris.Wrap(err, "cleanup failed")
     }
 
     // Proceed with standard deletion
@@ -92,14 +118,14 @@ func (ex *ResourceExtension) Delete(
     armClient *genericarmclient.GenericClient,
     obj genruntime.ARMMetaObject,
     next extensions.DeleteFunc,
-) (ctrl.Result, error) {
+) (extensions.DeleteResult, error) {
     resource := obj.(*myservice.MyResource)
 
     // Check if resource should be preserved in Azure
     if ex.shouldPreserve(resource) {
         log.V(Status).Info("Skipping Azure deletion, resource marked for preservation")
-        // Return success without calling next() - finalizer will be removed
-        return ctrl.Result{}, nil
+        // Report completion without calling next(), allowing ASO to remove the finalizer.
+        return extensions.DeleteCompleted(), nil
     }
 
     // Proceed with normal deletion
@@ -107,7 +133,9 @@ func (ex *ResourceExtension) Delete(
 }
 ```
 
-### Pattern 3: Soft Delete with Purge Option
+### Pattern 3: Temporarily Block Deletion
+
+Use `BlockDelete()` when a prerequisite can become ready later. ASO keeps the finalizer, updates the Ready condition with the supplied message, and retries the deletion.
 
 ```go
 func (ex *ResourceExtension) Delete(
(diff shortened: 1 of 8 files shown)


v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -184,7 +184,49 @@ func (r *azureDeploymentReconcilerInstance) StartDeleteOfResource(ctx context.Co
...
+			msg := fmt.Sprintf(
+				"Resource deletion blocked: %s",
+				result.Message(),
+			)
+
+			r.Log.V(Verbose).Info(msg)
+
+			return ctrl.Result{}, result.CreateConditionError()
+		}
+
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

v2/pkg/genruntime/extensions/delete_result.go
```diff
@@ -0,0 +1,127 @@
...
+	pollerResponse *genericarmclient.PollerResponse[genericarmclient.GenericDeleteResponse]
+}
+
+// ProceedWithDelete is returned if normal deletion of the resource should proceed.
+// with action `Proceed`.
+func ProceedWithDelete() DeleteResult {
+	return DeleteResult{
+		action: deleteResultTypeProceed,
+	}
+}
```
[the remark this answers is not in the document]
theunrepentantgeek: Removed, as unused.

v2/pkg/genruntime/extensions/delete_result.go
```diff
@@ -0,0 +1,127 @@
...
+
+// OperationID returns the operation ID associated monitoring the deletion operation, if any.
+func (r DeleteResult) OperationID() (string, bool) {
+	if r.pollerResponse == nil {
+		return "", false
+	}
+
+	return r.pollerResponse.ID, true
+}
```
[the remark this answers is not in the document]
theunrepentantgeek: Applied.

v2/pkg/genruntime/extensions/deleter.go
```diff
@@ -40,7 +39,7 @@ type Deleter interface {
 		resolver *resolver.Resolver,
 		armClient *genericarmclient.GenericClient,
 		obj genruntime.ARMMetaObject,
-		next DeleteFunc) (ctrl.Result, error)
+		next DeleteFunc) (DeleteResult, error)
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

v2/api/apimanagement/customizations/product_extensions.go
```diff
@@ -42,19 +41,19 @@ func (extension *ProductExtension) Delete(
...
 	id, err := genruntime.GetAndParseResourceID(typedObj)
 	if err != nil {
-		return ctrl.Result{}, eris.Wrapf(err, "failed to get the ARM ResourceId for %s", productName)
+		return extensions.DeleteResult{}, eris.Wrapf(err, "failed to get the ARM ResourceId for %s", productName)
 	}
 
 	if id.Parent == nil {
-		return ctrl.Result{}, eris.Wrapf(err, ". APIM Product had no parent ID: %s", id.String())
+		return extensions.DeleteResult{}, eris.Wrapf(err, ". APIM Product had no parent ID: %s", id.String())
 	}
```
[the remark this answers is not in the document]
theunrepentantgeek: Fixed.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -182,9 +182,62 @@ func (r *azureDeploymentReconcilerInstance) StartDeleteOfResource(ctx context.Co
...
+		msg := fmt.Sprintf(
+			"Resource deletion blocked: %s",
+			result.Message(),
+		)
+
+		r.Log.V(Verbose).Info(msg)
+
+		return ctrl.Result{}, result.CreateConditionError()
+
+	default:
```
theunrepentantgeek, line 252: 🤖 TheUnrepentantRobot says:

**High:** This default makes every result other than Block or Monitor—including `DeleteResult{}`, the unused Proceed state, and any future state—mean deletion completed. An extension can accidentally return the zero value with a nil error and cause ASO to remove the finalizer while the Azure resource still exists. Please switch explicitly over Complete/Block/Monitor and return an error for zero or unknown actions.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -182,9 +182,62 @@ func (r *azureDeploymentReconcilerInstance) StartDeleteOfResource(ctx context.Co
...
+			return ctrl.Result{},
+				eris.Wrapf(err, "couldn't create DELETE resume token for resource %q", r.Obj.AzureName())
+		}
+
+		operationId, ok := result.OperationID()
+		if !ok {
+			return ctrl.Result{}, eris.Errorf("couldn't get operation ID for resource %q", r.Obj.AzureName())
+		}
+
+		SetPollerResumeToken(r.Obj, operationId, token)
```
theunrepentantgeek, line 231: 🤖 TheUnrepentantRobot says:

**High:** `MonitorDelete()` accepts and persists an arbitrary operation ID here, but `DetermineDeleteAction()` only resumes monitoring when that ID equals `genericarmclient.DeletePollerID`. A custom purge poller, including the pattern documented in this PR, will rerun the deleter rather than resume the stored operation. Please either add a canonical resume strategy for custom operation IDs or stop exposing arbitrary pollers through `MonitorDelete()`.

docs/hugo/content/contributing/extension-points/deleter.md
```diff
@@ -6,13 +6,39 @@ weight: 30
...
+) (extensions.DeleteResult, error)
+```
+
+## Deletion Results
+
+Return one of the following results to describe the deletion state:
+
+| Result | Meaning |
+| --- | --- |
+| `extensions.ProceedWithDelete()` | Continue a composed deletion flow. Most extensions should call `next()` and return its result instead. |
```
theunrepentantgeek: 🤖 TheUnrepentantRobot says:

**High:** This documents `extensions.ProceedWithDelete()`, but that constructor does not exist and the reconciler has no Proceed handling. The unused `deleteResultTypeProceed` makes the contract more misleading. Please remove this result and keep continuation expressed by calling and returning `next()`.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -240,27 +293,28 @@ func (r *azureDeploymentReconcilerInstance) DeleteNotPossibleInAzure(ctx context
...
-		err = eris.Wrap(err, msg)
-	}
-
-	return ctrl.Result{},
-		conditions.NewReadyConditionImpactingError(
-			err,
-			conditions.ConditionSeverityWarning,
-			conditions.ReasonDeletionNotSupported,
-		)
+	return extensions.BlockDelete(msg), nil
```
theunrepentantgeek: 🤖 TheUnrepentantRobot says:

**Medium:** This changes the canonical failure from `ReasonDeletionNotSupported` to the generic `ReasonReconcileBlocked`, and any non-404 error returned by `getStatus()` above is now discarded. That loses the underlying ARM diagnostic and changes condition/retry semantics. Please preserve the status error and allow this blocking result to carry `ReasonDeletionNotSupported`.

theunrepentantgeek (commented):
🤖 TheUnrepentantRobot says:

Deep code-quality review: these four issues should be addressed before merge. GitHub does not allow authors to request changes on their own PR, so this is submitted as a comment review.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
theunrepentantgeek: Done.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
theunrepentantgeek: We don't _currently_ need other kinds of poller, but I've added a check to catch that situation.

docs/hugo/content/contributing/extension-points/deleter.md
theunrepentantgeek: Updated docs.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
theunrepentantgeek: Fixed.

docs/hugo/content/contributing/extension-points/deleter.md
```diff
@@ -6,13 +6,40 @@ weight: 30
...
+
+| Result | Meaning |
+| --- | --- |
+| `extensions.BlockDelete(message)` | Keep the finalizer and retry later. ASO exposes `message` in the resource's Ready condition. |
+| `extensions.DeleteCompleted()` | Deletion is complete. ASO can remove the finalizer without issuing another ARM DELETE request. |
+| `extensions.MonitorDelete(pollerResponse)` | A long-running deletion operation has started. ASO stores its resume token and requeues until it completes. |
+
+Return `extensions.DeleteResult{}` with an error. ASO ignores the result when the error is non-nil and retries reconciliation according to the error classification.
+
+If you want the deletion to proceed, call the `next()` function, which invokes the standard ARM DELETE operation. The result from `next()` can be returned directly or modified before returning.
```
matthchr, line 42: > If you want the deletion to proceed, call the `next()` function, which invokes the standard ARM DELETE operation. The result from `next()` can be returned directly or modified before returning.

Slightly confused (might get clarified as I get further in the PR), when should I do `extensions.MonitorDelete(pollerResponse)` and when should I do `next`?

Aren't they both "the deletion proceeding"?

v2/api/apimanagement/customizations/product_extensions.go
```diff
@@ -71,9 +71,10 @@ func (extension *ProductExtension) Delete(
 	if err != nil {
 		// If the product (or its parent) is already gone, treat as success
 		if genericarmclient.IsNotFoundError(err) {
-			return next(ctx, log, resolver, armClient, obj)
+			return extensions.DeleteCompleted(), nil
```
matthchr, line 74: is it actually safe to call DeleteCompleted here? How do we know that `next` is done? I guess we know because we are the extension and the `next` we've been passed is the standard delete impl, and since the err is `NotFound` it means the resource is gone and we know that's what `next` is going to try to delete anyway?

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
```diff
@@ -182,9 +182,79 @@ func (r *azureDeploymentReconcilerInstance) StartDeleteOfResource(ctx context.Co
...
+
+		SetPollerResumeToken(r.Obj, operationId, token)
+
+		// Normally don't need to set both of these fields but because retryAfter can be 0 we do
+		return ctrl.Result{
+			Requeue:      true,
+			RequeueAfter: result.RetryAfter(),
+		}, nil
+
+	case result.BlockDeletion():
```
matthchr, line 239: Do we actually need this case? Is this not just making `err` a `ReadyConditionImpactingError` of the appropriate type? What's the advantage of encoding that in a BlockDeletion rather than having the extension do:
```
return DeleteResult{}, conditions.MakeReadyConditionImpactingError(...)
```

(which seems slightly more idiomatic Go to me, though maybe only slightly?)

v2/pkg/genruntime/extensions/delete_result_test.go
```diff
@@ -0,0 +1,122 @@
...
+	g.Expect(err).ToNot(HaveOccurred())
+	result := MonitorDelete(&genericarmclient.PollerResponse[genericarmclient.GenericDeleteResponse]{Poller: poller})
+
+	token, err := result.OperationToken()
+
+	g.Expect(err).ToNot(HaveOccurred())
+	g.Expect(token).ToNot(BeEmpty())
+}
+
+func TestDeleteResult_CreateConditionError(t *testing.T) {
```
matthchr, line 105: minor: as mentioned not sold we need to bother with this variant.

matthchr (approved):
Approved with some comments

docs/hugo/content/contributing/extension-points/deleter.md
theunrepentantgeek: Calling `next()` means "I want ASO to do the normal DELETE`

Returning `extensions.MonitorDelete(pollerResponse)` is for when the extension started a long-running-operation to handle the DELTETE and wants ASO to monitor that operation until it finishes.

v2/api/apimanagement/customizations/product_extensions.go
theunrepentantgeek: Correct - we've just tried to delete the product ourselves, and it failed with a 404, so we know it's already gone.

v2/internal/reconcilers/arm/azure_generic_arm_reconciler_instance.go
theunrepentantgeek: It's a design choice - the difference between

* the delete extension failed & we returned an error explaining why; and
* the delete extension correctly determined that the DELETE shouldn't happen yet

The call to `result.CreateConditionError()` creates exactly the error you're suggesting, but this way (a) the user don't have to worry about getting it right, and (b) the code is more declarative about what's happening with the DELETE.

Issue #5529, opened 2026-07-08
Release Plan v2.21


theunrepentantgeek:
Tracking issue for final changes in the v2.21 release planned for this week.

_Everything below is up for discussion, please comment below and `@tag` us as needed._

# Committed work in progress

These in-flight PRs will be included in v2.21 unless something serious blocks them.

- ✅ PR #5232 
- ~~PR #5462~~ Pulled, rescheduled for v2.22
- ✅ PR #5553 
- ✅ PR #5554 -- design only; implementation will be in a later release.
- ✅ PR #5620 
- ✅ PR #5628

# Desired work in progress

We would like to include these PRs in v2.21 if practical.

- ~~PR #5612~~ - contingent the rework being completed
- ✅ PR #5625
- ✅ PR #5609 - once we have PR #5625 merged
- ✅ PR #5615

# Deferred

We don't currently believe these PRs will be ready for the release - please discuss below (or on the PR) if you think this is mistaken.

* PR #5564 - we are keen to support `VaultKey` but think this PR needs more work before it's ready for testing and release. There's also a dependency on PR #5612 to consider.

# Breaking changes

These breaking changes will be included in v2.21.

## containerservice

The property `managedCluster.aadProfile.ServerAppSecret` is now correctly identified as a secret.

## synapse

The property `workspace.encryption.kekIdentity.useSystemAssignedIdentity` is now correctly typed as bool instead of v1.JSON (no effect on valid configurations, but might break invalid ones)

# Upcoming breaking changes

These breaking changes will be included in future releases

## TLS 1.3 (v2.22)

In v2.22 we'll default to using TLS 1.3, but will provide a command line flag to enable TLS 1.2 if needed.

PR #5620, opened 2026-08-11
feat: add --tls-min-version flag for webhook and metrics servers

marek-veber:
## Summary

- Add a `--tls-min-version` flag (default: `VersionTLS12`) to the ASO controller manager, allowing operators to configure the minimum TLS version used by both the webhook server (port 9443) and the metrics server (port 8443).
- Pass `TLSOpts` with the configured `MinVersion` to both `webhook.Options` and metrics `server.Options` in controller-runtime.
- This enables deployments requiring TLS 1.3 only (e.g. OpenShift Modern TLS profile) to enforce it via `--tls-min-version=VersionTLS13`.

## Motivation

TLS scans configured for the Modern TLS Profile show that ASO controller manager endpoints fail compliance because they offer TLS 1.2 with legacy ciphers (including SHA-1 based ones like `ECDHE-*-AES*-SHA`). Currently there is no way to configure the minimum TLS version — the webhook and metrics servers rely on Go's default (TLS 1.2) with no `TLSOpts` set.

Other Kubernetes controllers in the same ecosystem already support this. For example, CAPZ (Cluster API Provider Azure) exposes a `--tls-min-version` flag via the upstream [cluster-api `util/flags` package](https://github.com/kubernetes-sigs/cluster-api/blob/main/util/flags/manager.go#L60), which applies `TLSOpts` to both its webhook and metrics servers. This PR brings ASO in line with that pattern.

## Test plan

- [x] `go build ./cmd/controller/...` passes
- [x] `go test ./cmd/controller/app/...` passes
- [ ] Deploy with `--tls-min-version=VersionTLS13` and verify TLS scan shows TLS 1.3 only on ports 8443 and 9443
- [ ] Deploy with default (`VersionTLS12`) and verify backward compatibility
- [ ] Verify invalid values (e.g. `VersionTLS11`) are rejected at startup

### DIFF
--- a/v2/cmd/controller/app/flags.go
+++ b/v2/cmd/controller/app/flags.go
@@ -6,6 +6,7 @@ Licensed under the MIT license.
 package app
 
 import (
+	"crypto/tls"
 	"flag"
 	"fmt"
 
@@ -27,6 +28,7 @@ type Flags struct {
 	CRDManagementMode    string
 	CRDPatterns          string // This is a ';' delimited string containing a collection of patterns
 	CRDLabels            string // This is a ',' or ';' delimited string containing labels to apply to managed CRDs
+	TLSMinVersion        string
 }
 
 // parseCRDLabels parses the --crd-labels flag, rejecting any label reserved for ASO's own use.
@@ -48,7 +50,7 @@ func parseCRDLabels(value string) (map[string]string, error) {
 
 func (f Flags) String() string {
 	return fmt.Sprintf(
-		"MetricsAddr: %s, SecureMetrics: %t, ProfilingMetrics: %t, MetricsCertDir: %s, HealthAddr: %s, WebhookPort: %d, WebhookCertDir: %s, EnableLeaderElection: %t, CRDManagementMode: %s, CRDPatterns: %s, CRDLabels: %s",
+		"MetricsAddr: %s, SecureMetrics: %t, ProfilingMetrics: %t, MetricsCertDir: %s, HealthAddr: %s, WebhookPort: %d, WebhookCertDir: %s, EnableLeaderElection: %t, CRDManagementMode: %s, CRDPatterns: %s, CRDLabels: %s, TLSMinVersion: %s",
 		f.MetricsAddr,
 		f.SecureMetrics,
 		f.ProfilingMetrics,
@@ -60,6 +62,7 @@ func (f Flags) String() string {
 		f.CRDManagementMode,
 		f.CRDPatterns,
 		f.CRDLabels,
+		f.TLSMinVersion,
 	)
 }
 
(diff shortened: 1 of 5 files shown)


theunrepentantgeek (approved):

v2/cmd/controller/app/flags.go
```diff
@@ -80,6 +83,20 @@ func InitFlags(flagSet *flag.FlagSet) *Flags {
 		"Instructs the operator on how it should manage the Custom Resource Definitions. One of 'auto', 'none'")
 	flagSet.StringVar(&result.CRDPatterns, "crd-pattern", "", "Install these CRDs. CRDs already in the cluster will also always be upgraded.")
 	flagSet.StringVar(&result.CRDLabels, "crd-labels", "", "Comma-separated (or semicolon-separated) labels to apply to all managed CRDs (for example, example.com/owner=aso,environment=production). Labels reserved by the operator (app.kubernetes.io/name, app.kubernetes.io/version and the serviceoperator.azure.com/ prefix) cannot be set.")
+	flagSet.StringVar(&result.TLSMinVersion, "tls-min-version", "VersionTLS12", "The minimum TLS version in use by the webhook and metrics servers. Possible values: VersionTLS12, VersionTLS13.")
```
matthchr: We're a bit inconsistent on when we use flags vs `v2/internal/config/vars.go`, but I think in general tend to prefer config.

It has a more robust "parse + verify" approach as well that you can just plug into.

Can you swap over to use that instead of flags? (Can use the same defaults) w/ `envOrDefault`.

matthchr: Can you also make sure to add a few basic unit-tests for a case like TLS13 (succeeeds) vs TLSabcd (fails, not a valid TLS version)? Code looks right to me but unit test in `v2/internal/config/vars_test.go` would be ideal.

matthchr (commented):
Thanks for this PR and the callout that we are not doing this. Great improvement.

Had a few minor logistical comments, but overall looks great to me. Be aware that we are doing a release soon (see https://github.com/Azure/azure-service-operator/issues/5529). We will make an effort to wait for this PR to merge for that release. If you don't have time to get to these comments in a timely fashion, please let us know and we will see what we can do to help.

v2/cmd/controller/app/flags.go
marek-veber: Done in 9f10fde. Moved the TLS min version handling out of `flags.go` into `v2/internal/config/vars.go`: added a `TLS_MIN_VERSION` env var read via `envOrDefault` (same `VersionTLS12` default) and a `ParseTLSMinVersion` helper using the parse+verify approach. `setup.go` now consumes `cfg.TLSMinVersion`, and the `--tls-min-version` flag / `Flags.TLSVersion()` were removed. Should be safe to resolve.

marek-veber: Added in 9f10fde — table-driven `Test_ParseTLSMinVersion` in `v2/internal/config/vars_test.go` covering VersionTLS12/VersionTLS13 (succeed) and TLSabcd / VersionTLS11 / empty (fail). Should be safe to resolve.

v2/cmd/controller/app/flags.go
```diff
@@ -80,6 +83,20 @@ func InitFlags(flagSet *flag.FlagSet) *Flags {
...
+var tlsVersionMap = map[string]uint16{
+	"VersionTLS12": tls.VersionTLS12,
+	"VersionTLS13": tls.VersionTLS13,
+}
+
+func (f Flags) TLSVersion() (uint16, error) {
+	v, ok := tlsVersionMap[f.TLSMinVersion]
+	if !ok {
+		return 0, fmt.Errorf("invalid TLS version %q, must be one of: VersionTLS12, VersionTLS13", f.TLSMinVersion)
+	}
```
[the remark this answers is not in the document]
marek-veber: Addressed in 9f10fde. The logic moved to `v2/internal/config` and is now covered by table-driven tests in `vars_test.go` for valid (VersionTLS12/VersionTLS13) and invalid (VersionTLS11, TLSabcd, empty) values. Should be safe to resolve.

matthchr (approved):

PR #5625, opened 2026-08-17
Refactor policy context 

theunrepentantgeek:
## What this PR does

Refactors @andreidorin-oprea's work from PR #5608 to pass ReconcilePolicy through as a direct parameter instead of tunnelling through the existing context.Context that's passed around.

Currently sharing for comparison and discussion only.

gif


andreidorin-oprea:
Thank you for picking this up and doing the refactor yourself — and apologies for the slow reply, I was on vacation and didn't see your comment on #5608 until now.

You're right about the context tunnelling, and I like where this landed. Passing `ReconcilePolicies` explicitly makes the dependency visible at the call site, and moving the type into `pkg/common/annotations` with `ForAnnotation` as a method reads better than the free functions I had.

I've rebased #5609 — the consumer — on top of this to check it holds up in practice. The port was small, which I think is a good sign for the design:

- `reconcilers.ReconcilePolicyFromContext(ctx).AllowsModify()` → `reconcilePolicies.Effective.AllowsModify()`
- `reconcilers.ReconcilePolicyForAnnotation(ctx, ann)` → `policies.ForAnnotation(ann)`
- threading `reconcilePolicies` through the `next()` calls

`go build ./...`, `go vet ./...`, the unit tests (including your new `annotations` and `extensions` ones) and all three databasewatcher CRUD recordings are green on top of it.

One thing to flag: this branch is based on `87d1ac441`, which predates the #5607 merge, so it doesn't have the databasewatcher group and is currently 11 commits behind `main`. I rebased it onto `main` locally to do the above and all 8 commits replayed with no conflicts, so it should be painless whenever you want to refresh it.

Happy either way on the mechanics — land this and I'll rebase #5609 onto `main` afterwards, or if you'd rather not carry it, I'm glad to take the commits back onto #5608 and finish it there. Just let me know which you prefer, and whether you'd like me to close #5608 in favour of this.

docs/hugo/content/contributing/extension-points/post-reconciliation-checker.md
```diff
@@ -71,6 +71,7 @@ func (ex *ResourceExtension) PostReconcileCheck(
     resourceResolver *resolver.Resolver,
     armClient *genericarmclient.GenericClient,
     log logr.Logger,
+    reconcilePolicies annotations.ReconcilePolicies,
```
matthchr: minor: terminology here is `ReconcilePolicies` which I think is a bit weird because there aren't really a collection of policies, there is a single policy.

Admittedly, this is giving the hierarchy too, but it still resolves to a single policy so wondering if we want this to be:

```suggestion
    reconcilePolicy annotations.ReconcilePolicy,
```

(or similar)

edit: could be `ResolvedReconcilePolicy` or something if we wanted.

v2/internal/reconcilers/generic/generic_reconciler.go
```diff
@@ -425,6 +431,45 @@ func (gr *GenericReconciler) mergeReconcilePolicy(ctx context.Context, log logr.
...
+	}
+}
+
+// namespaceReconcilePolicy returns the reconcile-policy annotation of the given namespace, and whether
+// the namespace could be read at all.
+func (gr *GenericReconciler) namespaceReconcilePolicy(
+	ctx context.Context,
+	log logr.Logger,
+	namespace string,
+) (string, bool) {
```
matthchr: Maybe just return the error here? Caller can/should log it?

v2/pkg/common/annotations/reconcile_policy.go
```diff
@@ -26,6 +28,49 @@ const (
 	ReconcilePolicyDetachOnDelete = ReconcilePolicyValue("detach-on-delete")
 )
 
+// ReconcilePolicies are the reconcile policies in effect during a reconcile, none of which a resource
```
matthchr: minor: Don't mention "resource extension" here. That's about how this is used, nothing to do with what it _is_.

v2/internal/reconcilers/generic/generic_reconciler.go
```diff
@@ -425,6 +431,45 @@ func (gr *GenericReconciler) mergeReconcilePolicy(ctx context.Context, log logr.
 			"policyAnnotation", policyStr,
 		)
 	}
+
+	// What a resource in this namespace carrying no annotation of its own would get. Unlike the policy
```
matthchr: Can't this result in some funky reported policies?

Namespace failed to be read, so we end up with `Effective: manage` + `Inherited: skip` even though the resource itself doesn't have a reconcile policy specified (which would suggest that `Inherited` ought to be `manage` since that's what the resource has...

My expectation is that this is a rare case... I am not sold that it makes sense to handle the namespace error differently here compared to above.

v2/internal/reconcilers/generic/generic_reconciler_test.go
```diff
@@ -0,0 +1,244 @@
...
+	"github.com/Azure/azure-service-operator/v2/internal/config"
+	"github.com/Azure/azure-service-operator/v2/internal/util/kubeclient"
+	"github.com/Azure/azure-service-operator/v2/pkg/common/annotations"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime"
+	"github.com/Azure/azure-service-operator/v2/pkg/genruntime/conditions"
+)
+
+// Inherited decides whether a resource other than the one being reconciled may be touched, so unlike the
+// effective policy it has to fail closed
+func Test_MergeReconcilePolicy_givenAnnotations_returnsExpectedPolicies(t *testing.T) {
```
matthchr: very minor:

```suggestion
func Test_MergeReconcilePolicy_GivenAnnotations_ReturnsExpectedPolicies(t *testing.T) {
```

v2/internal/reconcilers/generic/generic_reconciler_test.go
```diff
@@ -0,0 +1,244 @@
...
+	_ logr.Logger,
+	_ record.EventRecorder,
+	_ genruntime.MetaObject,
+	policies annotations.ReconcilePolicies,
+) error {
+	r.updateStatusPolicies = &policies
+	return nil
+}
+
+func Test_CreateOrUpdate_passesResolvedPoliciesToSelectedPath(t *testing.T) {
```
matthchr:
```suggestion
func Test_CreateOrUpdate_PassesResolvedPoliciesToSelectedPath(t *testing.T) {
```

v2/pkg/genruntime/extensions/postreconciliation_checker_test.go
```diff
@@ -0,0 +1,69 @@
...
+	armClient *genericarmclient.GenericClient,
+	log logr.Logger,
+	policies annotations.ReconcilePolicies,
+	next PostReconcileCheckFunc,
+) (PostReconcileCheckResult, error) {
+	r.received = policies
+	return next(ctx, obj, owner, resourceResolver, armClient, log, policies)
+}
+
+func TestCreatePostReconciliationChecker_passesPoliciesToExtension(t *testing.T) {
```
matthchr:
```suggestion
func TestCreatePostReconciliationChecker_PassesPoliciesToExtension(t *testing.T) {
```

theunrepentantgeek:
Appreciate the feedback @andreidorin-oprea; I'll progress this PR and get it merged in as timely a manner as possible.

v2/pkg/common/annotations/reconcile_policy.go
```diff
@@ -26,6 +28,49 @@ const (
...
+
+	policy, _ := ParseReconcilePolicy(annotation, r.Default)
+	return policy
+}
+
+// ParseReconcilePolicy parses provided reconcile policy.
+// defaultPolicyValue is read from DEFAULT_RECONCILE_POLICY env variable or set to 'manage' when missing
+func ParseReconcilePolicy(policy string, defaultReconcilePolicy ReconcilePolicyValue) (ReconcilePolicyValue, error) {
+	// policy is read from CR annotation, if it's empty it being read from defaultReconcilePolicy
+	switch policy {
```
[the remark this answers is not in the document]
theunrepentantgeek: Updated.

v2/internal/reconcilers/generic/generic_reconciler_test.go
```diff
@@ -0,0 +1,244 @@
...
+			namespaceMissing:  true,
+			expectedEffective: annotations.ReconcilePolicyManage,
+			expectedInherited: annotations.ReconcilePolicySkip,
+		},
+		// The namespace we cannot read may be the one that says to leave everything alone
+		"Unreadable namespace, nothing annotated": {
+			namespaceMissing:  true,
+			expectedEffective: annotations.ReconcilePolicyManage,
+			expectedInherited: annotations.ReconcilePolicySkip,
+		},
```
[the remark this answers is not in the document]
theunrepentantgeek: Done.

docs/hugo/content/contributing/extension-points/post-reconciliation-checker.md
theunrepentantgeek: After an offline discussion, I've renamed the existing struct to try and improve clarity/meaning:

``` go
// ResolvedReconcilePolicies are the resolved reconcile policies in effect during a reconcile, none of which a
// resource extension can work out on its own. Read them with Effective or ForAnnotation rather than combining them.
type ResolvedReconcilePolicies struct {
	// Effective is the policy for the resource being reconciled, resolved from its own annotation, Namespace,
	// and Global.
	Effective ReconcilePolicyValue

	// Namespace is the policy for resources in this namespace carrying no annotation of their own, resolved
	// from the namespace annotation or Global.
	Namespace ReconcilePolicyValue

	// Global is the policy resolved from the operator configuration or the built-in default. An unusable
	// annotation falls back to this policy.
	Global ReconcilePolicyValue
}
```

v2/internal/reconcilers/generic/generic_reconciler.go
theunrepentantgeek: Agreed. Done.

v2/pkg/common/annotations/reconcile_policy.go
theunrepentantgeek: Updated.

v2/internal/reconcilers/generic/generic_reconciler.go
theunrepentantgeek: Opted to return with an error if the namespace can't be read - this should be vanishingly rare, and the error may be a bellweather of more extensive problems. The error return will provide a fast retry, so if it's truly a transient error, we'll goal-seek quickly anyway.

v2/internal/reconcilers/generic/generic_reconciler.go
theunrepentantgeek: Done.

v2/internal/reconcilers/generic/generic_reconciler_test.go
theunrepentantgeek: Done.

v2/internal/reconcilers/generic/generic_reconciler_test.go
theunrepentantgeek: Done.

v2/pkg/genruntime/extensions/postreconciliation_checker_test.go
theunrepentantgeek: Done.

v2/pkg/common/annotations/reconcile_policy.go
```diff
@@ -26,6 +28,53 @@ const (
...
+	Namespace ReconcilePolicyValue
+
+	// Global is the policy resolved from the operator configuration or the built-in default. An unusable
+	// annotation falls back to this policy.
+	Global ReconcilePolicyValue
+}
+
+// ForAnnotation returns the reconcile policy for a resource other than the one being reconciled, given
+// that resource's own reconcile-policy annotation.
+func (r ResolvedReconcilePolicies) ForAnnotation(annotation string) ReconcilePolicyValue {
```
theunrepentantgeek: 🤖 TheUnrepentantRobot says:

This helper applies the namespace policy resolved for the resource currently being reconciled, but its contract says it can be used for any "resource other than the one being reconciled." The docs narrow that only to a resource belonging to the same operator, and one operator can watch multiple namespaces. An extension following that guidance could therefore evaluate a resource using the wrong namespace policy. Can we either constrain and name/document this explicitly as same-namespace-only (for example, `ForSameNamespaceAnnotation`), or make the target namespace part of the resolution contract?

theunrepentantgeek: 🤖 TheUnrepentantRobot says:

Addressed in `3c4afe8f`. `ResolvedReconcilePolicies` now records the namespace for which its namespace policy was resolved, and `ForAnnotation(string)` has been replaced with `ForResource(metav1.Object)`. `ForResource` verifies that the target resource is in that namespace and returns an error on a mismatch before resolving its annotation. Tests cover same-namespace fallback, valid and invalid annotations, cross-namespace rejection, and matching empty namespaces; the extension documentation now directs callers to use the checked resource API.

theunrepentantgeek:
@andreidorin-oprea, I've made more changes to this branch to improve clarity and catch potential issues, would be interested in your thoughts if you have time to review.

v2/pkg/common/annotations/reconcile_policy.go
```diff
@@ -26,6 +31,65 @@ const (
...
+	// Effective is the policy for the resource being reconciled, resolved from its own annotation, Namespace,
+	// and Global.
+	Effective ReconcilePolicyValue
+
+	// Namespace is the policy for resources in this namespace carrying no annotation of their own, resolved
+	// from the namespace annotation or Global.
+	Namespace ReconcilePolicyValue
+
+	// NamespaceName is the namespace from which Namespace was resolved.
+	NamespaceName string
```
andreidorin-oprea: Nit: NamespaceName seems a bit odd across the other fields

andreidorin-oprea (approved):

v2/internal/reconcilers/generic/generic_reconciler.go
```diff
@@ -397,79 +405,84 @@ func (gr *GenericReconciler) mergeReconcilePolicy(
 	ctx context.Context,
 	log logr.Logger,
 	obj genruntime.MetaObject,
-) annotations.ReconcilePolicies {
-	// If no configured default policy, we set it to 'manage'
-	defaultReconcilePolicy := gr.Config.DefaultReconcilePolicy
-	if defaultReconcilePolicy == "" {
-		defaultReconcilePolicy = annotations.ReconcilePolicyManage
+) (annotations.ResolvedReconcilePolicies, error) {
+	// If no global policy is configured, use the built-in default of 'manage'.
```
matthchr, line 409: minor: comment here is possibly inaccurate now since DefaultReconcilePolicy is configurable so the default may not actually be `manage`.

Though I note the comment had this issue before you changed this as well.

v2/internal/reconcilers/generic/generic_reconciler.go
```diff
@@ -397,79 +405,84 @@ func (gr *GenericReconciler) mergeReconcilePolicy(
...
-	if namespaceRead {
-		inheritedPolicy, err = annotations.ParseReconcilePolicy(namespacePolicyStr, defaultReconcilePolicy)
-		if err != nil {
-			log.V(Verbose).Info(
-				"Namespace has an unusable reconcile policy. Applying default policy instead",
-				"policyAnnotation", namespacePolicyStr,
-			)
-		}
+	// Resolve the namespace policy independently because resources other than the one being reconciled may
+	// rely on it. Unlike the effective policy, this fails closed when the namespace cannot be read.
```
matthchr: > Unlike the effective policy, this fails closed when the namespace cannot be read.

Not accurate anymore? When the namespace can't be read we error. When the namespace value can't be parsed is how we end up here, I think? Which is not really "fails closed" because we apply the global policy instead (that's not closed, I think in the alternative impl this ended up `skipped`?

Not saying it needs to be skipped now but I think the doc should be corrected.

v2/pkg/common/annotations/reconcile_policy.go
matthchr: Could be `PolicyNamespace` or something? Not sure that's better.

matthchr (approved):
LGTM, had a few nits.

Feel free to ping me to reapprove if you fix the nits

v2/pkg/common/annotations/reconcile_policy.go
theunrepentantgeek: I've made it a little better but there's still room for improvement if someone has a good idea.

v2/internal/reconcilers/generic/generic_reconciler.go
theunrepentantgeek: This comment is talking about whether the default is configured or not, so I'm a little confused.

v2/internal/reconcilers/generic/generic_reconciler.go
theunrepentantgeek: Updated doc. 

And your understanding is correct - we only end up here if the value configured on the namespace is invalid.

v2/internal/reconcilers/generic/generic_reconciler.go
matthchr: Uhh yeah you're right. I am confused, ignore me. This comment is correct.

matthchr (approved):

PR #5628, opened 2026-08-17
fix: improve CRD comparison perf

matthchr:
## What this PR does

Fixes the issue discussed in this PR comment https://github.com/Azure/azure-service-operator/pull/5588#discussion_r3669880256

## Checklist

- [x] this PR contains tests

### DIFF
--- a/v2/internal/crdmanagement/crd_installation_instruction.go
+++ b/v2/internal/crdmanagement/crd_installation_instruction.go
@@ -15,7 +15,6 @@ type DiffResult string
 
 const (
 	NoDifference      = DiffResult("NoDifference")
-	SpecDifferent     = DiffResult("SpecDifferent")
 	VersionDifferent  = DiffResult("VersionDifferent")
 	MetadataDifferent = DiffResult("MetadataDifferent")
 )
@@ -24,8 +23,6 @@ func (i DiffResult) DiffReason(crd apiextensions.CustomResourceDefinition) strin
 	switch i {
 	case NoDifference:
 		return fmt.Sprintf("No difference between existing and goal CRD %q", makeMatchString(crd))
-	case SpecDifferent:
-		return fmt.Sprintf("The spec was different between existing and goal CRD %q", makeMatchString(crd))
 	case VersionDifferent:
 		return fmt.Sprintf("The version was different between existing and goal CRD %q", makeMatchString(crd))
 	case MetadataDifferent:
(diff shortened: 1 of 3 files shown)


v2/internal/crdmanagement/manager.go
```diff
@@ -245,46 +258,25 @@ func (m *Manager) FindMatchingCRDs(
...
+		for _, comparator := range comparators {
+			diffReason := NoDifference
+			if !comparator.Compare(existingCRD, goalCRD) {
+				diffReason = comparator.DifferenceReason
 			}
-		}
 
-		if equal {
-			matching[goalCRD.Name] = goalCRD
+			comparisonResults = append(
```
theunrepentantgeek, line 275: Shouldn't we skip adding the reason if the comparator indicates there's no difference?

I'm thinking that returning `NoDifference, NoDifference, VersionDifferent` is kinda weird, but that's what we can get.

Returning just `VersionDifferent` might be clearer ... would probably need something at the end to return `NoDifference` if all the comparators returned true.

v2/internal/crdmanagement/crd_installation_instruction.go
```diff
@@ -15,7 +15,6 @@ type DiffResult string
 
 const (
 	NoDifference      = DiffResult("NoDifference")
-	SpecDifferent     = DiffResult("SpecDifferent")
```
theunrepentantgeek, line 18: Why skip comparing the spec?

theunrepentantgeek (approved):
A couple questions but otherwise looks sound.

v2/internal/crdmanagement/crd_installation_instruction.go
matthchr: Spec wasn't being compared before either, I just forgot to remove the reason. If you see, we were only doing "VersionDifferent" and "MetadataDifferent" comparisons below.

v2/internal/crdmanagement/manager.go
matthchr: Somewhat torn here. On the one hand, the surface you suggest is definitely a bit cleaner. On the other hand, having a single entry for each comparator is "nice" because it makes the comparator -> difference mapping quite clear. Though "comparison by index" is more awkward than "by name" so maybe if we really want/need more clarity there (such as if a future refactor makes the DiffReason's dynamic where a single comparator can return a dynamic diff reason and multiple comparisons could report the same diff reason, so diff reason alone wasn't enough to indicate which comparator was used) we would just add a name rather than requiring an index match.

I've made the change suggested though as currently it seems cleaner.

theunrepentantgeek (approved):

theunrepentantgeek (approved):

PR #5672, opened 2026-09-07
Add pprof options to the code generator

theunrepentantgeek:
🤖 TheUnrepentantRobot says:

## What this PR does

The ASO generator can consume substantial CPU and memory, but previously required custom instrumentation to investigate performance. This adds persistent `--cpu-prof <file>` and `--memory-prof <file>` options to every generator subcommand.

Both options can be used together. CPU sampling covers the selected command's execution, while memory profiling writes the cumulative `allocs` profile after a final GC. Profiles are finalised on success, returned errors, and panics, and startup failures clean up partial files without hiding the original error.

gif

## Checklist

- [x] this PR contains tests


v2/tools/generator/root.go
```diff
@@ -19,22 +20,23 @@ import (
 
 // Execute kicks off the command line
 func Execute() {
-	cmd, err := newRootCommand()
+	cmd, profiler, err := newRootCommand()
```
matthchr, line 23: minor: returned as a pair, passed as a pair -- could put into their own structure so we have:
```
cmd, err := newRootCommand()
```

but `cmd` has `.cmd` (ok maybe `cmd` variable naming is bad -- `result.cmd`) and `result.profiler`?

v2/tools/generator/profiling_test.go
```diff
@@ -0,0 +1,276 @@
...
+	profiler := newProfiler()
+	profiler.memoryProfilePath = memoryPath
+
+	g.Expect(profiler.start()).To(Succeed())
+
+	var allocations [][]byte
+	deadline := time.Now().Add(200 * time.Millisecond)
+	for time.Now().Before(deadline) {
+		allocations = append(allocations, allocateProfileBytes(1024))
+		if len(allocations) > 1_000 {
```
matthchr, line 56: minor, style: 
```suggestion
		if len(allocations) > 1000 {
```

maybe? Not sure we use 1_000 style elsewhere in the repo?

matthchr: Applies throughout this file.

v2/tools/generator/profiling_test.go
```diff
@@ -0,0 +1,276 @@
...
+//nolint:paralleltest // Keep profiler lifecycle tests serialized for consistency.
+func TestProfiler_InvalidMemoryPathReportsCreateFailureAndLeavesHandlesNil(t *testing.T) {
+	g := NewGomegaWithT(t)
+
+	tempDir := t.TempDir()
+	invalidMemoryPath := filepath.Join(tempDir, "missing", "memory.pprof")
+	profiler := newProfiler()
+	profiler.memoryProfilePath = invalidMemoryPath
+
+	g.Expect(profiler.start()).To(MatchError(ContainSubstring("creating memory profile")))
```
matthchr, line 146: Test seems very similar to one above it (it also asserts nil for all these fields I believe?)

PR #5673, opened 2026-09-07
feat: approve the private endpoint connection a SharedPrivateLink opens

andreidorin-oprea:
## What this PR does

A `SharedPrivateLink` is how a database watcher reaches a target, or its own data store, over a private
endpoint. Azure opens the private endpoint connection behind it pending the resource owner's approval and,
unlike a `PrivateEndpoint`, reports nothing about that connection back on the link: `properties.status` on
`sharedPrivateLinkResources` is `null` before approval and still `null` after it, on every deployed API
version. ASO therefore marked the link `Ready` while the connection behind it was still pending, and a
watcher could not collect from a target whose public network access was disabled.

`SharedPrivateLinkExtension` now reads the connection on the resource the link points at, which is the only
place its state is reported, and completes it:

| Connection state | Result |
| --- | --- |
| `Approved` | The link goes `Ready`. |
| `Pending` | The connection is approved, and the link stays short of `Ready` until Azure confirms it. |
| `Rejected`, `Disconnected` | Somebody decided this deliberately, so the link reports it and leaves it be. |
| anything else | Waited on rather than acted upon. |

Approving is long running on some providers, so the operation is followed across reconciles through a poller
resume token, exactly as a watcher start already is. A link therefore approves once, however long the
provider takes.

### Special notes

**Why this approves when `PrivateEndpoint` deliberately does not.** Approval only happens when the link names
its resource as a Kubernetes resource rather than by ARM ID. That reference is what supplies the API version
its connections are read and written with - the link's own `databasewatcher` API version is not valid for
`Microsoft.Sql` or `Microsoft.Kusto` - and it means the operator is completing a handshake on a resource the
user already manages with it, rather than approving on some third party's behalf. A link to a bare ARM ID
keeps the readiness it has always had: with no API version its connections cannot even be read, so reporting
anything about them would be a guess.

Before approving, the link refuses when the resource is managed by another operator, or asks for a credential
the link cannot prove is its own, and when either resource's reconcile policy forbids modification. Azure
asks for `Owner` on the resource to approve a connection against it, which the credential that created the
link need not have, so `AuthorizationFailed` is reported as approval being required and who has to give it
rather than as an error on every reconcile. The comparisons a `Target` already made against its `Watcher` are
now shared as `differingOperator` and a widened `differingCredential`, so both extensions refuse a foreign
resource on the same terms.

**This check calls ARM on every reconcile**, which the extension guidance would rather avoid. There is no
alternative here: the status field that should carry this is never populated, and an extension holding an
`ARMMetaObject` has no type-specific accessor for another resource's private endpoint connections. One list
per reconcile is the cost of reading the only source of truth.

**A watcher running when a connection is approved still needs restarting** before it uses that connection, as
[the documentation](https://learn.microsoft.com/en-us/azure/azure-sql/database-watcher-manage) notes. This PR
does not restart one: stopping a running watcher interrupts collection for every target on it, so it seemed
better to raise than to decide here.

Nothing sequences the two, so this is not limited to watchers that were already running. `TargetExtension`
starts a watcher as soon as it finds it `Stopped` while reconciling a `Target`, without consulting any
`SharedPrivateLink`, and the two post-reconcile checks run on different resources with no ordering between
them. Whether the start lands before or after the approval is therefore a race on first creation too.
Sequencing it would mean gating the start on link readiness inside `TargetExtension`, which is a change to
merged behaviour and a policy call I would rather leave to a maintainer than make here.

**A watcher can now be held stopped.** Review asked whether anyone would want ASO not to start a watcher.
They would: a watcher stopped deliberately - to stop paying for collection, or to keep targets provisioned
but not yet watched - was started again on the next reconcile of any of its targets, and nothing short of
`reconcile-policy: skip` could hold it. `Watcher.operatorSpec.autoStart` opts out. Unset still starts, so
nothing changes for anyone today, and a start already in flight is still followed to its end.

An Azure support case was opened on the `null` status (`2609040050002363`). Microsoft confirmed the field
cannot be read from the watcher at all - "from the watcher side this is not possible", the API "is not
behaving as expected" - and recommends exactly what this PR does: list the private endpoint connections on
the target resource and read `properties.privateLinkServiceConnectionState.status` there.

Their suggested way to pick the connection out is its name prefix, which they note is "current observed
behavior, not a documented contract". This PR matches on `properties.privateEndpoint.id` instead: Azure names
the managed private endpoint after the link exactly, where the connection appends a GUID, so a link cannot
claim a connection belonging to a link whose name it merely prefixes. The recording shows both -
`privateEndpoints/asotest-spl-owpjxz` against
`privateEndpointConnections/asotest-spl-owpjxz-3f8ab5eb-...`.

The product group has since settled it: the documentation will be updated to describe the current API
behaviour, rather than the field being populated. So reading the connection on the target resource is not a
workaround waiting on a fix - it is the supported way to learn whether a shared private link carries traffic.
Were the field ever populated anyway, this extension keeps working unchanged.

### Testing

`Test_DatabaseWatcher_Target_v20241001preview_CRUD` creates a `SharedPrivateLink`, so its recording is
regenerated. Re-recorded live, it captures Azure naming the connection after the link and a GUID of its own,
opening it `Pending` carrying the link's request message, and reporting it `Approved` once the operator has
approved it. That test also carried a comment saying `Status.Status` "isn't populated yet"; it is never
populated, and the assertion now says so and checks the connection on the server instead.

Alongside it are unit tests for the connection-name match, including a link whose name merely prefixes
another's, and `httptest` tests covering approve-once-and-follow-the-operation, every connection state, a
connection Azure has yet to open, a resource with no ARM ID, an ARM ID reference, both reconcile policies, a
foreign operator, a differing credential, and an unauthorized approval. States and statuses are covered in
the casing ARM documents and in another, and `autoStart` both for the opt-out reading nothing from ARM and
for an opt-out arriving while a start is still running.

## Checklist

- [x] this PR contains tests


docs/hugo/content/guide/database-watchers.md
theunrepentantgeek, line 1: Is there any situation where someone would want ASO _not_ to automatically start the watcher? That is, are there circumstances where this behaviour needs to be turned off?

If so, we can use `$operatorSpecProperties` in the config file to add a property to the `operatorSpec`.

v2/api/databasewatcher/customizations/shared_private_link_extensions.go
```diff
@@ -0,0 +1,421 @@
...
+			fmt.Sprintf(
+				"cannot tell which private endpoint connection on %s this link opened, so it has to be approved there",
+				resource.GetName(),
+			),
+		), nil
+	}
+
+	connection := &connections[0]
+
+	switch state := connection.state(); state {
```
theunrepentantgeek, line 160: ARM states are _supposed_ to be case-invariant (e.g. `Succeeded`) but there have been cases where different casing is returned (e.g., `canceled` instead of `Canceled`).

It's probably good practice to do case-insensitive comparisons on the state here, just in case.

v2/api/databasewatcher/customizations/shared_private_link_extensions.go
```diff
@@ -0,0 +1,421 @@
...
+// foreignPrivateLinkResource reports why the link cannot approve a connection on the resource it points at,
+// and is empty when it can. Approving writes to that resource with the link's credential, so the operator
+// and the credential behind both have to be the same.
+func foreignPrivateLinkResource(
+	link *databasewatcher.SharedPrivateLink,
+	resource genruntime.ARMMetaObject,
+) (string, bool) {
+	if differingOperator(link, resource) {
+		return fmt.Sprintf(
+			"cannot approve the private endpoint connection on %s, which is managed by the operator in %s while this link is managed by the operator in %s",
```
theunrepentantgeek:
```suggestion
			"cannot approve the private endpoint connection on %s; it is managed by the operator in %s while this link is managed by the operator in %s",
```

theunrepentantgeek (commented):
Some minor comments, but overall this looks really good.

v2/api/databasewatcher/customizations/shared_private_link_extensions.go
andreidorin-oprea: Applied, and the same to the credential message just below it, which read the same way.

v2/api/databasewatcher/customizations/shared_private_link_extensions.go
andreidorin-oprea: Done. `state()` now folds the status onto the casing ARM documents before anything compares it, so both the switch and `allApproved` are case-insensitive, and a state ARM does not document still reaches the message reporting it verbatim.

The watcher status in `target_extensions.go` had the same `== "Running"` / `!= "Stopped"` comparisons, so it shares the fold - a `running` there would have started an already-running watcher a second time. Say the word if you would rather that went in its own PR.

docs/hugo/content/guide/database-watchers.md
andreidorin-oprea: Yes - a watcher somebody stopped deliberately. Today the next reconcile of any of its targets reads it `Stopped` and starts it again, so nothing short of `reconcile-policy: skip`, which suppresses everything else too, can hold one stopped. Collection costs while a watcher runs, and there are targets you may want provisioned before they are watched, so it is worth being able to hold.

Added it as you suggested: `$operatorSpecProperties` on `Watcher` gives `operatorSpec.autoStart`. Unset still starts, so nothing changes for anyone today. The check sits after the resume-token block and before the watcher is read, so opting out costs no ARM call, and a start already in flight is still followed to its end rather than leaving its poller token on the target. Documented in the guide under "Starting the watcher", with tests for the opt-out and for an opt-out arriving mid-start.

v2/api/databasewatcher/customizations/shared_private_link_extensions.go
theunrepentantgeek: Including it here makes sense.

andreidorin-oprea:
@theunrepentantgeek all three of your comments are addressed in the four commits since your review: the connection-state comparisons are case-insensitive (and the watcher status too, per your reply), the message suggestion is applied, and `Watcher.operatorSpec.autoStart` gives the opt-out you asked about - unset still starts, so nothing changes for anyone today.

Checks are green; `integration-tests` is skipped. Happy to re-run the live recording if you want that before merge. Anything else you would like changed, or is this good to approve?

theunrepentantgeek (approved):

theunrepentantgeek:
> Anything else you would like changed, or is this good to approve?

Looks good to me, I've approved and added to our merge queue. 

Thanks for taking the time to iterate on this one, I think it came out to a good result.

PR #5677, opened 2026-09-08
feat: support unhealthyPodEvictionPolicy on the PodDisruptionBudget

Ranexia04:
## What this PR does

Adds `podDisruptionBudget.unhealthyPodEvictionPolicy` to the operator's PodDisruptionBudget. Only rendered when set, so existing releases keep the Kubernetes default of `IfHealthyBudget`. Setting it to `AlwaysAllow` lets a not-Ready replica be evicted during a node drain instead of blocking it, per the [Kubernetes drain recommendation](https://kubernetes.io/docs/tasks/run-application/configure-pdb/#pdb-unhealthy-pod-eviction-policy).

Without it, a single crash-looping operator replica takes `disruptionsAllowed` to 0 and blocks a node drain indefinitely, even though the pod is Not Ready and serving nothing.

### Special notes

The value defaults to `""` and is guarded by a falsy check in the template, following the existing `azureOperatorMode` pattern in this chart, so `helm template` output is byte-identical to `main` with default values and `controller:validate-helm` still matches the kustomize output.

There is no unit-test harness for this chart, so the checklist item for tests is left unticked rather than deleted — happy to add coverage if you'd like it somewhere.

## Checklist

- [ ] this PR contains tests

### DIFF
--- a/v2/charts/azure-service-operator/templates/policy_v1_poddisruptionbudget_azureserviceoperator-pdb.yaml
+++ b/v2/charts/azure-service-operator/templates/policy_v1_poddisruptionbudget_azureserviceoperator-pdb.yaml
@@ -12,4 +12,7 @@ spec:
     matchLabels:
       control-plane: controller-manager
   minAvailable: {{ .Values.podDisruptionBudget.minAvailable }}
+  {{- with .Values.podDisruptionBudget.unhealthyPodEvictionPolicy }}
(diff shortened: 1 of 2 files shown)


Ranexia04:
@microsoft-github-policy-service agree company="Mapped"

theunrepentantgeek (approved):

matthchr (approved):

PR #5686, opened 2026-09-09
fix: dataprotection hybrid migration release reporting

theunrepentantgeek:
## What this PR does

The `dataprotection` group already generates hybrid `v20*` API packages, but its migration release was omitted from the release-reporting map. Register v2.21.0 as the migration release so the new-style versions report when they actually became available instead of inheriting the original `v1api` release dates.

This also regenerates the API reference indexes, updating `v20230101` and `v20231101` to show `v2.21.0` in the Supported From column.

### Special notes

Controller generation and unit tests pass. The full controller quick-check task reaches lint but is blocked by an unrelated existing `unparam` finding in `v2/test/perf/perf_helpers_test.go:201`.

gif

## Checklist

- [ ] this PR contains tests


matthchr (approved):
Just correcting docs - LGTM

PR #5688, opened 2026-09-09
feat: Add ARO-HCP (HcpOpenShiftCluster) support for v20260901preview

marek-veber:
## Summary

- Adds ASO support for the ARO-HCP (hosted control plane) resources under the `redhatopenshift` group, API version `v20260901preview`: `HcpOpenShiftCluster`, `HcpOpenShiftClustersNodePool`, and `HcpOpenShiftClustersExternalAuth` — generated from `v2/azure-arm/redhatopenshift.yaml` in the usual way, alongside a hand-written extension (`hcp_open_shift_cluster_extension.go`) for pre-reconcile checks and CSR-based admin-credential/kubeconfig export.
- Adds the hand-written controller CRUD integration test required by ASO's own contributing guide, backed by a recorded cassette covering the full lifecycle (create, patch, delete, verified-absent) of all three resources. Verified against live Azure and replays in ~85s with no credentials needed.
- Adds unit tests for the hand-written extension (`hcp_open_shift_cluster_extension_test.go`), covering the CSR/kubeconfig handling, secrets logic, and the pre-reconcile delete-blocking check.
- Genericizes the `v20260901preview` sample manifests (placeholder subscription ID and `aso-sample-*` names, matching the rest of the sample suite) and narrows `samples_tester.go`'s exclusion from the whole `redhatopenshift` group down to the `v1api` (service-principal-based) and `v20260901preview` (pending self-contained sample prerequisites) directories individually.

Recording the integration test surfaced two real behaviors worth calling out for review:
- `HcpOpenShiftClustersNodePool` (and `HcpOpenShiftClustersExternalAuth`) `spec.azureName` is capped at 15 characters by the ARM API, unlike the cluster's 54-character limit — the test sets an explicit short `AzureName` rather than defaulting to the generated Kubernetes object name.
- Azure rejects deleting a cluster's last remaining node pool while the cluster still exists (`Forbidden: The last node pool can not be deleted from a cluster.`), so the test deletes the external-auth and the cluster explicitly and relies on ARM's cascade to remove the node pool.

## Test plan

- [x] `go build ./...` and `go vet` clean
- [x] `go test ./api/redhatopenshift/...` — unit tests for generated types and the hand-written extension
- [x] `Test_RedHatOpenShift_HcpOpenShiftCluster_v20260901preview_CRUD` recorded against live Azure (full create/patch/delete lifecycle for cluster, node pool, and external-auth)
- [x] Same test replays cleanly from the committed cassette with no Azure credentials
- [x] Cassette inspected for redaction correctness (no subscription/tenant IDs, no auth headers, no secrets)
- [ ] Sample suite (`Test_Samples_CreationAndDeletion`) coverage for `v20260901preview` — documented as a follow-up in `testdata/redhatopenshift-hcp/README.md`; the samples aren't yet self-contained (they reference external VNet/NSG/identity/Key Vault prerequisites) so they remain excluded from automatic sample testing


v2/api/redhatopenshift/customizations/hcp_open_shift_cluster_extension.go
```diff
@@ -0,0 +1,281 @@
...
+		return nil, nil
+	}
+
+	id, err := genruntime.GetAndParseResourceID(typedObj)
+	if err != nil {
+		return nil, err
+	}
+
+	var adminCredentials string
+	if requestedSecrets.Contains(adminCredentialsKey) {
```
theunrepentantgeek, line 111: I'm a little worried about the need for this to call a poller and wait for the credential to be available, but our design for the `KubernetesSecretExporter` extension doesn't allow much of a choice.

What's the typical wall-clock time for this? Is it seconds, tens-of-seconds, minutes?

theunrepentantgeek (approved):

v2/api/redhatopenshift/customizations/hcp_open_shift_cluster_extension.go
marek-veber: The current CRUD cassette does not request `operatorSpec.secrets.adminCredentials`, so it does not provide an end-to-end measurement for this path. The implementation performs the initial LRO poll immediately, then polls every 15 seconds, with a 5-minute timeout. That means the normal completion latency is the RP operation time plus at most one 15-second polling interval; the 5 minutes is only a safety bound for an unhealthy/stuck request. I will follow up with a recording that exercises admin credential export so we can report an observed wall-clock value.

theunrepentantgeek (approved):

v2/api/redhatopenshift/customizations/hcp_open_shift_cluster_extension.go
marek-veber: I checked the live ASO controller logs for both provisioned clusters. The complete admin-credential export path (from “Getting Kubernetes secrets for export” to “Successfully retrieved Kubernetes secrets for export”, including key generation, ARM LRO polling, kubeconfig processing, and Secret preparation) took:
* `capz-test-20260917-194140`: 04:25:44.530–04:26:00.961, **16.43 s**;
* `capz-test-20260918-061357`: 04:27:39.304–04:27:57.517, **18.21 s**.

So the observed wall-clock time here was 16–18 seconds, with the 5-minute timeout remaining only as a stuck-operation safety bound.

marek-veber: I repeated the measurement with the new timing logs enabled on a live HCP cluster.

For `capz-test-20260918-211825/mv1-tests`:

- Initial `BeginRequestAdminCredential` ARM call: **153.84014 ms**
- ARM LRO polling: **15.833898195 s**
- Total admin-credential request: **15.987769985 s**

The exporter log timestamps were `19:40:38.686538Z` to `19:40:54.674278Z`, confirming approximately **15.99 seconds wall-clock time**. This is consistent with the earlier observed range of **16–18 seconds** across the provisioned clusters. The five-minute timeout remains only a safety bound for a stuck request.

theunrepentantgeek (approved):

v2/api/redhatopenshift/customizations/hcp_open_shift_cluster_extension.go
theunrepentantgeek: Good to know, thanks for getting these measurements. This _shouldn't_ have much of an effect on ASO's reconciliation of other resource types, but we will need to be mindful of this if we get bug reports about slow updates.

theunrepentantgeek:
CI is failing on a code formatting issue; you should be able to fix by running `task format-code` and committing the result.

theunrepentantgeek (approved):

theunrepentantgeek:
Thanks @marek-veber - much appreciate all the effort that's gone into getting this across the line.

PR #5694, opened 2026-09-10
feat: Add AKS api 2026-05-01

markjgardner:
Import ContainerService 2026-03-01 and migrate the group to hybrid versioning so AKS agent pools can use the upstream AzureContainerLinux OS SKU enum value.

## What this PR does

The primary driver behind this PR is to enable support for `osSKU==AzureContainerLinux` for AKS nodepools. That particular OS variant became available with containerservice api version 2026-03-01.

### Special notes

The legay --> hybrid conversion generated a large number of new files as part of this commit. If there is a better way of handling this, please let me know.

nervous

## Checklist

- [x] this PR contains tests


markjgardner:
I feel a little less worried after seeing @theunrepentantgeek changes in #5685. This looks very similar, if just greater scale. Let me know how you guys want to handle this.

theunrepentantgeek:
Yes, the move to hybrid versioning adds a lot of files - this is expected. 

_However_, there's a hard 2MiB limit for each CRD (due to an etcd limitation) and I suspect that the additional versions included by moving `containerservice` to hybrid versioning will be a problem.

Let's see how the build works.

theunrepentantgeek:
> Let's see how the build works.

We're getting close to the line, but we may be Ok.

Before this PR, our CI reports this:

```
[controller:check-crd-size] Checking CRD sizes...
[controller:check-crd-size] Checking CRDs against size limit of 1.5MiB (1572864 bytes)...
[controller:check-crd-size] [WRN] Large CRD (>1MiB): containerservice.azure.com_managedclusters.yaml - 1.03MiB
[controller:check-crd-size]
[controller:check-crd-size] [INF] CRD Size Summary:
[controller:check-crd-size] Total CRDs: 293
[controller:check-crd-size] Large CRDs (>1MiB): 1
[controller:check-crd-size] Oversized CRDs (>1.5MiB): 0
[controller:check-crd-size]
[controller:check-crd-size] [INF] All CRDs are within the size limit
```

After this PR, it's significantly larger, but still under the 1.5MiB limit:

```
[controller:check-crd-size] [WRN] Large CRD (>1MiB): containerservice.azure.com_managedclusters.yaml - 1.39MiB
```

This check is only an approximation of what Kubernetes does, so we need to test to be sure.

v2/azure-arm/containerservice.yaml
```diff
@@ -103,8 +103,13 @@ groupModelConfiguration:
         OIDCIssuerProfile: $.Status.OidcIssuerProfile.IssuerURL
         PrincipalId: $.Status.Identity.PrincipalId
     ManagedCluster_Spec:
+      AutoScalerProfile:
+        $conversionStrategy: manual
```
theunrepentantgeek, line 107: Is this necessary in the new version?

Resorting to manual conversion is intended to be vanishingly rare, and I'd be surprised to see it necessary more than once in any specific resource conversion chain.

The handwritten code in `conversion_overrides.go` looks very much like the code that would be generated natively - it doesn't have any of the special handling usually associated with the need for manual conversion.

Recommendation: Remove these manual overrides and test whether they're actually required (we generate unit tests to ensure conversions work, so if there's a problem those should find it).

v2/internal/testcommon/samples_tester.go
```diff
@@ -57,6 +57,14 @@ var exclusions = []*regexp.Regexp{
 	// file present so scripts/v2/check_samples.py is satisfied, but skip it in the samples test.
 	regexp.MustCompile(`cdn/v.*20210601/.*_profilesendpoint.yaml`),
 
+	// The containerservice TrustedAccessRoleBinding sample needs a Microsoft.MachineLearningServices
+	// workspace as its source resource, and AML now rejects the workspace payload our sample refs
+	// produce with "ValidationError: Missing dependent resources in workspace json" (it wants
+	// additional dependencies such as Application Insights). Older containerservice versions only
+	// still pass because their recordings predate that service-side change. Keep the sample file
+	// present so scripts/v2/check_samples.py is satisfied, but skip it in the samples test.
+	regexp.MustCompile(`containerservice/v20260301/.*_trustedaccessrolebinding.yaml`),
```
theunrepentantgeek, line 66: How complex would it be to upgrade the sample to work with the new requirement?

theunrepentantgeek (commented):
This looks very good - just a couple queries to answer before we consider merging.

v2/azure-arm/containerservice.yaml
markjgardner: The problem is that nested type names zig-zag across versions:

version | name
-- | --
v20251002preview | ManagedClusterLoadBalancerProfileOutboundIPPrefixes
v1api20250801 | ManagedClusterLoadBalancerProfile_OutboundIPPrefixes
v20260301 | ManagedClusterLoadBalancerProfileOutboundIPPrefixes

Is there a different/better way to handle those?

v2/internal/testcommon/samples_tester.go
markjgardner: It's not a big change, but it requires updating the shared AML sample and re-recording the `machinelearningservices` samples too, which feels like it belongs in its own PR (out of scope for what I wanted to do here). But you can also tell me I'm being lazy and I would accept that.

v2/azure-arm/containerservice.yaml
theunrepentantgeek: Yuck. I wonder what it is in the upstream Swagger that's causing that. :cry:

You should be able to align things by adding `$nameInNextVersion` to the YAML config, something like this:

``` yaml
2025-08-01:
    ManagedClusterLoadBalancerProfile_OutboundIPPrefixes:
        $nameInNextVersion: ManagedClusterLoadBalancerProfile_OutboundIPPrefixes
2025-10-01preview:
    ManagedClusterLoadBalancerProfileOutboundIPPrefixes:
        $nameInNextVersion: ManagedClusterLoadBalancerProfile_OutboundIPPrefixes
```

Part of the reason this might look odd is a choice we made when constructing the graph of resource versions: preview versions convert _back_ to the previous stable version, not _forward_.

If needed, you can use `$nameInNextVersion` for properties too.

v2/internal/testcommon/samples_tester.go
theunrepentantgeek: No, that's fair enough. Please create an issue so that we don't forget to do it though.

markjgardner: Created #5696

v2/azure-arm/containerservice.yaml
markjgardner: Unfortunately that doesn't work either. I asked copilot to investigate why and this is what it came back with:

> **Root cause: two graphs that disagree.**
> 
> 
> `$nameInNextVersion` isn't a rename — `searchForRenamedType` *walks the package-level link graph* looking for the target name. But conversions are generated from a *separate, per-resource* graph. Importing 2026-03-01 creates a new `v1api20250801/storage/compat` package (8 unrelated types — `AppMonitoring`, `HostedSystemProfile`; absent on `main`), and the two graphs then diverge:
> 
> 
> 
> - **Package-level** (`GroupConversionGraphBuilder`) — compat is derived from `…/v1api20250801/storage`, so stage 1 links `2025-08-01 → compat` and drops 2025-08-01 from the candidate list. By the time `previewReferencesConvertBackward` runs, `refs[i-1]` is **compat**. Empirically: `preview → compat → 2026-03-01`.
> 
> - **Resource-level** (`ResourceConversionGraphBuilder`) — operates on *type names*, and `ManagedCluster` has no compat variant, so it correctly gives `preview → v1api20250801/storage`.
> 
> 
> 
> So the generator emits `AssignProperties_From_ManagedCluster_Spec(v1api20250801/storage.ManagedCluster_Spec)` — correct — then resolves the nested rename against the *other* graph, which never visits `v1api20250801/storage`, the one package holding the underscored name. Hence the error.
> 
> 
> No YAML config can bridge that, because the walk simply never reaches the package that has the type. The zig-zag is the trigger, but the compat package is the actual blocker — and it only appears because of this PR.

This seems plausible to me, because I notice that depending on which value I use for `$nameInNextVersion` it will fail at step 59 (Repair property bag serialization for properties that skip resource or object versions) for v20260301 compatible names or step 63 (Inject property assignment functions AssignFrom() and AssignTo() into resources and objects) for v20250801 compatible names. I'm happy to dig in deeper here, but I think I'll need some hand holding to figure this out.

Looking at the API specs, it looks like the containerservices api changed these property definitions from nested definitions (inferred type names) to references with explicitly defined type names between [v20250801](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/stable/2025-08-01/managedClusters.json#L4745) and [v20251002preview](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/preview/2025-10-02-preview/managedClusters.json#L10756).

theunrepentantgeek: I'll dig into this and work out what's going on.

FWIW, in case you're curious, my first step will be to run the code generator with `--debug containerservice`. This dumps stage-by-stage diagnostics into a temp folder. Super useful, but not for the faint hearted!

theunrepentantgeek:
>The legacy --> hybrid conversion generated a large number of new files as part of this commit. If there is a better way of handling this, please let me know.

This PR doesn't actually move to hybrid versioning - I just spotted this comment in commit `aedaba4c`:

> Migrating the group to hybrid mode instead would duplicate every existing version, pushing the ManagedCluster CRD to 2.00MiB and over the 1.5MiB Kubernetes/etcd limit, which broke envtest CRD installation. ADR-2025-05 already calls this out for ManagedCluster. With legacy mode the CRD is 1.39MiB.

Looks like Copilot recognized the problem, and left things configured for Legacy mode to avoid breaking things.

v2/azure-arm/containerservice.yaml
theunrepentantgeek: **TL,DR:** I've found and fixed a bug in how conversions are hooked up. Now this is fixed, no `manual` conversions are needed, just a block of renames. I've pushed a fix to this PR.

This was a doozy of a subtle bug - easy enough to explain now I've found it, but a real head-scratcher to untangle even though I wrote the original code!

In the way of these things, it's a confluence of two different decisions interacting in an unexpected way.

The first is that when constructing the conversion graph, we link preview versions to the chronologically earlier version.

The second is that we introduced the `compat` subpackage to fix a problem known as [skipping properties](https://azure.github.io/azure-service-operator/design/adr-2023-09-skipping-properties/).

Together, these resulted in the preview versions being linked to the `compat` subpackage, instead of the correct version.

I've fixed this in PR #5703 - the writeup there has some nice diagrams if you want to wrap your head around the problem.

theunrepentantgeek (approved):

v2/azure-arm/containerservice.yaml
markjgardner: Thanks for all the hard work @theunrepentantgeek!

markjgardner:
> > The legacy --> hybrid conversion generated a large number of new files as part of this commit. If there is a better way of handling this, please let me know.
> 
> This PR doesn't actually move to hybrid versioning - I just spotted this comment in commit `aedaba4c`:
> 
> > Migrating the group to hybrid mode instead would duplicate every existing version, pushing the ManagedCluster CRD to 2.00MiB and over the 1.5MiB Kubernetes/etcd limit, which broke envtest CRD installation. ADR-2025-05 already calls this out for ManagedCluster. With legacy mode the CRD is 1.39MiB.
> 
> Looks like Copilot recognized the problem, and left things configured for Legacy mode to avoid breaking things.

Yeah, I got really nervous when I saw the initial commit on this PR and the number of new files it produced. So I went back and did some nudging to get to a more compact implementation which was [aedaba4](https://github.com/Azure/azure-service-operator/pull/5694/commits/aedaba4c35260ec9c1aa90f34f741b3fb9fea7be). I didn't update the PR description, sorry about that.

markjgardner:
@matthchr Could you approve this PR? Since @theunrepentantgeek merged in his bug fix and removed the manual conversions his approval no longer counts. Much appreciated.

v2/azure-arm/containerservice.yaml
```diff
@@ -223,9 +243,65 @@ groupModelConfiguration:
         $importConfigMapMode: optional
       ObjectId:
         $importConfigMapMode: optional
-    ScaleProfile:
-      Autoscale:
-        $conversionStrategy: manual
-    ScaleProfile_STATUS:
-      Autoscale:
-        $conversionStrategy: manual
+  2026-03-01:
```
matthchr: Latest merged API version is 2026-06-01. Can we use that instead?

matthchr: or, if that's not available in our submodule (it might not be), 2026-05-01?

markjgardner: Looks like 20260501 is the most recent supported version. I wasn't sure how aggressive you were willing to accept when upgrading so I picked the lowest version that supported my base requirement.

I ran a quick test and the CRD size is going to be pretty tight with 20260501:
| version | bytes |
|--------|-------|
| v20260301 | 1461188 |
| v20260501 | 1482054 |

I'm re-recording now and I'll push the update when it finishes.

theunrepentantgeek (approved):

matthchr (approved):

PR #5707, opened 2026-09-17
feat: make the leader election lease timings configurable

andreidorin-oprea:
## What this PR does

Adds `--lease-duration`, `--renew-deadline` and `--retry-period`, and exposes them in the Helm chart under `leaderElection`. The defaults are unchanged (15s / 10s / 2s), so this is a no-op unless you set them.

The timings are currently hardcoded in `SetupControllerManager`. On a large cluster the operator cannot finish its initial informer sync inside the 10s renew deadline: the renewal is starved while every watched kind is listed for the first time, the lease expires, and the manager exits.

```
I0916 23:50:18.86  "Starting Controller" (x199)  "Starting EventSource" (x453)
E0916 23:52:18.52  "failed to start manager" err="leader election lost"
```

Because both replicas do this, the standby takes the lease, reaches the same point and exits too. On our production cluster - 199 installed CRDs, ~40k ASO resources - the operator flapped across six pods for roughly 50 minutes after a `v2.20 -> v2.21` upgrade and only settled once a pod happened to win the race. It needed manual intervention on three consecutive releases. There is no way to work around this short of editing the deployment after Helm has run.

### Special notes

- The timings reach `crdmanagement.NewLeaderElector` through `ctrlOptions`, so the CRD leader elector picks them up too. I rewrote the comment above them: it said they were pinned to the defaults so the CRD elector could reuse them, and the reuse is still the reason, but "to the default" no longer holds.
- Validated in `main.go` right after flag parsing. Without that, `leaseDuration <= renewDeadline` only fails later inside client-go's leader elector, after startup is already underway.
- `v2/config/manager/manager.yaml` gets the same three args so `controller:validate-helm` keeps matching the chart against the kustomize output.
- The docs entry lands under **Common mistakes** next to the other pod-startup symptoms, and the existing "Resource reports webhook error when applied" entry now links to it, since ASO pod restarts are already named there as a cause.
- It documents the failover trade-off: a leader that is killed rather than shut down gracefully is only replaced after `leaseDuration`. Graceful restarts are unaffected, because `LeaderElectionReleaseOnCancel` is already set.

gif

## Checklist

- [x] this PR contains tests


docs/hugo/content/guide/diagnosing-problems/_index.md
```diff
@@ -60,6 +60,32 @@ spec:
...
+
+The error looks like this, shortly after the pod has started its controllers:
+```
+E0916 23:52:08.898297  1 main.go:47] "failed to start manager" err="leader election lost"
+```
+
+The pod exits, the standby replica takes over, reaches the same point and exits too, so the operator never
+finishes starting. This happens when the initial informer sync is slower than the lease renewal deadline: the
+renewal is starved while every watched kind is listed for the first time, the lease expires, and the manager
+stops. It gets more likely the more CRDs are installed and the more resources exist in the cluster.
```
matthchr: minor: may be worth clarifying that this should only happen for large cluster with a large number of resources and CRDs

v2/charts/azure-service-operator/templates/apps_v1_deployment_azureserviceoperator-controller-manager.yaml
```diff
@@ -61,6 +61,9 @@ spec:
         {{- end }}
         - --health-addr=:8081
         - --enable-leader-election
+        - --lease-duration={{ .Values.leaderElection.leaseDuration }}
```
matthchr: minor, curious @theunrepentantgeek's thoughts too, should these be `--leader-lease-duration`, `--leader-renew-deadline`? Or maybe `--election` instead of `--leader`? Basically some prefix to make it clear these values are all about leader election at the cmdline flags level.

matthchr (approved):
Looks good to me, but had a few minor comments.

v2/charts/azure-service-operator/templates/apps_v1_deployment_azureserviceoperator-controller-manager.yaml
theunrepentantgeek: I agree, tweaking the names slightly would be a good idea to improve clarity. I'd suggest these:

`--leader-lease-duration`
`--leader-renew-deadline`
`--leader-retry-interval`

* Common prefix ties them together, and indicates what they're configuring.
* _interval_ instead of _period_ because the later may be interpreted as the time allowed for all retries.

theunrepentantgeek (commented):
Overall looks good. 

Should we consider changing the defaults from 15s/10s/2s to something larger?

v2/charts/azure-service-operator/templates/apps_v1_deployment_azureserviceoperator-controller-manager.yaml
matthchr: FWIW on the last, `RetryPeriod` is the controller-runtime field, so matching that exactly (with a prefix like `--leader-retry-period` has some value. I don't have strong opinions though if @andreidorin-oprea or @theunrepentantgeek you feel strongly another way. My opinions are slightly stronger on the "I think a prefix here in the cmdline arg would be good", so as long as we get that I am happy with either name.

theunrepentantgeek: Matching the config from controller-runtime is a strong argument; I'm comfortable with using `--leader-retry-period`; the common prefix is much more important.

andreidorin-oprea: Done in f4a21e3 - they are now `--leader-lease-duration`, `--leader-renew-deadline` and `--leader-retry-period`. Went with `retry-period` to keep the controller-runtime field name, per the consensus above. The Helm values are already nested under `leaderElection`, so those are unchanged.

docs/hugo/content/guide/diagnosing-problems/_index.md
andreidorin-oprea: Done in f4a21e3. The entry now opens with it rather than mentioning it at the end:

> This only affects large clusters - ones with many installed CRDs, many ASO resources, or both. Smaller clusters start well inside the default lease timings and never hit it.

andreidorin-oprea:
@theunrepentantgeek on raising the defaults - I'd rather leave them at 15s/10s/2s.

They have held up for us across more than two years of running ASO and repeatedly growing what it manages, and this is the first time we have needed to move them. A default that only one end of the size range needs is better expressed as a knob than as a new default, and raising it slows failover for every operator that was never affected.

Happy to be argued out of it, but it feels like a change that wants a benchmark behind it rather than a guess - something that shows where the initial sync actually crosses the renew deadline as CRD and resource counts grow. I don't have that data, only the one production data point in the PR description.

Review feedback is pushed in f4a21e3: flags renamed with the `--leader-` prefix, and the docs entry now leads with the large-cluster scoping.

matthchr (approved):

theunrepentantgeek:
> They have held up for us across more than two years of running ASO and repeatedly growing what it manages, and this is the first time we have needed to move them.

Good to know. I had a background concern that maybe this was a common failure that warranted changing the defaults. Sounds like its not, so I've hit merge.

Thanks again for the contributions.

